PR #2395, opened 2018-11-23
Increase resource limits

mrIncompetent:
This PR will very aggressively increase the container resource limits.
The purpose of the limits is to avoid containers going "crazy".
For actual resource planning, the current resource request should be used (Which will be maintained by the VerticalPodAutoscaler - which will come as a follow up).

This PR also fixes the DeepEqual comparisons of Deployments & StatefulSets, by increasing the `MaxDepth` of the DeepEqual check from 10 to 20. This will now correctly identify changes of the container resources and trigger finally an update.

**Release note**:
```release-note
[ACTION REQUIRED] Resource limits for control plane containers have been increased. This might require additional resources for the seed cluster
```

### DIFF
--- a/api/pkg/resources/apiserver/deployment.go
+++ b/api/pkg/resources/apiserver/deployment.go
@@ -24,8 +24,8 @@ var (
 			corev1.ResourceCPU:    resource.MustParse("100m"),
 		},
 		Limits: corev1.ResourceList{
-			corev1.ResourceMemory: resource.MustParse("1Gi"),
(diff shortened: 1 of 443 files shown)


alvaroaleman (commented):
Generally speaking I am fine with the change, but your release not should contain an `[Action Required]` and some info about how much resources are no required by default per cluster.

And we _really_ need a VPA :S

alvaroaleman (approved):

PR #2396, opened 2018-11-23
Ensure all resources have resource limit and request set

mrIncompetent:
Add missing resource requests and limits.
Additionally a check has been implemented in the tests to make sure all Containers have resource requests and limits.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/resources/machinecontroller/deployment.go
+++ b/api/pkg/resources/machinecontroller/deployment.go
@@ -21,7 +21,7 @@ var (
 		},
 		Limits: corev1.ResourceList{
 			corev1.ResourceMemory: resource.MustParse("512Mi"),
-			corev1.ResourceCPU:    resource.MustParse("100m"),
+			corev1.ResourceCPU:    resource.MustParse("2"),
 		},
 	}
 )
--- a/api/pkg/resources/machinecontroller/deployment.go
+++ b/api/pkg/resources/machinecontroller/deployment.go
@@ -3,14 +3,27 @@ package machinecontroller
 import (
 	"fmt"
 
-	"github.com/kubermatic/kubermatic/api/pkg/resources/apiserver"
-
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/apiserver"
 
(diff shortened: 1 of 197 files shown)


api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -3,16 +3,29 @@ package machinecontroller
...
 
+var (
+	controllerResourceRequirements = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
+			corev1.ResourceCPU:    resource.MustParse("25m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+			corev1.ResourceCPU:    resource.MustParse("100m"),
```
alvaroaleman: In the e2e tests the machine-controller temporarily consumes up to four full cpus. We do not usually spawn that many instances with Kubermatic, but I would definitely set this up to 1 core. Unfortunatelly I have no idea about how much memory it needs

api/pkg/resources/metrics-server/deployment.go
```diff
@@ -3,17 +3,30 @@ package metricsserver
...
 
+var (
+	defaultResourceRequirements = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
+			corev1.ResourceCPU:    resource.MustParse("25m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+			corev1.ResourceCPU:    resource.MustParse("250m"),
```
alvaroaleman: What for does it need so much cpu?

mrIncompetent: I reduced it to `150m`. It uses `~26m` on a medium size cluster (60 nodes | 4k pods)

api/pkg/resources/machinecontroller/deployment.go
mrIncompetent: I increased the CPU limit to `2`

alvaroaleman (approved):

PR #2401, opened 2018-11-25
add missing vault to helm container

mrIncompetent:
Add missing vault to helm container

**Release note**:
```release-note
NONE
```

### DIFF
--- a/containers/helm/Dockerfile
+++ b/containers/helm/Dockerfile
@@ -1,8 +1,13 @@
 FROM alpine
 
-RUN apk add --no-cache ca-certificates curl
+RUN apk add --no-cache -U ca-certificates unzip jq bash curl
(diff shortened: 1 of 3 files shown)


containers/helm/Dockerfile
```diff
@@ -1,8 +1,13 @@
 FROM alpine
 
-RUN apk add --no-cache ca-certificates curl
+RUN apk add --no-cache -U ca-certificates wget unzip jq bash curl
```
alvaroaleman: I think the wget is not required here :P

mrIncompetent: Removed with https://github.com/kubermatic/kubermatic/pull/2401/commits/bed8b8e45d82f9dc7ac9ecd61013642a18381089

Though i have not created a new image as it would not have any impact

alvaroaleman (dismissed):

alvaroaleman (approved):

PR #2403, opened 2018-11-26
add TLS verification for dex

zreigz:
Since Dex is configured with TLS there is a need to add the CA cert to initiate trust.\
Depends on https://github.com/kubermatic/secrets/pull/204

**Documentation**:
https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/user-cluster-authentication-via-dex.md

**Release note**:
```release-note
Kubermatic controller can now use TLS verification
[Action required] to enable the TLS verification for kubermatic-controller-manager the users must set `-feature-gates=OpenIDAuthPlugin=true` and provide `-oidc-issuer-url`, `-oidc-issuer-client-id` and `--oidc-ca-file` when running the controller.
[Action required] user need CA bundle file with a correct chain of certificates
```


api/pkg/resources/resources.go
```diff
@@ -657,6 +661,21 @@ func getClusterCAFromLister(name string, cluster *kubermaticv1.Cluster, lister c
 	return certs[0], key, nil
 }
 
+// GetDexCAFromLister returns the Dex CA from the lister
+func GetDexCAFromLister(lister corev1lister.SecretLister) ([]*x509.Certificate, error) {
+	caCertSecret, err := lister.Secrets("oauth").Get(DexCASecretName)
```
alvaroaleman: I am not sure if its ideal to hardcode the namespace and name of the Secret. IMHO we should make that a config option that can default to `"oauth/"DexCASecretName`

Also I think it may make sense to check on startup if that secret exists and has the correct key and fail if not, to make errors there obvious

zreigz: Thanks, I like this idea. It can be configurable and we can check if the correct secret/key exists during the validation.

zreigz:
@alvaroaleman @p0lyn0mial @mrIncompetent PTAL

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -4,7 +4,9 @@ import (
 	"flag"
 	"fmt"
 	"net/url"
+	"strings"
 
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
p0lyn0mial: nit: we usually group imports, for example https://github.com/kubermatic/kubermatic/blob/81ba106b702ddd39baf3ba32e87160bce2ff7336/api/pkg/controller/rbac/sync_project.go

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -13,12 +15,16 @@ import (
...
 	kubeinformers "k8s.io/client-go/informers"
+	certutil "k8s.io/client-go/util/cert"
 
 	"github.com/kubermatic/kubermatic/api/pkg/features"
 
 	backupcontroller "github.com/kubermatic/kubermatic/api/pkg/controller/backup"
 )
 
+// DexCAFileName CA file name
+const DexCAFileName = "ca.pem"
```
p0lyn0mial: make it private, also consider changing the name to `caBundle.pem` just to highlight it can contain a certificate chain.

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -89,9 +96,10 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
 	flag.StringVar(&c.monitoringScrapeAnnotationPrefix, "monitoring-scrape-annotation-prefix", "monitoring.kubermatic.io", "The prefix for monitoring annotations in the user cluster. Default: monitoring.kubermatic.io -> monitoring.kubermatic.io/port, monitoring.kubermatic.io/path")
 	flag.StringVar(&rawFeatureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for various features.")
-	flag.StringVar(&c.oidcCAFile, "oidc-ca-file", "", "The path to the certificate for the CA that signed your identity provider’s web certificate.")
+	flag.BoolVar(&c.skipOidcTLSVerify, "skip-oidc-tls-verify", false, "The flag used to skip TLS verification against dex.")
 	flag.StringVar(&c.oidcIssuerURL, "oidc-issuer-url", "", "URL of the OpenID token issuer. Example: http://auth.int.kubermatic.io")
 	flag.StringVar(&c.oidcIssuerClientID, "oidc-issuer-client-id", "", "Issuer client ID")
+	flag.StringVar(&c.oidcDexSecretCAName, "oidc-dex-secret-ca-name", "oauth/dex-ca", "The secret name which contains CA bundle for TLS verification against dex. This flag is optional. The default value is oauth/dex-ca.")
```
p0lyn0mial: add a note that the secret key must be `ca.pem` (or `caBundle.pem` if you decide to change it)

api/pkg/resources/resources.go
```diff
@@ -142,6 +143,10 @@ const (
 	ApiserverEtcdClientCertificateSecretName = "apiserver-etcd-client-certificate"
 	//ApiserverFrontProxyClientCertificateSecretName is the name for the secret containing the apiserver's client certificate for proxy auth
 	ApiserverFrontProxyClientCertificateSecretName = "apiserver-proxy-client-certificate"
+	// DexCASecretName is the name of the secret that contains the Dex CA bundle
+	DexCASecretName = "dex-ca"
```
p0lyn0mial, line 149: how do we feel about the fact that the secret we would create in the user's cluster will have a different name than the one provided via the flag ?

api/pkg/resources/certificates/dex-ca.go
```diff
@@ -0,0 +1,52 @@
...
+		ca, err := getCA()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get dex public keys: %v", err)
+		}
+
+		if se.Data == nil {
+			se.Data = map[string][]byte{}
+		}
+
+		if _, exists := se.Data[resources.DexCAFileName]; exists {
```
p0lyn0mial: shouldn't we use `dataCAKey` instead ?

api/pkg/resources/certificates/dex-ca.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type dexCAGetter func() ([]*x509.Certificate, error)
+
+// GetDexCACreator returns a function to create a secret containing a CA bundle with the specified name
+func GetDexCACreator(name, commonName, dataCAKey string, getCA dexCAGetter) func(resources.SecretDataProvider, *corev1.Secret) (*corev1.Secret, error) {
```
p0lyn0mial: `commonName` arg is not used.

api/pkg/resources/apiserver/deployment.go
```diff
@@ -232,6 +232,11 @@ func Deployment(data resources.DeploymentDataProvider, existing *appsv1.Deployme
 					MountPath: "/etc/kubernetes/pki/front-proxy/ca",
 					ReadOnly:  true,
 				},
+				{
+					Name:      resources.DexCASecretName,
+					MountPath: "/etc/kubernetes/dex/ca",
```
p0lyn0mial, line 237: this path is different than the one passed to `--oidc-ca-file` is this intentional ?

api/pkg/resources/data.go
```diff
@@ -185,21 +192,26 @@ func (d *TemplateData) OIDCAuthPluginEnabled() bool {
...
 }
 
-// OIDCIssuerClientID return the issuer client ID
+// OIDCIssuerClientID returns the issuer client ID
 func (d *TemplateData) OIDCIssuerClientID() string {
 	return d.oidcIssuerClientID
 }
 
+// OIDCDexSecretCAName returns the secret name for Dex CA bundle
+func (d *TemplateData) OIDCDexSecretCAName() string {
```
p0lyn0mial: I think that this method is not used anywhere, what's the purpose of having it then ?

p0lyn0mial (changes requested):

api/cmd/kubermatic-controller-manager/options.go
zreigz: done

api/cmd/kubermatic-controller-manager/options.go
zreigz: done

api/cmd/kubermatic-controller-manager/options.go
zreigz: done

api/pkg/resources/resources.go
zreigz: the flag purpose was to hide hardcoded path to the secret. Because this secret name is for the cluster we should use a constant name like for others to keep consistency

api/pkg/resources/certificates/dex-ca.go
zreigz: yes, done

api/pkg/resources/certificates/dex-ca.go
zreigz: yes, removed

api/pkg/resources/apiserver/deployment.go
zreigz: this path indicates just directory where secret was mounted. The `--oidc-ca-file` indicates the full path for CA bundle file

api/pkg/resources/data.go
zreigz: yes, removed

zreigz:
PTAL

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -116,6 +122,18 @@ func (o controllerRunOptions) validate() error {
 		if len(o.oidcIssuerClientID) == 0 {
 			return fmt.Errorf("%s feature is enabled but \"oidc-issuer-client-id\" flag was not specified", OpenIDAuthPlugin)
 		}
+
+		if !o.skipOidcTLSVerify {
```
zreigz: I think the dex secret exists independently on this flag, I should skip this check and move outside the TLS verification

mrIncompetent:
Is there a specific reason why we load the CA from the Kubernetes API instead of just mounting it?

It feels weird having now 2 sources for the controller input:
- Files passed via flag (ConfigMaps or Secrets, which get mounted to the Pod)
- Files which exist as secret and need to be accessed via the Kubernetes API

zreigz:
@mrIncompetent @alvaroaleman @p0lyn0mial PTAL

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -166,6 +174,24 @@ func (o controllerRunOptions) validate() error {
 	return nil
 }
 
+// validateDexSecretWithCABundle
+func (o controllerRunOptions) validateDexSecretWithCABundle() error {
+
+	f, err := os.Open(o.oidcCAFile)
```
alvaroaleman, line 182: This should be closed in a `defer` statement

api/pkg/controller/backup/backup_controller.go
```diff
@@ -380,9 +384,15 @@ func (c *Controller) sync(key string) error {
...
-	serviceLister corev1lister.ServiceLister
+	cluster          *kubermaticv1.Cluster
+	secretLister     corev1lister.SecretLister
+	serviceLister    corev1lister.ServiceLister
+	oidcCABundleFile string
+}
+
+// GetDexCA returns the chain of public certificates for TLS verification of the Dex
+func (d *secretData) GetDexCA() ([]*x509.Certificate, error) {
+	return resources.GetDexCAFromFile(d.oidcCABundleFile)
```
alvaroaleman: Why was all this code needed? The backup controller does not seem to be using the`DexCA` anywhere or am I missing it?

alvaroaleman: Ah, because we need to satisfy the `SecretDataProvider` interface. Wouldn't it be easier to just implement this as a `return nil, errors.New("not implemented)` and not carry around the `CABundleFile` since we don't use it all?

Not quite sure about this one, what do others think?

api/pkg/controller/cluster/resources_test.go
```diff
@@ -181,7 +184,19 @@ func TestSecretV2CreatorsKeepAdditionalData(t *testing.T) {
 	}
 	serviceLister := listerscorev1.NewServiceLister(serviceIndexer)
 
-	templateData := resources.NewTemplateData(cluster, dc, "", secretLister, nil, serviceLister, "", "", "", resource.Quantity{}, "", "", false, false, "", nil, "", "", "")
+	file, err := ioutil.TempFile("", "caBundle.pem")
```
alvaroaleman, line 188: Always close file handlers, please

api/pkg/resources/kubeconfig_test.go
```diff
@@ -87,6 +87,9 @@ func (fake *fakeDataProvider) GetRootCA() (*triple.KeyPair, error)        { retu
 func (fake *fakeDataProvider) GetOpenVPNCA() (*ECDSAKeyPair, error)       { return &ECDSAKeyPair{}, nil }
 func (fake *fakeDataProvider) InClusterApiserverURL() (*url.URL, error)   { return &url.URL{}, nil }
 func (fake *fakeDataProvider) InClusterApiserverAddress() (string, error) { return "", nil }
+func (fake *fakeDataProvider) GetDexCA() ([]*x509.Certificate, error) {
+	return []*x509.Certificate{{}}, nil
```
alvaroaleman: Nit: You can just `return nil, nil` here

api/pkg/resources/data.go
```diff
@@ -180,6 +182,11 @@ func NewTemplateData(
 	}
 }
 
+// GetDexCA returns the chain of public certificates of the Dex
+func (d *TemplateData) GetDexCA() ([]*x509.Certificate, error) {
+	return GetDexCAFromFile(d.oidcCAFile)
```
alvaroaleman, line 187: I think it makes more sense to store the whole `CA` in memory instead of reading it from disk every time it is needed

api/pkg/controller/backup/backup_controller.go
zreigz: yes, I agree I will implement your solution and test it but I also didn't notice the usage for this function in backup controller

api/pkg/resources/data.go
zreigz: I agree I've put the whole cert into memory. I can revert it if others disagree

api/pkg/resources/kubeconfig_test.go
zreigz: done

api/pkg/controller/cluster/resources_test.go
zreigz: done

api/cmd/kubermatic-controller-manager/options.go
zreigz: done

zreigz:
PTAL

api/pkg/resources/resources.go
```diff
@@ -657,6 +667,22 @@ func getClusterCAFromLister(name string, cluster *kubermaticv1.Cluster, lister c
 	return certs[0], key, nil
 }
 
+// GetDexCAFromFile returns the Dex CA from the lister
+func GetDexCAFromFile(caBundleFilePath string) ([]*x509.Certificate, error) {
+	if dexCACerts == nil {
```
alvaroaleman: Is this concurrency-safe? Wouldn't it be cleaner to read the file in the `cmd` and pass it as an arg into the Controller when creating the controller?

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -166,6 +176,31 @@ func (o controllerRunOptions) validate() error {
 	return nil
 }
 
+// validateDexSecretWithCABundle
+func (o controllerRunOptions) validateDexSecretWithCABundle() error {
```
mrIncompetent: Is the function name still correct?
As it does not validate a secret anymore

api/pkg/resources/data.go
mrIncompetent: My 2 cents:
I think its nice to have it read it from filesystem everytime (It will land in the page cache anyway and we won't cause any disk IO). It avoid having to restart the controller on file change.

Though it differs then from our docker-secret handling - where we load it during startup into memory.

api/cmd/kubermatic-controller-manager/options.go
zreigz: done

api/pkg/resources/resources.go
zreigz: I have reverted this for reading from filesystem everytime to avoid having to restart the controller on file change.

alvaroaleman (approved):

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -151,6 +156,11 @@ func (o controllerRunOptions) validate() error {
 		return fmt.Errorf("moniotring-scrape-annotation-prefix is undefined")
 	}
 
+	// Validate OIDC CA file
+	if err := o.validateCABundle(); err != nil {
```
p0lyn0mial, line 160: shouldn't we validate `caBundle` only if the `OpenIDAuthPlugin` feature has been enabled ?

api/hack/run-controller.sh
```diff
@@ -22,6 +22,7 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
   -backup-container=../config/kubermatic/static/backup-container.yaml \
   -cleanup-container=../config/kubermatic/static/cleanup-container.yaml \
   -docker-pull-config-json-file=../../secrets/seed-clusters/dev.kubermatic.io/.dockerconfigjson \
+  -oidc-ca-file=../../secrets/seed-clusters/dev.kubermatic.io/caBundle.pem \
```
p0lyn0mial, line 25: `OpenIDAuthPlugin` is optional, I'd rather have all (`oidc-issuer-url` and `oidc-issuer-client-id` ) or nothing. I'm fine with having it `disabled` on the local machine. A small comment explaining how to enable it would be appreciated. Does it make sense ?

p0lyn0mial (changes requested):
Please also make sure that the release note is correct.

api/cmd/kubermatic-controller-manager/options.go
zreigz: We create a cluster dex-ca secret and mount it in cluster api deployment independently even if the `OpenIDAuthPlugin` feature is disabled. It makes  `-oidc-ca-file` mandatory right now. In this case validation is OK

api/hack/run-controller.sh
zreigz: Like in the previous comment it is a mandatory flag. Otherwise, we had to make this configurable: https://github.com/kubermatic/kubermatic/pull/2403/files#diff-a4282b98d1211368af8912d690d2794eR203 and depends on the feature `OpenIDAuthPlugin` we would have to create secret and mount it to the cluster API deployment

api/cmd/kubermatic-controller-manager/options.go
p0lyn0mial: we could make it optional in the deployment too, for example https://github.com/kubermatic/kubermatic/blob/master/config/kubermatic/templates/kubermatic-controller-manager-dep.yaml#L60

so what do you think it is not worth it ?

zreigz: Yes, but creating cluster secrets in the controller it is not configurable in our source code right now. We create all secrets from the list: https://github.com/kubermatic/kubermatic/blob/ac952343cd56f7c0ead6b3f53949959f4d53e888/api/pkg/controller/cluster/resources.go#L196

So secret can exist in the system even if it's not used

zreigz: Let's do this in the separate PR

p0lyn0mial (approved):

api/hack/run-controller.sh
mrIncompetent: @zreigz can you do a follow up PR describing how to get that file?

mrIncompetent (approved):

PR #2408, opened 2018-11-26
Increase resources of prometheus on cloud eu to 4GiB of memory

alvaroaleman:

Fixes prometheus-kubermatic crashlooping

**Documentation**:

**Release note**:

```release-note
NONE
```


xrstf:
Why would we not update the cloud values.yaml in our secrets repo instead? This PR is putting non-build-related configuration into an unexpected place IMHO.

mrIncompetent:
> Why would we not update the cloud values.yaml in our secrets repo instead? This PR is putting non-build-related configuration into an unexpected place IMHO.

I agree with @xrstf here - the values.yaml is the correct place.
The only things we should override in the CI are the commit hashes, which we want to deploy

xrstf:
I've created kubermatic/secrets#201 to replace this.

alvaroaleman:
Valid objection, closing this one

PR #2437, opened 2018-11-27
Support deleting LoadBalancer services on cluster deletion

### LINKED ISSUE #1985: OpenStack clusters are not deleted properly when a LoadBalancer exists in the cluster
When we try to delete an OpenStack cluster that has a Service with the type "LoadBalancer", the cluster hangs at the `kubermatic.io/cleanup-openstack-network` finalizer because the cluster's network can't be deleted before the OpenStack loadbalancer including its pool, listener and security group are deleted.

Either the Service resource needs to be deleted manually before deleting the cluster (which will trigger Kubernetes to delete the LB resources in OpenStack), or the cluster creation logic needs to check for OpenStack resources that are attached to the network and clean these up as well.

bashofmann:

On OpenStack any LoadBalancer services need to be deleted on cluster
deletion, because otherwise the OpenStack network can not be deleted
since there is still a LoadBalancer, Security Group and Floating IP
attached to it.

Fixes #1985

I'm not quite sure, if I'm adding this logic in the right place. Especially using the constant in two different packages is not really nice.
 On the one hand it's fairly related to OpenStack, on the other hand, the function just talks with the Kubernetes API and may be also necessary or at least not hurting for other providers.

**Release note**:

```release-note
Delete OpenStack LoadBalancers on cluster deletion to prevent errors deleting the OpenStack network
```

### DIFF
--- a/api/pkg/controller/cluster/deleting.go
+++ b/api/pkg/controller/cluster/deleting.go
@@ -6,9 +6,11 @@ import (
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	kuberneteshelper "github.com/kubermatic/kubermatic/api/pkg/kubernetes"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
+	"github.com/kubermatic/kubermatic/api/pkg/provider/cloud/openstack"
 
 	"github.com/kubermatic/machine-controller/pkg/node/eviction"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 )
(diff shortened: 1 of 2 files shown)


kdomanski:
I'd prefer to make the `deletingLoadBalancerServices` function provider-agnostic and instead check for the finalizer befor it's call in `cleanupCluster`. @alvaroaleman what do you think?

alvaroaleman:
> I'd prefer to make the deletingLoadBalancerServices function provider-agnostic and instead check for the finalizer befor it's call in cleanupCluster. @alvaroaleman what do you think?

Yes, that would be perfect. Also the finalizer itself should be provider-agnostic

mrIncompetent:
Please have a look at https://app.zenhub.com/workspaces/all-kubermatic--58863654794153994b62af76/issues/kubermatic/kubermatic/2448

It will be cloud provider agnostic and it is in our sprint.

mrIncompetent:
@bashofmann We implemented the backend part in https://github.com/kubermatic/kubermatic/pull/2514
With the difference: We're not automatically adding the finalizer.

Can this PR be closed ?

bashofmann:
@mrIncompetent Yes this can be closed then. Would be nice to have https://github.com/kubermatic/kubermatic/issues/2449 soon then.

PR #2444, opened 2018-11-28
Make fluentd crashes visible

alvaroaleman:
So far we've used a fluentd image that contained a supervisor
and on top of that had fluentds own supervising capabiltiy enabled.

That meant that whenever the fluentd worker crashed it would get
restarted, which made the restartCount of the pod stay at zero
even after the process got repeatedly oom-killed by the kernel.

Fixes #

**Documentation**:

**Release note**:

```release-note
NONE
```

CC @p0lyn0mial @mrIncompetent @xrstf

### DIFF
--- a/config/logging/fluentd/templates/fluentd-daemonset.yaml
+++ b/config/logging/fluentd/templates/fluentd-daemonset.yaml
@@ -21,6 +21,13 @@ spec:
         effect: NoSchedule
       containers:
       - name: fluentd
+        command:
+        - /bin/sh
+        - -c
+        - >-
+            sed -i 's;#!/usr/bin/dumb-init.*;#!/bin/sh;g' /fluentd/entrypoint.sh &&
+            if ! grep -q supervisor /fluentd/entrypoint.sh; then sed -i 's;--gemfile;--no-supervisor --gemfile;g' /fluentd/entrypoint.sh; fi &&
+            exec /fluentd/entrypoint.sh
         image: '{{ .Values.logging.fluentd.image.repository }}:{{ .Values.logging.fluentd.image.tag }}'
         imagePullPolicy: {{ .Values.logging.fluentd.image.pullPolicy }}
         securityContext:


kron4eg (approved):

config/logging/fluentd/templates/fluentd-daemonset.yaml
```diff
@@ -21,6 +21,13 @@ spec:
         effect: NoSchedule
       containers:
       - name: fluentd
+        command:
+        - /bin/sh
+        - -c
+        - >-
+            sed -i 's;#!/usr/bin/dumb-init.*;#!/bin/sh;g' /fluentd/entrypoint.sh &&
+            if ! grep -q supervisor /fluentd/entrypoint.sh; then sed -i 's;--gemfile;--no-supervisor --gemfile;g' /fluentd/entrypoint.sh; fi &&
```
xrstf, line 29: Just my 2 cents, maybe I'm wrong:

* Can we ue `FLUENTD_OPT` to contain `--no-supervisor` instead of `sed`ing the entrypoint script?
* If so, that should be enough to make the errors visible, no? `dumb-init` should die if its children die: ["If your process dies, dumb-init will also die, taking care to clean up any other processes that might still remain."](https://github.com/Yelp/dumb-init#what-dumb-init-does)

xrstf: Ah shit, I'm late to the party.

alvaroaleman: Thats indeed more elegant. I first assumed dumb-init was doing all the restart until I figured fluent has a supervisor on its own. Can you create a follow-up to do that @xrstf ?

xrstf: Can do

xrstf: => #2446

PR #2446, opened 2018-11-28
make fluentd crash configuration more stable

xrstf:
This is a follow-up to #2444. It improves two points:

* The `no-supervisor` flag can be specified by using the `FLUENTD_OPT` variable used by the [entrypoint.sh](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.3/debian-elasticsearch/entrypoint.sh).
* dumb-init itself does not restart processes, so it's not required to remove it from the entrypoint script at all.

Putting the two together allows us to remove the customizations to the entrypoint script alltogether.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/logging/fluentd/templates/fluentd-daemonset.yaml
+++ b/config/logging/fluentd/templates/fluentd-daemonset.yaml
@@ -21,13 +21,6 @@ spec:
         effect: NoSchedule
       containers:
       - name: fluentd
-        command:
-        - /bin/sh
-        - -c
-        - >-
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):

alvaroaleman:
@xrstf Could you also create a PR to have this fix in `release/v2.8`?  Cherry picking wont work because my PR that this PR reverts isn't in there :disappointed:

kron4eg (approved):
oh wow, nice

PR #2461, opened 2018-11-30
make user email addresses case insensitive

### LINKED ISSUE #2452: Adding a member to the project is not case sensitive
**User Story** It turns out that email addresses can be stored in the database in any form. For example as `lukasz@loodse.com` or `Lukasz@loodse.com`.  To make the end user experience better we could change our backend to look up the email addresses in a case-insensitive way. 

**Acceptance criteria** 
 - Inviting `lukasz@loodse.com` or `Lukasz@loodse.com` should work and add exactly the same user to a project.
- Additionally we should change "create user" logic in the backend so that we don't end up with two users in the system (resources)

zreigz:
**What this PR does / why we need it**: fixes problem with users email adress. It makes email address case insensitive.

Fixes #2452

**Documentation**:

**Release note**:
```release-note
NONE
```


alvaroaleman:
I would really like to see this change but I remember @p0lyn0mial mentioning there were some issues at a customer site because they had the same users with different casing. Are those issues resolved @p0lyn0mial ?

p0lyn0mial:
> I would really like to see this change but I remember @p0lyn0mial mentioning there were some issues at a customer site because they had the same users with different casing. Are those issues resolved @p0lyn0mial ?

In the past our system generated duplicates. This issue was resolved. No-one has complained since then so one would think that we are good to go :) However back then we didn't take "case-insensitive" into account. So I'd write a small tool to make sure that there are no users with almost the same email addresses (case sensitivity). Although the chance they are any are very low.

zreigz:
I've created an application to check if user emails are unique in the cluster. The question is what we should do if we find two different users with the same email in the case-sensitive scenario.
https://github.com/kubermatic/kubermatic/pull/2468

zreigz:
My second thought is we usually use some external service for authentication like google. I don't know others but emails are always unique. I can't imagine a situation where we have the same emails with a different letter size.

p0lyn0mial:
> I can't imagine a situation where we have the same emails with a different letter size.

good point, the email addresses should be stable and immutable once created.

zreigz:
The mentioned issue on slack was related with already registered email: `Error 400: cannot add the user = bryan.cornies@sjrb.ca to the project j54jngwjmr because the user doesn't exist.` The user was: `Bryan.Cornies@sjrb.ca,` This PR fix this particular error

p0lyn0mial (approved):

PR #2468, opened 2018-11-30
[WIP] check if user email is unique

zreigz:
**What this PR does / why we need it**: This application checks if user email is unique in the cluster

**Release note**:
```release-note
```


p0lyn0mial:
I think we can close it ?

PR #2471, opened 2018-11-30
Disable debug logs on control plane

mrIncompetent:
Disable debug logs on control plane.

This will dramatically reduce the number of logs the control plane components will emit.

So far i was under the impression that `--v=4` is the default "info" log level - which is wrong.
Its just the level of verbosity - actually fairly obvious.

Looking at kubeadm, it does not specify the `--v` flag at all.

```
# Before the change:
~3200 messages/minute
# After the change:
~55 messages/second
```

**Release note**:
```release-note
Control plane components are no longer logging at debug level
```


mrIncompetent:
cc @kdomanski @toschneck

alvaroaleman (approved):

bashofmann:
@mrIncompetent Could this be back-ported to 2.8?

PR #2487, opened 2018-12-05
Update available Kubernetes versions

kdomanski:
* add newer K8S versions
* Drop support for creation of 1.8 and 1.9 clusters
* Automatically update to a version that has a fix for CVE-2018-1002105
* Add missing upgrade paths

**Release note**:
```release-note
Kubernetes will be automatically updated to versions that contain a fix for CVE-2018-1002105 and `v1.8`, `v1.9` cluster creation is now dropped
```


config/kubermatic/static/master/updates.yaml
```diff
@@ -16,12 +16,19 @@ updates:
...
-  #We need to skip 1.10.0 as it broke OpenStack.
-  to: ^1.10.1
+  to: ^1.10.11
   automatic: false
+# CVE-2018-1002105
+- from: <= 1.10.10
+  to: 1.10.11
+  automatic: true
+# Allow to next minor release
+- from: 1.10.*
```
alvaroaleman, line 34: Sholdn't we also auto update `>=1.11.0, <1.11.5` to `1.11.5`?

kdomanski: Done.

kdomanski: btw would `from: >=1.11.0, <1.11.5` be valid syntax?

alvaroaleman: yes

config/kubermatic/static/master/updates.yaml
```diff
@@ -16,12 +16,29 @@ updates:
...
-  to: 1.10.1
+  to: ^1.10.11
   automatic: false
 
 # ======= 1.10 =======
 # Allow to change to any patch version
 - from: 1.10.*
-  #We need to skip 1.10.0 as it broke OpenStack.
-  to: ^1.10.1
+  to: ^1.10.11
```
alvaroaleman, line 27: I think this one is implicit by the one  below

kdomanski: It is, but only because `1.10.11` is the only valid release in this minor version. If `v1.10.12` was released and enabled, this rule would allow upgrading to it.

config/kubermatic/static/master/updates.yaml
```diff
@@ -6,7 +6,7 @@ updates:
   automatic: false
 # Allow to next minor release
 - from: 1.8.*
-  to: 1.9.0
+  to: ^1.10.11
   automatic: false
```
alvaroaleman, line 11: Why not automatic: true?

config/kubermatic/static/master/updates.yaml
```diff
@@ -16,12 +16,29 @@ updates:
...
 # Allow to next minor release
 - from: 1.9.*
-  to: 1.10.1
+  to: ^1.10.11
   automatic: false
```
alvaroaleman, line 22: should be automatic: true as well

alvaroaleman (approved):

PR #2491, opened 2018-12-06
Ci conformance testing

alvaroaleman:

Fixes

**Documentation**:

**Release note**:

```release-note
NONE
```


mrIncompetent:
Why not follow the issue and:
> It deploys the kubermatic version of that PR into a dedicated namespace

?
That way, we would also test if the kubermatic chart works with its default values.

alvaroaleman:
@mrIncompetent I'm fine with changing it like that but maybe in a follow-up?

alvaroaleman:
And also, maybe doing both? That way we know if only the chart is broken or more

mrIncompetent:
> @mrIncompetent I'm fine with changing it like that but maybe in a follow-up?

sgtm

> And also, maybe doing both? That way we know if only the chart is broken or more

The chart should be our main focus as that is what customers use - not the dev/hack scripts.
Whats the benefit in maintaining both? We test the hack script?

alvaroaleman:
> The chart should be our main focus as that is what customers use - not the dev/hack scripts.
Whats the benefit in maintaining both? We test the hack script?

If we do not run everything as a presubmit we know on the next day that only the chart is broken, not the actual code

PR #2493, opened 2018-12-07
Use dynamic client

mrIncompetent:
Use dynamic client for the reconciling of StatefulSets in the cluster controller.

**Release note**:
```release-note
NONE
```

#### Why going with `StatefulSetCreator` + `StatefulSetObjectWrapper` ?
Using `ObjectCreator` has the disadvantage that all functions which would return a StatefulSet, would need to follow that interface & do a type conversion on their own. So we would need to duplicate that code for each function which creates a StatefulSet.
This is because Go does no implicit interface matching on function types.

Additionally:
The `runtime.Object` interface is very minimal:
```go
type Object interface {
	GetObjectKind() schema.ObjectKind
	DeepCopyObject() Object
}
```
Which would break* existing `something.Name` or `something.Name()` access - which we use a lot in the tests - see the resources/test package. 
*Sure, we also could do type conversions all over the place... But (IMO) that will decrease readability by a lot


mrIncompetent:
@alvaroaleman Squash + force pushing purged all comments :(
PTAL

api/pkg/resources/resources.go
```diff
@@ -725,3 +729,14 @@ func ClusterIPForService(name, namespace string, serviceLister corev1lister.Serv
...
 }
+
+// StatefulSetObjectWrapper adds a wrapper so the StatefulSetCreator matches ObjectCreator
+// This is needed as golang does not support function interface matching
+func StatefulSetObjectWrapper(create StatefulSetCreator) ObjectCreator {
+	return func(data *TemplateData, existing runtime.Object) (runtime.Object, error) {
+		if existing != nil {
+			return create(data, existing.(*appsv1.StatefulSet))
+		}
+		return create(data, nil)
```
alvaroaleman: Should we pass in an empty statefulset here and remove all the redundant code we have in every single creator that checks if existing is nil and if not creates an empty object?

mrIncompetent: Good idea!
I'll do that :)

mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/2493/commits/47c7bcdcbf8f6362640662c155b6aaf56a1522c6

mrIncompetent:
@alvaroaleman 
All fixed PTAL :)

api/pkg/resources/prometheus/statefulset.go
```diff
@@ -35,7 +35,7 @@ var (
 )
 
 // StatefulSet returns the prometheus StatefulSet
-func StatefulSet(data resources.StatefulSetDataProvider, existing *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
+func StatefulSet(data *resources.TemplateData, existing *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
 	var set *appsv1.StatefulSet
```
alvaroaleman, line 39: Please remove that check here as well, other than that, LGTM

mrIncompetent: The Prometheus StatefulSet does still use the old way. 
This PR only touches the cluster controller so far - just to get a reference implementation.
I'll update the rest in follow-up PR's :)

alvaroaleman (approved):

PR #2499, opened 2018-12-10
Ban possibility to add SSH keys with already existing name

### LINKED ISSUE #2426: Ban possibility to add SSH keys with already existing name
**User Story**
Atm it is possible to add a SSH key that has the same name as an already existing one (within the same project). This should not be possible, as it confuses users, because "the same" SSH key is shown twice (or even more often).

**Acceptance criteria**
It should only be possible to successfully add a new SSH key to a project, if its name isn't used within the project yet. Trying to add a SSH key with an already existing name should cause an error.

kgroschoff:
Ban possibility to add SSH keys with already existing name

Fixes #2426 

@p0lyn0mial Duplicated names for clusters and projects are already forbidden

**Release note**:

```release-note bugfix
SSH keys can no longer be added with duplicate names
```


api/pkg/handler/ssh.go
```diff
@@ -27,6 +27,17 @@ func createSSHKeyEndpoint(keyProvider provider.SSHKeyProvider, projectProvider p
 			return nil, kubernetesErrorToHTTPError(err)
 		}
 
+		existingKeys, err := keyProvider.List(project, nil)
+		if err != nil {
+			return nil, kubernetesErrorToHTTPError(err)
+		}
+
+		for _, key := range existingKeys {
```
p0lyn0mial: do you think we could move it to the provider and specify an option ? Similarly to what we do for clusters, https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/cluster.go#L131

If we ever add `update the key endpoint` then we don't have to change the code and we could reuse already existing :)/

p0lyn0mial (commented):
Can you please add a unit test ?

Have you checked if duplicates cannot be created for `clusters` and `project` resources ? ( I have just seen special notes :)

kgroschoff:
@p0lyn0mial ready for another review 😄

api/pkg/provider/types.go
```diff
@@ -112,6 +112,8 @@ type ClusterProvider interface {
 type SSHKeyListOptions struct {
 	// ClusterName gets the keys that are being used by the given cluster name
 	ClusterName string
+	// KeyName gets the ssh keys with the given name in the spec
```
p0lyn0mial: typo, please rename it to `SSHKeyName` :)

api/pkg/handler/ssh_test.go
```diff
@@ -192,6 +192,7 @@ func TestCreateSSHKeysEndpoint(t *testing.T) {
 		{
 			Name:             "scenario 1: a user can create ssh key that will be assigned to the given project",
 			Body:             `{"name":"my-second-ssh-key","spec":{"publicKey":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC8LlXSRW4HUYAjzx1+r5JzpjXIDDyFkWZzBQ8aU14J8LdMyQsU6/ZKuO5IKoWWVoPi0e63qSjkXPTjnUAwpE62hDm6uLaPgIlc3ND+8d9xbItS+gyXk9TSkC3emrsCWpS76W3KjLwyz5euIfnMCQZSASM7F5CrNg6XSppOgRWlyY09VEKi9PmvEDKCy5JNt6afcUzB3rAOK3SYZ0BYDyrVjuqTcMZwRodryxKb/jxDS+qQNplBNuUBqUzqjuKyI5oAk+aVTYIfTwgBTQyZT7So/u70gSDbRp9uHI05PkH60IftAHdYu4TJTmCwJxLW/suOEx3PPvIsUP14XQUZgmDJEuIuWDlsvfOo9DXZNnl832SGvTyhclBpsauWJ1OwOllT+hlM7u8dwcb70GD/OzCG7RSEatVoiNtg4XdeUf4kiqqzKZEqpopHQqwVKMhlhPKKulY0vrtetJxaLokEwPOYyycxlXsNBK2ei/IbGan+uI39v0s30ySWKzr+M9z0QlLAG7rjgCSWFSmy+Ez2fxU5HQQTNCep8+VjNeI79uO9VDJ8qvV/y6fDtrwgl67hUgDcHyv80TzVROTGFBMCP7hyswArT0GxpL9q7PjPU92D43UEDY5YNOZN2A976O5jd4bPrWp0mKsye1BhLrct16Xdn9x68D8nS2T1uSSWovFhkQ== lukasz@loodse.com "}}`,
+			RewriteSSHKeyID:  true,
```
p0lyn0mial, line 195: ah, good call :)

p0lyn0mial:
let's fix that small typo and merge it :)

p0lyn0mial (approved):

api/pkg/provider/types.go
kgroschoff: ups 😲 fixed

PR #2503, opened 2018-12-10
add startup container

mrIncompetent:
Add startup container. 
This is necessary as upstream never build a new container with the latest [script](https://github.com/kubernetes/contrib/blob/master/startup-script/manage-startup-script.sh).

The latest version creates checkpoints for each passed `STARTUP_SCRIPT` by hashing its content.
This ensures that changing `STARTUP_SCRIPT` will trigger a new run.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/containers/startup-script/Dockerfile
+++ b/containers/startup-script/Dockerfile
@@ -1,0 +1,4 @@
+FROM alpine
+RUN apk add --no-cache bash util-linux
+ADD manage-startup-script.sh /
+CMD /manage-startup-script.sh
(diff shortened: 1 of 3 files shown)


mrIncompetent:
When approved, i'll change the tag in the release script

alvaroaleman:
/lgtm

Maybe also add an addon for this? It wont do anything unless activated but my assumption is that we will need this more often :)

mrIncompetent:
> Maybe also add an addon for this? It wont do anything unless activated but my assumption is that we will need this more often :)

Only as a follow up and if we know exactly what we want to do with it.
Right now it only helps debugging

alvaroaleman (approved):

alvaroaleman:
Testing the cherrypick bots permissions...


/cherrypick release/v2.8

PR #2505, opened 2018-12-11
Add VerticalPodAutoscaler resources for all resources

### LINKED ISSUE #2424: Use the VerticalPodAutoscaler to scale all pods in the cluster namespace
**User Story**
To give Pods realistic resource guarantees (resource request), the VerticalPodAutoscaler should be used.
It will automatically update a pods resource.request based on the usage.

It will ensure that pods will have a realistic resource request set, which will ensure that it will only be scheduled on nodes with enough resources.
Additionally it will enable operators to have a more realistic overview about the resource usage.

Since the VerticalPodAutoscaler will only modify `resources.request`, the `resources.limit` must be increased enough so it meets even the resource needs of a strongly utilized user cluster.

**Acceptance criteria**
- [x] All pods in the cluster namespace have increased the `resources.limit`
- [x] A [VerticalPodAutoscaler](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/examples/redis.yaml) resource must be deployed for every pod in the cluster namespace

Outlook:
Potentially a HPA can be used to scale the Apiserver based on the CPU usage - though this must be tested thoroughly.

mrIncompetent:
- Add `VerticalPodAutoscaler` resources to all components
- Add the dynamic client to the monitoring controller
- Abstract away the controller specific `TemplateData` struct. 
  This was done so we can use the `EnsureObject` function across all controllers (Cluster, Monitoring, and more to come). Without this refactoring, all controller would have needed to implement handling for the `TemplateData` struct. 

Closes #2424

**Release note**:
```release-note
Experimantal support for VerticalPodAutoscaler has been added. The VPA resources use the PodUpdatePolicy=initial
```


api/pkg/resources/ensure.go
```diff
@@ -567,3 +529,53 @@ func EnsureObject(data *TemplateData, namespace string, rawcreate ObjectCreator,
 
 	return nil
 }
+
+// EnsureVerticalPodAutoscalers will create and update the VerticalPodAutoscalers coming from the passed VerticalPodAutoscalerCreator slice
+func EnsureVerticalPodAutoscalers(creators []VerticalPodAutoscalerCreator, namespace string, client ctrlruntimeclient.Client, informerFactory ctrlruntimecache.Cache, wrapper ...ObjectModifier) error {
+	store, err := informerutil.GetSyncedStoreFromDynamicFactory(informerFactory, &v1beta1.VerticalPodAutoscaler{})
+	if err != nil {
+		if _, crdNotRegistered := err.(*meta.NoKindMatchError); crdNotRegistered {
```
alvaroaleman: Shouldn't we check for this in cmd and simply crash if its not there?

api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -24,6 +24,9 @@ var (
 			corev1.ResourceCPU:    resource.MustParse("2"),
 		},
 	}
+
+	// VerticalPodAutoscaler returns a VerticalPodAutoscaler which can be applied to the machine-controller Deployment
+	VerticalPodAutoscaler = resources.GetVerticalPodAutoscaler(name, resources.BaseAppLabel(name, nil))
```
alvaroaleman, line 29: If we do it like this anyways for everything, wouldn'it be easier to use a wrapper for everything in `ensureObject` that does a type assertion on Depoyment and Statefulset and if there is a match, adds the VerticalPodAutoscaler Resource with an owner Ref on the Deployment/Statefulset?

api/pkg/resources/ensure.go
mrIncompetent: It does not break any functionality - and the VPA is marked as v1beta1.
IMO keeping it optional is a good compromise.

api/pkg/resources/machinecontroller/deployment.go
mrIncompetent: That's brilliant!

But why a OwnerRef on the Deployment/StatefulSet instead of the cluster?

alvaroaleman: Because hypothetically we could decide we do not want to deploy Deployment XYZ anymore in a newer version of Kubermatic and then we'd have strain VPA CRs

api/pkg/resources/ensure.go
alvaroaleman: I disagree because we are going to run e2e tests and if in those e2e tests we run leverage a VPA that does not exist $customer, test may pass for us but not for the customer. Aside, Beta is new Stable in Kubernetes world

api/pkg/resources/machinecontroller/deployment.go
mrIncompetent: Nice idea, i'll adapt the pull!

api/pkg/resources/ensure.go
mrIncompetent: The VPA only adjusts the `resource.request` of pods - What would be the scenario, where this could influence the cluster behavior?

alvaroaleman: Well it may be different which may result in the pod crashing due to insufficient resources

mrIncompetent: > which may result in the pod crashing due to insufficient resources

Which applies for both ways. With or without VPA.
The VPA only set's the `resource.request` - not limit. It does not protect pods from getting OOM killed.

mrIncompetent: Moved the check to startup: https://github.com/kubermatic/kubermatic/pull/2505/commits/810763068f82c455b6e4797ad798d3873388f95c

api/pkg/resources/machinecontroller/deployment.go
mrIncompetent: That would mean to update the Deployment handling here in this PR as well.
I'll wait with that until the discussion around the object reconciling is done.

alvaroaleman:
@mrIncompetent Can we add the VPA deployment/Webhook etc to our chart? Other than that LGTM

config/vertical-pod-autoscaler/templates/tls.yaml
```diff
@@ -0,0 +1,14 @@
+{{- $ca := genCA "deployment-admission-controller" 3650 -}}
+{{- $cn := .Chart.Name -}}
+{{- $altName1 := printf "%s.%s" .Chart.Name .Release.Namespace -}}
+{{- $altName2 := printf "%s.%s.svc" .Chart.Name .Release.Namespace -}}
+{{- $cert := genSignedCert $cn nil (list $altName1 $altName2) 3650 $ca -}}
```
mrIncompetent: This will always generate a new certificate & cause the secret to update.
Even though this is not very nice, the VPA handles this case by deleting the `MutatingWebhookConfiguration` on each start.

The admission controller deployment uses `checksum/tls: {{ include (print $.Template.BasePath "/tls.yaml") . | sha256sum }}` to ensure we always roll the pod in case the secret changes.

alvaroaleman:
@mrIncompetent Why use a dedicated chart? This is a non-optional part of Kubermatic now

mrIncompetent:
@alvaroaleman PTAL

config/vertical-pod-autoscaler/templates/tls.yaml
alvaroaleman: Looking forward to the Kubermatic Operator

alvaroaleman:
@mrIncompetent failures seem to be real, the exact same error was in the test before. Maybe the apiserver gets scaled when the conformance tester lists nodes? Also we probably should wrapt all calls from the conformance tester to the user clusters apiserver in a retry loop

mrIncompetent:
I'm suspecting the preemtible nodes here - the machine-controller got unhealthy as well during the test.
I saw this behaviour before when a node got replaced on our GKE cluster.

Fix for the test tool: https://github.com/kubermatic/kubermatic/pull/2526

mrIncompetent:
@kron4eg PTAL

PR #2506, opened 2018-12-11
move resources into the individual controller package where they belong

mrIncompetent:
Move resources into the individual controller package where they belong.
Also this simplifies the DexCA function, which was more complex than actually needed.

**Release note**:
```release-note
NONE
```


api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -451,8 +451,10 @@ func (cc *Controller) processNextItem() bool {
 	}
 	defer cc.queue.Done(key)
 
-	if err := cc.syncCluster(key.(string)); err != nil {
-		cluster, errGetCluster := cc.clusterLister.Get(key.(string))
+	sKey := key.(string)
+	if err := cc.syncCluster(sKey); err != nil {
+		glog.Errorf("Failed syncinc '%s': %v", sKey, err)
```
alvaroaleman, line 456: Why do you re-introduce that?

alvaroaleman: And also, why not on  a high `V` level?

api/pkg/controller/cluster/resources.go
```diff
@@ -231,7 +232,7 @@ func (cc *Controller) ensureSecrets(c *kubermaticv1.Cluster, data *resources.Tem
 
 	for _, op := range operations {
 		if err := resources.EnsureSecret(op.name, data, op.create, cc.secretLister.Secrets(c.Status.NamespaceName), cc.kubeClient.CoreV1().Secrets(c.Status.NamespaceName)); err != nil {
-			return fmt.Errorf("failed to ensure that the Secret exists: %v", err)
+			return fmt.Errorf("failed to ensure that the Secret '%s' exists: %v", op.name, err)
```
alvaroaleman: Can you reword this to `failed to ensure Secret '%s'`? The secret may exist, what failed is to ensure its desired state, not its existence

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: I don't see any reason not to log an error during sync.
Otherwise i would need to watch the events.

We run into an error, so we log with `glog.Errorf`

api/pkg/controller/cluster/resources.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/2506/commits/e81424bdd6a505a21bad3a88845a5e4758b0eb04

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: But the error may be transient so if you handle a higher amount of clusters this may result in a lot of errors and these are debug logs that should not be turned on by default. Kubernetes controller-manager also does not log errors for any error in a sync loop it encounters

mrIncompetent:
I'm closing this one and issue separate PR's

PR #2507, opened 2018-12-11
Add PATCH endpoint for node deployments

### LINKED ISSUE #2453: An HTTP endpoint for updating node deployments
**Description** Consider using `PATCH` instead of `PUT` when implementing the endpoint.

maciaszczykm:
**What this PR does / why we need it**: Adds PATCH endpoint for node deployments to allow updating them.

Fixes #2453

**Special notes for your reviewer**: It was tested with changes for Dashboard: https://github.com/kubermatic/dashboard-v2/commit/95b39c257a9addafaf3cf5dcf76c2a3cf0fef3e0

I will add some tests.

**Documentation**:

**Release note**:

```release-note
NONE
```


maciaszczykm:
It is ready for a review.

alvaroaleman:
Why can I approve for `api/cmd/kubermatic-api/OWNERS` and `api/pkg/handler/OWNERS `? :thinking:

api/pkg/handler/node_v1.go
```diff
@@ -845,46 +841,158 @@ func decodeGetNodeDeploymentForCluster(c context.Context, r *http.Request) (inte
...
 
-func getNodeDeploymentForCluster(projectProvider provider.ProjectProvider) endpoint.Endpoint {
+func getNodeDeployment() endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(NodeDeploymentReq)
 		clusterProvider := ctx.Value(clusterProviderContextKey).(provider.ClusterProvider)
 		userInfo := ctx.Value(userInfoContextKey).(*provider.UserInfo)
 
-		_, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{})
+		cluster, err := clusterProvider.Get(userInfo, req.ClusterID, &provider.ClusterGetOptions{})
```
p0lyn0mial: please bring back checking if a project exists
`_, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{})`

api/pkg/handler/node_v1.go
```diff
@@ -845,46 +841,158 @@ func decodeGetNodeDeploymentForCluster(c context.Context, r *http.Request) (inte
...
+
+func decodePatchNodeDeployment(c context.Context, r *http.Request) (interface{}, error) {
+	var req PatchNodeDeploymentReq
+
+	clusterID, err := decodeClusterID(c, r)
+	if err != nil {
+		return nil, err
+	}
+
+	nodeDeploymentID := mux.Vars(r)["nodedeployment_id"]
```
p0lyn0mial: do you think we could move it to a separate method so that checking is in one place? Just like we do with `decodeDcReq` and `decodeClusterID`. Changing requires updating `decodeDeleteNodeDeployment`, `decodeGetNodeDeployment` and `decodePatchNodeDeployment`

api/pkg/handler/node_v1.go
```diff
@@ -845,46 +841,158 @@ func decodeGetNodeDeploymentForCluster(c context.Context, r *http.Request) (inte
...
+		req := request.(PatchNodeDeploymentReq)
+		clusterProvider := ctx.Value(clusterProviderContextKey).(provider.ClusterProvider)
+		userInfo := ctx.Value(userInfoContextKey).(*provider.UserInfo)
+
 		cluster, err := clusterProvider.Get(userInfo, req.ClusterID, &provider.ClusterGetOptions{})
 		if err != nil {
 			return nil, kubernetesErrorToHTTPError(err)
 		}
 
+		project, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{})
```
p0lyn0mial: check if the project exits before getting the cluster (line 913)

alvaroaleman:
> Why can I approve for api/cmd/kubermatic-api/OWNERS and api/pkg/handler/OWNERS ? ￼

The reason is this: https://github.com/kubermatic/kubermatic/pull/2513

@p0lyn0mial should also approve, after that, feel free to `/hold cancel`

api/pkg/handler/node_v1.go
```diff
@@ -845,46 +841,158 @@ func decodeGetNodeDeploymentForCluster(c context.Context, r *http.Request) (inte
...
+			return nil, fmt.Errorf("failed to create machine deployment from template: %v", err)
+		}
+
+		// We need to ensure that the name and resource version are set and the selector stays the same.
+		md.Name = req.NodeDeploymentID
+		md.ResourceVersion = machineDeployment.ResourceVersion
+		md.Spec.Selector = machineDeployment.Spec.Selector
+		md.Spec.Template.ObjectMeta.Labels = machineDeployment.Spec.Template.ObjectMeta.Labels
+
+		md, err = machineClient.ClusterV1alpha1().MachineDeployments(md.Namespace).Update(md)
```
p0lyn0mial: I am afraid that we can not "patch the machine deployment in that way. Because `Update` means replace and the object we are passing to it is created from `patchedNodeDeployment` which doesn't fully mirror already existing object. We overwrite most fields from `ObjectMetadata` and `Status`, is that correct ?

p0lyn0mial (changes requested):

api/pkg/handler/node_v1.go
maciaszczykm: Ah, so this is the purpose of these calls. Didn't check that precisely. I will do it.

api/pkg/handler/node_v1.go
maciaszczykm: Of course.

api/pkg/handler/node_v1.go
maciaszczykm: You are right, this may bring some issues.

Above lines are used to ensure that most important fields like resource version will not be overridden. They will come from GET that is executed before. I agree that this is not the ideal solution, but the status will be fine anyways. Internal logic will keep it up to date. I was not sure how to solve this, because:

- In the perfect world, I would be able to use PATCH method directly from Machine Deployment endpoint, but we are not exposing structures directly (using internal and external ones, and rightly so). So we have our API that exposes `NodeDeployments` PATCH endpoint, and incoming patch request bodies cannot be easily mapped to call PATCH `MachineDeployment` internally.
- Using current solution is also not the best one because we have to ensure that incoming patch object will not change important fiels (i.e. resource version, status etc.). Concurrency will not be an issue, as concurrent edits will be blocked (resource version has to be latest to update an object).

What if we would allow patching only the `spec` of `NodeDeployment` and get all other information from previous GET?

Do you have any suggestions here?

p0lyn0mial: for future reference: we had a chat and decided to update only the spec.

maciaszczykm:
@p0lyn0mial PTAL

api/pkg/handler/node_v1.go
```diff
@@ -875,20 +881,140 @@ func getNodeDeploymentForCluster(projectProvider provider.ProjectProvider) endpo
...
+		if err != nil {
+			return nil, kubernetesErrorToHTTPError(err)
+		}
+
+		cluster, err := clusterProvider.Get(userInfo, req.ClusterID, &provider.ClusterGetOptions{})
+		if err != nil {
+			return nil, kubernetesErrorToHTTPError(err)
+		}
+
+		project, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{})
```
p0lyn0mial: this line can be removed, since we get project in line `928`

api/pkg/handler/node_v1.go
```diff
@@ -875,20 +881,140 @@ func getNodeDeploymentForCluster(projectProvider provider.ProjectProvider) endpo
...
+		}
+
+		patchedMachineDeployment, err := machineresource.Deployment(cluster, patchedNodeDeployment, dc, keys)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create machine deployment from template: %v", err)
+		}
+
+		// Only spec will be updated by a patch.
+		// It ensures that the name and resource version are set and the selector stays the same.
+		machineDeployment.Spec = patchedMachineDeployment.Spec
```
p0lyn0mial, line 988: I think that detecting updates to other fields than `spec` can be difficult ? If so, I would appreciate a comment (swagger spec) that explains which fields can be updated. If not we could return an error.

api/pkg/handler/node_v1_test.go
```diff
@@ -845,7 +846,155 @@ func TestGetNodeDeploymentForCluster(t *testing.T) {
...
+			Body:             fmt.Sprintf(`{"spec":{"replicas":%v}}`, replicasUpdated),
+			ExpectedResponse: fmt.Sprintf(`{"id":"venus","name":"venus","creationTimestamp":"0001-01-01T00:00:00Z","spec":{"replicas":%v,"template":{"cloud":{"digitalocean":{"size":"2GB","backups":false,"ipv6":false,"monitoring":false,"tags":["kubermatic","kubermatic-cluster-defClusterID"]}},"operatingSystem":{"ubuntu":{"distUpgradeOnBoot":true}},"versions":{"kubelet":""}},"paused":false},"status":{}}`, replicasUpdated),
+			cluster:          "keen-snyder",
+			HTTPStatus:       http.StatusOK,
+			project:          genDefaultProject().Name,
+			ExistingAPIUser:  genDefaultAPIUser(),
+			NodeDeploymentID: "venus",
+			ExistingMachineDeployments: []*clusterv1alpha1.MachineDeployment{
+				{
+					ObjectMeta: metav1.ObjectMeta{
```
p0lyn0mial: do you think we could wrap it into `genDefaultMachineDeployment(...)` ? If so I would appreciate updating the tests that use `MachineDeployment`

api/pkg/handler/node_v1_test.go
```diff
@@ -845,7 +846,155 @@ func TestGetNodeDeploymentForCluster(t *testing.T) {
...
+	for _, tc := range testcases {
+		t.Run(tc.Name, func(t *testing.T) {
+			req := httptest.NewRequest(http.MethodPatch, fmt.Sprintf("/api/v1/projects/%s/dc/us-central1/clusters/%s/nodedeployments/%s",
+				genDefaultProject().Name, genDefaultCluster().Name, tc.NodeDeploymentID), strings.NewReader(tc.Body))
+			res := httptest.NewRecorder()
+			kubermaticObj := []runtime.Object{}
+			machineDeploymentObjets := []runtime.Object{}
+			kubernetesObj := []runtime.Object{}
+			kubermaticObj = append(kubermaticObj, tc.ExistingKubermaticObjs...)
+			for _, existingMachineDeployment := range tc.ExistingMachineDeployments {
```
p0lyn0mial, line 921: can this be simplified to `machineDeploymentObjets = append(machineDeploymentObjets, tc.ExistingMachineDeployment...)` ?

api/pkg/handler/routes_v1.go
```diff
@@ -1500,24 +1504,50 @@ func (r Routing) createNodeDeploymentForCluster() http.Handler {
...
+		)(getNodeDeployment(r.projectProvider)),
+		decodeGetNodeDeployment,
+		encodeJSON,
+		r.defaultServerOptions()...,
+	)
+}
+
+// swagger:route GET /api/v1/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/nodedeployments/{nodedeployment_id} project patchNodeDeployment
+//
+//     Patches a node deployment that is assigned to the given cluster.
```
p0lyn0mial: that's the place I would add the comment about what can be updated

p0lyn0mial (changes requested):

api/pkg/handler/node_v1.go
maciaszczykm: Yep, it would be difficult. I will update comments for swagger.

api/pkg/handler/node_v1_test.go
maciaszczykm: Seems like it can't. Types won't match if I try to change it.

maciaszczykm:
@p0lyn0mial PTAL :)

PR #2516, opened 2018-12-13
remove unecessary files from the kubernetes test files

mrIncompetent:
**What this PR does / why we need it**:
Remove unecessary files from the kubernetes test files

**Release note**:
```release-note
NONE
```

PR #2521, opened 2018-12-17
adds "kubernetes" suffix to a directory name

p0lyn0mial:
**What this PR does / why we need it**: turns out that `kubetest` assumes that the last
part of that path contains "kubernetes". If not then it complains `directory: must run from kubernetes directory root`. 

See: https://github.com/kubernetes/test-infra/blob/master/kubetest/main.go#L216

**Release note**:

```
NONE
```

### DIFF
--- a/containers/conformance-tests/install.sh
+++ b/containers/conformance-tests/install.sh
@@ -9,7 +9,7 @@ apt update && apt install -y git-crypt curl
 TMP_ROOT="./.install-tmp"
 
 for VERSION in 1.10 1.11 1.12; do
-    DIRECTORY="${ROOT_DIR}/${VERSION}"
+    DIRECTORY="${ROOT_DIR}/${VERSION}-kubernetes"
     if [ ! -d "${DIRECTORY}" ]; then
 
         FULL_VERSION=$(curl -Ss https://storage.googleapis.com/kubernetes-release/release/stable-${VERSION}.txt)


p0lyn0mial:
`e2e` tests for `kubeone` will use the image so that that they don't have to download additional binaries. The test use `kubetest` to run the conformance test.

containers/conformance-tests/install.sh
```diff
@@ -9,7 +9,7 @@ apt update && apt install -y git-crypt curl
 TMP_ROOT="./.install-tmp"
 
 for VERSION in 1.10 1.11 1.12; do
-    DIRECTORY="${ROOT_DIR}/${VERSION}"
+    DIRECTORY="${ROOT_DIR}/${VERSION}-kubernetes"
```
p0lyn0mial, line 12: will it brake already existing tests ?

thz:
nit first:
s/"kubernets" prefix/"kubernetes" suffix/

p0lyn0mial:
> nit first:
s/"kubernets" prefix/"kubernetes" suffix/

I've updated the title.

alvaroaleman:
@mrIncompetent Does some magic within the `conformance-tester` to make that work, maybe you can look there? Since it works I feel like this change should not be required

p0lyn0mial:
> @mrIncompetent Does some magic within the conformance-tester to make that work, maybe you can look there? Since it works I feel like this change should not be required

Wouldn't be better to adjust the directory to `kubetest`'s expectations ? Otherwise everyone that wants to use the image must make additional changes.

containers/conformance-tests/install.sh
mrIncompetent: Yes, it will.
We either have to adapt the conformance test tool, or we create a symlink

mrIncompetent:
If you decide to change the image (What should be simpler than updating the code), you need to update the image tag & release the image as well. Those images are not being build & released by the CI.

See https://github.com/kubermatic/kubermatic/blob/master/containers/conformance-tests/release.sh#L3

I would recommend waiting until https://github.com/kubermatic/kubermatic/pull/2516 has been merged, as we reduced the image size there quite a bit.

mrIncompetent:
@p0lyn0mial whats the state here?

mrIncompetent:
Ping @p0lyn0mial

p0lyn0mial:
sorry for being slow, do we want to change the image ? If yes, I can update places that use this img when I have time :)

mrIncompetent:
@p0lyn0mial You opened this PR because I assume you wanted to use this image with Kubetest.
In case you don't need this anymore, please close this PR.

I remember @alvaroaleman had plans to to replace kubetest with Ginkgo ? So we might not use kubetest in the future at all

mrIncompetent:
Closing since this seems to be not relevant anymore.
In case it becomes relevant again, reopen and consider https://github.com/kubermatic/kubermatic/pull/2521#issuecomment-448167643

PR #2522, opened 2018-12-17
Explicitly enable ip forwarding

mrIncompetent:
Explicitly enable IP forwarding.
This will ensure the VPN will work regardless if the host of the seed cluster has IP forwarding enabled or not.
This can eliminate useless debugging.

**Release note**:
```release-note bugfix
Fixed an issue with kubelets being unreachable by the apiserver on some OS configurations.
```

### DIFF
--- a/api/pkg/resources/test/fixtures/deployment-aws-1.10.0-openvpn-server.yaml
+++ b/api/pkg/resources/test/fixtures/deployment-aws-1.10.0-openvpn-server.yaml
@@ -132,17 +132,24 @@ spec:
       initContainers:
       - args:
         - -c
-        - "\n\t\t\t\t# do not give a 10.20.0.0/24 route to clients (nodes) but\n\t\t\t\t#
-          masquerade to openvpn-server's IP instead:\n\t\t\t\tiptables -t nat -A POSTROUTING
-          -o tun0 -s 10.20.0.0/24 -j MASQUERADE\n\n\t\t\t\t# Only allow outbound traffic
-          to services, pods, nodes\n\t\t\t\tiptables -P FORWARD DROP\n\t\t\t\tiptables
-          -A FORWARD -m state --state ESTABLISHED,RELATED -j ACCEPT\n\t\t\t\tiptables
-          -A FORWARD -i tun0 -o tun0 -s 10.20.0.0/24 -d 172.25.0.0/16 -j ACCEPT\n\t\t\t\tiptables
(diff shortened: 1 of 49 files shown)


api/pkg/resources/test/fixtures/deployment-aws-1.10.0-openvpn-server.yaml
```diff
@@ -132,17 +132,24 @@ spec:
...
-          -A FORWARD -m state --state ESTABLISHED,RELATED -j ACCEPT\n\t\t\t\tiptables
-          -A FORWARD -i tun0 -o tun0 -s 10.20.0.0/24 -d 172.25.0.0/16 -j ACCEPT\n\t\t\t\tiptables
-          -A FORWARD -i tun0 -o tun0 -s 10.20.0.0/24 -d 10.10.10.0/24 -j ACCEPT\n\t\t\t\tiptables
-          -A FORWARD -i tun0 -o tun0 -s 10.20.0.0/24 -d 192.0.2.0/24 -j ACCEPT\n\n\t\t\t\tiptables
-          -A INPUT -m state --state ESTABLISHED,RELATED -j ACCEPT\n\t\t\t\tiptables
-          -A INPUT -i tun0 -p icmp -j ACCEPT\n\t\t\t\tiptables -A INPUT -i tun0 -j
-          DROP\n\t\t\t\t"
+        - |
+          # Always set IP forwarding
+          sysctl -w net.ipv4.ip_forward=1
```
thz, line 137: Did you test that?
I thought you can only change sysctl with the `security.alpha.kubernetes.io/*sysctl`-annotations since around 1.10 or 1.11?
And it would require more than just NET_ADMIN, right?

mrIncompetent: Yes it did test that, if you want to retest:
Create a new cluster, exec into the openvpn server pod & issue the above command.

mrIncompetent: The versions i tested this on:
- V1.11
- V1.13

thz: It seems my memory was about sysctls which count as node-level. Enabling ip_forward in this way should be okay.

alvaroaleman:
/cherrypick release/v2.8

wrong bot...

PR #2534, opened 2018-12-18
unify date format

### LINKED ISSUE #2257: Unify date format
**User Story** Noticed when working on #2254. Date format differs depending on currect localization, i.e.:

UTC:`"creationTimestamp":"2013-02-03T19:54:00Z"`
+1: `"creationTimestamp":"2013-02-03T20:54:00+01:00"`

Date is the same, but format is different. It may cause potential issues as it is applied during JSON unmarshal among others. In my case it caused test failure, because in tests we are checking mostly for UTC.

**Acceptance criteria** Like Kubernetes we should make sure that specific rule is used and follow it, i.e. https://github.com/kubernetes/kubernetes/blob/7cbb9995189c5ecc8182da29cd0e30188c911401/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L167-L176

/cc @p0lyn0mial

zreigz:
**What this PR does / why we need it**: create a custom type (for fields that hold time) and write custom Marshal and Unmarshal methods that would always encode/decode to desired format.

Fixes #2257 

**Release note**:
```release-note bugfix
Timestamp format has been unified throughout the Kubermatic API.
```


api/cmd/kubermatic-api/swagger.json
```diff
@@ -4114,16 +4090,10 @@
       "type": "object",
       "properties": {
         "creationTimestamp": {
-          "description": "CreationTimestamp is a timestamp representing the server time when this object was created.",
-          "type": "string",
-          "format": "date-time",
-          "x-go-name": "CreationTimestamp"
+          "$ref": "#/definitions/Time"
```
p0lyn0mial: could we bring back the previous description ? Before we described the purpose of the filed `CreationTime` now we describe the generic type `Time`.

api/pkg/api/v1/time.go
```diff
@@ -0,0 +1,85 @@
...
+	return Time{time.Date(year, month, day, hour, min, sec, nsec, loc)}
+}
+
+// Now returns the current local time.
+func Now() Time {
+	return Time{time.Now()}
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (t Time) MarshalJSON() ([]byte, error) {
```
p0lyn0mial: why this method doesn't have a pointer receiver ? 

I would remove a pointer receiver from all methods except `UnmarshalJSON` or I would add a pointer receiver to all methods.

api/pkg/api/v1/time_test.go
```diff
@@ -0,0 +1,55 @@
+package v1_test
```
p0lyn0mial, line 1: :)

api/pkg/api/v1/time_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	. "github.com/kubermatic/kubermatic/api/pkg/api/v1"
```
p0lyn0mial, line 8: Import dot is not recommended, can we remove it ?

api/cmd/kubermatic-api/swagger.json
zreigz: uff, was hard but magic annotation helped `// swagger:strfmt date-time`

api/pkg/api/v1/time.go
zreigz: done

api/pkg/api/v1/time_test.go
zreigz: unfortunately not :(

xrstf: https://github.com/kubermatic/kubermatic/pull/8956 fixes it :)

PR #2548, opened 2018-12-19
Use https scraping for controller-manager and scheduler

mrIncompetent:
Use HTTPS for scraping for scraping the controller-manager and scheduler.
The HTTPS endpoints got introduced in those components with 1.12 & 1.13 together with the deprecation of the HTTP endpoints.
Both components will use auth-delegation to validate requests against the API server - thus the RBAC resources are needed.

This enables 1.13.0 as well.

**Release note**:
```release-note
NONE
```


api/pkg/resources/apiserver/deployment.go
```diff
@@ -84,9 +84,9 @@ func Deployment(data resources.DeploymentDataProvider, existing *appsv1.Deployme
 	dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
 		Labels: podLabels,
 		Annotations: map[string]string{
-			"prometheus.io/scrape-apiserver": "true",
-			"prometheus.io/path":             "/metrics",
-			"prometheus.io/port":             strconv.Itoa(int(externalNodePort)),
+			"prometheus.io/scrape_with_kube_cert": "true",
```
alvaroaleman, line 90: That is smart! But we have to switch beetween the bare scrape and the cert scrape annotation based on the version of the cluster, right?

api/pkg/resources/scheduler/deployment.go
```diff
@@ -206,6 +194,15 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
 		"--kubeconfig", "/etc/kubernetes/kubeconfig/kubeconfig",
 	}
 
+	// With 1.13 we're using the secure port for scraping metrics as the insecure port got marked deprecated
+	if cluster.Spec.Version.Semver().Minor() >= 13 {
```
alvaroaleman, line 201: Shouldn't we already do it in 1.12?

alvaroaleman (commented):
Some nits but overall a real nice change!

More of a thought to discuss: Would it maybe make sense to add another Presubmit that tests 1.13, considering how unflaky that job is?

Or should we just add a big pile of periodics for the time being?

api/pkg/resources/apiserver/deployment.go
alvaroaleman: scratch the comment

api/pkg/resources/scheduler/deployment.go
mrIncompetent: The scheduler only supports it for 1.13+ :/

mrIncompetent:
Regarding the testing:
I would add presubmits for all supported versions(still only aws) . They can all run in parallel and it should not take really longer. 

On top maybe a periodic which tests all(providers+versions)

alvaroaleman:
/approve
/lgtm

E2Es give so much confidence for changes like this, just awesome

mrIncompetent:
@thetechnick please have a look at the prom changes

mrIncompetent:
Just realized @thetechnick is on vacation.
@kron4eg you have some time for a review?

PR #2549, opened 2018-12-20
Update conformance test versions & Extend CI e2e coverage

mrIncompetent:
- Update versions to test in the conformance test tool
- Remove the version limitation from the CI conformance test script

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,7 +15,6 @@ linters:
     - ineffassign
     - goconst
     - gocyclo
-    - maligned
     - misspell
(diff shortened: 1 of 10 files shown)


alvaroaleman:
@mrIncompetent This will quadruple the run time. We should instead make the version configurable via an env var and add three new presubmits, one per version

mrIncompetent:
@alvaroaleman Why will this quadruple the run time?
All clusters are being created in parallel & so are the ginkgo tests

PR #2555, opened 2018-12-21
Conformance tester: fail testcase if conformance tests failed

alvaroaleman:

Currently the conformance-tester happily returns with a RC0 if the conformance tests failed: https://prow.loodse.com/log?job=pull-kubermatic-e2e-aws-coreos-1.13&id=1076120506983256065

This PR fixes that

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/cmd/conformance-tests/runner.go
+++ b/api/cmd/conformance-tests/runner.go
@@ -353,6 +353,7 @@ func (r *testRunner) testCluster(
 		return nil
 	})
 	if err != nil {
+		report.Errors++
 		log.Errorf("Failed to successfully run kubernetes e2e tests: %v", err)
 	}
 


mrIncompetent:
I cannot access the logs via the link above anymore :/

But i remember that the tests did not actually fail, the ginkgo run reported a flakey test.
A flakey test results in ginkgo returning with statuscode 1.

Instead of interpreting a non-0 statuscode as a failure, i would prefer to always collect the JUnit reports & check the report.

alvaroaleman:
> I cannot access the logs via the link above anymore :/

https://storage.cloud.google.com/prow-data/pr-logs/pull/kubermatic_kubermatic/2548/pull-kubermatic-e2e-aws-coreos-1.13/1079687257834655744/build-log.txt?authuser=1&folder=true&organizationId=1081285000144

> But i remember that the tests did not actually fail, the ginkgo run reported a flakey test.
> A flakey test results in ginkgo returning with statuscode 1.

In that case we would have to check the junit reports during the retry loop because currently we retry whenever ginkgo does not return with rc0.

mrIncompetent:
> In that case we would have to check the junit reports during the retry loop because currently we retry whenever ginkgo does not return with rc0.

Yes, I think that's the more elegant solution

alvaroaleman:
@mrIncompetent but just to verify: Are you sure the tests currently fail if ginkgo reported failures? Because the log output doesn't look like that

alvaroaleman:
Closing in favor of #2563

PR #2556, opened 2018-12-28
add missing drone build step

mrIncompetent:
We removed the `build` step from drone, which causes the `docker` step to fail as the `docker` step depends on the binaries produced during the `build` step.

**Release note**:
```release-note
NONE
```


nikhita (approved):
/lgtm
/approve

PR #2563, opened 2019-01-02
Analyze the JUnit report in case ginkgo fails, to accept flakey tests

mrIncompetent:
This will make our conformance test command analyze the reports in case ginkgo returns with a non-0 status code.
This is required as ginkgo returns with status code 1 in case it encounters flakey tests.
Tests a being marked as flakey if ginkgo had to retry a tests. In case the test still did not succeed after several retries, it will be marked as failure - So tests which are only marked as flakey succeeded after some retries.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/cmd/conformance-tests/runner.go
+++ b/api/cmd/conformance-tests/runner.go
@@ -282,6 +282,14 @@ func (r *testRunner) testCluster(
 	const maxTestAttempts = 3
 	var err error
 	totalStart := time.Now()
+	log.Info("Starting to test cluster...")
+	defer log.Infof("Finished testing cluster after %s", time.Since(totalStart))
+
+	// We'll store the report there and all kinds of logs
+	scenarioFolder := path.Join(r.reportsRoot, scenarioName)
+	if err := os.MkdirAll(scenarioFolder, os.ModePerm); err != nil {
+		return nil, fmt.Errorf("failed to create the scenario folder '%s': %v", scenarioFolder, err)
+	}
 
 	report := &reporters.JUnitTestSuite{
 		Name: scenarioName,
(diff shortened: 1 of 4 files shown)


api/cmd/conformance-tests/runner.go
```diff
@@ -335,7 +335,7 @@ func (r *testRunner) testCluster(
 		log.Infof("[Attempt: %d/%d] Starting E2E tests...", attempt, maxTestAttempts)
 		if err := r.runE2E(log, cluster, kubeconfigFilename, cloudConfigFilename, reportsDir, apiNodes, dc); err != nil {
 			log.Warnf("[Attempt: %d/%d] failed to run E2E tests: %v", attempt, maxTestAttempts, err)
```
alvaroaleman: I would be in favor of removing that log message, it is very confusing to have the tests print it and then exit without an error

alvaroaleman (commented):
With this change we will still run the tests again if ginkgo returned a rc != 0 or am I missing something?

mrIncompetent:
> With this change we will still run the tests again if ginkgo returned a rc != 0 or am I missing something?

Tests will be retried if no reports could be collected (No junit xml's could be found -> Ginkgo failed hard) or if one of the tests in the test suite failed.

api/cmd/conformance-tests/runner.go
mrIncompetent: Would you be ok with changing the log level to debug?

alvaroaleman: IMHO this should only return an error if there was an error in the junit results or there are no results but not care about the return code. This would also avoid unnecesary retries

mrIncompetent: I'll move the report collection into the `runE2E` function

api/cmd/conformance-tests/runner.go
```diff
@@ -322,54 +330,56 @@ func (r *testRunner) testCluster(
...
-		log.Errorf("Failed to successfully run kubernetes e2e tests: %v", err)
-	}
-
-	if ginkgoReport != nil {
-		report.Time = time.Since(totalStart).Seconds()
-		report.Tests += ginkgoReport.Tests
-		report.Errors += ginkgoReport.Errors
-		report.Failures += ginkgoReport.Failures
-		report.TestCases = append(report.TestCases, ginkgoReport.TestCases...)
+		if ginkgoRes != nil {
```
alvaroaleman: Are you sure this cover all cases? I would expect `ginkoRes` to be nil if there is a `runErr`. Maybe just setting a success boolean inside the loop and switching off of it here would be better?

api/cmd/conformance-tests/runner.go
```diff
@@ -732,51 +739,94 @@ func (r *testRunner) runE2E(
...
+	log = log.WithField("ginkgo-log", file.Name())
 
-			cmd.Stdout = writer
-			cmd.Stderr = writer
-			if err := cmd.Run(); err != nil {
-				return fmt.Errorf("failed to execute ginkgo run '%s': %v. Output can be found at %s", run.name, err, logFile)
-			}
-			return nil
+	writer := bufio.NewWriter(file)
+	defer writer.Flush()
```
alvaroaleman, line 785: Not completely sure about how golang orders the execution of funcs in `defer` statements, but doesn't this mean that the file can be already closed when you call `write.Flush`?

Also we wont even see any errors about that in the logs as you just silently ignore them here (Shouldn't that make the linting fail btw?)

api/cmd/conformance-tests/runner.go
alvaroaleman: Actually I don't understand this condition at all. It will be always non-nil if `executeGinkgoRun` didn't fail, that doesn't tell anything about the [not-] success of the run

api/cmd/conformance-tests/runner.go
```diff
@@ -732,51 +739,94 @@ func (r *testRunner) runE2E(
...
+	}
+	if _, err := writer.Write([]byte(strings.Join(cmd.Args, argSeparator))); err != nil {
+		return nil, fmt.Errorf("failed to write command to log: %v", err)
+	}
+
+	// We're clearing up the temp dir on every run
+	if err := os.RemoveAll(run.reportsDir); err != nil {
+		return nil, fmt.Errorf("failed to clean temporary reports directory: %v", err)
+	}
+	if err := os.MkdirAll(run.reportsDir, os.ModePerm); err != nil {
```
alvaroaleman: This one confuses me for two reasons:

a) The report dir also holds logs of other Ginkgo run (e.G. if we are in the parallel run it already holds the logs of the serial run). Doesn't this mean that we just delete them?
b) If we cant to clean, wouldn't it be more logical to create the dir when entering `executeGinkgoRun` and deleting it when leaving, instead of `deleting+recreating` at the end?

Maybe the best is to always `delete+recreate` when entering `executeGinkgoRun`? Right now we don't clean it if we errord out before getting to this point here

api/cmd/conformance-tests/runner.go
mrIncompetent: Defer functions are popped onto a stack.

> Also we wont even see any errors about that in the logs as you just silently ignore them here (Shouldn't that make the linting fail btw?)

Do we actually care? We cannot handle the error correctly, nor is this a long running process. The file handle will be latest released when the process dies.

Not sure about linting though

api/cmd/conformance-tests/runner.go
mrIncompetent: > a) The report dir also holds logs of other Ginkgo run (e.G. if we are in the parallel run it already holds the logs of the serial run). Doesn't this mean that we just delete them?

No the current code just stores the logs if Ginkgo in a temp file unter /tmp
For storing the temporary ginkgo reports, a temp folder will be created (Used above).
Each ginkgo run gets its own temp folder, so nothing will be overwritten.
I'm cleaning that directory before the run to avoid running into issues with retries of certain ginkgo runs, as all retries use the same temp folder.

> Maybe the best is to always delete+recreate when entering executeGinkgoRun? Right now we don't clean it if we errord out before getting to this point here

Theoretically yes, but the ginkgo command needs to have the flag set to the individual folder.
The current logic is:
- Get all ginkgo runs (Parallel, Serial)
- Iterate over them
  - Execute run & Retry if needed

api/cmd/conformance-tests/runner.go
mrIncompetent: I updated the logic to make it more readable + added some comments:
https://github.com/kubermatic/kubermatic/pull/2563/commits/13fdf1b5ff4dca60b38fd7173dfda5e738346906

api/cmd/conformance-tests/runner.go
alvaroaleman: > Defer functions are popped onto a stack.

:+1: 

> Do we actually care? We cannot handle the error correctly, nor is this a long running process. The file handle will be latest released when the process dies.

For debugging purposes if something failed, but may be less likely here.

Would be nice if linting failed because there may be other places in which this matters

api/cmd/conformance-tests/runner.go
alvaroaleman: > No the current code just stores the logs if Ginkgo in a temp file unter /tmp

Ah okay.

> Theoretically yes, but the ginkgo command needs to have the flag set to the individual folder.

Sure, but that does not prevent us to do a `os.RemoveAll` directly followed by a `os.MkdirAll` when entering this func as opposed to when leaving it. It still means it is there when we execute Ginkgo

alvaroaleman (commented):
One nit, then its good to go

PR #2576, opened 2019-01-03
Make prometheues resource requests and limits configurable in cluster resource

bashofmann:

Make user cluster prometheues resource requests and limits configurable in cluster resource, similarly to etcd or apiserver.

**Release note**:

```release-note monitoring
Cluster Prometheus resource requests and limits are now configurable in cluster resource
```


alvaroaleman:
@bashofmann Does sys11 have a github org we can mark as trusted so you wont need the `/ok-to-test` in the future?

bashofmann:
Yes (https://github.com/syseleven), though our fork lives in our internal Gitlab and not on Github. Or is it enough if we are part of the organization (which we are) and the PR does not need to come a fork from there?

xrstf:
Nice work :-)

/lgtm
/approve

PR #2578, opened 2019-01-03
Blacklist upstream alerts

### LINKED ISSUE #2367: Fix CPUThrottlingHigh alert
**Description**:

Currently, the alert `CPUThrottlingHigh` is one of the major causes of alert fatigue. The alert occured 1912 times in the week of 12 Nov 2018 - 19 Nov 2018. Due to this, it is easy to miss out on the "real" alerts.

This is a known issue upstream, see the following issues for reference:
-  https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/108 and https://github.com/coreos/prometheus-operator/issues/2063 on the prometheus side
- https://github.com/kubernetes/kubernetes/issues/51135 is a lengthy discussion about the same. I've highlighted the relevant portions below.

The alerts are triggered correctly when the CPU throttles, but these throttling occurrences are unnecessary in reality. The CPU throttles due to a [bug](https://bugzilla.kernel.org/show_bug.cgi?id=198197) in the Linux Kernel, where even though the CFS quota isn't burned during the CFS period, throttling kicks in anyway. This [gist](https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1) describes the problem more intuitively.

**Possible Solutions**:

There are various solutions that we can do it silence the alerts. However, none of them would truly fix the underlying problem unless it's fixed in the kernel itself.

1. Bump the values of the cpu limits and `cpuThrottlingPercent`: https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/108#issuecomment-437875182

2. Make CFS quota period configurable through a flag in the kubelet: https://github.com/kubernetes/kubernetes/pull/63437. But this feature is only available as alpha from v1.12. 5ms is being used by Monzo: https://github.com/kubernetes/kubernetes/issues/51135#issuecomment-384908627 instead of the default 100ms.

3. Disable CPU limits entirely: https://github.com/kubernetes/kubernetes/issues/51135#issuecomment-373454012. However, this might not be such a good idea (https://github.com/kubernetes/kubernetes/issues/51135#issuecomment-385346661).

4. Disable CFS quota using `--cpu-cfs-quota=false` instead of the default the default 5ms period. See https://github.com/kubernetes/kubernetes/issues/67577#issuecomment-414936286 and https://twitter.com/mikkeloscar/status/1037443700513796103 for more details where people have used this.

xrstf:
Sometimes alerts turn out to be unstable, triggered by external bugs or just plain unfitting for a Kubermatic setup. This PR adds a way to blacklist certain alerts by their name so that we do not need to manually silence them (and document for customers that they should also silence an alert, which is kind of a weird thing to do in the first place).

Fixes #2367

**Release note**:
```release-note
disable CPUThrottlingHigh alert
```


xrstf:
/hold

This might be obsoleted by #2583

xrstf:
We don't need this anymore.

PR #2582, opened 2019-01-04
Force IAP pods to restart on config change

bashofmann:
Force IAP pods to restart on config change, otherwise a config change will not trigger a restart of the pod and the new config will not be applied.

**Release note**:

```release-note
NONE
```


maciaszczykm:
It seems like master build is failing since this change: https://drone.loodse.com/kubermatic/kubermatic/5759

PR #2583, opened 2019-01-04
Cleanup Alerts

### LINKED ISSUE #2209: Use kube_node_status_condition metric for monitoring nodes
**User Story**
Monitor and alert on `kube_node_status_condition{condition="Ready"}` metric in user-cluster prometheus

### LINKED ISSUE #2367: Fix CPUThrottlingHigh alert

### LINKED ISSUE #2462: Replace alert EtcdHighNumberOfFailedGRPCRequests
**User Story**
The `EtcdHighNumberOfFailedGRPCRequests` alert gets triggert for every restart of the Kubernetes API server as it seems to not gracefully shutdown etcd-watch connections.

The alert gets triggert because the API server might connect to a different etcd member, which will result in a 100% error rate on the old etcd member.

For a default etcd cluster this alert is fairly reasonable, but not in a scenario where only only member at a time is being used.

The alert should be replaced with an alert which expresses etcd connection errors from the API server side.

**Acceptance criteria**
- [ ] The `EtcdHighNumberOfFailedGRPCRequests` alert has been disabled
- [ ] A new alert got introduced which fires as soon as the API server has connection issues with the etcd
- [ ] The new alert does not get fired when the apiserver restarts
- [ ] The new alert does not get fired when any etcd member goes offline (Reconnects from the AP server are a common scenario)

xrstf:
This PR removes jsonnet as the source of alerts for Kubermatic, but (mostly) keeps the current set of alerts. The main reason for removing it was that customizing alerts is too cumbersome and sometimes not possible at all. Also, removing individual alerts that do not fit Kubermatic setups is ... ugly (see #2578).

The existing alerts have been moved into static YAML files inside the `rules/` directory and reformatted slightly. Only the CPU Throttling alert has been disabled so far. Having the alerts grouped in individual files makes it much, much easier to reason about them and also makes it possible to remove some by just removing that file.

---

Apart from removing jsonnet, this PR also doubles down on the "alert as soon as possible and as close to the source as possible" rule (reminder: We decided to go that route to scale better and reduce the load on the seed-level prometheus). The checks for "APIServer is gone" and such have thus been moved into the cluster namespaces and the `up` metrics are not federated up into the seed anymore. Likewise, machine-controller stuff has also been pushed down and only pre-aggregated metrics for worker and machine count are now federated.

---

Etcd failures are now handled by checking for EtcdHighNumberOfFailedThingies, but rather if dependent services experience issues. Testing showed that if etcd goes down, apiserver and controller-manager and such also go down and it's much easier and more stable to alert on those conditions than on the flaky etcd intermittent failures. To make this useful, this PR reduces the time from 15 to 3 minutes before sending an alert if apiserver/cm/... is down.

---

A final change is that namespaces with `prow-e2e-` in their name are not scraped anymore to reduce disk usage and CPU load on the seed-level Prometheus.

Fixes #2462, fixes #2367, fixes #2209

**Release note**:
```release-note monitoring
Alerts for for control-plane components now reside in cluster namespaces
```


kron4eg:
@xrstf so what are customizations points now? Where other people can plug-in?

/lgtm

xrstf:
> @xrstf so what are customizations points now? Where other people can plug-in?

How do you mean? They can still do what they were allowed to do before this PR: Place new rule files into `rules/`, remove existing files (now with greater fidelity), use jsonnet to generate their YAML files, ..

Am I overlooking something?

thetechnick (approved):

api/pkg/resources/prometheus/configmap-rules.go
```diff
@@ -142,141 +142,187 @@ groups:
...
+      severity: critical
+
+- name: kubernetes-nodes
+  rules:
+  - alert: KubernetesNodeNotReady
+    annotations:
+      message: '{{ $labels.node }} has been unready for more than an hour.'
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubenodenotready
+    expr: kube_node_status_condition{condition="Ready",status="true"} == 0
+    for: 1h
```
kron4eg: probably that timeout is too long.

xrstf: I copied it from the seed-level rule but would be fine to reduce it.

xrstf: Lowered it to 30 minutes.

config/monitoring/prometheus/rules/kubermatic.yaml
```diff
@@ -0,0 +1,20 @@
...
+    for: 10m
+    labels:
+      severity: warning
+
+  - alert: KubermaticClusterDeletionTakesTooLong
+    annotations:
+      message: Cluster {{ $labels.cluster }} is stuck in deletion for more than 30min.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubermaticclusterdeletiontakestoolong
+    expr: (time() - kubermatic_cluster_deleted) > 30*60
+    for: 0m
```
kron4eg, line 18: `0m` doesn't makes sense. Probably a typo?

xrstf: Also copied from the original code. 0m makes sense here because we already express the timeout in the `expr`.

config/monitoring/prometheus/rules/kubernetes-absent.yaml
```diff
@@ -0,0 +1,83 @@
...
+    expr: absent(up{job="apiserver"} == 1)
+    for: 15m
+    labels:
+      severity: critical
+
+  - alert: KubeControllerManagerDown
+    annotations:
+      message: KubeControllerManager has disappeared from Prometheus target discovery.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubecontrollermanagerdown
+    expr: absent(up{job="controller-manager"} == 1)
```
kron4eg, line 26: in GKE environment this alert will always fire. Not sure how to handle that... Probably we can silence it indefinitely.

config/monitoring/prometheus/rules/kubernetes-absent.yaml
```diff
@@ -0,0 +1,83 @@
...
+    for: 15m
+    labels:
+      severity: critical
+
+  - alert: KubeSchedulerDown
+    annotations:
+      message: KubeScheduler has disappeared from Prometheus target discovery.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubeschedulerdown
+    expr: absent(up{job="scheduler"} == 1)
```
kron4eg, line 35: same goes here

config/monitoring/prometheus/rules/kubernetes-absent.yaml
xrstf: As we do already, yep. Maybe at some point we can make it optional, but that should not be the scope of this PR.

config/monitoring/prometheus/rules/kubernetes-absent.yaml
xrstf: As we do already, yep. Maybe at some point we can make it optional, but that should not be the scope of this PR.

config/monitoring/prometheus/rules/kubernetes-absent.yaml
```diff
@@ -0,0 +1,83 @@
...
+  - alert: KubeletDown
+    annotations:
+      message: Kubelet has disappeared from Prometheus target discovery.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubeletdown
+    expr: absent(up{job="kubelet"} == 1)
+    for: 15m
+    labels:
+      severity: critical
+
+  - alert: KubermaticAPIDown
```
kron4eg: probably it's a good idea to split generic kubernetes alerts and kubermatic specific

config/monitoring/prometheus/rules/kubernetes-apps.yaml
```diff
@@ -0,0 +1,160 @@
...
+    for: 1h
+    labels:
+      severity: critical
+
+  - alert: KubePodNotReady
+    annotations:
+      message: Pod {{ $labels.namespace }}/{{ $labels.pod }} has been in a non-ready state for longer than an hour.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubepodnotready
+    expr: sum by (namespace, pod) (kube_pod_status_phase{job="kube-state-metrics", phase=~"Pending|Unknown"}) > 0
+    for: 1h
```
kron4eg: me personally would like to know about "stuck" pod a little bit earlier then 1 hour later it hang

config/monitoring/prometheus/rules/kubernetes-absent.yaml
xrstf: Can do, it's not better or worse than keeping all the "absent" rules in one file.

config/monitoring/prometheus/rules/kubernetes-apps.yaml
```diff
@@ -0,0 +1,160 @@
...
+      severity: warning
+
+  - alert: KubeJobFailed
+    annotations:
+      message: Job {{ $labels.namespaces }}/{{ $labels.job }} failed to complete.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubejobfailed
+    expr: kube_job_status_failed{job="kube-state-metrics"}  > 0
+    for: 1h
+    labels:
+      severity: warning
```
kron4eg, line 160: One very useful addition here would be `kube_cronjob_spec_suspend` metric. CronJobs are getting suspended once it failed N times in raw, and it's hugely useful to know about this :)

xrstf: => future PR

config/monitoring/prometheus/rules/kubernetes-system.yaml
```diff
@@ -0,0 +1,119 @@
+groups:
+- name: kubernetes-system
+  rules:
+  - alert: KubeNodeNotReady
+    annotations:
+      message: '{{ $labels.node }} has been unready for more than an hour.'
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubenodenotready
+    expr: kube_node_status_condition{job="kube-state-metrics",condition="Ready",status="true"} == 0
```
kron4eg, line 8: This looks like duplicate or `KubernetesNodeNotReady` alert

xrstf: Yeah but it's the one for the seed cluster.

config/monitoring/prometheus/rules/kubernetes-system.yaml
```diff
@@ -0,0 +1,119 @@
...
+    labels:
+      severity: warning
+
+  - alert: KubeClientErrors
+    annotations:
+      message: >
+        Kubernetes API server client {{ $labels.job }}/{{ $labels.instance }} is experiencing
+        {{ printf "%0.0f" $value }} errors / second.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubeclienterrors
+    expr: sum(rate(ksm_scrape_error_total{job="kube-state-metrics"}[5m])) by (instance, job) > 0.1
```
kron4eg: Does this alert message match the expression?

config/monitoring/prometheus/rules/node.rules.yaml
```diff
@@ -0,0 +1,179 @@
...
+      sum(irate(node_network_transmit_drop{app="node-exporter",device="eth0"}[1m]))
+
+  - record: node:node_net_saturation:sum_irate
+    expr: |
+      sum by (node) (
+        (irate(node_network_receive_drop{app="node-exporter",device="eth0"}[1m]) +
+        irate(node_network_transmit_drop{app="node-exporter",device="eth0"}[1m]))
+      * on (namespace, pod) group_left(node)
+        node_namespace_pod:kube_pod_info:
+      )
```
kron4eg, line 179: something bad happening with naming of those recording rules, they are pretty inconsistent. Some are named with `node:` prefix and some are not. Should we bring them to the single form?

xrstf: That's all fine. Some of the metrics are grouped by node, some are not. Those which are not are not prefixed with "node:", so they start with ":....". Looks weird, but it's consistent with Prometheus' naming rules.

Issue #2439, opened 2018-11-27
[OpenStack] Make allocation of FloatingIP's optional


mrIncompetent:
**User Story**
As a user of OpenStack, i might not wan't a Floating IP for all my nodes.
Currently all Kubermatic nodes though get one - as the relevant flag in the Machine, which kubermatic creates, is hardcoded.

As a FloatingIP though is mandatory for some OpenStack environments, it must be possible to enforce the allocation of FloatingIP's on a datacenter level.

**Acceptance criteria**
- [x] A user can decide via a CheckBox in the Dashboard to allocate FloatingIP's for the nodes
- [x] A setting in the datacenters.yaml can enforce the allocation of FloatingIP's
- [ ] If the datacenter is configured to always allocate FloatingIP's, the Dashboard won't show the option

p0lyn0mial:
@mrIncompetent  Do we want to store a list of `FloatingIP` in the `datacenters.yaml` file ? If yes would it be okay to sort the list and assign the first one from the list to all nodes ?

mrIncompetent:
@p0lyn0mial 

> Do we want to store a list of FloatingIP in the datacenters.yaml file ?

No. How did you came up with this idea?

p0lyn0mial:
> No. How did you came up with this idea?

@mrIncompetent Because during a cluster creation a user can choose a `FloatingIP` from the list (a dropdown menu). Should we hide/disable the dropdown when there is a value in the datacenters.yaml file ?

mrIncompetent:
@p0lyn0mial Do you mean the `Floating IP pool` which can be selected in the extended view when creating a OpenStack cluster?

p0lyn0mial:
> @p0lyn0mial Do you mean the Floating IP pool which can be selected in the extended view when creating a OpenStack cluster?

@mrIncompetent yes, this issue is about something else :) ?

floreks:
@mrIncompetent should `datacenters.yaml` be updated to enforce floating IP usage for any of our providers or just leave it as is for now?

PR #2585, opened 2019-01-07
Add option to enforce floating IP usage for openstack

### LINKED ISSUE #2544: [OpenStack] Extend OpenStackSpec with FloatingIP field
**User Story** 
We need to introduce a new field to the `OpenStackSpec` struct to inform us if floating IP usage is enforced by the datacenter. 

Should it be called `enforceFloatingIP` boolean or differently?

**Acceptance criteria**
 - [x] Add new field to the `api/pkg/provider/datacenter.go:OpenstackSpec` struct
 - [x] Add new field to the `api/pkg/api/v1/types.go:OpenstackDatacenterSpec` struct
 - [x] Adapt datacenter handler `apiSpec` method to use new field
 - [x] Adapt related tests

floreks:

I have added an option to enforce floating IP usage on the datacenter level. Additionally, when creating a node it will be possible to specify `useFloatingIP` field. The user will be able to decide if he wants to assign a floating IP to a new node, but only in case, it is not enforced by the datacenter first.

Related to #2439 
Fixes #2544 

#### TODO:
 - [x] Fix existing tests
 - [x] Add/adapt some tests

**Release note**:

```release-note
NONE
```

Should I add a note about adding new field to the API struct?


api/pkg/api/v1/types.go
```diff
@@ -563,6 +568,9 @@ type OpenstackNodeSpec struct {
 	// Additional metadata to set
 	// required: false
 	Tags map[string]string `json:"tags,omitempty"`
+	// Defines whether floating ip should be used
+	// required: false
+	UseFloatingIP bool `json:"useFloatingIP,omitempty"`
```
p0lyn0mial, line 573: I'm not sure if I understand why we need this filed, is it because we read it in "Add Node" page ? If yes, isn't that field redundant ? The dashboard knows it because it has access to `Datacenters` and `CloudSpec` files

api/pkg/api/v1/types.go
```diff
@@ -407,14 +408,18 @@ func newPublicAWSCloudSpec(internal *kubermaticv1.AWSCloudSpec) (public *PublicA
...
+}
 
 func newPublicOpenstackCloudSpec(internal *kubermaticv1.OpenstackCloudSpec) (public *PublicOpenstackCloudSpec) {
 	if internal == nil {
 		return nil
 	}
 
-	return &PublicOpenstackCloudSpec{}
+	return &PublicOpenstackCloudSpec{
+		FloatingIPPool: internal.FloatingIPPool,
```
p0lyn0mial, line 421: once set all future nodes will have the same `IPPool` assigned ?

p0lyn0mial (changes requested):
I don't see validation code for "adding" and "updating" which should check if an `IPPool` has been assigned when required ("EnforceFloatingIP").

api/pkg/provider/datacenter_test.go
```diff
@@ -33,7 +34,28 @@ datacenters:
...
+#============OpenStack=============
+#==================================
+  syseleven-dbl1:
+    location: Syseleven - dbl1
+    seed: europe-west3-c
+    country: DE
+    provider: openstack
+    spec:
+      openstack:
+        auth_url: https://api.cbk.cloud.syseleven.net:5000/v3
```
mrIncompetent: Please do not hardcode customer names/addresses in our codebase :)

api/pkg/provider/datacenter_test.go
```diff
@@ -33,7 +34,28 @@ datacenters:
...
+        availability_zone: dbl1
+        region: dbl
+        dns_servers:
+        - "37.123.105.116"
+        - "37.123.105.117"
```
mrIncompetent: Please use the google DNS servers:
- 8.8.8.8
- 8.8.4.4

api/pkg/provider/datacenter_test.go
```diff
@@ -33,7 +34,28 @@ datacenters:
     country: NL
     spec:
       digitalocean:
-        region: ams3`
+        region: ams3
+#==================================
+#============OpenStack=============
+#==================================
+  syseleven-dbl1:
+    location: Syseleven - dbl1
```
mrIncompetent: See https://github.com/kubermatic/kubermatic/pull/2585/files#r245946182

api/pkg/provider/datacenter_test.go
```diff
@@ -59,6 +81,28 @@ datacenters:
 			IsSeed:           false,
 			SeedDNSOverwrite: nil,
 		},
+		"syseleven-dbl1": {
+			Location: "Syseleven - dbl1",
+			Seed:     "europe-west3-c",
+			Country:  "DE",
+			Spec: DatacenterSpec{
+				Openstack: &OpenstackSpec{
+					AuthURL:          "https://api.cbk.cloud.syseleven.net:5000/v3",
```
mrIncompetent: See https://github.com/kubermatic/kubermatic/pull/2585/files#r245946182

api/pkg/provider/datacenter_test.go
floreks: Damn, you are right. Sorry.

Removed.

api/pkg/api/v1/types.go
floreks: This structure is sent to the backend when creating a new node. In case datacenter does not enforce floating IP usage, this field directly reflects users' choice whether or not newly created nodes should have a floating IP assigned.

api/pkg/api/v1/types.go
floreks: Actually, no. In [here](https://github.com/kubermatic/dashboard-v2/pull/943/files#diff-92ed9bff35147125e4b965386f6efe7cR49) frontend decides if added checkbox should be disabled based on this information.

Basically, if floating ip pool will not be specified at all, then checkbox will be unchecked and disabled when creating a new node. In case it will be specified, but not enforced by datacenter, then user can decide whether or not he wants to assign a floating IP to the created node.

api/pkg/api/v1/types.go
p0lyn0mial: oh, so the user can decide not only on a cluster level but also on a node level.

api/pkg/validation/cluster.go
```diff
@@ -246,6 +246,9 @@ func ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 		if spec.Openstack.Tenant == "" {
 			return errors.New("no tenant specified")
 		}
+		if spec.Openstack.FloatingIPPool == "" && dc.Spec.Openstack != nil && dc.Spec.Openstack.EnforceFloatingIP {
```
floreks, line 249: It would be nice to introduce some kind of mini validation framework to simplify code and validation process. Right now it is split into multiple places and a bit confusing. In example [here](https://github.com/kubermatic/kubermatic/blob/master/api/pkg/validation/cluster.go#L187) and [here](https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L51).

floreks:
PTAL

api/pkg/validation/cluster.go
mrIncompetent: Please create an issue for that so we keep track of this technical dept

mrIncompetent:
/approve

code looks good, but i'll leave the lgtm to @p0lyn0mial

api/pkg/validation/cluster.go
floreks: Done https://github.com/kubermatic/kubermatic/issues/2600

PR #2586, opened 2019-01-08
drop the AllowedNodeVersions field

### LINKED ISSUE #2543: Drop the AllowedNodeVersions field
**User Story**
The `AllowedNodeVersions` was never really used except for passing it to the Dashboard that did not use it either. With the new controlplane-kubelet version checking, the field is being deprecated. 

**Acceptance criteria**
- [ ] Remove the `AllowedNodeVersions` of `pkg/version.MasterVersion`.
- [ ] Remove the `AllowedNodeVersions` of `pkg/api/v1.MasterVersion`.
- [ ] remove the `allowed-node-versions` section and `allowedNodeVersions` fields in `versions.yaml`

zreigz:
**What this PR does / why we need it**: the AllowedNodeVersions was never really used except for passing it to the Dashboard that did not use it either. With the new controlplane-kubelet version checking, the field is being deprecated.

Fixes #2543 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -3501,13 +3501,6 @@
       "description": "MasterVersion describes a version of the master components",
       "type": "object",
       "properties": {
-        "allowedNodeVersions": {
-          "type": "array",
-          "items": {
-            "type": "string"
-          },
-          "x-go-name": "AllowedNodeVersions"
-        },
         "default": {
           "type": "boolean",
           "x-go-name": "Default"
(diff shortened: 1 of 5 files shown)


kdomanski:
Will it be okay to remove the field from the API response while it's still present in the data structure on the Dashboard's side?

zreigz:
@maciaszczykm can we  remove `AllowedNodeVersions` from the backend without impacting for the Dashboard

maciaszczykm:
@zreigz Looks like it should be safe. There are some leftovers but not much of them (entity & some mocks). I think you should check locally with `npm run start:local` and your branch.

zrzut ekranu 2019-01-8 o 13 56 19

I think we should do the same on the Dashboard side if it will work.

zreigz:
@p0lyn0mial I've checked it locally with dashboard and works fine

PR #2588, opened 2019-01-08
Implementing an endpoint for project update (rename)

### LINKED ISSUE #2510: Update project HTTP endpoint
**User Story** As a user of kubermatic I would like to change the name of already existing projects.

**Acceptance criteria**
 - changing the name of the project works
 - there is a unit test
 - make sure that the name is not taken

glower:
With the new HTTP endpoint it is possible to rename the project

Fixes #2510

**Documentation**:

**Release note**:

```release-note
Endpoint PUT /api/v1/projects/{project_id} for a renaming of the project is implemented
```


api/hack/run-api.sh
```diff
@@ -24,7 +24,7 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
   -worker-name="$(tr -cd '[:alnum:]' <<< $KUBERMATIC_WORKERNAME | tr '[:upper:]' '[:lower:]')" \
   -internal-address=127.0.0.1:18085 \
   -prometheus-url=http://localhost:9090 \
-  -address=127.0.0.1:8080 \
+  -address=127.0.0.1:8081 \
```
kdomanski: Why?

glower: it was not intended, will revert this

glower: done

api/pkg/provider/kubernetes/project.go
```diff
@@ -92,6 +92,38 @@ func (p *ProjectProvider) New(user *kubermaticapiv1.User, projectName string) (*
 	return p.clientPrivileged.Create(project)
 }
 
+// Rename renames the given project as the given user
+// and make sure that the name is not taken for other projects of the user
+func (p *ProjectProvider) Rename(user *kubermaticapiv1.User, projectInternalName, newProjectName string) (*kubermaticapiv1.Project, error) {
```
p0lyn0mial: could we change the name and the signature of this method so that it knows how to update project resource not only the name ? `For example Update(userInfo *UserInfo, newProject *kubermaticv1.Project) (*kubermaticv1.Project, error)`

have a look on these as a reference
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L92
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L139

api/pkg/provider/kubernetes/project.go
```diff
@@ -92,6 +92,38 @@ func (p *ProjectProvider) New(user *kubermaticapiv1.User, projectName string) (*
...
+	if user == nil {
+		return nil, errors.New("a user is missing but required")
+	}
+	projects, err := p.projectLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+	var renamedProject *kubermaticapiv1.Project
+	for _, project := range projects {
```
p0lyn0mial: could we extract this logic to a separate method `List` and use it in the handler ?
similarly to https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L127 and https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L86

p0lyn0mial (changes requested):

api/pkg/provider/kubernetes/project.go
glower: Yes, I was not sure about this and just implemented the rename functionality. It makes sense to refactor this to the update function. Will do this

api/pkg/provider/kubernetes/project.go
glower: Yes, I will have a look at how we can simplify the logic here

api/pkg/provider/kubernetes/project.go
glower: done

api/pkg/provider/kubernetes/project.go
glower: done

glower:
Some tests are not working correctly so I added a skip to fix it later. I think it is because our fake client not checking RBAC rules (not sure about it). I will create a follow-up task for rather fix the tests or remove it.

kdomanski:
If the tests are not working then the code is not working. Will not approve until this is resolved.

glower:
After discussion with @p0lyn0mial we decided to remove these tests, because rbac roles are not enforced/checked in unit tests

api/pkg/provider/kubernetes/project.go
```diff
@@ -92,6 +78,58 @@ func (p *ProjectProvider) New(user *kubermaticapiv1.User, projectName string) (*
 	return p.clientPrivileged.Create(project)
 }
 
+// List returns a list of projects that belong to a given user
+// using UserProjectsListOptions you can filter by a specific name of the project
+func (p *ProjectProvider) List(user *kubermaticapiv1.User, options *provider.UserProjectsListOptions) ([]*kubermaticapiv1.Project, error) {
```
p0lyn0mial: I'm sorry but i have just realised that it would be better to use `userInfo *provider.UserInfo` instead of `user *kubermaticapiv1.User` because once we have service accounts they could also map to `UserInfo` which holds only an email and the group and no further changes in the code would be required. 

Do you think we could use `ProjectMemberProvider.List` function https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/member.go#L66 to get a list of project a user belongs to ? We could extend `provider.ProjectMemberListOptions` to filter only the owners.

api/pkg/handler/project.go
```diff
@@ -81,9 +90,47 @@ func deleteProjectEndpoint(projectProvider provider.ProjectProvider) endpoint.En
...
+		}
+
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
+
+		kubermaticProject, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{IncludeUninitialized: true})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		if len(req.Name) != 0 {
```
p0lyn0mial: nit: not needed it has to be `> 0` because of line 104 ?

api/pkg/provider/kubernetes/project.go
glower: Yes, this makes sense. I will rework the code

api/pkg/handler/project.go
```diff
@@ -176,3 +226,43 @@ func decodeDeleteProject(c context.Context, r *http.Request) (interface{}, error
 	}
 	return DeleteProjectRq{ProjectReq: req.(common.ProjectReq)}, err
 }
+
+// UserProjectsListOptions allows to set filters for listing user projects
+type UserProjectsListOptions struct {
```
p0lyn0mial: this doesn't have to be exported.

api/pkg/handler/project.go
```diff
@@ -176,3 +226,43 @@ func decodeDeleteProject(c context.Context, r *http.Request) (interface{}, error
...
+	// ProjectName set the project name for the given user
+	ProjectName string
+}
+
+func listUserProjects(email string, projectProvider provider.ProjectProvider, memberMapper provider.ProjectMemberMapper, options *UserProjectsListOptions) ([]*kubermaticapiv1.Project, error) {
```
p0lyn0mial: I think that for `create` and `update` endpoints we need to get (filter) only projects owned by the user.

api/pkg/handler/project_test.go
```diff
@@ -17,6 +17,151 @@ import (
...
+			ExistingAPIUser: apiv1.User{
+				ObjectMeta: apiv1.ObjectMeta{
+					ID: testUserName,
+				},
+				Email: testUserEmail,
+			},
+			ExpectedResponse: `{"error":{"code":403,"message":"forbidden: The user \"john@acme.com\" doesn't belong to the given project = some-ID"}}`,
+		},
+		{
+			Name:            "scenario 4: rename a project where user is an editor",
```
p0lyn0mial: we could remove this test scenario because the test framework doesn't check if the user is an editor or an owner only kube-api server. So this is the same  as `scenario 2`

api/pkg/handler/user.go
```diff
@@ -291,11 +291,13 @@ func (r Routing) userInfoMiddleware() endpoint.Middleware {
...
 				return nil, k8cerrors.New(http.StatusInternalServerError, "unable to get authenticated user object")
 			}
+
+			var projectID string
 			prjIDGetter, ok := request.(common.ProjectIDGetter)
-			if !ok {
-				return nil, k8cerrors.NewBadRequest("you can only use userInfoMiddleware for endpoints that interact with project")
+			if ok {
+				projectID = prjIDGetter.GetProjectID()
+				// return nil, k8cerrors.NewBadRequest("you can only use userInfoMiddleware for endpoints that interact with project")
```
p0lyn0mial: rm not used code.

api/pkg/handler/project.go
glower: make sense, done

api/pkg/handler/project.go
glower: fixed

api/pkg/handler/user.go
glower: fixed

api/pkg/handler/project.go
```diff
@@ -81,9 +85,43 @@ func deleteProjectEndpoint(projectProvider provider.ProjectProvider) endpoint.En
...
+		}
+
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
+
+		kubermaticProject, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{IncludeUninitialized: true})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		existingProject, err := listUserProjects(userInfo.Email, projectProvider, memberMapper, &userProjectsListOptions{ProjectName: req.Name, Group: "owners"})
```
p0lyn0mial: please use `rbac.OwnerGroupNamePrefix` instead.

api/pkg/handler/project.go
```diff
@@ -176,3 +227,47 @@ func decodeDeleteProject(c context.Context, r *http.Request) (interface{}, error
...
+
+func listUserProjects(email string, projectProvider provider.ProjectProvider, memberMapper provider.ProjectMemberMapper, options *userProjectsListOptions) ([]*kubermaticapiv1.Project, error) {
+	userMappings, err := memberMapper.MappingsFor(email)
+	if err != nil {
+		return nil, err
+	}
+
+	userProjects := []*kubermaticapiv1.Project{}
+	for _, mapping := range userMappings {
+		if len(options.Group) > 0 && !strings.HasPrefix(mapping.Spec.Group, options.Group) {
```
p0lyn0mial: please use `rbac.ExtractGroupPrefix` and compare prefixes.

api/pkg/handler/project.go
glower: fixed

api/pkg/handler/project_test.go
glower: fixed

api/pkg/handler/project.go
```diff
@@ -81,9 +81,43 @@ func deleteProjectEndpoint(projectProvider provider.ProjectProvider) endpoint.En
...
+
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
+		user := ctx.Value(middleware.UserCRContextKey).(*kubermaticapiv1.User)
+
+		kubermaticProject, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{IncludeUninitialized: true})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		alreadyExistingProjects, err := projectProvider.List(&provider.ProjectListOptions{ProjectName: req.Name, OwnerUID: user.UID})
```
p0lyn0mial, line 107: @glower i changed the endpoint to get the list of projects directly from the cache - as we agreed offline.

glower:
lgtm

PR #2592, opened 2019-01-08
Correctly scope IAP chart checksums

### LINKED ISSUE #2591: Master build is failing
See: https://github.com/kubermatic/kubermatic/pull/2582#issuecomment-452349393

```
Error: UPGRADE FAILED: render error in "iap/templates/deployments.yaml": template: iap/templates/deployments.yaml:22:28: executing "iap/templates/deployments.yaml" at <include (print $.Tem...>: error calling include: template: iap/templates/configmaps.yaml:1:17: executing "iap/templates/configmaps.yaml" at <.Values.iap.deployme...>: can't evaluate field iap in type interface {}
```

Last 3 build have failed.

bashofmann:
Fixes #2591

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/iap/templates/deployments.yaml
+++ b/config/iap/templates/deployments.yaml
@@ -1,4 +1,3 @@
-{{- $root := . -}}
 {{- range .Values.iap.deployments }}
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


xrstf:
Can we just use `$` instead of declaring a new variable?

bashofmann:
@xrstf Does it still work then, that the hash changes if value that is included in the template changes?

xrstf:
Yes, I just checked. Changing the name of a deployment yields a new hash. As documented in https://github.com/helm/helm/blob/master/docs/chart_template_guide/variables.md:

> However, there is one variable that is always global - `$` - this variable will always point to the root context. This can be very useful when you are looping in a range need to know the chart's release name.

bashofmann:
Makes sense, that's of course also why `$.Template.BasePath` works.

PR #2608, opened 2019-01-10
Cleanup e2e tests on timeout

alvaroaleman:

This PR makes sure that the cleanup in the e2e tests gets called, even when they timeout and hence [get a SIGINT from Prow](https://github.com/kubernetes/test-infra/blob/c23317e88f679ccadf6922e1cb51dfd58f35a3e6/prow/entrypoint/run.go#L228)

This requires wrapping everything in a subshell as Bash does not forward signals and can not react to signals via traps as long as a child is running. By wrapping in a subshell that is forked we circumvent this by just running short sleeps and waiting for the cleanup to be done which implies that either the subshell finished (successfully or not) or we got a SIGINT

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

/assign @thz 
/cc @mrIncompetent @nikhita

### DIFF
--- a/api/hack/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci-run-minimal-conformance-tests.sh
@@ -58,6 +58,20 @@ function cleanup {
 }
 trap 'cleanup 2' INT
 
+echodate "Getting secrets from Vault"
+export VAULT_ADDR=https://vault.loodse.com/
+export VAULT_TOKEN=$(vault write \
+  --format=json auth/approle/login \
+  role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID \
+  | jq .auth.client_token -r)
+vault kv get -field=kubeconfig \
+  dev/seed-clusters/dev.kubermatic.io > /tmp/kubeconfig
+vault kv get -field=values.yaml \
+  dev/seed-clusters/dev.kubermatic.io > /tmp/values.yaml
+export KUBECONFIG=/tmp/kubeconfig
+export VALUES_FILE=/tmp/values.yaml
+echodate "Successfully got secrets from Vault"
+
 # Bash can not forward signals and doesn't react to signals
 # when a command is running, so we have to move everything into
 # a subshell and fork it, then do short sleeps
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/hold

We should discuss/brainstorm if this is the best possible approach

api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -2,16 +2,34 @@
...
+  # makes the subshell enter this func again because of
+  # the EXIT trap
+  if [ -e /tmp/cleanup-going ]; then return; fi
+  touch /tmp/cleanup-going
+
+  echodate "Starting cleanup"
+
+  # Kill all sub-processes to make sure they don't create
+  # any new resources and dont block termination of the job
+  jobs -p|xargs -r -I ^ kill -9 ^ &>dev/null
```
thz: missing / for `/dev/null`?

api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -58,26 +70,41 @@ vault kv get -field=values.yaml \
...
+# a subshell and fork it, then do short sleeps
+(
+# Enable the exit trap for the subshell
+trap 'cleanup $?' exit
 
+docker ps &>/dev/null || start-docker.sh
+
+echodate "Unlocking secrets repo"
+cd $(go env GOPATH)/src/github.com/kubermatic/secrets
+echo $KUBERMATIC_SECRETS_GPG_KEY_BASE64 | base64 -d > /tmp/git-crypt-key
```
thz, line 86: are you really sure /tmp is on tmpfs and not aufs? I would like to be sure that secrets are never accidently on disk.

api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -164,3 +191,13 @@ echo "Starting conformance tests"
   -versions="$VERSIONS" \
   -providers=aws \
   -exclude-distributions="ubuntu,centos"
+
+# Test if cleanup on timeout works
+sleep 1d
```
thz: this sleep is just for testing and going to be removed, right?

alvaroaleman: yes

api/hack/ci-run-minimal-conformance-tests.sh
alvaroaleman: Not aufs because not even GKE is that insane but not tmpfs either. I will fixup the job to put a tmpfs on /tmp

mrIncompetent:
@alvaroaleman @thz @nikhita would you be ok with brainstorming this after the daily on tuesday?

alvaroaleman:
> @alvaroaleman @thz @nikhita would you be ok with brainstorming this after the daily on tuesday?

SGTM

alvaroaleman:
We will simply kill the conformance tests via timeout after a timeout instead, see #2685

Issue #2359, opened 2018-11-15
 Use the “underscore test” for the handler package


zreigz:
**Description**
Go doesn’t let you include multiple packages in one folder except for when you use a separate test package.

Using a separate `_test` package means that user can’t access the unexported fields and functions in the package. The test can see only exported API, so it makes the code more usable, safe and complete.

maciaszczykm:
+1 from me. Like I said in https://github.com/kubermatic/kubermatic/pull/2348#discussion_r233378650 it would be good to use it in all packages in my opinion.

p0lyn0mial:
+1

PR #2611, opened 2019-01-11
move cluster and node test to _test package

zreigz:
**What this PR does / why we need it**:  this is the fourth step to moving unit tests from handler package to handler_test package

The issue for this task: #2359 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/handler/cluster_test.go
+++ b/api/pkg/handler/cluster_test.go
@@ -1,4 +1,4 @@
-package handler
+package handler_test
 
 import (
 	"encoding/json"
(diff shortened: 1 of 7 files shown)


api/pkg/handler/cluster_test.go
```diff
@@ -12,6 +12,7 @@ import (
```
p0lyn0mial, line 12: shouldn't we change the package name for this file to `handler_test` ?

api/pkg/handler/node_v1_test.go
```diff
@@ -12,8 +12,10 @@ import (
 	"time"
```
p0lyn0mial, line 13: same as https://github.com/kubermatic/kubermatic/pull/2611/files#r247130307

zreigz: yes, it was the purpose, please hold on

zreigz:
@p0lyn0mial PTAL

api/pkg/handler/cluster_test.go
zreigz: done

api/pkg/handler/node_v1_test.go
zreigz: fixed

PR #2614, opened 2019-01-12
Use encrypted weave on Hetzner

alvaroaleman:

Fixes #

I'd love to rename the `canal` addon to something more generic like `networking` but I think this will cause issues for existing clusters, because they already have a `canal` addon which we can't remove because that will create service disruptions - And allowing both the `canal` and a new `networking` addon would mean we can never change anything in that new addon because that would mean the two fight each other in older clusters

**Documentation**:

**Release note**:

```release-note
Clusters created on Hetzner will now by default use encrypted Weave as CNI
```

/hold

Addon image tag has to get bumped

---
This change is [Reviewable](https://reviewable.io/reviews/kubermatic/kubermatic/2614)

### DIFF
--- a/addons/canal/configmap.yaml
+++ b/addons/canal/configmap.yaml
@@ -1,3 +1,4 @@
+{{ if not .Cluster.Spec.Cloud.Hetzner }}
 # This ConfigMap can be used to configure a self-hosted Canal installation.
 kind: ConfigMap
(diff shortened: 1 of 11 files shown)


alvaroaleman:
Closing, we don't want different addons hidden in the same addon

PR #2647, opened 2019-01-17
new endpoint for node deployment events

### LINKED ISSUE #2635: Implement new endpoint in backend for node deployment events
**User Story**
Add a new endpoint for backend: `GET /api/v1/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/nodedeployments/{nodedeployment_id}/nodes/events`

**Acceptance criteria**
Lists node deployment events. If the query parameter `warning` is set to true then only warning events are retrieved.
If the value is false then normal events are returned. If the query parameter is missing method returns all events.

zreigz:
**What this PR does / why we need it* added new endpoint for node deployment events: `/api/v1/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/nodedeployments/{nodedeployment_id}/nodes/events`
Lists node deployment events. If the query parameter `type` is set to `warning` then only warning events are retrieved. If the value `normal`  then normal events are returned. If the query parameter is missing method returns all events.

Fixes #2635 

**Release note**:
```release-note
Added new edpoint: `/api/v1/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/nodedeployments/{nodedeployment_id}/nodes/events` with query parameter `type`. When `?type=warning` then endpoint returns only warning type events. When `?type=normal` then endpoint returns only normal type events. Without query parameters the endpoint returns all events.
```

### DIFF
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -667,8 +667,8 @@ type NodeDeploymentSpec struct {
 
 // EventList is an events response structure.
 type EventList struct {
-	// The object that those events are about.
-	InvolvedObjectName string `json:"involvedObjectName"`
+	// The object name that those events are about.
+	Name string `json:"Name"`
 
 	// List of events
 	Events []Event `json:"events"`
@@ -676,10 +676,6 @@ type EventList struct {
 
 // Event is a report of an event somewhere in the cluster.
 type Event struct {
-	// This should be a short, machine understandable string that gives the reason
-	// for the transition into the object's current status.
-	Reason string `json:"reason,omitempty"`
-
 	// A human-readable description of the status of this operation.
 	Message string `json:"message,omitempty"`
 
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -701,7 +701,7 @@ type Event struct {
 	// The number of times this event has occurred.
 	Count int32 `json:"count,omitempty"`
 
-	// Type of this event (Normal, Warning), new types could be added in the future
+	// Type of this event (normal, warning), new types could be added in the future
 	Type string `json:"type,omitempty"`
 
 	// The time at which the event was first recorded.
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -665,29 +665,17 @@ type NodeDeploymentSpec struct {
 	Paused *bool `json:"paused,omitempty"`
 }
 
-// ObjectReference contains enough information to let you inspect or modify the referred object.
-type ObjectReference struct {
-	// Kind of the referent.
-	Kind string `json:"kind,omitempty"`
+// EventList is an events response structure.
+type EventList struct {
+	// The object that those events are about.
+	InvolvedObjectName string `json:"involvedObjectName"`
 
-	// Namespace of the referent.
-	Namespace string `json:"namespace,omitempty"`
-
-	// Name of the referent.
-	Name string `json:"name,omitempty"`
+	// List of events
+	Events []Event `json:"events"`
 }
 
 // Event is a report of an event somewhere in the cluster.
 type Event struct {
-	// Namespace of the event.
-	Namespace string `json:"namespace,omitempty"`
-
-	// Name of the event.
-	Name string `json:"name,omitempty"`
-
-	// The object that this event is about.
-	InvolvedObject ObjectReference `json:"involvedObject"`
-
 	// This should be a short, machine understandable string that gives the reason
 	// for the transition into the object's current status.
 	Reason string `json:"reason,omitempty"`
@@ -695,18 +683,6 @@ type Event struct {
 	// A human-readable description of the status of this operation.
 	Message string `json:"message,omitempty"`
 
-	// The component reporting this event. Should be a short machine understandable string.
-	Source string `json:"source,omitempty"`
-
-	// The number of times this event has occurred.
-	Count int32 `json:"count,omitempty"`
-
 	// Type of this event (normal, warning), new types could be added in the future
 	Type string `json:"type,omitempty"`
-
-	// The time at which the event was first recorded.
-	FirstTimestamp Time `json:"firstTimestamp,omitempty"`
-
-	// The time at which the most recent occurrence of this event was recorded.
-	LastTimestamp Time `json:"lastTimestamp,omitempty"`
 }
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -2,7 +2,6 @@ package v1
 
 import (
 	"encoding/json"
-
 	"github.com/Masterminds/semver"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
@@ -664,4 +663,50 @@ type NodeDeploymentSpec struct {
 	Template NodeSpec `json:"template"`
 
 	Paused *bool `json:"paused,omitempty"`
+}
+
+// ObjectReference contains enough information to let you inspect or modify the referred object.
+type ObjectReference struct {
+	// Kind of the referent.
+	Kind string `json:"kind,omitempty"`
+
+	// Namespace of the referent.
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name of the referent.
+	Name string `json:"name,omitempty"`
+}
+
+// Event is a report of an event somewhere in the cluster.
+type Event struct {
+	// Namespace of the event.
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name of the event.
+	Name string `json:"name,omitempty"`
+
+	// The object that this event is about.
+	InvolvedObject ObjectReference `json:"involvedObject"`
+
+	// This should be a short, machine understandable string that gives the reason
+	// for the transition into the object's current status.
+	Reason string `json:"reason,omitempty"`
+
+	// A human-readable description of the status of this operation.
+	Message string `json:"message,omitempty"`
+
+	// The component reporting this event. Should be a short machine understandable string.
+	Source string `json:"source,omitempty"`
+
+	// The number of times this event has occurred.
+	Count int32 `json:"count,omitempty"`
+
+	// Type of this event (Normal, Warning), new types could be added in the future
+	Type string `json:"type,omitempty"`
+
+	// The time at which the event was first recorded.
+	FirstTimestamp Time `json:"firstTimestamp,omitempty"`
+
+	// The time at which the most recent occurrence of this event was recorded.
+	LastTimestamp Time `json:"lastTimestamp,omitempty"`
 }
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -668,7 +668,7 @@ type NodeDeploymentSpec struct {
 // EventList is an events response structure.
 type EventList struct {
 	// The object name that those events are about.
-	Name string `json:"Name"`
+	Name string `json:"name"`
 
 	// List of events
 	Events []Event `json:"events"`
--- a/api/pkg/handler/event.go
+++ b/api/pkg/handler/event.go
@@ -19,7 +19,7 @@ func GetMachineEvents(client kubernetes.Interface, machine clusterv1alpha1.Machi
 		return nil, err
 	}
 
-	return CreateEventList(events.Items), nil
+	return createEventList(events.Items), nil
 }
 
 // FilterEventsByType filters kubernetes API event objects based on event type.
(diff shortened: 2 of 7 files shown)


zreigz:
PTAL

api/pkg/handler/event.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, event := range events {
+		kubermaticEvent := ToEvent(event)
+		kubermaticEvents = append(kubermaticEvents, kubermaticEvent)
+	}
+	return kubermaticEvents
+}
+
+// ToEvent converts event api Event to Event model object.
+func ToEvent(event corev1.Event) v1.Event {
```
maciaszczykm: Can be private?

api/pkg/handler/event.go
```diff
@@ -0,0 +1,73 @@
...
+		if event.Type == eventType {
+			result = append(result, event)
+		}
+	}
+
+	return result
+}
+
+// CreateEventList converts array of api events to kubermatic array events
+func CreateEventList(events []corev1.Event) []v1.Event {
```
maciaszczykm: As above.

api/pkg/handler/node_v1.go
```diff
@@ -1077,3 +1077,101 @@ func deleteNodeDeployment(projectProvider provider.ProjectProvider) endpoint.End
...
+	req.ClusterID = clusterID
+	req.NodeDeploymentID = nodeDeploymentID
+	req.DCReq = dcr.(common.DCReq)
+
+	req.Warning = r.URL.Query().Get("warning")
+
+	return req, nil
+}
+
+func listNodeDeploymentNodesEvents() endpoint.Endpoint {
```
maciaszczykm, line 1143: Can we extract the code from there and `listNodeDeploymentNodes`?

api/pkg/handler/node_v1_test.go
```diff
@@ -1167,3 +1291,20 @@ func genTestCluster(isControllerReady bool) *kubermaticv1.Cluster {
 	}
 	return cluster
 }
+
+func genTestEvent(eventName, eventType, eventReason, eventMessage string) *corev1.Event {
```
maciaszczykm, line 1295: +1

api/pkg/api/v1/types.go
```diff
@@ -665,3 +664,49 @@ type NodeDeploymentSpec struct {
...
+
+	// Namespace of the referent.
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name of the referent.
+	Name string `json:"name,omitempty"`
+}
+
+// Event is a report of an event somewhere in the cluster.
+type Event struct {
```
maciaszczykm, line 678: I was wondering if we need all these fields, but each one of them seems important.

api/pkg/handler/node_v1.go
zreigz: at least I will create some common method because 70% are equal, thanks

mrIncompetent:
I'm not a big fan of using an optional bool to express 3 states.
- Not set: Show all
- Set but false: Show only normal events
- Set but true: Show only warning events

I would prefer something like `type={warning|normal}` as its more understandable.

But just a personal opinion - feel free to ignore if you have a strong opinion on that one

api/pkg/handler/event_test.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+
+		})
+	}
+
+}
+
+// equal tells whether a and b contain the same elements.
+// A nil argument is equivalent to an empty slice.
+func equal(a, b []v1.Event) bool {
```
mrIncompetent, line 70: Can this be replaced with a deep equal?
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/compare.go#L19

zreigz: it seems like deep equal has a problem with the uint64 comparison
```
panic: an unexported field was encountered, nested like this:  [recovered]
	panic: an unexported field was encountered, nested like this: uint64 [recovered]
	panic: an unexported field was encountered, nested like this: time.Time -> uint64 [recovered]
	panic: an unexported field was encountered, nested like this: v1.Time -> time.Time -> uint64 [recovered]
	panic: an unexported field was encountered, nested like this: v1.Event -> v1.Time -> time.Time -> uint64 [recovered]
	panic: an unexported field was encountered, nested like this: []v1.Event -> v1.Event -> v1.Time -> time.Time -> uint64 [recovered]
	panic: an unexported field was encountered, nested like this: []v1.Event -> v1.Event -> v1.Time -> time.Time -> uint64
```

mrIncompetent: Thanks for trying!

zreigz: I am aware of using deep equal but in this case, it was impossible :(

api/pkg/handler/event.go
zreigz: done

api/pkg/api/v1/types.go
p0lyn0mial, line 678: I’d prefer to start as simple as possible, it’s easier to add new fields than removing. It would be nice to see how dashboard could make use of all that fields. 

The endpoint returns events for all nodes that belongs to a deployment perhaps it would be nice if we could group events by machine. Do you think we could start with something like the following ?

```
[
 {
 “name”: “kubermatic-fznjchfpqg-s549d-8478c84f95-2c2zk.157aed7ac5f5e7b4”,
 “events”: [ “Successfully created instance”, “Found instance at cloud provider” , “Requested an instance at cloud provider”]
 },  
 {“name”: “kubermatic-fznjchfpqg-s549d-8478c84f95-mkfpc.157aede5f061f6a4”,
 “events”: [ “Successfully created instance”, “Found instance at cloud provider” , “Requested an instance at cloud provider”]
 },
]
```
or alternatively we could return events as object so that adding new fields doesn’t require changing dashboard 

```
[
 {
 “name”: “kubermatic-fznjchfpqg-s549d-8478c84f95-2c2zk.157aed7ac5f5e7b4”,
 “events”: [ {“msg”: “Successfully created instance”}, {“msg”: “Found instance at cloud provider”}, {“msg”: “Requested an instance at cloud provider”}]
 },  
 {“name”: “kubermatic-fznjchfpqg-s549d-8478c84f95-mkfpc.157aede5f061f6a4”,
 “events”: [ {“msg”: “Successfully created instance”}, {“msg”: “Found instance at cloud provider”}, {“msg”: “Requested an instance at cloud provider”}]
 },
]
```

api/pkg/handler/event.go
zreigz: done

api/pkg/handler/node_v1.go
zreigz: done

api/pkg/api/v1/types.go
zreigz: make it simple sounds good. I can do like you described but additionally we can also stick with an event type

zreigz:
@p0lyn0mial  PTAL
I've simplified result for this:
```
[{"involvedObjectName":"venus-1","events":[{"reason":"Started","message":"message started","type":"Normal"},{"reason":"Killed","message":"message killed","type":"Warning"}]}]
```

p0lyn0mial:
> [{"involvedObjectName":"venus-1","events":[{"reason":"Started","message":"message started","type":"Normal"},{"reason":"Killed","message":"message killed","type":"Warning"}]}]

@zreigz thx, i'm thinking if we could change `involvedObjectName` to something else, since we know from the HTTP path that the events concern nodes thus we could rename it to `node` or `name` or `nodeName`, wdyt ?

zreigz:
@p0lyn0mial just `name` sounds good

p0lyn0mial:
also do you think we could remove `reason` field ? It's hard to find a use case for it. The `message` field gives enough info.

zreigz:
The reason is strong part of the event structure. This is real example for machine
```
Events:
  FirstSeen	LastSeen	Count	From			SubObjectPath	Type		Reason			Message
  ---------	--------	-----	----			-------------	--------	------			-------
  1h		10m		21	machine-controller			Warning		CreateError		trying to create a VM: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="SkuNotAvailable" Message="The requested size for resource '/subscriptions/528474a9-8cc6-46c8-9b74-1dbc31c4bb7f/resourceGroups/cluster-npm554cwfr/providers/Microsoft.Compute/virtualMachines/kubermatic-npm554cwfr-b572k-fd57c87f7-jr6tn' is currently not available in location 'westeurope' zones '' for subscription '528474a9-8cc6-46c8-9b74-1dbc31c4bb7f'. Please try another size or deploy to a different location or zones. See https://aka.ms/azureskunotavailable for details.". Unable to create a machine.

```

p0lyn0mial:
> The reason is strong part of the event structure. 

It's seems like it's meant for computers not humans. We started this pull to give operators/humans more information for troubleshooting.

zreigz:
OK, I will remove it :)

zreigz:
@p0lyn0mial PTAL

api/pkg/api/v1/types.go
```diff
@@ -665,3 +664,21 @@ type NodeDeploymentSpec struct {
 
 	Paused *bool `json:"paused,omitempty"`
 }
+
+// EventList is an events response structure.
+type EventList struct {
+	// The object name that those events are about.
+	Name string `json:"Name"`
```
p0lyn0mial: start with a small letter `json:"name"`

api/pkg/handler/node_v1.go
```diff
@@ -1077,3 +1087,85 @@ func deleteNodeDeployment(projectProvider provider.ProjectProvider) endpoint.End
...
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		machines, err := getMachinesForNodeDeployment(clusterProvider, cluster, req.NodeDeploymentID)
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		for _, machine := range machines.Items {
+			eventList, err := GetMachineEvents(customerClusterClient, machine)
```
p0lyn0mial: check if `len(eventList) > 0` so that we don't return an empty list`

api/pkg/handler/node_v1.go
```diff
@@ -1077,3 +1087,85 @@ func deleteNodeDeployment(projectProvider provider.ProjectProvider) endpoint.End
...
+	nodeDeploymentID, err := decodeNodeDeploymentID(c, r)
+	if err != nil {
+		return nil, err
+	}
+
+	req.ClusterID = clusterID
+	req.NodeDeploymentID = nodeDeploymentID
+	req.DCReq = dcr.(common.DCReq)
+
+	req.Type = r.URL.Query().Get("type")
```
p0lyn0mial, line 1132: check if `type` holds  correct "type" `warning/normal` and return an error if not.

api/pkg/handler/event.go
```diff
@@ -0,0 +1,66 @@
+package handler
+
+import (
+	"github.com/kubermatic/kubermatic/api/pkg/api/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
```
p0lyn0mial: rm an empty line.

api/pkg/handler/event.go
```diff
@@ -0,0 +1,66 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+
+	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+// GetMachineEvents returns kubernetes API event objects assigned to the machine.
+func GetMachineEvents(client kubernetes.Interface, machine clusterv1alpha1.Machine) (v1.EventList, error) {
```
p0lyn0mial: Could `event.go` file contain generic methods ? So that they could be reused if we decided to add events for another resource. 

If yes could we move machine specific functions `GetMachineEvents` and `createMachineEventList` to `listNodeDeploymentNodesEvents` ?

api/pkg/handler/event_test.go
```diff
@@ -0,0 +1,87 @@
...
+				Events: []v1.Event{
+					genEvent("test1", corev1.EventTypeWarning),
+					genEvent("test2", corev1.EventTypeWarning),
+					genEvent("test3", corev1.EventTypeNormal),
+					genEvent("test4", corev1.EventTypeNormal),
+				},
+			},
+		},
+		{
+			Name:   "scenario 1, filter out normal event types",
```
p0lyn0mial: rename to `scenario 2`

api/pkg/handler/node_v1_test.go
```diff
@@ -874,6 +874,130 @@ func TestListNodeDeploymentNodes(t *testing.T) {
 	}
 }
 
+func TestListNodeDeploymentNodesEvents(t *testing.T) {
```
p0lyn0mial, line 877: 👍

p0lyn0mial (changes requested):

api/pkg/handler/event_test.go
zreigz: done

api/pkg/handler/event.go
zreigz: done

api/pkg/handler/event.go
zreigz: done

api/pkg/handler/node_v1.go
zreigz: done

api/pkg/handler/node_v1.go
zreigz: done

api/pkg/api/v1/types.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

p0lyn0mial:
/lgtm

waiting for others before approving
@maciaszczykm @mrIncompetent PTAL

PR #2648, opened 2019-01-17
OpenStack: on subnet creation, verify that existing subnet CIDRs collide

### LINKED ISSUE #2250: [OpenStack] validate cidr collisions
**User Story**
As a user i can specify a custom network when creating a cluster via kubermatic.
Specifying a subnet is optional.

In this situation kubermatic will use the specified network & will try to create a new subnet.
This subnet though will always have the CIDR `192.168.1.0/24` - which might collide with an existing subnet within the network.

A validation must be added to correctly identify potential collisions.
When the validation fails an appropriate error message must be returned explaining why the cluster cannot be created.

**Acceptance criteria**
- [ ] a validation exists in the `ValidateCloudSpec` inside the OpenStack code https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/cloud/openstack/provider.go

kdomanski:
Fixes #2250

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/cloud/openstack/provider.go
+++ b/api/pkg/provider/cloud/openstack/provider.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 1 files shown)


mrIncompetent:
/lgtm
/approve

Nice and simple!

PR #2650, opened 2019-01-18
better slack templates for alerts

xrstf:
This PR changes the Slack notifications for alerts.

* Alerts are not grouped anymore, because that made creating good-looking Slack messages extremely hard.
* We have a proper fallback text for mobile notifications.
* Each single alert takes up less space in Slack.

Old:

screenshot_2019-01-18_10-38-47

New:

screenshot_2019-01-18_10-38-26

Customers not using our templates and instead use the defaults will not notice any changes, hence the empty changelog entry.

We require the current beta version from Alertmanager because they added the `match` template function, which makes it possible to match on the seed cluster name and display fancy flags without hardcoding the seed cluster names into the template. A stable 0.16.0 is supposed to be released next week (prometheus/alertmanager#1714).

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/monitoring/alertmanager/kubermatic.tmpl
+++ b/config/monitoring/alertmanager/kubermatic.tmpl
@@ -1,21 +1,39 @@
-{{ define "slack.kubermatic.title.range" }}
-{{range .Alerts }}
-{{ if eq .Status "firing" }}:fire:{{ else }}:white_check_mark:{{ end }}{{ .Labels.alertname }}{{ if eq .Status "firing" }}:fire:{{ else }}:white_check_mark:{{ end }}
-{{ end }}
-{{ end }}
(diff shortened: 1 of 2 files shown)


xmudrii:
/lgtm
Alerts now look nice :heart_eyes: 

Should we document switch to Alertmanager 0.16.0 somehow or only when we switch to a stable version?

xrstf:
I would document it with the coming PR when we update to 0.16.0 -- should happen before we cut 2.9.

mrIncompetent:
@xrstf This does not work dev. The alertmanager pod is crashing with:
```
Loading configuration file failed" file=/etc/alertmanager/config/alertmanager.yaml err="template: kubermatic.tmpl:10: function \"match\" not defined"
```

PR #2660, opened 2019-01-21
Add helm check

mrIncompetent:
Add's helm check which ensures users dont upgrade the kubermatic chart which could result in the deletion of all cluster.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/kubermatic/crd/crd-vpa.yaml
+++ b/config/kubermatic/crd/crd-vpa.yaml
@@ -1,4 +1,3 @@
-{{ if .Values.kubermatic.deployVPA }}
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
@@ -46,4 +45,3 @@ spec:
     kind: VerticalPodAutoscalerCheckpoint
     shortNames:
     - vpacheckpoint
-{{ end }}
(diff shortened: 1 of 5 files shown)


config/kubermatic/crd/crd-vpa.yaml
```diff
@@ -1,4 +1,3 @@
-{{ if .Values.kubermatic.deployVPA }}
```
mrIncompetent, line 1: Moved into the `/crd` folder inside the kubermatic chart

config/kubermatic/templates/hook-migration-crd-check.yaml
```diff
@@ -0,0 +1,48 @@
...
+      containers:
+      - name: check-crd-migration
+        image: "quay.io/kubermatic/helm:2.11.0-2"
+        command: ["/bin/bash"]
+        args:
+        - "-c"
+        - |
+          set -euo pipefail
+
+          curl -L https://storage.googleapis.com/kubernetes-helm/helm-{{ .Values.kubermatic.checks.crd.helmVersion }}-linux-amd64.tar.gz | tar -xvz
```
mrIncompetent, line 25: Always freshly download helm to ensure user can actually configure the desired helm version.

PR #2665, opened 2019-01-21
Version skew check for MachineDeployments

### LINKED ISSUE #2539: Add apiserver-kubelet verifying function for Kubernetes v1.10-v1.13
**User Story**
A version matching mechanism was already being planned and its stub exist in the `versions.yaml` structure as well as the `MasterVersion` struct in the form of  `allowedNodeVersions` field. That solution however has two downsides:
 - It depends on the apiserver version being among supported versions which are listed in `versions.yaml`. This results in inability to get valid kubelet versions for unsupported master versions, and thus inability to upgrade the kubelets first.
 - The added config complexity and YAGNI-breaking customizability.

**Acceptance criteria**
- [ ] A function is created, that accepts a pair of kubernetes versions: one of the control plane and one of the kubelet. It returns nil for supported combinations and an error for unsupported.
- [ ] In particular, the function returns false if the kubelet version is higher than the control plane version
- [ ] ~~If the given master version is not `v1.10`-`v1.13`, then the function returns an error~~
- [ ] The code itself refers to relevant Kubernetes release notes commits/lines for the rules that are verified here, e.g.: https://github.com/kubernetes/kubernetes/blob/83ba7bfd499145e514dc65fdb48a00b7e1ce6fc7/CHANGELOG-1.13.md#urgent-upgrade-notes

### LINKED ISSUE #2540: Enforce kubelet-apiserver version constaints on machine(deployment) creation or update
**User Story**
The Kubermatic API server must refuse to create Machines or MachineDeployments with kubelet versions incompatible with the current control plane version. 

**Acceptance criteria**
- [ ] ~`createNodeForCluster` will verify whether the desired kubelet version is supported by the cluster version (using #2539) and it will respond with error to the called if there versions are incompatible or there is an error during the check. No changes will be committed to the cluster in the case of incompatibility.~  **Creating machines is now deprecated**
  - [ ] ~`TestCreateNodeForCluster` verifies the check is in place~
- [ ] `createNodeDeployment` will verify whether the desired kubelet version is supported by the cluster version (using #2539) and it will respond with error to the called if there versions are incompatible or there is an error during the check. No changes will be committed to the cluster in the case of incompatibility.
  - [ ] `TestCreateNodeDeployment` verifies the check is in place
- [ ] `patchNodeDeployment` will verify whether the desired kubelet version is supported by the cluster version (using #2539) and it will respond with error to the called if there versions are incompatible or there is an error during the check. No changes will be committed to the cluster in the case of incompatibility.
  - [ ] `TestPatchNodeDeployment` verifies the check is in place

kdomanski:
Fixes #2539
Fixes #2540 

**Release note**:

```release-note bugfix
NodeDeployments can no longer be configured to provision kubelets at versions incompatible with the control plane.
```

/assign @p0lyn0mial

### DIFF
--- a/api/pkg/handler/node_v1.go
+++ b/api/pkg/handler/node_v1.go
@@ -600,7 +600,6 @@ func createNodeDeployment(sshKeyProvider provider.SSHKeyProvider, projectProvide
 			return nil, k8cerrors.NewBadRequest("cannot create node deployment without cloud provider")
 		}
 
-		//TODO: We need to make the kubelet version configurable but restrict it to versions supported by the control plane
 		if nd.Spec.Template.Versions.Kubelet != "" {
 			kversion, err := semver.NewVersion(nd.Spec.Template.Versions.Kubelet)
 			if err != nil {
--- a/api/pkg/handler/node_v1.go
+++ b/api/pkg/handler/node_v1.go
@@ -22,7 +22,6 @@ import (
 	machineconversions "github.com/kubermatic/kubermatic/api/pkg/machine"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	machineresource "github.com/kubermatic/kubermatic/api/pkg/resources/machine"
-	apierrors "github.com/kubermatic/kubermatic/api/pkg/util/errors"
 	k8cerrors "github.com/kubermatic/kubermatic/api/pkg/util/errors"
 
 	corev1 "k8s.io/api/core/v1"
@@ -203,7 +202,7 @@ func getNodeForCluster(projectProvider provider.ProjectProvider) endpoint.Endpoi
 			return nil, err
 		}
 		if machine == nil && node == nil {
-			return nil, apierrors.NewNotFound("Node", req.NodeID)
+			return nil, k8cerrors.NewNotFound("Node", req.NodeID)
 		}
 
 		if machine == nil {
--- a/api/pkg/handler/node_v1.go
+++ b/api/pkg/handler/node_v1.go
@@ -985,6 +985,15 @@ func patchNodeDeployment(sshKeyProvider provider.SSHKeyProvider, projectProvider
 			return nil, fmt.Errorf("cannot decode patched cluster: %v", err)
 		}
 
+		//TODO: We need to make the kubelet version configurable but restrict it to versions supported by the control plane
+		kversion, err := semver.NewVersion(patchedNodeDeployment.Spec.Template.Versions.Kubelet)
+		if err != nil {
+			return nil, k8cerrors.NewBadRequest("failed to parse kubelet version: %v", err)
+		}
+		if err = ensureVersionCompatible(cluster.Spec.Version.Semver(), kversion); err != nil {
+			return nil, k8cerrors.NewBadRequest(err.Error())
+		}
+
 		dc, found := dcs[cluster.Spec.Cloud.DatacenterName]
 		if !found {
 			return nil, fmt.Errorf("unknown cluster datacenter %s", cluster.Spec.Cloud.DatacenterName)
--- a/api/pkg/handler/node_v1.go
+++ b/api/pkg/handler/node_v1.go
@@ -34,8 +34,7 @@ import (
 )
 
 const (
-	kubeletVersionConstraint = ">= 1.8"
-	errGlue                  = " & "
+	errGlue = " & "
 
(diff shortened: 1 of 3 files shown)


api/pkg/handler/node_v1_test.go
```diff
@@ -935,6 +959,32 @@ func TestPatchNodeDeployment(t *testing.T) {
...
+			ExpectedResponse:           fmt.Sprintf(`{"error":{"code":400,"message":"kubelet version 9.6.0 is not compatible with control plane version 9.9.9"}}`),
+			cluster:                    "keen-snyder",
+			HTTPStatus:                 http.StatusBadRequest,
+			project:                    test.GenDefaultProject().Name,
+			ExistingAPIUser:            test.GenDefaultAPIUser(),
+			NodeDeploymentID:           "venus",
+			ExistingMachineDeployments: []*clusterv1alpha1.MachineDeployment{genTestMachineDeployment("venus", `{"cloudProvider":"digitalocean","cloudProviderSpec":{"token":"dummy-token","region":"fra1","size":"2GB"}, "operatingSystem":"ubuntu", "operatingSystemSpec":{"distUpgradeOnBoot":true}}`, nil)},
+			ExistingKubermaticObjs:     test.GenDefaultKubermaticObjects(genTestCluster(true)),
+		},
+		// Scenario 4: Downgrade to too old kubelet version
```
p0lyn0mial: nit: upgrade to too new kubelet version

api/pkg/handler/node_v1_test.go
```diff
@@ -493,6 +493,30 @@ func TestCreateNodeDeployment(t *testing.T) {
...
+			HTTPStatus:             http.StatusBadRequest,
+			ProjectID:              test.GenDefaultProject().Name,
+			ClusterID:              test.GenDefaultCluster().Name,
+			ExistingKubermaticObjs: test.GenDefaultKubermaticObjects(genTestCluster(true)),
+			ExistingAPIUser:        test.GenDefaultAPIUser(),
+		},
+
+		// scenario 4
+		{
+			Name:                   "scenario 4: kubelet version is too old",
```
p0lyn0mial: nit: change to `kubelet version is too new`

api/pkg/handler/version_skew.go
```diff
@@ -0,0 +1,47 @@
...
+		return errors.New("ensureVersionCompatible: kubelet is nil")
+	}
+
+	// Kubelet must be the same major version and no more than 2 minor versions behind the control plane.
+	// https://kubernetes.io/docs/setup/version-skew-policy/
+	// https://github.com/kubernetes/website/blob/076efdf364651859553681a75f60c957de729023/content/en/docs/setup/version-skew-policy.md
+	compatible := kubelet.Major() == controlPlane.Major() && kubelet.Minor() >= (controlPlane.Minor()-2) && kubelet.Minor() <= controlPlane.Minor()
+
+	if !compatible {
+		return ErrVersionSkew{
```
p0lyn0mial: could we just return `return fmt.Sprintf("kubelet version %s is not compatible with control plane version %s", e.Kubelet, e.ControlPlane)` instead of defining a new `ErrVersionSkew` type ?

api/pkg/handler/node_v1.go
```diff
@@ -602,20 +600,17 @@ func createNodeDeployment(sshKeyProvider provider.SSHKeyProvider, projectProvide
 			return nil, k8cerrors.NewBadRequest("cannot create node deployment without cloud provider")
 		}
 
-		//TODO: We need to make the kubelet version configurable but restrict it to master version
+		//TODO: We need to make the kubelet version configurable but restrict it to versions supported by the control plane
```
p0lyn0mial: can we rm `TODO` ? This pull actually allows to configure kubelet version.

api/pkg/handler/version_skew.go
kdomanski: Why?

api/pkg/handler/node_v1.go
kdomanski: Fixed.

api/pkg/handler/node_v1_test.go
kdomanski: Fixed.

api/pkg/handler/node_v1_test.go
kdomanski: Fixed.

PR #2666, opened 2019-01-21
update node-exporter to 0.17

xrstf:
Unfortunately NE 0.16 has changed the naming of a ton of metrics, so this PR is kind of bloated.

**Release note**:
```release-note monitoring
Updated node-exporter to `v0.17` (**note: breaking changes to metric names might require updates to customized dashboards**)
```


xrstf:
Did a smoketest deployment without any real issues. This is good to merge.

NB: After this is being deployed, we must manually delete the pods because the DaemonSet is set to deploy=onDelete.

PR #2671, opened 2019-01-23
Add initial skeleton of user cluster controller using kubebuilder

nikhita:
This PR adds the user cluster controller using kubebuilder.

Currently, this controller only reconciles `APIService`s since creators for it don't require the `Cluster` CRD. Other resources will be achieved as a follow-up.

I guess we should add the release note on the final PR?

```release-note
NONE
```


api/pkg/controller/usercluster/controller.go
```diff
@@ -0,0 +1,95 @@
...
+	ctx context.Context
+	client.Client
+}
+
+// Reconcile makes changes in response to objects in the user cluster.
+func (r *reconcileUserCluster) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	// TODO: figure out how to pass the complete cluster object
+	// and the required data to the Reconcile method
+	// rUserCluster := reconciler{client: r.Client, ctx: r.ctx}
+	// err := rUserCluster.Reconcile(cluster, data)
```
nikhita: The problem I was facing here was that...there are some creators who want the complete Cluster object so passing through flags wasn't turning out to be enough. Suggestions on how to mitigate this? :thinking:

api/pkg/controller/usercluster/controller.go
```diff
@@ -0,0 +1,95 @@
...
+	// if err != nil {
+	// 		return reconcile.Result{
+	//  there is no need to set Requeue to true at the same time as RequeueAfter
+	// TODO: figure out how to requeue in exponential manner
```
alvaroaleman: Maybe ask in their slack channel?

api/pkg/controller/usercluster/controller.go
alvaroaleman: Lets start off by giving the user cluster controller a lister for the seed for the kubermatic cluster CRD. We will figure out how to get rid of that in a follow-up

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	for _, create := range creators {
+		role, err := create(data, nil)
+		if err != nil {
+			return fmt.Errorf("failed to build Role: %v", err)
+		}
+
+		existing := &rbacv1.Role{}
+		err = r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: role.Namespace, Name: role.Name}, existing)
```
alvaroaleman: Do these use the lister underneath?

api/pkg/controller/usercluster/controller.go
alvaroaleman: Actually, better idea: Lets for now leave all resources that need this inside the cluster-controller and move them over in a follow-up, i.E. not handle them here for the time being

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -0,0 +1,559 @@
...
+		machinecontroller.NodeSignerClusterRoleBinding,
+		vpnsidecar.DnatControllerClusterRoleBinding,
+		metricsserver.ClusterRoleBindingResourceReader,
+		metricsserver.ClusterRoleBindingAuthDelegator,
+		scheduler.ClusterRoleBindingAuthDelegator,
+		controllermanager.ClusterRoleBindingAuthDelegator,
+	}
+
+	if len(c.Spec.MachineNetworks) > 0 {
+		creators = append(creators, ipamcontroller.ClusterRoleBinding)
```
alvaroaleman: We can btw just always create this, it wont harm

alvaroaleman (commented):
I am actually mostly fine with this. What is missing right now is mainly:

* Remove the existing resource handling from the cluster controller for all resources that are now in the UCC
* Deploy the UCC
* The script that is used to run the UCC locally

We can work on getting that done together. If that is done and E2E tests pass, we can merge this and do the missing resources in follow-ups

api/pkg/controller/usercluster/reconciler.go
nikhita: Yes.

nikhita:
@alvaroaleman Updated, PTAL.

alvaroaleman:
What is still missing is to make the cluster controller deploy this controller and create a kubeconfig.
Looks good! What is still missing though is to make the cluster controller deploy the user cluster controller. This requires:

* A kubeconfig secret like this: https://github.com/kubermatic/kubermatic/blob/9c54bf86c0383152e837ea90397c4f263d48e13d/api/pkg/controller/cluster/resources.go#L212 The certificates "organization" field must be set to `system:masters` so the controller gets put into that group for which a cluster-admin clusterrolebinding exists by default
* A deployment that deploys it in the seed like this: https://github.com/kubermatic/kubermatic/blob/9c54bf86c0383152e837ea90397c4f263d48e13d/api/pkg/controller/cluster/resources.go#L167

We can do that together if that helps

nikhita:
@alvaroaleman Updated, PTAL. Happy to discuss this more over a call if needed too! :)

api/pkg/resources/usercluster/deployment.go
```diff
@@ -0,0 +1,115 @@
...
+
+var (
+	defaultResourceRequirements = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
+			corev1.ResourceCPU:    resource.MustParse("25m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+			corev1.ResourceCPU:    resource.MustParse("2"),
```
alvaroaleman: That is definitely too much, can we make that half a core/500 milli cores?

api/pkg/resources/usercluster/deployment.go
```diff
@@ -0,0 +1,115 @@
...
+		apiserverIsRunningContainer, err := apiserver.IsRunningInitContainer(data)
+		if err != nil {
+			return nil, err
+		}
+		dep.Spec.Template.Spec.InitContainers = []corev1.Container{*apiserverIsRunningContainer}
+
+		dep.Spec.Template.Spec.Containers = []corev1.Container{
+			{
+				Name:            name,
+				Image:           data.ImageRegistry(resources.RegistryGCR) + "/kubermatic/usercluster-controller", // TODO: fix image
```
alvaroaleman: You can do it like the IPAM controller and use the commit that was built as tag: https://github.com/kubermatic/kubermatic/blob/f07996f2df698fa07b7eb8f4c94ae963c6ddb2f3/api/pkg/resources/ipamcontroller/deployment.go#L89

alvaroaleman: And you can also just use the `kubermatic/api` image, it contains all binaries that are built via `make build`

alvaroaleman (commented):
Modulo the image LGTM :+1: 

I am a bit surprised this passed the tests, because of the wrong image the `user-cluster-controller` pod should have went into ImagePullBackoff and that should have made the tests fail :thinking:

nikhita:
@alvaroaleman Updated, PTAL.

> I am a bit surprised this passed the tests, because of the wrong image the `user-cluster-controller` pod should have went into ImagePullBackoff and that should have made the tests fail

The test just checks that the deployment fixtures match the deployment created by the `DeploymentCreator`, they don't "actually" create the deployment.

alvaroaleman:
> The test just checks that the deployment fixtures match the deployment created by the DeploymentCreator, they don't "actually" create the deployment.

The e2e tests do

api/pkg/controller/usercluster/controller.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// add adds a new Controller to mgr with r as the reconcile.reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	if err = c.Watch(&source.Kind{Type: &apiregistrationv1beta1.APIService{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
alvaroaleman, line 35: Right now this is for a follow-up tasks but: We have to make sure that we always put the same key into the workq independently of what informer triggered it to make sure we get

* The deduplication (i.e. if five informers trigger we still have to run Reconcile only once)
* The lock for the item, i.e. we should never have more than one worker working at a time

This would be easy to achieve if we could directly access the workqeue by just always putting a `const` on it, not sure how its done with the kubebuilder abstractions

alvaroaleman (commented):
LGTM except that for some reason the pulling didn't work in the e2e tests so putting a hold on it until we know what the root cause for that is.

/lgtm
/approve
/hold

nikhita:
@alvaroaleman PTAL at the last commit. Is this what you meant? I'm wondering if its ok to push to quay if we can't use `docker manifest inspect` there...

alvaroaleman:
Yeah its okay to not do that but please also remove the pushing to dockerhub, it costs time and is not needed. This then implies removing the whole check and unconditionally pushing. Also I think we don't need the docker experimental features enabled anymore then.

alvaroaleman:
@nikhita The script configures dockerhub as location for the images in the helm command :S

nikhita:
@alvaroaleman Ok to cancel the hold?

nikhita:
Discussed on the daily call, removing the hold.

/hold cancel
/pony yay

PR #2674, opened 2019-01-23
update cert-manager to 0.6.0

xrstf:
Thew new cert-manager update is supposed to be the shizzle and they "**strongly recommend** all users upgrade to the v0.6 release as soon as possible!"

This PR updates the cert-manager chart and adds the new ca-sync CronJob as well as the new webhook.

I am not sure whether we need to do something about https://github.com/jetstack/cert-manager/pull/1136.

**Release note**:
```release-note
[ACTION REQUIRED] Updated cert-manager to `v0.6.0` (see https://cert-manager.readthedocs.io/en/latest/admin/upgrading/index.html)
```


alvaroaleman:
The hold is just to make sure this PR doesnt merge before you know stuff actually works :) Feel free to remove it  if you know it does work

xrstf:
I removed a domain from on of our certs and the cert-manager immediately got a new cert. Re-adding the domain again yielded a new certificate. Seems like everything is working as intended.

PR #2676, opened 2019-01-23
marks the remaining node HTTP endpoints as deprecated

p0lyn0mial:
**What this PR does / why we need it**: marks the remaining node HTTP endpoints as deprecated

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/handler/node_v1.go
+++ b/api/pkg/handler/node_v1.go
@@ -44,7 +44,7 @@ const (
 	initialConditionParsingDelay = 5
 )
 
-func deleteNodeForCluster(projectProvider provider.ProjectProvider) endpoint.Endpoint {
+func deleteNodeForClusterLegacy(projectProvider provider.ProjectProvider) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(DeleteNodeForClusterReq)
 		clusterProvider := ctx.Value(middleware.ClusterProviderContextKey).(provider.ClusterProvider)
@@ -91,7 +91,7 @@ func deleteNodeForCluster(projectProvider provider.ProjectProvider) endpoint.End
 	}
 }
 
-func listNodesForCluster(projectProvider provider.ProjectProvider) endpoint.Endpoint {
+func listNodesForClusterLegacy(projectProvider provider.ProjectProvider) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req := request.(ListNodesForClusterReq)
 		clusterProvider := ctx.Value(middleware.ClusterProviderContextKey).(provider.ClusterProvider)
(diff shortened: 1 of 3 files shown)


api/pkg/handler/node_v1.go
```diff
@@ -44,7 +44,7 @@ const (
 	initialConditionParsingDelay = 5
 )
 
-func deleteNodeForCluster(projectProvider provider.ProjectProvider) endpoint.Endpoint {
+func deleteNodeForClusterLegacy(projectProvider provider.ProjectProvider) endpoint.Endpoint {
```
maciaszczykm, line 47: I am not sure if we are going to keep that functionality or not. Deleting single nodes can work as a way of "restarting" them.

maciaszczykm (commented):
/lgtm
/approve
/hold

Adding hold for deleting single nodes as it is not obvious for me at the moment. cc @alvaroaleman WDYT?

api/pkg/handler/node_v1.go
p0lyn0mial: This pull doesn't remove any endpoint it only marks them as deprecated. We don't have a deprecation policy in place so I would assume that in the future releases we could remove it.

alvaroaleman: Is this endpoint used when deleting a single node from the nodeDeployment view? If yes I'd vote for keeping it to force the rotation of a node

alvaroaleman:
Is the dashboard adjusted for this? We must continue to have a UI that can manage single nodes until all customers of ours use Kubermatic >= 2.9

p0lyn0mial:
> Is the dashboard adjusted for this? We must continue to have a UI that can manage single nodes until all customers of ours use Kubermatic >= 2.9

From the dashboard point of view nothing has changed because I haven't deleted any endpoints, this pull only adds deprecation note to the swagger spec.

api/pkg/handler/node_v1.go
maciaszczykm: > Is this endpoint used when deleting a single node from the nodeDeployment view?

I think it is.

p0lyn0mial: before removing this endpoint we will create a new one under `nodedeployments` path.

maciaszczykm:
> From the dashboard point of view nothing has changed because I haven't deleted any endpoints, this pull only adds deprecation note to the swagger spec.

Are we going to use deprecated method for long time from now? I mean that list and get will be not used soon, but delete might be used for a long time from now.

maciaszczykm:
/hold cancel

Everything seems to be clear right now. We have discussed it on Slack. If we will delete `delete` endpoint it will be replcated by new one with `nodedeployments` in the path. To be more consistent and have everything under one path.

p0lyn0mial:
> Installing Kubermatic via Helm
Error: forwarding ports: error upgrading connection: error dialing backend: EOF
Starting cleanup

/test pull-kubermatic-e2e-aws-coreos-1.11

p0lyn0mial:
@alvaroaleman the tests are not stable it would be nice if we could make them optional or improve their stability :)

alvaroaleman:
The tests will never become stable if we make them optional :)

kdomanski:
@p0lyn0mial I'm pretty sure the customers don't need to know that we've changed names of a few internal functions. I removed the release note.

p0lyn0mial:
@kdomanski this pull marked some public (not internal) endpoints as deprecated. It leaks to OpenAPI spec. Everyone who uses the spec to generate documentation or clients will see it.

alvaroaleman:
@p0lyn0mial We don't have public endpoints, because we don't have a public API

p0lyn0mial:
@alvaroaleman good to know, I have been doing it wrong for the whole time ;)
@alvaroaleman this is actually very convenient because it means every time we add/remove an endpoint we don't have to announce to the external world (release note). It also means that we have zero deprecation policy and we can actually remove endpoints.

alvaroaleman:
Yup. The api will become a public one when we have service accounts. Until then it's for use by the dashboard only.

PR #2678, opened 2019-01-23
RBAC manager for user cluster

### LINKED ISSUE #2696: create user cluster RBAC manager
**User Story**
 Resources that are in the cluster are protected and have proper permissions assigned.

**Acceptance criteria**
RBAC controller on cluster creation adds Roles and RoleBindings for specific groups: `owners`, `editors` and `viewers`. Roles are assigned for `machinedeployments` and `machines` resources

zreigz:
**What this PR does / why we need it**: new RBAC manager for user cluster

Fixes #2696

```release-note
new RBAC manager for user cluster
```


api/cmd/rbac-user-cluster/main.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+
+	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
```
p0lyn0mial: could we move registration to its own file `register.go` to be consistent with k8s, for example 

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/code-generator/_examples/MixedCase/apis/example/v1/register.go

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/apps/v1/register.go

api/cmd/rbac-user-cluster/main.go
```diff
@@ -0,0 +1,64 @@
...
+		glog.Fatal(err)
+	}
+
+	mgr, err := manager.New(cfg, manager.Options{})
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	glog.Info("registering runtime objects")
+	schemeBuilder.Register(runtimeObjects...)
```
p0lyn0mial: could be registered in `init` function https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/code-generator/_examples/MixedCase/apis/example/v1/register.go#L35

api/cmd/rbac-user-cluster/controllers.go
```diff
@@ -0,0 +1,18 @@
+package main
```
p0lyn0mial: please add `OWNERS` and `README` files.

api/cmd/rbac-user-cluster/controllers.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac-user-cluster/machinedeployment"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// AddToManager adds all Controllers to the Manager
+func AddToManager(m manager.Manager) error {
```
p0lyn0mial: make it private.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		// If we can't get a clientset we can't do anything else
+		glog.Fatal(err)
```
p0lyn0mial: please return the error instead.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+	err = c.Watch(&source.Kind{Type: &apiextensionv1beta1.CustomResourceDefinition{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ReconcileMachineDeployment reconciles a MachineDeployment object
+type ReconcileMachineDeployment struct {
```
p0lyn0mial: can we just rename it to `reconciler` ? Initially it's going to watch over `machinedeployments` but later on we could add additional CRDs for example `machines` without creating a new controller. I think that filtering the resources we want to reconcile is enough, wdyt ?

p0lyn0mial: perhaps `reconciler` doesn't have to be in its own package `machinedeployment` having it in `rbac_user_cluster` is enough.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+const machinedeploymentCRDName = "machinedeployments.cluster.k8s.io"
+
+// Add creates a new Namespace Controller and adds it to the Manager.
+// The Manager will set fields on the Controller and Start it.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
```
p0lyn0mial: I think that it would be prudent to expose at least one metric reporting health status (can be very primitive) to our monitoring system so that we know that the controller is up and running.

https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/rbac/rbac_controller.go#L37


can be implement as a follow-up.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+		glog.Fatal(err)
+	}
+
+	return &ReconcileMachineDeployment{Client: mgr.GetClient(), clientset: clientset}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("machinedeployment-rbac-controller", mgr, controller.Options{Reconciler: r})
```
p0lyn0mial: if we decide to rename the reconciler then we need to change the name of the controller too.

api/pkg/controller/rbac-user-cluster/machinedeployment/reconciler.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) ensureRBACRole(groupName, name string, crd *apiextensionv1beta1.CustomResourceDefinition) error {
+
+	if crd.DeletionTimestamp != nil {
+		glog.Infof("deleting Role %s", name)
```
p0lyn0mial: I would suggest removing logging it will be too noisy, although they can be useful for debugging, in that case we could add log level.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+	crd := &apiextensionv1beta1.CustomResourceDefinition{}
+	err = r.Get(context.TODO(), request.NamespacedName, crd)
+	if err != nil {
+		return reconcile.Result{}, nil
+	}
+
+	if crd.Name == machinedeploymentCRDName {
+		err = rdr.Reconcile(crd)
+		if err != nil {
+			return reconcile.Result{}, err
```
p0lyn0mial: we definitely want to re-enqueue on certain errors.

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Reconcile makes changes in response to MachineDeployment changes
+func (r *ReconcileMachineDeployment) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	var err error
+	rdr := Reconciler{Clientset: r.clientset}
+	crd := &apiextensionv1beta1.CustomResourceDefinition{}
+	err = r.Get(context.TODO(), request.NamespacedName, crd)
+	if err != nil {
+		return reconcile.Result{}, nil
```
p0lyn0mial: we should check the error and try to add the object to the queue. For example we don't have re-enqueue if the resource doesn't exist but we have to in case of timeouts.

api/pkg/controller/rbac-user-cluster/machinedeployment/reconciler.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to generate the RBAC Role: %v", err)
+			}
+			glog.Infof("creating a new Role %s", name)
+			if _, err := r.Clientset.RbacV1().Roles(metav1.NamespaceSystem).Create(newRole); err != nil {
+				return fmt.Errorf("failed to create the RBAC Role: %v", err)
+			}
+			return nil
+		}
+		return err
```
p0lyn0mial: in case the role already exits we should make sure that it matches the generated one and hasn't been changed manually.

api/pkg/controller/rbac-user-cluster/machinedeployment/reconciler.go
```diff
@@ -0,0 +1,101 @@
...
+				return nil
+			}
+			return fmt.Errorf("failed to delete the RBAC Role: %v", err)
+		}
+		return nil
+	}
+
+	if _, err := r.Clientset.RbacV1().Roles(metav1.NamespaceSystem).Get(name, metav1.GetOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			newRole, err := common.GenerateRBACRole(groupName, policyAPIGroup, []string{"machinedeployments", "machines"})
```
p0lyn0mial: we watch only for one resource namely `machinedeployments` but here we generate a role for two resources `machinedeployment` and `machines`. Additionally the names are hardcoded but could be taken from  from `crd` argument variable.

api/pkg/controller/rbac-user-cluster/common/mapper.go
```diff
@@ -0,0 +1,78 @@
+package common
```
p0lyn0mial: could we move it to `rbac_user_cluster` pkg ?

p0lyn0mial (changes requested):

api/pkg/controller/rbac-user-cluster/common/mapper.go
```diff
@@ -0,0 +1,78 @@
...
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// generateVerbsForGroup generates a set of verbs for a group
+func generateVerbsForGroup(groupName string) ([]string, error) {
+	// verbs for owners
+	if groupName == rbac.OwnerGroupNamePrefix {
+		return []string{"list", "get", "update", "delete"}, nil
```
p0lyn0mial: I think that `create` is missing - i'll double check the verbs.

api/pkg/controller/rbac-user-cluster/common/mapper.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// verbs for editors
+	if groupName == rbac.EditorGroupNamePrefix {
+		return []string{"list", "get", "update"}, nil
```
p0lyn0mial: editors can `delete` and `create` too.

api/cmd/rbac-user-cluster/main.go
zreigz: done

api/cmd/rbac-user-cluster/controllers.go
zreigz: done

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
zreigz: done

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
zreigz: done

api/pkg/controller/rbac-user-cluster/machinedeployment/controller.go
zreigz: done

api/pkg/controller/rbac-user-cluster/common/mapper.go
zreigz: done

api/pkg/controller/rbac-user-cluster/common/mapper.go
zreigz: done

api/pkg/controller/rbac-user-cluster/common/mapper.go
zreigz: done

api/cmd/rbac-user-cluster/controllers.go
zreigz: done

api/pkg/controller/rbac-user-cluster/machinedeployment/reconciler.go
zreigz: done

api/pkg/controller/rbac-user-cluster/machinedeployment/reconciler.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/cmd/rbac-user-cluster/controllers.go
```diff
@@ -0,0 +1,47 @@
...
+
+// addToManager adds all Controllers to the Manager
+func addToManager(m manager.Manager) error {
+	metrics := newMetrics()
+	addToManagerFuncs := []func(manager.Manager) error{crd.Add}
+	for _, f := range addToManagerFuncs {
+		if err := f(m); err != nil {
+			return err
+		}
+		glog.V(0).Info("New controller added")
```
p0lyn0mial: would be nice if we could print the controller name.

api/cmd/rbac-user-cluster/controllers.go
```diff
@@ -0,0 +1,47 @@
...
+			Help:      "The number of running controllers",
+		}),
+	}
+
+	cm.controllers.Set(0)
+	return cm
+}
+
+// addToManager adds all Controllers to the Manager
+func addToManager(m manager.Manager) error {
```
p0lyn0mial: could we rename it to `registerControllers` ?

api/cmd/rbac-user-cluster/controllers.go
```diff
@@ -0,0 +1,47 @@
+package main
+
+import (
```
p0lyn0mial: make sure that imports are correctly grouped, usually it is
```
standard library

external libraries

kubermatic

kubernetes
```

for example
```
	"fmt"
	"strings"

	"github.com/golang/glog"

	rbacv1 "k8s.io/api/rbac/v1"
	"k8s.io/apimachinery/pkg/api/equality
```

api/cmd/rbac-user-cluster/main.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	mgr, err := manager.New(cfg, manager.Options{})
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	glog.V(0).Info("registering components")
```
p0lyn0mial: i tried to run the app locally but this line wasn't printed
```
./_build/rbac-user-cluster -logtostderr=1 -v=6
I0207 08:19:21.301974   10603 loader.go:359] Config loaded from file /home/lukasz/.kube/config
```

api/pkg/controller/rbac-user-cluster/apis/apiextensions/v1beta1/register.go
```diff
@@ -0,0 +1,35 @@
+package v1beta1
```
p0lyn0mial: move the pkg to ` api/pkg/controller/rbac-user-cluster/`

api/pkg/controller/rbac-user-cluster/crd/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	var err error
+	rdr := Reconciler{Clientset: r.clientset}
+	crd := &apiextensionv1beta1.CustomResourceDefinition{}
+	err = r.Get(context.TODO(), request.NamespacedName, crd)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{
+			//  there is no need to set Requeue to true at the same time as RequeueAfter
```
p0lyn0mial: I'd add a comment that we would like to investigate how to requeue items in an exponential manner.

api/pkg/controller/rbac-user-cluster/crd/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	machinesCRDName          = "machines.cluster.k8s.io"
+	controllerName           = "crd-rbac-controller"
+
+	// time in sec
+	requeueAfter1sec = time.Second
+)
+
+// Add creates a new Namespace Controller and adds it to the Manager.
+// The Manager will set fields on the Controller and Start it.
+func Add(mgr manager.Manager) error {
```
p0lyn0mial: this could return a name of the controller.

api/pkg/controller/rbac-user-cluster/apis/apiextensions/v1beta1/register.go
```diff
@@ -0,0 +1,35 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name use in this package
+const GroupName = "apiextensions.k8s.io"
+
+// SchemeGroupVersion is group version used to register these objects
+var SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: "v1beta1"}
```
p0lyn0mial: replace with `apiextensionv1beta1.SchemeGroupVersion`

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+
+	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	policyAPIGroup = "cluster.k8s.io"
+	crdAPIVersion  = "apiextensions.k8s.io/v1beta1"
```
p0lyn0mial: use `apiextensionv1beta1.SchemeGroupVersion`

api/cmd/rbac-user-cluster/main.go
```diff
@@ -0,0 +1,43 @@
...
+	if err := v1beta1.AddToScheme(mgr.GetScheme()); err != nil {
```
p0lyn0mial: could this be replaced by `apiextensionv1beta1.AddToScheme(mgr.GetScheme())` ?

p0lyn0mial: if yes are there any drawbacks ?

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac-user-cluster"
+
+	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	policyAPIGroup = "cluster.k8s.io"
```
p0lyn0mial: if possible take the name from a public API just like we can with `apiextensionv1beta1.SchemeGroupVersion`, this also applies to line 20 `crdKind`

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to generate the RBAC Role: %v", err)
+	}
+
+	role, err := r.Clientset.RbacV1().Roles(metav1.NamespaceSystem).Get(rbacName, metav1.GetOptions{})
+	if err != nil {
+		// create role if not exist
+		if errors.IsNotFound(err) {
+			glog.V(3).Infof("creating a new Role %s", rbacName)
```
p0lyn0mial: this is debug information, I'd increase the log level to `4`, see https://github.com/eBay/Kubernetes/blob/master/docs/devel/logging.md, you might want to check other places too.

api/pkg/controller/rbac-user-cluster/mapper.go
```diff
@@ -0,0 +1,81 @@
...
+// generateVerbsForGroup generates a set of verbs for a group
+func generateVerbsForGroup(groupName string) ([]string, error) {
+	// verbs for owners
+	if groupName == rbac.OwnerGroupNamePrefix {
+		return []string{"create", "list", "get", "update", "delete"}, nil
+	}
+
+	// verbs for editors
+	if groupName == rbac.EditorGroupNamePrefix {
+		return []string{"delete", "create", "list", "get", "update"}, nil
```
p0lyn0mial: verbs for editors are exactly the same as for owners we could remove this line.

api/cmd/rbac-user-cluster/main.go
zreigz: It will display but propably later:
```
I0207 09:57:06.305947   16982 loader.go:359] Config loaded from file /home/lukasz/Downloads/kubeconfig-admin-wmvg2rkd6c
I0207 09:57:06.470396   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/api?timeout=32s 200 OK in 163 milliseconds
I0207 09:57:06.495738   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:06.538664   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/api/v1?timeout=32s 200 OK in 28 milliseconds
I0207 09:57:06.585364   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apiregistration.k8s.io/v1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:06.614012   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apiregistration.k8s.io/v1beta1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:06.646274   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/extensions/v1beta1?timeout=32s 200 OK in 30 milliseconds
I0207 09:57:06.689344   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apps/v1?timeout=32s 200 OK in 39 milliseconds
I0207 09:57:06.717514   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apps/v1beta2?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:06.749343   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apps/v1beta1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:06.782290   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/events.k8s.io/v1beta1?timeout=32s 200 OK in 28 milliseconds
I0207 09:57:06.968395   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/authentication.k8s.io/v1?timeout=32s 200 OK in 185 milliseconds
I0207 09:57:07.000030   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/authentication.k8s.io/v1beta1?timeout=32s 200 OK in 31 milliseconds
I0207 09:57:07.034938   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/authorization.k8s.io/v1?timeout=32s 200 OK in 33 milliseconds
I0207 09:57:07.062792   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/authorization.k8s.io/v1beta1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:07.092464   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/autoscaling/v1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:07.119100   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/autoscaling/v2beta1?timeout=32s 200 OK in 26 milliseconds
I0207 09:57:07.152231   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/autoscaling/v2beta2?timeout=32s 200 OK in 30 milliseconds
I0207 09:57:07.177723   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/batch/v1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:07.205064   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/batch/v1beta1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:07.235374   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/certificates.k8s.io/v1beta1?timeout=32s 200 OK in 28 milliseconds
I0207 09:57:07.263362   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/networking.k8s.io/v1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:07.290954   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/policy/v1beta1?timeout=32s 200 OK in 26 milliseconds
I0207 09:57:07.322305   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/rbac.authorization.k8s.io/v1?timeout=32s 200 OK in 28 milliseconds
I0207 09:57:07.351958   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/rbac.authorization.k8s.io/v1beta1?timeout=32s 200 OK in 28 milliseconds
I0207 09:57:07.378791   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/storage.k8s.io/v1?timeout=32s 200 OK in 26 milliseconds
I0207 09:57:07.406238   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/storage.k8s.io/v1beta1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:07.434903   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/admissionregistration.k8s.io/v1beta1?timeout=32s 200 OK in 26 milliseconds
I0207 09:57:07.464389   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/admissionregistration.k8s.io/v1alpha1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:07.494503   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/apiextensions.k8s.io/v1beta1?timeout=32s 200 OK in 29 milliseconds
I0207 09:57:07.525003   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/scheduling.k8s.io/v1beta1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:07.550856   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/coordination.k8s.io/v1beta1?timeout=32s 200 OK in 25 milliseconds
I0207 09:57:07.583560   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/crd.projectcalico.org/v1?timeout=32s 200 OK in 31 milliseconds
I0207 09:57:07.611589   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/cluster.k8s.io/v1alpha1?timeout=32s 200 OK in 27 milliseconds
I0207 09:57:07.646106   16982 round_trippers.go:405] GET https://wmvg2rkd6c.europe-west3-c.dev.kubermatic.io:30413/apis/metrics.k8s.io/v1beta1?timeout=32s 200 OK in 29 milliseconds
I0207 09:57:07.647989   16982 main.go:27] registering components
I0207 09:57:07.648275   16982 main.go:33] setting up controller
I0207 09:57:07.649064   16982 controllers.go:45] New controller crd-rbac-controller added
```

api/cmd/rbac-user-cluster/controllers.go
zreigz: done

api/cmd/rbac-user-cluster/controllers.go
zreigz: done

api/cmd/rbac-user-cluster/controllers.go
zreigz: done

api/pkg/controller/rbac-user-cluster/apis/apiextensions/v1beta1/register.go
zreigz: now it is deleted

api/pkg/controller/rbac-user-cluster/crd/controller.go
zreigz: done

api/pkg/controller/rbac-user-cluster/apis/apiextensions/v1beta1/register.go
zreigz: done

api/cmd/rbac-user-cluster/main.go
zreigz: works fine with `apiextensionv1beta1.AddToScheme(mgr.GetScheme())`

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
zreigz: for kind is a problem. Even tests for this package use string with CRD name. There is no const with definition for this

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
zreigz: done

api/pkg/controller/rbac-user-cluster/mapper.go
zreigz: done

api/pkg/controller/rbac-user-cluster/crd/reconciler.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/cmd/user-cluster-controller/main.go
```diff
@@ -0,0 +1,104 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+)
+
+type controllerRunOptions struct {
+	internalAddr string
+}
+
+func main() {
```
p0lyn0mial: @nikhita this will be our share of the pie, PTAL :)

api/cmd/user-cluster-controller/controllers.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	cm.controllers.Set(0)
+	return cm
+}
+
+// registerControllers adds all Controllers to the Manager
+func registerControllers(m manager.Manager) error {
+	metrics := newMetrics()
+	addToManagerFuncs := []func(manager.Manager) (string, error){crd.Add}
```
p0lyn0mial: here you can register new controllers.

api/cmd/user-cluster-controller/README.md
```diff
@@ -0,0 +1,6 @@
+# rbac-user-cluster
```
nikhita: I see the stuff in the cmd as the controller-manager instead of just one controller...can we move this part to where the controller resides? :grimacing:

zreigz: initially, it was intended for RBAC only, the plan has changed so I will change it

api/pkg/controller/user-cluster-controller/crd/controller.go
```diff
@@ -0,0 +1,110 @@
+package crd
```
p0lyn0mial: @zreigz I think we should change the directory name to `rbac-user-cluster`.

zreigz: yes, I like this idea

api/pkg/controller/rbac-user-cluster/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	policyAPIGroup = "cluster.k8s.io"
+	crdAPIVersion  = apiextensionv1beta1.GroupName + "/v1beta1"
+	crdKind        = "CustomResourceDefinition"
+
+	debugLevel = 4
+)
+
+// reconciler creates and deletes Kubernetes resources to achieve the desired state of an RBAC Definition
+type reconciler struct {
+	// TODO replace clientset for sigs.k8s.io/controller-runtime/pkg/client.Client
```
zreigz: To use `controller-runtime` client we need newer version at least `v0.1.8` where fakeClient list method was fixed:
https://github.com/kubernetes-sigs/controller-runtime/pull/213
It work with k8s v1.12 and we have v1.11

zreigz: also, it requires newer go-client

api/cmd/user-cluster-controller-manager/OWNERS
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+  - team-ui
```
alvaroaleman, line 4: Please add the lifecycle team here, as the files here will be shared once #2671 merges

nikhita: @zreigz @p0lyn0mial fwiw, I've rebased my PR on top of this, so there shouldn't be too many merge conflicts.

zreigz: team-lifecycle added

api/pkg/controller/rbac-user-cluster/OWNERS
```diff
@@ -0,0 +1,12 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+  - team-ui
+  - team-lifecycle
```
alvaroaleman: I actually meant to add `team-lifecycle` to _only_ the `cmd/user-cluster-controller` package as that one is shared, the rbac controller itself is something team-lifecycle should not have approve permissions on IMHO

api/pkg/controller/rbac-user-cluster/reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+			glog.V(debugLevel).Infof("creating a new Role %s", defaultRole.Name)
+			if err := r.client.Create(r.ctx, defaultRole); err != nil {
+				return fmt.Errorf("failed to create the RBAC Role: %v", err)
+			}
+			return nil
+		}
+		return err
+	}
+	// compare rules with default. If don't match update for default
+	if !RolesMatches(role, defaultRole) {
```
nikhita: Out of curiousity, why don't we directly do a `equality.Semantic.DeepEqual(role, defaultRole)` here?

api/pkg/controller/rbac-user-cluster/controller.go
```diff
@@ -0,0 +1,93 @@
...
+	err = r.Get(r.ctx, request.NamespacedName, crd)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{
+			//  there is no need to set Requeue to true at the same time as RequeueAfter
+			// TODO: figure out how to requeue in exponential manner
+			RequeueAfter: requeueAfter1sec,
+		}, nil
```
alvaroaleman: I _believe_ the only thing that has to be done here is to actually return the error instead of swallowing it, the controller will then call `Reconcile` again: https://github.com/kubernetes-sigs/controller-runtime/blob/bd4677326d3a1e201f986bf798638fd38fbe3680/pkg/reconcile/reconcile.go#L82

CC @nikhita because I believe this has to be fixed in the user cluster controller as well

zreigz: good point, thanks

api/pkg/controller/rbac-user-cluster/reconciler.go
zreigz: the `role` object is a full object from API but `defaultRole` is a just template with the most important fields so they don't fully match

api/pkg/controller/rbac-user-cluster/controller.go
alvaroaleman: Yeah it just uses the `AddRateLimited` from the workqueue if there was an error: https://github.com/kubernetes-sigs/controller-runtime/blob/0dd6edb02bf23e3669a3edc50ef00a38719173b0/pkg/internal/controller/controller.go#L216

nikhita: Re user cluster controller, it returns an error:

https://github.com/kubermatic/kubermatic/blob/3da7568e8aa898eb836f5182611e8dbc115b56d3/api/pkg/controller/usercluster/controller.go#L54-L58

alvaroaleman: Maybe just remove the `RequeueAfter` then? It is a no-op right now but would make the code easier to understand if it wasn't there

zreigz: we decided to use `RequeueAfter: requeueAfter1sec` in some unexpected cases like a network problem to delay another check

alvaroaleman: I would argue every error that happens here is unexpected and the normal backoff mechanism of the q should be used instead of trying to be smarter

zreigz: ok, so I will remove it

api/pkg/controller/rbac-user-cluster/role/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Errorf("failed to generate the RBAC Cluster Role: %v", err)
+	}
+
+	clusterRole := &rbacv1.ClusterRole{}
+	err = r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: defaultClusterRole.Name}, clusterRole)
+	if err != nil {
+		// create clusterRole if not exist
+		if errors.IsNotFound(err) {
+			glog.V(debugLevel).Infof("creating a new Cluster Role %s", defaultClusterRole.Name)
+			if err := r.client.Create(r.ctx, defaultClusterRole); err != nil {
```
p0lyn0mial: in what namespace is the role created ?

api/pkg/controller/rbac-user-cluster/role/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+		err := r.ensureRBACRole(groupName)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *reconciler) ensureRBACRole(groupName string) error {
```
p0lyn0mial: nit: consider changing the name to `ensureRBACClusterRole`.

api/pkg/controller/rbac-user-cluster/role/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+	defaultClusterRole, err := rbacusercluster.GenerateRBACClusterRole(groupName)
+	if err != nil {
+		return fmt.Errorf("failed to generate the RBAC Cluster Role: %v", err)
+	}
+
+	clusterRole := &rbacv1.ClusterRole{}
+	err = r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: defaultClusterRole.Name}, clusterRole)
```
p0lyn0mial: are we searching for the role in all namespaces ? shouldn't we search `kube-system` only ?

api/pkg/controller/rbac-user-cluster/role/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+
+// reconciler creates and deletes Kubernetes resources to achieve the desired state of an RBAC Definition
+type reconciler struct {
+	ctx    context.Context
+	client controllerclient.Client
+}
+
+// Reconcile creates, updates, or deletes Kubernetes resources to match
+//   the desired state defined in an RBAC Definition
+func (r *reconciler) Reconcile() error {
```
p0lyn0mial: update the comment.

api/pkg/controller/rbac-user-cluster/role/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+			if err := r.client.Create(r.ctx, defaultClusterRole); err != nil {
+				return fmt.Errorf("failed to create the RBAC Cluster Role: %v", err)
+			}
+			return nil
+		}
+		return err
+	}
+	// compare rules with default. If don't match update for default
+	if !rbacusercluster.RolesMatches(clusterRole, defaultClusterRole) {
+		glog.V(debugLevel).Infof("updating the clusterRole %s because it doesn't match the expected one", defaultClusterRole.Name)
```
p0lyn0mial: change `clusterRole` to `Cluster Role` as in other places.

api/pkg/controller/rbac-user-cluster/role/controller.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Reconcile makes changes in response to cluster role related changes
+func (r *reconcileClusterRole) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	var err error
+
+	rdr := reconciler{client: r.Client, ctx: r.ctx}
+	cr := &rbacv1.ClusterRole{}
+	err = r.Get(r.ctx, request.NamespacedName, cr)
+	if err != nil {
```
p0lyn0mial: seems like this could be simplified to 
```
if !errors.IsNotFound(err) {
return err
}

err := rdr.Reconcile()
```

api/pkg/controller/rbac-user-cluster/role/controller.go
```diff
@@ -0,0 +1,75 @@
...
+	client.Client
+}
+
+// Reconcile makes changes in response to cluster role related changes
+func (r *reconcileClusterRole) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	var err error
+
+	rdr := reconciler{client: r.Client, ctx: r.ctx}
+	cr := &rbacv1.ClusterRole{}
+	err = r.Get(r.ctx, request.NamespacedName, cr)
```
p0lyn0mial: can we throw all roles not in `kube-system` namespace out ?

api/pkg/controller/rbac-user-cluster/role/controller.go
p0lyn0mial: or even 
```
err := rdr.Reconcile()
```

because the object we reconcile is not necessarily the one we are interested in, since we watch all roles in a namespace.

api/pkg/controller/rbac-user-cluster/matcher.go
```diff
@@ -0,0 +1,23 @@
+package rbacusercluster
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+)
+
+// RolesMatches compares roles OwnerReferences and Rules
+func RolesMatches(existingRole, requestedRole *rbacv1.ClusterRole) bool {
```
p0lyn0mial: nit: rename to `ClusterRoleMatches`

api/pkg/controller/rbac-user-cluster/mapper.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	clusterPolicyAPIGroup = "cluster.k8s.io"
+
+	machinedeployments = "machinedeployments"
```
p0lyn0mial, line 18: not needed anymore.

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+func (r *reconciler) ensureRBACRoleBinding(groupName string) error {
+	defaultClusterBinding := rbacusercluster.GenerateRBACClusterRoleBinding(groupName)
+
+	// check if cluster role already exists to make binding
+	clusterRole := &rbacv1.ClusterRole{}
+	if err := r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: defaultClusterBinding.Name}, clusterRole); err != nil {
+		glog.Error("getting  cluster role ", defaultClusterBinding.Name, " failed with error: ", err)
```
p0lyn0mial: consider changing to `unable to create ClusterRoleBinding because the corresponding ClusterRole doesn't exist`

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return nil
+}
+
+func (r *reconciler) ensureRBACRoleBinding(groupName string) error {
+	defaultClusterBinding := rbacusercluster.GenerateRBACClusterRoleBinding(groupName)
+
+	// check if cluster role already exists to make binding
+	clusterRole := &rbacv1.ClusterRole{}
+	if err := r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: defaultClusterBinding.Name}, clusterRole); err != nil {
```
p0lyn0mial: consider adding a comment explaining that if the `ClusterRole` does not exit we will return an error and the request will be retried.

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+
+// reconciler creates and deletes Kubernetes resources to achieve the desired state of an RBAC Definition
+type reconciler struct {
+	ctx    context.Context
+	client controllerclient.Client
+}
+
+// Reconcile creates, updates, or deletes Kubernetes resources to match
+//   the desired state defined in an RBAC Definition
+func (r *reconciler) Reconcile() error {
```
p0lyn0mial: does it have to be public ?

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *reconciler) ensureRBACRoleBinding(groupName string) error {
+	defaultClusterBinding := rbacusercluster.GenerateRBACClusterRoleBinding(groupName)
```
p0lyn0mial: the corresponding `ClusterRole` could be the owner of the corresponding `ClusterRoleBinding`, consider adding `OwnerReference`.

api/pkg/controller/rbac-user-cluster/role/controller.go
```diff
@@ -0,0 +1,75 @@
...
+// add adds a new Controller to mgr with r as the reconcile.reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to ClusterRoles
+	err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestForObject{})
```
alvaroaleman: As this controller doesn't reconcile individual ClusterRoles but a set of ClusterRoles, wouldn't it make sense to not use `EnqueueRequestForObject` but something like this to deduplicate  multiple events and make sure only one `Reconcile` is running at a time? https://github.com/kubermatic/kubermatic/pull/2822/files#diff-365f1d5947b4b54405fa2ec44a05e7b4R22

Also, wouldn't it make sense to add a `Predicates` func to the watch so this reconciler only runs if a ClusterRole it cares about got changed, not if any clusterrole got changed?

p0lyn0mial: Alvaro we are interested in individual ClusterRoles but I think we need to see the whole set (all roles in a namespace) to handle "individual ClusterRoles don't exist" scenario. For example our controller starts and the roles were not created, we need to create them for the first time. I think that `Predicate` for a name would rule out all the roles and nothing would be scheduled.

p0lyn0mial: regarding `EnqueueRequestsFromMapFunc` if i understand it correctly we could return a list of roles (namespace/name) we would like to additionally enqueue for each event.

p0lyn0mial: so we would have to implement `EnqueueRequestsFromMapFunc` that would return `[]{'kubermatic:viewers", "kubermatic:owners"}`  and `Predicate` that would be interested ony in resources with the following names `'kubermatic:viewers", "kubermatic:owners"`, is that correct ?

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
zreigz: This is a method from the framework interface

api/pkg/controller/rbac-user-cluster/role/controller.go
alvaroaleman: Oh you are right regarding using predicates, that would create issues for the initial creation.

> so we would have to implement EnqueueRequestsFromMapFunc that would return []{'kubermatic:viewers", "kubermatic:owners"} and Predicate that would be interested ony in resources with the following names 'kubermatic:viewers", "kubermatic:owners", is that correct ?

No. The `EnqueueRequestsFromMapFunc` basically returns a slice of `type.NamespacedName` and each of these will then trigger a run of the `Reconcile`. If you have a `Reconcile` that `reconciles` exactly one thing (For example the cluster controller does this) then it makes sense to have something like `EnqueueRequestForObject`.

But because this controller reconciles _multiple_ ClusterRoles per `Reconcile` run,  the `types.NamespacedName` that is returned should contain a `const` for Name, to make sure that:

* Only one `Reconcile` runs at a time
* If multiple events happen while a `Reconcile` is running, only one item is put in the Workq


The current approach means:

* On startup, `Reconcile` will get triggered for every existing Clusterrole in the cluster (65 times in the sample cluster I have here)
* Every single change to a ClusterRole will trigger another Reconcile (3 here I believe)

Total of 68 `Reconcile` runs

Whereas with the approach of enq a string constant we would:

* Initially trigger one reconcile, then have the key that is currently being worked on be put onto the workq but not trigger the Reconcile, as it already runs for that key
* All changes to ClusterRoles enq the same string const that is already in the workq so nothing happens

=> Total of two `Reconcile` runs

p0lyn0mial: Alvaro thanks for sharing, this is definitely something we should investigate, if you don't mind as a follow up. We would like to merge this pull to unblock API and testing. 

@zreigz is actually going to work on it right after the merge.

p0lyn0mial:
/lgtm

@alvaroaleman is there anything else we could improve/fix ?

api/pkg/controller/rbac-user-cluster/role/reconciler.go
p0lyn0mial: this comment is not relevant as `ClusteRole` resources are global/cluster-wide.

api/pkg/controller/rbac-user-cluster/matcher_test.go
```diff
@@ -0,0 +1,130 @@
+package rbacusercluster_test
```
alvaroaleman, line 1: Is it normal to use a different package name for `_test.go` files? Havent seen that before

api/pkg/controller/rbac-user-cluster/rolebinding/controller.go
```diff
@@ -0,0 +1,57 @@
...
+// add adds a new Controller to mgr with r as the reconcile.reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to ClusterRoleBindings
+	err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, &handler.EnqueueRequestForObject{})
```
alvaroaleman: Same as above, `EneueueRequestForObjcet` doesn't really make sense here IMHO

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (r *reconciler) ensureRBACClusterRoleBinding(groupName string) error {
+	defaultClusterBinding := rbacusercluster.GenerateRBACClusterRoleBinding(groupName)
+
+	// check if ClusterRole already exists to make binding
+	clusterRole := &rbacv1.ClusterRole{}
+	// if the corresponding ClusterRole does not exit then the reconciler return an error and the request will be retried.
+	if err := r.client.Get(r.ctx, controllerclient.ObjectKey{Namespace: metav1.NamespaceAll, Name: defaultClusterBinding.Name}, clusterRole); err != nil {
+		glog.Error("unable to create ClusterRoleBinding because the corresponding ClusterRole doesn't exist")
```
alvaroaleman: Why does the ClusterRole have to exist to create the ClusterRoleBinding?

alvaroaleman (commented):
I must say I find it very confusing that we have two controllers here. The only advantage of using more than one controller is that it parallelizes things, but performance deoesn't seem to be a concern here. You do know that you can make once controller watch different resources, e.G. one controller can watch both ClusterRoles and ClusterRoleBindings?

The drawback is that a lot more code is required because each controllers needs some scaffolding and understanding the "bigger picture" is rather difficult. But thats your teams call.

api/pkg/controller/rbac-user-cluster/matcher_test.go
zreigz: Using a separate _test package means that user can’t access the unexported fields and functions in the package. The test can see only exported API, so it makes the code more usable, safe and complete.

api/pkg/controller/rbac-user-cluster/rolebinding/reconciler.go
zreigz: done

zreigz:
@alvaroaleman @p0lyn0mial PTAL

alvaroaleman:
@zreigz There is currently a race in the cluster-controller/cloud-controller, which is why we hit AWS ratelimits.  https://github.com/kubermatic/kubermatic/pull/2844 fixes this, until that is merged, retesting will likely not help

zreigz:
@alvaroaleman thanks :)

PR #2685, opened 2019-01-24
Set a 90m timeout for the conformance tests

alvaroaleman:

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

/assign @mrIncompetent 

Superseeds #2608

### DIFF
--- a/api/hack/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci-run-minimal-conformance-tests.sh
@@ -148,7 +148,7 @@ helm upgrade --install --wait --timeout 300 \
 echo "Finished installing Kubermatic"
 
 echo "Starting conformance tests"
-./conformance-tests \
+timeout -s 9 90m ./conformance-tests \
   -debug \
   -worker-name=$BUILD_ID \
   -kubeconfig=$KUBECONFIG \


api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -148,7 +148,7 @@ helm upgrade --install --wait --timeout 300 \
 echo "Finished installing Kubermatic"
 
 echo "Starting conformance tests"
-./conformance-tests \
+timeout -s 9 90m ./conformance-tests \
```
thz, line 151: You're so brutal with your `-s 9`... How about a `-k 95m` instead?

(Had this comment on the left-side of the diff... Ughh...)

alvaroaleman: The tests have no feelings, the users waiting for them to return do

thz: It is not about feelings. It is about giving the process a chance to exit gracefully and SIGKILLing it a few moments after that. You're probably right that in this particular moment with the current implementation of the process it might make no difference. But when you somewhen change it you now need to remember that you will not have a chance to catch a TERM for whatever graceful termination. TERM and then KILL is quite a standard in UNIX'ish environments.

alvaroaleman: We never had that chance before and the runner has no functionality implemented to act on signals so no, I do not see the point of making something possible we do not use

PR #2690, opened 2019-01-24
Use internal API server service (Without ClusterIP) + Do custom endpoint reconciling to avoid handling the NodePort

mrIncompetent:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/cluster/userclusterresources.go
+++ b/api/pkg/controller/cluster/userclusterresources.go
@@ -601,3 +601,68 @@ func (cc *Controller) userClusterEnsureServices(c *kubermaticv1.Cluster) error {
 
 	return nil
 }
+
+func (cc *Controller) reconcileAPIServerEndpoints(c *kubermaticv1.Cluster) error {
+	const endpointsName = "kubernetes"
+	if !c.Status.Health.Apiserver {
+		return nil
+	}
+
+	s, err := cc.serviceLister.Services(c.Status.NamespaceName).Get(resources.ApiserverExternalServiceName)
+	if err != nil {
+		return fmt.Errorf("failed to get service '%s': %v", resources.ApiserverExternalServiceName, err)
+	}
+
+	ipSet, err := getExternalIPv4Set(c.Address.ExternalName)
+	if err != nil {
+		return err
+	}
+
+	endpoints := &corev1.Endpoints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: endpointsName,
+		},
+		Subsets: []corev1.EndpointSubset{
+			{
+				Ports: []corev1.EndpointPort{
+					{
+						Name:     "https",
+						Port:     s.Spec.Ports[0].NodePort,
+						Protocol: corev1.ProtocolTCP,
+					},
+				},
+			},
+		},
+	}
+
+	for _, ip := range ipSet.List() {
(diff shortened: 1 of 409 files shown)


api/pkg/controller/cluster/userclusterresources.go
```diff
@@ -601,3 +601,68 @@ func (cc *Controller) userClusterEnsureServices(c *kubermaticv1.Cluster) error {
 
 	return nil
 }
+
+func (cc *Controller) reconcileAPIServerEndpoints(c *kubermaticv1.Cluster) error {
```
alvaroaleman, line 605: IMHo this should run inside the user cluster controller manages as a dedicated controller and also have a lister for the services in the seed NS of the cluster to ensure we react quickly to any changes

api/pkg/resources/apiserver/service.go
```diff
@@ -20,15 +20,16 @@ func Service(data resources.ServiceDataProvider, existing *corev1.Service) (*cor
 	se.Labels = resources.BaseAppLabel(name, nil)
 
 	se.Spec.Type = corev1.ServiceTypeClusterIP
+	se.Spec.ClusterIP = corev1.ClusterIPNone
```
alvaroaleman, line 23: Why?

alvaroaleman (commented):
Would love to see the `Use internal API server service` part.

Is there a reason why that must be done together with the custom endpoint reconciling?

mrIncompetent:
This PR does change 3 things (Which i'm not certain if they are worth it):
- Use the `apiserver` service instead of the `apiserver-external` service
  - That itself should not have any impact. The `apiserver-external` & `apiserver` are both ClusterIP services after all.
- Use RoundRobin DNS (`clusterIP=none`) instead of a ClusterIP for the `apiserver` service. This was mostly a "esoteric" change with the hope of better connection handling for clients.
- Use custom endpoint reconciliation 
  - This is required since we now use RoundRobin DNS instead of a ClusterIP. As we cannot re-map ports of a service.

Just using the `apiserver` service instead of the `apiserver-external` has technically no impact at all.

We might close this PR after all as it seems our connection issues might have been caused by GKE.

mrIncompetent:
@alvaroaleman 
Please give some feedback on https://github.com/kubermatic/kubermatic/pull/2690#issuecomment-463147474

Otherwise i'll close this PR by the end of the week

alvaroaleman:
> This PR does change 3 things (Which i'm not certain if they are worth it):

> Use the apiserver service instead of the apiserver-external service
That itself should not have any impact. The apiserver-external & apiserver are both ClusterIP services after all.
Use RoundRobin DNS (clusterIP=none) instead of a ClusterIP for the apiserver service. This was mostly a "esoteric" change with the hope of better connection handling for clients.
Use custom endpoint reconciliation
This is required since we now use RoundRobin DNS instead of a ClusterIP. As we cannot re-map ports of a service.
Just using the apiserver service instead of the apiserver-external has technically no impact at all.

> We might close this PR after all as it seems our connection issues might have been caused by GKE.

The reason I wanted this is that I do not want our controller manager to do the extra hop out of the seed and back in when it talks to user clusters. But this PR does not achieve that anyways so I'd vote for just closing it. If you do want it, just re-open.

PR #2691, opened 2019-01-25
Remove static users from dex config

floreks:
Removes static clients configuration from dex. We can add user manually using dex client now. No need to hardcode them.

**Release note**:

```release-note
NONE
```


zreigz (approved):
/lgtm
/approve

PR #2699, opened 2019-01-25
moves handler/user.go into its own pkg

p0lyn0mial:
**What this PR does / why we need it**: simply moves handler/user.go into its own pkg

**Release note**:

```
NONE
```


p0lyn0mial:
@zreigz PTAL

zreigz (approved):
/lgtm
/approve

PR #2702, opened 2019-01-25
Improve e2e tests

alvaroaleman:

This pull adds several improvements to the e2e tests:

* The PVC and LB test are run at the end of the suite and each attempt has a distinct namespace, this way we don't have to explicitly clean up the namespaces which saves us some time
* All `echo` outputs in the `ci-run-minimal-conformance-tests.sh` scripts now have a date prefix
* The Junit artifacts are now uploaded to GCS which makes the visible in Deck, e.G. here: https://prow.loodse.com/view/gcs/prow-data/pr-logs/pull/kubermatic_kubermatic/2702/pull-kubermatic-e2e-aws-coreos-1.11/1088869637627580417
* We now use the git HEADs sha as image tag and check if an image with that tag already exists prior to building Kubermatic and the Docker image
* We retry pushing the Docker image
* We retry the `helm install`
* We build the `conformance-tests` cli _after_ the helm install to make sure we fail as fast as possible when something with the helm chart is wrong

/assign @mrIncompetent 

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/hack/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci-run-minimal-conformance-tests.sh
@@ -83,9 +83,6 @@ export KUBECONFIG=/tmp/kubeconfig
 export VALUES_FILE=/tmp/values.yaml
 echodate "Successfully got secrets from Vault"
 
-echodate "Building conformance-tests cli"
-time go build -v github.com/kubermatic/kubermatic/api/cmd/conformance-tests
-echodate "Finished building conformance-tests cli"
 
 if [[ ! -f $HOME/.docker/config.json ]]; then
   docker ps &>/dev/null || start-docker.sh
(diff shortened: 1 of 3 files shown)


api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -58,26 +81,31 @@ vault kv get -field=values.yaml \
...
+echodate "Successfully got secrets from Vault"
 
 
 if [[ ! -f $HOME/.docker/config.json ]]; then
-  echo "Logging into quay.io"
-  docker login -u $QUAY_IO_USERNAME -p $QUAY_IO_PASSWORD quay.io
-  echo "Logging into dockerhub"
+  docker ps &>/dev/null || start-docker.sh
+  mkdir  -p $HOME/.docker
+  echo '{"experimental": "enabled"}' > ~/.docker/config.json
```
mrIncompetent, line 90: Why do we need to activate the docker experimental features?

api/hack/ci-run-minimal-conformance-tests.sh
```diff
@@ -58,26 +81,31 @@ vault kv get -field=values.yaml \
...
-echo "Building kubermatic-controller-manager"
-time make -C api build
-echo "Finished building conformance-tests and kubermatic-controller-manager"
-
-echo "Building docker image"
-./api/hack/push_image.sh ${BUILD_ID}
-echo "Finished building and pushing docker images"
+# Only build kubermatic binaries and docker image if it doesn't exist yet
+# We use dockerhub because docker manifest inspect doesn't seem to work on quay
+if ! docker manifest inspect docker.io/kubermatic/api:$GIT_HEAD_HASH &>/dev/null; then
```
mrIncompetent, line 98: Add a comment to the `push_image.sh` script that we have a dependency on the docker hub images in our e2e tests

api/hack/ci-run-minimal-conformance-tests.sh
alvaroaleman: Because the `docker manifest inspect docker.io/kubermatic/api` is an experimental feature

api/hack/ci-run-minimal-conformance-tests.sh
alvaroaleman: Why? The `push_image.sh` script is not used at all here

mrIncompetent: Sorry missed that you replaced the usage :+1:

PR #2709, opened 2019-01-28
remove OwnerRef from in-cluster metrics-server service

mrIncompetent:
As this Service is located inside the user cluster, the GC must have been removing it on a regular basis.

**Release note**:
```release-note
NONE
```


alvaroaleman:
Thank you for making my day :heart: 

/lgtm
/approve

mrIncompetent:
v2.8 is not affected.
We moved the metrics-server from the user cluster to seed with v2.9.

But I would rather merge master into v2.9 as we had a lot of smaller bugfixes and no feature development.

PR #2714, opened 2019-01-29
[WIP] Use go test for Kubermatic e2e tests

mrIncompetent:
**Release note**:
```release-note
NONE
```


api/cmd/e2e/context.go
```diff
@@ -0,0 +1,40 @@
...
+	clusterLister          kubermaticv1lister.ClusterLister
+	kubermaticClient       kubermaticclientset.Interface
+	kubeClient             kubernetes.Interface
+	clusterClientProvider  *clusterclient.Provider
+	dcs                    map[string]provider.DatacenterMeta
+	nodeCount              int
+	workerName             string
+	deleteClustersWhenDone bool
+	workingDir             string
+	testBinRoot            string
```
alvaroaleman, line 23: I would much prefer it if we passed the `t *testing.T` through so individual test funcs can call `t.Errorf()`

api/cmd/e2e/ginkgo.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+)
+
+func testConformanceParallel(ctx *TestContext, t *testing.T) {
+	testConformance(ctx, t, ctx.nodeCount*7, "parallel", `\[Conformance\]`, `\[Serial\]`)
+}
+
+func testConformanceSerial(ctx *TestContext, t *testing.T) {
+	// Broken on aws. See https://github.com/kubernetes/kubernetes/pull/72965
+	skip := `should not cause race condition when used for configmap`
```
alvaroaleman: Instead of skipping this unconditionally, could you change the code to only test this if cluster minor version is 13? The info is available in the `TestContext`

api/cmd/e2e/ginkgo.go
```diff
@@ -0,0 +1,80 @@
...
+	"path"
+	"testing"
+	"time"
+)
+
+func testConformanceParallel(ctx *TestContext, t *testing.T) {
+	testConformance(ctx, t, ctx.nodeCount*7, "parallel", `\[Conformance\]`, `\[Serial\]`)
+}
+
+func testConformanceSerial(ctx *TestContext, t *testing.T) {
```
alvaroaleman, line 17: Could you please not use `ctx` as variable name for the `TestContext`? By convention, `ctx` is a `context.Context`, this is IMHO confusing

api/cmd/e2e/ginkgo.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"testing"
+	"time"
+)
+
+func testConformanceParallel(ctx *TestContext, t *testing.T) {
```
alvaroaleman, line 13: Instead of having one signature for the gingko type tests and another one for the LB and PV test, could we define a `type testExecutor(testContext *TestContext)`? Then we can put all of them into a slice.

Also I'd like the test funcs themselve to call `t.Parallel()` if they support it instead of doing that in `executeClusterTests`

Also I think it would be better if the tests themselves check if they are supported by using some info from `TestContext` and if not, call `t.Skip()`

This would make the funcs completely self-contained instead of having the required info scattered around

api/cmd/e2e/ginkgo.go
```diff
@@ -0,0 +1,80 @@
...
+		fmt.Sprintf("--repo-root=%s", kubernetesDir),
+		fmt.Sprintf("--report-dir=%s", ctx.workingDir),
+		fmt.Sprintf("--report-prefix=%s", name+"_"),
+		fmt.Sprintf("--kubectl-path=%s", path.Join(binDir, "kubectl")),
+		fmt.Sprintf("--kubeconfig=%s", ctx.clusterContext.kubeconfig),
+		fmt.Sprintf("--num-nodes=%d", ctx.nodeCount),
+		"--provider=local",
+	}
+
+	cmdCtx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
```
alvaroaleman: Instead of hardcoding this, wouldn't it make more sense to put the timeout in the layer above? Parallel and Serial have different timeouts I think. Also a timeout of 1h feels way too big for both of them.

api/cmd/e2e/ginkgo.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		t.Fatalf("failed to create ginkgo logfile: %v", err)
+	}
+	defer f.Close()
+	defer f.Sync()
+
+	cmd.Stdout = f
+	cmd.Stderr = f
+
+	// Convenience function to flush the logs to the file more often. Enables watching the ginkgo logs
```
alvaroaleman, line 64: :+1:

api/cmd/e2e/custom_tests.go
```diff
@@ -0,0 +1,238 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	storageTestNamespace = "pvc-test"
+	lbTestNamespace      = "lb-test"
+)
+
+func testStorageSupport(ctx *TestContext, r *R) {
+	defer deleteNamespace(storageTestNamespace, ctx, r)
```
alvaroaleman, line 27: Is this required? I am pretty sure ginkgo conformance tests dont care about other namespaces that exist as long as their names dont interfer, so I'd prefer it if we simply made the storage and lb  tests use a unique namespace name per run so we don't have to clean anything up so the whole thing is faster

api/cmd/e2e/run.sh
```diff
@@ -0,0 +1,19 @@
...
+
+go test \
+  -parallel 999 \
+  -tags=e2e \
+  -run=TestE2E/.*/aws/coreos \
+  github.com/kubermatic/kubermatic/api/cmd/e2e \
+  -kubeconfig=/home/henrik/go/src/github.com/kubermatic/secrets/seed-clusters/dev.kubermatic.io/kubeconfig \
+  -datacenters=/home/henrik/go/src/github.com/kubermatic/secrets/seed-clusters/dev.kubermatic.io/datacenters.yaml \
+  -test-bin-dir=/opt/kube-test/ \
+  -timeout=6h \
```
alvaroaleman, line 17: Please:

* Lower the timeout significantly (90 min?)
* Make this use helm
* Add an exit trap that does cleanup

alvaroaleman (commented):
A couple of nits, generally speaking I like the PR very much.

Could you also please already add optional presubmits that leverage this so we can verify it actually works?

PR #2715, opened 2019-01-29
add promtool container

mrIncompetent:
**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/hack/verify-user-cluster-prometheus-configs.sh
+++ b/api/hack/verify-user-cluster-prometheus-configs.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "This script requires write access to /etc/etcd/pki/client & /etc/kubernetes as it needs to write dummy certs."
+echo "This is required because the promtool checks if files inside the prometheus.yaml actually exist."
+mkdir -p /etc/etcd/pki/client
+touch /etc/etcd/pki/client/apiserver-etcd-client.crt
+touch /etc/etcd/pki/client/apiserver-etcd-client.key
+mkdir -p /etc/kubernetes
+touch /etc/kubernetes/prometheus-client.crt
+touch /etc/kubernetes/prometheus-client.key
+
+cd "$(dirname "$0")/../"
+
+for CM in pkg/resources/test/fixtures/configmap-*-prometheus.yaml; do
+  echo "Checking ${CM} ..."
+  cat ${CM} | yaml2json | jq -r '.data["rules.yaml"]' > rules.yaml
+  promtool check rules rules.yaml
+  rm rules.yaml
+
+  cat ${CM} | yaml2json | jq -r '.data["prometheus.yaml"]' > prometheus.yaml
+  promtool check config prometheus.yaml
+  rm prometheus.yaml
+done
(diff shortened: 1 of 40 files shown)


containers/promtool/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM alpine
```
alvaroaleman, line 1: Why not simply use `quay.io/prometheus/prometheus:v2.7.0`? Then we don't have to maintain yet another image

mrIncompetent: It does not contain make

alvaroaleman: Ah, good point.

containers/promtool/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+RUN apk add --no-cache -U curl make
+
+RUN curl -L https://github.com/prometheus/prometheus/releases/download/v2.7.0/prometheus-2.7.0.linux-amd64.tar.gz | tar -xvz && \
```
alvaroaleman, line 5: :+1:

mrIncompetent:
@alvaroaleman I added a hack script which checks if the user cluster prometheus configs are valid.

api/pkg/resources/test/fixtures/configmap-aws-1.10.0-prometheus.yaml
```diff
@@ -9,7 +9,13 @@ data:
     rule_files:
     - "/etc/prometheus/config/rules*.yaml"
     scrape_configs:
-    some test
```
mrIncompetent, line 12: Had to replace it with some valid scrape config

api/hack/verify-user-cluster-prometheus-configs.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "This script requires write access to /etc/etcd/pki/client & /etc/kubernetes as it needs to write dummy certs."
+echo "This is required because the promtool checks if files inside the prometheus.yaml actually exist."
+mkdir -p /etc/etcd/pki/client
+touch /etc/etcd/pki/client/apiserver-etcd-client.crt
```
mrIncompetent, line 7: I don't have any better ideas to avoid this.
The promtool has no flag to deactivate those checks

alvaroaleman: Its fine, but maybe replace the echo with a comment? And in csae you are bored and like long discussions, you can create an upstream PR to add a flag to deactivate those :)

mrIncompetent: > And in csae you are bored and like long discussions, you can create an upstream PR to add a flag to deactivate those :)

Saw today a 83 comment long(Just 2 participants) discussion on the `runbook` field within the Prometheus alert type.
Not even close to be that bored.

mrIncompetent: > Its fine, but maybe replace the echo with a comment? 

Thought about that, but wanted to make it more visible without going the extra mile with permission checking + handling.

alvaroaleman: :+1:

PR #2722, opened 2019-01-31
Migrate service reconciling

mrIncompetent:
This migrates the service reconciling to the new dynamic client.

**Release note**:
```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubermatic/kubermatic/2722)


api/pkg/controller/monitoring/resources.go
```diff
@@ -169,26 +169,6 @@ func (c *Controller) ensureStatefulSets(cluster *kubermaticv1.Cluster, data *res
 	return resources.EnsureStatefulSets(creators, cluster.Status.NamespaceName, c.dynamicClient, c.dynamicCache, resources.ClusterRefWrapper(cluster))
 }
 
-// GetServiceCreators returns all service creators that are currently in use
-func GetServiceCreators() []resources.ServiceCreator {
```
mrIncompetent, line 173: Those services are not used at all... I just removed them.

api/pkg/controller/cluster/resources.go
```diff
@@ -135,28 +135,22 @@ func (cc *Controller) ensureNamespaceExists(c *kubermaticv1.Cluster) (*kubermati
 }
 
 // GetServiceCreators returns all service creators that are currently in use
-func GetServiceCreators() []resources.ServiceCreator {
+func GetServiceCreators(data *resources.TemplateData) []resources.ServiceCreator {
```
alvaroaleman, line 138: Why the data? It isnt used anywhere

api/pkg/controller/cluster/userclusterresources.go
```diff
@@ -567,9 +567,9 @@ func (cc *Controller) userClusterEnsureServices(c *kubermaticv1.Cluster) error {
 		return err
 	}
 
-	for _, create := range GetUserClusterServiceCreators() {
+	for _, create := range GetUserClusterServiceCreators(data) {
 		var existing *corev1.Service
-		service, err := create(data, nil)
```
alvaroaleman, line 572: Why pass that in instead of nil?

api/pkg/controller/monitoring/resources.go
alvaroaleman: Althought I'd argue that if someone wants data from kubhestate metrics they are required - althought I presume they don't work anyways currently, as we'd have to manually manage the endpoints

api/pkg/resources/prometheus/statefulset.go
```diff
@@ -54,7 +54,6 @@ func StatefulSetCreator(data *resources.TemplateData) resources.StatefulSetCreat
 
 		set.Spec.Replicas = resources.Int32(1)
 		set.Spec.UpdateStrategy.Type = appsv1.RollingUpdateStatefulSetStrategyType
-		set.Spec.ServiceName = resources.PrometheusServiceName
```
alvaroaleman, line 57: Isn't that requierd for a statefulset?

alvaroaleman (commented):
I would prefer it if we kept all in-clouster resources untouched except for urgent fixes until  @nikhita is done with the user cluster controller

api/pkg/resources/prometheus/statefulset.go
mrIncompetent: Nope

api/pkg/controller/cluster/userclusterresources.go
mrIncompetent: I changed the `metricsserver.ExternalNameServiceCreator` function to match the existing ServiceCreators. All ServiceCreators expect a valid Service struct.

For the StatefulSets & VerticalPodAutoscalers we ensure that in a wrapper function: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/resources.go#L747
(For the Services as well, just not here)

The wrapper is not used here, as it converts the create function to a ObjectCreator, which is not usable here - as it uses API calls instead of a lister.
For simplicity i would just keep it that way.
It's a necessary change for the UCC anyway as its supposed to use the dynamic client + ensure functions.

api/pkg/controller/cluster/resources.go
mrIncompetent: Used here: https://github.com/kubermatic/kubermatic/pull/2722/files/fa22915b688d8fca6684355c58070e89ed16b90c#diff-a4282b98d1211368af8912d690d2794eR143

mrIncompetent:
@alvaroaleman 
> I would prefer it if we kept all in-clouster resources untouched except for urgent fixes until @nikhita is done with the user cluster controller

I agree, but in this case its a change in favor of the UCC as those changes would be made anyway when moving to the Lister approach (Or are we planning on doing API calls?).

api/pkg/resources/apiserver/service.go
```diff
@@ -4,73 +4,61 @@ import (
...
+	return func(se *corev1.Service) (*corev1.Service, error) {
+		se.Name = resources.ApiserverInternalServiceName
+		se.Labels = resources.BaseAppLabel(name, nil)
 
-	if len(se.Spec.Ports) == 0 {
+		se.Spec.Type = corev1.ServiceTypeClusterIP
+		se.Spec.Selector = map[string]string{
+			resources.AppLabelKey: name,
+		}
 		se.Spec.Ports = []corev1.ServicePort{
```
alvaroaleman, line 20: Sometimes I hate git(hub) for being not smart enough to simply show that this is the same code as before but was just moved...

api/pkg/resources/metrics-server/service.go
```diff
@@ -5,29 +5,24 @@ import (
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Service returns a service for the metrics server
-func Service(data resources.ServiceDataProvider, existing *corev1.Service) (*corev1.Service, error) {
-	se := existing
-	if se == nil {
-		se = &corev1.Service{}
-	}
+// ServiceCreator returns the function to reconcile the DNS service
```
alvaroaleman: I'd argue its still the metrics-server service

api/pkg/resources/resources.go
```diff
@@ -366,7 +362,7 @@ type StatefulSetCreator = func(existing *appsv1.StatefulSet) (*appsv1.StatefulSe
 type VerticalPodAutoscalerCreator = func(existing *autoscalingv1beta.VerticalPodAutoscaler) (*autoscalingv1beta.VerticalPodAutoscaler, error)
 
 // ServiceCreator defines an interface to create/update Services
-type ServiceCreator = func(data ServiceDataProvider, existing *corev1.Service) (*corev1.Service, error)
+type ServiceCreator = func(existing *corev1.Service) (*corev1.Service, error)
```
alvaroaleman, line 365: Wouldn't it make sense to have one data struct for all resources, so we can eventually unify their handling?

alvaroaleman (commented):
> I agree, but in this case its a change in favor of the UCC as those changes would be made anyway when moving to the Lister approach (Or are we planning on doing API calls?).

Certainly not. My main concern is that I do not want @nikhita to have to rebase her work. But then again it probably makes most sense to move the relevant resources over in a package that is owned by the UCC and not have her PR change anything inside `api/pkg/resources` other than removing or WDYT?

api/pkg/resources/resources.go
mrIncompetent: You mean a single function signature for all resources?
Like https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/resources.go#L408 ?

mrIncompetent:
> Certainly not. My main concern is that I do not want @nikhita to have to rebase her work. 

Good point! I did a quick check of the current UCC PR - and rebased it onto this PR without any conflicts + compiles. Sure there might be other issues but it might be good enough :)

> But then again it probably makes most sense to move the relevant resources over in a package that is owned by the UCC and not have her PR change anything inside api/pkg/resources other than removing or WDYT?

:+1:

mrIncompetent:
cc @xrstf Any objections? As this removes the prom & kube-state-metrics services

xrstf:
If we don't need the services to scrape kube-state-metrics, I'm fine with this.

/approve

PR #2735, opened 2019-02-01
Use codegen to implement the typed resource reconcile functions

mrIncompetent:
This PR adds codegeneration for the Reconcile* (Renamed from Ensure*) functions.

Why?
- Having typed functions avoids type conversions in our "business logic"
- Using codegen makes it easier to maintain them

**Release note**:
```release-note
NONE
```


api/pkg/resources/reconcile.go
```diff
@@ -0,0 +1,112 @@
+// This file is generated. DO NOT EDIT.
```
alvaroaleman: Can we rename this file in a way that makes it obvious its generated? Maybe zse the `zz_generated_reconcicle.go` patten from kube?

api/cmd/codegen/reconcile/main.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+
+	informerutil "github.com/kubermatic/kubermatic/api/pkg/util/informer"
+
+  "k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimecache "sigs.k8s.io/controller-runtime/pkg/cache"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+{{ range .Resources }}
+{{- if .ResourceImportPath }}
+	{{ .ImportAlias }} "{{ .ResourceImportPath }}"
```
alvaroaleman: How do we avoid duplicate reports?

api/Makefile
```diff
@@ -1,7 +1,7 @@
 export CGO_ENABLED?=0
 REPO=kubermatic/api
 GO?=go
-CMD=$(filter-out OWNERS, $(notdir $(wildcard ./cmd/*)))
+CMD=$(filter-out OWNERS codegen, $(notdir $(wildcard ./cmd/*)))
```
alvaroaleman: Having this in cmd doesn't really feel right since this is not an app, althought I don't have a great idea for a better place off of the top of my head either

alvaroaleman (commented):
> Having typed functions avoids type conversions in our "business logic"

But we already had them before?

mrIncompetent:
> But we already had them before?

Completely right, that argument is just pro typed reconcile funcs. Not pro codegen directly

api/cmd/codegen/reconcile/main.go
mrIncompetent: If `. ResourceImportPath` is not set, no new import will be added. So it's up to the caller to make this sure. 
Not particular pretty but I can do a more solid handling by deduplicating the imports via a map. As you prefer

api/pkg/resources/reconcile.go
mrIncompetent: Nice idea! I'll change it

PR #2742, opened 2019-02-03
fix bad volume mount for Prometheus 2.7+

xrstf:
The Prometheus 2.7.1 update PR missed changing the tsdb path parameter.

**Release note**:
```release-note
NONE
```


xmudrii:
/lgtm

Feel free to approve if you're ready to ship it

PR #2745, opened 2019-02-05
enforce kubelet-apiserver version constaints on cluster upgrade

### LINKED ISSUE #2542: Enforce kubelet-apiserver version constaints on cluster upgrade
**User Story**
The Kubermatic API server must refuse to upgrade a cluster to a version incompatible with existing nodes.

**Acceptance criteria**
- [ ] `patchCluster` fetches all Machines and MachineDeployments within the cluster. It checks (using #2539) whether all Machines and MachineDeployments are running kubelets compatible with the desired cluster control-plane version. If there are incompatible kubeletes in the cluster, the endpoint returns an errror and no changes are committed to the cluster.
 - [ ] `TestPatchCluster` runs a test scenario for the case of incompatible versions

kdomanski:
To prevent clusters from being upgraded to versions incompatible with existing nodes

Fixes #2542

**Release note**:

```release-note
Clusters can no longer be upgraded to versions incompatible with existing nodes.
```

### DIFF
--- a/api/pkg/handler/version_skew.go
+++ b/api/pkg/handler/version_skew.go
@@ -3,8 +3,15 @@ package handler
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/Masterminds/semver"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/v1/common"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
+
+	kubermaticapiv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1alpha1clientset "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
 )
 
 // errVersionSkew denotes an error condition where a given kubelet/controlplane version pair is not supported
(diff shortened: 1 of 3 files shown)


api/pkg/handler/version_skew.go
```diff
@@ -45,3 +52,88 @@ func ensureVersionCompatible(controlPlane *semver.Version, kubelet *semver.Versi
...
+	}
+
+	// get deduplicated list of all used kubelet versions
+	kubeletVersions, err := getKubeletVersions(machineClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the list of kubelet versions used in the cluster: %v", err)
+	}
+
+	// this is where the incompatible versions shall be saved
+	incompatibleVersionsSet := map[string]bool{}
```
maciaszczykm, line 71: 👍

api/pkg/handler/version_skew.go
```diff
@@ -45,3 +52,88 @@ func ensureVersionCompatible(controlPlane *semver.Version, kubelet *semver.Versi
...
+				incompatibleVersionsSet[kubeletVersion.String()] = true
+				continue
+			}
+
+			// other error types
+			return nil, fmt.Errorf("failed to check compatibility between kubelet %q and control plane %q: %v", kubeletVersion, clusterVersion, err)
+		}
+	}
+
+	// deduplicate
```
maciaszczykm: This comment was a bit confusing to me at the first glance as map should be already "deduplicated". Isn't it rather conversion back to slice?

api/pkg/handler/version_skew.go
```diff
@@ -45,3 +52,88 @@ func ensureVersionCompatible(controlPlane *semver.Version, kubelet *semver.Versi
...
+	incompatibleVersionsList := []string{}
```
maciaszczykm: nit: `var incompatibleVersionsList []string` would be a bit better in case of empty `incompatibleVersionsSet`.

api/pkg/handler/version_skew.go
```diff
@@ -45,3 +52,88 @@ func ensureVersionCompatible(controlPlane *semver.Version, kubelet *semver.Versi
...
+	if err != nil {
+		return nil, common.KubernetesErrorToHTTPError(err)
+	}
+
+	kubeletVersionsSet := map[string]bool{}
+
+	// first let's go through the legacy non-MD nodes
+	for _, m := range machineList.Items {
+		// Do not list Machines that are controlled, i.e. by Machine Set.
+		if len(m.ObjectMeta.OwnerReferences) != 0 {
```
maciaszczykm: What about turning the if other way around? In this case:

```go
if len(m.ObjectMeta.OwnerReferences) == 0 {
    ver := strings.TrimSpace(m.Spec.Versions.Kubelet)
    kubeletVersionsSet[ver] = true
}
```

Seems to be easier to understand.

api/pkg/handler/version_skew.go
```diff
@@ -45,3 +52,88 @@ func ensureVersionCompatible(controlPlane *semver.Version, kubelet *semver.Versi
...
+		kubeletVersionsSet[ver] = true
+	}
+
+	// now the deployments
+	for _, md := range machineDeployments.Items {
+		ver := strings.TrimSpace(md.Spec.Template.Spec.Versions.Kubelet)
+		kubeletVersionsSet[ver] = true
+	}
+
+	// deduplicated list
```
maciaszczykm, line 130: In this form, it is easier to understand that we are creating a deduplicated list here :)

maciaszczykm (commented):
Looks good to me. Left a few minor comments.

api/pkg/handler/version_skew.go
maciaszczykm: https://stackoverflow.com/questions/29164375/correct-way-to-initialize-empty-slice

kdomanski:
@maciaszczykm I addressed the issues you pointed out.

PR #2750, opened 2019-02-06
Delete in-cluster resources before deleting nodes

alvaroaleman:

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/pkg/clusterdeletion/deletion.go
+++ b/api/pkg/clusterdeletion/deletion.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	kubermaticapiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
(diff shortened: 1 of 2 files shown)


mrIncompetent:
Why are we doing this change?

alvaroaleman:
> Why are we doing this change?

Because in a situation where a user deployed a controller for one of PVs/LBs and wants them to cleaned up on deletion, we currently end up in a dead lock, as we delete all nodes and hence the respective controller pods before deleting the PVs/LBs

mrIncompetent:
> > Why are we doing this change?

> Because in a situation where a user deployed a controller for one of PVs/LBs and wants them to cleaned up on deletion, we currently end up in a dead lock, as we delete all nodes and hence the respective controller pods before deleting the PVs/LBs

In case we delete a PVC which gets consumed by a Pod, the PVC won't disappear as long the Pod is running. Deleting a PVC does not trigger a pod to restart - instead the running Pod blocks the PVC deletion.
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#storage-object-in-use-protection

So the correct solution would rather be to delete the PVC & delete the consumer?

alvaroaleman:
> So the correct solution would rather be to delete the PVC & delete the consumer?

Good point. Will fix it up later

mrIncompetent:
@alvaroaleman Shall we close this PR ?

alvaroaleman:
I actually want to keep it as a reminder to myself

mrIncompetent:
@alvaroaleman You mind if we only  cherry-pick this into v2.10 ?

alvaroaleman:
No.

/cherrypick release/v2.10

PR #2758, opened 2019-02-06
add an endpoint that returns the api server's build version string

kdomanski:
In the UI we will be able to show not only the UI version, but the version of the backend as well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -2617,6 +2617,29 @@
         }
       }
     },
+    "/api/v1/version": {
+      "get": {
+        "produces": [
+          "application/json"
+        ],
+        "tags": [
+          "versions"
+        ],
+        "summary": "Get versions of running Kubermatic components.",
+        "operationId": "getKubermaticVersion",
+        "responses": {
+          "200": {
+            "description": "KubermaticVersions",
+            "schema": {
+              "$ref": "#/definitions/KubermaticVersions"
+            }
+          },
+          "default": {
+            "$ref": "#/responses/errorResponse"
+          }
+        }
+      }
+    },
     "/api/v1/versions": {
       "get": {
         "description": "Lists all versions which don't result in automatic updates",
@@ -3709,6 +3732,18 @@
             "$ref": "#/definitions/NamedAuthInfo"
           },
           "x-go-name": "AuthInfos"
+        }
+      },
+      "x-go-package": "github.com/kubermatic/kubermatic/api/pkg/api/v1"
+    },
+    "KubermaticVersions": {
+      "type": "object",
+      "title": "KubermaticVersions describes the versions of running Kubermatic components.",
+      "properties": {
+        "api": {
+          "description": "Version of the Kubermatic API server.",
+          "type": "string",
+          "x-go-name": "API"
         }
       },
       "x-go-package": "github.com/kubermatic/kubermatic/api/pkg/api/v1"
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"encoding/json"
+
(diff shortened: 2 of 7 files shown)


api/cmd/kubermatic-api/swagger.json
```diff
@@ -2524,6 +2524,29 @@
...
     },
+    "/api/v1/version": {
+      "get": {
+        "produces": [
+          "application/json"
+        ],
+        "tags": [
+          "versions"
+        ],
+        "summary": "Get versions of running Kubermatic components.",
```
p0lyn0mial, line 2628: "components" so in the future we will list the version of controllers as well ? In that case how we will get a version of kubermatic controller manager for example ?

api/pkg/resources/resources.go
```diff
@@ -40,6 +40,9 @@ import (
 // KUBERMATICCOMMIT is a magic variable containing the git commit hash of the current (as in currently executing) kubermatic api. It gets feeded by Makefile as a ldflag.
 var KUBERMATICCOMMIT string
 
+// KUBERMATICGITTAG is a magic variable containing the output of `git describe` for the current (as in currently executing) kubermatic api. It gets fed by Makefile as an ldflag.
+var KUBERMATICGITTAG = "manual_build"
```
p0lyn0mial, line 44: nit: thinking about a different name how about `unofficial_build`, `dev_build`, `unstable_build` ?

api/pkg/api/v1/types.go
```diff
@@ -682,3 +683,10 @@ type Event struct {
 	// Type of this event (normal, warning), new types could be added in the future
 	Type string `json:"type,omitempty"`
 }
+
+// KubermaticVersions describes the versions of running Kubermatic components.
+// swagger:model KubermaticVersions
+type KubermaticVersions struct {
+	// Version of the Kubermatic API server.
+	API string `json:"api"`
```
p0lyn0mial, line 695: would be nice to print Major,Minor and Patch as well. For example:
```
Major:"2", Minor:"9", Patch:"1", GitCommit:"2344234"
```
What do you think ?

api/cmd/kubermatic-api/swagger.json
kdomanski: Maybe in the future, maybe not. So far the cotroller-manager is always deployed at the same versions as API server.

api/pkg/resources/resources.go
kdomanski: Most builds are dev and unstable, the point of the default value is that this does not come from the CI, so no commit information is available.

api/pkg/api/v1/types.go
kdomanski: I think it would be unnecessary expansion of the scope. The point of the PR is to display the version in the UI.

p0lyn0mial: @maciaszczykm would it be better if API version is structured (the example above) or not (string) ?

It would be good if all components return their version always in the same format independent from how other programs would like to present it.

maciaszczykm: Both are fine, but I'd use `semver` in that case. It is only an opinion though :)

With `semver` we can stay a bit more consistent (seems like kubelet version is a string in a few structs) and type is representing contents more precisely. In the future other clients might use it too (or not?). With `string` we keep it simple to only display it.

kdomanski: @p0lyn0mial What is the use case for structured version? The endpoint returns the same exact string that `git describe` does. Seems wasteful to parse it just because we can.

p0lyn0mial: @kdomanski usually representation is different than presentation (xml, structured logging). Clients (dashboard, CLI) can then display the version however they like without parsing and guessing. Worse if not all components follow the same format assuming it will be a string.

@kdomanski I don't want to block it, feel free to rebase and merge it.

PR #2760, opened 2019-02-06
Do not tolerate authentication-tolerate-lookup-failure

alvaroaleman:

Currently if the lookup of the authentication config lookup fails, the scheduler and controller-manager will just continue starting up, but treat all requests as anonymous and thus deny them. This patch makes a lookup of the authentication config a fatal error.

Fixes the sporadic `Kubernetes scheduler has disappeared from Prometheus target discovery.`

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/pkg/resources/scheduler/deployment.go
+++ b/api/pkg/resources/scheduler/deployment.go
@@ -200,6 +200,8 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
 		flags = append(flags, "--authorization-kubeconfig", "/etc/kubernetes/kubeconfig/kubeconfig")
 		// We're going to use the https endpoints for scraping the metrics starting from 1.13. Thus we can deactivate the http endpoint
 		flags = append(flags, "--port", "0")
+		// Force the authentication lookup to succeed, otherwise if it fails all requests will be treated as anonymous and thus fail
+		flags = append(flags, "--authentication-tolerate-lookup-failure", "false")
 
 	}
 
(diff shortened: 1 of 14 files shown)


alvaroaleman:
cc @xrstf

api/pkg/resources/scheduler/deployment.go
```diff
@@ -200,6 +200,8 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
 		flags = append(flags, "--authorization-kubeconfig", "/etc/kubernetes/kubeconfig/kubeconfig")
 		// We're going to use the https endpoints for scraping the metrics starting from 1.13. Thus we can deactivate the http endpoint
 		flags = append(flags, "--port", "0")
+		// Force the authentication lookup to succeed, otherwise if it fails all requests will be treated as anonymous and thus fail
+		flags = append(flags, "--authentication-tolerate-lookup-failure", "false")
```
mrIncompetent: This flag does not exist in v1.13.0.
It got introduced as a Bugfix with https://github.com/kubernetes/kubernetes/pull/71806

But we can simply omit the flag(for v1.13.0) as failing was the default behavior.

mrIncompetent (changes requested):

api/pkg/resources/scheduler/deployment.go
alvaroaleman: Wow, that is a detailled knowledge of the releases. Will fix it up.

alvaroaleman: Updated, PTAL

mrIncompetent (approved):

PR #2766, opened 2019-02-07
Control plane monitoring

xrstf:
We never actually monitored schedulers and controller-managers inside seed clusters, so our scraping rules and the alerts were more or less entirely broken. This PR attempts to rectify the situation.

The diff for this is rather large because of all the rule changes, but don't fear -- they were just moved around and not actually changed in any way.

With this PR, the following changes:

* Scheduler and controller-manager health are now deduced from looking at the kube-state-metrics metrics instead of having to scrape them individually. Scraping them securely (!) turns out to be a rather complicated beast. Plus it would require customers to potentially change their scheduler/cm deployments and fiddle with kubeadm. Relying on KSM instead makes the alerting much simpler.
* The rules have been re-organized by "group" (general, kubermatic, managed). This has been done in an effort to make it easier to disable certain pieces. "managed" in this context means scheduler+cm ("control-plane" would have been a confusing name because the apiserver is part of the general group).
  The old organization by "concern" was nice, but would not allow to get rid of alerts for schedulers specifically.
* Custom alerting rules are no longer accidentally ignored.

We decided to keep the alerts for scheduler+cm enabled by default, so customers on GKE would have to disable them via Helm.

**Release note**:
```release-note
fix Prometheus scraping/alerting rules for Kubernetes scheduler and controller-manager
```


xrstf:
@kron4eg I don't really totally love the new way the config files are configured in the `values.yaml`. If someone wants to override something, they would have to copy the entire list of files and would miss out on new files in the chart in future versions.

Another idea would be to make it a map like

```
prometheus:
  ruleFiles:
    '/etc/prometheus/rules/ark.yaml': true
    '/etc/prometheus/rules/cadvisor.yaml': true
```

WDYT?

xrstf:
cc @alvaroaleman if you're interested

kron4eg:
@xrstf sounds terrible :D glob syntax is perfect, expecially considering the fact that it can be used as a automated integration point for 3rd party alerts without changing anything in values.yaml

xrstf:
> glob syntax is perfect, expecially considering the fact that it can be used as a automated integration point for 3rd party alerts without changing anything in values.yaml

Yeah no shit, that's why I was asking for a better suggestion. ;-)

xrstf:
@kron4eg This is now pretty well tested and can be reviewed.

config/monitoring/prometheus/rules/src/compile.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env sh
+
+cd $(dirname $0)
+
+comment="# This file has been generated, do not edit."
+
+for file in */*.yaml; do
+  newfile=$(dirname $file)-$(basename $file)
+  echo "$file => $newfile"
+  yq r $file -j | jq 'del(.groups[].rules[].runbook)' | (echo "$comment"; yq r -) > ../$newfile
```
kron4eg, line 18: probably somewhere those dependencies should be documented somehow, like how to install or at least where to grab them.

config/monitoring/prometheus/templates/clusterrole.yaml
```diff
@@ -11,6 +11,7 @@ rules:
   - services
   - endpoints
   - pods
+  - pods/proxy
```
kron4eg: do we still need pods/proxy?

kron4eg (commented):
looks good, albeit had couple of comments

config/monitoring/prometheus/templates/clusterrole.yaml
xrstf: True, this can go away.

xrstf:
ping @kron4eg

PR #2771, opened 2019-02-08
Allow choosing an Openstack subnet in the UI

alvaroaleman:

Currently its impossible to choose a subnet after a network was chosen for Openstack in the "Extended Settings" page. This fixes that.

Also, it changes the `getAllNetworks` function in the openstack helper to take a `osnetworks.ListOpts` parameter to force consumers to think about the question if its possible to filter.

Fixes #

**Documentation**:

**Release note**:

```release-note
Fixed a bug that made it impossible to choose a subnet on Openstack after a network was chosen
```

/assign @mrIncompetent 

cc @kgroschoff

### DIFF
--- a/api/pkg/provider/cloud/openstack/helper.go
+++ b/api/pkg/provider/cloud/openstack/helper.go
@@ -58,9 +58,9 @@ type NetworkWithExternalExt struct {
 	osextnetwork.NetworkExternalExt
 }
 
-func getAllNetworks(netClient *gophercloud.ServiceClient) ([]NetworkWithExternalExt, error) {
+func getAllNetworks(netClient *gophercloud.ServiceClient, opts osnetworks.ListOpts) ([]NetworkWithExternalExt, error) {
 	var allNetworks []NetworkWithExternalExt
-	allPages, err := osnetworks.List(netClient, nil).AllPages()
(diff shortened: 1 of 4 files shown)


api/pkg/provider/cloud/openstack/helper.go
```diff
@@ -362,9 +362,19 @@ func getTenants(authClient *gophercloud.ProviderClient, region string) ([]osproj
...
-func getSubnetForNetwork(netClient *gophercloud.ServiceClient, networkID string) ([]ossubnets.Subnet, error) {
+func getSubnetForNetwork(netClient *gophercloud.ServiceClient, networkIDOrName string) ([]ossubnets.Subnet, error) {
 	var allSubnets []ossubnets.Subnet
 
+	networks, err := getAllNetworks(netClient, osnetworks.ListOpts{Name: networkIDOrName})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list networks: %v", err)
+	}
+
+	networkID := networkIDOrName
```
kdomanski, line 381: What is the purpose of this assignment? This value will persist only if `getAllNetworks` with the `Name` filter returns either zero results or more than 1, which IMO should be error conditions here.

PR #2776, opened 2019-02-11
ensure no clashes with top-level keys in values.yaml

xrstf:
Some of our charts use their top-top-level directory inside their values.yaml, like the logging charts. This leads to us creating a broken example values.yaml because we have the same top-level keys multiple times.

This PR extends the sync-charts.sh script to filter the YAML line by line and keep track of known top-level keys. This works under the assumption that charts with the same top-level key are added one after another with no gaps to the values.yaml.

For the next point release (i.e. 2.10) of Kubermatic we should instead fix the charts, but this would be too much for a bugfix release in the 2.9 branch.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/hack/sync-charts.sh
+++ b/api/hack/sync-charts.sh
@@ -110,6 +110,24 @@ for CHART in ${BACKUP_CHARTS}; do
   echo "" >> ${TARGET_VALUES_FILE}
 done
 
+# merge duplicate top-level keys
+KNOWN_KEYS=()
+
+while IFS= read LINE; do
+  MATCH=$(echo "${LINE}" | grep -oE "^[a-zA-Z0-9]+" || true)
+
+  if [ -z "${MATCH}" ]; then
+    echo "${LINE}"
+  else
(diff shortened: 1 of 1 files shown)


alvaroaleman:
/shrug

What do you think about changing the publishing to

a )just render a concatenated `values.examples.yaml` inside this repo and add a presubmit that ppl dont forget
b) Publish the results post-merge?


That way we can like see the generated file and spot mistakes, right now its more like "I think this will do the right thing"

xrstf:
I would much rather spend time fixing the charts than building larger hacks around them.

alvaroaleman:
> I would much rather spend time fixing the charts than building larger hacks around them.

Sure. What I meant is: We currently can only review the scripts that generate these charts but not the result of that generation. IMHO it would make sense to version the result of that generation inside this repo so we can check if it matches our expectations.

I am pretty sure if we had had that already we would have noticed the issue with the duplicate top level keys ourselves.

PR #2779, opened 2019-02-11
Restructure events responses

maciaszczykm:
**What this PR does / why we need it**: Restructures event responses for the Dashboard client. Events specific for Nodes will be put in REST subresource or directly in the structures. This change is connected only to group endpoint (grouping events from all Nodes created by Node Deployment).

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: n/a

**Documentation**:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -681,6 +681,6 @@ type Event struct {
 	// Type of this event (i.e. normal or warning). New types could be added in the future.
 	Type string `json:"type,omitempty"`
 
-	// The object name that those events are about.
+	// The object reference that those events are about.
 	InvolvedObject ObjectReference `json:"involvedObject"`
 }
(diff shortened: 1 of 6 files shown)


api/pkg/api/v1/types.go
```diff
@@ -665,20 +664,23 @@ type NodeDeploymentSpec struct {
...
 }
 
-// EventList is an events response structure.
-type EventList struct {
-	// The object name that those events are about.
-	Name string `json:"name"`
-
-	// List of events
-	Events []Event `json:"events"`
+type ObjectReference struct {
```
maciaszczykm: @p0lyn0mial I've restored this one as it felt better to have data to identify object not only by its name, but namespace and kind too. If you don't like it I can revert it to the name only, however, I would prefer it in the current form.

maciaszczykm: Having unified `Event` structure will allow us to create generic event list component in the UI, then we can display it not only in Node Deployments details but on all other pages where we will have `[]Event`.

PR #2780, opened 2019-02-11
Validate via events that LB services actually got deleted

alvaroaleman:

Ensures we properly wait until all services of type Loadbalancer are gone

Fixes #

**Documentation**:

**Release note**:

```release-note
Fixed a bug that prevented Loadbalancers from getting properly cleaned up on cluster deletion
```

/assign @mrIncompetent

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,4 +37,3 @@ issues:
   - "`updateManager` can be `github.com/kubermatic/kubermatic/api/pkg/handler.UpdateManager`"
   - "`store` can be `github.com/kubermatic/kubermatic/api/vendor/k8s.io/client-go/tools/cache.KeyGetter`"
   - 'func `\(\*ClusterProvider\)\.withImpersonation` is unused'
-  - 'cyclomatic complexity 36 of func `\(\*Controller\)\.cleanupInClusterResources`'
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,3 +37,4 @@ issues:
   - "`updateManager` can be `github.com/kubermatic/kubermatic/api/pkg/handler.UpdateManager`"
   - "`store` can be `github.com/kubermatic/kubermatic/api/vendor/k8s.io/client-go/tools/cache.KeyGetter`"
   - 'func `\(\*ClusterProvider\)\.withImpersonation` is unused'
+  - 'cyclomatic complexity 36 of func `\(\*Controller\)\.cleanupInClusterResources`'
--- a/api/pkg/controller/cluster/deleting.go
+++ b/api/pkg/controller/cluster/deleting.go
@@ -18,7 +18,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
+
 	controllerruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -85,6 +87,8 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
 		}
 
 		for _, service := range serviceList.Items {
+			// Need to change the scope so the inline func in the updateCluster call always has the service from the current iteration
+			service := service
 			if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
 				if err := client.CoreV1().Services(service.Namespace).Delete(service.Name, &metav1.DeleteOptions{}); err != nil {
 					return nil, fmt.Errorf("failed to delete Service '%s/%s' from user cluster: %v", service.Namespace, service.Name, err)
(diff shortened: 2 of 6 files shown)


alvaroaleman:
TODO:

Make sure we only go through this code path if the in-tree cloudprovider for AWS, Azure or Openstack is used

.golangci.yml
```diff
@@ -37,3 +37,4 @@ issues:
   - "`updateManager` can be `github.com/kubermatic/kubermatic/api/pkg/handler.UpdateManager`"
   - "`store` can be `github.com/kubermatic/kubermatic/api/vendor/k8s.io/client-go/tools/cache.KeyGetter`"
   - 'func `\(\*ClusterProvider\)\.withImpersonation` is unused'
+  - 'cyclomatic complexity 36 of func `\(\*Controller\)\.cleanupInClusterResources`'
```
mrIncompetent: I would like to have this resolved instead of silenced.
Maybe in a follow up?

api/pkg/controller/cluster/deleting.go
```diff
@@ -82,7 +93,19 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
 				if err := client.CoreV1().Services(service.Namespace).Delete(service.Name, &metav1.DeleteOptions{}); err != nil {
 					return nil, fmt.Errorf("failed to delete Service '%s/%s' from user cluster: %v", service.Namespace, service.Name, err)
 				}
-				deletedSomeResource = true
+				deletedLBs = append(deletedLBs, fmt.Sprintf("%s/%s", service.Namespace, service.Name))
```
mrIncompetent: We should save the individual service in the cluster here, otherwise we risk losing all data, if the last delete call fails

api/pkg/controller/cluster/deleting.go
```diff
@@ -119,6 +142,57 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
 		return c, nil
 	}
 
+	// Check for the event, as that is the only way we can safely know the Loadbalancer is actually gone
+	if val, exists := c.Annotations[deletedLBAnnotationName]; exists {
+		deletedLBsFromAnnotationSlice := strings.Split(val, ",")
+		deletedLBsFromAnnotation := sets.NewString(deletedLBsFromAnnotationSlice...)
```
mrIncompetent: Can we make this a 1 line statement and maybe rename it to something like `lbNames` ?

api/pkg/controller/cluster/deleting.go
```diff
@@ -119,6 +142,57 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
...
+			}
+
+		}
+		c, err = cc.updateCluster(c.Name, func(c *kubermaticv1.Cluster) {
+			if deletedLBsFromAnnotation.Len() > 0 {
+				c.Annotations[deletedLBAnnotationName] = strings.Join(deletedLBsFromAnnotation.List(), ",")
+			} else {
+				newAnnotations := map[string]string{}
+				for k, v := range c.Annotations {
+					if k != deletedLBAnnotationName {
```
mrIncompetent: Why not delete the single map key with:
```go
delete(c.Annotations, deletedLBAnnotationName);
```
?

mrIncompetent:
How do we handle the case when we don't get the event (Because it could not be deleted, or the event could not be created)?

Should we specify a timeout which:
- Skips the deletion?
- Emits a counter metric + we have an alert in place?

api/pkg/controller/cluster/deleting.go
alvaroaleman: Can you elaborate on what you mean by `We should save the individual service in the cluster here`?

alvaroaleman:
> How do we handle the case when we don't get the event (Because it could not be deleted, or the event could not be created)?

Maybe just use a 2h timeout since `DeletionTimestamp`?

api/pkg/controller/cluster/deleting.go
mrIncompetent: We should update the cluster object on each loop iteration.
Otherwise we'll lose all deleted service names in case the last Delete call failed

mrIncompetent:
> Maybe just use a 2h timeout since `DeletionTimestamp`?

2h timeout sounds good, with or without a metric afterwards?

api/pkg/controller/cluster/deleting.go
alvaroaleman: Quick idea: What about tracking that inside the cluster? Updating the cluster crd and updating the loop on each iteration will create a lot of load, as the update triggers a reconciliation and the reconiliation triggers an update

alvaroaleman:
@mrIncompetent PTAL, I think all comments have been addressed:

* We update the annotation after each LoadBalancer service deletion
* The code was in my biased opinion as the author drastically simplified and does not require a linter exception anymore
* We only do this whole checking if one of AWS, Azure or Openstack is used
* There is a 2h timeout on waiting for the event
* There is a metric for LoadBalancers that did not emit the event within the grace period
* Added a check for a caching issue (We have to wait for the Annotation to be in the lister after updating it) along the way

alvaroaleman:
The failures are due to #2807, we changed the chart to only have one setting for both the deployment of the VPA and the feature gate in #2805, this makes the e2e tests fail as Helm tries to create the cluster-wide VPA resources but they already exist

api/pkg/controller/cluster/deleting.go
```diff
@@ -78,11 +87,38 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
...
+					return nil, fmt.Errorf("failed to update cluster when trying to add UID of deleted LoadBalancer: %v", err)
+				}
+				// Wait for the update to appear in the lister as we use the data from the lister later on to verify if the LoadBalancers
+				// are gone
+				if err := wait.Poll(10*time.Millisecond, 5*time.Second, func() (bool, error) {
+					clusterFromLister, err := cc.clusterLister.Get(c.Name)
+					if err != nil {
+						return false, err
+					}
+					if deletedLBAnnotationValue, annotationExists := clusterFromLister.DeepCopy().Annotations[deletedLBAnnotationName]; annotationExists {
```
mrIncompetent: You could avoid doing the `annotationExists` check. If it does not exist, `deletedLBAnnotationValue` is an empty string - enough for the `Contains` check afterwards

mrIncompetent:
Added my last comment :)
When that is addressed i'm happy to lgtm this

alvaroaleman:
Fixed all the map accesses to be less explicit and also fixed a tiny bug that kind of made the clusters always stay around for 2h :)

alvaroaleman:
This PR changes nothing in `.golanci.yml`....

PR #2781, opened 2019-02-11
Quote certain helm values

kron4eg:
For those could be yaml-blocks, without quotes they don't render

```release-note
Quote potential yaml block values in helm template
```


kron4eg:
With this change now it's possible to specify yaml blocks instead of those huge oneliners like base64 datacenters.yaml. 

To obtain such "wrapped" base64 output:

```bash
base64 -b 70 < datacenters.yaml
```

kron4eg:
PTAL @alvaroaleman @kdomanski @mrIncompetent @xrstf

kdomanski:
OMG <3

/approve
/lgtm

PR #2792, opened 2019-02-12
ensure Grafana dashboards are consistent and have no static variables

xrstf:
We previously accidentally had a dashboard with "cached" nodenames and their variable being set to "never ever refresh", leading customers to think we purposefully hardcoded the node names into dashboards.

This PR extends our efforts to sanitise and cleanup dashboard JSONs.

**Release note**:
```release-note
Fixed invalid variable caching in Grafana dashboards
```


kron4eg:
OMG magic 
/lgtm 
/approve

PR #2799, opened 2019-02-13
check if a cluster is healthy before deletion

p0lyn0mial:
**What this PR does / why we need it**: prevents a cluster deletion when the cluster is not healthy and removal of additional resources (volumes, load balancers) was requested.

**Release note**:

```
NONE
```

### DIFF
--- a/api/pkg/handler/cluster.go
+++ b/api/pkg/handler/cluster.go
@@ -27,6 +27,8 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/resources/cluster"
 	"github.com/kubermatic/kubermatic/api/pkg/util/errors"
 	"github.com/kubermatic/kubermatic/api/pkg/validation"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 func createClusterEndpoint(cloudProviders map[string]provider.CloudProvider, projectProvider provider.ProjectProvider, dcs map[string]provider.DatacenterMeta) endpoint.Endpoint {
@@ -177,9 +179,14 @@ func deleteCluster(sshKeyProvider provider.SSHKeyProvider, projectProvider provi
 			}
 		}
 
+		existingCluster, err := clusterProvider.Get(userInfo, req.ClusterID, &provider.ClusterGetOptions{CheckInitStatus: true})
 		if req.DeleteVolumes || req.DeleteLoadBalancers {
-			existingCluster, err := clusterProvider.Get(userInfo, req.ClusterID, &provider.ClusterGetOptions{})
 			if err != nil {
+				if kerrors.IsServiceUnavailable(err) {
+					return nil, errors.NewWithDetails(http.StatusServiceUnavailable,
+						"the cluster must be healthy when cleanup of additional resources (e.g. volumes) was requested, try one more time",
+						err.Error())
+				}
 				return nil, common.KubernetesErrorToHTTPError(err)
 			}
 			if req.DeleteLoadBalancers {
--- a/api/pkg/handler/cluster_test.go
+++ b/api/pkg/handler/cluster_test.go
@@ -87,6 +87,44 @@ func TestDeleteClusterEndpointWithFinalizers(t *testing.T) {
 			HeaderParams:    map[string]string{},
 			ExpectedUpdates: 0,
 		},
+		{
+			Name:             "scenario 4: deletion of a cluster fails for an unhealthy cluster with finalizers",
+			Body:             ``,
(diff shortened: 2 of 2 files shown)


alvaroaleman:
@p0lyn0mial It should not block on an unhealthy cluster, it should simply not add the finalizers

p0lyn0mial:
> @p0lyn0mial It should not block on an unhealthy cluster, it should simply not add the finalizers

wouldn't be better if an operator could decide what he wants ? For example if he hasn' created any additional resources he can uncheck the checkboxes and he is good, otherwise he waits until the cluster is healthy.

alvaroaleman:
> wouldn't be better if an operator could decide what he wants ? For example if he hasn' created any additional resources he can uncheck the checkboxes and he is good, otherwise he waits until the cluster is healthy.

So you expect kubermatic users to know that they must not check the cleanup boxes if their cluster wasn't healthy before because that will make Kubermatic run into a deadlock?

I'd argue that is optimistic at best.

p0lyn0mial:
> So you expect kubermatic users to know that they must not check the cleanup boxes if their cluster wasn't healthy before because that will make Kubermatic run into a deadlock?

This should be both transparent and reliable for the users. A request to cleanup additional resources should succeed or fail but not fail silently.

kdomanski:
So, a broken cluster cannot be properly cleaned up. On purpose.
Can you please elaborate on the "why we need it" ?

p0lyn0mial:
> Can you please elaborate on the "why we need it" ?

We need this pull because some clusters especially on `dev` are stuck. There were marked for deletion but they weren't because of pending finalisers. Even at the time of writing there is one faulty cluster on dev (`rjzkzrlbhc`). 

My primary concern with "simply not adding the finalizes when a cluster is unhealthy" as suggested by Alvaro is reliability. From an end-user perspective additional resources (that cost money) will not be deleted because a cluster was (at that moment) not healthy - worse there will be no message back saying - "your request failed" - instead we will return "OK".


For more see ongoing discussion in slack: https://loodse.slack.com/archives/C7KHDK8QJ/p1550053637010400?thread_ts=1550052664.010300&cid=C7KHDK8QJ

p0lyn0mial:
also the feature was added quite long time ago and I don't recall reports on that matter, are we sure this is the root cause ?

alvaroaleman:
Yes I am sure this is the root cause, the reason we see an increase in issues now is that we configured the dashboard to add these finalizers by default.

Another angle to approach this problem could BTW be to just have the cluster controller check if there is a namespace for the user cluster and if not, just remove the finalizers. Wdyt?

p0lyn0mial:
> Yes I am sure this is the root cause, the reason we see an increase in issues now is that we configured the dashboard to add these finalizers by default.

It was enabled 10 days ago on 4.02 (https://github.com/kubermatic/secrets/commit/54372afb140ee21ccf210ef423ae592a917f11cb) and if I'm reading the alerts (on `alerting-dev`) correctly the issues started yesterday.

> Another angle to approach this problem could BTW be to just have the cluster controller check if there is a namespace for the user cluster and if not, just remove the finalizers. Wdyt?

Will it make things more reliable for the end users ? What would you do as a user of the system when you found out after some time that some resources were actually not deleted ?

alvaroaleman:
> It was enabled 10 days ago on 4.02 (kubermatic/secrets@54372af) and if I'm reading the alerts (on alerting-dev) correctly the issues started yesterday.

No, we had quite some occurences of this on cloud already, thats not just since yesterday.


> Will it make things more reliable for the end users ? What would you do as a user of the system when you found out after some time that some resources were actually not deleted ?

There are only two situations in which this could happen:

A) User manually deletes the namespace of his/her cluster in the seed and deletes the cluster CR
B) Namespace was never created because of some issue like cloud provider permissions

In case b) there is nothing to delete anyways. For case a) someone actively interfered with Kubermatic and that is a situation in which we can not guarantee anything anyways so I am fine with not having the cleanup happen there.

alvaroaleman:
@p0lyn0mial If you think the reasoning outlined above is not sufficient, lets please have a call today about the topic. I want this issue to get resolved as its creating an annoying amount of manual labour and spurious alerts.

p0lyn0mial:
There are actually more options, for example one of the components e.g. `machine-controller` might crash and mark the cluster as unhealthy because the `Status.Health. MachineControllerDeploymentName` will be set to false - https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/cluster/health.go#L13

I created an alternative implementation that doesn't add finalizes, feel free to merge it, https://github.com/kubermatic/kubermatic/pull/2806

We can set up a call next week when I'm back from my holidays.

alvaroaleman:
As discussed in #2806, closing this because we will add a check for this case in the cluster controller instead

PR #2801, opened 2019-02-13
Update Event structure

maciaszczykm:
**What this PR does / why we need it**: Updates Event structure.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: n/a

**Special notes for your reviewer**: n/a

**Documentation**:

**Release note**:

```release-note
NONE
```


api/pkg/api/v1/types.go
```diff
@@ -674,4 +676,11 @@ type Event struct {
 
 	// The object name that those events are about.
 	InvolvedObjectName string `json:"involvedObjectName"`
+
+	// The time at which the most recent occurrence of this event was recorded.
+	// swagger:strfmt date-time
```
p0lyn0mial, line 681: wow :)

api/pkg/api/v1/types.go
```diff
@@ -674,4 +676,11 @@ type Event struct {
...
+	LastTimestamp Time `json:"lastTimestamp,omitempty"`
```
p0lyn0mial, line 682: oh, wait shouldn't we use our `Time` struct ? https://github.com/kubermatic/kubermatic/blob/a5dc4f75d0757d789f712aeb19e582e1859efca2/api/pkg/api/v1/time.go

p0lyn0mial:
/hold

not sure about `Time` struct.

api/pkg/api/v1/types.go
maciaszczykm: That's the one I am using. It is in the same package :)

p0lyn0mial: I must have checked a wrong file :)

PR #2806, opened 2019-02-14
do not add finalizers when cluster is unhealthy

p0lyn0mial:
**What this PR does / why we need it**: when deletion of a cluster was requested and the cluster is unhealthy additional finalizers (if requested) for volumes and/or load balancers removal will not be added.

**Release note**:

```
NONE
```

### DIFF
--- a/api/pkg/handler/cluster.go
+++ b/api/pkg/handler/cluster.go
@@ -27,6 +27,8 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/resources/cluster"
 	"github.com/kubermatic/kubermatic/api/pkg/util/errors"
 	"github.com/kubermatic/kubermatic/api/pkg/validation"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 func createClusterEndpoint(cloudProviders map[string]provider.CloudProvider, projectProvider provider.ProjectProvider, dcs map[string]provider.DatacenterMeta) endpoint.Endpoint {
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
if we decide to merge this pull then we can close https://github.com/kubermatic/kubermatic/pull/2799

alvaroaleman:
@p0lyn0mial I am actually starting to think that it would be better to implement this in the controller instead. Any thoughts on that?

p0lyn0mial:
@alvaroaleman may I know what have you decided ? Do you want to merge one of the two pull we have prepared ?

alvaroaleman:
@p0lyn0mial I havent decided anything at all, my thinking right now is:

> @p0lyn0mial I am actually starting to think that it would be better to implement this in the controller instead. Any thoughts on that?

WDYT?

p0lyn0mial:
I assume you are referring to controller manager, if only we could display cluster events that would be awesome, otherwise the end users may not know why a cluster is pending deletion. In fact this is no different from knowing for example why a cluster wasn't fully initialised. I think that in the near future we will add an endpoint for retrieving the events and we will adjust the dashboard.

alvaroaleman:
> I assume you are referring to controller manager

Yes

> if only we could display cluster events that would be awesome, otherwise the end users may not know why a cluster is pending deletion. In fact this is no different from knowing for example why a cluster wasn't fully initialised. I think that in the near future we will add an endpoint for retrieving the events and we will adjust the dashboard.

I agree, events in the UI will be a big step forward. But it doesn't really have anything to do with the decision weather the API or the controller should take care of handling the case that is causing issues here?

Do you think the check is better placed in the API or in the controller? IMHO its the controller, otherwise we distribute the special case handling in a weird way among different binaries

p0lyn0mial:
> Do you think the check is better placed in the API or in the controller?

Yes, I agree. It would be better if the check was in the controller.
Will you add it ? Can we close the pulls ?

alvaroaleman:
> Will you add it ? Can we close the pulls ?

Yes, will do. Thanks for your effort anyways, thought!

PR #2810, opened 2019-02-15
Fix user-cluster-controller-manager binary name

nikhita:
`s/usercluster-controller-manager/user-cluster-controller-manager` :woman_facepalming: 

/kind bug
/priority critical-urgent
/assign @alvaroaleman 

**Release note**:

```release-note
NONE
```


alvaroaleman:
Once #2809 is in we will notice mistakes like this by the e2e tests failing

PR #2813, opened 2019-02-15
Delete metrics-server addon

alvaroaleman:

This PR removes the metrics server addon from the addon container to make sure we don't run into issues if a customer forgets to delete it from the `defaultAddons` list

Fixes #

**Documentation**:

**Release note**:

```release-note
Remove the metrics-server from the addons container image as its now created in the seed cluster by Kubermatic
```

/assign @mrIncompetent


mrIncompetent:
When a addon gets deleted (Deleting the addon custom resource), the addon-controller will parse the addons manifest files & issue a `kubectl delete -f manifest.yaml`.

Deleting the manifests will break the deletion.

mrIncompetent:
If our goal is to prevent that certain addon names exist in the `defaultAddons` list, we should introduce a validation in the controller startup.

alvaroaleman:
> If our goal is to prevent that certain addon names exist in the defaultAddons list, we should introduce a validation in the controller startup.

Sounds reasonable :+1: 

/close

PR #2817, opened 2019-02-15
ClusterConnectionProvider: Allow using internal apiserver address

alvaroaleman:

    This change makes us not use the external apiserver address when
    components running in the seed connect to the user cluster, saving one
    unnecessary hop

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/cmd/kubermatic-controller-manager/controllers.go
+++ b/api/cmd/kubermatic-controller-manager/controllers.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/kubermatic/kubermatic/api/pkg/cluster/client"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/addon"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/addoninstaller"
 	backupcontroller "github.com/kubermatic/kubermatic/api/pkg/controller/backup"
@@ -122,13 +121,6 @@ func createClusterController(ctrlCtx *controllerContext) (runner, error) {
 		return nil, fmt.Errorf("failed to load ImagePullSecret from %s: %v", ctrlCtx.runOptions.dockerPullConfigJSONFile, err)
 	}
 
-	var clientProvider *client.Provider
-	if ctrlCtx.runOptions.kubeconfig != "" {
-		clientProvider = client.NewExternal(ctrlCtx.kubeInformerFactory.Core().V1().Secrets().Lister())
-	} else {
-		clientProvider = client.NewInternal(ctrlCtx.kubeInformerFactory.Core().V1().Secrets().Lister())
-	}
-
 	return cluster.NewController(
 		ctrlCtx.kubeClient,
 		ctrlCtx.dynamicClient,
@@ -136,7 +128,7 @@ func createClusterController(ctrlCtx *controllerContext) (runner, error) {
 		ctrlCtx.runOptions.externalURL,
 		ctrlCtx.runOptions.dc,
 		dcs,
-		clientProvider,
+		ctrlCtx.clientProvider,
 		ctrlCtx.runOptions.overwriteRegistry,
 		ctrlCtx.runOptions.nodePortRange,
 		ctrlCtx.runOptions.nodeAccessNetwork,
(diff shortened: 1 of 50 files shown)


mrIncompetent:
The only consumer of this functionality would be the addon controller?

alvaroaleman:
> The only consumer of this functionality would be the addon controller?

No, the whole cluster controller

api/pkg/cluster/client/client.go
```diff
@@ -22,19 +23,30 @@ import (
...
-	s, err := p.secretLister.Secrets(c.Status.NamespaceName).Get(resources.AdminKubeconfigSecretName)
+	var s *corev1.Secret
+	var err error
+	if p.useExternalAddress {
+		// Load the admin kubeconfig secret, it uses the external apiserver address
+		s, err = p.secretLister.Secrets(c.Status.NamespaceName).Get(resources.AdminKubeconfigSecretName)
+	} else {
+		// Load the UserClusterControllerKubeconfigSecret, it has cluster-admin privileges
+		// and uses the internal apiserver address
+		s, err = p.secretLister.Secrets(c.Status.NamespaceName).Get(resources.UserClusterControllerKubeconfigSecretName)
```
mrIncompetent: This is not explicit enough.

Clearly separate this by using 2(4) dedicated functions: 
- `GetExternalAdminKubeconfig` (`GetExternalClient`)
- `GetInternalAdminKubeconfig` (`GetInternalClient`)

Also do not reuse the UCC Kubeconfig. Instead introduce a dedicated internal admin kubeconfig which then can be used by the UCC.

config/kubermatic/templates/kubermatic-api-dep.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.kubermatic.isMaster }}
```
mrIncompetent: Move this into a dedicated PR

api/pkg/cluster/client/client.go
alvaroaleman: > Also do not reuse the UCC Kubeconfig. Instead introduce a dedicated internal admin kubeconfig which then can be used by the UCC.

SGTM

> Clearly separate this by using 2(4) dedicated functions:

I disagree here, because that would double the whole Interface as we would also have every client getter that is part of it doubled. There is no case I can think of where someone wants to have some internal and some external clients.

What about replacing the `New` with `NewInternal` and `NewExternal` instead?

mrIncompetent: > What about replacing the `New` with `NewInternal` and `NewExternal` instead?

Great idea!

mrIncompetent:
Doesn't this break local development?

alvaroaleman:
> Doesn't this break local development?

Oh it indeed does. What about just adding a flag to the cluster-controller?

mrIncompetent:
> > Doesn't this break local development?

> Oh it indeed does. What about just adding a flag to the cluster-controller?

sgtm

alvaroaleman:
@mrIncompetent Updated, PTAL

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -93,6 +94,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.StringVar(&c.inClusterPrometheusScrapingConfigsFile, "in-cluster-prometheus-scraping-configs-file", "", "The file containing the custom scraping configs for the prometheus running in the cluster-foo namespaces.")
 	flag.StringVar(&c.monitoringScrapeAnnotationPrefix, "monitoring-scrape-annotation-prefix", "monitoring.kubermatic.io", "The prefix for monitoring annotations in the user cluster. Default: monitoring.kubermatic.io -> monitoring.kubermatic.io/port, monitoring.kubermatic.io/path")
 	flag.StringVar(&rawFeatureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for various features.")
+	flag.BoolVar(&c.runOutsideOfSeed, "run-outside-of-seed", false, "Whether the controller runs outside of the seed cluster")
```
mrIncompetent: Would it be possible to just check if `-kubeconfig` is set?
I can't imagine any scenario where someone would set `-kubeconfig=/some/kubeconfig` but `-run-outside-of-seed=false` or vice versa

alvaroaleman: smart!

alvaroaleman: Updated, PTAL

api/cmd/kubermatic-controller-manager/controllers.go
```diff
@@ -122,14 +122,21 @@ func createClusterController(ctrlCtx *controllerContext) (runner, error) {
 		return nil, fmt.Errorf("failed to load ImagePullSecret from %s: %v", ctrlCtx.runOptions.dockerPullConfigJSONFile, err)
 	}
 
+	var clientProvider *client.Provider
```
floreks: `clientProvider` creation logic is identical in 3 places in this file. Is it time to refactor it to a method?

floreks:
Basically, code LGTM. I'll quickly test it.

floreks:
Tested on my local kubermatic setup and didn't see any errors related to controller trying to go out of the cluster and back in :slightly_smiling_face:. Waiting for resolving conflicts.

alvaroaleman:
@floreks As long as you are fine with the idea and approach of the change, you can just `/approve`, that will stay even throught new pushes. Henrik will do the final `\lgtm` (which gets removed when pushing new changes)

PR #2818, opened 2019-02-16
Cloud controller

alvaroaleman:

    Manage cloud provider infrastucture with a dedicated controller
    
    This results in a speedup when creating new clusters, as we parallelize
    the creation of the control plane and the cloud provider resources

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


api/cmd/kubermatic-controller-manager/main.go
```diff
@@ -64,36 +60,29 @@ func main() {
...
+	if err := kubermaticv1.SchemeBuilder.AddToScheme(mgr.GetScheme()); err != nil {
+		glog.Fatalf("failed to add kubermatic scheme to mgr: %v", err)
 	}
 
-	dynamicCache, err := cache.New(config, cache.Options{})
-	if err != nil {
-		glog.Fatalf("failed to create dynamic informer cache: %v", err)
-	}
+	dynamicClient := mgr.GetClient()
+	dynamicCache := mgr.GetCache()
```
alvaroaleman, line 78: What I realized when putting this pull together is that we allocate listers twice, once via the shared informer factory and once via the cache from controller-runtime, this could explain the increase in memory usage sys11 noticed.

Long term we should probably just rewrite all controllers to be controller-runtime based, its some labour but not exactly difficult. Maybe something for a working student or a new employee?

api/pkg/controller/cloud/cloud_controller.go
```diff
@@ -0,0 +1,119 @@
...
+		if finalizers.Has(clustercontroller.InClusterLBCleanupFinalizer) ||
+			finalizers.Has(clustercontroller.InClusterPVCleanupFinalizer) ||
+			finalizers.Has(clustercontroller.NodeDeletionFinalizer) {
+			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+		}
+		_, err = prov.CleanUpCloudProvider(cluster, r.updateCluster)
+		return reconcile.Result{}, err
+	}
+
+	_, err = prov.InitializeCloudProvider(cluster, r.updateCluster)
```
alvaroaleman, line 107: So far the check if all cloud provider resources were successfully created was implicit in the "control plane is ready check". This is not the case anymore, but we must communicate this back, otherwise the frontend may attempt to create machines when not all infra was created yet. WDYT about adding a field in the status of the cluster crd?

api/pkg/controller/cluster/deleting.go
```diff
@@ -168,43 +151,14 @@ func (cc *Controller) cleanupInClusterResources(c *kubermaticv1.Cluster) (*kuber
 		return c, nil
 	}
 
-	c, err = cc.updateCluster(c.Name, func(c *kubermaticv1.Cluster) {
-		// In case we should keep the LB's, we must remove some cloud provider specific finalizers.
-		// Otherwise we might break the LB
-		if !shouldDeleteLBs {
-			// OpenStack
-			c.Finalizers = kuberneteshelper.RemoveFinalizer(c.Finalizers, openstack.OldNetworkCleanupFinalizer)
```
alvaroaleman, line 176: So far this has been mostly dead code, because if both finalizers were unset we never got to this point. We only got here if exactly one finalizer was set.

Now that we do have cloud provider resource management in a distinct controller, this is harmful as it will start removing finalizers once in-cluster resources are cleaned up.

I am not sure whats the best way to communicate "There was never a LB cleanup finalizer" and "There was never a PV cleanup finalizer". Maybe just ignore this case for now? As mentioned above, the existing code has been mostly defunct anyways.

alvaroaleman (commented):
/assign @mrIncompetent

api/cmd/kubermatic-controller-manager/main.go
mrIncompetent: > Long term we should probably just rewrite all controllers to be controller-runtime based

:+1: 

> Long term we should probably just rewrite all controllers to be controller-runtime based, its some labour but not exactly difficult. Maybe something for a working student or a new employee?

:+1:

api/pkg/controller/cloud/cloud_controller.go
mrIncompetent: How about using conditions?

alvaroaleman: Actually, after trying to implement conditions I start disliking the idea, because:

* Whenever we want to get the state, we have to iterate
* Whenever we want to modify the state, we have to iterate within the `modify` passed into the `updateCluster` as the index of the condition could have changed
* We can not guarantee a condition doesn't get set twice

Is there anything to be gained for that complexity?

mrIncompetent: > * Whenever we want to get the state, we have to iterate
> * Whenever we want to modify the state, we have to iterate within the `modify` passed

True, but we can simply have a helper function? The etcd operator did something like this: https://github.com/coreos/etcd-operator/blob/master/pkg/apis/etcd/v1beta2/status.go#L182

> * We can not guarantee a condition doesn't get set twice

When using a helper function like above this should not happen? (Sure there might be edge cases).

> Is there anything to be gained for that complexity?

I just find this the most intuitive approach. How would the status field look like?

alvaroaleman: > When using a helper function like above this should not happen? (Sure there might be edge cases).

OFC but you have to use these helper functions everytime, implying you have to know they exist and have to find them. That is a lot more complicated than e.G. `if cluster.Status.Health.Etcd`.

>  just find this the most intuitive approach. How would the status field look like?

Right now I would just make it a boolean like the existing health fields, so `.Status.Health.CloudProviderInfrastructureReady bool`

alvaroaleman: Any opinion on that?

api/pkg/controller/cluster/deleting.go
mrIncompetent: Create a issue so we keep track of it. This must be reworked before we cut the next release

api/pkg/controller/cloud/cloud_controller.go
mrIncompetent: For the time being i'm fine with that variable

api/pkg/controller/cluster/deleting.go
alvaroaleman: Done: https://github.com/kubermatic/kubermatic/issues/2839

PR #2819, opened 2019-02-17
Do not create VPAs when the feature is disabled

alvaroaleman:

Currently we always create the VPA CRs, even when the feature is disabled. The only thing different is that we set their mode to `off`. With this PR we don't create them in the first place, which shortens the duration of a cluster sync and simplifies the code.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


mrIncompetent:
Going this way can lead to situations where the VPA will still be used even though the feature flag has been disabled, as the existing VPA resources won't be deleted after the flag has been disabled.

The old code ensures that all VPA resources get disabled (mode: off) when the feature flag gets disabled.

PR #2820, opened 2019-02-17
Only sync charts on releases, not for each commit

alvaroaleman:

Currently we sync the charts on every commit in a release branch. This has already caused some confusion, especially as the installer repo does not contain the release tags. This PR changes that behavior to only sync when there is a tag on the commit and also changes the commit message to indicate for which release the sync was.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


alvaroaleman:
Holding because the github push event most likely does not contain the branch when triggered by a tag

PR #2822, opened 2019-02-18
ucc: reconcile more resources

nikhita:
Follow up of https://github.com/kubermatic/kubermatic/pull/2671

This PR moves reconciliation of `Role`, `RoleBinding`, `ClusterRole`, `ClusterRoleBinding` to the user cluster controller. The creators for these resources take `TemplateData` as an argument, but they are never used.

For example:

https://github.com/kubermatic/kubermatic/blob/3da7568e8aa898eb836f5182611e8dbc115b56d3/api/pkg/resources/machinecontroller/role.go#L80

This PR does not remove the `$typeDataProvider` interface as an argument from the creator though -- this is because some other functions like [`EnsureRole`](https://github.com/kubermatic/kubermatic/blob/3da7568e8aa898eb836f5182611e8dbc115b56d3/api/pkg/resources/ensure.go#L39) use this interface (to use `TemplateData`).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/cluster/userclusterresources.go
+++ b/api/pkg/controller/cluster/userclusterresources.go
@@ -8,17 +8,13 @@ import (
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
-	"github.com/kubermatic/kubermatic/api/pkg/resources/controllermanager"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/ipamcontroller"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/machinecontroller"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/metrics-server"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/openvpn"
-	"github.com/kubermatic/kubermatic/api/pkg/resources/scheduler"
-	"github.com/kubermatic/kubermatic/api/pkg/resources/vpnsidecar"
 
 	admissionv1alpha1 "k8s.io/api/admissionregistration/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -38,23 +34,7 @@ func (cc *Controller) reconcileUserClusterResources(cluster *kubermaticv1.Cluste
 		}
 	}
 
-	if err = cc.userClusterEnsureRoles(cluster, client); err != nil {
-		return nil, err
-	}
-
 	if err = cc.userClusterEnsureConfigMaps(cluster, client); err != nil {
-		return nil, err
-	}
-
-	if err = cc.userClusterEnsureRoleBindings(cluster, client); err != nil {
-		return nil, err
-	}
-
-	if err = cc.userClusterEnsureClusterRoles(cluster, client); err != nil {
-		return nil, err
-	}
-
-	if err = cc.userClusterEnsureClusterRoleBindings(cluster, client); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


api/pkg/controller/cluster/userclusterresources.go
```diff
@@ -38,26 +34,10 @@ func (cc *Controller) reconcileUserClusterResources(cluster *kubermaticv1.Cluste
...
 
-	if err = cc.userClusterEnsureRoles(cluster, client); err != nil {
-		return nil, err
-	}
-
 	if err = cc.userClusterEnsureConfigMaps(cluster, client); err != nil {
 		return nil, err
 	}
 
-	if err = cc.userClusterEnsureRoleBindings(cluster, client); err != nil {
```
nikhita, line 49: @alvaroaleman IIRC this needs to stay to be able to reconcile a single RoleBinding? :thinking:

alvaroaleman: Does it? Which one?

api/pkg/controller/usercluster/controller.go
```diff
@@ -35,6 +36,18 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err = c.Watch(&source.Kind{Type: &apiregistrationv1beta1.APIService{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return err
 	}
+	if err = c.Watch(&source.Kind{Type: &rbacv1.Role{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
alvaroaleman: Instead of running a fulll reconciliation every time a any of the objects we watch changes, can we use a `EnqueueRequestsFromMapFunc` to enq a string const? Basically by always returning a `NamespacedName` where `Ǹamespace` is empty and `Name` is the string const.

A sample on how `EnqueueRequestsFromMapFunc` can be used is here: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/handler#ex-EnqueueRequestsFromMapFunc

api/pkg/controller/cluster/userclusterresources.go
nikhita: I vaguely remembered, maybe I'm confusing this with something else. Never mind. :see_no_evil:

alvaroaleman: I _think_ it does not. Initially I thought we would have to create one clusterrolebinding so this controller has the required permissions, but we are using the default `cluster-admin` clusterrolebinding which is created by the API server itself. Maybe thats it what you are remembering?

nikhita:
Updated, but I'm not entirely sure if this is correct. :thinking: 

PTAL.

api/pkg/controller/usercluster/controller.go
```diff
@@ -32,7 +31,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	if err = c.Watch(&source.Kind{Type: &apiregistrationv1beta1.APIService{}}, &handler.EnqueueRequestForObject{}); err != nil {
+	if err = c.Watch(&source.Kind{Type: &apiregistrationv1beta1.APIService{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: apiServiceMapper}); err != nil {
+		return err
+	}
+	if err = c.Watch(&source.Kind{Type: &rbacv1.Role{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: roleMapper}); err != nil {
```
alvaroaleman: What I meant was: have a `ToRequests` func that is identicall for _all_ objects we watch and looks like this, having a hardcoded key that never changes:

```
return []reconcile.Request{
￼ 			{NamespacedName: types.NamespacedName{
￼ 				Name:      "reoncicle-user-cluster-resources",
￼ 				Namespace: "",
￼ 			}}
```

We only have one reconciliation loop and we have no "controlling object" that we can enq. Hence our reconciliation loop covers _all_ objects. To not have annoying issues we must:

* Ensure only one reconciliation is running at a time
* Ensure that independently of how many objects change, only one item gets put into the workq

Otherwise if we e.G. create 7 rolebindings, we end up with having 7 new keys added to the workq (as every modification triggers another reconiliation) that each get worked on concurrently and while the first reconciliation is still running. That will just create issues and not gain us anything.

alvaroaleman:
> This PR does not remove the `$typeDataProvider` interface as an argument from the creator though -- this is because some other functions like [`EnsureRole`](https://github.com/kubermatic/kubermatic/blob/3da7568e8aa898eb836f5182611e8dbc115b56d3/api/pkg/resources/ensure.go#L39) use this interface (to use `TemplateData`).


Can we:

* Never pass nil into functions, this is very prone to create issues at some later point because no one expects that?
* Find out what data exactly is required and create a `userClusterControllerData` struct that has only fields for that data and that gets passed into all of the `creator` funcs
* Pass that data via CLI argument into the user-cluster-controller

nikhita:
closing in favor of https://github.com/kubermatic/kubermatic/pull/3086

PR #2837, opened 2019-02-20
add promtail and loki to the logging stack

xrstf:
Elasticsearch is unreliable on preemptible nodes. This PR adds Loki+Promtail as an alternative logging stack.

This requires Grafana 6.0, which is not yet released. That's why there is a new `grafana-master` deployment.

Right now promtail does not work at all, but this PR can serve as a reference for later when the whole stack is not pre-alpha anymore.

**Release note**:
```release-note
add promtail and Loki as part of the logging stack
```

### DIFF
--- a/config/logging/loki/Chart.yaml
+++ b/config/logging/loki/Chart.yaml
@@ -1,0 +1,3 @@
+name: loki
+version: 0.1.0
+description: Logging Server
--- a/config/logging/loki/templates/configmap.yaml
+++ b/config/logging/loki/templates/configmap.yaml
@@ -1,0 +1,39 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: loki
+data:
+  loki.yaml: |
+    auth_enabled: {{ .Values.logging.loki.config.auth_enabled }}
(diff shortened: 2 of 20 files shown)


mrIncompetent:
We now maintain 2 logging stacks because Elasticsearch does not run reliably on preemtible nodes?

alvaroaleman:
Lets please make this optional and disabled by default. I am fine with using it to help ourselves and even showing it to customers if they opt-in but it should not be enabled by default

xrstf:
> We now maintain 2 logging stacks because Elasticsearch does not run reliably on preemtible nodes?

No, we experimented with Loki and didn't want to throw the work away, in the hopes that one day we can replace ELK with PLG (Promtail, Loki, Grafana). That's why this branch is held and blocked.

We will not maintain two production-ready loggings stacks. Even having Loki as an optional alternative is something I am not sure about. But let's cross that bridge when Loki becomes usable.

mrIncompetent:
How about moving this into the infra repo?

xrstf:
Parking at the infra repo instead.

PR #2838, opened 2019-02-20
Add endpoint to upgrade all node deployments within the cluster

maciaszczykm:

Fixes #

**Special notes for your reviewer**: Adds endpoint to upgrade all node deployments within the cluster.

**Documentation**: n/a

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/handler/upgrade.go
+++ b/api/pkg/handler/upgrade.go
@@ -100,7 +100,7 @@ type UpgradeClusterNodeDeploymentsReq struct {
 	common.GetClusterReq
 
 	// in: body
-	Body []byte
+	Body apiv1.MasterVersion
 }
 
 func DecodeUpgradeClusterNodeDeploymentsReq(c context.Context, r *http.Request) (interface{}, error) {
@@ -149,10 +149,9 @@ func upgradeClusterNodeDeployments(projectProvider provider.ProjectProvider) end
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
-		kubeletVersion := string(req.Body)
 		var updateErrors []error
 		for _, machineDeployment := range machineDeployments.Items {
-			machineDeployment.Spec.Template.Spec.Versions.Kubelet = kubeletVersion
+			machineDeployment.Spec.Template.Spec.Versions.Kubelet = req.Body.Version.String()
 			_, err = machineClient.ClusterV1alpha1().MachineDeployments(metav1.NamespaceSystem).Update(&machineDeployment)
 			if err != nil {
 				updateErrors = append(updateErrors, err)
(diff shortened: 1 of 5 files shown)


api/pkg/handler/upgrade.go
```diff
@@ -91,6 +94,78 @@ func getClusterNodeUpgrades(updateManager UpdateManager, projectProvider provide
...
+
+		machineDeployments, err := machineClient.ClusterV1alpha1().MachineDeployments(metav1.NamespaceSystem).List(metav1.ListOptions{})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		var updateErrors []error
+		for _, machineDeployment := range machineDeployments.Items {
+			machineDeployment.Spec.Template.Spec.Versions.Kubelet = req.Body.Version.String()
+			_, err = machineClient.ClusterV1alpha1().MachineDeployments(metav1.NamespaceSystem).Update(&machineDeployment)
```
p0lyn0mial, line 165: before updating we must check if the kubelet version is compatible with the cluster version (use `ensureVersionCompatible` method ?) Would be nice if we could cover this scenario with a unit test.

api/pkg/handler/upgrade.go
```diff
@@ -91,6 +94,78 @@ func getClusterNodeUpgrades(updateManager UpdateManager, projectProvider provide
...
+			if err != nil {
+				updateErrors = append(updateErrors, err)
+			}
+		}
+
+		if len(updateErrors) > 0 {
+			return nil, fmt.Errorf("failed to update some node deployments: %+q", updateErrors)
```
p0lyn0mial: how does the client parse the response ? wouldn't be better to return human readable message along with a list of additional errors ? e.g. `{"error": "failed to update some node deployments", "details": ["first error", "second error"]"`

p0lyn0mial (changes requested):

api/pkg/handler/upgrade.go
maciaszczykm: It is parsed like every other error message. Sample notification will show `Error 500: failed to update some node deployments: ["err1", "err2"]`.

api/pkg/handler/upgrade.go
maciaszczykm: Yes, sure it should be done. I will do it.

api/pkg/handler/upgrade.go
maciaszczykm: Done.

api/pkg/handler/upgrade.go
maciaszczykm: Done.

maciaszczykm:
@p0lyn0mial PTAL

PR #2854, opened 2019-02-22
Openshift controller

alvaroaleman:

Fixes #

**Documentation**:

**Release note**:

```release-note
```


api/pkg/resources/ensure.go
```diff
@@ -332,8 +333,56 @@ func EnsureObject(namespace string, rawcreate ObjectCreator, store informerStore
 	return nil
 }
 
+// EnsureNamedObjectV2 will generate the Object with the passed create function & create or update it in Kubernetes if necessary.
+// Different to EnsureObject, EnsureNamedObjectV2 requires the name of the resource being passed so the generation just for the name gets avoided.
+// It differs from EnsureNamedObject mainly in its signature and is intended to be used from a controller-runtime based controller
+func EnsureNamedObjectV2(ctx context.Context, namespacedName types.NamespacedName, rawcreate ObjectCreator, client ctrlruntimeclient.Client, emptyObject runtime.Object) error {
```
alvaroaleman, line 339: This currently has no tests, I will add a test based on https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/envtest in the following days, as I don't see too much value in testing this with a fake client, as one of the main goals of this is to ensure idempotency

api/pkg/controller/openshift/resources/controlplane_configmap.go
```diff
@@ -0,0 +1,292 @@
...
+- (?i)//127\.0\.0\.1(:|\z)
+- (?i)//localhost(:|\z)
+- (?i)//116\.203\.105\.73(:|\z)
+- (?i)//kubernetes\.default(:|\z)
+- (?i)//kubernetes\.default\.svc\.cluster\.local(:|\z)
+- (?i)//kubernetes(:|\z)
+- (?i)//openshift\.default(:|\z)
+- (?i)//openshift\.default\.svc(:|\z)
+- (?i)//172\.30\.0\.1(:|\z)
+- (?i)//alvaro\-openshift\-controller(:|\z)
```
mrIncompetent: Lets remove this one :)

api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -0,0 +1,200 @@
...
+			nn(osData.Cluster().Status.NamespaceName, secretName), resources.SecretObjectWrapper(secretCreator), r.Client, &corev1.Secret{}); err != nil {
+			return fmt.Errorf("failed to ensure Secret %s: %v", secretName, err)
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) getAllConfigmapCreators() []openshiftresources.NamedConfigMapCreator {
+	return []openshiftresources.NamedConfigMapCreator{openshiftresources.OpenshiftControlPlaneConfigMapCreator}
```
mrIncompetent, line 168: This is the only place where you actually use the interface `openshiftresources.NamedConfigMapCreator` - so why not just define it here?

All other places just implicitly implement it

api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -0,0 +1,200 @@
...
+		if err := resources.EnsureNamedObjectV2(ctx,
+			nn(osData.Cluster().Status.NamespaceName, configMapName), resources.ConfigMapObjectWrapper(configMapCreator), r.Client, &corev1.ConfigMap{}); err != nil {
+			return fmt.Errorf("failed to ensure ConfigMap %s: %v", configMapName, err)
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) getAllDeploymentCreators() []openshiftresources.NamedDeploymentCreator {
+	return []openshiftresources.NamedDeploymentCreator{openshiftresources.APIDeploymentCreator}
```
mrIncompetent, line 183: Same as https://github.com/kubermatic/kubermatic/pull/2854/files#r260243814

api/pkg/controller/openshift/openshift_controller.go
mrIncompetent: Though no blocker

api/pkg/controller/cluster/resources.go
```diff
@@ -161,14 +161,17 @@ func GetDeploymentCreators(data resources.DeploymentDataProvider) []resources.De
...
 		dns.DeploymentCreator(data),
 		metricsserver.DeploymentCreator(data),
 		usercluster.DeploymentCreator(data),
 	}
 
+	if cluster := data.Cluster(); cluster != nil && cluster.Annotations["kubermatic.io/openshift"] == "" {
+		creators = append(creators, apiserver.DeploymentCreator(data))
+		creators = append(creators, scheduler.DeploymentCreator(data))
+		creators = append(creators, controllermanager.DeploymentCreator(data))
+	}
```
mrIncompetent, line 173: That is something we should remove at some point(Not now)...
Having the `openshift-controller` to depend on resources coming from the `cluster-controller` is bad as we then have the worst of both worlds:
- "Separate" controllers with some degree of duplication
- Some "if openshift then ... else" inside one controller to avoid more duplication

alvaroaleman: Yeah I completely agree, whats here is definitely not a long term solution, we should have two separate controller and only one of them is active, depending on if its openshift or kubernetes. I just want to get to the point where we can spawn nodes and schedule workloads on them, once we have that I'll copy everything over

api/pkg/controller/openshift/openshift_controller.go
alvaroaleman: Then I have to export `openshiftresources.openshiftData` because `openshiftcontroller.openshiftdata` will not be accepted because of the function interface matching. That is IMHO worse.

alvaroaleman:
@mrIncompetent PTAL

api/pkg/controller/openshift/openshift_controller.go
mrIncompetent: sgtm

PR #2863, opened 2019-02-25
add e2e tests for user cluster RBAC controller

### LINKED ISSUE #2802: Add an E2E test
TBD

zreigz:
**What this PR does / why we need it**: adds e2e tests for user cluster RBAC controller
Fixes #2802

```release-note
NONE
```

### DIFF
--- a/api/cmd/conformance-tests/usercluster_controller_tests.go
+++ b/api/cmd/conformance-tests/usercluster_controller_tests.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac-user-cluster"
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func (r *testRunner) testUserclusterControllerRBAC(log *logrus.Entry, cluster *kubermaticv1.Cluster, kubeClient, seedKubeClient kubernetes.Interface) error {
+	log.Info("Testing user cluster RBAC controller")
+
+	clusterNamespace := fmt.Sprintf("cluster-%s", cluster.Name)
+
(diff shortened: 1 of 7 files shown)


alvaroaleman:
/lgtm
/approve
/hold

So @p0lyn0mial can have a look pre-merge.

Thanks a lot for this! First contribution to the e2e tests from the UI/API team :)

zreigz:
@alvaroaleman thanks for the review

api/cmd/conformance-tests/usercluster_controller_tests.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// check user cluster resources: ClusterRoles and ClusterRoleBindings
+	for _, resourceName := range rbacResourceNames() {
+		log.Info("Getting a Cluster Role: ", resourceName)
+		clusterRole, err := kubeClient.RbacV1().ClusterRoles().Get(resourceName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get Cluster Role: %s, error: %v", clusterRole, err)
+		}
+
+		defaultClusterRole, err := rbacusercluster.GenerateRBACClusterRole(resourceName)
```
p0lyn0mial, line 39: I would prefer not to use "production" code from the tests. Simply because is not possible to detect changes. For example a developer changes `GenerateRBACClusterRole` incorrectly and the tests are happy because they use exactly the same function to determine the correctness. WDYT ?

zreigz: Your scenario should be cover by unit tests. E2e tests check rather a deployment and existence in the system, not the logic.

p0lyn0mial: I disagree, those test should not make any assumption about what other tests check or don't. They should be independent. 

Approving as I don't want to block this pull.

p0lyn0mial: Feel free to merge it :)

alvaroaleman: The main difference between e2e/integration tests and unit tests is that the former check the resulting behavior, whereas unit tests verify implementation details on various layers. That is mainly useful to detect if the code adheres to a given specification but not of that specification makes any sense. It helps a lot in detecting changes but again, doesn't tell if the result works. E2E/integration tests can tell you whether the result works or not, but if it doesn't work, they will probably not be very helpfull in telling you why.

For this example at hand that would mean to check if a user with a given group can/can not access specific resources.

PR #2876, opened 2019-02-27
migrate serviceaccount handling to the dynamic client

mrIncompetent:
Migrate ServiceAccount handling to the dynamic client

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/monitoring/resources.go
+++ b/api/pkg/controller/monitoring/resources.go
@@ -40,20 +40,6 @@ func (c *Controller) getClusterTemplateData(cluster *kubermaticv1.Cluster) (*res
 	), nil
 }
 
-func (c *Controller) ensureServiceAccounts(cluster *kubermaticv1.Cluster, data *resources.TemplateData) error {
-	creators := []resources.ServiceAccountCreator{
-		prometheus.ServiceAccount,
-	}
-
-	for _, create := range creators {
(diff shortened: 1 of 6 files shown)


alvaroaleman:
E2E tests will succeed even if this doens't work, right?

api/pkg/controller/monitoring/resources.go
```diff
@@ -176,3 +162,16 @@ func (c *Controller) ensureServices(cluster *kubermaticv1.Cluster, data *resourc
 
 	return resources.ReconcileServices(creators, cluster.Status.NamespaceName, c.dynamicClient, c.dynamicCache, resources.ClusterRefWrapper(cluster))
 }
+
+// GetServiceCreators returns all service creators that are currently in use
+func GetServiceAccountCreators() []resources.ServiceAccountCreator {
```
alvaroaleman, line 167: I personally would prefer it if we used `NamedRESOURCECreatorGetter` for everything btw, whats your take on that?

mrIncompetent:
> E2E tests will succeed even if this doens't work, right?

Yes, we don't have any test which validates Prometheus

api/pkg/controller/monitoring/resources.go
mrIncompetent: Yes, i agree, but i would like to issue a dedicated PR to switch all resources to the `NamedRESOURCECreatorGetter`.
I first would like to remove the need for all custom `Ensure***` functions to not mix different refactoring steps

alvaroaleman:
Lets merge this before #2882 as this makes the latter less ugly

PR #2881, opened 2019-02-28
extend health endpoint by CloudProviderInfrastructure

kgroschoff:
With adding `CloudProviderInfrastructure` to health endpoint, we could check in ui, if cluster is ready so we could start node deployment creation. 

As this is part of the [AllHealthy()](https://github.com/kubermatic/kubermatic/blob/master/api/pkg/crd/kubermatic/v1/cluster.go#L347) method which is checked [here](https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/cluster.go#L160), we will get error `Cluster components are not ready yet`, if we have no possibility to also check it in ui.

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -3215,6 +3215,10 @@
           "type": "boolean",
           "x-go-name": "Apiserver"
         },
+        "cloudProviderInfrastructure": {
+          "type": "boolean",
+          "x-go-name": "CloudProviderInfrastructure"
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
/lgtm
/approve

thanks!

p0lyn0mial:
I have seen `Finished with error: failed to setup cluster: failed waiting for control plane to become ready:` error many times, usually re-running the tests helps. 

/retest

PR #2893, opened 2019-03-01
Add doc on resource reconciling

mrIncompetent:
**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/resource-handling.md
+++ b/docs/resource-handling.md
@@ -100,6 +100,11 @@ It offers:
 func ReconcileSecrets(namedGetters []NamedSecretCreatorGetter, namespace string, client ctrlruntimeclient.Client, informerFactory ctrlruntimecache.Cache, objectModifiers ...ObjectModifier) error 
 ```
 
+The signature of all `Reconcile*` functions will change in the future to be more convenient for controller-runtime controllers:
+```go
+func ReconcileSecrets(ctx context.Context, namedGetters []NamedSecretCreatorGetter, client ctrlruntimeclient.Client, objectModifiers ...ObjectModifier) error 
+``` 
+
 ### SecretCreator
 
 A typed creator function. Prefer the [NamedSecretCreatorGetter](#namedsecretcreatorgetter) instead. 
--- a/docs/resource-handling.md
+++ b/docs/resource-handling.md
@@ -1,0 +1,123 @@
+# Resource handling in the Kubermatic codebase
+
+One of Kubermatic's main task's is to reconcile Kubernetes resources.
+The following will describe how we structure the code in Kubermatic to achieve a unified handling of various Kubernetes resources from different controllers.
+
+## Reconciling
+
+Our reconciling is designed in a way, that every Object will be overwritten on each sync.
+
+## EnsureNamedObject
+
+EnsureNamedObject is a generic "reconcile" function which will update the existing object or create it.
+If the existing object does not differ from the "wanted" object, EnsureNamedObject will issue any API call. 
+```go
+func EnsureNamedObject(name string, namespace string, rawcreate ObjectCreator, store informerStore, client ctrlruntimeclient.Client) error
+```
+
+## ObjectCreator
+
+ObjectCreator defines an interface to create/update a runtime.Object.
+It uses `runtime.Object` as that is the minimal interface which gets satisfied by all Kubernetes objects.
+That way we can have a single function to reconcile all Objects within a single function.
+`existing` will be the already existing object coming from the cache. If no object is found in the cache `existing` is nil.
+```go
+type ObjectCreator = func(existing runtime.Object) (runtime.Object, error)
+```
+
+### Example implementation
+
+The following will demonstrate a potential `ObjectCreator` implementation.
+```go
+func MyWonderfulConfigMap(existing runtime.Object) (runtime.Object, error) {
+	var cm *corev1.ConfigMap
+	if existing == nil {
+		// In case there is no already existing object to update
+		cm = &corev1.ConfigMap{}
+	} else {
+		// We case for better usability here
+		cm = existing.(*corev1.ConfigMap)
+	}
+	
+	cm.Name = "wonderful-config"
+	cm.Data = map[string]string{
+		"config.yaml": "some-config",
+	}
+
+	return cm, nil
+}
+```
+
+When having to maintain multiple resources the first 7 lines start to become an error prone & annoying boilerplate code.
+For this we created the typed `Reconcile*` functions.  
+
+## ReconcileSecrets aka "Reduce the type casting" 
+
+To avoid certain boilerplate code we introduced typed `Reconcile*` functions.
+Those functions are being generated using `go generate`: https://github.com/kubermatic/kubermatic/tree/master/api/codegen/reconcile
+The result is written into: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/zz_generated_reconcile.go
+
+`ReconcileSecrets` is a typed convenience function around `EnsureNamedObject`.
+It offers:
+- Typed creator functions `SecretCreator` (`func(existing *corev1.Secret) (*corev1.Secret, error)`)
+- Automatic nil checks & struct initialization
+- Informer allocation from a passed in `InformerFactory`
+- Unified modifier functions to allow to apply certain modifications to all passed in `SecretCreator` functions 
+- Sets resource name based the name coming from the `NamedSecretCreatorGetter` 
+
+```go
+func ReconcileSecrets(namedGetters []NamedSecretCreatorGetter, namespace string, client ctrlruntimeclient.Client, informerFactory ctrlruntimecache.Cache, objectModifiers ...ObjectModifier) error 
+```
+
+### NamedSecretCreatorGetter
+
+The `NamedSecretCreatorGetter` is a simple function definition to combine the name of the resource + the creator function.
+This avoids the need to call the creator function twice (1st time to get the objects name + second time to get the actual creator)
+
+```go
+type NamedSecretCreatorGetter = func() (name string, create SecretCreator)
+```
+
+### SecretCreator
+
+A typed creator function.
+```go
+type SecretCreator = func(existing *corev1.Secret) (*corev1.Secret, error)
+```
+
+### Example SecretCreator implementation
+
+```go
+func MyWonderfulSecret(existing corev1.Secret) (corev1.Secret, error) {
+	existing.Name = "wonderful-secret"
+	existing.Data = map[string][]byte{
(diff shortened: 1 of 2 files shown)


mrIncompetent:
Once we migrated completely to the controller-runtime, this doc must be updated to reflect the removal of the informer.

docs/resource-handling.md
```diff
@@ -0,0 +1,123 @@
+# Resource handling in the Kubermatic codebase
+
+One of Kubermatic's main task's is to reconcile Kubernetes resources.
+The following will describe how we structure the code in Kubermatic to achieve a unified handling of various Kubernetes resources from different controllers.
+
+## Reconciling
+
+Our reconciling is designed in a way, that every Object will be overwritten on each sync.
```
alvaroaleman: This already assumes familarity with some concepts, maybe something like `that we got notified whenever an object change, compare it to a desired state and if there is a diff update it`?

docs/resource-handling.md
```diff
@@ -0,0 +1,123 @@
...
+
+## EnsureNamedObject
+
+EnsureNamedObject is a generic "reconcile" function which will update the existing object or create it.
```
alvaroaleman: Plase use `EnsureNamedObject` to make sure its obvious this is code. Also I would link it to the code

docs/resource-handling.md
```diff
@@ -0,0 +1,123 @@
...
+If the existing object does not differ from the "wanted" object, EnsureNamedObject will issue any API call. 
+```go
+func EnsureNamedObject(name string, namespace string, rawcreate ObjectCreator, store informerStore, client ctrlruntimeclient.Client) error
```
alvaroaleman: I personally wouldn't nention `EnsureNamedObject` at all. It has a really weird signature, taking the store as argument. The only reason we need is is that the `cluster controller` uses some parts of `cluster-runtime` (dynamic cache, dynamic store), but doesn't use others (cache-backed dynamic client).

Can we instead just mention `EnsureNamedObjectV2` as that is what ppl should use?

docs/resource-handling.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+## ObjectCreator
+
+ObjectCreator defines an interface to create/update a runtime.Object.
```
alvaroaleman: `is a function type...`

alvaroaleman (commented):
Some comments. Also what about just going through this with @nikhita and check how much she understand? We both are pretty deep into the topic so none of us has an unbiased view.

Also IMHO it would probably be helpful to do this more on a use-case basis, as in:

* If you want to add a new object, create a `rawCreator` for your type
* If you want to alter an existing resources, wrap the `rawCreator`

docs/resource-handling.md
mrIncompetent: I'm not sure what you mean :thinking:

docs/resource-handling.md
mrIncompetent: https://github.com/kubermatic/kubermatic/pull/2893#issuecomment-468622725

docs/resource-handling.md
alvaroaleman: replace

```
EnsureNamedObject is a generic "reconcile"
```

with

```
`EnsureNamedObject` is a generic "reconcile"
```

so it gets highlighted as code

docs/resource-handling.md
alvaroaleman: > Once we migrated completely to the controller-runtime, this doc must be updated to reflect the removal of the informer.

I disagree. The ppl that need this doc will not work on the relevant parts in the cluster controller. And the cluster-controller is the only place where we are using `EnsureNamedObject`. Everything else should not use it.

docs/resource-handling.md
```diff
@@ -0,0 +1,226 @@
...
+The following will describe how we structure the code in Kubermatic to achieve a unified handling of various Kubernetes resources from different controllers.
+
+## Reconciling
+
+Our reconciling is designed in a way, that we got notified whenever an object change, compare it to a desired state and if there is a diff update it.
+
+## EnsureNamedObject
+
+`EnsureNamedObject` is a generic "reconcile" function which will update the existing object or create it.
+If the existing object does not differ from the "wanted" object, `EnsureNamedObject` will issue any API call.
```
nikhita: nit: not issue any API call?

docs/resource-handling.md
```diff
@@ -0,0 +1,226 @@
...
+
+```go
+func MyWonderfulSecret(existing corev1.Secret) (corev1.Secret, error) {
+	existing.Name = "wonderful-secret"
+	existing.Data = map[string][]byte{
+		"user":     []byte("foo"),
+		"password": []byte("bar"),
+	}
+
+	return existing, nil
```
nikhita: For this one, can we do something like this to ensure proper handling of `nil`?

```go
secret := existing
if secret == nil {
   secret = &corev1.Secret{}
}

secret.Name = "wonderful-secret"
...
```

docs/resource-handling.md
```diff
@@ -0,0 +1,226 @@
...
+```go
+func MyWonderfulMyNewTypeCreator(data dataProvider) MyNewTypeCreator {
+	return func(existing *myapiv1.MyNewType) (*myapiv1.MyNewType, error) {
+		existing.Foo = "bar"
+		return existing, nil
+	}
+}
+```
+
+### Reconcile your new resource
```
nikhita: Can we just add a line that this reconciliation needs to go in the controller?

(because writing the creator func above isn't part of the controller and it just confused me for a second :see_no_evil: )

docs/resource-handling.md
```diff
@@ -0,0 +1,226 @@
...
+if err := ReconcileMyNewTypes(creators, "some-namespace", client, informerFactory); err != nil {
+	return fmt.Errorf("failed to reconcile MyNewTypes: %v", err)
+}
+```
+
+## Wrap/Modify existing objects
+
+For wrapping/modifying existing resources we have 2 options:
+- Pass `ObjectModifier` functions to the typed `Reconcile*` functions
+  Good if you wan't to modify all resources of a specific type
```
nikhita: nit: s/wan't/want

nikhita:
> Once we migrated completely to the controller-runtime, this doc must be updated to reflect the removal of the informer.

@mrIncompetent IIUC this doc is not meant to reflect reconciliation using controller-runtime...or is it? If yes, I think it would be useful to write down how to use the generated functions by controller-runtime (in brief).

nikhita:
Also.... can we also talk a bit about how funcs like `userClusterEnsureConfigMaps` funcs work? I think the gist is somewhat covered in the doc but could we also say something in more explicit terms (in much nicer words of course :)):

- it will use `data` for the creators and create an object (say `cm`).
- it will then use the client to `GET` an object (of the same type) with that name (say `cm.Name`)
- if it doesn't find such an object, it will create it using the client
- we then deepcopy the object to make sure we don't mutate it
- compare with the existing object created in the first point -> if it is equal, don't do anything. if not, update the object.

When I first saw this code, it took me a while to understand why we were using `Creators`, etc and @alvaroaleman walked me through these funcs which made the _why_ part much clearer :grimacing:

docs/resource-handling.md
```diff
@@ -0,0 +1,226 @@
+# Resource handling in the Kubermatic codebase
+
+One of Kubermatic's main task's is to reconcile Kubernetes resources.
+The following will describe how we structure the code in Kubermatic to achieve a unified handling of various Kubernetes resources from different controllers.
+
+## Reconciling
+
+Our reconciling is designed in a way, that we got notified whenever an object change, compare it to a desired state and if there is a diff update it.
```
nikhita, line 25: re desired state...can we say that we get the desired state using the creators?

mrIncompetent:
> Also.... can we also talk a bit about how funcs like userClusterEnsureConfigMaps funcs work? I think the gist is somewhat covered in the doc but could we also say something in more explicit terms (in much nicer words of course :)):

`userClusterEnsureConfigMaps` should be replaced by `ReconcileConfigMaps`.
There shall be no custom reconciling implementation. 
We only introduced `userClusterEnsureConfigMaps` because we had no way of writing a generic reconciling function which covers plain HTTP + Cache usage - As all code which reconciled user cluster resources had no cache.

mrIncompetent:
> @mrIncompetent IIUC this doc is not meant to reflect reconciliation using controller-runtime...or is it? If yes, I think it would be useful to write down how to use the generated functions by controller-runtime (in brief).

The above mentioned functions can be used by controllers using the controller-runtime.
Only caveat is that each reconciler needs to have the "informerFactory":
```go
informerFactory := mgr.GetCache()
```
As its a required parameter for the `Reconcile*` functions - but that should drop as soon as i migrated the cluster+monitoring controller (which i started to do)

docs/resource-handling.md
mrIncompetent: done with https://github.com/kubermatic/kubermatic/pull/2893/commits/cbaaf2882bb5a6445959916e02c4bb2a0faa8145

docs/resource-handling.md
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/2893/commits/cbaaf2882bb5a6445959916e02c4bb2a0faa8145

docs/resource-handling.md
mrIncompetent: I added a comment with https://github.com/kubermatic/kubermatic/pull/2893/commits/cbaaf2882bb5a6445959916e02c4bb2a0faa8145

docs/resource-handling.md
mrIncompetent: Fixed with https://github.com/kubermatic/kubermatic/pull/2893/commits/cbaaf2882bb5a6445959916e02c4bb2a0faa8145

mrIncompetent:
@nikhita @alvaroaleman 
I updated the doc to include:
- A TOC
- Example code which covers the integration in legacy controllers + controller-runtime controllers
- A info on a potential change of the function signature of `EnsureNamedObject` + the typed `Reconcile*` functions.
- Added a info which tells to prefer the typed `Reconcile*` functions over the `EnsureNamedObject`

mrIncompetent:
Merging this one as there has been no comment in the last 8 days and having it in is more useful than an open PR...
/approve
/lgtm

nikhita:
/retest

:/

mrIncompetent:
oh come on
/retest

nikhita:
Worst case, if it flakes again, _mayyybeee_ you can do a `/override <test-name>` since this PR just has docs changes?

But I don't want this to set a precedent for using that command too. :see_no_evil:

mrIncompetent:
@nikhita oh you shouldn't have told me that :smiling_imp:

PR #2897, opened 2019-03-01
move tests to _test package

### LINKED ISSUE #2359: Use the “underscore test” for the handler package

zreigz:
**What this PR does / why we need it**:  this is the last step to moving unit tests from handler package to handler_test package

Fixes #2359

```release-note
NONE
```


api/pkg/handler/routing_test.go
```diff
@@ -1,116 +1,29 @@
...
 	testUserID    = test.UserID
 	testUserName  = test.UserName
 	testUserEmail = test.UserEmail
 )
 
-func getUser(email, id, name string, admin bool) apiv1.User {
-	return test.GetUser(email, id, name, admin)
-}
-
 func TestUpRoute(t *testing.T) {
```
p0lyn0mial: does this test has any real value ? It test an endpoint that always returns OK. How do you feel about removing it ?

zreigz: agree

zreigz: I removed the file

p0lyn0mial:
fyi the tests are failing because there are no sufficient resources, you might want to wait until Monday :) (Henrik is working on it)

PR #2903, opened 2019-03-01
Define PDB for Elasticseasrch

xrstf:
This attempts to stabilize the ES cluster on systems with premtible nodes by forbidding too many unavailable nodes.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/logging/elasticsearch/templates/es-data-pdb.yaml
+++ b/config/logging/elasticsearch/templates/es-data-pdb.yaml
@@ -1,0 +1,12 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  labels:
+    component: elasticsearch
+  name: es-data
+spec:
+  maxUnavailable: 1
+  selector:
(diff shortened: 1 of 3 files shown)


kron4eg:
GKE preemption has nothing to do with EvictionAPI. It just kills the whole node, after 30 seconds after initial signal. https://cloud.google.com/compute/docs/instances/preemptible#preemption_process

What I mean is — this change wouldn't help preemptive nodes in no way.

xrstf:
You are slightly wrong. On *our* GKE setups we have Henrik's [preemtible-manager](https://github.com/kubermatic/infra/tree/master/apps/preemtible-manager), which drains and then deletes all nodes that are older than 6 hours. Having a PDB will make sure that when it's draining the second node, this will hang/wait until the just killed ES pod is rescheduled.

kron4eg:
Didn't know about preemtible-manager, that changes the game completely.

/shrug
/lgtm
/approve

PR #2907, opened 2019-03-02
Use NamedDeploymentCreators

alvaroaleman:

Moves the handling of Deployments from Deployment creators to NamedDeploymentCreators. This has two advantages:

* We don't have to always call `create` at least twice, once just to get the name
* In the openshift controller, this gets rid of the `NamedDeploymentCreator` type, which so far forced all `NamedDeploymentCreators` to have the same signature and take args even when they didn't use them (happend e.G. for the machine-controller)

Adjusting the codegen unfortunately required to adjust the whole codebase, sorry for the big PR

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


mrIncompetent:
ping prow

PR #2909, opened 2019-03-02
Split Kubermatic alerts into master and seed

### LINKED ISSUE #2861: Disable KubermaticAPIDown alerts for non-master seeds
**User Story**

We don't deploy the API and UI into non-master seeds anymore as they are not used there. This currently results in alerts being triggered there.

**Acceptance criteria**

* Alerts for API and UI are not deployed onto non-master seed clusters

/cc @xrstf @kron4eg

xrstf:
The Kubermatic chart now allows to not deploy master components, so we need to make sure that there are no false positive alerts being fired all the time.

Existing installations will continue to work because their `kubermatic-*.yaml` expression in the values.yaml will match both new alert YAML files.

Fixes #2861 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/monitoring/prometheus/rules/general-cert-manager.yaml
+++ b/config/monitoring/prometheus/rules/general-cert-manager.yaml
@@ -1,0 +1,18 @@
+# This file has been generated, do not edit.
+groups:
+- name: cert-manager
+  rules:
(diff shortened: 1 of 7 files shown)


kron4eg (approved):
/lgtm 
/approve

PR #2914, opened 2019-03-04
Remove dependency of the API onto the typed cluster-api client

alvaroaleman:

Removes the dependency to the typed cluster api clientset from everywhere except the IPAM controller. The purpose of this effort is to reduce the transient dependency on the client-go version that is used by the cluster-api project.

IMHO the code also gets easier as we don't need a variety of typed clients but instead can use just one client for everything.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


api/pkg/handler/node_v1.go
```diff
@@ -357,14 +350,14 @@ func getMachineForNode(node *corev1.Node, machines []clusterv1alpha1.Machine) *c
 	return nil
 }
 
-func findMachineAndNode(name string, machineClient clusterv1alpha1clientset.Interface, kubeClient kubernetes.Interface) (*clusterv1alpha1.Machine, *corev1.Node, error) {
-	machineList, err := machineClient.ClusterV1alpha1().Machines(metav1.NamespaceSystem).List(metav1.ListOptions{IncludeUninitialized: true})
-	if err != nil {
+func findMachineAndNode(ctx context.Context, name string, client ctrlruntimeclient.Client) (*clusterv1alpha1.Machine, *corev1.Node, error) {
```
alvaroaleman, line 353: What has to be noted here is that so far the `machineClient` was sometimes an impersoanted client and sometimes not, but the `kubeClient` has _always_ been an admin client. I am not completely sure if its okay to change it like this but then again, why should one of the clients have more permissions than the user?

zreigz:
I think something is missing here. I have errors `Error 500: failed to load machines from cluster: no kind is registered for the type v1alpha1.MachineList in scheme "k8s.io/client-go/kubernetes/scheme/register.go:61"`

api/pkg/handler/node_v1.go
zreigz: cc @p0lyn0mial

p0lyn0mial: only the legacy endpoints should use admin `machineClient`, the new endpoints must use impersonated version. `kubeClient` is only used (node_v1) by `nodes` endpoints to find corresponding machines and as we don't add `RBAC Roles/Bindings` to `nodes` we use the admin client.

alvaroaleman: I dont follow. So is it okay to use or not use the same impersonation for the `kubeClient` as for the `machineClient`?

alvaroaleman:
> I think something is missing here. I have errors Error 500: failed to load machines from cluster: no kind is registered for the type v1alpha1.MachineList in scheme "k8s.io/client-go/kubernetes/scheme/register.go:61"

I've added the scheme to the two Getters for a `UserClusterConnectionProvider`:

* https://github.com/kubermatic/kubermatic/pull/2914/files#diff-8d546d64b62ef142f8ce1b8487fc0a85R39
* https://github.com/kubermatic/kubermatic/pull/2914/files#diff-8d546d64b62ef142f8ce1b8487fc0a85R49

But still the same error :( @p0lyn0mial Do you have an idea what could be missing?

api/pkg/handler/node_v1.go
p0lyn0mial: no, no impersonation for `kubeClient` - there are no RBAC Roles/Bindings in place for `node` resources.

p0lyn0mial:
> But still the same error :( @p0lyn0mial Do you have an idea what could be missing?

@alvaroaleman I will look into it right in the morning.

alvaroaleman:
> @alvaroaleman I will look into it right in the morning.

Found out myself, had nothing to do with scheme or not scheme. The problem is the impersonated users do not have enough permissions for api discovery which results in the discovery client that backs the ctlruntime client not working at all. Its easely fixed (once you know and understand the issue after 4h debugging) [by adding the `system:authenticated` group to the impersonation config](https://github.com/kubermatic/kubermatic/pull/2914/commits/0894dec33aa9bad79814a280c63604855748b0d8)

api/pkg/handler/node_v1.go
alvaroaleman: Any objections against extending the existing roles to allow a get on `node` resources?

p0lyn0mial: sure go ahead, we were planing to add RBAC to more resources if not all in the future.

alvaroaleman: Done

zreigz:
Now everything seems to work.
waiting for @p0lyn0mial  before approval
/lgtm

alvaroaleman:
/hold

RBAC permissions for nodes dont seem to properly work yet.

zreigz:
or user cluster RBAC controller didn't start correctly

zreigz:
when I start my local env and execute `./run-controller.sh` the `usercluster-controller` never start from deployment

alvaroaleman:
> when I start my local env and execute ./run-controller.sh the usercluster-controller never start from deployment

What do you mean by "never start from deployment"?

I have now adjusted the permissions for the roles created by the rbac controller, also fixed up its script so it can actually be tested (had a conflicting port) and manually verified that its possible via UI to:

* Create clusters
* Update MachineDeployments
* Show existing machines that are not part of a machineDeployment
* Delete existing machines that are not part of a machineDeployment

@zreigz @p0lyn0mial PTAL

zreigz:
@alvaroaleman when I create a cluster and check my pods I have something like this:
```
kubectl get pods -n cluster-7jx6bvcglv
NAME                                         READY   STATUS             RESTARTS   AGE
apiserver-687b795495-r2tbc                   3/3     Running            0          24m
controller-manager-597fdccf4d-zlm4q          2/2     Running            0          24m
dns-resolver-6b7bcf64b-hz78m                 2/2     Running            0          24m
dns-resolver-6b7bcf64b-wh77s                 2/2     Running            0          24m
etcd-0                                       1/1     Running            0          24m
etcd-1                                       1/1     Running            0          24m
etcd-2                                       1/1     Running            0          24m
kube-state-metrics-8db456c6b-xbfn6           1/1     Running            0          24m
machine-controller-5f56bf6b7f-sf8lk          1/1     Running            0          24m
machine-controller-webhook-95cd7457b-lphkb   1/1     Running            0          24m
metrics-server-7857c57586-g8fl4              3/3     Running            0          24m
metrics-server-7857c57586-vf9j6              3/3     Running            0          24m
openvpn-server-69bcd66fd6-2m9m4              2/2     Running            0          24m
prometheus-0                                 1/1     Running            0          24m
scheduler-7cbccf94cd-jbxjx                   2/2     Running            0          24m
usercluster-controller-68f469586d-jj6nz      0/1     ImagePullBackOff   0          24m
lukasz@lukasz:~/Downloads$ kubectl logs usercluster-controller-68f469586d-jj6nz -n cluster-7jx6bvcglv
Error from server (BadRequest): container "usercluster-controller" in pod "usercluster-controller-68f469586d-jj6nz" is waiting to start: trying and failing to pull image

```

alvaroaleman:
@zreigz Not sure why that happens but I doubt it has anything to do with this PR as it doesn't change anything about the deployment of the user cluster controller. Also the error you are seeing does not happen in CI, otherwise it would fail with "waiting for controlplane to get ready"

zreigz:
yes, it's my internal problem for @p0lyn0mial everything works fine. Also @maciaszczykm has the same problem with running our system locally. For him also usercluster-controller can not start. I will investigate it later.
/lgtm

api/pkg/cluster/client/client.go
```diff
@@ -16,34 +16,48 @@ import (
...
 )
 
+// UserClusterConnectionProvider describes the interface available for accessing
+// resources inside the user cluster
+type UserClusterConnectionProvider interface {
+	GetClient(*kubermaticv1.Cluster, ...ConfigOption) (kubernetes.Interface, error)
+	GetApiextensionsClient(*kubermaticv1.Cluster, ...ConfigOption) (apiextensionsclientset.Interface, error)
+	GetAdmissionRegistrationClient(*kubermaticv1.Cluster, ...ConfigOption) (admissionregistrationclientset.AdmissionregistrationV1beta1Interface, error)
+	GetKubeAggregatorClient(*kubermaticv1.Cluster, ...ConfigOption) (aggregationclientset.Interface, error)
+	GetDynamicClient(*kubermaticv1.Cluster, ...ConfigOption) (ctrlruntimeclient.Client, error)
```
p0lyn0mial, line 30: this is really nice, the interface for interacting with API is clean :) What's the current status of the library (`controller-runtime/pkg/client`) ? Isn't in alpha ? Are there any known limitations ?

p0lyn0mial:
> yes, it's my internal problem for @p0lyn0mial everything works fine

ack, tested on my local machine and it actually works ! :)

api/pkg/cluster/client/client.go
alvaroaleman: >  Isn't in alpha ?

No, where did you get that from?

We already use it in various places like e.G.:

* User Cluster Controller and User Cluster RBAC controller
* Openshift controller
* Partly in the cluster controller (dynamic cache from controller-runtime)

We plan to eventually make all controllers controller-runtime based because it is a lot easier to use than direct client-go and also reduces the amount of boilerplate needed for controllers drastically, making the code faster to read and the actual business logic easier to find. So if there are actual issues I would vote to just fix them upstream.

It is also used by e.G. the cluster-api project. I can't find anything [in the docs](https://github.com/kubernetes-sigs/controller-runtime/blob/master/VERSIONING.md#users) that says it was alpha.

> Are there any known limitations ?

The fake client seems to be a bit more limited and there is no `DeleteCollection` although there is a [open WIP PR](https://github.com/kubernetes-sigs/controller-runtime/pull/324) for that. Other than that none that I am aware of.

p0lyn0mial: Usually major number of 0 is used to indicate pre release versions, see https://github.com/kubernetes-sigs/controller-runtime/releases

other than that it looks really nice :)

api/cmd/kubermatic-api/main.go
```diff
@@ -114,9 +114,14 @@ func createInitProviders(options serverRunOptions) (providers, error) {
 			kubermaticSeedInformerFactory := kubermaticinformers.NewSharedInformerFactory(kubermaticSeedClient, informer.DefaultInformerResyncPeriod)
 			defaultImpersonationClientForSeed := kubernetesprovider.NewKubermaticImpersonationClient(cfg)
 
+			userClusterConnectionProvider, err := client.NewExternal(kubeInformerFactory.Core().V1().Secrets().Lister())
+			if err != nil {
+				glog.Fatalf("failed to get usreClusterConnectionProvider: %v", err)
```
mrIncompetent: typo + return an error here. Only die with a `fatal` inside the `main` func

alvaroaleman: fixed

api/pkg/controller/rbac-user-cluster/reconciler.go
```diff
@@ -25,6 +25,7 @@ type reconciler struct {
 
 // Reconcile creates and updates ClusterRoles and ClusterRoleBinding to achieve the desired state
 func (r *reconciler) Reconcile(resourceName string) error {
+	glog.Infof("Reconciling RBAC for %s", resourceName)
```
mrIncompetent: Following https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/devel/logging.md

this might rather be a `glog.V(4)` ? WDYT?

api/pkg/controller/rbac-user-cluster/mapper.go
```diff
@@ -58,6 +58,11 @@ func GenerateRBACClusterRole(resourceName string) (*rbacv1.ClusterRole, error) {
 				Resources: []string{machinedeployments, machines},
 				Verbs:     verbs,
 			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"nodes"},
```
p0lyn0mial, line 63: i'm going to add an issue to our backlog to create different verbs for different groups as we do for other resources.

/cc @zreigz

alvaroaleman: I don't follow. This is a per-resource setting, not a per-group one. And what's the advantage of factoring it out?

p0lyn0mial: no, this is per-group (this method will be called for each group we have - owners, editors, viewers), see: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/rbac-user-cluster/controller.go#L24


the advantage is a unified access to a cluster, for example readers can only view resources.

api/pkg/controller/rbac-user-cluster/reconciler.go
alvaroaleman: Updated

mrIncompetent:
Sorry for the delay!
/lgtm

PR #2920, opened 2019-03-05
Use EnsureNamedObject for reconciling CRDs

nikhita:
- Generates Reconciliation functions for CRDs using `EnsureNamedObject`
- Adds a field `UseNamedObject` field to the `reconcileFunctionData` struct so that `EnsureNamedObject` can be used for multiple types selectively.

/cc @mrIncompetent @alvaroaleman 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/codegen/reconcile/main.go
+++ b/api/codegen/reconcile/main.go
@@ -24,13 +24,15 @@ func main() {
 				ResourceImportPath: "k8s.io/api/core/v1",
 			},
 			{
-				ResourceName: "Secret",
-				ImportAlias:  "corev1",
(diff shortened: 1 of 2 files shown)


alvaroaleman:
I am fine with the change by itself but I wonder on what basis we decide if we want to use `EnsureNamedObject` or `EnsureObject`. Do we want to eventually use `EnsureNamedObject` everywhere?

mrIncompetent:
> Do we want to eventually use EnsureNamedObject everywhere?

Yes - we just need to migrate all creators

PR #2921, opened 2019-03-05
Openshift: Deploy distinct set of addons

alvaroaleman:

Currently the addoninstaller and addon controller always deploy the same set of addons. This PR changes that so that they have a distinct config based on weather its an Openshift or Kubernetes cluster.

Also this PR adds required networking addon for Openshift.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


openshift_addons/networking/os-networking.yaml
```diff
@@ -0,0 +1,201 @@
+apiVersion: v1
```
alvaroaleman, line 2: It would be better if we deployed the Namespace via the user cluster controller as it takes a moment for it to be ready, making always at least one reconciliation in the addon-controller fail

openshift_addons/release.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -ex
+export TAG=v0.1
+
+docker build -t quay.io/kubermatic/openshift-addons:${TAG} .
```
alvaroaleman, line 6: This is a private repo for now

alvaroaleman: Changed that to public again, because otherwise we have to pass docker credentials into the `verify-addons-up-to-date` presubmit

config/kubermatic/values.yaml
```diff
@@ -51,20 +51,29 @@ kubermatic:
...
-      defaultAddons:
-      - canal
-      - dashboard
-      - dns
-      - kube-proxy
-      - openvpn
-      - rbac
-      - kubelet-configmap
-      - default-storage-class
+      kubernetes:
```
mrIncompetent, line 54: As this is backwards incompatible, please create an issue for the docs or start a page for the v2.9 -> v2.10 migration

openshift_addons/networking/openshift-node-configmap-deployer.yaml
```diff
@@ -0,0 +1,90 @@
...
+      labels:
+        app: create-bind-mounts
+    spec:
+      hostPID: true
+      serviceAccount: sdn
+      serviceAccountName: sdn
+      hostNetwork: true
+      containers:
+      - name: startup-script
+        image: quay.io/kubermatic/startup-script:v0.1.0
```
mrIncompetent, line 21: Wanted to comment on this regarding breakage of "custom docker registry".
But then i realized that we don't respect the custom registry at all in the addons.

I created https://github.com/kubermatic/kubermatic/issues/2924 to follow up on that.

openshift_addons/networking/os-networking.yaml
mrIncompetent: Might be worth to be added to the OpenShift todos doc

config/kubermatic/values.yaml
alvaroaleman: What about starting a document inside the kubermatic/kubermatic repo "migration docs 2.10"? Otherwise it is public already

mrIncompetent: You could specify `publishDate` on the page, so it won't be published :)
https://gohugo.io/content-management/front-matter/

alvaroaleman: Nice! Here you go: https://github.com/kubermatic/docs/pull/109

openshift_addons/networking/os-networking.yaml
alvaroaleman: I've added it inside the addon as todo instead. Doing this is very simple and shouldn't take more than 30mins, I am just afraid I'll create merge conflicts with Nikhitas current work

alvaroaleman:
@mrIncompetent PTAL

PR #2931, opened 2019-03-05
Add NodeCSRApprover controller for openshift

alvaroaleman:

Adds a NodeCSRApprover controller to the usercluster controller manager that automatically approves CSRs from nodes. This is required in order for Openshift nodes to be able to join a cluster.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


api/pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		return err
+	}
+	for _, condition := range csr.Status.Conditions {
+		if condition.Type == certificatesv1beta1.CertificateApproved {
+			return nil
+		}
+	}
+
+	if csr.Spec.Username == "system:serviceaccount:openshift-infra:node-bootstrapper" ||
```
alvaroaleman: Actually I realized the auto approval for `system:serviceaccount:openshift-infra:node-bootstrappe` can be done via a rolebinding, the one for `system:node` not thought. The latter is a serving cert whereas the former is a client cert.

As a sidenote, maybe we could use this mechanism for kubernetes as well to have the kubelets request serving certs?

Also I could probably make the checking a bit stricter (verify usages and check the group name instead of `UserName starts with string X`. WDYT?

mrIncompetent: Using it for the kube kubelet serving cert: :+1: 

Making it more strict: :+1: 

We should also check if the IP's in the cert match the Node's IP

alvaroaleman: > Using it for the kube kubelet serving cert: ￼

Done

> Making it more strict: ￼

Done

>  We should also check if the IP's in the cert match the Node's IP

That is unfortunately not possible. A sample CSR requests the following SANs:

`hetzner-machinedeployment-5d5f5bcbc9-n2f4f, IP Address:159.69.183.155, IP Address:2A01:4F8:1C17:73E2:0:0:0:1, IP Address:172.17.0.1, IP Address:172.25.1.1`

Hostname and IPv4 address are on the node object, IPv6 address is not. The 172.17.0.1 and 172.25.1.1 are within the cluster IP range.

IMHO the current state is fine. The certs can only be used for serving due to their usage, so worst case is someone using a node identity and then does a mitm for another service in the cluster. Compared to "We dont verify at all in Kubernetes" that state is still much better and we can iterate later if needed

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -69,6 +70,12 @@ func main() {
 	if err := registerControllers(mgr, metrics); err != nil {
 		glog.Fatal(err)
 	}
+	if runOp.openshift {
```
mrIncompetent, line 73: This can be removed since its not valid for Kubernetes as well?

alvaroaleman: I think we could configure the nodes to request serving certificates. But we currently dont do that so right now it would be pointless to active it for Kubernetes

api/pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -0,0 +1,121 @@
...
+		Type:   certificatesv1beta1.CertificateApproved,
+		Reason: "Kubermatic NodeCSRApprover controller approved node serving cert",
+	}
+	csr.Status.Conditions = append(csr.Status.Conditions, approvalCondition)
+
+	if _, err := r.certClient.UpdateApproval(csr); err != nil {
+		return fmt.Errorf("failed to update approval for CSR %q: %v", csr.Name, err)
+	}
+
+	glog.V(4).Infof("Successfully approved CSR %q", csr.Name)
```
mrIncompetent: Worth a 2 i guess:
https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/devel/logging.md

alvaroaleman: Its debug logging on the debug log level. Whats wrong with that?

mrIncompetent: I would consider approving a cert approval a system state change, but i don't insist

alvaroaleman: IMHO this is a silly discussion because the UCCM always gets deployed with `v=4`. There is no option to custmize that. So in practise this thing is binary: Include it in the log level shown to customers or not, AKA log at level <= 4 or > 4. Both 4 and 2 are <= 4, so its always shown either way. Updated the pull.

PR #2933, opened 2019-03-06
add elasticsearch lifecycle hooks

mrIncompetent:
This follows the best practices described in the ES docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html

Instead of moving data, this just deactivates the shard allocation, which should be enough for our use case.

**Release note**:
```release-note
Add lifecycle hooks to the Elasticsearch StatefulSet
```


kron4eg:
/lgtm
/approve
/hold

Hold to let @xrstf see it too

xrstf:
/hold cancel

Love it!

PR #2936, opened 2019-03-06
add proposal for service account

zreigz:
**What this PR does / why we need it**: add a proposal for a service account

```release-note
NONE
```


zreigz:
@p0lyn0mial PTAL

PR #2939, opened 2019-03-06
Update machine-controller to v1.0.7 to allow initialization

alvaroaleman:

Updates the machine-controller to v1.0.7, this is required to allow the ipam-controller to work based off of annotations instead of initializers.

Fixes #

**Documentation**:

**Release note**:

```release-note
NONE
```

/assign @mrIncompetent

### DIFF
--- a/api/Gopkg.toml
+++ b/api/Gopkg.toml
@@ -121,7 +121,7 @@ required = [
 
 [[override]]
   name = "github.com/kubermatic/machine-controller"
-  version = "v1.0.6"
+  version = "v1.0.7"
 
 [[override]]
   name = "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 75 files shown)


api/Gopkg.toml
```diff
@@ -121,7 +121,7 @@ required = [
 
 [[override]]
   name = "github.com/kubermatic/machine-controller"
-  version = "v1.0.6"
+  version = "v1.0.7"
```
mrIncompetent, line 124: Even though we changed that file nothing in the vendor dir changed?

alvaroaleman: Apparently, yes

alvaroaleman:
Actually I will close this. It was a terrible idea to define a const that is bascially an api in the same place were all the business logic is. Also it contradicts the idea of not being dependent on the cluster api typed clients.

/close

alvaroaleman:
Actually, even with the const defined in the wrong package we need the functionality for #2940 so reopening. Sorry for the noise :sa: 

/reopen

PR #2940, opened 2019-03-06
Ipam controller: Move into UCCM and work based off of Annotation

alvaroaleman:

* Rewrites the IPAM controller to be controller-runtime based
* Moves it into the UCCM
* Makes it work based off of an Annotation as Initializers got removed upstream and will not be shipped with 1.14

One thing to note thought: The functionality the Ipam controller provides is currently inaccessible for Kubermatic users because there is no API endpoint that can create uninitialized machines. This was the case before and this continues to be the case. Fixing this is IMHO better left to the API/UI team thought.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


api/pkg/controller/ipam/ipam_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// ControllerName is the name of this controller
+	ControllerName                 = "kubermatic_ipam_controller"
+	annotationMachineUninitialized = "machine-controller.kubermatic.io/initializers"
```
alvaroaleman, line 34: Ideally we would import this but since its currently in the controller package and I don't want to pull that with all its dependencies in so yeah, I've redefined it here

mrIncompetent:
> One thing to note thought: The functionality the Ipam controller provides is currently inaccessible for Kubermatic users because there is no API endpoint that can create uninitialized machines. This was the case before and this continues to be the case. Fixing this is IMHO better left to the API/UI team thought.

Do we have a issue for this?

/lgtm
/approve

api/pkg/controller/ipam/ipam_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	for _, cidr := range r.cidrRanges {
+		ip, err := r.getNextFreeIPForCIDR(cidr, usedIps)
+		if err == nil {
+			return ip, cidr, nil
+		}
+	}
+
+	return nil, Network{}, cidrExhaustedError{}
```
p0lyn0mial: why not simply return a standard error (`errors.New()`)? From what I have seen there is no code that makes use of `cidrExhaustedError` type.

api/pkg/handler/v1/common/version_skew.go
```diff
@@ -102,7 +101,7 @@ func CheckClusterVersionSkew(ctx context.Context, userInfo *provider.UserInfo, c
 
 // getKubeletVersions returns the list of all kubelet versions used by a given cluster's Machines and MachineDeployments
 func getKubeletVersions(ctx context.Context, client ctrlruntimeclient.Client) ([]string, error) {
-	listOpts := &ctrlruntimeclient.ListOptions{Raw: &metav1.ListOptions{IncludeUninitialized: true}}
+	listOpts := &ctrlruntimeclient.ListOptions{}
```
p0lyn0mial: nit: `emptyListOpts`

p0lyn0mial (commented):
I haven't reviewed IPAM controller because I think that the logic hasn't changed only the wiring. Two tiny comments and this lgtm.

Many thanks for adding the issue to our backlog.

PR #2943, opened 2019-03-06
Conformance tests: Enable monitor mode

alvaroaleman:
**What this PR does / why we need it**:

Fixes cleanup when the job receives a signal:

```
$ cat test.sh 
#!/bin/bash

set -o monitor

trap 'echo FOO' EXIT
echo "sleeping"
sleep 1000

$ docker run --rm -v $PWD:/script -e ENTRYPOINT_OPTIONS='{"args":["/script/test.sh"],"process_log":"/log","marker_file":"/marker"}' gcr.io/k8s-prow/entrypoint:v20190301-1278de5-with-bash /entrypoint
sleeping
{"component":"entrypoint","level":"error","msg":"Entrypoint received interrupt: terminated","time":"2019-03-06T21:13:02Z"}
FOO
{"component":"entrypoint","level":"error","msg":"Process did not exit before 15s grace period","time":"2019-03-06T21:13:17Z"}
{"component":"entrypoint","error":"os: process already finished","level":"error","msg":"Could not kill process after grace period","time":"2019-03-06T21:13:17Z"}
{"component":"entrypoint","error":"process aborted","level":"error","msg":"Error executing test process","time":"2019-03-06T21:13:17Z"}

vs

$ cat test.sh 
#!/bin/bash

#set -o monitor

trap 'echo FOO' EXIT
echo "sleeping"
sleep 1000

$ docker run --rm -v $PWD:/script -e ENTRYPOINT_OPTIONS='{"args":["/script/test.sh"],"process_log":"/log","marker_file":"/marker"}' gcr.io/k8s-prow/entrypoint:v20190301-1278de5-with-bash /entrypoint
sleeping
{"component":"entrypoint","level":"error","msg":"Entrypoint received interrupt: terminated","time":"2019-03-06T21:27:58Z"}
{"component":"entrypoint","level":"error","msg":"Process did not exit before 15s grace period","time":"2019-03-06T21:28:13Z"}
{"component":"entrypoint","error":"process aborted","level":"error","msg":"Error executing test process","time":"2019-03-06T21:28:13Z"}
```
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Release note**:

```release-note
none
```

PR #2944, opened 2019-03-06
Wait for cloud infra before attempting to render the cloud-config

alvaroaleman:

Currently we often times create a cloud-config before the cloud-provider infrastructure is ready, resulting in the cloud-config getting updated shortly after and the apiserver & controller-manager to start off with two different replicasets with one pod each, because no pod is ready yet because they are blocked on etcd startup/apiserver startup.

With this change we wont create the cloud-config before the cloud-provider infrastructure is ready, making sure at the point in time where we create the apiserver & controller manager deployments an up-to-date cloud-config based on the ready infrastructure is available.

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/cluster/resources.go
+++ b/api/pkg/controller/cluster/resources.go
@@ -52,6 +52,10 @@ func (cc *Controller) ensureResourcesAreDeployed(cluster *kubermaticv1.Cluster) 
 
 	// Wait until the cloud provider infra is ready before attempting
 	// to render the cloud-config
+	// TODO: Model resource deployment as a DAG so we don't need hacks
+	// like this combined with tribal knowledge and "someone is noticing this
+	// isn't working correctly"
+	// https://github.com/kubermatic/kubermatic/issues/2948
 	if !cluster.Status.Health.CloudProviderInfrastructure {
 		cc.enqueueAfter(cluster, 1*time.Second)
 		return nil
--- a/api/pkg/controller/cluster/resources.go
+++ b/api/pkg/controller/cluster/resources.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


api/pkg/controller/cluster/resources.go
```diff
@@ -44,6 +45,18 @@ func (cc *Controller) ensureResourcesAreDeployed(cluster *kubermaticv1.Cluster)
...
 	}
 
+	// check that all StatefulSets are created
+	if err := cc.ensureStatefulSets(cluster, data); err != nil {
+		return err
+	}
+
+	// Wait until the cloud provider infra is ready before attempting
+	// to render the cloud-config
+	if !cluster.Status.Health.CloudProviderInfrastructure {
```
mrIncompetent, line 59: Move that one level above so `ensureResourcesAreDeployed` does not get executed when `CloudProviderInfrastructure` is not true. 
Otherwise we risk running into the same thing again.

alvaroaleman: No, that defeats the purpose of moving the cloud controller reconiliation into a dedicated controller in the first place. The purpose was to allow as much parallelization as possible.  The only thing that needs the cloud provider infra is the cloud config (and transiently the apiserver and controller-manager deployment).

Ideally this whole thing would be a graph where every resource can specifiy its requirements but unfortunately we are not there yet

mrIncompetent: Good point, but then please create a code TODO + Issue to remove this with a graph

alvaroaleman: Good idea, there: https://github.com/kubermatic/kubermatic/issues/2948
Very unfortunate that I travel for my holidays, would really like to work on that...

alvaroaleman:
/hold

To make sure #2943 merges first

PR #2945, opened 2019-03-07
Remove drone

nikhita:
We now use prow for presubmits and postsubmits so drone can be removed. :wastebasket: 

/assign @alvaroaleman 

**Release note**:

```release-note
NONE
```


alvaroaleman:
/approve
/lgtm
/hold

I want #2943 to merge first, as it avoids stale clusters in the CI env

PR #2951, opened 2019-03-07
deploy dex before kubermatic

mrIncompetent:
The kubermatic API server depends on DEX which is why it must be deployed first.

**Release note**:
```release-note
NONE
```


alvaroaleman:
When iterating over a bash hashmap, the elements are sorted by the internal hash used to build the hashmap: https://stackoverflow.com/questions/29161323/how-to-keep-associative-array-order

So the ordering you see has nothing to do with the ordering in which the elements are deployed. And because the postubmit works I presume dex gets deplyoed before Kubermatic: https://prow.loodse.com/?type=postsubmit

mrIncompetent:
Closed in favor of https://github.com/kubermatic/kubermatic/pull/2953

PR #2954, opened 2019-03-07
Blackbox-Exporter

xrstf:
Adding the exporter chart allows us to easily monitor external reachability of ingresses. To not force it on customers, we will keep the actual probe configuration in our values.yaml (i.e. specify it in the values.yaml unter `prometheus.scraping.configs`). The newly defined alerts should not fire if the exporter is not installed.

Since the new chart is so simple, we will publish it in our installer distribution.

**Release note**:
```release-note
add blackbox-exporter chart
```


kron4eg (commented):
/lgtm
/approve

PR #2955, opened 2019-03-07
update conformance test image

mrIncompetent:
Update conformance test image.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/containers/conformance-tests/install-kube-tests.sh
+++ b/containers/conformance-tests/install-kube-tests.sh
@@ -7,11 +7,10 @@ set -euox pipefail
 TMP_ROOT="./.install-tmp"
 
 declare -A FULL_VERSIONS
-FULL_VERSIONS["1.10"]=$(curl -Ss https://storage.googleapis.com/kubernetes-release/release/stable-1.10.txt)
-FULL_VERSIONS["1.11"]=$(curl -Ss https://storage.googleapis.com/kubernetes-release/release/stable-1.11.txt)
-FULL_VERSIONS["1.12"]=$(curl -Ss https://storage.googleapis.com/kubernetes-release/release/stable-1.12.txt)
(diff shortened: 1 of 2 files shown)


containers/conformance-tests/install-kube-tests.sh
```diff
@@ -6,10 +6,18 @@ set -euox pipefail
 
 TMP_ROOT="./.install-tmp"
 
-for VERSION in 1.10 1.11 1.12 1.13; do
+declare -A FULL_VERSIONS
+FULL_VERSIONS["1.10"]=$(curl -Ss https://storage.googleapis.com/kubernetes-release/release/stable-1.10.txt)
```
alvaroaleman: Does it make sense to keep 1.10?

Also I'd actually prefer it if we did this explicitly because right now we don't even know what patch version we re using for testing

alvaroaleman (commented):
Upstream added platform-specific test tarballs, can we maybe use them to reduce the download times when creating this image? https://github.com/kubernetes/kubernetes/pull/74065

containers/conformance-tests/install-kube-tests.sh
mrIncompetent: > Does it make sense to keep 1.10?

Yes. We have still tests for v1.10 (Although they don't run by default).

> Also I'd actually prefer it if we did this explicitly because right now we don't even know what patch version we re using for testing

Fixed in https://github.com/kubermatic/kubermatic/pull/2955/commits/d04451cfc10ce0ab81d6b8840f472d588343e834

mrIncompetent:
> Upstream added platform-specific test tarballs, can we maybe use them to reduce the download times when creating this image? kubernetes/kubernetes#74065

If i read the PR correctly and the logs i got, it should be used automatically as we use the `get-kube-binaries.sh` script.

PR #2959, opened 2019-03-07
do not send pod annotations via fluentbit to elasticsearch

xrstf:
Annotations do not provide much value when viewing logs in Kibana, but what they do is clutter the UI and blow up the index disk usage noticibly. Let's disable sending annotations at all, saving us the hassle to define Elasticsearch Mapping settings.

**Release note**:
```release-note
pod annotations are no longer logged in Elasticsearch
```


kron4eg:
lol wtf with this `This pull-request has been approved by: [empty list here]`

PR #2960, opened 2019-03-07
define a post-start hook to provision the default kibana index pattern

xrstf:
This makes Kibana usable right after it was started for the first time, without the customer having to fiddle with it manually.

**Release note**:
```release-note
Kibana is automatically initialized in new installations.
```


kron4eg (approved):
/lgtm
/approve

PR #2962, opened 2019-03-07
Openshift controller: Handle all required secrets

alvaroaleman:

Makes the openshift controller manage all secrets used for Openshift

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/openshift/openshift_controller.go
+++ b/api/pkg/controller/openshift/openshift_controller.go
@@ -9,6 +9,11 @@ import (
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/apiserver"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/certificates"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/etcd"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/machinecontroller"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/openvpn"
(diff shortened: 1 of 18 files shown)


api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -397,6 +434,31 @@ func (r *Reconciler) deployments(ctx context.Context, osData *openshiftData) err
 	return nil
 }
 
+func (r *Reconciler) ensureNamespace(ctx context.Context, c *kubermaticv1.Cluster) error {
+	if c.Status.NamespaceName == "" {
+		if err := r.updateCluster(ctx, c.Name, func(c *kubermaticv1.Cluster) {
```
mrIncompetent: As i just realized the `updateCluster` function does not return the cluster object nor does it modify a passed cluster pointer,

doesn't this mean this function will try to create a namespace without name in the first execution?

api/pkg/resources/apiserver/tls-serving-certificate.go
```diff
@@ -55,7 +63,7 @@ func TLSServingCertificateCreator(data resources.SecretDataProvider) resources.N
 					fmt.Sprintf("%s.%s.svc.cluster.local", resources.ApiserverInternalServiceName, data.Cluster().Status.NamespaceName),
 				},
 				IPs: []net.IP{
-					*externalIP,
+					net.ParseIP(externalIP),
```
mrIncompetent: Add a check to validate that `net.ParseIP` did not return `nil` - as it does when it fails parsing the ip

api/pkg/controller/openshift/openshift_controller.go
alvaroaleman: Good point, that signature didn't make too much sense. Updated it to take a cluster pointer instead that will later on contain the updated cluster

api/pkg/resources/apiserver/tls-serving-certificate.go
alvaroaleman: Fixed

PR #2970, opened 2019-03-08
Fix apiserver init check

mrIncompetent:
This replaces the `etcdctl endpoint health` with a `etcdctl put quorum-check something`.

The difference is that `etcdctl endpoint health` checks if all specified endpoints are healthy - in our case 3. This can lead to downtime of the API server if only 1 etcd is down.

`etcdctl put quorum-check something` Does a simple write to etcd. If it succeeds we have a quorum and know that the API server can start.

**Release note**:
```release-note
Ensure that the API server init container only waits for etcd quorum and not all members.
```

### DIFF
--- a/api/pkg/resources/apiserver/deployment.go
+++ b/api/pkg/resources/apiserver/deployment.go
@@ -109,7 +109,8 @@ func DeploymentCreator(data resources.DeploymentDataProvider) resources.Deployme
 				Command: []string{
 					"/bin/sh",
 					"-ec",
-					fmt.Sprintf("until ETCDCTL_API=3 /usr/local/bin/etcdctl --cacert=/etc/etcd/pki/client/ca.crt --cert=/etc/etcd/pki/client/apiserver-etcd-client.crt --key=/etc/etcd/pki/client/apiserver-etcd-client.key --dial-timeout=2s --endpoints='%s' endpoint health; do echo waiting for etcd; sleep 2; done;", strings.Join(etcdEndpoints, ",")),
+					// Write a key to etcd. If we have quorum it will succeed.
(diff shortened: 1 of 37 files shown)


api/pkg/resources/apiserver/deployment.go
```diff
@@ -109,7 +109,8 @@ func DeploymentCreator(data resources.DeploymentDataProvider) resources.Deployme
 				Command: []string{
 					"/bin/sh",
 					"-ec",
-					fmt.Sprintf("until ETCDCTL_API=3 /usr/local/bin/etcdctl --cacert=/etc/etcd/pki/client/ca.crt --cert=/etc/etcd/pki/client/apiserver-etcd-client.crt --key=/etc/etcd/pki/client/apiserver-etcd-client.key --dial-timeout=2s --endpoints='%s' endpoint health; do echo waiting for etcd; sleep 2; done;", strings.Join(etcdEndpoints, ",")),
+					// Write a key to etcd. If we have quorum it will succeed.
+					fmt.Sprintf("until ETCDCTL_API=3 /usr/local/bin/etcdctl --cacert=/etc/etcd/pki/client/ca.crt --cert=/etc/etcd/pki/client/apiserver-etcd-client.crt --key=/etc/etcd/pki/client/apiserver-etcd-client.key --dial-timeout=2s --endpoints='%s' put quorum-check something; do echo waiting for etcd; sleep 2; done;", strings.Join(etcdEndpoints, ",")),
```
alvaroaleman: Can we maybe prefix this with `kubermatic/` so we have one defined keyspace?

mrIncompetent: Done

PR #2976, opened 2019-03-08
Notify the API caller which cluster upgrades are restricted by kubelet versions 

### LINKED ISSUE #2541: Notify the API caller which cluster upgrades are restricted by kubelet versions
**User Story**
When a user of the Kubermatic API is calling the `.../upgrades` endpoint to get the list of available upgrades, it needs to know which upgrades are not possible at the time due to incompatible kubeletes running in the cluster.

**Acceptance criteria**
 - [x] The v1 API structure `MasterVersion` has an additional boolean field `restrictedByKubeletVersion`
 - [x] `getClusterUpgrades` fetches all Machines and MachineDeployments within the cluster. When responding with the list of master versions, it marks `restrictedByKubeletVersion`=`true` for all versions that are deemed to by incompatible with existing Machines/MachineDeployments (using #2539)
 - [x] `TestGetClusterUpgradesV1` contains a scenario that verifies `restrictedByKubeletVersion`=`true`  for incompatible master-kubelet pairs

maciaszczykm:
**What this PR does / why we need it**: Notifies the API caller which cluster upgrades are restricted by kubelet versions .

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Fixes #2541.

**Special notes for your reviewer**: n/a

**Documentation**: n/a

**Release note**:

```release-note
Mark cluster upgrades as restricted if kubelet version is incompatible.
```

### DIFF
--- a/api/pkg/handler/upgrade_test.go
+++ b/api/pkg/handler/upgrade_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/Masterminds/semver"
 	"github.com/go-test/deep"
@@ -29,9 +30,6 @@ import (
 func TestGetClusterUpgrades(t *testing.T) {
 	t.Parallel()
 
-	mdWithOldKubelet := genTestMachineDeployment("venus", `{"cloudProvider":"digitalocean","cloudProviderSpec":{"token":"dummy-token","region":"fra1","size":"2GB"}, "operatingSystem":"ubuntu", "operatingSystemSpec":{"distUpgradeOnBoot":true}}`, nil)
-	mdWithOldKubelet.Spec.Template.Spec.Versions.Kubelet = "v1.4.0"
-
 	tests := []struct {
 		name                       string
 		cluster                    *kubermaticv1.Cluster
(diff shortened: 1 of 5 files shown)


api/pkg/handler/upgrade_test.go
```diff
@@ -80,6 +86,51 @@ func TestGetClusterUpgradesV1(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "upgrade available but restricted by kubelet versions",
+			cluster: &kubermaticv1.Cluster{
```
p0lyn0mial: could you use `test.GenCluster` function ? for example:
```
cluster: func() *kubermaticv1.Cluster {
c := test.GenCluster(...)
// modify c
return c
}
```

p0lyn0mial:
one `nit` and this lgtm

api/pkg/handler/upgrade_test.go
maciaszczykm: Of course 👍

maciaszczykm:
@p0lyn0mial PTAL

p0lyn0mial:
@maciaszczykm thx

/lgtm
/approve

PR #2977, opened 2019-03-08
Fluentbit glog Parser

xrstf:
This adds a new parser for glog outputs, so we can more easily filter by warnings, errors, files etc. in Kibana.

Note that the documentation on Fluentbit is not really all that clear and that the regex is applied to *everything* it reads from the container runtime's logs. I.e. you are not just matching against

    E0308 11:31:46.678784       1 reflector.go:205] a lot of text here

but instead the whole, raw line from the file:

    2019-03-08T11:31:46.679082406Z stderr F E0308 11:31:46.678784       1 reflector.go:205] a lot of text here

This is why the regex is not anchored at the beginning and we allow for the timestamp and other junk to preceed our glog format.

cc @mrIncompetent 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/cert-manager/templates/ca-sync-job.yaml
+++ b/config/cert-manager/templates/ca-sync-job.yaml
@@ -8,6 +8,8 @@ spec:
     metadata:
       labels:
         app: ca-helper
+      annotations:
+        fluentbit.io/parser: glog
     spec:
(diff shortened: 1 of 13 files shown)


kron4eg (approved):
/approve

PR #2985, opened 2019-03-11
Update addons

mrIncompetent:
This updates the addons we install in user clusters.

**Release note**:
```release-note
Update user cluster addons
```


addons/canal/daemonset.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         # container programs network policy and routes on each
         # host.
         - name: calico-node
-          image: quay.io/calico/node:v2.6.2
```
mrIncompetent, line 40: Calico 3.6 fails with stating that one should use `v2.6.5+` before using `v3.0` due to a migration...

addons/dashboard/kubernetes-dashboard.yaml
```diff
@@ -0,0 +1,162 @@
...
+  labels:
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard
+  namespace: kube-system
+spec:
+  ports:
+  - port: 443
+    targetPort: 8443
+  selector:
+    k8s-app: kubernetes-dashboard
```
mrIncompetent, line 162: I just use the exact file which gets used upstream which should ease future comparisons

addons/dns/coredns-deployment.yaml
```diff
@@ -0,0 +1,171 @@
...
+        cache 30
+        loop
+        reload
+        loadbalance
+    }
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: coredns
```
mrIncompetent, line 57: Finally switching to CoreDNS

addons/kube-proxy/daemonset.yaml
```diff
@@ -1,73 +1,63 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
mrIncompetent, line 1: I replaced the manifest with what Kubeadm uses... Mostly the sorting of the keys is different

mrIncompetent:
/assign @kdomanski 

I verified that the changed addons work on new & existing clusters

addons/canal/daemonset.yaml
thz: Did you try updating `calico/cni` as well?
We need a big step from `v1.11.0` to `v3.5.1` here. If this requires another intermediate step it would be good to include it here too.

mrIncompetent: My bad, i now replaced all canal files with the files from upstream (That will make reviewing later much easier).
It now also uses the latest CNI version (For v2.6)

Source: https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/hosted/canal/

thz: 👍

mrIncompetent:
Image tag has been updated to `v0.2.0`

PR #2986, opened 2019-03-12
allow to re-enable the Grafana login form

xrstf:
In some situations we do not have Keycloak in front of cluster services and would still like to allow customers to login manually, i.e. when port-forwarding to Grafana for debugging cluster issues. This change makes it possible to have a plain ol' login form.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/monitoring/grafana/config/grafana.ini
+++ b/config/monitoring/grafana/config/grafana.ini
@@ -3,7 +3,7 @@ admin_password = {{ .Values.grafana.password | b64dec }}
 admin_user = {{ .Values.grafana.user | b64dec }}
 
 [auth]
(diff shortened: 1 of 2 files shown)


kron4eg:
>This change makes it possible to have a plain ol' login form.

In that case we should allow passing custom [auth config](http://docs.grafana.org/auth/overview/) down to grafana.

xrstf:
We have the admin user already, which is useless right now. Without this PR, we cannot even use the Grafana on ci (yes, we now have Grafana on that cluster ;-)).

kron4eg (approved):
/lgtm
/approve

PR #2989, opened 2019-03-12
Create initial node deployment with the cluster

### LINKED ISSUE #2928: Improve initial node deployment creation
**User Story** 
At the moment when the user is creating a cluster, there are separate requests to create the cluster and to create initial node deployment. Request to create initial node deployment is send from the UI after the user is redirected to the cluster details page and the cluster is healthy. In the meantime, initial node data is stored in the local storage. This causes potential issues, i.e. when the user closes the browser before initial node deployment creation or if the local storage is corrupted/removed.

To improve the situation we can start with changes in the API. We can wait for a cluster to be created in the initial request and create initial node deployment. It shouldn't take too long but it should fix most of the problems.

The target solution for this issue seems to be implementation of a new controller that will handle initial node deployment creation and will store machine deployments/some other CRDs in the seed cluster as a cache/queue and when the clusters will be ready then the controller can add machine deployments into the clusters.

**Acceptance criteria**
- [x] API should handle the creation of initial node deployments.
- [x] There should be a metric and alert for cases where creation will fail after a certain timeout.
- [ ] There should be tests to cover the changes in the API.
- [ ] UI should be adjusted and tested as well.

cc @mrIncompetent @alvaroaleman @kgroschoff @p0lyn0mial @thetechnick   

Feel free to edit this issue if you feel that description could be changed.

maciaszczykm:
**What this PR does / why we need it**: Contains changes that allow creating initial node deployments with the cluster, in the same requests.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Fixes #2928.

**Special notes for your reviewer**: Has to be merged at the same time as https://github.com/kubermatic/dashboard-v2/pull/1125.

**Documentation**: n/a

**Release note**:

```release-note
Initial node deployments are now created with the cluster.
```


api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +65,92 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+				err = common.KubernetesErrorToHTTPError(err)
+				break
+			}
+
+			// since we use an impersonated client when sending a request to the cluster
+			// there have to be RBAC Roles/Bindings in place otherwise we get 403 errors
+			// at the moment the cluster health doesn't check if the RBAC was generated
+			// so we might open a connection and get an access denied error
+			//
+			// TODO: we could extend the cluster health to report when the controller has finished generating RBAC
```
p0lyn0mial: Checking `Status.Health.AllHealthy()` is not enough. Since we use an impersonated client, RBAC has to be generated before we open a connection the cluster otherwise we might get occasionally 403. 

@mrIncompetent do you think we could extend cluster health to also include RBAC readiness in the cluster ?

mrIncompetent: I think we need such a "checkpoint" on the cluster.

But:
The necessary RBAC permissions are being applied by a controller which has no access to the actual cluster object?

p0lyn0mial: > The necessary RBAC permissions are being applied by a controller which has no access to the actual cluster object?

I think so, the controller (UCCM) runs in `cluster-abc` namespace and adds RBAC to a cluster created by the user.

> I think we need such a "checkpoint" on the cluster.

Not sure if this is the best solution, all I'm saying is that we need a way to know that UCCM has finished his job before we say that the cluster is ready.

mrIncompetent: > Not sure if this is the best solution, all I'm saying is that we need a way to know that UCCM has finished his job before we say that the cluster is ready.

But how do we know its ready?
The 2 potential solutions which come to my mind here:
- Lets the UCCM write a condition in the cluster object when it successfully completed the sync loop
  - This has the downside that we now need to make the UCC "seed cluster aware" which means we need to talk to 2 clusters now. 
- Let the UCCM write some checkpoint resource in the cluster which gets checked by the cluster controller
  - The cluster controller does a "is reachable" check during each sync anyway

p0lyn0mial: since the CM connects to the clusters anyway the second option seems compelling. It could read a "checkpoint" resource and report the status back all the way up to the method that checks if the cluster is healthy.

maciaszczykm:
Added metric to report the number of failures. How do we add them to alert manager?

/hold

maciaszczykm:
PTAL @p0lyn0mial

api/pkg/handler/v1/cluster/cluster.go
mrIncompetent: Sounds good, would you mind creating an Issue for that?

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+		// Create the initial node deployment in the background.
+		go createInitialNodeDeployment(&req.Body.NodeDeployment, newCluster, project, sshKeyProvider, dcs, clusterProvider, userInfo)
+
 		return convertInternalClusterToExternal(newCluster), nil
 	}
 }
 
+// TODO: We can add an event to the cluster after successful creation or failure.
+func createInitialNodeDeployment(nodeDeployment *apiv1.NodeDeployment, cluster *kubermaticapiv1.Cluster, project *kubermaticapiv1.Project,
+	sshKeyProvider provider.SSHKeyProvider, dcs map[string]provider.DatacenterMeta, clusterProvider provider.ClusterProvider, userInfo *provider.UserInfo) {
```
p0lyn0mial: we need to add a guard here - to recover from a panic, for example we could use `utilruntime.HandleCrash()` ("k8s.io/apimachinery/pkg/util/runtime) or create our own function.

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
 		}
+
+		// Create the initial node deployment in the background.
+		go createInitialNodeDeployment(&req.Body.NodeDeployment, newCluster, project, sshKeyProvider, dcs, clusterProvider, userInfo)
+
 		return convertInternalClusterToExternal(newCluster), nil
 	}
 }
 
+// TODO: We can add an event to the cluster after successful creation or failure.
```
p0lyn0mial: rm the comment.

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
 		if err != nil {
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
+
+		// Create the initial node deployment in the background.
+		go createInitialNodeDeployment(&req.Body.NodeDeployment, newCluster, project, sshKeyProvider, dcs, clusterProvider, userInfo)
```
p0lyn0mial: do we want to make a `NodeDeployment` optional ? If not provided we create a cluster only.

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	defer cancelCtx()
+
+	nd, err := machineresource.Validate(nodeDeployment, cluster.Spec.Version.Semver())
+	if err != nil {
+		glog.V(5).Infof("initial node deployment for cluster %s is not valid: %v", cluster.Name, err)
+		return
+	}
+
+	interval := 10 * time.Second
```
p0lyn0mial: we could use `wait.Poll`(`"k8s.io/apimachinery/pkg/util/wait"`) method instead, wdyt ?

api/pkg/metrics/middleware.go
```diff
@@ -25,10 +25,18 @@ var httpRequestsDuration = prometheus.NewHistogramVec(
 	[]string{"method", "route"},
 )
 
+var InitNodeDeploymentFailures = prometheus.NewCounterVec(prometheus.CounterOpts{
```
p0lyn0mial: I would prefer to pass the metric as a dependency to  `CreateEndpoint` function. This file `middleware.go` contains metrics that enhance all HTTP request, whereas we want to add a metric to only one endpoint.

api/pkg/metrics/middleware.go
```diff
@@ -25,10 +25,18 @@ var httpRequestsDuration = prometheus.NewHistogramVec(
...
+	Name: "init_node_deployment_failures",
+	Help: "The number of times initial node deployment couldn't be created within the 5m timeout",
```
p0lyn0mial: not sure if putting a timeout in the comment is good, it can easily get out of sync with the code.

api/pkg/metrics/middleware.go
```diff
@@ -25,10 +25,18 @@ var httpRequestsDuration = prometheus.NewHistogramVec(
...
+},
+	[]string{"cluster"},
```
p0lyn0mial: should we also add a data center name so that we can locate the faulty cluster quicker?

/cc @mrIncompetent

api/pkg/metrics/middleware.go
```diff
@@ -25,10 +25,18 @@ var httpRequestsDuration = prometheus.NewHistogramVec(
 	[]string{"method", "route"},
 )
 
+var InitNodeDeploymentFailures = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: "init_node_deployment_failures",
```
p0lyn0mial: @mrIncompetent  do we want to define an alert for that metric ?

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/Masterminds/semver"
```
p0lyn0mial: we usually group imports, add a new line so that they are separated from std imports.

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -159,3 +162,33 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc pro
...
+		nd.Spec.Template.Cloud.Hetzner == nil &&
+		nd.Spec.Template.Cloud.VSphere == nil &&
+		nd.Spec.Template.Cloud.Azure == nil {
+		return nil, k8cerrors.NewBadRequest("node deployment needs to have cloud provider data")
+	}
+
+	if nd.Spec.Template.Versions.Kubelet != "" {
+		kubeletVersion, err := semver.NewVersion(nd.Spec.Template.Versions.Kubelet)
+		if err != nil {
+			return nil, k8cerrors.NewBadRequest("failed to parse kubelet version: %v", err)
```
p0lyn0mial: as below.

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -159,3 +162,33 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc pro
...
+		}
+
+		if err = common.EnsureVersionCompatible(controlPlaneVersion, kubeletVersion); err != nil {
+			return nil, k8cerrors.NewBadRequest(err.Error())
```
p0lyn0mial: as below.

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -159,3 +162,33 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc pro
...
+// Validate if the node deployment structure fulfills certain requirements. It returns node deployment with updated
+// kubelet version if it wasn't specified.
+func Validate(nd *apiv1.NodeDeployment, controlPlaneVersion *semver.Version) (*apiv1.NodeDeployment, error) {
+	if nd.Spec.Template.Cloud.Openstack == nil &&
+		nd.Spec.Template.Cloud.Digitalocean == nil &&
+		nd.Spec.Template.Cloud.AWS == nil &&
+		nd.Spec.Template.Cloud.Hetzner == nil &&
+		nd.Spec.Template.Cloud.VSphere == nil &&
+		nd.Spec.Template.Cloud.Azure == nil {
+		return nil, k8cerrors.NewBadRequest("node deployment needs to have cloud provider data")
```
p0lyn0mial: it feels weird to return api errors from a function that is not necessarily called from the API server

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+
+			if err = client.Create(ctx, md); err != nil {
+				glog.V(5).Infof("initial node deployment for cluster %s created: %v", cluster.Name, err)
+				return
+			}
+		}
+
+		if time.Now().After(deadline) {
+			glog.V(5).Info("couldn't create initial node deployment, timed out waiting for cluster to be ready")
+			metrics.InitNodeDeploymentFailures.With(prometheus.Labels{"cluster": cluster.Name}).Add(1)
```
p0lyn0mial: could we also add an event to the cluster saying that we were unable to add a ND for the cluster? In the future the dashboard will display events for a cluster.

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +67,89 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+	defer cancelCtx()
+
+	nd, err := machineresource.Validate(nodeDeployment, cluster.Spec.Version.Semver())
+	if err != nil {
+		glog.V(5).Infof("initial node deployment for cluster %s is not valid: %v", cluster.Name, err)
+		return
+	}
+
+	interval := 10 * time.Second
+	timeout := 5 * time.Minute
```
p0lyn0mial: In the worst case scenario, the end user will not receive any status/info back from the API. He might start to think that a ND will not be added to the cluster and could send a new request. I'd add an event saying that a ND is going to be added to the cluster.

p0lyn0mial (changes requested):

api/pkg/metrics/middleware.go
mrIncompetent: Adding the seed-dc name would be great!

api/pkg/metrics/middleware.go
mrIncompetent: Yes! But we can do that in a follow up in case it would block this PR.
cc @xrstf

api/pkg/handler/v1/cluster/cluster.go
p0lyn0mial: sure, np.

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Makes sense. I will add an issue to track this soon.

maciaszczykm: `this` means adding events related to initial node deployment creation.

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Yes, that's why there is a `TODO` above. Would you like to have it as a part of this PR?

api/pkg/resources/machine/machinedeployment.go
maciaszczykm: True, I will change it.

api/pkg/handler/v1/cluster/cluster.go
p0lyn0mial: ah :) Your call :) I'm fine with a follow-up.

api/pkg/handler/v1/cluster/cluster.go
p0lyn0mial: added an issue https://app.zenhub.com/workspaces/all-kubermatic-58863654794153994b62af76/issues/kubermatic/kubermatic/3019

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: https://github.com/kubermatic/kubermatic/issues/3027

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: https://github.com/kubermatic/kubermatic/issues/3027

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Yep, could be useful for the future.

maciaszczykm:
PTAL @p0lyn0mial

maciaszczykm:
If it will be approved I will try to merge in the evening and then UI part as soon as e2e will pass.

api/pkg/handler/routes_v1.go
```diff
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"github.com/prometheus/client_golang/prometheus"
```
p0lyn0mial: grouping :(

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +68,116 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
 		if err != nil {
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
+
+		// Create the initial node deployment in the background.
+		if req.Body.NodeDeployment != nil {
+			go createInitialNodeDeployment(req.Body.NodeDeployment, newCluster, project, sshKeyProvider, dcs,
```
p0lyn0mial: we could avoid repeating calling `initNodeDeploymentFailures` in `createInitialNodeDeployment` function by doing:
```
go func() {
 defer utilruntime.HandleCrash()
 err := createInitialNodeDeployment(...)
 if err != . il {
   initNodeDeploymentFailures.With(prometheus.Labels{"cluster": cluster.Name, "seed_dc": seedDC}).Add(1)
 }
}()
```

what do you think ?

api/cmd/kubermatic-api/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	}, []string{"code", "method"}),
+	HTTPRequestsDuration: prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "http_request_duration_seconds",
+			Help:    "A histogram of latencies for requests.",
+			Buckets: []float64{.005, .01, .025, .05, 0.1, 0.25, 0.5, 1, 1.25, 1.85, 2, 5},
+		},
+		[]string{"method", "route"},
+	),
+	InitNodeDeploymentFailures: prometheus.NewCounterVec(
```
p0lyn0mial, line 26: could you add an item for creating an alert for the metric ?

api/pkg/handler/routes_v1.go
maciaszczykm: Sorry... Is there any check for GoLand for it?

p0lyn0mial: I'm afraid there is not, let me know once you find one :)

api/cmd/kubermatic-api/metrics.go
maciaszczykm: https://github.com/kubermatic/kubermatic/issues/3041

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Awesome, it improves a lot :)

maciaszczykm:
PTAL @p0lyn0mial

p0lyn0mial:
this lgtm waiting for @mrIncompetent before approval

thanks for the pull :)

api/cmd/kubermatic-api/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+		prometheus.CounterOpts{
+			Name: "init_node_deployment_failures",
```
mrIncompetent: Add `kubermatic_api` to all metric names

api/cmd/kubermatic-api/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		[]string{"cluster", "seed_dc"},
+	),
+}
+
+// registerMetrics registers metrics for the API.
+func registerMetrics() {
+	prometheus.MustRegister(metrics.HTTPRequestsTotal)
+	prometheus.MustRegister(metrics.HTTPRequestsDuration)
+	prometheus.MustRegister(metrics.InitNodeDeploymentFailures)
```
mrIncompetent, line 39: For gauge & counter metrics: Initialize them with 0 within the register func:
For example: `metrics.HTTPRequestsTotal.Add(0)`

api/cmd/kubermatic-api/metrics.go
```diff
@@ -0,0 +1,52 @@
+package main
```
mrIncompetent, line 1: Why not just declare them in the package where they actually get used?

api/pkg/handler/test/hack/hack.go
```diff
@@ -69,3 +70,15 @@ func generateDefaultOicdCfg() *common.OIDCConfiguration {
...
+	return common.ServerMetrics{
+		InitNodeDeploymentFailures: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "init_node_deployment_failures",
+				Help: "The number of times initial node deployment couldn't be created within the timeout",
+			},
+			[]string{"cluster", "seed_dc"},
+		),
+	}
+}
```
mrIncompetent, line 84: Why only partially initializes this ~metric~ struct?

api/cmd/kubermatic-api/metrics.go
maciaszczykm: We wanted to group API's metrics in one file. Isn't that good?

cc @p0lyn0mial

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -56,10 +68,106 @@ func CreateEndpoint(cloudProviders map[string]provider.CloudProvider, projectPro
...
+	clusterProvider provider.ClusterProvider, userInfo *provider.UserInfo) error {
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	defer cancelCtx()
+
+	nd, err := machineresource.Validate(nodeDeployment, cluster.Spec.Version.Semver())
+	if err != nil {
+		return fmt.Errorf("node deployment is not valid: %v", err)
+	}
+
+	err = wait.Poll(5*time.Second, 5*time.Minute, initClusterCondition(ctx, cluster.Name, clusterProvider, userInfo))
```
mrIncompetent: Can we increase the timeout to 30min ?
If you have new worker nodes which need to pull all images, there's a good chance that it will take >5min

maciaszczykm: Of course, I will do it.

api/pkg/handler/test/hack/hack.go
maciaszczykm: So basically having this init inside a function where we also register the metrics and set counters to 0?

api/cmd/kubermatic-api/metrics.go
mrIncompetent: I prefer to just add the metrics in the packages where they are actually needed & register them in an init func.
Like: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/cluster/metrics.go

Then you don't need to do any pluming - You now have a hard dependency and there's no easy way to unit-test those, but its easier from my point of view.
But that's just my opinion - this don't must be changed

api/pkg/handler/v1/cluster/cluster.go
p0lyn0mial: the timeout (5 min) is for waiting for a cluster to become ready, we are not waiting for worker nodes.

api/pkg/handler/test/hack/hack.go
mrIncompetent: Ah wait, i'm an idiot. That's test code. Sorry :man_facepalming:

api/pkg/handler/v1/cluster/cluster.go
mrIncompetent: > the timeout (5 min) is for waiting for a cluster to become ready, we are not waiting for worker nodes.

Misunderstanding:
If the worker nodes of the seed cluster are "fresh" and need to pull the images for the control plane pods, this step can take definitely more than 5min.

api/cmd/kubermatic-api/metrics.go
p0lyn0mial: make sure to update alerts when the name changes.

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: > the timeout (5 min) is for waiting for a cluster to become ready, we are not waiting for worker nodes.

True, perhaps 10 minutes will be a compromise here.

p0lyn0mial: oh, I see, they need to download Kubernetes images. Thanks.

api/cmd/kubermatic-api/metrics.go
maciaszczykm: Hmm, what to do with labels? Leave them empty? Seems like there is no Add() method directly on the metric object.

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Okay, then 30m it is 🙂

api/cmd/kubermatic-api/metrics.go
p0lyn0mial: no strong opinion here, I figured it would be nice to declare them in the main pkg as it improves readability - you immediately know what type of metrics are collected by the API server.

api/cmd/kubermatic-api/metrics.go
mrIncompetent: Ah sh*t.
Haven't noticed that. So my comment is useless, sorry!

api/cmd/kubermatic-api/metrics.go
maciaszczykm: Yep, I did.

api/pkg/handler/v1/cluster/cluster.go
maciaszczykm: Done.

api/cmd/kubermatic-api/metrics.go
maciaszczykm: Okay, thanks, no problem 🙂

p0lyn0mial:
thanks for reviewing 

/lgtm
/approve

api/cmd/kubermatic-api/metrics.go
maciaszczykm: 3 x no strong opinion 😄 

What about keeping it w/o changes and updating if it won't work for us in the long run?

maciaszczykm:
Thank you both for review and tips :)

p0lyn0mial:
@kron4eg could you please tag the pull ?

/hold cancel

config/monitoring/prometheus/rules/kubermatic-master-kubermatic.yaml
```diff
@@ -14,7 +14,7 @@ groups:
     annotations:
       message: Kubermatic API is returning a high rate of HTTP 5xx responses.
       runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubermaticapitoomanyerrors
-    expr: sum(rate(http_requests_total{role="kubermatic-api",code=~"5.."}[5m])) >
+    expr: sum(rate(kubermatic_api_http_requests_total{role="kubermatic-api",code=~"5.."}[5m])) >
```
kron4eg: * this file is generated (there is a notice in the head: "# This file has been generated, do not edit.")
* renaming `http_requests_total` to `kubermatic_api_http_requests_total` is bad idea in general

config/monitoring/prometheus/rules/src/kubermatic-master/kubermatic.yaml
```diff
@@ -18,7 +18,7 @@ groups:
     annotations:
       message: Kubermatic API is returning a high rate of HTTP 5xx responses.
       runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-kubermaticapitoomanyerrors
-    expr: sum(rate(http_requests_total{role="kubermatic-api",code=~"5.."}[5m])) > 0.1
+    expr: sum(rate(kubermatic_api_http_requests_total{role="kubermatic-api",code=~"5.."}[5m])) > 0.1
```
kron4eg: same applies here

config/monitoring/prometheus/rules/kubermatic-master-kubermatic.yaml
p0lyn0mial: could you say why renaming is a bad idea ?

kron4eg: * metrics names is public API and need to be threaded the same way
* this breaks ability to aggregate across all `http_requests_total` from all the services
* it adds nothing (we were able to filter only to kubermatic `http_requests_total{role="kubermatic-api"}` previously just fine)

maciaszczykm: @mrIncompetent are you okay with reverting this?

mrIncompetent: I'm basically just following https://prometheus.io/docs/practices/naming/#metric-names

> A metric name should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. For metrics specific to an application, the prefix is usually the application name itself. Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries

mrIncompetent: But apparently i misinterpreted this one: https://groups.google.com/forum/#!topic/prometheus-developers/XjlOnDCK9qc

kron4eg: yeah,
>Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries

`http_requests_total` is one of those standard metrics that a useful when unified across different services.

maciaszczykm:
Removed the prefix from the metrics names. @kron4eg PTAL.

api/cmd/kubermatic-api/metrics.go
kron4eg: This metric is custom, and probably should get new prefixed name per @mrIncompetent comment.

maciaszczykm:
@kron4eg PTAL

kron4eg (approved):
/lgtm

PR #2998, opened 2019-03-13
add service account binding controller

### LINKED ISSUE #2932: RBAC Controller monitors `user/sa` resources
**Description** Since service accounts are a special type of users we decided to use existing `User` resources for service accounts. The only change is to add an additional field `type` for filtering. The good thing is that binding (binding users to groups) will work unchanged as for regular users. Thus the controller should watch for `user` resources with `type=TODO` and add RBAC Roles/Binding.

zreigz:
**What this PR does / why we need it**: new controller watches SA objects and creates binding to the project.

Fixes #2932

```release-note
add service account binding controller
```

### DIFF
--- a/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
+++ b/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
@@ -104,11 +104,11 @@ func (r *reconcileServiceAccountProjectBinding) ensureServiceAccountProjectBindi
 		if err := r.createBinding(sa, projectName); err != nil {
 			return err
 		}
-		// remove labelGroup from sa
-		delete(sa.Labels, labelGroup)
-		if err := r.Update(r.ctx, sa); err != nil {
-			return err
-		}
+	}
+	// remove labelGroup from sa
+	delete(sa.Labels, labelGroup)
+	if err := r.Update(r.ctx, sa); err != nil {
+		return err
 	}
 	return nil
 }
--- a/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
+++ b/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
@@ -106,10 +106,13 @@ func (r *reconcileServiceAccountProjectBinding) ensureServiceAccountProjectBindi
 		}
 	}
 	// remove labelGroup from sa
-	delete(sa.Labels, labelGroup)
-	if err := r.Update(r.ctx, sa); err != nil {
-		return err
+	if _, ok := sa.Labels[labelGroup]; ok {
+		delete(sa.Labels, labelGroup)
+		if err := r.Update(r.ctx, sa); err != nil {
+			return err
+		}
 	}
+
 	return nil
 }
 
--- a/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
+++ b/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
@@ -3,9 +3,10 @@ package serviceaccount
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/rand"
-	"strings"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 
@@ -100,7 +101,14 @@ func (r *reconcileServiceAccountProjectBinding) ensureServiceAccountProjectBindi
 		}
 	}
 	if !bindingExist {
-		return r.createBinding(sa, projectName)
+		if err := r.createBinding(sa, projectName); err != nil {
+			return err
+		}
+		// remove labelGroup from sa
+		delete(sa.Labels, labelGroup)
+		if err := r.Update(r.ctx, sa); err != nil {
+			return err
+		}
 	}
 	return nil
 }
@@ -109,13 +117,6 @@ func (r *reconcileServiceAccountProjectBinding) createBinding(sa *kubermaticv1.U
 	group, ok := sa.Labels[labelGroup]
 	if !ok {
 		return fmt.Errorf("label %s not found for sa %s", labelGroup, sa.Name)
-	}
-
-	// remove labelGroup from sa
-	delete(sa.Labels, labelGroup)
-	err := r.Update(r.ctx, sa)
-	if err != nil {
-		return err
 	}
 
 	binding := &kubermaticv1.UserProjectBinding{
--- a/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
+++ b/api/pkg/controller/rbac/service-account/service_account_binding_controller.go
@@ -3,9 +3,9 @@ package serviceaccount
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"strings"
-
-	"github.com/golang/glog"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 
@@ -57,8 +57,6 @@ func (r *reconcileServiceAccountProjectBinding) Reconcile(request reconcile.Requ
 }
 
 func (r *reconcileServiceAccountProjectBinding) ensureServiceAccountProjectBinding(saName string) error {
-	glog.V(4).Infof("Reconciling SA binding for %s", saName)
-
 	sa := &kubermaticv1.User{}
 	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: saName}, sa); err != nil {
 		if kerrors.IsNotFound(err) {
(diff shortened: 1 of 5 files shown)


api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	resourceName := request.Name
+	if !strings.HasPrefix(resourceName, "serviceaccount") {
+		return reconcile.Result{}, nil
+	}
+
+	return reconcile.Result{}, r.ensureServiceAccountProjectBinding(resourceName)
+}
+
+func (r *reconcileServiceAccountProjectBinding) ensureServiceAccountProjectBinding(saName string) error {
+	glog.V(4).Infof("Reconciling SA binding for %s", saName)
```
p0lyn0mial: this is redundant and to verbose, we could remove it, wdyt ?

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+
+	sa := &kubermaticv1.User{}
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: saName}, sa); err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil
```
p0lyn0mial, line 64: why don't simply return an error from `Get` ?

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: kubermaticv1.SchemeGroupVersion.String(),
+					Kind:       kubermaticv1.UserKindName,
+					UID:        user.GetUID(),
+					Name:       user.Name,
+				},
+			},
+			Name:   genBindingName(project.Name, user.Spec.ID),
```
p0lyn0mial: name should be random.

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	project := &kubermaticv1.Project{}
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: projectName}, project); err != nil {
+		return err
+	}
+
+	binding := &kubermaticv1.UserProjectBinding{}
+	bindingName := genBindingName(projectName, sa.Spec.ID)
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: bindingName}, binding); err != nil {
```
p0lyn0mial: list all bindings for the given `sa` and `project` from the cache instead.

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		if kerrors.IsNotFound(err) {
+			return r.createBinding(sa, project)
+		}
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconcileServiceAccountProjectBinding) createBinding(user *kubermaticv1.User, project *kubermaticv1.Project) error {
```
p0lyn0mial: nit: change `user` to `sa`

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+
+	binding := &kubermaticv1.UserProjectBinding{}
+	bindingName := genBindingName(projectName, sa.Spec.ID)
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: bindingName}, binding); err != nil {
+		if kerrors.IsNotFound(err) {
+			return r.createBinding(sa, project)
+		}
+		return err
+	}
+
```
p0lyn0mial, line 95: remove `user.Labels[labelGroup]` if exist

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(projectName) == 0 {
+		projectName = sa.GetLabels()[kubermaticv1.ProjectIDLabelKey]
+	}
+
+	if len(projectName) == 0 {
+		return fmt.Errorf("unable to find owing project for the service account = %s", sa.Name)
+	}
+
+	project := &kubermaticv1.Project{}
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: projectName}, project); err != nil {
```
p0lyn0mial: why do we have to retrieve the project ? can we simply use `projectName` ?

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconcileServiceAccountProjectBinding) createBinding(user *kubermaticv1.User, project *kubermaticv1.Project) error {
+	group, ok := user.Labels[labelGroup]
+	if !ok {
+		return fmt.Errorf("label %s not found", labelGroup)
```
p0lyn0mial: please add the name of the sa to the log msg.

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: kubermaticv1.SchemeGroupVersion.String(),
+					Kind:       kubermaticv1.UserKindName,
+					UID:        user.GetUID(),
+					Name:       user.Name,
+				},
+			},
+			Name:   genBindingName(project.Name, user.Spec.ID),
+			Labels: map[string]string{kubermaticv1.ProjectIDLabelKey: project.Name},
```
p0lyn0mial: why do we need the label ?

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+		Spec: kubermaticv1.UserProjectBindingSpec{
+			ProjectID: project.Name,
+			UserEmail: user.Spec.Email,
+			Group:     group,
+		},
+	}
+
+	if err := r.Create(r.ctx, binding); err != nil {
```
p0lyn0mial: change to `return r.Create(r.ctx, binding)`

api/pkg/handler/test/helper.go
```diff
@@ -294,6 +294,29 @@ func GenUser(id, name, email string) *kubermaticapiv1.User {
 	}
 }
 
+// GenServiceAccount generates a Service Account resource
+func GenServiceAccount(id, name, group, projectName string) *kubermaticapiv1.User {
+	return &kubermaticapiv1.User{
```
p0lyn0mial: use `GenUser` method and pass `id, name and email` then add the `group` label

api/pkg/controller/rbac/service-account/service_account_binding_test.go
```diff
@@ -0,0 +1,92 @@
...
+		{
+			name:   "scenario 1: this test creates binding for service account",
+			saName: "serviceaccount-abcd",
+			existingKubermaticObjects: []runtime.Object{
+				test.GenDefaultProject(),
+				test.GenServiceAccount("abcd", "test", "editors", "my-first-project-ID"),
+			},
+			expectedBindingName: "sa-my-first-project-ID-abcd",
+			expectedBinding: &kubermaticv1.UserProjectBinding{
+				ObjectMeta: metav1.ObjectMeta{
```
p0lyn0mial: please use `GenBinding` for example https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/user/user_test.go#L59

api/pkg/controller/rbac/sync_project_binding_controller.go
```diff
@@ -162,16 +162,9 @@ func (r *reconcileSyncProjectBinding) userForBinding(projectBinding *kubermaticv
...
-		if ref.Kind == kubermaticv1.ProjectKindName {
-			var err error
-			projectFromCache := &kubermaticv1.Project{}
-			if err = r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: ref.Name}, projectFromCache); err != nil {
-				return nil, err
-			}
-			return projectFromCache, nil
-		}
+	projectFromCache := &kubermaticv1.Project{}
+	if err := r.Get(r.ctx, client.ObjectKey{Namespace: metav1.NamespaceAll, Name: projectBinding.Spec.ProjectID}, projectFromCache); err != nil {
```
p0lyn0mial, line 166: nice :)

p0lyn0mial (changes requested):

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: removed

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: when SA is deleted then we return nil, otherwise, it will call the Reconcile method again and again

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: this code https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/rbac/sync_resource.go#L29 expects Project type, we use User for OwnerRef

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (r *reconcileServiceAccountProjectBinding) createBinding(sa *kubermaticv1.User, projectName string) error {
+	group, ok := sa.Labels[labelGroup]
+	if !ok {
+		return fmt.Errorf("label %s not found for sa %s", labelGroup, sa.Name)
+	}
+
+	// remove labelGroup from sa
+	delete(sa.Labels, labelGroup)
```
p0lyn0mial: we need to remove labels after creating a binding otherwise this is not reliable, you can move it to `ensureServiceAccountProjectBinding` function

p0lyn0mial (changes requested):

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/pkg/controller/rbac/service-account/service_account_binding_controller.go
```diff
@@ -0,0 +1,143 @@
...
+			break
+		}
+	}
+	if !bindingExist {
+		if err := r.createBinding(sa, projectName); err != nil {
+			return err
+		}
+		// remove labelGroup from sa
+		delete(sa.Labels, labelGroup)
+		if err := r.Update(r.ctx, sa); err != nil {
```
p0lyn0mial, line 111: what if a binding was created but this call failed for some reason ? In that case we never remove the label. Please move it to line `113`.

zreigz: done

PR #3029, opened 2019-03-18
RBAC controller monitors ServiceAccount resources

### LINKED ISSUE #2932: RBAC Controller monitors `user/sa` resources

p0lyn0mial:
**What this PR does / why we need it**: the controller generates RBAC for ServiceAccount (aka. User) resources. At the moment only the owners can create and manipulate SA (mapper.go).

Fixes #2932 

**Release note**:

```
RBAC generator controller adds RBAC Roles/Bindings for ServiceAccount resources.
```


zreigz (approved):
looks very good :)
/lgtm
/approve

PR #3037, opened 2019-03-19
add support for SA creation

### LINKED ISSUE #3034: Add create SA endpoint
**Description** TBD

zreigz:
**What this PR does / why we need it**: new endpoint added for service accounts. It creates a new SA resource.
Fixes: #3034

```release-note
a new endpoint for service account creation: `POST /api/v1/projects/{project_id}/serviceaccounts`
[Action required] user need to set a new parameter `--domain` for a domain name on which the server is deployed
```


api/pkg/provider/kubernetes/sa.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(name) == 0 || len(group) == 0 {
+		return nil, kerrors.NewBadRequest("Service account name and group cannot be empty when creating a new SA resource")
+	}
+
+	uniqueID := rand.String(10)
+	uniqueName := fmt.Sprintf("serviceaccount-%s", uniqueID)
+
+	user := kubermaticv1.User{}
+	user.Name = uniqueName
+	user.Spec.Email = fmt.Sprintf("%s@kubermatic.io", uniqueName)
```
p0lyn0mial: the email is going to end up in the tokens, it would be nice to have a subdomain encoded as well, for example,`@dev.kubermatic.io` or `@cloud.kubermatic.io`. The tokens are not encrypted so if a user doesn't remember where it came from looking at the subdomain could be helpful.

api/pkg/provider/kubernetes/sa.go
```diff
@@ -0,0 +1,61 @@
...
+	user.Spec.ID = uniqueID
+	user.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion: kubermaticv1.SchemeGroupVersion.String(),
+			Kind:       kubermaticv1.ProjectKindName,
+			UID:        project.GetUID(),
+			Name:       project.Name,
+		},
+	}
+	user.Labels = map[string]string{"group": group}
```
p0lyn0mial: now I have realised that `initialGroup` would be a better name for the label :(, do we want to change that ?

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,88 @@
...
+			name:     "scenario 1, create service account `test` for owners group",
+			userInfo: &provider.UserInfo{Email: "john@acme.com", Group: "owners-abcd"},
+			project:  createProject("abcd"),
+			saName:   "test",
+			saGroup:  "owners",
+			existingKubermaticObjects: []runtime.Object{
+				createAuthenitactedUser(),
+				createProject("abcd"),
+			},
+			expectedSA: func() *kubermaticv1.User {
```
p0lyn0mial: could you add a function that generates `SA` ? Similar to `createProject` - although it could have more arguments.

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,88 @@
...
+		name                      string
+		existingKubermaticObjects []runtime.Object
+		project                   *kubermaticv1.Project
+		userInfo                  *provider.UserInfo
+		saName                    string
+		saGroup                   string
+		expectedSA                *kubermaticv1.User
+	}{
+		{
+			name:     "scenario 1, create service account `test` for owners group",
```
p0lyn0mial: this test should fail, `sa` cannot belong to the `owners` group. Since the validation is done in the handler we cold change the group name.

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+
+			sa, err := target.CreateServiceAccount(tc.userInfo, tc.project, tc.saName, tc.saGroup)
+
+			// validate
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if !equality.Semantic.DeepEqual(sa.OwnerReferences, tc.expectedSA.OwnerReferences) {
```
p0lyn0mial: wouldn't be better to compare everything except a name which is autogenerated ?

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,98 @@
...
+		if err != nil {
+			return nil, err
+		}
+		saFromRequest := req.Body
+		project, err := projectProvider.Get(userInfo, req.ProjectID, &provider.ProjectGetOptions{})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+
+		sa, err := serviceAccountProvider.CreateServiceAccount(userInfo, project, saFromRequest.Name, saFromRequest.Group)
```
p0lyn0mial: before creating a `SA` we should check if the name (`saFromRequest`) was not taken.

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,98 @@
...
+		if err != nil {
+			return nil, err
```
p0lyn0mial: change to `return nil, common.KubernetesErrorToHTTPError(err)`

api/pkg/controller/rbac/mapper.go
```diff
@@ -35,6 +35,12 @@ var AllGroupsPrefixes = []string{
 	ViewerGroupNamePrefix,
 }
 
+// ServiceAccountGroupsPrefixes holds a list of groups with prefixes that we will generate RBAC Roles/Binding for service account.
+var ServiceAccountGroupsPrefixes = []string{
```
p0lyn0mial: please move `ServiceAccountGroupsPrefixes` where it's used - API :)

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,88 @@
...
+		saName                    string
+		saGroup                   string
+		expectedSA                *kubermaticv1.User
+	}{
+		{
+			name:     "scenario 1, create service account `test` for owners group",
+			userInfo: &provider.UserInfo{Email: "john@acme.com", Group: "owners-abcd"},
+			project:  createProject("abcd"),
+			saName:   "test",
+			saGroup:  "owners",
```
p0lyn0mial: no `owners` group

api/pkg/handler/v1/serviceaccount/sa_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
+	kubermaticapiv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/test"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/test/hack"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
p0lyn0mial, line 16: grouping :)

api/pkg/handler/v1/serviceaccount/sa.go
p0lyn0mial: group name is wrong, should be `owners-projectID`, use `rbac.GenerateActualGroupNameFor(Name, GroupPrefix)`

p0lyn0mial (changes requested):

zreigz:
@p0lyn0mial PTAL

api/pkg/provider/kubernetes/sa.go
zreigz: done

api/pkg/provider/types.go
```diff
@@ -321,3 +321,9 @@ func DatacenterCloudProviderName(spec *DatacenterSpec) (string, error) {
 	}
 	return clouds[0], nil
 }
+
+// ServiceAccountProvider declares the set of methods for interacting with kubermatic service account
+type ServiceAccountProvider interface {
+	CreateServiceAccount(userInfo *UserInfo, project *kubermaticv1.Project, name, group string) (*kubermaticv1.User, error)
```
p0lyn0mial: change to `Create`

api/pkg/provider/types.go
```diff
@@ -321,3 +321,9 @@ func DatacenterCloudProviderName(spec *DatacenterSpec) (string, error) {
...
+	GetServiceAccountByNameForProject(userInfo *UserInfo, serviceAccountName, projectName string) (*kubermaticv1.User, error)
```
p0lyn0mial: change to `List` make it generic and add an additional parameter for filtering, for example https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L75

we also need to add a guard to validate if only the owners of a project can `List` `SA`s, for example
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/member.go#L92

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,121 @@
...
+// Validate validates AddReq request
+func (r AddReq) Validate() error {
+	if len(r.ProjectID) == 0 || len(r.Body.Name) == 0 || len(r.Body.Group) == 0 {
+		return k8cerrors.NewBadRequest("the name, project ID and group cannot be empty")
+	}
+
+	isRequestedGroupPrefixValid := false
+	for _, existingGroupPrefix := range serviceAccountGroupsPrefixes {
+		if existingGroupPrefix == r.Body.Group {
+			isRequestedGroupPrefixValid = true
```
p0lyn0mial: nit: you don't need `isRequestedGroupPrefixValid` variable, you could return `nil` as soon as the code enters here.

```
if existingGroupPrefix == r.Body.Group {
  return nil
}

...
return fmt.Errorf("invalid group name %s", r.Body.Group)
}

File: api/pkg/controller/service-account/service_account_binding_controller.go [hunk 14]
```diff
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	label "github.com/kubermatic/kubermatic/api/pkg/provider/kubernetes"
```
p0lyn0mial: could you change the alias, it reads badly `label.ServiceAccountLabelGroup`

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,163 @@
...
+		name                      string
+		existingKubermaticObjects []runtime.Object
+		projectName               string
+		saName                    string
+		userInfo                  *provider.UserInfo
+		expectedSA                *kubermaticv1.User
+		expectedError             string
+	}{
+		{
+			name:        "scenario 2, get existing service account",
```
p0lyn0mial: nit: update the name to `scenario 1`

api/pkg/provider/kubernetes/sa_test.go
```diff
@@ -0,0 +1,163 @@
...
+				if !equality.Semantic.DeepEqual(sa, tc.expectedSA) {
+					t.Fatalf("expected %v got %v", tc.expectedSA, sa)
+				}
+			}
+		})
+	}
+}
+
+func createSA(name, projectName, group string) *kubermaticv1.User {
+	sa := &kubermaticv1.User{}
```
p0lyn0mial: we could use `GenUser(...)` function to emphasise that a `SA` is not really different from a `User`

config/kubermatic/templates/kubermatic-api-dep.yaml
```diff
@@ -34,6 +34,7 @@ spec:
           - -versions=/opt/master-files/versions.yaml
           - -updates=/opt/master-files/updates.yaml
           - -internal-address=0.0.0.0:8085
+          - -domain={{ .Values.kubermatic.domain }}
```
p0lyn0mial, line 37: should we announce it in the release note ? So that customers know they have to pass an additional param.

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// AddEndpoint adds the given service account to the given project
+func AddEndpoint(projectProvider provider.ProjectProvider, serviceAccountProvider provider.ServiceAccountProvider) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(AddReq)
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
+		err := req.Validate()
+		if err != nil {
+			return nil, err
```
p0lyn0mial: make sure we return a `BadRequest` error.

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,121 @@
...
+
+	isRequestedGroupPrefixValid := false
+	for _, existingGroupPrefix := range serviceAccountGroupsPrefixes {
+		if existingGroupPrefix == r.Body.Group {
+			isRequestedGroupPrefixValid = true
+			break
+		}
+	}
+	if !isRequestedGroupPrefixValid {
+		return k8cerrors.NewBadRequest("invalid group name %s", r.Body.Group)
```
p0lyn0mial: if the validation is done in the decoding function then the users gets `BadRequest` since this function will not be run there could you return a normal error (`fmt.Errorf`) and wrap it in the endpoint ? This improves readability

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,121 @@
...
+		// check if service account name is already reserved in the project
+		existingSA, err := serviceAccountProvider.GetServiceAccountByNameForProject(userInfo, saFromRequest.Name, project.Name)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return nil, common.KubernetesErrorToHTTPError(err)
+			}
+		}
+
+		if existingSA != nil {
+			return nil, common.KubernetesErrorToHTTPError(fmt.Errorf("the given name: '%s' for service account already exists", saFromRequest.Name))
```
p0lyn0mial: please return `errors.NewAlreadyExists` instead.

p0lyn0mial (changes requested):

api/pkg/provider/kubernetes/sa.go
zreigz: done

api/pkg/provider/types.go
zreigz: done

api/pkg/provider/types.go
zreigz: done

api/pkg/handler/v1/serviceaccount/sa.go
zreigz: done

api/pkg/controller/service-account/service_account_binding_controller.go
zreigz: done

api/pkg/provider/kubernetes/sa_test.go
zreigz: done

api/pkg/provider/kubernetes/sa_test.go
zreigz: done

config/kubermatic/templates/kubermatic-api-dep.yaml
zreigz: I've updated `release notes`

api/pkg/handler/v1/serviceaccount/sa.go
zreigz: done

api/pkg/handler/v1/serviceaccount/sa.go
zreigz: done

api/pkg/handler/v1/serviceaccount/sa.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/pkg/handler/v1/serviceaccount/sa.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// AddEndpoint adds the given service account to the given project
+func AddEndpoint(projectProvider provider.ProjectProvider, serviceAccountProvider provider.ServiceAccountProvider) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(AddReq)
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
+		err := req.Validate()
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
```
p0lyn0mial: this is `errors.BadRequest`, otherwise we will return `500`

zreigz: done

api/pkg/provider/kubernetes/sa.go
```diff
@@ -0,0 +1,130 @@
...
+
+	}
+	if len(options.ServiceAccountName) == 0 {
+		return resultList, nil
+	}
+
+	filteredList := make([]*kubermaticv1.User, 0)
+	for _, sa := range resultList {
+		if sa.Spec.Name == options.ServiceAccountName {
+			filteredList = append(filteredList, sa)
```
p0lyn0mial, line 125: could we  `break` as soon as we found `sa` :) ?
```
if sa.Spec.Name == options.ServiceAccountName {
			filteredList = append(filteredList, sa)
                         break
}
```

zreigz: done

p0lyn0mial:
@zreigz we need an `OWNER` file for `service-account` controller.

config/kubermatic/templates/kubermatic-api-dep.yaml
mrIncompetent: Please update the docs: https://github.com/kubermatic/docs/tree/master/content/upgrading/2.9_to_2.10

api/pkg/provider/kubernetes/sa.go
mrIncompetent: 2 Things:
You're iterating over `[]*kubermaticv1.User` but your variable is called `sa` -> this seems a bit confusing

If you are returning `filteredList` in this func, but `break` as soon as you added the first item - why do you return a slice instead of a pointer?

p0lyn0mial: > You're iterating over []*kubermaticv1.User but your variable is called sa -> this seems a bit confusing

ServiceAccounts are a special type of users, they have an email address, belong to a project and a group within that project. They are used to access our API just like regular users.

> If you are returning filteredList in this func, but break as soon as you added the first item - why do you return a slice instead of a pointer?

This is a generic `List` method with options for filtering, it's a common patter for all providers, see: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L85

api/pkg/provider/kubernetes/sa.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	serviceAccounts, err := p.serviceAccountLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	resultList := make([]*kubermaticv1.User, 0)
+	for _, sa := range serviceAccounts {
+		if strings.HasPrefix(sa.Name, "serviceaccount") {
```
mrIncompetent, line 93: can we use something more expressive like `kubermatic-api-serviceaccount` as prefix?

api/pkg/provider/kubernetes/sa.go
```diff
@@ -0,0 +1,131 @@
...
+			for _, owner := range sa.GetOwnerReferences() {
+				if owner.APIVersion == kubermaticv1.SchemeGroupVersion.String() && owner.Kind == kubermaticv1.ProjectKindName && owner.Name == project.Name {
+					resultList = append(resultList, sa)
```
mrIncompetent, line 96: This might be simpler & faster if you use a custom indexer:
Example(UID): https://github.com/kube-node/nodeset/blob/master/pkg/controller/nodeset/node/controller.go#L139

api/pkg/provider/kubernetes/sa.go
p0lyn0mial: the name of a service account is `serviceaccount-ID@dev.kubermatic.io` isn't that enough ? Will we have a different type of `SA` in the future ?

mrIncompetent: Oh wasn't aware, thanks for clarification!

api/pkg/provider/kubernetes/sa.go
p0lyn0mial: we need to create an indexer first :) Iterating over a list in memory is quite fast - practical for billions of elements :)

p0lyn0mial: but I agree it looks better :)

api/pkg/provider/kubernetes/sa.go
p0lyn0mial: > You're iterating over []*kubermaticv1.User but your variable is called sa -> this seems a bit confusing

Also we didn't want to create a new type and generate code for it, plus we could reuse already existing code for example for mapping a service account to a group.

mrIncompetent: > ServiceAccounts are a special type of users, they have an email address, belong to a project and a group within that project. They are used to access our API just like regular users.

:+1: 

> This is a generic List method with options for filtering, it's a common patter for all providers, see: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/types.go#L85

Having a list function which accepts options for filtering :+1: 
Considering the exact `name` as filter during a List operating is a questionable design decision.

mrIncompetent:
Waiting for the docs approval before approving

api/pkg/provider/kubernetes/sa.go
p0lyn0mial: > Considering the exact name as filter during a List operating is a questionable design decision.

We need to list all resources to check if the name was not taken, we could extend the interface by adding more methods like `GetServiceAccountByName`/`ListServiceAccountByName` or use a single `List` method with options.

zreigz:
Documentation is ready https://github.com/kubermatic/docs/pull/112
@mrIncompetent PTAL

PR #3038, opened 2019-03-19
change rbac-generator to master-controller-manager

### LINKED ISSUE #2991: Change rbac-generator name for kubermatic-master-controller-manager
**User Story**
Currently, rbac-generator is deployed only on master nodes. It monitors RBAC resources and also SAs.
In the future, it will monitor more resources. The current name doesn't fit and we would like to change it for more generic.

zreigz:
**What this PR does / why we need it**: change rbac-generator to master-controller-manager

Fixes #2991

-->
```release-note
rbac-generator changed to master-controller-manager
```


api/hack/run-master-controller-manager.sh
```diff
@@ -4,12 +4,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-make -C $(dirname $0)/.. rbac-generator
+make -C $(dirname $0)/.. master-controller-manager
```
p0lyn0mial, line 7: update `Makefile` as well, https://github.com/kubermatic/kubermatic/blob/master/api/Makefile#L80

p0lyn0mial:
also make sure to announce it on `dev`

api/hack/run-master-controller-manager.sh
zreigz: done

api/pkg/controller/master-controller/informer_provider.go
```diff
@@ -1,4 +1,4 @@
-package rbac
+package mastercontroller
```
p0lyn0mial: instead of renaming `rbac` to `mastercontroller` we should take `service-account` controller out. the master controller is created as a binary in cmd/master-controller-manager. So renaming `rbac-generator` binary is enough, WDYT ?

zreigz: agree, I have to revert this change

api/pkg/controller/rbac/user-project-binding/sync_project_binding_controller.go
```diff
@@ -1,10 +1,11 @@
-package rbac
+package userprojectbinding
```
p0lyn0mial, line 1: you can have only one name in `rbac` package, we can move this controller to a separate pkg in another pull.

zreigz:
In next PR I will change CI secret for rbac.replicas and other in values for rbac https://github.com/kubermatic/kubermatic/blob/master/config/kubermatic/values.yaml#L121

mrIncompetent:
/approve

When changing the values.yaml, make sure you update the v2.9-2.10 upgrade docs as well in case you make backwards incompatible changes

PR #3047, opened 2019-03-21
fix Prometheus backup

xrstf:
We often run into issues when creating a Prometheus snapshot. The HTTP call responds with

    {"status":"error","errorType":"internal","error":"create snapshot: snapshot head block: write compaction: found chunk with minTime: 1553171533343 maxTime: 1553173843343 outside of compacted minTime: 1553169600000 maxTime: 1553173840566"}

This has been reported in https://github.com/prometheus/prometheus/issues/5105 and a PR exists since February, but has not yet been merged. The issue mentioned above mentions a workaround that would allow us to create snapshots, even thought they might miss the last few hours of data.

Missing the last few hours of data is better than not having snapshot at all. Until the PR is merged, this is a good bandaid.

**Release note**:
```release-note
improve Prometheus backups in high traffic environments
```


kron4eg (approved):
/lgtm
/approve

PR #3053, opened 2019-03-22
display errors if happened

zreigz:
**What this PR does / why we need it**: currently, all errors are digested and not displayed if happened

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/service-account/service_account_binding_controller.go
+++ b/api/pkg/controller/service-account/service_account_binding_controller.go
@@ -5,13 +5,14 @@ import (
 	"fmt"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/rand"
+	"github.com/golang/glog"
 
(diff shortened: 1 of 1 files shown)


api/pkg/controller/service-account/service_account_binding_controller.go
```diff
@@ -54,7 +55,11 @@ func (r *reconcileServiceAccountProjectBinding) Reconcile(request reconcile.Requ
 		return reconcile.Result{}, nil
 	}
 
-	return reconcile.Result{}, r.ensureServiceAccountProjectBinding(resourceName)
+	if err := r.ensureServiceAccountProjectBinding(resourceName); err != nil {
+		glog.Error(err)
```
p0lyn0mial: we should at least add the name fo the controller to the error msg, there's more than one controller logging to the same "file", wdyt ?

zreigz: yes, agree

zreigz:
@p0lyn0mial PTAL

nikhita (approved):
/lgtm
/approve

PR #3059, opened 2019-03-26
make addon controller respect cluster pause

mrIncompetent:
Make addon controller respect cluster pause.

**Release note**:
```release-note
Addon controller now respects the pause flag on a cluster
```

### DIFF
--- a/api/pkg/controller/addon/addon_controller.go
+++ b/api/pkg/controller/addon/addon_controller.go
@@ -270,6 +270,10 @@ func (c *Controller) sync(key string) error {
 		return nil
 	}
 
+	if cluster.Spec.Pause {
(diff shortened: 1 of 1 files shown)


mrIncompetent:
/assign @themue 

cc @bashofmann

api/pkg/controller/addon/addon_controller.go
```diff
@@ -270,6 +270,10 @@ func (c *Controller) sync(key string) error {
 		return nil
 	}
 
+	if cluster.Spec.Pause {
```
themue, line 274: Would like to see little hint about sense, like for other condition checks. Would also a similar logging make sense?

PR #3062, opened 2019-03-26
Update Elasticsearch chart to 6.6.2

xrstf:
This aligns our chart with helm/stable. To be a little bit more precise, this brings us:

* A mediocre PodSecurityPolicy (because our init containers still require root permissions) and thereby dedicated ServiceAccounts for data and master pods.
* A better configuration for generic ES env variables and more importantly the MINIMUM_MASTER_NODES flag, hopefully preventing split brain situations and keeping clusters on preemtible nodes alive.
* The configuration is now templated instead of relying solely on env variables.
* Node selectors and affinities can be configured.
* We switched to the `-oss` Docker images to get rid of auto-enabled Machine Learning stuff.
* Files inside the volumes are chowned to elasticsearch:elasticsearch instead of elasticsearch:root.
* The node discovery was switched to using the `es-master` service. I don't know why this would be better, but the helm/stable chart does it and so do we now. To be clear, the helm/stable chart has a `es-discovery` service and uses that, but it's just a different name for our es-master service.

**Release note**:
```release-note
update Elasticsearch to 6.6.2
```


kron4eg (approved):
/lgtm
/approve

PR #3064, opened 2019-03-26
increase the conntrack limit

### LINKED ISSUE #3026: Investigate increasing the conntrack limit via kube-proxy
**User Story**
During a recent outage we've encountered that we reached the conntrack limit within the seed cluster.

As this can potentially happen within a user cluster as well, we should check if we should set a higher conntrack limit.

Kube proxy does offer the following flags:
```
--conntrack-max-per-core int32     Default: 32768
    Maximum number of NAT connections to track per CPU core (0 to leave the limit as-is and ignore conntrack-min).
--conntrack-min int32     Default: 131072
    Minimum number of conntrack entries to allocate, regardless of conntrack-max-per-core (set conntrack-max-per-core=0 to leave the limit as-is).
--conntrack-tcp-timeout-close-wait duration     Default: 1h0m0s
    NAT timeout for TCP connections in the CLOSE_WAIT state
--conntrack-tcp-timeout-established duration     Default: 24h0m0s
    Idle timeout for established TCP connections (0 to leave as-is)
```

mrIncompetent:
Increase the conntrack table limit 

Fixes #3026

**Release note**:
```release-note
NONE
```

### DIFF
--- a/addons/kube-proxy/configmap.yaml
+++ b/addons/kube-proxy/configmap.yaml
@@ -19,10 +19,11 @@ data:
     configSyncPeriod: 15m0s
     conntrack:
       max: null
-      maxPerCore: 32768
-      min: 131072
-      tcpCloseWaitTimeout: 1h0m0s
-      tcpEstablishedTimeout: 24h0m0s
+      maxPerCore: 0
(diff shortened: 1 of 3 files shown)


mrIncompetent:
/hold
The addons image must be updated

mrIncompetent:
@thz i updated the image PTAL

PR #3068, opened 2019-03-27
Add uccm healthchecks

### LINKED ISSUE #3019: Report status of UCCM
**Description** The dashboard and the API really on [`Status.Health.AllHealthy`](https://github.com/kubermatic/kubermatic/blob/58a683fe3cf017fd6ef08dd0218b38bcaad3e768/api/pkg/crd/kubermatic/v1/cluster.go#L347]) method to determine if a cluster is ready. For example the dashboard will not allow for adding Node Deployment to unhealthy cluster. Since cluster creation is a complex process having a single method is really nice. Recently we have added `UCCM` controller for setting up various things in the clusters, for example for generating `RBAC` for `NodeDeployments`. The issue is that there is no coordination between `UCCM` and `AllHealthy` method. It can happen that all requests to add a node deployment fail because there are no RBAC Roles/Bindings in place - `UCCM` hasn't finished its job. To fix this issue we need a way to report status of `UCCM` to other components. One way of doing this has been suggested in #2989:

```
Let the UCCM write some checkpoint resource in the cluster which gets checked by the cluster controller. The cluster controller does a "is reachable" check during each sync anyway and report the status back all the way up to "AllHealthy" method.
```

**Note**
Alternatively, we could create a new method Status.Ready() to report "all required set up has been done" as AllHealthy implies a slightly different meaning.

mrIncompetent:
This introduces (Potentially misuses) readiness probes to the user cluster controllers.
The main purpose for those is to communicate the "reconciling state" back to the seed cluster.

When adding the corresponding ReadinessProbes to the UCCM deployment, we now have a way to identify if relevant user cluster controllers finished reconciling at least once (Creating relevant RBAC rules, CRD's, etc.).
This works because the cluster controller waits for all pods to reach the "Ready" phase.

The reason why i currently wrote this is mainly because its really simple and even works when we extend the UCCM to reconcile more resources.
For example:
In the next version we extend the UCCM to reconcile more RBAC resources, which are required by the api+dashboard. 
When using a ConfigMap as checkpoint file, we now would need to introduce a new checkpoint for the new RBAC resources.
With the readiness probe we don't have to extend anything as it only turns ready when the reconciling has finished at least once (Which implies the new resources have been created).

Fixes #3019

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/rbac-user-cluster/controller.go
+++ b/api/pkg/controller/rbac-user-cluster/controller.go
@@ -62,10 +62,10 @@ func Add(mgr manager.Manager, registerReconciledCheck func(name string, check he
 		return err
 	}
 
-	// A very simple but limited way to express the successful reconciling to the seed cluster
-	registerReconciledCheck(fmt.Sprintf("%s-%s", controllerName, "reconciling_successfully"), func() error {
-		if !reconcile.reconcilingSuccessfully {
-			return errors.New("last reconcile failed or did not happen yet")
+	// A very simple but limited way to express the first successful reconciling to the seed cluster
+	registerReconciledCheck(fmt.Sprintf("%s-%s", controllerName, "reconciled_successfully_once"), func() error {
+		if !reconcile.reconciledSuccessfullyOnce {
+			return errors.New("no successful reconciliation so far")
 		}
 		return nil
 	})
@@ -77,7 +77,7 @@ func Add(mgr manager.Manager, registerReconciledCheck func(name string, check he
 type reconcileRBAC struct {
 	ctx context.Context
 	client.Client
-	reconcilingSuccessfully bool
+	reconciledSuccessfullyOnce bool
 }
 
 // Reconcile makes changes in response to Cluster Role and Cluster Role Binding related changes
@@ -86,10 +86,9 @@ func (r *reconcileRBAC) Reconcile(request reconcile.Request) (reconcile.Result, 
 
 	if err := rdr.Reconcile(request.Name); err != nil {
 		glog.Errorf("RBAC reconciliation failed: %v", err)
-		r.reconcilingSuccessfully = false
 		return reconcile.Result{}, err
 	}
-	r.reconcilingSuccessfully = true
 
+	r.reconciledSuccessfullyOnce = true
 	return reconcile.Result{}, nil
 }
--- a/api/pkg/controller/rbac-user-cluster/controller.go
+++ b/api/pkg/controller/rbac-user-cluster/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/golang/glog"
 	"github.com/heptiolabs/healthcheck"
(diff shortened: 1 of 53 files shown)


api/pkg/controller/rbac-user-cluster/controller.go
```diff
@@ -59,23 +61,35 @@ func Add(mgr manager.Manager) error {
 	if err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: mapFn}); err != nil {
 		return err
 	}
+
+	// A very simple but limited way to express the successful reconciling to the seed cluster
+	registerReconciledCheck(fmt.Sprintf("%s-%s", controllerName, "reconciling_successfully"), func() error {
+		if !reconcile.reconcilingSuccessfully {
```
p0lyn0mial: isn't that a race condition ? We have at least two actors, one of them is a `kubelet` trying to probe and the second one is the thread in `Reconcile` method.

api/pkg/controller/rbac-user-cluster/controller.go
```diff
@@ -59,23 +61,35 @@ func Add(mgr manager.Manager) error {
...
 
 // Reconcile makes changes in response to Cluster Role and Cluster Role Binding related changes
 func (r *reconcileRBAC) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	rdr := reconciler{client: r.Client, ctx: r.ctx}
 
-	err := rdr.Reconcile(request.Name)
-	if err != nil {
+	if err := rdr.Reconcile(request.Name); err != nil {
 		glog.Errorf("RBAC reconciliation failed: %v", err)
+		r.reconcilingSuccessfully = false
```
p0lyn0mial: we can reach this line many times, each time preventing the end users from creating clusters even if they could - for example, some transient error.

api/pkg/controller/rbac-user-cluster/controller.go
```diff
@@ -2,10 +2,12 @@ package rbacusercluster
...
 import (
 	"context"
-
-	"k8s.io/apimachinery/pkg/types"
+	"errors"
+	"fmt"
 
 	"github.com/golang/glog"
+	"github.com/heptiolabs/healthcheck"
+	"k8s.io/apimachinery/pkg/types"
```
p0lyn0mial, line 11: grouping :)

api/pkg/controller/rbac-user-cluster/controller.go
mrIncompetent: You mean nodes?

But yes, good point. We should probably not invalidate the flag. 
It should be initialized with false during startup & only be set to true during reconciling. As we only care for the first sync after startup to suceed

api/pkg/controller/rbac-user-cluster/controller.go
mrIncompetent: Got me ;)

api/pkg/controller/rbac-user-cluster/controller.go
mrIncompetent: How could that race?

p0lyn0mial: I think it's equivalent to https://play.golang.org/p/5bh1UftbHbs

two threads are accessing the same variable without synchronisation.
running the app with `-race` flag should blow the app.

api/pkg/controller/rbac-user-cluster/controller.go
p0lyn0mial: yeah, setting the flag only once seems to be a better options in that case.

api/pkg/controller/rbac-user-cluster/controller.go
mrIncompetent: Thanks for the example! I always thought that it only matters for maps/slices.

I added a mutex to guard the access to the variable with https://github.com/kubermatic/kubermatic/pull/3068/commits/6c76131af960fa25febe9defb8c8444718e97213

p0lyn0mial:
I agree, the implementation is simple, extensible and it works  - an elegant hack :)

/lgtm
/approve

nikhita (approved):
/approve

PR #3069, opened 2019-03-27
filter out service accounts from members

zreigz:
**What this PR does / why we need it**: filter out service accounts from members

```release-note
filter out service accounts from the member's list
```

### DIFF
--- a/api/pkg/provider/kubernetes/member.go
+++ b/api/pkg/provider/kubernetes/member.go
@@ -17,11 +17,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+const serviceAccount = "serviceaccount"
+
 // NewProjectMemberProvider returns a project members provider
-func NewProjectMemberProvider(createMasterImpersonatedClient kubermaticImpersonationClient, membersLister kubermaticv1lister.UserProjectBindingLister) *ProjectMemberProvider {
+func NewProjectMemberProvider(createMasterImpersonatedClient kubermaticImpersonationClient, membersLister kubermaticv1lister.UserProjectBindingLister, userLister kubermaticv1lister.UserLister) *ProjectMemberProvider {
 	return &ProjectMemberProvider{
 		createMasterImpersonatedClient: createMasterImpersonatedClient,
 		membersLister:                  membersLister,
+		userLister:                     userLister,
 	}
 }
 
@@ -34,6 +37,9 @@ type ProjectMemberProvider struct {
 
 	// membersLister local cache that stores bindings for members and projects
 	membersLister kubermaticv1lister.UserProjectBindingLister
+
+	// userLister local cache that stores users
+	userLister kubermaticv1lister.UserLister
 }
 
 // Create creates a binding for the given member and the given project
@@ -80,6 +86,14 @@ func (p *ProjectMemberProvider) List(userInfo *provider.UserInfo, project *kuber
 	projectMembers := []*kubermaticapiv1.UserProjectBinding{}
 	for _, member := range allMembers {
 		if member.Spec.ProjectID == project.Name {
+
+			// skip if user is service account type
+			if strings.HasPrefix(member.Spec.UserEmail, serviceAccount) {
+				userEmailParts := strings.Split(member.Spec.UserEmail, "@")
+				if _, err := p.userLister.Get(userEmailParts[0]); err == nil {
+					continue
+				}
+			}
 			projectMembers = append(projectMembers, member)
 		}
 	}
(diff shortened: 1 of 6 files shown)


zreigz:
@p0lyn0mial PTAL

api/pkg/provider/kubernetes/member.go
```diff
@@ -77,9 +79,22 @@ func (p *ProjectMemberProvider) List(userInfo *provider.UserInfo, project *kuber
...
+	}
+
 	projectMembers := []*kubermaticapiv1.UserProjectBinding{}
 	for _, member := range allMembers {
 		if member.Spec.ProjectID == project.Name {
+
+			// skip if user is service account type
+			if strings.HasPrefix(member.Spec.UserEmail, serviceAccount) {
+				userEmailParts := strings.Split(member.Spec.UserEmail, "@")
+				if _, err := masterImpersonatedClient.Users().Get(userEmailParts[0], metav1.GetOptions{}); err == nil {
```
p0lyn0mial: why do we have send `GET` request ?

api/pkg/provider/kubernetes/member_test.go
```diff
@@ -61,81 +62,65 @@ func TestListBinding(t *testing.T) {
...
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "differentProjectBinding",
-					},
-					Spec: kubermaticv1.UserProjectBindingSpec{
-						ProjectID: createProject("abcd").Name,
-						UserEmail: "bob@acme.com",
-						Group:     fmt.Sprintf("owners-%s", createProject("abcd").Name),
-					},
-				},
+				createBinding("abcdBinding", "my-first-project-ID", "bob@acme.com", "owners"),
```
p0lyn0mial, line 73: sweet !

api/pkg/provider/kubernetes/member_test.go
```diff
@@ -61,81 +62,65 @@ func TestListBinding(t *testing.T) {
...
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "cdBinding",
-					},
-					Spec: kubermaticv1.UserProjectBindingSpec{
-						ProjectID: createProject("1234").Name,
-						UserEmail: "bob@acme.com",
-						Group:     fmt.Sprintf("owners-%s", createProject("123").Name),
-					},
-				},
+				createBinding("abcdBinding", "my-first-project-ID", "bob@acme.com", "owners"),
```
p0lyn0mial, line 78: thanks for refactoring this :)

api/pkg/provider/kubernetes/member.go
zreigz: The email address is not enough to be sure if it SA or regular user. `Get` checks if it's a user with an email like SA and it should be displayed or real SA and must be omitted.
The user can have an email like "serviceaccount-bob@loods.com" but it's not SA.

api/pkg/provider/kubernetes/member.go
```diff
@@ -80,6 +86,14 @@ func (p *ProjectMemberProvider) List(userInfo *provider.UserInfo, project *kuber
 	projectMembers := []*kubermaticapiv1.UserProjectBinding{}
 	for _, member := range allMembers {
 		if member.Spec.ProjectID == project.Name {
+
+			// skip if user is service account type
+			if strings.HasPrefix(member.Spec.UserEmail, serviceAccount) {
+				userEmailParts := strings.Split(member.Spec.UserEmail, "@")
```
p0lyn0mial, line 94: should we check the length of `userEmailParts` slice before processing ? The next line can blow if the slice is empty.

api/pkg/provider/kubernetes/member.go
```diff
@@ -80,6 +86,14 @@ func (p *ProjectMemberProvider) List(userInfo *provider.UserInfo, project *kuber
 	projectMembers := []*kubermaticapiv1.UserProjectBinding{}
 	for _, member := range allMembers {
 		if member.Spec.ProjectID == project.Name {
+
+			// skip if user is service account type
```
p0lyn0mial: please add a comment why we are doing this.

api/pkg/provider/kubernetes/member.go
zreigz: yes, I can add the validation but generally, it shouldn't happen.
The email is generated in the code for SA (should be always correct), for members is already validated:
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/user/user.go#L276

api/pkg/provider/kubernetes/member.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

PR #3071, opened 2019-03-27
Add metrics server entry

mrIncompetent:
This PR adds a static entry for the metrics-server DNS name which gets used from the API server.

The API server does a call to `https://metrics-server.cluster-xxxxxxx.cluster.local:443` which might get forwarded to the user cluster DNS resolver.
We cannot use a trailing dot as its not allowed by the validation.

**Release note**:
```release-note
NONE
```


mrIncompetent:
After doing some more tests on this i found that this does not solve the issue :/

PR #3072, opened 2019-03-27
fix node resources dashboard

xrstf:
This adds two missing recording rules and fixes the disk overview chart because we now have the node name available directly and do not need the old, complicated metric merge logic anymore. Also the old chart was unreadable and would have annoyed customers because it was stacking and limited to 100% -- meaning if three disks are at 40%, they would destroy the chart.

**Release note**:
```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve

PR #3073, opened 2019-03-27
fix kibana to match new elasticsearch chart

xrstf:
After updating Elasticsearch recently, we also must use a matching OSS image for Kibana or it will fail because no _xpack index has been setup.

**Release note**:
```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve

Issue #3022, opened 2019-03-18
Add node overview dashboard for user clusters 


mrIncompetent:
**User Story**
As a admin of Kubermatic i want to have a dashboard which shows the utilization of the worker nodes of individual user clusters.

Basically a clone of the seed cluster's "Nodes overview" dashboard where i could select the cluster to look at.

**Acceptance criteria**
- [ ] A the new dashboard has been added

xrstf:
This would require installing node-exporter into user clusters, hoping that users will not nuke it. Are we sure we want to do that?

mrIncompetent:
Even if users nuke it, the addon controller will reinstall it.

Idea is to add it as optional addon (Always enabled for dev, cloud, run & ci)

PR #3089, opened 2019-04-01
add node-exporter addon

xrstf:
We want to provide node dashboards for the user cluster nodes, so we need to install node-exporter into the user clusters. This PR adds a simple daemonset, identical to our current Helm chart.

Relates to #3022

I am very unsure about the labelling here. Some of our addons are using the deprecated `kubernetes.io/cluster-service: "true"`, some are not. Many are using `k8s-app` labels, even though from a short look it's only the Prometheus Operator guys pushing hard for these labels. The canonical documentation seems to recommend the labels I have applied, but I'm not 100% sure.

On that note, some addons have references to `kubermatic`, which kind of breaks our whitelabelling (no pun intended) guidelines, as this is visible to the customers, no?

**Release note**:
```release-note
In order to provide Grafana dashboards for user cluster resource usage, the node-exporter is now deployed by default as an addon into user clusters.
```


alvaroaleman:
@xrstf Do we also want to enable them by default in the chart? Also should we maybe extend the release note to explain _why_ we add it/What it is needed for?

xrstf:
> Do we also want to enable them by default in the chart?

Good catch, yes.

> Also should we maybe extend the release note to explain why we add it/What it is needed for?

Can do.

alvaroaleman:
/approve

LGTM modulo the addon image tag

PR #3090, opened 2019-04-01
removes a redundant field from resourcesToProcess struct

p0lyn0mial:
**What this PR does / why we need it**: simply removes a redundant field from resourcesToProcess struct

**Release note**:

```
NONE
```


zreigz (approved):
/lgtm
/approve

PR #3094, opened 2019-04-02
move core reconciling functions into own subpackage

mrIncompetent:
Move core reconciling functions into own subpackage.

This is mostly to avoid potential cyclic dependencies (Which i ran into)

**Release note**:
```release-note
NONE
```


mrIncompetent:
I'm not going to bother the seed team folks for this PR

PR #3096, opened 2019-04-02
Openshift userdata plugin

### LINKED ISSUE #3118: OS: Allow deployment on all supported providers
Currently its only possible to create Openshift clusters on Hetzner. We should extend this to allow creating them on all providers the machine-controller supports.

*Acceptance criteria*

* The openshift controller doesn't deploy a fork of the machine-controller but the upstream version
* Openshift clusters can be created on all providers that can be used for Kubernetes clusters

alvaroaleman:

Fixes #3118

**Documentation**:

**Release note**:

```release-note
none
```

/hold


api/cmd/openshift-userdata/main.go
```diff
@@ -0,0 +1,22 @@
+package main
```
themue: Why a different path scheme for this package, not i.e. `api/cmd/userdata/openshift`?

Question: How will it be named as binary? Important for todays way of loading.

alvaroaleman: The main reason is that we have one maketarget for all binaries that names the binary like the directory the main.go is in and I dislike `openshift` as binary name.

This is not the name the binary has after being copied into the machine-controller container, see `pkg/controller/openshift/reddources/machinecontroller.go` for details

alvaroaleman:
@themue This is ready now, PTAL

/hold cancel

api/Makefile
```diff
@@ -1,5 +1,5 @@
 export CGO_ENABLED?=0
-REPO=kubermatic/api
+REPO=quay.io/kubermatic/api
```
themue, line 2: What's the motivation here?

api/hack/run-controller.sh
```diff
@@ -3,6 +3,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -x
```
themue, line 6: Still needed?

api/pkg/controller/openshift/resources/machinecontroller.go
```diff
@@ -1,52 +1,54 @@
...
-			if d.Spec.Template.Spec.Containers[idx].Name == containerName {
-				d.Spec.Template.Spec.Containers[idx].Image = image
-				containerWasFound = true
-				break
+		return name, func(in *appsv1.Deployment) (*appsv1.Deployment, error) {
+			d, err := creator(in)
+			if err != nil {
+				return nil, err
 			}
+			d.Spec.Template.Spec.Volumes = append(d.Spec.Template.Spec.Volumes, corev1.Volume{Name: "userdata-plugins",
```
themue, line 26: That's quite interesting. Havn't yet thought of creating/initialising this way. Have to take a deeper look into the Go docs for the API here.

themue (commented):
Only one question, rest LGTM.

/lgtm

api/Makefile
alvaroaleman: We don't use the dockerhub repo anymore

api/hack/run-controller.sh
alvaroaleman: Yes, because there is a long time of no feedback at all when something goes wrong during startup and this is not set

alvaroaleman:
@themue Can you `/approve`? :)

themue:
Sure, after these two helpful answers.

/approve

PR #3086, opened 2019-03-30
Move user cluster resource handling from the cluster controller to the UCCM

mrIncompetent:
**What this PR does / why we need it**:
This PR moves almost all user cluster resource handling from the cluster controller to the user cluster controller.

Things which are currently still kept inside the cluster controller:
- OpenVPN client cert
  Let the user cluster controller generate it? Or generate it in the cluster controller and just propagate it via the user cluster controller?

Regardless of the above mentioned open issue, this PR works(According to the tests) and can be merged.

**Release note**:
```release-note
NONE
```

PR #3098, opened 2019-04-02
Migrate the cluster controller to controller-runtime

mrIncompetent:
This PR migrates the cluster controller to the controller-runtime

I'm sorry for the size.

**Release note**:
```release-note
NONE
```


api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -190,348 +160,165 @@ func NewController(
...
-}
-
-func (cc *Controller) updateCluster(name string, modify func(*kubermaticv1.Cluster)) (updatedCluster *kubermaticv1.Cluster, err error) {
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() (err error) {
-		//Get latest version from cache
-		cacheCluster, err := cc.clusterLister.Get(name)
-		if err != nil {
-			return err
-		}
+	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &kubermaticv1.Cluster{}}
```
alvaroaleman, line 158: Are you sure we set the ownerRef everywhere? In the openshift controller I've created a mapping func that enqs the cluster that has the namespace of the object in its status: https://github.com/kubermatic/kubermatic/blob/2d83e9b42620b3cccd586b2c3840acc3fbb93304/api/pkg/controller/openshift/openshift_controller.go#L82

mrIncompetent: Currently we do set the ownerRef on all objects to point to the cluster object (Except for VPA objects - where we use the Deployment/StatefulSet actually).

I though about triggering the Cluster on each event from any object inside the Namespace, but that would increase the amount of events. (I'm thinking about Pod updates mainly)
Though: Going with that approach would prevent potential bugs which could get triggered by a incomplete resource list.

Any strong opinion here?

alvaroaleman: I am not sure if we really set the OwnerRef everywhere which is why I'd prefer to use the namespace approach.

> I though about triggering the Cluster on each event from any object inside the Namespace, but that would increase the amount of events.

How would that even look like?

api/cmd/kubermatic-controller-manager/controllers.go
```diff
@@ -191,6 +175,7 @@ func createBackupController(ctrlCtx *controllerContext) (runner, error) {
 		backupcontroller.NewMetrics(),
 		ctrlCtx.kubermaticClient,
```
alvaroaleman, line 176: Not for this PR, but we could think about getting rid of all occurrences of typed clients :)

api/pkg/controller/backup/backup_controller.go
```diff
@@ -381,50 +382,6 @@ func (c *Controller) sync(key string) error {
 	return c.ensureCronJob(cluster)
 }
 
-type secretData struct {
```
alvaroaleman, line 384: so awesome :)

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -1,51 +1,48 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
+)
+
+const (
+	ControllerName = "kubermatic_kubernetes_controller"
```
alvaroaleman, line 44: Should we move it into a matching folder, maybe in a follow-up to not completely destroy the reviability?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -129,43 +110,25 @@ func NewController(
...
-		dynamicClient:           dynamicClient,
-		kubermaticClient:        kubermaticClient,
+
+	reconciler := &Reconciler{
+		Client:                  dynamicClient,
 		userClusterConnProvider: userClusterConnProvider,
 		dynamicCache:            dynamicCache,
+		workerName:              workerName,
 
 		queue:    workqueue.NewNamedRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(1*time.Second, 5*time.Minute), "cluster"),
```
alvaroaleman: What for is the queue needed?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -190,348 +153,165 @@ func NewController(
...
-				return err
-			}
-			// Apply modifications
-			modify(currentCluster)
-			// Update the cluster
-			updatedCluster, err = cc.kubermaticClient.KubermaticV1().Clusters().Update(currentCluster)
-			if err != nil {
-				return err
-			}
+	for _, t := range typesToWatch {
```
alvaroaleman, line 172: This is pretty damn neat :)

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -190,348 +153,165 @@ func NewController(
...
-	for i := 0; i < workerCount; i++ {
-		go wait.Until(cc.runWorker, time.Second, stopCh)
+func (r *Reconciler) updateClusterError(ctx context.Context, cluster *kubermaticv1.Cluster, reason kubermaticv1.ClusterStatusError, message string) error {
+	if cluster.Status.ErrorReason == nil || *cluster.Status.ErrorReason != reason {
+		err := r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
+			c.Status.ErrorMessage = &message
+			c.Status.ErrorReason = &reason
+		})
+		if err != nil {
+			return fmt.Errorf("failed to set error status on cluster to: errorReason='%s' errorMessage='%s'. Could not update cluster: %v", reason, message, err)
```
alvaroaleman: You can just use `%q` to achieve the same as the `'%s'`

api/pkg/controller/cluster/pending.go
```diff
@@ -13,118 +16,114 @@ const (
...
+	return &reconcile.Result{}, nil
 }
 
 // ensureClusterNetworkDefaults will apply default cluster network configuration
-func (cc *Controller) ensureClusterNetworkDefaults(c *kubermaticv1.Cluster) (*kubermaticv1.Cluster, error) {
-	var err error
-	if len(c.Spec.ClusterNetwork.Services.CIDRBlocks) == 0 {
-		c, err = cc.updateCluster(c.Name, func(c *kubermaticv1.Cluster) {
+func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	var modifiers []func(*kubermaticv1.Cluster)
```
alvaroaleman, line 101: nice :)))

api/pkg/resources/machine/machine.go
```diff
@@ -59,8 +57,7 @@ func Machine(c *kubermaticv1.Cluster, node *apiv1.Node, dc provider.DatacenterMe
 		// We use OverwriteCloudConfig for Vsphere to ensure we always
 		// use the credentials passed in via frontend for the cloud-provider
 		// functionality
-		templateData := resources.NewTemplateData(c, &dc, "", nil, nil, nil, "", "", "", resource.Quantity{}, "", "", false, false, "", nil, "", "", "", false)
-		overwriteCloudConfig, err := cloudconfig.CloudConfig(templateData)
+		overwriteCloudConfig, err := cloudconfig.CloudConfig(c, &dc)
```
alvaroaleman, line 60: :)))

api/pkg/resources/test/fixtures/deployment-aws-1.10.0-usercluster-controller.yaml
```diff
@@ -54,6 +54,7 @@ spec:
         - name: NAMESPACE
           valueFrom:
             fieldRef:
+              apiVersion: v1
```
alvaroaleman, line 57: Do you know why this is added now?

alvaroaleman (commented):
Such an awesome Pull :)

Regarding the enq for owner vs enq cluster that owns Namespace: Maybe we can leave it as-is and do a follow-up to not set the owner ref on objects inside the cluster NS anymore and then do the enquing via NS?

/lgtm
/approve
/hold

In case you want to do sth about the enq for ns/via ownerRef

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: /lgtm cancel

This one should actually be cleaned up prior to merge

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: Nevermind i'm an idiot.
I'll clean that up in a follow up as suggested. I'll use the code from the openshift controller (Maybe i can add some indexer sugar so we don't have to iterate over all clusters :) )

api/cmd/kubermatic-controller-manager/controllers.go
mrIncompetent: OH YES PLEASE

api/pkg/resources/test/fixtures/deployment-aws-1.10.0-usercluster-controller.yaml
mrIncompetent: Yes, i missed setting this field in https://github.com/kubermatic/kubermatic/pull/3086 (And that field gets defaulted)

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: We used that in the past for some things and i can vaguely remember that it caused some "interesting" escaping. Though i cannot remember and also not replicate in go playground.
I'll update the code accordingly

mrIncompetent: done with https://github.com/kubermatic/kubermatic/pull/3098/commits/a41f0c11b8f3f15a9f67022ad38e720c41dd10b6

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3098/commits/3d22e8abaca082fe8074ae6bd8c49514a6a20ffb

PR #3099, opened 2019-04-02
improve Kibana initialization

### LINKED ISSUE #2064: Document how to configure the logging stack

xrstf:
Initializing Kibana means two things:

1. Create an index-pattern.
2. Set this index-pattern to be the default.

There are two obvious approaches for Initializing an application:

a) Use an initContainer.
b) Use a post-start lifecycle hook.

Previously, we used option b), because it allowed us to access Kibana after it started. Unfortunately, this restricted our toolset to only
things available inside the Kibana Docker image (i.e. no jq).

For 1), there are two options: Use the Kibana API or interact directly with Elasticsearch. Using ES directly *does not work* because even if you create an index pattern manually, Kibana will reject it (presumably because we don't configure all the various fields from the log entries). This means that we **have** to use Kibana's API to create the index pattern. This also means we need a running Kibana process and thereby we can remove approach a (initContainer) from the list of options.

The second task requires us to get the index pattern's ID after it has been created. This means parsing either the document from inside ES or to parse Kibana API's response to our creation request. In both cases, without jq this is just terrible. But we have no jq in the Kibana image, so we cannot perform task 2 inside the post-start lifecycle hook and would need an initContainer with a custom Docker image.

So now we cannot use either approach a or b. This leaves us with a single option: Use a sidecar container. It's kind of misusing the concept of sidecars, but it works. And it allows us to continuously reconcile the configuration by checking if the index pattern still exists and recreating it if it's missing. Kind of like a "mini operator".

cc @alvaroaleman

Fixes #2064

**Release note**:
```release-note
improve Kibana usability by auto-provisioning index patterns
```


config/logging/kibana/config/ensure-index-pattern.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ -n "$DEBUG" ]]; then
+  set -x
+fi
+
+kibana() {
+   curl --silent -H 'kbn-xsrf:anything' "$@"
```
alvaroaleman, line 10: That works everywhere?

config/logging/kibana/values.yaml
```diff
@@ -12,3 +12,16 @@ logging:
         # need more cpu upon initialization, therefore burstable class
         cpu: 1000m
         memory: 600Mi
+
+    setupContainer:
+      image:
+        repository: quay.io/kubermatic/helm
```
kron4eg: a helm image as setupContainer which runs `curl`? is that correct?

xrstf: Yes, I copied this from the check-crd-migration container. And since this container is named `setup-index-pattern`, I put its configuration under `setupContainer` in the values.yaml.

config/logging/kibana/config/ensure-index-pattern.sh
xrstf: This worked when I tested the script. Had it as an alias first, but the `$(...)` subshells did not see my alias. This function worked.

config/logging/kibana/values.yaml
kron4eg: I mean, helm is obviously not used in this container, so why `helm` image?

xrstf: Because I did not want to create a new Docker image for this.

kron4eg: There're plenty of generic images, like alpine or ubuntu or whatever else busybox, using here `kubermatic/helm` is just misleading. I was confused.

xrstf: Yes, of course, but I did not want to use a public image for cases where we want to support offline setups. I know that we do not right now, but we might at some point.

Also, neither alpine nor ubuntu have jq installed and dynamically installing jq on demand when the pod starts is a sure-fire way to break stuff when for some reason an apt/apk repository is unavailable.

xrstf:
@kron4eg I hope you're happy now :)

kron4eg (approved):
/lgtm
/approve

PR #3104, opened 2019-04-03
set oauth settings in OpenShift config

mrIncompetent:
Set OAuth settings in OpenShift config

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/openshift/resources/controlplane_configmap.go
+++ b/api/pkg/controller/openshift/resources/controlplane_configmap.go
@@ -23,7 +23,15 @@ const (
 	openshiftContolPlaneConfigKeyName      = "master-config.yaml"
 )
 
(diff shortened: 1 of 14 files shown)


api/pkg/controller/openshift/resources/controlplane_configmap.go
```diff
@@ -314,30 +329,52 @@ networkConfig:
...
+            - email
+        extraScopes:
+          - email
+          - profile
+        urls:
+          authorize: {{ .OIDC.IssuerURL }}/auth
+          token: {{ .OIDC.IssuerURL }}/token
+  masterCA: /etc/kubernetes/pki/ca/ca.crt
+  assetPublicURL: "{{ .ClusterURL }}/console"
+  masterPublicURL: "{{ .ClusterURL }}"
```
mrIncompetent, line 367: These are settings which we will probably change in the future, but i thought it will be easier to have this ground work in already.

PR #3106, opened 2019-04-03
Do Openshift address reconciliation in Openshift controller

alvaroaleman:

Does the Openshift address reconciliation in the Openshift controller. As a sideeffect, this avoids writing a defunct token into Openshift Cluster CRs

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/cmd/kubermatic-controller-manager/controllers.go
+++ b/api/cmd/kubermatic-controller-manager/controllers.go
@@ -117,10 +117,12 @@ func createOpenshiftController(ctrlCtx *controllerContext) (runner, error) {
 		ctrlCtx.mgr,
 		ctrlCtx.runOptions.workerCount,
 		ctrlCtx.runOptions.workerName,
+		ctrlCtx.runOptions.dc,
 		ctrlCtx.dcs,
 		ctrlCtx.runOptions.overwriteRegistry,
 		ctrlCtx.runOptions.nodeAccessNetwork,
 		ctrlCtx.dockerPullConfigJSON,
+		ctrlCtx.runOptions.externalURL,
 		openshiftcontroller.OIDCConfig{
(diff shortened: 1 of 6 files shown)


api/pkg/controller/openshift/resources/controlplane_configmap.go
```diff
@@ -331,49 +331,49 @@ networkConfig:
...
-  masterURL: "{{ .ClusterURL }}"
-  sessionConfig:
-    sessionMaxAgeSeconds: 300
-    sessionName: ssn
-    sessionSecretsFile: ""
-  templates: null
-  tokenConfig:
-    accessTokenMaxAgeSeconds: 86400
-    authorizeTokenMaxAgeSeconds: 300
+#oauthConfig:
```
mrIncompetent: I've fixed the run scripts so they contain the necessary flags which are used here:
https://github.com/kubermatic/kubermatic/pull/3109

alvaroaleman: Yeah we should discuss what the real solution looks like. Even after Openshift is done, we shouldn't force ppl to use this if they only care about Kubernetes clusters.

WDYT about altering the template to omit the oauth config when the feature is disabled?

alvaroaleman:
@mrIncompetent Would you be fine with merging this if I remove 9bd2a5b3cbc32839fb2c7fff50e1a5c320cafb7e from the Pull?

mrIncompetent:
> @mrIncompetent Would you be fine with merging this if I remove 9bd2a5b from the Pull?

Yes :)

alvaroaleman:
@mrIncompetent Updated and removed the oauth changes, PTAL

PR #3109, opened 2019-04-04
Set oidc flag in hack script

### LINKED ISSUE #3108: Update hack script to contain the OIDC flag
**User Story**
With OpenShift we're going to use Dex right out of the box.
This requires us to always set the OIDC flag (At least during development). Otherwise the OpenShift control plane won't start.

**Acceptance criteria**
- [ ] Hack scripts have been updated to set the OIDC flags

mrIncompetent:
This PR updates our codebase to do certain templating based on the existence of the actual feature flag (`-features=OIDC=true`) and not on the existence of a feature based flag (`-oidc-issuer-url`).
Also this PR updates the hack scripts to contain the OIDC flags.

Fixes #3108

**Release note**:
```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/options.go
+++ b/api/cmd/kubermatic-api/options.go
@@ -69,7 +69,9 @@ func newServerRunOptions() (serverRunOptions, error) {
 }
 
 func (o serverRunOptions) validate() error {
-	// As long as OpenShift support has not reached the testing stage, we only validate those flags when the OIDC feature flag was set
+	// OpenShift always requires those flags, but as long as OpenShift support is not stable/testable
+	// we only validate them when the OIDCKubeCfgEndpoint feature flag is set (Kubernetes specific).
+	// Otherwise we force users to set those flags without any result (for Kubernetes clusters)
 	// TODO: Enforce validation as soon as OpenShift support is testable
 	if o.featureGates.Enabled(OIDCKubeCfgEndpoint) {
 		if len(o.oidcIssuerClientSecret) == 0 {
--- a/api/cmd/kubermatic-api/options.go
+++ b/api/cmd/kubermatic-api/options.go
@@ -69,6 +69,8 @@ func newServerRunOptions() (serverRunOptions, error) {
 }
 
 func (o serverRunOptions) validate() error {
+	// As long as OpenShift support has not reached the testing stage, we only validate those flags when the OIDC feature flag was set
+	// TODO: Enforce validation as soon as OpenShift support is testable
 	if o.featureGates.Enabled(OIDCKubeCfgEndpoint) {
 		if len(o.oidcIssuerClientSecret) == 0 {
 			return fmt.Errorf("%s feature is enabled but \"oidc-client-secret\" flag was not specified", OIDCKubeCfgEndpoint)
--- a/api/cmd/kubermatic-controller-manager/options.go
+++ b/api/cmd/kubermatic-controller-manager/options.go
@@ -133,16 +133,6 @@ func (o controllerRunOptions) validate() error {
 		if len(o.oidcIssuerClientSecret) == 0 {
 			return fmt.Errorf("%s feature is enabled but \"oidc-issuer-client-secret\" flag was not specified", OpenIDAuthPlugin)
 		}
-
-	} else {
-		// don't pass OpenID issuer flags if OpenIDAuthPlugin disabled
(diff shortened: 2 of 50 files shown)


api/cmd/kubermatic-api/options.go
```diff
@@ -69,6 +69,8 @@ func newServerRunOptions() (serverRunOptions, error) {
 }
 
 func (o serverRunOptions) validate() error {
+	// As long as OpenShift support has not reached the testing stage, we only validate those flags when the OIDC feature flag was set
```
alvaroaleman: This comment is really confusing and requires to know that:

* OIDC is not optional for Openshift
* We re-use the same feature gate for both Kubernetes and Openshift

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -133,16 +133,6 @@ func (o controllerRunOptions) validate() error {
 		if len(o.oidcIssuerClientSecret) == 0 {
 			return fmt.Errorf("%s feature is enabled but \"oidc-issuer-client-secret\" flag was not specified", OpenIDAuthPlugin)
 		}
-
-	} else {
```
alvaroaleman, line 137: Why is this removed? I like it if the thing crashes if a config parameter for a feature that is disabled is passed

api/hack/run-api.sh
```diff
@@ -27,5 +27,10 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
   -address=127.0.0.1:8080 \
   -oidc-url=https://dev.kubermatic.io/dex \
   -oidc-authenticator-client-id=kubermatic \
+  -oidc-issuer-url="$(vault kv get -field=oidc-issuer-url dev/seed-clusters/dev.kubermatic.io)" \
```
alvaroaleman: Shouldn't the featuregate be enabled as well or am I missing it?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -82,8 +88,7 @@ type Reconciler struct {
 	oidcIssuerClientID string
 
 	// Temporary feature flags. TODO: Replace with the featureGates map
-	enableVPA                      bool
-	enableEtcdDataCorruptionChecks bool
+	features Features
```
alvaroaleman, line 90: What "featureGates" map is the comment talking about and is that still a TODO after this pull?

api/cmd/kubermatic-api/options.go
mrIncompetent: The feature flag only targets Kubernetes.

For OpenShift we always use those flags to configure the OpenShift internal OAuth server.
That is the reason why we're not validating them for OpenShift yet.
Otherwise administrators would need to set those flags with v2.10 even they don't use the feature

api/cmd/kubermatic-controller-manager/options.go
mrIncompetent: See https://github.com/kubermatic/kubermatic/pull/3109#discussion_r272234536

api/hack/run-api.sh
mrIncompetent: See https://github.com/kubermatic/kubermatic/pull/3109#discussion_r272234536

api/cmd/kubermatic-api/options.go
mrIncompetent: I'll update the comment

mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3109/commits/147e85a270587e40874d4ef524c7d2b1b115c1c3

api/hack/run-api.sh
zreigz: there is no `oidc-issuer-url` in API options. Use 'oidc-url' instead

zreigz (changes requested):

api/hack/run-api.sh
zreigz: generally, this line can be skipped

mrIncompetent: Removed with https://github.com/kubermatic/kubermatic/pull/3109/commits/79b3b49568f3326b469715ba8dc5e28cfb1c0963

mrIncompetent:
@zreigz PTAL

PR #3111, opened 2019-04-04
add endpoint for service account token

### LINKED ISSUE #3107: Add endpoint to create token for SA
**User Story**
Create a new endpoint for token creation.
**Acceptance criteria**
Endpoint generates Secret with JWT token

zreigz:
**What this PR does / why we need it**: add a new endpoint to be able to create a token for the service account

Fixes #3107
Depends on: https://github.com/kubermatic/secrets/pull/250

```release-note
[Action required] the users must set `service-account-signing-key` flag when running the kubermatic-api.
It must be a key with 32 bytes or longer. A user can generate this key using command: "openssl rand -hex 32"
[Action required] The value for the `service-account-signing-key flag must be stored in `auth` section for `kubermtic` in values.yaml, for example:

kubermatic:
  auth:
    serviceAccountKey: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```


zreigz:
don't merge before: https://github.com/kubermatic/secrets/pull/250

api/cmd/kubermatic-api/main.go
```diff
@@ -26,6 +26,8 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/kubermatic/kubermatic/api/pkg/serviceaccount"
```
p0lyn0mial: imports :)

api/pkg/provider/kubernetes/sa_token_test.go
```diff
@@ -0,0 +1,112 @@
...
+		saToSync          *kubermaticv1.User
+		projectToSync     string
+		expectedOwnerRefs []metav1.OwnerReference
+	}{
+		{
+			name:          "scenario 1, create token",
+			userInfo:      &provider.UserInfo{Email: "john@acme.com", Group: "owners-abcd"},
+			saToSync:      createSA("test-1", "my-first-project-ID", "viewers", "1"),
+			projectToSync: "my-first-project-ID",
+			expectedOwnerRefs: []metav1.OwnerReference{
```
p0lyn0mial: wouldn't be better to compare the whole object (`Secret`) ?

api/pkg/provider/kubernetes/sa_token_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			impersonationClient, _, _, err := createFakeKubernetesClients([]runtime.Object{})
+			if err != nil {
+				t.Fatalf("unable to create fake clients, err = %v", err)
+			}
+
+			tokenGenerator, err := serviceaccount.JWTTokenGenerator([]byte(test.TestServiceAccountHashKey))
```
p0lyn0mial: we could inject a fake token generator and we could easily compare expected and actual objects. 

In this test we would like to know if the provider generates a valid object. We could inspect: namespace, OwnerRef, Name and the token (but we could treat the token as an opaque value)

api/pkg/api/v1/types.go
```diff
@@ -203,6 +203,14 @@ type ServiceAccount struct {
 	Group string `json:"group"`
 }
 
+// ServiceAccountToken represent an API service account token
+// swagger:model ServiceAccountToken
+type ServiceAccountToken struct {
+	ObjectMeta
+	// Token the JWT token
+	Token string `json:"token"`
```
p0lyn0mial: this value can be optional, for example it will not be filled on `GET` and `LIST` requests.

api/pkg/provider/kubernetes/sa_token.go
```diff
@@ -0,0 +1,68 @@
...
+			Kind:       kubermaticv1.UserKindName,
+			UID:        sa.GetUID(),
+			Name:       sa.Name,
+		},
+	}
+	secret.Labels = map[string]string{kubermaticv1.ProjectIDLabelKey: projectID}
+	secret.Data = make(map[string][]byte)
+	token, err := p.tokenGenarator.GenerateToken(serviceaccount.Claims(sa.Spec.Email, projectID, secret.Name))
+	if err != nil {
+		return nil, err
```
p0lyn0mial: I think we need to wrap the err into a kubernetes error so that:
 - we stay consistent - all providers return kubernetes errors
 - we can translate kubernetes error into kubermatic api errror (common.KubernetesErrorToHTTPError(err))

api/pkg/provider/kubernetes/sa_token.go
```diff
@@ -0,0 +1,68 @@
...
+func (p *ServiceAccountTokenProvider) Create(userInfo *provider.UserInfo, sa *kubermaticv1.User, projectID string) (*v1.Secret, error) {
+	if userInfo == nil {
+		return nil, kerrors.NewBadRequest("userInfo cannot be nil")
+	}
+	if sa == nil {
+		return nil, kerrors.NewBadRequest("service account cannot be nil")
+	}
+
+	secret := &v1.Secret{}
+	secret.Name = fmt.Sprintf("sa-token-%s", rand.String(10))
```
p0lyn0mial, line 58: so, the users cannot specify a name for tokens ? wouldn't be better to let them specify human readable names ? for example `kubermatic_ci_token`, `ci_e2e_tests`

api/pkg/handler/v1/serviceaccount/token_test.go
```diff
@@ -0,0 +1,98 @@
...
+				t.Fatalf("failed to create test endpoint due to %v", err)
+			}
+
+			ep.ServeHTTP(res, req)
+
+			if res.Code != tc.httpStatus {
+				t.Fatalf("expected HTTP status code %d, got %d: %s", tc.httpStatus, res.Code, res.Body.String())
+			}
+
+			if tc.httpStatus == http.StatusCreated {
```
p0lyn0mial: it always needs to return `StatusCreated`

api/pkg/handler/v1/serviceaccount/token_test.go
```diff
@@ -0,0 +1,98 @@
...
+				var saToken apiv1.ServiceAccountToken
+				err = json.Unmarshal(res.Body.Bytes(), &saToken)
+				if err != nil {
+					t.Fatal(err)
+				}
+
+				tokenAuthenticator := serviceaccount.JWTTokenAuthenticator([]byte(test.TestServiceAccountHashKey))
```
p0lyn0mial: I'd prefer if we could change `CreateTestEndpointAndGetClients` method to return `tokenAuthenticator` instead. WDYT ?

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type TokenGenerator interface {
+	// GenerateToken generates a token which will identify the given
+	// ServiceAccount. privateClaims is an interface that will be
+	// serialized into the JWT payload JSON encoding at the root level of
+	// the payload object. Public claims take precedent over private
+	// claims i.e. if both claims and privateClaims have an "exp" field,
+	// the value in claims will be used.
+	GenerateToken(claims *jwt.Claims, privateClaims *TokenClaim) (string, error)
```
p0lyn0mial: since the interface is `TokenGenerator` we could change the method to `Generate`, wdyt ?

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+type TokenAuthenicator interface {
```
p0lyn0mial: typo in `Authenicator`, should be `Authenticator`

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+	// AuthenticateToken checks given token and transform it to private claim object
+	AuthenticateToken(tokenData string) (*TokenClaim, error)
```
p0lyn0mial: since the interface is `TokenAuthenticator` we could change the method to `Authenticate`, wdyt ?

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+// AuthenticateToken decrypts signed token data to TokenClaim object and checks if token expired
+func (a *jwtTokenAuthenticator) AuthenticateToken(tokenData string) (*TokenClaim, error) {
+
+	tok, err := jwt.ParseSigned(tokenData)
+	if err != nil {
+		return nil, err
+	}
+
+	public := &jwt.Claims{}
+	private := &TokenClaim{}
```
p0lyn0mial: why do you call it `private` ? `TokenClaim` is not encrypted.
could we call them `customClaims`?

api/pkg/api/v1/types.go
```diff
@@ -203,6 +203,14 @@ type ServiceAccount struct {
 	Group string `json:"group"`
 }
 
+// ServiceAccountToken represent an API service account token
+// swagger:model ServiceAccountToken
+type ServiceAccountToken struct {
```
p0lyn0mial, line 218: Should we add an `Expiry` field ?

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if err := tok.Claims(a.key, private, public); err != nil {
+		return nil, err
+	}
+
+	err = public.Validate(jwt.Expected{
```
p0lyn0mial, line 107: shouldn't we use `ValidateWithLeeway` function instead ? what's the difference ?

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,111 @@
...
+		ProjectID: projectID,
+		TokenID:   tokenID,
+	}
+
+	return sc, pc
+}
+
+// JWTTokenGenerator returns a TokenGenerator that generates signed JWT tokens, using the given privateKey.
+func JWTTokenGenerator(privateKey []byte) (TokenGenerator, error) {
+	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.HS256, Key: privateKey}, &jose.SignerOptions{})
```
p0lyn0mial, line 59: what are the requirements for the length of the key ? I think that for `HS256` the key must have `32` bytes in size.

See:
https://tools.ietf.org/html/rfc7518#section-3.2

api/cmd/kubermatic-api/options.go
```diff
@@ -87,18 +91,24 @@ func (o serverRunOptions) validate() error {
 			return fmt.Errorf("%s feature is enabled but \"oidc-issuer-client-id\" flag was not specified", OIDCKubeCfgEndpoint)
 		}
 	}
+
+	if len(o.serviceAccountHashKey) == 0 {
```
p0lyn0mial: should we validate the size of they key ?

api/cmd/kubermatic-api/options.go
```diff
@@ -58,6 +61,7 @@ func newServerRunOptions() (serverRunOptions, error) {
 	flag.BoolVar(&s.oidcIssuerOfflineAccessAsScope, "oidc-issuer-offline-access-as-scope", true, "Set it to false if OIDC provider requires to set \"access_type=offline\" query param when accessing the refresh token")
 	flag.StringVar(&rawFeatureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for various features.")
 	flag.StringVar(&s.domain, "domain", "localhost", "A domain name on which the server is deployed")
+	flag.StringVar(&s.serviceAccountHashKey, "service-account-hash-key", "", "Hash key authenticates the service account's token value using HMAC. It is recommended to use a key with 32 or 64 bytes.")
```
p0lyn0mial: please change the name to `service-account-signing-key`

p0lyn0mial (changes requested):

api/pkg/serviceaccount/jwt.go
zreigz: This call `ValidateWithLeeway` with default DefaultLeeway:
```
// DefaultLeeway defines the default leeway for matching NotBefore/Expiry claims.
DefaultLeeway = 1.0 * time.Minute

func (c Claims) Validate(e Expected) error {
	return c.ValidateWithLeeway(e, DefaultLeeway)
}
```

api/cmd/kubermatic-api/options.go
zreigz: done

api/pkg/serviceaccount/jwt.go
zreigz: it also works when size is different

api/cmd/kubermatic-api/options.go
zreigz: we don't validate other keys

api/pkg/api/v1/types.go
zreigz: done

api/pkg/provider/kubernetes/sa_token.go
zreigz: doen

api/pkg/api/v1/types.go
zreigz: done

api/pkg/provider/kubernetes/sa_token_test.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/pkg/provider/kubernetes/sa_token.go
```diff
@@ -0,0 +1,182 @@
...
+	kubernetesImpersonatedClient, err := createKubernetesImpersonationClientWrapperFromUserInfo(userInfo, p.kubernetesImpersonationClient)
+	if err != nil {
+		return nil, kerrors.NewInternalError(err)
+	}
+
+	return kubernetesImpersonatedClient.CoreV1().Secrets(kubermaticNamespace).Create(secret)
+}
+
+// List gets tokens for the project
+func (p *ServiceAccountTokenProvider) List(userInfo *provider.UserInfo, project *kubermaticv1.Project, sa *kubermaticv1.User, options *provider.ServiceAccountTokenListOptions) ([]*apiv1.PublicServiceAccountToken, error) {
```
p0lyn0mial: return `[]*v1.Secret` instead `[]*apiv1.PublicServiceAccountToken`

api/pkg/provider/kubernetes/sa_token.go
```diff
@@ -0,0 +1,182 @@
...
+
+	kubernetesImpersonatedClient, err := createKubernetesImpersonationClientWrapperFromUserInfo(userInfo, p.kubernetesImpersonationClient)
+	if err != nil {
+		return nil, kerrors.NewInternalError(err)
+	}
+
+	return kubernetesImpersonatedClient.CoreV1().Secrets(kubermaticNamespace).Create(secret)
+}
+
+// List gets tokens for the project
```
p0lyn0mial: please, update the comment - this method gets tokens for the given service account and project

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		token, ok := secret.Data["token"]
+		if !ok {
+			return nil, errors.New(http.StatusInternalServerError, "can not find token data")
+		}
+
+		publicClaim, _, err := serviceAccountTokenProvider.GetTokenAuthenticator().Authenticate(string(token))
+		if err != nil {
+			return nil, errors.New(http.StatusInternalServerError, "can not encode expiration time")
```
p0lyn0mial: please change the error msg `unable to create a token for %s due to %v", saName, err`

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,119 @@
...
+type addTokenReq struct {
+	common.ProjectReq
+	idReq
+	// in: body
+	Body apiv1.ServiceAccountToken
+}
+
+// Validate validates addTokenReq request
+func (r addTokenReq) Validate() error {
+	if len(r.Body.Name) == 0 || len(r.ProjectID) == 0 || len(r.ServiceAccountID) == 0 {
```
p0lyn0mial, line 85: should we also check the max length of the name ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	return req, nil
+}
+
+func convertInternalTokenToExternal(internal *v1.Secret, expiry apiv1.Time) *apiv1.ServiceAccountToken {
+	externalToken := &apiv1.ServiceAccountToken{}
+	externalToken.Expiry = expiry
+	externalToken.ID = internal.Name
+	externalToken.Name = internal.Labels["name"]
```
p0lyn0mial: what if empty ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,119 @@
...
+	externalToken.Token = string(internal.Data["token"])
```
p0lyn0mial: what if empty ?

api/pkg/handler/routes_v1.go
```diff
@@ -1409,6 +1416,34 @@ func (r Routing) deleteServiceAccount() http.Handler {
 	)
 }
 
+// swagger:route POST /api/v1/projects/{project_id}/serviceaccounts/{serviceaccount_id}/tokens tokens addTokenToServiceAccount
+//
+//     Adds the given token to the given service account
```
p0lyn0mial: the EP doesn't add, it generates a token.

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,119 @@
...
+	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/middleware"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/v1/common"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
+	"github.com/kubermatic/kubermatic/api/pkg/util/errors"
+
+	"k8s.io/api/core/v1"
+)
+
+// CreateTokenEndpoint adds the given token to the service account
```
p0lyn0mial: please update the comment, the EP creates a token for the given service account.

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,113 @@
...
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+// Now stubbed out to allow testing
+var Now = time.Now
+
+type TokenGenerator interface {
```
p0lyn0mial, line 15: should we update the name to `JWTTokenGenerator` ? Also please add a comment to the exported type.

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,113 @@
...
+	// Generate generates a token which will identify the given
+	// ServiceAccount. privateClaims is an interface that will be
+	// serialized into the JWT payload JSON encoding at the root level of
+	// the payload object. Public claims take precedent over private
+	// claims i.e. if both claims and privateClaims have an "exp" field,
+	// the value in claims will be used.
+	Generate(claims *jwt.Claims, privateClaims *TokenClaim) (string, error)
+}
+
+type TokenAuthenticator interface {
```
p0lyn0mial, line 26: should we update the name to `JWTTokenAuthenticator` ? Also please add a comment to the exported type.

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,113 @@
...
+	// Authenticate checks given token and transform it to custom claim object
+	Authenticate(tokenData string) (*jwt.Claims, *TokenClaim, error)
+}
+
+type TokenClaim struct {
```
p0lyn0mial: could we change the name to `CustomTokenClaim` ? Also please add a comment to the exported type.

api/pkg/serviceaccount/jwt.go
```diff
@@ -0,0 +1,113 @@
...
+var Now = time.Now
+
+type TokenGenerator interface {
+	// Generate generates a token which will identify the given
+	// ServiceAccount. privateClaims is an interface that will be
+	// serialized into the JWT payload JSON encoding at the root level of
+	// the payload object. Public claims take precedent over private
+	// claims i.e. if both claims and privateClaims have an "exp" field,
+	// the value in claims will be used.
+	Generate(claims *jwt.Claims, privateClaims *TokenClaim) (string, error)
```
p0lyn0mial: change `privateClaims` to `customClaims` and check other places too.

api/pkg/serviceaccount/jwt.go
p0lyn0mial: Sure it works for different sizes but it doesn't mean it's secure. 

In our case the minimum size of the key is `32` bytes (see the spec) - we need to validate it.

the spec says:
```
 key of the same size as the hash output (for instance, 256 bits for
   "HS256") or larger MUST be used with this algorithm.  (This
   requirement is based on Section 5.3.4 (Security Effect of the HMAC
   Key)
```

api/cmd/kubermatic-api/options.go
```diff
@@ -87,18 +91,24 @@ func (o serverRunOptions) validate() error {
 			return fmt.Errorf("%s feature is enabled but \"oidc-issuer-client-id\" flag was not specified", OIDCKubeCfgEndpoint)
 		}
 	}
+
+	if len(o.serviceAccountSigningKey) == 0 {
```
p0lyn0mial: > we don't validate other keys

maybe we should ?

In our case the minimum size of the key is 32 bytes (see the spec) - we need to validate it.

the spec says:

 key of the same size as the hash output (for instance, 256 bits for
   "HS256") or larger MUST be used with this algorithm.  (This
   requirement is based on Section 5.3.4 (Security Effect of the HMAC
   Key)

api/pkg/api/v1/types.go
```diff
@@ -204,6 +204,23 @@ type ServiceAccount struct {
...
+	ObjectMeta
+	// Expiry is a timestamp representing the time when this token will expire.
+	// swagger:strfmt date-time
+	Expiry Time `json:"expiry,omitempty"`
+}
+
+// ServiceAccountToken represent an API service account token
+// swagger:model ServiceAccountToken
+type ServiceAccountToken struct {
+	PublicServiceAccountToken
```
p0lyn0mial, line 219: this is not how `PublicCloudSpec` is implemented, could you add a custom `MarshalJSON` method to `ServiceAccountToken`, just like https://github.com/kubermatic/kubermatic/blob/master/api/pkg/api/v1/types.go#L331

api/cmd/kubermatic-api/options.go
```diff
@@ -58,6 +61,7 @@ func newServerRunOptions() (serverRunOptions, error) {
 	flag.BoolVar(&s.oidcIssuerOfflineAccessAsScope, "oidc-issuer-offline-access-as-scope", true, "Set it to false if OIDC provider requires to set \"access_type=offline\" query param when accessing the refresh token")
 	flag.StringVar(&rawFeatureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for various features.")
 	flag.StringVar(&s.domain, "domain", "localhost", "A domain name on which the server is deployed")
+	flag.StringVar(&s.serviceAccountSigningKey, "service-account-signing-key", "", "Hash key authenticates the service account's token value using HMAC. It is recommended to use a key with 32 or 64 bytes.")
```
p0lyn0mial: please update the description - this is not a `Hash key`, it's a key used for HMAC.

Additionally recommended key size is:
```
key of the same size as the hash output (for instance, 256 bits for
"HS256") or larger MUST be used with this algorithm. (This
requirement is based on Section 5.3.4 (Security Effect of the HMAC
Key)
```

so in our case is `32` or larger.

api/hack/run-api.sh
```diff
@@ -31,5 +31,6 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
   -oidc-issuer-client-secret="$(vault kv get -field=oidc-issuer-client-secret dev/seed-clusters/dev.kubermatic.io)" \
   -oidc-issuer-redirect-uri="$(vault kv get -field=oidc-issuer-redirect-uri dev/seed-clusters/dev.kubermatic.io)" \
   -oidc-issuer-cookie-hash-key="$(vault kv get -field=oidc-issuer-cookie-hash-key dev/seed-clusters/dev.kubermatic.io)" \
+  -service-account-signing-key="$(vault kv get -field=service-account-signing-key dev/seed-clusters/dev.kubermatic.io)" \
```
p0lyn0mial, line 34: will you add a key to vault ?

p0lyn0mial (changes requested):

api/pkg/provider/kubernetes/sa_token.go
zreigz: done

api/pkg/provider/kubernetes/sa_token.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/handler/routes_v1.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/serviceaccount/jwt.go
zreigz: comment added I can't change the name because this name `JWTTokenGenerator` exists in the package for the function

api/pkg/serviceaccount/jwt.go
zreigz: comment added I can't change the name because this name `JWTTokenAuthenticator` exists in the package

api/pkg/serviceaccount/jwt.go
zreigz: done

api/pkg/serviceaccount/jwt.go
zreigz: done

api/cmd/kubermatic-api/options.go
zreigz: done

api/cmd/kubermatic-api/options.go
zreigz: done

api/hack/run-api.sh
zreigz: yes, waiting for review: https://github.com/kubermatic/secrets/pull/250

zreigz:
@p0lyn0mial PTAL

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -0,0 +1,134 @@
...
+	// in: body
+	Body apiv1.ServiceAccountToken
+}
+
+// Validate validates addTokenReq request
+func (r addTokenReq) Validate() error {
+	if len(r.Body.Name) == 0 || len(r.ProjectID) == 0 || len(r.ServiceAccountID) == 0 {
+		return fmt.Errorf("the name, service account ID and project ID cannot be empty")
+	}
+	if len(r.Body.Name) > 50 {
```
p0lyn0mial: the `len` function gives you the number of bytes not individual characters

PR #3112, opened 2019-04-04
Use autodetection for openstack bs version

mrIncompetent:
This PR enables the automatic detection of the BlockStorage API version for OpenStack clusters according to https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage

Blocked by https://github.com/kubermatic/machine-controller/pull/480

**Release note**:
```release-note
Enable automatic detection of the OpenStack BlockStorage API version within the cloud config
```


alvaroaleman:
@mrIncompetent Why does this require kubermatic/machine-controller#480 ?

mrIncompetent:
@alvaroaleman 

> @mrIncompetent Why does this require kubermatic/machine-controller#480 ?

Because we would otherwise deploy the control plane with "auto" & the kubelets would use v2 (As it was hardcoded).
This will lead to differences in environments where v3 is available.

PR #3130, opened 2019-04-05
allow creating machines that need IPAM initialization

### LINKED ISSUE #2956: Allow creating machines that need IPAM initialization
**User Story**

We have an IPAM-Controller that allows to configure a Network on a Cluster and then automatically configures IPs on machines to cover cases where customers do not have DHCP available. In order for this to work, the machines must be created with an Annotation `machine-controller.kubermatic.io/initializers: ipam`.

Right now there is no way to do this, effectively making the feature unavailable.

**Acceptance criteria**

* When a machineDeployment is created and `len(cluster.Spec.MachineNetworks) > 0`, the API must make sure that `machinedeployment.Spec.Template.Metadata.Annotations["machine-controller.kubermatic.io/initializers"]  = "ipam"` is always set

zreigz:
**What this PR does / why we need it**: allow creating machines that need IPAM initialization

Fixes #2956 

```release-note
NONE
```

### DIFF
--- a/api/pkg/resources/machine/machinedeployment.go
+++ b/api/pkg/resources/machine/machinedeployment.go
@@ -44,6 +44,12 @@ func Deployment(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc provider.D
 	md.Spec.Replicas = &nd.Spec.Replicas
 	md.Spec.Template.Spec.Versions.Kubelet = nd.Spec.Template.Versions.Kubelet
 
+	if len(c.Spec.MachineNetworks) > 0 {
+		md.Spec.Template.Annotations = map[string]string{
+			"machine-controller.kubermatic.io/initializers": "ipam",
+		}
+	}
+
 	if nd.Spec.Paused != nil {
 		md.Spec.Paused = *nd.Spec.Paused
 	}


api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -146,6 +146,9 @@ func createInitialNodeDeployment(nodeDeployment *apiv1.NodeDeployment, cluster *
 		return err
 	}
 
+	if len(cluster.Spec.MachineNetworks) > 0 {
+		md.Spec.Template.Annotations["machine-controller.kubermatic.io/initializers"] = "ipam"
```
alvaroaleman: Please make sure the map is initialized, otherwise this will panic

api/pkg/handler/v1/node/node.go
```diff
@@ -110,6 +110,10 @@ func CreateNodeDeployment(sshKeyProvider provider.SSHKeyProvider, projectProvide
 			return nil, fmt.Errorf("failed to create machine deployment from template: %v", err)
 		}
 
+		if len(cluster.Spec.MachineNetworks) > 0 {
+			md.Spec.Template.Annotations["machine-controller.kubermatic.io/initializers"] = "ipam"
```
alvaroaleman: Please make sure the map is initialized, otherwise this will panic

zreigz:
@alvaroaleman now the code is in a better place, where MD is prepared for the creation
PTAL

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -44,6 +44,12 @@ func Deployment(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc provider.D
 	md.Spec.Replicas = &nd.Spec.Replicas
 	md.Spec.Template.Spec.Versions.Kubelet = nd.Spec.Template.Versions.Kubelet
 
+	if len(c.Spec.MachineNetworks) > 0 {
+		md.Spec.Template.Annotations = map[string]string{
```
alvaroaleman, line 48: Can we actually change this to:

```
if md.Spec.Template.Annotations == nil {
  md.Spec.Template.Annotations = map[string]string{}
}
md.Spec.Template.Annotations["machine-controller.kubermatic.io/initializers"] = "ipam"
```

?

The current code would overwrite existing annotations

zreigz: yes, no problem

zreigz: but does it really make sense? It's a new structure https://github.com/kubermatic/kubermatic/blob/7cc53e528fb5d684536c7f4ecd41c16fdfdf5c7e/api/pkg/resources/machine/machinedeployment.go#L26
we don't overwrite existing, we create a new one

alvaroaleman: Oh, you are right, my bad.

/lgtm
/approve

zreigz:
@alvaroaleman thanks for review :)

PR #3138, opened 2019-04-06
Migrate to aggregator routing

mrIncompetent:
This enables the "aggregator routing" on the kube API server.
This instructs the kube apiserver to call services directly via their endpoint IP's. 
As a result the kube API server does not need to have access to the in-cluster DNS.

Downside: User clusters which use GlusterFS or Rook might use a StorageClass which points to a in-cluster DNS name. To inform customers we must prepare a paragraph in our docs to instruct users to use the ClusterIP instead inside the StorageClass.

IMO we're simplifying our codebase + architecture with this change.

**Documentation**:
When we agree on this change i'll create a docs PR.

**Release note**:
```release-note
Enable the aggregator routing of the kube API server & remove the DNS deployments.
```


mrIncompetent:
This also fixes the issue with the broken API server -> metrics-server communication when the VPA kills one of the metrics-server pods ([The API sever would just stop to cleanup the namespaces because it could not connect to the metrics-server.](https://github.com/kubermatic/kubermatic/blob/master/docs/things-that-break.md#metrics-server-not-available)).

cc @thz (I was actually able to reproduce this bug when running the e2e tests against dev.kubermatic.io on a regular basis)

mrIncompetent:
> This also fixes the issue with the broken API server -> metrics-server communication when the VPA kills one of the metrics-server pods (The API sever would just stop to cleanup the namespaces because it could not connect to the metrics-server.).
>
> cc @thz (I was actually able to reproduce this bug when running the e2e tests against dev.kubermatic.io on a regular basis)
Unfortunately there's no metric to actually track this error, but i've build kibana query to search for all occurrences:


Nevermind, this PR does not fix the error:
https://kibana.dev.kubermatic.io/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(_source),index:'0390efe0-5547-11e9-92c8-638c2f4cfa4d',interval:auto,query:(language:lucene,query:'kubernetes.namespace_name:%20%22cluster-%22%20AND%20kubernetes.container_name:%20%22apiserver%22%20AND%20log:%20%22v1beta1.metrics.k8s.io%20failed%20with%22'),sort:!('@timestamp',desc))

Regardless of that bug - it's still a useful change.

alvaroaleman:
> Downside: User clusters which use GlusterFS or Rook might use a StorageClass which points to a in-cluster DNS name. To inform customers we must prepare a paragraph in our docs to instruct users to use the ClusterIP instead inside the StorageClass.

Tbh im a bit torn here. On the one hand less code to achieve the same result is always better, on the other hand breaking things that "just work" in other setups is pretty bad, especially since these things are usually managed by users and not operators of Kubetmatic, that may never read or know about ours docs and even if they do, not understand the problem. Also, fixing this for a Helm chart is hard to impossible, because you can't know the assigned cluster ips at deployment time. Furthermore this may become worse in the future as developing and using weird operators is becoming more common

How do other popular managed offerings like gke and aks handle this?

toschneck:
I see it like Alvaro. If we break such things nobody will care about the reason. User won't care about the "best architectural approach". They want to do as less config/work on there side as possible.

Even if we already know, that it may breaks some user cluster with GlusterFS/Rook, we couldn't be sure that maybe other components we are not aware of also break. 
In my opinion the risk is really high that our customers won't be satisfied with this change.

mrIncompetent:
Closing for above mentioned reasons.

PR #3149, opened 2019-04-09
double memory request for the kubermatic controller manager deployment

mrIncompetent:
Double memory request for the kubermatic controller manager deployment

**Release note**:
```release-note
Double memory request for the Kubermatic controller manager
```


xrstf:
/approve
/lgtm

Thought I wonder there the Github issue for this change is (*cough*, *cough*). Why do we need to increase the limit? I looked at weeks of data for the controller-manager and it never needed that much memory.

PR #3161, opened 2019-04-10
Update the openvpn container

mrIncompetent:
Update the openvpn container

**Release note**:
```release-note
NONE
```

### DIFF
--- a/openshift_addons/release.sh
+++ b/openshift_addons/release.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 set -ex
-export TAG=v0.7-dev1
+export TAG=v0.7
 
 docker build -t quay.io/kubermatic/openshift-addons:${TAG} .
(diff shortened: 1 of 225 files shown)


mrIncompetent:
/hold 

i need to update the images

openshift_addons/release.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 set -ex
-export TAG=v0.6
+export TAG=v0.7-dev1
```
alvaroaleman: Why was this changed?

mrIncompetent: The openvpn container runs as client inside the user cluster. We deploy it via an Addon.
The corresponding addon manifest is changed as well: https://github.com/kubermatic/kubermatic/pull/3161/files#diff-c7c1b998591380695968958c8ad5a1ca

alvaroaleman (commented):
/lgtm
/approve

PR #3173, opened 2019-04-11
Use EnsureObjectV2

mrIncompetent:
Use EnsureObjectV2

**Release note**:
```release-note
NONE
```


api/pkg/controller/monitoring/monitoring_controller.go
```diff
@@ -49,7 +48,6 @@ type Controller struct {
 	kubeClient              kubernetes.Interface
 	dynamicClient           ctrlruntimeclient.Client
 	userClusterConnProvider userClusterConnectionProvider
-	dynamicCache            ctrlruntimecache.Cache
```
mrIncompetent, line 51: We might miss some caches - but that's solves after i migrated this controller to the controller runtime

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -148,7 +148,7 @@ func (r *reconciler) reconcileRoles(ctx context.Context) error {
 		machinecontroller.KubePublicRoleCreator(),
 	}
 
-	if err := reconciling.ReconcileRoles(creators, metav1.NamespacePublic, r.Client, r.cache); err != nil {
+	if err := reconciling.ReconcileRoles(ctx, creators, metav1.NamespacePublic, r); err != nil {
```
alvaroaleman: That does work, but for clarity, can we pass `r.Client` instead?

alvaroaleman:
The build of `Image-loader` fails :(

mrIncompetent:
@alvaroaleman i had to rebase due to conflicts, PTAL

PR #3182, opened 2019-04-11
fix mc in util Docker image

xrstf:
Only the most recent version of mc is available in the top-level directory. This new URL is stable and works for current and older releases.

**Release note**:
```release-note
NONE
```


kron4eg (approved):
/lgtm 
/approve

PR #3184, opened 2019-04-11
fix envoy CLI args

xrstf:
envoy cannot parse foo=bar CLI flags

**Release note**:
```release-note
NONE
```


kron4eg (approved):
/lgtm 
/approve

PR #3185, opened 2019-04-11
removes "serviceaccount-" prefix when returning sa api object

### LINKED ISSUE #3215: remove "serviceaccount-" prefix from the ID field for v1.ServiceAccount resources
**Description**: TBD

p0lyn0mial:
**What this PR does / why we need it**: so that callers can use `api/v1/projects/{id}/serviceaccounts/7d4b5695vb`
instead of `api/v1/projects/{id}/serviceaccounts/serviceaccount-7d4b5695vb`

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): Fixes #3215.

**Release note**:

```
NONE
```

### DIFF
--- a/api/pkg/provider/kubernetes/sa.go
+++ b/api/pkg/provider/kubernetes/sa.go
@@ -14,7 +14,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 )
 
-const ServiceAccountLabelGroup = "initialGroup"
+const (
+	ServiceAccountLabelGroup = "initialGroup"
+	saPrefix                 = "serviceaccount-"
+)
 
 // NewServiceAccountProvider returns a service account provider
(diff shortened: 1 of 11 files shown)


api/pkg/provider/kubernetes/sa.go
```diff
@@ -116,6 +124,11 @@ func (p *ServiceAccountProvider) List(userInfo *provider.UserInfo, project *kube
 		}
 
 	}
+
+	for _, sa := range resultList {
+		sa.Name = removeSAPrefix(sa.Name)
```
zreigz, line 129: to avoid another loop it would be better to add this to line 105 and during validation (line 121) add SAPrefix. WDYT ?

PR #3190, opened 2019-04-12
migrate the monitoring-controller to controller-runtime

### LINKED ISSUE #3121: Migrate the monitoring controller to the controller-runtime
**User Story**
We started using the controller-runtime but we still have some old controllers which don't use it.
To avoid having the same informer multiple times & to reduce boilderplate code we should migrate the monitoring controller to the controller-runtime.

**Acceptance criteria**
- [ ] The monitoring controller uses no typed client

mrIncompetent:
Migrate the monitoring-controller to controller-runtime

Fixes #3121

**Release note**:
```release-note
NONE
```


api/pkg/controller/monitoring/monitoring_controller.go
```diff
@@ -141,238 +115,139 @@ func New(
...
+	c, err := controller.New(ControllerName, mgr, ctrlOptions)
 	if err != nil {
-		runtime.HandleError(fmt.Errorf("couldn't get key for object %#v: %v", cluster, err))
-		return
+		return err
 	}
 
-	c.queue.AddAfter(key, duration)
-}
+	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &kubermaticv1.Cluster{}}
```
alvaroaleman, line 124: This will actually not work with Openshift because the ownerRef is not set on most objects there. Although that probably doesn't matter as this controller doesn't use anything created by the openshift controller?

api/pkg/controller/monitoring/resources.go
```diff
@@ -26,36 +26,36 @@ func (c *Controller) getClusterTemplateData(ctx context.Context, client ctrlrunt
 		client,
 		cluster,
 		&dc,
-		c.dc,
-		c.overwriteRegistry,
-		c.nodePortRange,
-		c.nodeAccessNetwork,
+		r.dc,
```
alvaroaleman, line 29: I would prefer it if the monitoring controller had its own subset of this, it e.G. certainly doesn't need nodePortRage or nodeAccessNetwork. If ppl use this, they expect all fields to contain valid data, which is not the case

api/pkg/controller/monitoring/monitoring_controller.go
mrIncompetent: Should not matter afaik :thinking:

api/pkg/controller/monitoring/resources.go
alvaroaleman: Probably a follow up, thought.

/lgtm 
/approve

PR #3196, opened 2019-04-12
Openshift controller: Handle cluster cleanup

### LINKED ISSUE #3117: OS: Make the cluster controller not do anything on Openshift clusters
In order to keep our code readable and not create weird and by most ppl unknown interdependencies, the cluster controller should work only on Kubernetes clusters and not do anything for Openshift clusters.

*Acceptance criteria*

* The cluster controller doesn't do anything for Openshift clusters
* The cluster controller is renamed to `kubernetes-controller`

alvaroaleman:

Makes the Openshift controller handle cluster deletion, allowing the cluster controller to not be involved at all into Openshift

Fixes #3117 

**Documentation**:

**Release note**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/common/deletion/deletion.go
+++ b/api/pkg/controller/common/deletion/deletion.go
@@ -28,14 +28,14 @@ const (
 	deletedLBAnnotationName = "kubermatic.io/cleaned-up-loadbalancers"
 )
 
-func New(seedClient, userClusterClient controllerruntimeclient.Client) *deletion {
-	return &deletion{
+func New(seedClient, userClusterClient controllerruntimeclient.Client) *Deletion {
+	return &Deletion{
 		seedClient:        seedClient,
 		userClusterClient: userClusterClient,
 	}
 }
 
-type deletion struct {
+type Deletion struct {
 	seedClient        controllerruntimeclient.Client
 	userClusterClient controllerruntimeclient.Client
 }
@@ -43,7 +43,7 @@ type deletion struct {
 const reqeueInterval = 5 * time.Second
 
 // cleanupCluster is responsible for cleaning up a cluster
-func (d *deletion) CleanupCluster(ctx context.Context, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+func (d *Deletion) CleanupCluster(ctx context.Context, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
 	err := d.deletingNodeCleanup(ctx, cluster)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


api/pkg/controller/common/deletion/deletion.go
```diff
@@ -1,4 +1,4 @@
-package cluster
+package deletion
```
alvaroaleman: This is probably not a great package location. Maybe just `pkg/clusterdeletion` instead?

api/pkg/controller/common/deletion/deletion.go
```diff
@@ -21,112 +21,122 @@ import (
...
+	err := d.deletingNodeCleanup(ctx, cluster)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	// If we still have nodes, we must not cleanup other infrastructure at the cloud provider
 	if kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.NodeDeletionFinalizer) {
-		return nil
+		return &reconcile.Result{RequeueAfter: reqeueInterval}, nil
```
alvaroaleman: Alternatively I could maybe just have a wrapping inside `CleanupCluster` that makes sure a `&reconcile.Result{RequeueAfter: requeueInterval}` is always retured?

I mainly didn't want to do this in both the Cluster and Openshift controller

api/pkg/controller/common/deletion/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+const (
+	subsystem = "kubermatic_cluster"
+)
+
+var (
+	registerMetrics sync.Once
+	staleLBs        = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: subsystem,
```
alvaroaleman: This effectively renames the metric, as the subsystem used to be `kubermatic_cluster_controller`. Do you think this will cause issues?

api/pkg/controller/common/deletion/deletion.go
mrIncompetent: I can live with both :)

api/pkg/controller/common/deletion/deletion.go
mrIncompetent: You could check at the root if the cluster still has finalizers & then requeue.
Then you don't need to handle it further down

api/pkg/controller/common/deletion/metrics.go
mrIncompetent: Didn't we have an alert for this ?

alvaroaleman:
@mrIncompetent Tests are passing now, PTAL

PR #3206, opened 2019-04-15
Remove openshift references from values.yaml

alvaroaleman:
**What this PR does / why we need it**: Removes openshift references from the values.yaml as it wont be an official feature in the next release. We should revert this PR after 2.10 got branched  off

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


alvaroaleman:
Hm. This breaks the helm chart becaue the `copy-addons-openshift` container image can not be taken out from the values anymore. Maybe just keep the values file as-is for now and just clean up the docs?

mrIncompetent:
sgtm

PR #3212, opened 2019-04-15
fix Kibana's sidecar container

### LINKED ISSUE #3208: Kibana crashlooping
**User Story**
The kibana sidecar, which ensures that the index exists, crashloops because the usage of an unbound variable.

**Acceptance criteria**
- [ ] The Kibana pods does not crashloop

xrstf:
This makes the Kibana sidecar container not crash anymore due to `DEBUG` not being defined.

Fixes #3208 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/logging/kibana/config/ensure-index-pattern.sh
+++ b/config/logging/kibana/config/ensure-index-pattern.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-if [[ ${DEBUG+x} ]]; then
+if [[ ${DEBUG+false} ]]; then
   set -x
 fi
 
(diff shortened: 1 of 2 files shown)


config/logging/kibana/config/ensure-index-pattern.sh
```diff
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-if [[ -n "$DEBUG" ]]; then
+if [[ ${DEBUG+x} ]]; then
```
kron4eg: probably it will be better to replace this weird trick with more explicit `DEBUG=${DEBUG:-false}`

kron4eg (approved):
not that I'm very excited about whole ensure-index-pattern.sh and `${DEBUG+false}` in particular, but I'll approve this with hope that in the future we would find a better solution to "elastic search schema evolution"

/approve
/lgtm

Issue #2342, opened 2018-11-07
Test idempotency of controllers that create kubernetes objects


alvaroaleman:
**User Story**

As a developer I want my CI system to automatically test as much as possible and give me meaningful feedback. 

**Acceptance criteria**

* There is a test that verifies that the controllers won't update any resources after they got created

**Notes**

May be a good usecase for kind: https://github.com/kubernetes-sigs/kind

mrIncompetent:
Needs more details on the actual impact when things go wrong here

alvaroaleman:
Proposed approach: Add an additional alert just on dev.kubermatic.io on a high PUT request rate of the kubermatic controller manager

PR #3218, opened 2019-04-15
Enable controller-runtime metrics for Kubermatic controller manager

alvaroaleman:

Enables the controller-runtime metrics inside the Kubermatic controller manager. This aids in fixing #2342 as there is a metric for the number of  requests by method, e.G.

```
rest_client_requests_total{code="200",host="35.198.89.109",method="PUT"} 135
```

Unfortunately this means that we cant use the prometheus default Registry anymore, because controller-runtime registers its metrics in various `init()` calls, meaning that even if we replace its Registry with the Prometheus default one, it then wont have its metrics.

Instead I just started using the `ctrlruntime` registry everywhere

Fixes #

**Documentation**:

**Release note**:

```release-note
none
```


alvaroaleman:
cc @xrstf because this will get you a lot of shiny new metrics :)

alvaroaleman:
@mrIncompetent Changed the code to instrument both the default Prometheus registry and the ctrlruntimemetrics registry, PTAL

api/cmd/kubermatic-controller-manager/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	ctrlruntimemetrics.Registry.Unregister(prometheus.NewGoCollector())
+}
+
+// We have our own metrics server implementation that gathers the metrics from
+// both the default prometheus registry and the ctrltuntimemetrics registry.
+// The background is that the latter is not configurable at all and we don't
+// want to force developers into using it, because that is counterintuitive
+// and prone to be forgotten
+type metricsServer struct {
+	gatherers     prometheus.Gatherers
```
mrIncompetent, line 35: :scream: 
I didn't saw that. Nice!

mrIncompetent:
This is a beautiful PR!

/lgtm
/approve

PR #3221, opened 2019-04-15
harmonise chart metadata / define proper versions

xrstf:
This is primarily a spring cleanup:

* `apiVersion` is a mandatory field in Chart.yaml files.
* All charts have a (somewhat) more meaningful version number.
* Maintainer, homepage and sources updated.

---

I propose that we give our charts a meaningful version that we actually update with each revision (on a best-effort basis; kinda not sure how often we would have parallel PRs for the same charts). As Helm versions must be SemVer2 compatible, this means

* no leading "v" prefix
* "a.b.c-d" means "is a pre-release", not "is a patch-release", i.e. `1.0.0 > 1.0.0-1`.

I think this will make a `helm ls` much more meaningful, especially for clusters not permanently kept up-to-date like ours.

**Release note**:
```release-note
improve Helm charts metadata to make Helm-based workflows easier and aid in cluster updates
```

### DIFF
--- a/api/hack/ci/ci-sync-charts.sh
+++ b/api/hack/ci/ci-sync-charts.sh
@@ -6,7 +6,7 @@ apk add --no-cache -U git bash openssh
 cd $(dirname $0)/../../..
 git fetch
 export LATEST_VERSION=$(git describe --tags --abbrev=0)
-sed -i "s/__KUBERMATIC_TAG__/$LATEST_VERSION/g" config/kubermatic/values.yaml
+sed -i "s/__KUBERMATIC_TAG__/$LATEST_VERSION/g" config/kubermatic/values.yaml config/kubermatic/Chart.yaml
 git config --global user.email "dev@loodse.com"
 git config --global user.name "Prow CI Robot"
 git config --global core.sshCommand 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
--- a/config/logging/fluentbit/Chart.yaml
+++ b/config/logging/fluentbit/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentbit
-version: 1.0.4-1
+version: 1.0.0
 appVersion: 1.0.4
 description: fluentbit
 keywords:
--- a/config/logging/fluentbit/Chart.yaml
+++ b/config/logging/fluentbit/Chart.yaml
@@ -1,13 +1,15 @@
+apiVersion: v1
 name: fluentbit
-version: 0.1.0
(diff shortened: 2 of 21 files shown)


api/hack/ci/ci-sync-charts.sh
```diff
@@ -6,7 +6,7 @@ apk add --no-cache -U git bash openssh
 cd $(dirname $0)/../../..
 git fetch
 export LATEST_VERSION=$(git describe --tags --abbrev=0)
-sed -i "s/__KUBERMATIC_TAG__/$LATEST_VERSION/g" config/kubermatic/values.yaml
+sed -i "s/__KUBERMATIC_TAG__/$LATEST_VERSION/g" config/kubermatic/values.yaml config/kubermatic/Chart.yaml
```
xrstf, line 9: I have no idea if this works. `sed --help` mentioned multiple input files are okay.

config/minio/Chart.yaml
```diff
@@ -1,13 +1,16 @@
+apiVersion: v1
 name: minio
-version: 1.0.0
+version: 2019.4.9-1.22.30-1
```
xrstf: I know, this is ugly/cheating, but ... it works I guess.

api/hack/ci/ci-sync-charts.sh
xrstf: Well, works on my machine.

config/minio/Chart.yaml
kron4eg: would it make sent to adopt upstream style of version naming, like latest release is `RELEASE.2019-04-09T01-22-30Z`? Will it even work or helm would just spit an error?

config/logging/fluentbit/Chart.yaml
```diff
@@ -1,13 +1,15 @@
...
+version: 1.0.4-1
+appVersion: 1.0.4
 description: fluentbit
 keywords:
-- k8s
 - kubermatic
+- logging
 - fluentbit
-home: http://www.kubermatic.io/
+home: http://fluentbit.org/
```
kron4eg: this site announces latest release as v0.12.17 made a year ago... Looks like it's abandoned and https://fluentbit.io/ is used instead

config/minio/Chart.yaml
xrstf: Well the Helm docs clearly state that versions must be SemVer 2 and SemVer 2 is pretty strict about its `X.Y.Z-a.b.c+d.e.f` system.

kron4eg (approved):
/lgtm
/approve

PR #3226, opened 2019-04-16
returns a copy of resources from sa.List method

p0lyn0mial:
**What this PR does / why we need it**: do not modify original object that sit in the cache instead modify a copy

See #3215

**Release note**:

```
NONE
```


zreigz (approved):
/lgtm
/approve

PR #3229, opened 2019-04-16
there is no metrics endpoint on the cainjector

xrstf:
This fixes a copy-paste mistake when updating the cert-manager to 0.7.

**Release note**:
```release-note
NONE
```


kron4eg (approved):
/lgtm

ups :))

PR #3239, opened 2019-04-16
Add container linux update operator

### LINKED ISSUE #2080: Provide and configure locksmithd for CoreOS instances
**User Story**
As a cluster user who deploys multiple CoreOS instances, I'd like them to reboot one by one upon a CoreOS update being released, instead of (worst case) all of them at once.

**Acceptance criteria**
[ ] either locksmithd or coreos-update-operator are configured on all CoreOS nodes in user clusters
[ ] the configuration enforces only one node to be rebooted at a time

### LINKED ISSUE #2996: Add ContainerLinuxUpdateOperator
**User Story**
As a user of Kubermatic i want that my ContainerLinux based cluster does coordinated reboots.

The [ContainerLinuxUpdateOperator](https://github.com/coreos/container-linux-update-operator) does exactly that.

Issue though is that user clusters can have mixed nodes, which will make the pods fail if they run on a non ContainerLinux node.

To schedule the pods only on ContainerLinux nodes they need to be labeled. (By default there is no label on Nodes which would enable this)

**Acceptance criteria**
- [ ] Extend the user cluster controller to label all ContainerLinux nodes with the label `kubernetes.io/coreos=true`
- [ ] Add a addon for the ContainerLinuxUpdateOperator

mrIncompetent:
This PR adds the ContainerLinux controller which is responsible for:
- Labeling all Nodes which run ContainerLinux
- Deploy the ContainerLinuxUpdateOperator

When there is no ContainerLinux node in the cluster, the operator won't be deployed.
All operator componentes get deployed with a NodeSelector, so they only run on ContainerLinux nodes.

If the user removes all ContainerLinux nodes, the operator will not be removed - I don't think its worth the effort (Rare edge case + It does not hurt to have those resources).

Fixes #2996
Fixes #2080

**Release note**:
```release-note
Add the ContainerLinuxUpdateOperator to all clusters using ContainerLinux nodes
```

### DIFF
--- a/api/pkg/controller/container-linux/container_linux_controller.go
+++ b/api/pkg/controller/container-linux/container_linux_controller.go
@@ -1,0 +1,167 @@
+package containerlinux
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/container-linux/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName = "kubermatic_container_linux_controller"
+)
+
+type Reconciler struct {
+	ctrlruntimeclient.Client
+	overwriteRegistry string
+}
+
+func Add(mgr manager.Manager, overwriteRegistry string) error {
+
+	reconciler := &Reconciler{
+		Client:            mgr.GetClient(),
+		overwriteRegistry: overwriteRegistry,
+	}
+
+	ctrlOptions := controller.Options{
+		Reconciler: reconciler,
+		// Only use 1 worker to prevent concurrent operator deployments
+		MaxConcurrentReconciles: 1,
+	}
+	c, err := controller.New(ControllerName, mgr, ctrlOptions)
+	if err != nil {
+		return err
+	}
+
+	predicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newNode := e.ObjectNew.(*corev1.Node)
+			// Only sync if the node uses ContainerLinux & is missing the label
+			return isContainerLinuxNode(newNode) && !hasContainerLinuxLabel(newNode)
+		},
+	}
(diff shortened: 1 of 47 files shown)


api/pkg/controller/container-linux/container_linux_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	events := make(chan event.GenericEvent)
+	src := &source.Channel{Source: events, DestBufferSize: 1}
+	if err := c.Watch(src, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+
+	timer := time.NewTimer(syncInterval)
+	go func() {
+		<-timer.C
+		events <- event.GenericEvent{}
```
mrIncompetent: I really did not wanted to sync on each node event.

Syncing on time might not be the cleanest way - but IMO its an acceptable trade-off.
Current interval is `30s`

api/pkg/controller/container-linux/container_linux_controller.go
```diff
@@ -0,0 +1,175 @@
...
+	if err := c.Watch(src, &handler.Funcs{GenericFunc: func(event event.GenericEvent, queue workqueue.RateLimitingInterface) {
+		queue.Add(reconcile.Request{})
+	}}); err != nil {
+		return err
+	}
+
+	// Don't react on node events - There are just too many node events happening
+	// Instead just make sure we create a Watcher for nodes, so we get the cache
+	noOpHandler := &handler.Funcs{}
+	return c.Watch(&source.Kind{Type: &corev1.Node{}}, noOpHandler)
```
alvaroaleman: Why not simply use a predicate function that ingores update events where `oldNode.Status.NodeInfo.OperatingSystem == newNode.Status.NodeInfo.OperatingSystem` , somewhat like this https://github.com/kubermatic/zevenet-lb-controller/blob/cbab0452694f590d7513231c0a22505512b2af7a/pkg/controller/service/service_controller.go#L95 ?

api/pkg/controller/container-linux/resources/clusterrole.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+const (
+	ClusterRoleName = "container_linux_update_operator"
+)
+
+func ClusterRoleCreator() reconciling.NamedClusterRoleCreatorGetter {
```
alvaroaleman, line 13: Soomer or late we should write a CLI that can generate this code from a YAML document....

alvaroaleman:
Please make the PR title a bit more meaningful (`Add container linux update operator`?). 

More of a sidequestion, does the operator drain and taint nodes prior to upgrading them?

api/pkg/controller/container-linux/container_linux_controller.go
mrIncompetent: I've added a update predicate func to check if the node is using ContainerLinux + missing the label.

That should cover it :)

api/pkg/controller/container-linux/resources/clusterrole.go
mrIncompetent: Oh yes.

mrIncompetent:
> More of a sidequestion, does the operator drain and taint nodes prior to upgrading them?

Yes (Except it sets `node.spec.unschedulable=true` - So exactly what `kubectl drain does`)

mrIncompetent:
/retest

@alvaroaleman PTAL

mrIncompetent:
@alvaroaleman i screwed up the naming of the resources...

PR #3247, opened 2019-04-17
add endpoint to edit tokens

### LINKED ISSUE #3204: Add endpoint to edit token for SA
**User Story**
Add a new endpoint to edit the service account token

zreigz:
**What this PR does / why we need it**: new endpoints to edit token for the service account. 

Fixes #3204

```release-note
new endpoints to edit token for the service account:
PUT /api/v1/projects/{project_id}/serviceaccounts/{serviceaccount_id}/tokens/{token_id} - updates and regenerates token data
PATCH /api/v1/projects/{project_id}/serviceaccounts/{serviceaccount_id}/tokens/{token_id} - changes token name
```

### DIFF
--- a/api/pkg/handler/v1/serviceaccount/token.go
+++ b/api/pkg/handler/v1/serviceaccount/token.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"unicode/utf8"
 
@@ -149,8 +150,15 @@ func PatchTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccount
 		if err != nil {
 			return nil, errors.NewBadRequest(err.Error())
 		}
+		tokenReq := &apiv1.PublicServiceAccountToken{}
+		if err := json.Unmarshal(req.Body, tokenReq); err != nil {
+			return nil, errors.NewBadRequest(err.Error())
+		}
+		if len(tokenReq.Name) == 0 {
+			return nil, errors.NewBadRequest("new name can not be empty")
+		}
 
-		secret, err := updateToken(projectProvider, serviceAccountProvider, serviceAccountTokenProvider, userInfo, tokenGenerator, req.ProjectID, req.ServiceAccountID, req.TokenID, req.Body.Name, false)
+		secret, err := updateToken(projectProvider, serviceAccountProvider, serviceAccountTokenProvider, userInfo, tokenGenerator, req.ProjectID, req.ServiceAccountID, req.TokenID, tokenReq.Name, false)
 		if err != nil {
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
@@ -246,17 +254,13 @@ type updateTokenReq struct {
 	Body apiv1.PublicServiceAccountToken
 }
 
-type tokenPatch struct {
-	Name string `json:"name"`
-}
-
 // patchTokenReq defines HTTP request for patchServiceAccountToken
 // swagger:parameters patchServiceAccountToken
 type patchTokenReq struct {
 	commonTokenReq
 	tokenIDReq
 	// in: body
-	Body tokenPatch
+	Body []byte
 }
 
 // Validate validates addTokenReq request
@@ -300,8 +304,8 @@ func (r patchTokenReq) Validate() error {
 	if err := r.commonTokenReq.Validate(); err != nil {
 		return err
 	}
-	if len(r.Body.Name) == 0 {
-		return fmt.Errorf("new name can not be empty")
+	if len(r.Body) == 0 {
+		return fmt.Errorf("body can not be empty")
 	}
 
 	return nil
@@ -384,7 +388,8 @@ func DecodePatchTokenReq(c context.Context, r *http.Request) (interface{}, error
 	req.ServiceAccountID = tokenReq.ServiceAccountID
 	req.ProjectID = tokenReq.ProjectID
 
-	if err := json.NewDecoder(r.Body).Decode(&req.Body); err != nil {
+	req.Body, err = ioutil.ReadAll(r.Body)
+	if err != nil {
 		return nil, err
 	}
 
--- a/api/pkg/handler/v1/serviceaccount/token.go
+++ b/api/pkg/handler/v1/serviceaccount/token.go
@@ -192,7 +192,11 @@ func updateToken(projectProvider provider.ProjectProvider, serviceAccountProvide
 	}
 	existingName, ok := existingSecret.Labels["name"]
 	if !ok {
-		return nil, errors.New(http.StatusInternalServerError, fmt.Sprintf("can not find token name in secret %s", existingSecret.Name))
+		return nil, fmt.Errorf("can not find token name in secret %s", existingSecret.Name)
+	}
+
+	if newName == existingName && !regenerateToken {
+		return existingSecret, nil
 	}
 
 	if newName != existingName {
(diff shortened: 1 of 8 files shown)


api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -130,6 +236,34 @@ type commonTokenReq struct {
...
+type tokenIDReq struct {
+	// in: path
+	TokenID string `json:"token_id"`
+}
+
+// updateTokenReq defines HTTP request for updateServiceAccountToken
+// swagger:parameters updateServiceAccountToken
+type updateTokenReq struct {
+	commonTokenReq
+	tokenIDReq
```
p0lyn0mial, line 256: is `tokenIDReq` different from `commonTokenReq.idReq` ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -115,6 +116,111 @@ func ListTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccountP
...
+		if err != nil {
+			return nil, errors.NewBadRequest(err.Error())
+		}
+
+		existingSecret, err := updateTokenEndpoint(projectProvider, serviceAccountProvider, serviceAccountTokenProvider, userInfo, tokenGenerator, req.ProjectID, req.ServiceAccountID, req.TokenID, req.Body.Name, true)
+		if err != nil {
+			return nil, err
+		}
+
+		secret, err := serviceAccountTokenProvider.Update(userInfo, existingSecret)
```
p0lyn0mial: why not to move it to `updateTokenEndpoint` method ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -115,6 +116,111 @@ func ListTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccountP
...
+		externalToken, err := convertInternalTokenToPublicExternal(secret, tokenAuthenticator)
+		if err != nil {
+			return nil, errors.New(http.StatusInternalServerError, err.Error())
+		}
+
+		return externalToken, nil
+	}
+}
+
+func updateTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccountProvider provider.ServiceAccountProvider,
```
p0lyn0mial: could we change the name to `updateToken` ?

api/pkg/handler/v1/serviceaccount/token.go
zreigz: yes this one is `token_id` second `serviceaccount_id`

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -115,6 +116,111 @@ func ListTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccountP
...
+	serviceAccountTokenProvider provider.ServiceAccountTokenProvider, userInfo *provider.UserInfo, tokenGenerator serviceaccount.TokenGenerator,
+	projectID, saID, tokenID, newName string, regenerateToken bool) (*v1.Secret, error) {
+
+	project, err := projectProvider.Get(userInfo, projectID, &provider.ProjectGetOptions{})
+	if err != nil {
+		return nil, common.KubernetesErrorToHTTPError(err)
```
p0lyn0mial: you we could return `nil, err` and wrap in the handlers ?

api/pkg/handler/v1/serviceaccount/token.go
zreigz: good point, I will

api/pkg/handler/v1/serviceaccount/token.go
p0lyn0mial: should we then update the name of `commonTokenReq.idReq` ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -130,6 +236,34 @@ type commonTokenReq struct {
...
+	TokenID string `json:"token_id"`
+}
+
+// updateTokenReq defines HTTP request for updateServiceAccountToken
+// swagger:parameters updateServiceAccountToken
+type updateTokenReq struct {
+	commonTokenReq
+	tokenIDReq
+	// in: body
+	Body apiv1.ServiceAccountToken
```
p0lyn0mial: shouldn't we accept `v1.PublicServiceAccountToken` ?

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -130,6 +236,34 @@ type commonTokenReq struct {
...
+	Name string `json:"name"`
+}
+
+// patchTokenReq defines HTTP request for patchServiceAccountToken
+// swagger:parameters patchServiceAccountToken
+type patchTokenReq struct {
+	commonTokenReq
+	tokenIDReq
+	// in: body
+	Body tokenPatchName
```
p0lyn0mial: since this is a PATH we should accept any data not only the name, thus I'd change it to `Body []byte` and I'd add a comment saying that at the moment only a name can be updated. WDYT ?

api/pkg/handler/v1/serviceaccount/token.go
zreigz: yes, done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: the struct is self-descriptive and can be easily extended for the new fields.
I suggest staying with structure. I will change for a more generic name like: `tokenPatch` to don't suggest it's only for the name. WDYT?

api/pkg/handler/v1/serviceaccount/token.go
zreigz: yes, sure
done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

api/pkg/handler/v1/serviceaccount/token.go
zreigz: done

zreigz:
@p0lyn0mial PTAL

api/pkg/handler/v1/serviceaccount/token.go
p0lyn0mial: PATH implies that a caller can set whatever it wants, accepting well-defined fields we break this rule. Also changing it to `[]byte` is good for the future because we don't have to coordinate any changes with the clients because the API stays backward compatible.

zreigz: agree, done

api/pkg/handler/v1/serviceaccount/token.go
```diff
@@ -115,6 +117,113 @@ func ListTokenEndpoint(projectProvider provider.ProjectProvider, serviceAccountP
...
+		existingTokenList, err := serviceAccountTokenProvider.List(userInfo, project, sa, &provider.ServiceAccountTokenListOptions{TokenName: newName})
+		if err != nil {
+			return nil, err
+		}
+		if len(existingTokenList) > 0 {
+			return nil, errors.NewAlreadyExists("token", newName)
+		}
+		existingSecret.Labels["name"] = newName
+	}
+
```
p0lyn0mial, line 213: if the name hasn't changed and we don't need to regenerate the token then we should short circuit and return.

zreigz: done

PR #3248, opened 2019-04-17
use controller-runtime leader election

mrIncompetent:
Use controller-runtime leader election.

This also fixes the panic when receiving a signal

**Release note**:
```release-note
NONE
```


mrIncompetent:
Had a discussion with Alvaro about this topic.
It's prefereable to move the migration into the main binary to reduce the complexity.
This requires that we have direct access to the leaderelection code.

Issue #3238, opened 2019-04-16
Master-Seed Monitoring


xrstf:
We want to improve the monitoring experience for multi-seed Kubermatic installations. Specifically, these are the user-facing goals:

## Goals

* Provide master-Grafana with dashboards providing an overview over all seeds.
* Provide a single Alertmanager to view alerts across all seeds. Having the option to also *silence* them from there is a nice-to-have.
* Have alerting continue to work even if the master goes down.
* Allow access to monitoring services (Prometheus UI, Grafana, Alertmanager) even when master is down. Comfort is secondary, i.e. `kubectl port-forward` is fine, as long as the IAP can be circumvented.

### Prometheus

To continue the clear separation between seed and master, I propose installing Prometheus into the master cluster into its own dedicated namespace, `monitoring-master`. This "Master Prometheus" is then responsible for scraping the seed Prometheus pods and generating alerts based on the aggregated seed metrics. It can have a relatively short retention period of maybe 7 days or so.

### Grafana

Likewise, we will install a new Grafana into the `monitoring-master` namespace that has datasources configured for both the Master Prometheus and each Seed Prometheus. Having a datasource per seed makes it possible to re-use the existing dashboards by just switching between the datasources. As need arises, we will probably also define a few dashboards solely for the master cluster.

### Seed Proxy

To establish connections between master and seed, a new "seed-proxy" controller inside the master controller manager will take care of watching the master's kubeconfig/datacenters.yaml and managing pods that run `kubectl proxy`. This will neatly abstract away the nitty gritty details of the connection and leave both Prometheus and Grafana will a relatively straightforward configuration (i.e. `http://seed-<seed name>.kubermatic.svc.cluster.local/api/v1/proxy/...`). Another bonus is that for Prometheus we can simply use its Service Discovery to detect the seeds and don't need to fiddle with Prometheus' needlessly static scraping configs.

### Alertmanager

To provide a single view over all alerts, we will install an Alertmanager into the master. The biggest problem here is that alerts cannot be forwarded from one AM to another. Likewise, silencing an alert inside the master would have no effect on the seed's Alertmanager.

Having all alerts be sent to the master would require a communication channel from the seeds to the master, which does not yet exist. This would also lead to a single point of failure, so we want to avoid this.

[...]

PR #3252, opened 2019-04-17
Multi-Monitoring Stack

xrstf:
This PR allows to install the Prometheus/Grafana/Alertmanager multiple times into the same cluster, as long as it's into different namespaces. This is useful for the future master-monitoring stack discussed in #3238.

At the same time this also changes:

* Dashboard datasources can now be shown as a dropdown, so that when we have a master with multiple seeds, the user can select the datasource (= seed) to look at. This allows us to re-use the same dashboards we already have without adjusting them to run on aggregated data inside the master.
* The provisioning files for Grafana can now be templated.
* The Prometheus chart has a new flag to disable creating the cluster-wide resources. This is not perfect, but seems good enough for our usecase.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/monitoring/prometheus/values.yaml
+++ b/config/monitoring/prometheus/values.yaml
@@ -3,6 +3,10 @@ prometheus:
   host: ''
   storageSize: 100Gi
 
+  # if you install this chart multiple times inside a single
+  # cluster, set this to false on all but one installation
+  createGlobalRoles: true
+
   backup:
     enabled: true
     image:
--- a/config/monitoring/prometheus/values.yaml
+++ b/config/monitoring/prometheus/values.yaml
@@ -3,10 +3,6 @@ prometheus:
   host: ''
   storageSize: 100Gi
 
-  # if you install this chart multiple times inside a single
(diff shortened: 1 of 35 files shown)


config/monitoring/prometheus/values.yaml
```diff
@@ -3,6 +3,10 @@ prometheus:
   host: ''
   storageSize: 100Gi
 
+  # if you install this chart multiple times inside a single
+  # cluster, set this to false on all but one installation
+  createGlobalRoles: true
```
kron4eg: instead of such hacky way, same effect could be accomplished by giving clusterRoles distinct names based on `{{ .Release.Name }}`

kron4eg: obviously, `{{ .Release.Name }}` should be also taken into account in clusterRoleBingings too

kron4eg:
@xrstf seems like something went wrong:

I see such errors
```Error: no ClusterRole with the name "prometheus-kubermatic-prometheus" found``` in deploy logs

PR #3263, opened 2019-04-17
Seed-Proxy Controller for Master Clusters

xrstf:
We want to properly monitor the master cluster and scrape Prometheus inside the seed clusters. To make this easy and automated, this PR adds a new controller to the master-controller-manager that takes care of providing the configuration needed for Prometheus and Grafana to do their thing. Based on the datanceters.yaml and kubeconfig this controller

* provisions a ServiceAccount in each seed cluster,
* creates new roles that only have proxy permissions,
* binds SA to the roles for Grafana and Prometheus,
* copies the token and CA information from the seed into the master and creates a new Secret that holds the ServiceAccount credentials,
* adds a Deployment per seed that runs a single replica of a pod that just runs `kubectl proxy`. This Deployment makes use of the newly created secret from the steps above,
* creates a Service for the kubectl-proxy's opened port (8001),
* and lastly creates a ConfigMap with specialized Grafana config files.

It would have been possible to "simply" configure Prometheus to directly scrape `https://<seed-ip>:6443/api/proxy/monitoring/bla/bla/bla`, but configuring Prometheus dynamically sucks. By having dedicated "proxy pods",  we can instead use Prometheus' service discovery and hide the technical details inside the pods.

To keep the state handling simple, this controller explicitly does not watch resources inside the seeds, i.e. if someone deletes the ServiceAccount in a seed, it won't be reconciled until the reconcile loop is ran again on the master.

**Open Questions**
* A lot of stuff is hardcoded, like the namespace where Grafana is running. Can/should this be turned into CLI flags for the master-controller-manager?
* Given that the controller also creates a Grafana ConfigMap, maybe "seed proxy" is not a good name for the controller?

Relates to #3238

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/crd/client/informers/externalversions"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/metrics"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/signals"
 	"github.com/kubermatic/kubermatic/api/pkg/util/informer"
 	"github.com/kubermatic/kubermatic/api/pkg/util/workerlabel"
@@ -21,12 +22,14 @@ import (
 	kuberinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
 type controllerRunOptions struct {
 	kubeconfig   string
+	dcFile       string
 	masterURL    string
 	internalAddr string
 
@@ -43,6 +46,8 @@ type controllerContext struct {
 	kubeMasterInformerFactory       kuberinformers.SharedInformerFactory
 
 	mgr               manager.Manager
+	kubeconfig        *clientcmdapi.Config
+	datacenters       map[string]provider.DatacenterMeta
 	labelSelectorFunc func(*metav1.ListOptions)
 }
 
@@ -50,6 +55,7 @@ func main() {
 	var g run.Group
 	ctrlCtx := &controllerContext{}
 	flag.StringVar(&ctrlCtx.runOptions.kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	flag.StringVar(&ctrlCtx.runOptions.dcFile, "datacenters", "", "The datacenters.yaml file path")
 	flag.StringVar(&ctrlCtx.runOptions.masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&ctrlCtx.runOptions.workerName, "worker-name", "", "The name of the worker that will only processes resources with label=worker-name.")
 	flag.IntVar(&ctrlCtx.runOptions.workerCount, "worker-count", 4, "Number of workers which process the clusters in parallel.")
@@ -83,6 +89,16 @@ func main() {
 	ctrlCtx.kubermaticMasterInformerFactory = externalversions.NewFilteredSharedInformerFactory(ctrlCtx.kubermaticMasterClient, informer.DefaultInformerResyncPeriod, metav1.NamespaceAll, selector)
 	ctrlCtx.kubeMasterInformerFactory = kuberinformers.NewSharedInformerFactory(ctrlCtx.kubeMasterClient, informer.DefaultInformerResyncPeriod)
 	ctrlCtx.labelSelectorFunc = selector
+
+	ctrlCtx.datacenters, err = provider.LoadDatacentersMeta(ctrlCtx.runOptions.dcFile)
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	ctrlCtx.kubeconfig, err = clientcmd.LoadFromFile(ctrlCtx.runOptions.kubeconfig)
+	if err != nil {
+		glog.Fatal(err)
+	}
 
 	{
 		cfg, err := clientcmd.BuildConfigFromFlags(ctrlCtx.runOptions.masterURL, ctrlCtx.runOptions.kubeconfig)
--- a/api/pkg/controller/seed-proxy/controller.go
+++ b/api/pkg/controller/seed-proxy/controller.go
@@ -75,11 +75,12 @@ func Add(
 		pred predicate.Funcs
 	}
 
+	ownedByPred := makePredicateFuncs(managedByController)
 	typesToWatch := []watcher{
-		{obj: &appsv1.Deployment{}, pred: deploymentPredicate()},
-		{obj: &corev1.Service{}, pred: servicePredicate()},
-		{obj: &corev1.Secret{}, pred: secretsPredicate()},
-		{obj: &corev1.ConfigMap{}, pred: configMapPredicate()},
+		{obj: &appsv1.Deployment{}, pred: ownedByPred},
+		{obj: &corev1.Service{}, pred: ownedByPred},
+		{obj: &corev1.Secret{}, pred: ownedByPred},
+		{obj: &corev1.ConfigMap{}, pred: ownedByPred},
 	}
 
 	for _, t := range typesToWatch {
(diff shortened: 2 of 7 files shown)


api/cmd/master-controller-manager/main.go
```diff
@@ -84,6 +90,16 @@ func main() {
 	ctrlCtx.kubeMasterInformerFactory = kuberinformers.NewSharedInformerFactory(ctrlCtx.kubeMasterClient, informer.DefaultInformerResyncPeriod)
 	ctrlCtx.labelSelectorFunc = selector
 
+	ctrlCtx.datacenters, err = provider.LoadDatacentersMeta(ctrlCtx.runOptions.dcFile)
+	if err != nil {
+		glog.Fatal(err)
+	}
+
+	ctrlCtx.kubeconfig, err = clientcmd.LoadFromFile(ctrlCtx.runOptions.kubeconfig)
```
alvaroaleman, line 98: The kubeconfig is available right below your code (`clientcmd.BuildConfigFromFlags(ctrlCtx.runOptions.masterURL, ctrlCtx.runOptions.kubeconfig)`)

api/pkg/controller/seed-proxy/controller.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return nil
+}
+
+func managedByController(meta metav1.Object) bool {
+	labels := meta.GetLabels()
+	return labels[ManagedByLabel] == ControllerName
+}
+
+func secretsPredicate() predicate.Funcs {
```
alvaroaleman: There is no need to have a separate predicate func ber type or am I missing it?

api/pkg/controller/seed-proxy/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err
+	}
+
+	eventHandler := &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+		return []reconcile.Request{
+			{
+				NamespacedName: types.NamespacedName{
+					Namespace: "",
+					Name:      "seed-proxy-reconcile",
```
alvaroaleman: Wouldn't it be better to create a reconcile.Request per seed? Otherwise if one seed is broken (e.G. the kubeconfig doesn't work) it can block the reconciliation of all seeds

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (r *Reconciler) fetchServiceAccountSecret(ctx context.Context, client ctrlruntimeclient.Client) (*corev1.Secret, error) {
+	sa := &corev1.ServiceAccount{}
+	name := types.NamespacedName{
+		Namespace: ServiceAccountNamespace,
+		Name:      ServiceAccountName,
+	}
+
+	if err := r.Get(ctx, name, sa); err != nil {
```
alvaroaleman: Shouldn't you be using the passed in client instead of the one that is part of the Reconciler? The latter is not to the seed but to the master

api/pkg/controller/seed-proxy/resources.go
```diff
@@ -0,0 +1,335 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func secretName(contextName string) string {
```
alvaroaleman, line 21: What clears up the resources if someone removes a seed?

api/pkg/controller/seed-proxy/resources.go
```diff
@@ -0,0 +1,335 @@
...
+
+while true; do
+  kubectl proxy \
+	 --address=0.0.0.0 \
+	 --port=$PROXY_PORT \
+	 --accept-hosts='' \
+	 --accept-paths='^/api/v1/namespaces/monitoring/services/.*/proxy/.*$' \
+	 --reject-methods='^(POST|PUT|PATCH|DELETE)$'
+  sleep 1
+done
```
alvaroaleman: I'd actually prefer it if this crashes to improve visibility

api/pkg/controller/seed-proxy/controller.go
xrstf: It's not required per se, it just seemed cleaner and in older versions of this PR there were dedicated predicates for each type.

alvaroaleman: I'd argue "removing 75% of the current code for this" is cleaner ;)

api/pkg/controller/seed-proxy/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	MasterGrafanaNamespace        = "monitoring-master"
+	MasterGrafanaConfigMapName    = "grafana-seed-proxies"
+	KubectlProxyPort              = 8001
+	NameLabel                     = "app.kubernetes.io/name"
+	InstanceLabel                 = "app.kubernetes.io/instance"
+	ManagedByLabel                = "app.kubernetes.io/managed-by"
+)
+
+// Add creates a new Monitoring controller that is responsible for
+// operating the monitoring components for all managed user clusters
```
mrIncompetent: Comment does not match

api/cmd/master-controller-manager/main.go
xrstf: How would I get the contexts from the restclient.Config object later in my controller?

api/pkg/controller/seed-proxy/resources.go
xrstf: In coop with Henrik I put a garbage collect routine at the start of every reconcile run. This should remove any unused resources.

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -46,6 +51,121 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+
+	if err := r.garbageCollectDeployments(ctx, options); err != nil {
+		return fmt.Errorf("failed to garbage collect Deployments: %v", err)
+	}
+
+	if err := r.garbageCollectConfigMaps(ctx, options); err != nil {
+		return fmt.Errorf("failed to garbage collect ConfigMaps: %v", err)
+	}
+
+	if err := r.garbageCollectSecrets(ctx, options); err != nil {
```
mrIncompetent: My suggestion is still to set a OwnerRef on all objects pointing to this secret. 
Then we don't need to take care of all objects

api/pkg/controller/seed-proxy/controller.go
```diff
@@ -0,0 +1,177 @@
...
+		obj  runtime.Object
+		pred predicate.Funcs
+	}
+
+	ownedByPred := makePredicateFuncs(managedByController)
+	typesToWatch := []watcher{
+		{obj: &appsv1.Deployment{}, pred: ownedByPred},
+		{obj: &corev1.Service{}, pred: ownedByPred},
+		{obj: &corev1.Secret{}, pred: ownedByPred},
+		{obj: &corev1.ConfigMap{}, pred: ownedByPred},
```
alvaroaleman: This is still a needlessly complex premature optimization, please remove the `makePredicateFuncs` wrapping and make it just a `predicate.Funcs`, remove the `watcher` struct and make this just a slice of `runtime.Object` and just insert the predicateFunc in the line where you call `c.Watch`.

If this is needed at some point you can add it, but please keep it simple as long as possible.

api/pkg/controller/seed-proxy/reconciler.go
xrstf: I implemented owner references now.

api/pkg/controller/seed-proxy/controller.go
xrstf: I agree with the first part, but not the second. Creating the predicate.Funcs inside a function is IMHO more readable, but not enough to argue too hard.

api/pkg/controller/seed-proxy/reconciler.go
xrstf: Good catch! Didn't notice because on dev master and seed are inside the same cluster.

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	if err := reconciling.ReconcileServices(ctx, creators, MasterTargetNamespace, r.Client); err != nil {
+		return fmt.Errorf("failed to reconcile Services in the namespace %s: %v", MasterTargetNamespace, err)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) reconcileGrafana(ctx context.Context) error {
```
alvaroaleman: Why not directly call `ensureMasterGrafanaProvisioning`?

xrstf: I wanted to also update the Grafana deployment with an annotation for the ConfigMap's hash, so there were originally multiple tasks for reconciling Grafana. Now that there is only one task, yeah, sure, one could remove this function.

alvaroaleman:
@xrstf You should add the seed team to `api/cmd/master-controller-manager/OWNERS`

alvaroaleman:
and create an owners file containing the seed-team in `api/pkg/controller/seed-proxy`

PR #3276, opened 2019-04-18
Run migrations within the Kubermatic controller manager

alvaroaleman:

Fixes #

**Documentation**:

```release-note
Migrations are now executed only after the leader lease was acquired
```

### DIFF
--- a/api/pkg/crd/migrations/migrations.go
+++ b/api/pkg/crd/migrations/migrations.go
@@ -1,7 +1,6 @@
-package main
+package migrations
 
 import (
-	"flag"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


api/pkg/crd/migrations/migrations.go
```diff
@@ -35,51 +31,51 @@ type cleanupContext struct {
...
 	if err != nil {
-		glog.Fatal(err)
+		return fmt.Errorf("failed to create kubeClient: %v", err)
+	}
+	kubermatiClient, err := kubermaticclientset.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("failed to create Kuermatic client: %v", err)
 	}
 
-	ctx.config.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
```
alvaroaleman, line 65: I removed this, do you think this will cause any issues? I think this is the default anyways

mrIncompetent: I don't think so

PR #3287, opened 2019-04-19
Update prometheus

mrIncompetent:
This PR updates the Prometheus we deploy for each user cluster.

```release-note
Update the user cluster Prometheus to v2.9.1
```

### DIFF
--- a/api/pkg/resources/prometheus/statefulset.go
+++ b/api/pkg/resources/prometheus/statefulset.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
-	"github.com/kubermatic/kubermatic/api/pkg/resources/apiserver"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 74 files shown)


mrIncompetent:
I had a look at the logs while the conformance tests were running: Everything looks normal :+1:

api/pkg/resources/prometheus/statefulset.go
```diff
@@ -81,12 +80,6 @@ func StatefulSetCreator(data *resources.TemplateData) reconciling.NamedStatefulS
 				resourceRequirements = *data.Cluster().Spec.ComponentsOverride.Prometheus.Resources
 			}
 
-			apiserverIsRunningContainer, err := apiserver.IsRunningInitContainer(data)
-			if err != nil {
-				return nil, err
-			}
-			set.Spec.Template.Spec.InitContainers = []corev1.Container{*apiserverIsRunningContainer}
```
mrIncompetent, line 88: If the API server is crashlooping and the Prometheus got rescheduled - we're losing all monitoring & alerting

api/pkg/resources/prometheus/statefulset.go
```diff
@@ -151,7 +144,7 @@ func StatefulSetCreator(data *resources.TemplateData) reconciling.NamedStatefulS
 						PeriodSeconds:       5,
 						TimeoutSeconds:      3,
 						FailureThreshold:    6,
-						InitialDelaySeconds: 30,
+						InitialDelaySeconds: 5,
```
mrIncompetent, line 147: No need to wait 30s for the readinessprobe

PR #3296, opened 2019-04-23
Use zap logger for the addon controller

mrIncompetent:
This is just a example implementation on how we could utilize the [zap logger](https://github.com/uber-go/zap) for structured logging.

```release-note
NONE
```


alvaroaleman:
Wasnt the initial goal to have structured logging? That doesn't seem to be the case with `debug=true` or am I misunderstanding it?

mrIncompetent:
@alvaroaleman 
> Wasnt the initial goal to have structured logging? That doesn't seem to be the case with debug=true or am I misunderstanding it?

Correct. When using `-debug=true` it does not use JSON as output formatting.
I've wanted to start with a small example PR to get the discussion going.

I'm not happy with the `-debug` flag as well - i would rather prefer something like the the [prometheus-operator does it](https://github.com/coreos/prometheus-operator/blob/master/cmd/operator/main.go#L138): `-log-level` & `-log-format`
But that would mean we would need to copy/modify the [controller-runtime helper](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go) for zap. I'm fine with copying the helper but wanted to have a consensus on that

alvaroaleman:
> I'm not happy with the -debug flag as well - i would rather prefer something like the the prometheus-operator does it: -log-level & -log-format
But that would mean we would need to copy/modify the controller-runtime helper for zap. I'm fine with copying the helper but wanted to have a consensus on that

SGTM and writing our own helper doesn't seem too much of an effort

mrIncompetent:
I've replaced the 2 flags with:
```
-log-debug=[true|false]
-log-format=[JSON|Plain]
```
due to the surprising weird levels of Zap & the logr implementation of it: https://github.com/kubermatic/kubermatic/pull/3296/files#diff-c32863a883a0bfd98ac3bf2f2f234c09R42

mrIncompetent:
cc @xrstf @themue

api/pkg/controller/addon/addon_controller.go
```diff
@@ -249,7 +257,7 @@ func (r *Reconciler) GetTemplateFuncs() template.FuncMap {
 	return funcs
 }
 
-func (r *Reconciler) getAddonManifests(addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) ([]*bytes.Buffer, error) {
+func (r *Reconciler) getAddonManifests(log logr.Logger, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) ([]*bytes.Buffer, error) {
```
alvaroaleman: I must say I really hate this pattern of having to pass a logger around everywhere. Can we maybe just make it a package variable?

api/pkg/log/zap.go
```diff
@@ -0,0 +1,71 @@
+package log
```
alvaroaleman, line 1: If this is supposed to become our default logging and replace `glog`, can we document that somewhere?

api/pkg/controller/addon/addon_controller.go
mrIncompetent: No, we enrich the logger throughout the sync - which is not possible when using a package variable. Like addon & cluster name.
Enriching the logger is one of the benefits of passing down a log object. We now can clearly filter for all logs which belong to a:
- single sync
- cluster
- addon
- Whatever we want

I thought about adding it to the context but that makes using the logger a pain:
https://dave.cheney.net/2017/01/26/context-is-for-cancelation

api/pkg/log/zap.go
mrIncompetent: I'm happy to update our docs as soon we agree on this approach :)

mrIncompetent:
/hold

Docs must be updated

api/pkg/controller/addon/addon_controller.go
alvaroaleman: So we now have to pass a logger into everything where we want to log...?

mrIncompetent: Yes. There's no way around that without loosing information or passing around the required fields we want to attach to the log.

IMO its better compared to attaching the fields on each individual log statement:
```
	log.WithValues(
		"addon", addon.Name,
		"cluster", cluster.Name,
	).Info("foo")

	log.WithValues(
		"addon", addon.Name,
		"cluster", cluster.Name,
	).Info("bar")

	log.WithValues(
		"addon", addon.Name,
		"cluster", cluster.Name,
	).Info("baz")
```
Its prone for typos & missing information, which leads to frustration during filtering due to missing fields

alvaroaleman:
As this will change the (expected to be used) logging for everything in Kubermatic, can we discuss this in the dev team meeting?

mrIncompetent:
> As this will change the (expected to be used) logging for everything in Kubermatic, can we discuss this in the dev team meeting?

Sure

alvaroaleman:
Sidecomment to keep in mind: If someone actually needs debug logs, this change would mean they would have to both configure logging for glog and for our own logging

mrIncompetent:
/hold cancel
Docs have been updated.

@alvaroaleman PTAL

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -186,6 +198,10 @@ func (o controllerRunOptions) validate() error {
 		return errors.New("The metrics-server addon must be disabled, it is now deployed inside the seed cluster")
 	}
 
+	if !kubermaticlog.AvailableFormats.Has(o.log.format) {
+		return errors.New("invalid log-format specified. Available: " + kubermaticlog.AvailableFormats.String())
```
alvaroaleman: Nit: Its nice to print the invalid value in the error

api/hack/run-controller.sh
```diff
@@ -14,6 +14,7 @@ make kubermatic-controller-manager
 KUBERMATIC_WORKERNAME=${KUBERMATIC_WORKERNAME:-$(uname -n)}
 
 ./_build/kubermatic-controller-manager \
+  -debug \
```
alvaroaleman: This flag doens't exist or did I miss that?

api/pkg/controller/addon/addon_controller.go
```diff
@@ -67,16 +68,20 @@ type Reconciler struct {
...
+	log logr.Logger,
 	numWorkers int,
 	workerName string,
 	addonCtxVariables map[string]interface{},
 	kubernetesAddonDir string,
 	openshiftAddonDir string,
 	overwriteRegistey string,
-	KubeconfigProvider KubeconfigProvider) error {
-
+	KubeconfigProvider KubeconfigProvider,
```
alvaroaleman: Not from this PR but why is the variable name upper cased?

api/pkg/controller/addon/addon_controller.go
```diff
@@ -127,6 +132,8 @@ func Add(
 func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	log := r.log.WithValues("request", request)
+	log.V(1).Info("Processing")
```
alvaroaleman: Doesn't it have a `Debug` func? I'd prefer it if we stopped using the `V(N)` funcs :grimacing:

api/pkg/controller/addon/addon_controller.go
```diff
@@ -135,17 +142,18 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 		return reconcile.Result{}, err
 	}
+	log = r.log.WithValues("cluster", addon.Spec.Cluster.Name)
 
 	// Add a wrapping here so we can emit an event on error
-	err := r.reconcile(ctx, addon)
+	err := r.reconcile(ctx, log, addon)
 	if err != nil {
-		glog.Errorf("Failed to reconcile addon %s: %v", addon.Name, err)
+		log.Error(err, "Reconciling failed")
```
alvaroaleman: Would be nice to include the addon name here

api/pkg/controller/addon/addon_controller.go
```diff
@@ -418,26 +428,26 @@ func (r *Reconciler) getAddonLabel(addon *kubermaticv1.Addon) map[string]string
...
 
-func (r *Reconciler) writeCombinedManifest(manifest *bytes.Buffer, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) (string, fileHandlingDone, error) {
+func (r *Reconciler) writeCombinedManifest(log logr.Logger, manifest *bytes.Buffer, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) (string, fileHandlingDone, error) {
 	//Write combined Manifest to disk
 	manifestFilename := path.Join("/tmp", fmt.Sprintf("cluster-%s-%s.yaml", cluster.Name, addon.Name))
 	if err := ioutil.WriteFile(manifestFilename, manifest.Bytes(), 0644); err != nil {
 		return "", nil, fmt.Errorf("failed to write combined manifest to %s: %v", manifestFilename, err)
 	}
-	glog.V(4).Infof("wrote combined manifest for addon %s/%s to %s\n%s", addon.Name, addon.Namespace, manifestFilename, manifest.String())
+	log.Info("Wrote combined manifest", "file", manifestFilename, "content", manifest.String())
```
alvaroaleman: That is definitely debug log, we certainly dont want this shown by default

docs/logging.md
```diff
@@ -1,15 +1,54 @@
...
 
 ## Guidelines
 
-We're following the [guidelines from Kubernetes](https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/devel/logging.md).
-
+There are 2 things you will log:
+- Errors
+- Infos 
+
+Info logs are leveled, though only 2 levels exist. The higher the level, the higher the verbosity (More less important logs):
```
alvaroaleman: As mentioned elsewhere, Id really like to not have numeric levels but Instead just `log.Infof`, `log.Debug` and `log.Error` because thats more self-explanatory. I thought `Zap` supported that, do I remember that incorrectly?

Regarding the docs, I'd just remove the whole part until `Usage`, the sentence above is confusing and the `### Levels` paragraph explains the whole thing very well

api/pkg/controller/addon/addon_controller.go
mrIncompetent: No. It conflicts with the `logr` interface

docs/logging.md
mrIncompetent: > s mentioned elsewhere, Id really like to not have numeric levels but Instead just log.Infof, log.Debug and log.Error because thats more self-explanatory. I thought Zap supported that, do I remember that incorrectly?

Zap supports a `Debug` func, but not the `logr` interface: 
https://github.com/go-logr/logr

api/pkg/controller/addon/addon_controller.go
alvaroaleman: It conflicts? Why would the `logr` interface care if we implement something on top of it?

api/cmd/kubermatic-controller-manager/options.go
themue: Also looks a bit strange "lower-case sentence. Upper-case new sentence". Better "lower-case; lower-case ...".

docs/logging.md
themue: Going with @alvaroaleman , these `V(n)` feel somehow too fine granular. But have to take a look at `logr`. What's the benefit of it opposite to a small own wrapper for any logger, have to find out. 😉

api/pkg/log/zap.go
```diff
@@ -0,0 +1,80 @@
...
+
+var (
+	AvailableFormats = Formats{FormatJSON, FormatConsole}
+)
+
+func (f Formats) String() string {
+	const separator = ", "
+	var s string
+	for _, format := range f {
+
```
themue: s/\n//

api/pkg/log/zap.go
```diff
@@ -0,0 +1,80 @@
...
+		s = s + separator + string(format)
+	}
+	return strings.TrimPrefix(s, separator)
+}
+
+func (f Formats) Has(s string) bool {
```
themue: Wouldn't `Contains()` be better? More often to find in Go standard library.

api/pkg/log/zap.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, format := range f {
+		if s == string(format) {
+			return true
+		}
+	}
+	return false
+}
+
+func New(debug bool, format Format) logr.Logger {
```
themue: Returning a `logr.Logger` by in contructor in our package creates a new dependency in the using code. Defining an own wrapper hiding `logr` and returning only an interface (see `context.Context`) allows more flexibility in the future.

themue (commented):
Happy about structured logging but if possible (still have to check which interface to provide for users) would prefer more hiding of dependencies and implementation details inside of our own log package.

api/pkg/controller/addon/addon_controller.go
mrIncompetent: The function `WithValues` makes extending it impossible.
The logr.Logger interface defines:
```go
WithValues(keysAndValues ...interface{}) Logger
```

As soon as we would call `WithValues` we would loose the extended functionality.

alvaroaleman: I see. Modulo the other comments PR LGTM then

api/pkg/controller/addon/addon_controller.go
mrIncompetent: We have the addon name as key already set on the logger object :)
```golang
	defer cancel()
```

Request implements the Stringer interface, resulting in `namespace/name` (example `cluster-xxxxx/flannel`)

api/pkg/log/zap.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/pkg/log/zap.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

docs/logging.md
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/pkg/controller/addon/addon_controller.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/pkg/controller/addon/addon_controller.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/hack/run-controller.sh
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/cmd/kubermatic-controller-manager/options.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/7f4f4a1d4ab5b403bc6a2d98ce8a0ff97c0cb860

api/cmd/kubermatic-controller-manager/main.go
```diff
@@ -4,77 +4,81 @@ import (
...
 	controllerName = "kubermatic-controller-manager"
 )
 
 func main() {
-
 	options, err := newControllerRunOptions()
 	if err != nil {
-		glog.Fatalf("failed to create controller run options due to = %v", err)
+		fmt.Printf("Failed to create controller run options due to = %v", err)
+		os.Exit(1)
```
mrIncompetent, line 44: Avoids having to import glog here.
We don't have the logger yet due to the missing flag/options evaluation

api/pkg/log/zap.go
mrIncompetent: We now return a pointer to a zap.Logger

mrIncompetent:
@alvaroaleman @themue PTAL

api/cmd/kubermatic-controller-manager/main.go
alvaroaleman: Please add a trailing `\n`

api/cmd/kubermatic-controller-manager/main.go
```diff
@@ -4,77 +4,81 @@ import (
...
 	}
+
 	if err := options.validate(); err != nil {
-		glog.Fatal(err)
+		fmt.Print(err)
```
alvaroaleman: Please use `Println` or add a trailing `\n`

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -186,6 +199,10 @@ func (o controllerRunOptions) validate() error {
 		return errors.New("The metrics-server addon must be disabled, it is now deployed inside the seed cluster")
 	}
 
+	if !kubermaticlog.AvailableFormats.Contains(o.log.format) {
```
alvaroaleman: Pretty sure we want this sooner or later in other `cmds` as well. What about adding a `func (lo logOptions) validate` so the next person exports this instead of copy pasting it?

api/pkg/controller/addon/addon_controller.go
```diff
@@ -105,7 +110,7 @@ func Add(
 		addonList := &kubermaticv1.AddonList{}
 		listOptions := &ctrlruntimeclient.ListOptions{Namespace: cluster.Status.NamespaceName}
 		if err := client.List(context.Background(), listOptions, addonList); err != nil {
-			glog.Errorf("failed to get addons for cluster %s: %v", cluster.Name, err)
+			log.Error(err, "Failed to get addons for cluster", "cluster", cluster.Name)
```
alvaroaleman: Did you check how this prints, is `err` as first arg the correct approach?

api/cmd/kubermatic-controller-manager/main.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3296/commits/55888f0aa14c71f01e267a65206d00a88b3fa55b

api/pkg/controller/addon/addon_controller.go
mrIncompetent: Missed that in a bunch of places, the `zap.Logger` & `zap.SugardLogger` calls differ in a way that the latter requires you to use the `*w` functions(`Debugw`, `Infow`) for adding fields...

mrIncompetent: fixed with https://github.com/kubermatic/kubermatic/pull/3296/commits/55888f0aa14c71f01e267a65206d00a88b3fa55b

xrstf:
🎉

PR #3299, opened 2019-04-23
Add emptyDir mount to /tmp for the CoreDNS pod to avoid fatal glog invocations

mrIncompetent:
This is required as CoreDNS crashes in case any error gets logged from the kube-client.
We see this a lot on cloud.kubermatic.io since we use preemtible machines & the API server will become unavailable multiple times a day.
Adding a emptyDir to `/tmp` is the proposed fix from the CoreDNS maintainers.

See https://github.com/coredns/deployment/pull/138

```release-note
NONE
```

### DIFF
--- a/addons/dns/coredns-deployment.yaml
+++ b/addons/dns/coredns-deployment.yaml
@@ -94,6 +94,8 @@ spec:
         - name: config-volume
           mountPath: /etc/coredns
           readOnly: true
+        - name: tmp
+          mountPath: /tmp
         ports:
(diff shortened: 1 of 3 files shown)


mrIncompetent:
/hold

I need to update the tags

alvaroaleman:
wtf.

/lgtm
/approve

mrIncompetent:
@alvaroaleman Tag has been updated PTAL

/hold cancel

PR #3335, opened 2019-04-25
Add defaulting for Deployment containers

alvaroaleman:

Adds defaulting for Deployment containers to prevent reconciliation loops because someone forgot to set ImagePullPolicy/TerminationMessagePath/TerminationMessagePolicy

If you are okay with this approach I would probably like to even expand it further. Right now its too simple to get this wrong and too easy to miss in reviews.

Fixes #

**Documentation**:

```release-note
none
```

/assign @mrIncompetent

### DIFF
--- a/api/codegen/reconcile/main.go
+++ b/api/codegen/reconcile/main.go
@@ -202,6 +202,35 @@ func {{ .ResourceName }}ObjectWrapper(create {{ .ResourceName }}Creator) ObjectC
 func Reconcile{{ .ResourceName }}s(ctx context.Context, namedGetters []Named{{ .ResourceName }}CreatorGetter, namespace string, client ctrlruntimeclient.Client, objectModifiers ...ObjectModifier) error {
 	for _, get := range namedGetters {
 		name, create := get()
+{{ if eq .ResourceName "Deployments" }}
+
+		defaultContainer := func(c *corev1.Container) {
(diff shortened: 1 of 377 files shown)


mrIncompetent:
Great idea!

Though to have this re-usable we could treat them as wrappers (Like you already do), but maybe have them not generated? And instead generate the usage of it.

I'm thinking of something like:
- Define typed wrappers in https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/reconciling/wrapper.go
- Define wrappers in codegen via `reconcileFunctionData` https://github.com/kubermatic/kubermatic/blob/master/api/codegen/reconcile/main.go#L49
- The typed wrappers would be applied here: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/reconciling/zz_generated_reconcile.go#L214

So we can define a list of wrappers for each type, which then get applied to the Reconcile* funcs.

My main intention behind is to keep the static code as small as possible in the codegen. As maintaining static code as "text" is harder.

WDYT?

alvaroaleman:
SGTM

alvaroaleman:
/hold

If the tests pass, I'll remove the relevant properties from the existing Deployment Resources to make sure this actually works

alvaroaleman:
@mrIncompetent Updated, PTAL

PR #3340, opened 2019-04-25
symlink Grafana datasources

xrstf:
In contrast to dashboards, where we can configure Grafana to look into random paths to find JSON files, datasources must be a flat list of YAML files inside one specific directory with no support for sub-directories. This makes it complicated to provision Grafana in an environment like our master where some datasources (master prometheus) are configured statically via Helm and some are created dynamically via the new seed-proxy controller.

This PR makes it so that datasources are not mounted directly to `/etc/grafana/provisioning/datasources` anymore, but instead we now have an initContainer that takes a list of source directory and symlinks each file into the special datasource directory. Since Grafana does not yet allow to reload datasources at runtime (see https://github.com/grafana/grafana/pull/16579), it does not help to have a permanent sidecar that would reload Grafana everytime a mounted volume (most likely ConfigMaps) changes. Instead we will maybe work using annotations that are updated by the seed-proxy controller to trigger restarts.

Since we use a new configuration key (`grafana.provisioning.datasources.paths`) I don't expect any compatibility breaks.

The provision logic itself seems complicated (compared to a simple rsync), but it takes care of

* flattening the directory structure and ignoring any sub directories,
* providing a nice, readable log and
* prepending an index number so that two volumes with a `prometheus.yaml` will not end up overwriting each other (if you do not want a certain datasource, do not include it in your `paths` list)

Relates to #3238

```release-note
NONE
```

### DIFF
--- a/config/monitoring/grafana/templates/deployment.yaml
+++ b/config/monitoring/grafana/templates/deployment.yaml
@@ -14,6 +14,60 @@ spec:
       labels:
         app: grafana
     spec:
+      {{ if .Values.grafana.provisioning.datasources.paths }}
+      initContainers:
+      - image: quay.io/kubermatic/util:1.0.0-2
+        name: datasource-assembler
+        command: [/bin/bash]
+        args:
+        - -c
+        - |
+          set -euo pipefail
+
+          target=/etc/grafana/provisioning/datasources
+
+          create_symlinks() {
+            local base="$(realpath $1)/" # keep a trailing slash for the rel trim later
+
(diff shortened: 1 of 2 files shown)


kron4eg:
@xrstf what about hash of configs inside template (so grafana would restart in case of those changed)?

xrstf:
Can do. Wanted to do that via the seed-proxy controller, but that would probably not be a good idea.

kron4eg (approved):
/approve
/lgtm

PR #3343, opened 2019-04-25
checks projects OwnerRef before deleting a duplicated user

### LINKED ISSUE #3322: Add additional safety check for duplice user deletion
**User Story**
Duplicated users might lead to project deletion during migration.
The current migration logic only checks if a user has projects via the users project list `user.spec.projects`. To add an additional level of safety we should also check if there is a Project which has a OwnerRef pointing to the User - Otherwise we might delete a user which just lacks `user.spec.projects`

p0lyn0mial:
**What this PR does / why we need it**: Before a user is deleted the tool checks if the user doesn't belong to a project. It first checks user.spec.projects list and if it is empty it also checks UserProjectBindings.
Under normal conditions, the RBAC controller generates UserProjectBindings for project's owners (it examines OnwerReferences). However a faulty controller might have not created bindings so checking OwnerReferences is the last resort.

Fixes #3322

```release-note
NONE
```

### DIFF
--- a/api/cmd/projects-migrator/main.go
+++ b/api/cmd/projects-migrator/main.go
@@ -901,6 +901,21 @@ func doesUserOwnProjectAs(user kubermaticv1.User, ctx migrationContext, groupPre
 		}
 	}
 
+	// Under normal conditions, the RBAC controller generates UserProjectBindings for project's owners (it examines OnwerReferences).
+	// however a faulty controller might have not created bindings.
+	// so as a last resort examine OwnerReferences
+	allProjects, err := ctx.masterKubermaticClient.KubermaticV1().Projects().List(metav1.ListOptions{})
+	if err != nil {
+		return "", err
+	}
+	for _, project := range allProjects.Items {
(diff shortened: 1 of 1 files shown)


api/cmd/projects-migrator/main.go
```diff
@@ -901,6 +901,21 @@ func doesUserOwnProjectAs(user kubermaticv1.User, ctx migrationContext, groupPre
...
 
+	// Under normal conditions, the RBAC controller generates UserProjectBindings for project's owners (it examines OnwerReferences).
+	// however a faulty controller might have not created bindings.
+	// so as a last resort examine OwnerReferences
+	allProjects, err := ctx.masterKubermaticClient.KubermaticV1().Projects().List(metav1.ListOptions{})
+	if err != nil {
+		return "", err
+	}
+	for _, project := range allProjects.Items {
+		if ownerName := isOwnedByUser(project.OwnerReferences); len(ownerName) > 0 {
```
p0lyn0mial, line 912: this is how we assign `OwnerRef` to projects https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/project.go#L75

PR #3347, opened 2019-04-25
Remove -kubermatic suffix from Prometheus Chart

xrstf:
We want to install some of our charts multiple times into a cluster, especially Prometheus. In #3252 we started to make this easier by putting the release name into each identifier of the Prometheus chart (e.g. `prometheus-kubermatic-{{ .Release.Name }}`.

This was a good start, but leads to "stuttering" and ugly names, like a role being named `prometheus-kubermatic-prometheus` (guess which part is the release name). Having a `kubermatic` inside the names is only a hack someone made a long time ago to allow running a "Loodse Prometheus" next to the one used for Kubermatic. Considering that we "publish" the monitoring stack in github.com/kubermatic/monitoring and use it for non-Kubermatic customers, we should also think about not putting `kubermatic` anywhere into the charts.

This PR attemps to rectify all these issues by doing a "light" version of what many popular Helm charts already do: Use the release name unless the `nameOverride` variable is set. This allows us to for example install Prometheus as `prometheus-master` (release name) Into `monitoring-master` namespace and still have the resources be named `prometheus-config`, `prometheus-rules` etc.

For our dev master/seed cluster, we would have this configuration:

1. (seed prometheus): release name: `prometheus-seed`, namespace `monitoring`, name override `prometheus`
    * We would change the release name from `prometheus` to `prometheus-seed` because it makes the distinction to the master clearer *and* we can easily run this new release along the old (see below).
2. (master prometheus): release name: `prometheus-master`, namespace `monitoring-master`, name override `prometheus`

The biggest, gigantic, downside to this is that we lose data because the PVCs are not named `prometheus-kubermatic-db-[statefulset-name]-[index]` anymore, but `db-[statefulset-name]-[index]` (the name of the StatefulSet is already unique enough, no need to duplicate it into the PVC name). This means that customers "lose" their data (in the sense that it becomes inaccessible). A migration path could be that customers run Prometheus twice and keep the old for the duration of their retention period.

The Grafana and Alertmanager charts would need the same treatment, but at worst we lose configured silences.

Relates to #3238

```release-note
BREAK: refactored Prometheus Helm Chart for Master-Cluster monitoring, see documentation for migration notes 
```

### DIFF
--- a/api/cmd/kubermatic-api/options.go
+++ b/api/cmd/kubermatic-api/options.go
@@ -45,7 +45,7 @@ func newServerRunOptions() (serverRunOptions, error) {
 	flag.StringVar(&s.listenAddress, "address", ":8080", "The address to listen on")
 	flag.StringVar(&s.kubeconfig, "kubeconfig", "", "Path to the kubeconfig.")
 	flag.StringVar(&s.internalAddr, "internal-address", "127.0.0.1:8085", "The address on which the internal handler should be exposed")
-	flag.StringVar(&s.prometheusURL, "prometheus-url", "http://prometheus-kubermatic.monitoring.svc.local:web", "The URL on which this API can talk to Prometheus")
+	flag.StringVar(&s.prometheusURL, "prometheus-url", "http://prometheus.monitoring.svc.local:web", "The URL on which this API can talk to Prometheus")
 	flag.StringVar(&s.masterResources, "master-resources", "", "The path to the master resources (Required).")
 	flag.StringVar(&s.dcFile, "datacenters", "datacenters.yaml", "The datacenters.yaml file path")
 	flag.StringVar(&s.workerName, "worker-name", "", "Create clusters only processed by worker-name cluster controller")
--- a/config/iap/values.yaml
+++ b/config/iap/values.yaml
@@ -46,7 +46,7 @@ iap:
     #   client_secret: xxx
     #   encryption_key: xxx
     #   config: {} ## see https://www.keycloak.org/docs/latest/securing_apps/index.html#example-usage-and-configuration
-    #   upstream_service: prometheus-kubermatic.monitoring.svc.cluster.local
+    #   upstream_service: prometheus.monitoring.svc.cluster.local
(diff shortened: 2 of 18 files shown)


xrstf:
/hold

If we ever decide that we actually want this, we must update the postsubmit Prow job to deploy this as a different chart or else all things fall apart.

kron4eg (approved):
/lgtm
/approve

xrstf:
I have extended this PR with a more-or-less automated way to migrate existing configurations. By setting `prometheus.migration.enabled=true` in your values.yaml, the chart will *not* deploy the StatefulSet, but instead a Job that creates the PVCs for the new StatefulSet (`db-promtheus-0` and `db-prometheus-1`) and then rsyncs the data from `prometheus-kubermatic-db-prometheus-kubermatic-N` to `db-prometheus-N`. Once this job has finished, you disable the migration flag and deploy again and voila, you get a new StatefulSet that uses the new PVCs and boots up just fine.

At first I wanted to do this via an initContainer on the StatefulSet, but there is no way to retrieve the old PVC names and have them be mounted to the pods. Having a dedicated Job is the only halfway sane way of doing this.

The migration job will ensure that the target volume is empty and will create a lockfile so that once the volumes for replica 0 have been migrated, subsequent runs of the migration job would skip this volume (the job should never run more than once, but for safety sake there's a lockfile).

It's not a zero-effort migration, but this is something we can document easily:

```
helm upgrade --install --set prometheus.migration.enabled=true prometheus .
(wait....)
helm upgrade --install prometheus .
```

mrIncompetent:
@alvaroaleman @xrstf @kron4eg 

Why is this PR not progressing ?

xrstf:
I ran a test by installing the Prometheus chart under a new name into a new namespace, letting it scrape a bit of data for a few minutes, then performing the two migration steps and ended up with a working installation containing the pre-migration data.

Color me somewhat confident that this could work.

xrstf:
/hold cancel

Let's do this!

PR #3350, opened 2019-04-25
Add GCP support

### LINKED ISSUE #3120: Add GCP support to Kubermatic
**User Story**

As a user, I want to use Kubermatic to create Kubernetes Clusters on GCP

**Acceptance criteria**

* It is possible to use GCP as a provider to create Kubermatic-managed clusters on

nikhita:

This PR adds GCP support to Kubermatic.

Fixes #3120

```release-note
It is now possible to create Kubermatic-managed clusters on GCP.
```

/cc @alvaroaleman


api/pkg/api/v1/types.go
```diff
@@ -86,6 +86,11 @@ type OpenstackDatacenterSpec struct {
 	EnforceFloatingIP bool      `json:"enforce_floating_ip"`
 }
 
+// GCPDatacenterSpec specifies a datacenter of GCP.
+type GCPDatacenterSpec struct {
+	Region string `json:"region"`
```
nikhita: I'm not entirely sure where exactly we would randomly assign the zone depending on the region...

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	dc, ok := g.dcs[cluster.Spec.Cloud.DatacenterName]
+	if !ok {
+		return nil, fmt.Errorf("could not find datacenter %s", cluster.Spec.Cloud.DatacenterName)
+	}
+
+	if dc.Spec.GCP == nil {
+		return nil, fmt.Errorf("datacenter %q is not a valid GCP datacenter", cluster.Spec.Cloud.DatacenterName)
+	}
+
+	if cluster.Spec.Cloud.GCP.Network == "" && cluster.Spec.Cloud.GCP.Subnetwork == "" {
```
nikhita, line 47: Regarding subnet, my understanding is that it is ok to have `subnetwork` as empty because the network will then start with one subnet per region automatically...what am I missing here?

Also, related:

https://github.com/kubermatic/machine-controller/blob/8d20e37dcd84f41bf19d56e946fd091c34417750/pkg/cloudprovider/provider/gce/service.go#L62-L81

nikhita (commented):
Have a couple of questions

nikhita:
/hold
This also needs an update to machine-controller and especially: https://github.com/kubermatic/machine-controller/pull/507

alvaroaleman:
This is awesome :)))

Can you add GCP to `api/cmd/conformance-tests`?

api/pkg/api/v1/types.go
alvaroaleman: @mrIncompetent and me discussed this, we will instead require the user to enter the Zone suffix upon node creation and display a link to the Google docs explaining what it means to have nodes in different Zones

nikhita:
Updated with firewall support...but have some questions. :grimacing: 

Added the questions as TODOs in this commit - https://github.com/kubermatic/kubermatic/pull/3350/commits/1b26ad487f8770bdfe2402974a8aa68cfed54283

api/hack/ci/ci-run-minimal-conformance-tests-gcp.sh
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env bash
```
alvaroaleman: Instead of copying the script it was better if you made the `-providers` argument to the `conformance-tests` binary configurable and default to `aws`

nikhita: Yup, intended to do that. :+1:

nikhita:
Updated, but tests will fail. Needs https://github.com/kubermatic/machine-controller/pull/511.

api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -186,70 +195,134 @@ if [[ -n ${UPGRADE_TEST_BASE_HASH:-} ]]; then
...
-  -providers=aws \
-  -exclude-distributions="${EXCLUDE_DISTRIBUTIONS}" \
-  -kubermatic-delete-cluster=false
-
-# No upgradetest, just exit
-if [[ -z ${UPGRADE_TEST_BASE_HASH:-} ]]; then
-  echodate "Success!"
-  exit 0
+
+if if [[ $provider = "aws" ]]; then
```
alvaroaleman: Instead of duplicating this you could do something like:

```
if [[ $provider -eq aws ]]; then
  EXTRA_ARGS="-aws-access-key-id=${AWS_E2E_TESTS_KEY_ID}
     -aws-secret-access-key=${AWS_E2E_TESTS_SECRET}"
elif [[ $provider -eq gcp ]]; then
  EXTRA_ARGS="-my-gcp=arg"
fi
```
And then just append the `$EXTRA_ARGS` to the `./conformance-tests` call

api/pkg/resources/machine/common.go
```diff
@@ -224,6 +226,32 @@ func getDigitaloceanProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpe
...
+func getGCPProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc provider.DatacenterMeta) (*runtime.RawExtension, error) {
+	config := gce.CloudProviderSpec{
+		Zone:        providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.Zone},
+		MachineType: providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.MachineType},
+		DiskSize:    nodeSpec.Cloud.GCP.DiskSize,
+		DiskType:    providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.DiskType},
+	}
+
+	config.Tags = nodeSpec.Cloud.GCP.Tags
+	config.Tags = append(config.Tags, fmt.Sprintf("kubernetes-cluster-%s", c.Name))
```
alvaroaleman: Maybe guard this in a check if the tag is already there? Otherwise when ppl update their NodeDeployment, they will probably end up having the tags a lot of time on it :grimacing:

api/cmd/conformance-tests/scenarios_gcp.go
```diff
@@ -0,0 +1,103 @@
...
+	for _, v := range versions {
+		// Ubuntu
+		scenarios = append(scenarios, &gcpScenario{
+			version: v,
+			nodeOsSpec: kubermaticapiv1.OperatingSystemSpec{
+				Ubuntu: &kubermaticapiv1.UbuntuSpec{},
+			},
+		})
+		// CoreOS
+		scenarios = append(scenarios, &azureScenario{
```
alvaroaleman: I believe this should be a `gcpScenario` :P

api/cmd/conformance-tests/scenarios_gcp.go
```diff
@@ -0,0 +1,103 @@
...
+			nodeOsSpec: kubermaticapiv1.OperatingSystemSpec{
+				ContainerLinux: &kubermaticapiv1.ContainerLinuxSpec{
+					// Otherwise the nodes restart directly after creation - bad for tests
+					DisableAutoUpdate: true,
+				},
+			},
+		})
+		// CentOS
+		//TODO: Fix
+		//scenarios = append(scenarios, &azureScenario{
```
alvaroaleman: should be a `gcpScenario` as well

alvaroaleman:
Tests say (If they fail and you click on show some more lines you can get the events on the cluster which is extremely helpful for debugging):

> Warning  ReconcilingError  48s (x15 over 2m11s)  kubermatic_kubernetes_controller  unknown node dataceter set 'gcp-europe-west3-a'

I believe the datacenter name should be `gcp-westeurope`. Its the used in the `datacenters.yaml`

nikhita:
~~There is no e2e-gce secret in the prow cluster :/~~

edit: ignore this :grimacing:

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if cluster.Spec.Cloud.GCP.FirewallRuleName == "" {
+		if err := createFirewallRule(*cluster.Spec.Cloud.GCP, cluster.Name); err != nil {
+			return nil, err
+		}
+
+		if !kuberneteshelper.HasFinalizer(cluster, firewallCleanupFinalizer) {
+			cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
+				cluster.Finalizers = append(cluster.Finalizers, firewallCleanupFinalizer)
```
alvaroaleman: Please also set `cluster.Spec.Cloud.GCP.FirewallRuleName` here, otherwise we will create the rule over and over again

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,141 @@
...
+	svc, err := connectToComputeService(spec.ServiceAccount)
+	if err != nil {
+		return err
+	}
+
+	firewallService := compute.NewFirewallsService(svc)
+	tag := fmt.Sprintf("kubernetes-cluster-%s", clusterName)
+	firewallName := fmt.Sprintf("firewall-%s", clusterName)
+
+	firewallService.Insert(spec.ProjectID, &compute.Firewall{
```
alvaroaleman: This returns a `*FirewallInsertCall` whose ` .Do` you have to call in order for the thing to actually happen

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,141 @@
...
+		Name:       firewallName,
+		Network:    spec.Network,
+		TargetTags: []string{tag},
+	})
+	spec.FirewallRuleName = firewallName
```
alvaroaleman: This is a no-op as you pass in a literal and not a pointer. Maybe just return the `firewallName` and set it in the same call where you also set the finalizer?

api/pkg/api/v1/types.go
nikhita: Should this be Zone instead of Region? In `api/pkg/resources/cloudconfig/configmap.go`, to update the ProjectID and LocalZone in CloudConfig, the zone info will need to come from datacenterspec.

We don't really use this region info anywhere too.

alvaroaleman: This is a bit complicated. I think it should be region, the tricky part is thought, that users will be allowed to create machines in any zone of the region. This means that the configmap must include all zones of the region.

To not make this too easy, the zones are not always `a,b,c`: https://cloud.google.com/compute/docs/regions-zones/

Maybe we could add the allowed zones in the datacenter spec? WDYT @mrIncompetent ?

mrIncompetent: For simplicity:
Replace `Region` with `Zones []string`

Use the first zone of the slice for `ProjectID` `LocalZone`.

api/pkg/resources/controllermanager/deployment.go
```diff
@@ -182,7 +182,6 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
 		"--cluster-signing-cert-file", "/etc/kubernetes/pki/ca/ca.crt",
 		"--cluster-signing-key-file", "/etc/kubernetes/pki/ca/ca.key",
 		"--cluster-cidr", cluster.Spec.ClusterNetwork.Pods.CIDRBlocks[0],
-		"--configure-cloud-routes=false",
```
nikhita, line 185: @mrIncompetent can you elaborate on why this was false for all providers?

cc @alvaroaleman

mrIncompetent: It caused AWS nodes to get bad NodeConditions &  we don't need those routes as we use Flannel as overlay network for Pods.

We might wan't to reconsider this in the future to allow customers to use the native AWS/GCP networking, but i would like to keep it that way until someone actually requests this.

api/pkg/api/v1/types.go
alvaroaleman: Then we have to validate that all Zones are in the same Region.

> Use the first zone of the slice for ProjectID LocalZone
Seems that will work: https://github.com/kubernetes/kubernetes/blob/014a2d2552ba36135dbb292d6ba0831b8023524a/pkg/cloudprovider/providers/gce/gce.go#L182

We also have to set `multizone: true` or `regional: true` it seems:
* https://github.com/kubernetes/kubernetes/blob/014a2d2552ba36135dbb292d6ba0831b8023524a/pkg/cloudprovider/providers/gce/gce.go#L114
* https://github.com/kubernetes/kubernetes/blob/014a2d2552ba36135dbb292d6ba0831b8023524a/pkg/cloudprovider/providers/gce/gce.go#L338

api/pkg/resources/serviceaccountsecret.go
```diff
@@ -0,0 +1,23 @@
...
+// ServiceAccountSecretCreator returns a creator function to create a Google Service Account.
+func ServiceAccountSecretCreator(serviceAccount string) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return GoogleServiceAccountSecretName, func(se *corev1.Secret) (*corev1.Secret, error) {
+			se.Type = corev1.SecretTypeOpaque
+
+			if se.Data == nil {
+				se.Data = map[string][]byte{}
+			}
+			se.Data["serviceAccount"] = []byte(serviceAccount)
```
alvaroaleman: This must be base64 decoded first

api/pkg/controller/cluster/resources.go
```diff
@@ -212,6 +212,10 @@ func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconcili
 		creators = append(creators, apiserver.DexCACertificateCreator(data.GetDexCA))
 	}
 
+	if data.Cluster().Spec.Cloud.GCP != nil {
+		creators = append(creators, resources.ServiceAccountSecretCreator(data.Cluster().Spec.Cloud.GCP.ServiceAccount))
```
alvaroaleman, line 216: Please also add this to the openshift controller

api/pkg/resources/serviceaccountsecret.go
nikhita: I think the `GOOGLE_APPLICATION_CREDENTIALS` would work with the base64 encoded value :thinking: 

Ref: https://stackoverflow.com/a/47023291/5832720

nikhita: This worked on dev though :thinking:

alvaroaleman: It also worked on dev before there was any secret at all, so I'd argue thats a correlation and not an indication that this change works as intended:P

The difference in the stackoverflow awnswer is that they create a secret from a yaml file. That means you have to base64-encode your data before putting it into the yaml.

We use however the go types themselves and the associated marshalling already does the bsae64 encoding. Which means our secret will get base64 encoded twice, which means the file the `GOOGLE_APPLICATION_CREDENTIALS` var points to is still base64 encoded once and wont work.

nikhita: Done.

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -345,6 +346,15 @@ type PacketCloudSpec struct {
 	BillingCycle string `json:"billingCycle"`
 }
 
+// GCPCloudSpec specifies access data to GCP.
+type GCPCloudSpec struct {
+	ServiceAccount   string `json:"serviceAccount"`
+	Network          string `json:"network"`
+	Subnetwork       string `json:"subnetwork"`
+	Preemptible      bool   `json:"preemptible"`
```
mrIncompetent: Why is ` Preemtible` on the cluster level ?

api/pkg/provider/datacenter.go
```diff
@@ -94,6 +94,14 @@ type PacketSpec struct {
 	Facilities []string `yaml:"facilities"`
 }
 
+// GCPSpec describes a GCP datacenter
+type GCPSpec struct {
+	Region       string   `yaml:"region"`
+	ZoneSuffixes []string `yaml:"zone_suffixes"`
+	MultiZone    bool     `yaml:"multizone,omitempty"`
```
mrIncompetent: Isn't this implicitly set when specifying multiple zones ?

api/pkg/resources/machine/common.go
```diff
@@ -254,6 +256,45 @@ func getPacketProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
...
+	tag := fmt.Sprintf("kubernetes-cluster-%s", c.Name)
+	var exists bool
+	for _, t := range config.Tags {
+		if t == tag {
+			exists = true
+		}
+	}
+	if !exists {
+		config.Tags = append(config.Tags, tag)
+	}
```
mrIncompetent: Replace this with a Set:
```go
tags := sets.NewString(nodeSpec.Cloud.GCP.Tags...)
tags.Insert(fmt.Sprintf("kubernetes-cluster-%s", c.Name))
config.Tags = tags.List()
```

It covers the deduplication based on the nature of a [Set](https://en.wikipedia.org/wiki/Set_(abstract_data_type))

nikhita:
```
Flag --port has been deprecated, see --secure-port instead.
I0510 12:35:01.234465       1 serving.go:319] Generated self-signed cert in-memory
I0510 12:35:01.817011       1 controllermanager.go:155] Version: v1.14.0
I0510 12:35:01.817837       1 secure_serving.go:116] Serving securely on [::]:10257
I0510 12:35:01.818134       1 leaderelection.go:217] attempting to acquire leader lease  kube-system/kube-controller-manager...
I0510 12:35:19.053384       1 leaderelection.go:227] successfully acquired lease kube-system/kube-controller-manager
I0510 12:35:19.053897       1 event.go:209] Event(v1.ObjectReference{Kind:"Endpoints", Namespace:"kube-system", Name:"kube-controller-manager", UID:"7bf121d4-731f-11e9-ac81-8ad0ff3aa8b3", APIVersion:"v1", ResourceVersion:"780", FieldPath:""}): type: 'Normal' reason: 'LeaderElection' controller-manager-6b98d485d-p7mhg_07c79ce5-7320-11e9-a83b-7e1387a6c8a8 became leader
W0510 12:35:19.428412       1 plugins.go:118] WARNING: gce built-in cloud provider is now deprecated. The GCE provider is deprecated and will be removed in a future release
I0510 12:35:19.430074       1 gce.go:271] Using GCE provider config &{Global:{TokenURL: TokenBody: ProjectID:kubermatic-dev NetworkProjectID: NetworkName:global/networks/default SubnetworkName: SecondaryRangeName: NodeTags:[] NodeInstancePrefix: Regional:false Multizone:false APIEndpoint: ContainerAPIEndpoint: LocalZone:europe-west3-c AlphaFeatures:[]}}
I0510 12:35:19.430234       1 gce.go:882] Using existing Token Source &oauth2.reuseTokenSource{new:google.computeSource{account:""}, mu:sync.Mutex{state:0, sema:0x0}, t:(*oauth2.Token)(nil)}
E0510 12:35:19.570855       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:20.571117       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:21.971461       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:23.931668       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:26.676184       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:30.518144       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:35.898096       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:43.428261       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:35:53.973085       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
E0510 12:36:08.731242       1 gce.go:893] error fetching initial token: oauth2/google: can't get a token from the metadata service; not running on GCE
F0510 12:36:08.731404       1 controllermanager.go:213] error building controller context: cloud provider could not be initialized: could not init cloud provider "gce": timed out waiting for the condition
```

:disappointed: 

Looks like controller-manager can't see `GOOGLE_APPLICATION_CREDENTIALS`...checking why.

nikhita:
/test pull-kubermatic-e2e-gcp-coreos-1.14

want to check something

nikhita:
Rebased and cleaned up commits. PTAL.

The GCE test fails because kubermatic is not able to cleanup the cluster because the cleanup is done by Kubermatic installed from master, but passes otherwise.
Relevant logs: https://prow.loodse.com/view/gcs/prow-data/pr-logs/pull/kubermatic_kubermatic/3350/pull-kubermatic-e2e-gcp-coreos-1.14/1127855506614915074

I'll remove the cluster corresponding to the above test manually, but will do so after this gets lgtm and approved in case anyone wants to take a look at the logs. :)

alvaroaleman:
> The GCE test fails because kubermatic is not able to cleanup the cluster because the cleanup is done by Kubermatic installed from master, but passes otherwise.
Relevant logs: https://prow.loodse.com/view/gcs/prow-data/pr-logs/pull/kubermatic_kubermatic/3350/pull-kubermatic-e2e-gcp-coreos-1.14/1127855506614915074

That is actually not true, the deletion is not considered at all for the result of the tests here...The test you linked failed because of `/api/hack/ci-run-minimal-conformance-tests.sh: line 242: 8913 Killed timeout -s 9 90m ./conformance-tests $EXTRA_ARGS -debug -worker-name=$BUILD_ID -kubeconfig=$KUBECONFIG` because the tests exceeded the 90m timeout

/test pull-kubermatic-e2e-gcp-coreos-1.14

nikhita:
/test pull-kubermatic-e2e-gcp-coreos-1.14

:crossed_fingers:

api/pkg/api/v1/types.go
```diff
@@ -691,6 +713,19 @@ type PacketNodeSpec struct {
...
 
+// GCPNodeSpec gcp specific node settings
+// swagger:model GCPNodeSpec
+type GCPNodeSpec struct {
+	Zone                  string            `json:"zone"`
+	MachineType           string            `json:"machineType"`
+	DiskSize              int64             `json:"diskSize"`
+	DiskType              string            `json:"diskType"`
+	Preemptible           bool              `json:"preemptible"`
+	AssignPublicIPAddress bool              `json:"assignPublicIPAddress"`
```
alvaroaleman: Does it make sense to expose this in the API? It is not configurable and must be `true`

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,179 @@
...
+func (g *gcp) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
+	if spec.GCP.ServiceAccount == "" {
+		return fmt.Errorf("serviceAccount cannot be empty")
+	}
+	return nil
+}
+
+// CleanUpCloudProvider removes firewall rules and related finalizer.
+func (g *gcp) CleanUpCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
+	for _, finalizer := range cluster.Finalizers {
```
alvaroaleman: Can we replace this loop with a `if !kuberneteshelper.HasFinalizer(cluster, firewallCleanupFinalizer)`?

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -0,0 +1,179 @@
...
+	firewallService := compute.NewFirewallsService(svc)
+	tag := fmt.Sprintf("kubernetes-cluster-%s", clusterName)
+	firewallName := fmt.Sprintf("firewall-%s", clusterName)
+
+	_, err = firewallService.Insert(projectID, &compute.Firewall{
+		Name:    firewallName,
+		Network: spec.Network,
+		Allowed: []*compute.FirewallAllowed{
+			{
+				IPProtocol: "tcp",
```
alvaroaleman, line 163: Can we also create rules for the `icmp, esp, ah, sctp, ipip` protocols? Ref: https://cloud.google.com/vpc/docs/firewalls#protocols_and_ports

api/pkg/resources/cloudconfig/configmap.go
```diff
@@ -156,6 +199,7 @@ const (
 	// It is required when activating the vsphere cloud-provider in the controller
 	// manager on a non-ESXi host
 	// Upstream issue: https://github.com/kubernetes/kubernetes/issues/65145
-	FakeVMWareUUIDKeyName = "fakeVmwareUUID"
-	fakeVMWareUUID        = "VMware-42 00 00 00 00 00 00 00-00 00 00 00 00 00 00 00"
+	FakeVMWareUUIDKeyName          = "fakeVmwareUUID"
+	fakeVMWareUUID                 = "VMware-42 00 00 00 00 00 00 00-00 00 00 00 00 00 00 00"
+	GoogleServiceAccountVolumeName = "google-service-account-volume"
```
alvaroaleman: Can we define this in the `resources` package instead? Its not even used here

alvaroaleman (approved):
Last round of nits. GCP presubmit is green :tada:

nikhita:
/test pull-kubermatic-e2e-gcp-coreos-1.14

Addressed comments. Since there are some new changes, going to kick off the gcp tests again for a sanity check before merging.

nikhita:
oh yeah, this will need https://github.com/kubermatic/machine-controller/pull/533

api/pkg/resources/machine/common.go
```diff
@@ -258,14 +258,13 @@ func getPacketProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
...
-		Preemptible:           providerconfig.ConfigVarBool{Value: nodeSpec.Cloud.GCP.Preemptible},
-		Network:               providerconfig.ConfigVarString{Value: c.Spec.Cloud.GCP.Network},
-		Subnetwork:            providerconfig.ConfigVarString{Value: c.Spec.Cloud.GCP.Subnetwork},
+		Zone:        providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.Zone},
+		MachineType: providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.MachineType},
+		DiskSize:    nodeSpec.Cloud.GCP.DiskSize,
+		DiskType:    providerconfig.ConfigVarString{Value: nodeSpec.Cloud.GCP.DiskType},
+		Preemptible: providerconfig.ConfigVarBool{Value: nodeSpec.Cloud.GCP.Preemptible},
+		Network:     providerconfig.ConfigVarString{Value: c.Spec.Cloud.GCP.Network},
+		Subnetwork:  providerconfig.ConfigVarString{Value: c.Spec.Cloud.GCP.Subnetwork},
```
alvaroaleman: You should still set `AssignPublicIPAddress` here, just don't expose it as part of the `nodeSpec` via the API :)

nikhita:
Updated, PTAL.

nikhita:
/assign @maciaszczykm 
for approval for the ui/api part :)

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -345,6 +346,14 @@ type PacketCloudSpec struct {
 	BillingCycle string `json:"billingCycle"`
 }
 
+// GCPCloudSpec specifies access data to GCP.
+type GCPCloudSpec struct {
```
maciaszczykm, line 350: Seems like validation is not implemented for it: https://github.com/kubermatic/kubermatic/blob/gcp-support/api/pkg/validation/cluster.go#L190

nikhita: Thanks. Done :)

maciaszczykm:
/approve

Lint may fail though with validation method update :/

api/pkg/crd/kubermatic/v1/cluster.go
maciaszczykm: Np :)

api/pkg/validation/cluster.go
```diff
@@ -283,5 +286,9 @@ func ValidateCloudSpec(spec kubermaticv1.CloudSpec, dc provider.DatacenterMeta)
 		return nil
 	}
 
+	if spec.GCP != nil && spec.GCP.ServiceAccount != "" {
```
alvaroaleman: Should be `spec.GCP.ServiceAccount == ""`

nikhita: Done

PR #3355, opened 2019-04-26
add openshift support

### LINKED ISSUE #3371: Add OpenShift support in kubermatic API
**User Story**
A user can specify what kind of cluster can create: k8s or openshift
**Acceptance criteria**
`createCluster` endpoint additionally expects cluster type. Depends on the type the k8s or OpenShift cluster is created

zreigz:
**What this PR does / why we need it**: add support for OpenShift

Fixes #3371

```release-note
Add support for OpenShift. To create OpenShift cluster the `type` must be set for `openshift`. If `type` is missing or empty the kubernetes cluster will be created.
{"cluster":{"name":"cluster-name","type":"openshift", ....}
```


api/pkg/api/v1/types.go
```diff
@@ -327,6 +327,7 @@ type CreateClusterSpec struct {
 // swagger:model Cluster
 type Cluster struct {
 	ObjectMeta `json:",inline"`
+	Type       string        `json:"type"`
```
p0lyn0mial, line 337: Is `Type` only specific to `clusters` ? 

If not we could create `TypeMeta` and inline it or move `Type` to `ObjectMeta`, thoughts ?

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -43,12 +43,25 @@ const (
 	nodeDeploymentCreationStart   NodeDeploymentEvent = "NodeDeploymentCreationStart"
 	nodeDeploymentCreationSuccess NodeDeploymentEvent = "NodeDeploymentCreationSuccess"
 	nodeDeploymentCreationFail    NodeDeploymentEvent = "NodeDeploymentCreationFail"
+
+	openShiftClusterType  = "openshift"
```
p0lyn0mial: Those should be defined closer to the type (`api.Cluster`)

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -43,12 +43,25 @@ const (
...
+	kubernetesClusterType = "kubernetes"
 )
 
+// serviceAccountGroupsPrefixes holds a list of groups with prefixes that we will generate RBAC Roles/Binding for service account.
```
p0lyn0mial: update the comment.

api/pkg/provider/kubernetes/cluster.go
```diff
@@ -72,7 +72,7 @@ type ClusterProvider struct {
 }
 
 // New creates a brand new cluster that is bound to the given project
-func (p *ClusterProvider) New(project *kubermaticapiv1.Project, userInfo *provider.UserInfo, spec *kubermaticapiv1.ClusterSpec) (*kubermaticapiv1.Cluster, error) {
+func (p *ClusterProvider) New(project *kubermaticapiv1.Project, userInfo *provider.UserInfo, spec *kubermaticapiv1.ClusterSpec, clusterType string) (*kubermaticapiv1.Cluster, error) {
```
p0lyn0mial: I'd like to keep this interface more generic thus:
 - we could create a new obj that would hold `ClusterSpec` and `Annotation` (map)
- we could pass a `func(c *Cluster)` that would modify the resource before saving in the database

wdyt ?

p0lyn0mial (commented):
What about the update ? Users are not allowed to change the type of existing cluster ?

p0lyn0mial:
/hold

as discussed in slack we want to change one more thing.

PR #3384, opened 2019-04-30
Fix cert-manager CRDs

xrstf:
The CRDs require the Helm hooks to be installed before any other resource. Otherwise Kubermatic 2.9 -> 2.10 upgrades will fail.

```release-note
NONE
```

/cherrypick release/v2.10


kron4eg (approved):
/lgtm
/approve

Issue #3344, opened 2019-04-25
Support dedicated LB per user cluster


mrIncompetent:
## User Story
To avoid having the NodePort proxy LB as [SPOF](https://de.wikipedia.org/wiki/Single_Point_of_Failure) we should implement a second mode where each user clusters gets its own LB.

The "mode" should be made configurable via a flag of the kubermatic controller-manager:
```
-apiserver-expose-strategy=[LB|NodePort]
```

#### `-apiserver-expose-strategy=LB`
For each user cluster the `external-apiserver` service will be of type `LoadBalancer`.
To avoid complexity we won't utilize DNS. Thus `cluster.Address.ExternalName` must contain `service.Spec.LoadBalancerIP` of the `external-apiserver` service.

Note: `service.Spec.LoadBalancerIP` is typically an IP address, but on AWS we get a DNS record.

#### `-apiserver-expose-strategy=NodePort`
For each user cluster the `external-apiserver` service will be of type `NodePort`.
This is what we currently do.

**This change must be backwards compatible**
To be able to not break existing clusters, we should make the cluster-controller smart enough not try to change `service.ServiceType` to something different.

**Acceptance criteria**
- [ ] Support LB & NodePort as exposing mode for user clusters
- [ ] Docs have been updated to explain both modes & how to configure them
- [ ] The kubermatic helm chart has a values.yaml flag to configure `-apiserver-expose-strategy` on the kubermatic controller-manager

Unit tests exist which cover: 
- [ ] backwards compatibility
- [ ] Service creation for both modes
- [ ] Address handling for both modes

alvaroaleman:
@mrIncompetent Update this

mrIncompetent:
### Update
What we missed in the previous design was that we have 2 services which we export to the outside:
- APIServer
- OpenVPN server

To avoid having 2 LB's per cluster, we'll use envoy as TCP proxy.
Since we now have a very similar setup like with the NodePort proxy, we'll try to reuse it.

To allow reuse of the NodePort proxy we must change it, so it can work on a single namespace.


The overall design will be:
```
			 -> APIserver
			/
LB	->	Envoy(NodePortProxy)	
			\
			 -> OpenVPN
```

irozzo-1A:
> To avoid having the NodePort proxy LB as SPOF we should implement a second mode where each user clusters gets its own LB.

[...]

PR #3387, opened 2019-05-01
Allow exposing the usercluster APIServer via a service type LoadBalancer

alvaroaleman:

Goes towards fixing #3344 

**Documentation**:

```release-note
It is now possible to configure Kubermatic to create one service of type LoadBalancer per user cluster instead of exposing all of them via the nodeport-proxy on one central LoadBalancer service
```


alvaroaleman:
/assign @mrIncompetent 

Could you please test if `-apiserver-expose-strategy=LoadBalancer` actually works by creating a cluster and some machines in it? It is not testable on hcloud due to the lack of a LoadBalancer implementation and my connection is still too slow to run the kubermatic-controller-manager locally :/

alvaroaleman:
And should we cherrypick this into 2.10?

mrIncompetent:
> And should we cherrypick this into 2.10?

Yes, definitely.

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -92,6 +95,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.StringVar(&c.inClusterPrometheusRulesFile, "in-cluster-prometheus-rules-file", "", "The file containing the custom alerting rules for the prometheus running in the cluster-foo namespaces.")
 	flag.BoolVar(&c.inClusterPrometheusDisableDefaultRules, "in-cluster-prometheus-disable-default-rules", false, "A flag indicating whether the default rules for the prometheus running in the cluster-foo namespaces should be deployed.")
 	flag.StringVar(&c.dockerPullConfigJSONFile, "docker-pull-config-json-file", "config.json", "The file containing the docker auth config.")
+	flag.StringVar(&c.apiserverExposeStrategy, "apiserver-expose-strategy", "NodePort", "The strategy to expose the apiserver with, either NodePort which creates a NodePort with a `nodeport-proxy.k8s.io/expose: true` annotation or LoadBalancer, which creates a LoadBalancer")
```
mrIncompetent: Use a local variable for the "raw" flag input. That way you can also rename `apiServerExposeStrategyParsed` to `apiServerExposeStrategy`

api/pkg/controller/cluster/address_test.go
```diff
@@ -1,122 +0,0 @@
...
-	const testDomain = "dns-test.kubermatic.io"
-	ip, err := address.GetExternalIPv4(testDomain)
-	if err != nil {
-		t.Fatalf("failed to get the external IPv4 address for %s: %v", testDomain, err)
-	}
-
-	if ip != "192.168.1.1" {
-		t.Fatalf("expected to get 192.168.1.1. Got: %s", ip)
-	}
-}
```
mrIncompetent, line 32: I would like to keep that test :grimacing:

api/pkg/controller/cluster/address_test.go
```diff
@@ -1,122 +0,0 @@
...
-		t.Fatalf("external name is wrong. Expected=%s Got=%s", expectedExternalName, cluster.Address.ExternalName)
-	}
-
-	expectedURL := fmt.Sprintf("https://%s:%d", cluster.Address.ExternalName, TestExternalPort)
-	if cluster.Address.URL != expectedURL {
-		t.Fatalf("url is wrong. Expected=%s Got=%s", expectedURL, cluster.Address.URL)
-	}
-}
-
-func TestSeedDNSOverride(t *testing.T) {
```
mrIncompetent, line 80: We should document that it's not possible to use the seed dns override feature when using an LB

api/pkg/controller/cluster/address_test.go
alvaroaleman: I just moved it, it is now `api/pkg/resources/address/address_test.go`, I thought it was a bit silly to export that func just to make it testable from another package

api/pkg/controller/cluster/resources.go
```diff
@@ -36,6 +36,11 @@ func (r *Reconciler) ensureResourcesAreDeployed(ctx context.Context, cluster *ku
 		return err
 	}
 
+	// Set the hostname & url
+	if err := r.syncAddress(ctx, cluster); err != nil {
```
mrIncompetent, line 40: I don't think there's a big benefit in moving `syncAddress` right after the service reconciling.
There won't be any LB IP right after the creation of the service - so we would need to requeue anyway. 

I would prefer keeping it out of `ensureResourcesAreDeployed` for readability

alvaroaleman: I completely understand the sentiment and agree in that it doesn't fit very well into `resources.go`, the problem is that we can only do it _after_ the `Services` have been synced because:

* `SyncAddress` must know what type of service we have (LoadBalancer or NodePort)
* Thats why `SyncAddress` cant cope with `kerros.IsNotFound` anymore as it did before
* That is why `SyncAddress` has to come after Service reconciliation

If you have a better idea than putting it into `resources.go` regarding structure, I am all ears

mrIncompetent: sgtm, thanks for explaining!

mrIncompetent:
For some reason, this PR causes most Deployments to change after the first sync.
Which leads to almost every Deployment to start 2 times the pods which are intented,

I'm investigating...

api/pkg/resources/address/address.go
```diff
@@ -34,66 +32,90 @@ func SyncClusterAddress(ctx context.Context,
...
-		return nil, err
+	if len(service.Spec.Ports) < 1 {
+		return nil, fmt.Errorf("service %q has no port configured", serviceKey.String())
 	}
 
-	nodePort := service.Spec.Ports[0].NodePort
-	if cluster.Address.Port != nodePort {
+	// Port
+	port := service.Spec.Ports[0].Port
+	if service.Spec.Type == corev1.ServiceTypeNodePort {
```
alvaroaleman, line 96: @mrIncompetent Does that actually fix the issue? We have a race here between the NodePort controller and our controller. Maybe add a short `wait.Poll` that checks that `NodePort != ""` is when we have a NodePort service?

mrIncompetent: Yes it fixes the issue.

The controller creates the Service with just port 443 & waits 1 sync to update the port to point to the NodePort which got allocated. 
Which means Port is 443 after the first sync but points to the NodePort on the second sync.

alvaroaleman: Yeah but during the first sync we:

* Create the Services
* Sync the address (and use whatever is in the lister cache as basis for that)

So if the nodeport controller doesn't manage to set the nodeport in the timeframe between `We create the service` and `we sync the address` (Or that change is not in our cache yet)  we change the secrets and subsequentely the deployments we created during the first sync or do I miss something?

mrIncompetent: The API server allocates the NodePort

mrIncompetent:
As we missed considering the OpenVPN server in this PR i was thinking about possibilities.
- Create a LB service without selector & manage the `Endpoints` object
  - This has the disadvantage that we would need to take the Pod IP's as ClusterIP's are [not supported](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors) 
  - Also we probably would want to write a dedicate controller for this task to make it work independent of other controller failures
- Adapt the NodePortProxy and deploy it for each cluster (Scoped to the cluster's namespace)
  - Would involve minimal changes on the NodePortProxy side
  - We would not loose the envoy metrics

I'm favoring the latter.
@alvaroaleman

mrIncompetent:
Got feedback from a customers now:
They would really like to have a single IP & the metrics

PR #3389, opened 2019-05-01
remove ownerRef from VPA pointing to a cluster

mrIncompetent:
This PR removes the cluster OwnerRef on VPA objects.
Its not needed because we only wan't to have the target of the VPA as owner set (Deployment/Statefulset).

```release-note
NONE
```


alvaroaleman (commented):
/lgtm
/approve

PR #3390, opened 2019-05-01
Scrape cAdvisor metrics for user clusters & disable the graceful termination of Prometheus

mrIncompetent:
Scrape cAdvisor metrics for user clusters & disable the graceful termination of Prometheus

```release-note
cAdvisor metrics are now being scraped for user clusters
```


alvaroaleman (commented):
Why disable the graceful termination? Doesn't that increase the risk of a broken tsdb?

mrIncompetent:
> Why disable the graceful termination? Doesn't that increase the risk of a broken tsdb?

We use an emptyDir for storage

mrIncompetent:
/hold 
i screwed up the yaml formatting

PR #3391, opened 2019-05-01
[WIP] Replace the metrics-server with the prometheus metrics-adapter

mrIncompetent:
Replace the metrics-server with the prometheus metrics-adapter.

This PR has the following advantages:
- We have the cAdvisor metrics in Prometheus so we can utilize them (Dashboards / Alerts)
- With the cAdvisor metrics in Prometheus we avoid the overhead of the metrics-server own scraping(Scrapes the kubelet every 30sec). 
- Since the prometheus-metrics-adapter only talks to Prometheus, we can avoid having to run the OpenVPN sidecars - Reducing the number of containers per cluster by 4.

For concerns about stability/support:
RedHat uses it for OpenShift 4

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/cluster/resources.go
+++ b/api/pkg/controller/cluster/resources.go
@@ -231,6 +231,7 @@ func GetConfigMapCreators(data *resources.TemplateData) []reconciling.NamedConfi
 		cloudconfig.ConfigMapCreator(data),
 		openvpn.ServerClientConfigsConfigMapCreator(data),
 		dns.ConfigMapCreator(data),
+		metricsserver.ConfigMapCreator(),
 	}
 }
 
--- a/api/pkg/resources/metrics-server/configmap.go
+++ b/api/pkg/resources/metrics-server/configmap.go
@@ -1,0 +1,57 @@
+package metricsserver
+
+import (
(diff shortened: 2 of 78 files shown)


mrIncompetent:
/hold

Need to check what happens when Prometheus is not available & Update the resource requests/limits.

xrstf:
Can you please extend your PR with reasons as to why you made this change? Is the Prometheus adapter faster? Uses less resources? Is the metrics-server deprecated?

mrIncompetent:
@xrstf my fault, just extracted that changes from a testing PR of mine and haven't spent much time since then on this PR.

I've updated the PR description!

alvaroaleman:
Some openshift folks mentioned on Kubecon barcelona that they want to rewrite the adapter. Should we close this PR @mrIncompetent ?

mrIncompetent:
I've watched the recording.
They only mentioned that they want to rework the configuration & that they want to move it to a different org ?
Unless they mentioned something outside of the recording i would keep this PR until i have some time to test & polish this

mrIncompetent:
Closing as RedHat is currently writing a new version from scratch

PR #3393, opened 2019-05-01
Add Antiaffinity & PDBs to dns-resolver, Kubermatic and the monitoring components

alvaroaleman:

Adds Pod Antiaffinity and PDBs to the dns-resolver, the Kubermatic control plane and the monitoring components.

Fixes #

**Documentation**:

```release-note
Pod AntiAffinity and PDBs were added to the Kubermatic control plane components and the monitoring stack to spread them out if possible and reduce the chance of unavailability
```

### DIFF
--- a/config/kubermatic/templates/kubermatic-api-dep.yaml
+++ b/config/kubermatic/templates/kubermatic-api-dep.yaml
@@ -87,4 +87,14 @@ spec:
 {{ toYaml .Values.kubermatic.api.affinity | indent 8 }}
       tolerations:
 {{ toYaml .Values.kubermatic.api.tolerations | indent 8 }}
+---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: kubermatic-api
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      role: kubermatic-api
 {{ end }}
--- a/config/kubermatic/templates/kubermatic-controller-manager-dep.yaml
+++ b/config/kubermatic/templates/kubermatic-controller-manager-dep.yaml
@@ -173,3 +173,13 @@ spec:
 {{ toYaml .Values.kubermatic.controller.affinity | indent 8 }}
       tolerations:
 {{ toYaml .Values.kubermatic.controller.tolerations | indent 8 }}
+---
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager-v1
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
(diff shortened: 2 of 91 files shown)


config/kubermatic/templates/kubermatic-api-dep.yaml
```diff
@@ -5,6 +5,15 @@ metadata:
   name: kubermatic-api-v1
 spec:
   replicas: {{ .Values.kubermatic.api.replicas }}
+  affinity:
```
xrstf: Deployment spec's have affinity fields? https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#deploymentspec-v1-apps does not mention them.

Also, affinities are meant to be configured in the chart's values.yaml, same for all other charts and we already have *node*AntiAffinities for `kubernetes.io/hostname` -- are those equivalent to your podAntiAffinity?

xrstf (changes requested):

alvaroaleman:
> Deployment spec's have affinity fields? https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#deploymentspec-v1-apps does not mention them.

Ofc not :man_facepalming: Good catch!

> Also, affinities are meant to be configured in the chart's values.yaml, same for all other charts 

Yeah because I was looking for affinities at the wrong place I didn't see that. Fixed it up to be in the `values.yaml` everywhere.

> and we already have nodeAntiAffinities for kubernetes.io/hostname -- are those equivalent to your podAntiAffinity?

I think what you mean is `nodeAffinity` for `kubermatic.io/type`? Thats the only affinity I could find.

But those two are not the same. `nodeAffinity` is used to (not-) co-locate a set of pod on a set of nodes whereas `podAntiAffinity` is used to (not-) co-locate a set of pods with each other. Most of the times the latter is done with the purpose of spreading them out in order to reduce the chance of unavailability when a node fails.

Both of them have a `weight`, I have reduced the `weight` of the `podAntiAffinities` to `10` in all places where there was `nodeAffinity` to make sure the `nodeAffinity` is respected first. Let me know what you think.

config/kubermatic/templates/kubermatic-controller-manager-dep.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: controller-manager-v1
 spec:
   replicas: {{ .Values.kubermatic.controller.replicas }}
+  affinity:
```
xrstf: Leftover?

xrstf (changes requested):

alvaroaleman:
@xrstf Fixed the leftover

xrstf (approved):

PR #3401, opened 2019-05-02
extend versions and updates

### LINKED ISSUE #2421: [Kubermatic API] Extend Versions & updates handling to cover OpenShift versions

zreigz:
**What this PR does / why we need it**: extend versions and updates to support OpenShift cluster type

Fixes #2421

```release-note
extend versions and updates to support OpenShift cluster type.
The optional `type` field was added in `updates.yaml` and `versions.yaml`. By default, the `type` field is set to `kubernetes`  
```

### DIFF
--- a/api/pkg/controller/update/update_controller.go
+++ b/api/pkg/controller/update/update_controller.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/golang/glog"
 
+	"github.com/kubermatic/kubermatic/api/pkg/api/v1"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/semver"
 	"github.com/kubermatic/kubermatic/api/pkg/version"
@@ -34,7 +35,7 @@ type Reconciler struct {
 
 // Manager specifies a set of methods to find suitable update versions for clusters
 type Manager interface {
-	AutomaticUpdate(from string) (*version.MasterVersion, error)
+	AutomaticUpdate(from, clusterType string) (*version.MasterVersion, error)
 }
 
 // Add creates a new update controller
@@ -93,7 +94,12 @@ func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.Cluste
 		return nil
 	}
 
-	update, err := r.updateManager.AutomaticUpdate(cluster.Spec.Version.String())
+	clusterType := v1.KubernetesClusterType
+	if _, ok := cluster.Annotations["kubermatic.io/openshift"]; ok {
+		clusterType = v1.OpenShiftClusterType
+	}
+
+	update, err := r.updateManager.AutomaticUpdate(cluster.Spec.Version.String(), clusterType)
 	if err != nil {
 		return fmt.Errorf("failed to get automatic update for cluster for version %s: %v", cluster.Spec.Version.String(), err)
 	}
--- a/api/pkg/handler/v1/cluster/upgrade.go
+++ b/api/pkg/handler/v1/cluster/upgrade.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/Masterminds/semver"
 	"net/http"
 
+	"github.com/Masterminds/semver"
 	"github.com/go-kit/kit/endpoint"
 
 	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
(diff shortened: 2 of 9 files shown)


config/kubermatic/static/master/versions.yaml
```diff
@@ -2,41 +2,64 @@ versions:
...
+  type: "kubernetes"
 - version: "1.13.4"
   default: false
+  type: "kubernetes"
 - version: "1.13.5"
   default: true
+  type: "kubernetes"
 - version: "v1.14.0"
   default: false
+  type: "kubernetes"
```
alvaroaleman: Cam we maybe just have the `type` default to `kubernetes` in the code so ppl dont have to do anything?

alvaroaleman:
Can we maybe add a sanity check in the code that upgrades from openshift 3.11 are never allowed? Terrible things will happen if ppl try that :grimacing:

config/kubermatic/static/master/versions.yaml
zreigz: yes sure, I can add `kubernetes` as default in `LoadUpdates` and `LoadVersions` methods. Other methods will get initialized structures.

api/pkg/handler/v1/cluster/upgrade.go
```diff
@@ -223,14 +235,62 @@ func UpgradeNodeDeploymentsEndpoint(projectProvider provider.ProjectProvider) en
...
+// clusterType represents a request that contains the cluster type
+type clusterType struct {
+	Type string `json:"type"`
+}
+
+// versionsReq defines HTTP request for getMasterVersions
+// swagger:parameters getMasterVersions
+type versionsReq struct {
+	// in: body
+	Body clusterType
```
p0lyn0mial: usually query parameters are put in the URL address not in the body, do you think we could change it ?

api/pkg/handler/v1/cluster/upgrade.go
```diff
@@ -112,7 +124,7 @@ func GetNodeUpgrades(updateManager common.UpdateManager) endpoint.Endpoint {
 			return nil, fmt.Errorf("failed to parse control plane version: %v", err)
 		}
 
-		versions, err := updateManager.GetMasterVersions()
+		versions, err := updateManager.GetMasterVersions(apiv1.KubernetesClusterType)
```
p0lyn0mial, line 123: yeah, nodes version don't care about `Openshift` cluster type.

api/pkg/handler/v1/cluster/upgrade.go
```diff
@@ -51,21 +52,32 @@ func GetUpgradesEndpoint(updateManager common.UpdateManager, projectProvider pro
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
-		versions, err := updateManager.GetPossibleUpdates(cluster.Spec.Version.String())
+		clusterType := apiv1.KubernetesClusterType
+		isOpenShift, ok := cluster.Annotations["kubermatic.io/openshift"]
+		if ok {
+			if isOpenShift == "true" {
```
p0lyn0mial: are we actually interested in the value ? Checking if the key is present (`kubermatic.io/openshift`) is not enough ?

they code would simplify if we didn't have to check the value.

api/pkg/handler/v1/cluster/upgrade.go
zreigz: what about compatibility with current UI? If we would like to have it in URL I have to create a new endpoint for OpenShift, WDYT? The old one will stay not touched for k8s

p0lyn0mial: it will work exactly the same as with the body. The current UI will not add the query parameter and we will default to `kubernetes`

zreigz: done

api/pkg/handler/v1/cluster/upgrade.go
zreigz: done

zreigz:
@p0lyn0mial @alvaroaleman PTAL

api/pkg/controller/update/update_controller.go
```diff
@@ -93,7 +94,15 @@ func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.Cluste
 		return nil
 	}
 
-	update, err := r.updateManager.AutomaticUpdate(cluster.Spec.Version.String())
+	clusterType := v1.KubernetesClusterType
+	isOpenShift, ok := cluster.Annotations["kubermatic.io/openshift"]
+	if ok {
+		if isOpenShift == "true" {
```
p0lyn0mial: > are we actually interested in the value ? Checking if the key is present (kubermatic.io/openshift) is not enough ?
> they code would simplify if we didn't have to check the value.

here too

api/pkg/version/manager.go
```diff
@@ -46,10 +50,33 @@ func NewFromFiles(versionsFilename, updatesFilename string) (*Manager, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to load updates from %s: %v", updatesFilename, err)
 	}
+	for _, update := range updates {
```
p0lyn0mial, line 53: not sure if this is the best place as it will take the whole server down, wouldn't it be better to fail a request for an update?

zreigz: yes, good idea. Moved to `GetUpgradesEndpoint`

api/pkg/controller/update/update_controller.go
zreigz: done

zreigz:
@p0lyn0mial @alvaroaleman PTAL

alvaroaleman:
/lgtm

Could you update the `release-note` thought, remove the `[Action required]` and add the info that this field is optional and defaults to Kubernetes?

zreigz:
@alvaroaleman PTAL

PR #3402, opened 2019-05-02
compares the content of the sa token with the one in the db

### LINKED ISSUE #3399: regenerating a token doesn't revoke already created.
**Description**: regenerating a token doesn't invalidate already created, as a result all previously issued tokens are valid and can be used.

p0lyn0mial:
**What this PR does / why we need it**: previously token regeneration didn't work as expected because the new tokens didn't revoke already existing ones. The new tokens are almost exactly the same as the previous ones thus we need to compare the content of an incoming token with a token stored in the database.

Fixes #3399 

```release-note
NONE
```


p0lyn0mial:
/hold 

because I made a UT flaky.

zreigz (approved):
/lgtm
/approve

PR #3407, opened 2019-05-03
replace __KUBERMATIC__TAG__ when deploying via Prow

xrstf:
This makes it so that when Prow deployed Kubermatic, we can see the exact version via `helm ls`. The chart-sync script already takes care of that for publicly relased versions, but it's nice to have this for our dev versions as well.

```release-note
NONE
```


xrstf:
@alvaroaleman Does it make more sense to put this into the `hack/deploy.sh` script, so it applies to all clusters? The `GIT_HEAD_HASH` variable is exported, so it should be available in the `deploy.sh`.

alvaroaleman:
Yes, good idea

PR #3410, opened 2019-05-03
update Alertmanager to 0.17.0

xrstf:
Alertmanager 0.17 brings back an API for groups that could be useful for our master monitoring and the only breaking change is referring to the exporting of silences via `amtool`, which we never used.

This PR also changes the configuration to be in-sync with our other charts while still supporting the old `version` field.

```release-note
update Alertmanager to 0.17.0, deprecate version field in favor of image.tag in Helm values.yaml
```


kron4eg (approved):
/lgtm
/approve

PR #3414, opened 2019-05-05
Fix pdb upgrading

alvaroaleman:

Because we can currently not reconcile clusters because the PDB spec for metrics-server was changed

Fixes #

**Documentation**:

```release-note
none
```


api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -385,7 +385,8 @@ func (r *Reconciler) createClusterAccessToken(ctx context.Context, osData *opens
 		nn(metav1.NamespaceSystem, tokenOwnerServiceAccountName),
 		reconciling.ServiceAccountObjectWrapper(tokenOwnerServiceAccountCreator),
 		userClusterClient,
-		&corev1.ServiceAccount{})
+		&corev1.ServiceAccount{},
```
alvaroaleman, line 388: I definitely want to replace all of the `EnsureNamedObject` calls in the openshift controller with the corresponding `ReconcileTYPE` funcs, but since this PR will get cherrypicked I kept it as minimal as possible

PR #3427, opened 2019-05-07
fix user-cluster Prometheus not scraping kube-state-metrics

xrstf:
During refactorings we removed the Service for kube-state-metrics. As the service was the only element that contained Prometheus scraping annotations, we have since then not scraped it anymore.

This PR restores the annotation directly on the Deployment's spec. It also defines a new alert so that we notice the missing scrape target earlier.

```release-note
fixed kube-state-metrics in user-clusters not being scraped
```

### DIFF
--- a/api/pkg/resources/kubestatemetrics/deployment.go
+++ b/api/pkg/resources/kubestatemetrics/deployment.go
@@ -53,6 +53,11 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
 
 			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
 				Labels: podLabels,
+				Annotations: map[string]string{
+					// do not specify a port so that Prometheus automatically
+					// scrapes both the metrics and the telemetry endpoints
+					"prometheus.io/scrape": "true",
+				},
 			}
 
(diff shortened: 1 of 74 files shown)


api/pkg/resources/kubestatemetrics/deployment.go
```diff
@@ -53,6 +53,11 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
 
 			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
 				Labels: podLabels,
+				Annotations: map[string]string{
+					// do not specify a port so that Prometheus automatically
```
alvaroaleman, line 57: @xrstf Prometheus can get multiple targets out of a single annotation? Is that new?

mrIncompetent: Prometheus creates a target for every port:
https://github.com/prometheus/prometheus/blob/master/discovery/kubernetes/pod.go#L229

xrstf: I was surprised as well, but that how it works after adding the annotation :-)

PR #3430, opened 2019-05-07
fix nodeport-proxy nodeAffinity dummy

xrstf:
When a `requiredDuringSchedulingIgnoredDuringExecution` field is set, it must contain at least one node selector term. Thankfully we only ever defined a podAntiAffinity, so we can reduce the amount ot boilerplate for the other affinities to a bare minimum and work around this.

Customers who set their own nodeAffinity will continue to use theirs and everyone else, like us, will get an empty nodeAffinity with *no child nodes* in the YAML.

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve no-issue

PR #3434, opened 2019-05-07
fix forgotten nodePortProxy fallback

xrstf:
One more change even 8 eyes did not see and which I did not notice because without this, you could easily deploy the Chart, even though it would then later fail..

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve no-issue

Issue #3315, opened 2019-04-24
E2E test for SA (API)


p0lyn0mial:
**Describe**: TBD

Tasks:
- [x] 1.Update kind version to v.0.2.1
- [x] 2.Move `deploy.sh` and `expose.sh` to a docker img (https://github.com/kubermatic/dashboard-v2/pull/1231)
- [x] 3.Modify `deploy.sh` to accept an img tag as an argument (https://github.com/kubermatic/dashboard-v2/pull/1235)
- [x] 4.Modify `deploy.sh` to side-load an image to a cluster (https://github.com/kubermatic/dashboard-v2/pull/1235)
- [x] 5.Prepare a script that runs Kubermatic API E2E test in an isolated environment. (https://github.com/kubermatic/kubermatic/pull/3444)
- [x] 6.Prepare a prow job for running API E2E. (https://github.com/kubermatic/infra/pull/338)
- [x] 7.Prepare helpers for getting OIDC token required to auth in the API. (https://github.com/kubermatic/kubermatic/pull/3491)
- [x] 8.Write a few test scenarios (https://github.com/kubermatic/kubermatic/pull/3423)

PR #3444, opened 2019-05-08
adds ci_run_api_e2e scrip for running API E2E test

p0lyn0mial:
**What this PR does / why we need it**: the script is meant to be run in the CI.
It requires `e2e-kind` image that provides deployment and cluster creation scripts.

The script essentially builds a local image that has `Kubermatic` binaries from the current branch.
Pushes the image to the inner `Kubernetes` cluster and uses it to install `Kubermatic`.
Finally, it runs E2E test for the `API`.

See #3315 

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci_run_api_e2e.sh
+++ b/api/hack/ci/ci_run_api_e2e.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
The script will not be run until I prepare a prow job for it. Stay tuned.

PR #3445, opened 2019-05-08
reduce helm history size to 10

mrIncompetent:
This PR reduces the history size for Helm from 100 to 10.

We used 100 before to have sufficient valid releases we could rollback to.
This was required as helm would create a new failed release for each failed deployment - eventually deleting the last valid release, which results in a very bad state (Helm has no idea what is actually deployed...)

Since we now use the `--atomic` flag we can safely reduce that number. 
`--atomic` instructs helm to do a rollback if a deployment fails (Leaving us with a valid release in the end).

```release-note
NONE
```


nikhita:
bumping to check something

PR #3446, opened 2019-05-08
set GODEBUG env to prevent excessive Prometheus memory usage

xrstf:
This works around the increased memory in Prometheus 2.9.x. See https://github.com/prometheus/prometheus/issues/5524 for more details. Version 2.10 is due in two weeks or so and that's more painful than doing a temporary fix on our side.

https://github.com/prometheus/prometheus/issues/5524

```release-note
NONE
```


kron4eg (approved):
/lgtm 
/approve

PR #3450, opened 2019-05-09
simplify backup controller EnqueueRequestsFromMapFunc func

mrIncompetent:
Simplify backup controller EnqueueRequestsFromMapFunc func

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/backup/backup_controller.go
+++ b/api/pkg/controller/backup/backup_controller.go
@@ -116,15 +116,6 @@ func Add(
 		}
 
 		if ownerRef := metav1.GetControllerOf(a.Meta); ownerRef != nil && ownerRef.Kind == kubermaticv1.ClusterKindName {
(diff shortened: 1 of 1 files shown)


api/pkg/controller/backup/backup_controller.go
```diff
@@ -116,15 +116,6 @@ func Add(
 		}
 
 		if ownerRef := metav1.GetControllerOf(a.Meta); ownerRef != nil && ownerRef.Kind == kubermaticv1.ClusterKindName {
-			cluster := &kubermaticv1.Cluster{}
-			err := mgr.GetClient().Get(context.Background(), types.NamespacedName{Name: ownerRef.Name}, cluster)
```
mrIncompetent, line 120: No need to duplicate the logic. 
The controller checks if the cluster exists at the beginning of the sync

PR #3452, opened 2019-05-09
make max block duration in Prometheus configurable

xrstf:
We often have issues with Prometheus not coming up and getting OOMKilled after an unclean shutdown. The OOMKill is likely caused by the fact that Prometheus needs to load the [entire WAL into memory](https://github.com/prometheus/prometheus/issues/4842#issuecomment-447793131) on the next startup. Another argument is that after removing everything from the `wal/` directory, Prometheus comes up with no issues and significantly lower memory usage.

To maybe, hopefully prevent this issue, this PR starts to make use of the max-block-duration flag, in the hope that this will flush more frequently and never leave huge WAL chunks laying around. The default value for this flag was "10% of the retention period", so 36h in our setup.

```release-note
NONE
```

### DIFF
--- a/config/monitoring/prometheus/Chart.yaml
+++ b/config/monitoring/prometheus/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prometheus
-version: 2.0.1
+version: 2.0.2
 appVersion: v2.9.2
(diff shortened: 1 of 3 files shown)


kron4eg (approved):
/lgtm
/approve no-issue

PR #3467, opened 2019-05-10
Avoid enqueuing the same cluster twice

alvaroaleman:
`handler.EnqueueRequestForOwner` enqueues the owning object with the
triggering objects namespace, which doesn't work if the owning object
is non-namespaced.

We worked around this by removing the namespace from clusters in the
`Reconcile`. The problem is that we still have two possible queue keys
for the cluster, allowing two workers to concurrently work on the same
cluster, causing a lot of errors.

This PR changes the whole logic to not use the ownerreference but the
namespace of the objects to find the corresponding cluster.

Fixes #

**Documentation**:

```release-note
none
```

/cherrypick release/v2.10
/assign @mrIncompetent

### DIFF
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -8,6 +8,7 @@ import (
 
 	k8cuserclusterclient "github.com/kubermatic/kubermatic/api/pkg/cluster/client"
 	"github.com/kubermatic/kubermatic/api/pkg/clusterdeletion"
+	controllerutil "github.com/kubermatic/kubermatic/api/pkg/controller/util"
 	kubermaticscheme "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned/scheme"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
@@ -144,8 +145,6 @@ func Add(
 		return err
 	}
 
-	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &kubermaticv1.Cluster{}}
-
 	typesToWatch := []runtime.Object{
 		&corev1.Service{},
 		&corev1.ConfigMap{},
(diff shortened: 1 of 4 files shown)


api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -144,7 +144,20 @@ func Add(
 		return err
 	}
 
-	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &kubermaticv1.Cluster{}}
+	dynamicClient := mgr.GetClient()
+	enqueueClusterForNamespacedObject := &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+		clusterList := &kubermaticv1.ClusterList{}
+		if err := dynamicClient.List(context.Background(), &ctrlruntimeclient.ListOptions{}, clusterList); err != nil {
+			// TODO: Is there a better way to handle errors that occur here?
+			glog.Errorf("failed to list Clusters: %v", err)
```
mrIncompetent: Typically `utilruntime.HandleErr`.
It executes 2 thinks typically:
- Log the error using glog
- Increase our global error metric

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -144,7 +144,20 @@ func Add(
 		return err
 	}
 
-	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &kubermaticv1.Cluster{}}
+	dynamicClient := mgr.GetClient()
+	enqueueClusterForNamespacedObject := &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
```
mrIncompetent: Can we extract the handler for reuse?

alvaroaleman:
> Can we extract the handler for reuse?

So much code saved :sparkles:

api/pkg/controller/util/util.go
```diff
@@ -0,0 +1,31 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// EnqueueClusterForNamespacedObject enqueues the cluster that owns a namespaced object, if any
+// It is used by various controllers to react to changes in the resources in the cluster namespace
+func EnqueueClusterForNamespacedObject(client ctrlruntimeclient.Client) *handler.EnqueueRequestsFromMapFunc {
+	return &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+		clusterList := &kubermaticv1.ClusterList{}
+		if err := client.List(context.Background(), &ctrlruntimeclient.ListOptions{}, clusterList); err != nil {
+			utilruntime.HandleError(fmt.Errorf("failed to list Clusters: %v", err))
```
mrIncompetent, line 22: return here

alvaroaleman:
@mrIncompetent Updated, PTAL

PR #3477, opened 2019-05-13
add support to set node taints during NodeDeployment creation

mrIncompetent:
Add support to set node taints during NodeDeployment creation.

```release-note
NONE
```

### DIFF
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -602,6 +602,13 @@ type NodeVersionInfo struct {
 	Kubelet string `json:"kubelet"`
 }
 
+// TaintSpec defines a node taint
+type TaintSpec struct {
+	Key    string `json:"key"`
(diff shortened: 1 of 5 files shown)


api/pkg/api/v1/types.go
```diff
@@ -593,6 +600,8 @@ type NodeSpec struct {
 	// It will be applied to Nodes allowing users run their apps on specific Node using labelSelector.
 	// required: false
 	Labels map[string]string `json:"labels,omitempty"`
+	// List of taints to set on new nodes
+	Taints []TaintSpec `json:"taints"`
```
p0lyn0mial: I see that this value is not required thus can we tag it as ``json:"taints,omitempty"`` ?

mrIncompetent: Done :)

mrIncompetent:
@p0lyn0mial PTAL

PR #3478, opened 2019-05-13
do not create StatefulSet when migration is enabled

xrstf:
One more leftover from the latest Alertmanager renaming PR. Thankfully it seems that deploying the migration job and the StatefulSet at the same time did not cause any issues in the migration itself.

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve no-issue

PR #3480, opened 2019-05-14
Fix conformance tests for non aws and packet providers

nikhita:
Fix conformance tests for non aws and packet providers.

**Documentation**:

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/api/hack/ci/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci/ci-run-minimal-conformance-tests.sh
@@ -46,6 +46,7 @@ function cleanup {
       kubectl describe cluster -l worker-name=$BUILD_ID|egrep -vi 'APIKey|ProjectID'
     else
       echo "Provider $provider is not yet supported."
+      exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -212,10 +212,10 @@ fi
 
 echodate "Starting conformance tests"
 
-if [[ $provider = "aws" ]]; then
-  EXTRA_ARGS="-aws-access-key-id=${AWS_E2E_TESTS_KEY_ID}
-     -aws-secret-access-key=${AWS_E2E_TESTS_SECRET}"
-elif [[ $provider = "packet" ]]; then
```
alvaroaleman: This does look correct to me, how is it not working?

nikhita: because `EXTRA_ARGS` becomes an unbound variable if the provider is not aws or packet. :(

nikhita: Or wait! We do check for the provider before this. Checking if the logs with this error on slack were stale ones.

nikhita: Updated.

PR #3492, opened 2019-05-15
unify finalizer handling

mrIncompetent:
This PR unifies the finalizer handling across all controllers.
Instead of doing:
```golang
set := strings.NewSet(obj.Finalizers...)
set.Insert("myFancyFinalizer")
obj.Finalizers = set.List()
```
we now have helper functions which do all that:
```golang
kuberneteshelper.AddFinalizer(project, "myFancyFinalizer")
```

The helper functions use `metav1.Object` interface which makes the function applicable for all Kubernetes related objects.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/rbac/sync_project.go
+++ b/api/pkg/controller/rbac/sync_project.go
@@ -7,6 +7,7 @@ import (
 	"github.com/golang/glog"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	kuberneteshelper "github.com/kubermatic/kubermatic/api/pkg/kubernetes"
 
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -74,17 +75,13 @@ func (c *projectController) sync(key string) error {
 }
 
 func (c *projectController) ensureCleanupFinalizerExists(project *kubermaticv1.Project) error {
-	var err error
-	if !sets.NewString(project.Finalizers...).Has(CleanupFinalizerName) {
-		finalizers := sets.NewString(project.Finalizers...)
-		finalizers.Insert(CleanupFinalizerName)
-		project.Finalizers = finalizers.List()
-		project, err = c.masterClusterProvider.kubermaticClient.KubermaticV1().Projects().Update(project)
-		if err != nil {
+	if !kuberneteshelper.HasFinalizer(project, CleanupFinalizerName) {
+		kuberneteshelper.AddFinalizer(project, CleanupFinalizerName)
+		if _, err := c.masterClusterProvider.kubermaticClient.KubermaticV1().Projects().Update(project); err != nil {
 			return err
 		}
 	}
-	return err
+	return nil
 }
 
 func (c *projectController) ensureProjectIsInActivePhase(project *kubermaticv1.Project) error {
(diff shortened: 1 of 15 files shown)


api/pkg/controller/rbac/sync_project.go
```diff
@@ -75,10 +76,8 @@ func (c *projectController) sync(key string) error {
 
 func (c *projectController) ensureCleanupFinalizerExists(project *kubermaticv1.Project) error {
 	var err error
-	if !sets.NewString(project.Finalizers...).Has(CleanupFinalizerName) {
-		finalizers := sets.NewString(project.Finalizers...)
-		finalizers.Insert(CleanupFinalizerName)
-		project.Finalizers = finalizers.List()
+	if !kuberneteshelper.HasFinalizer(project, CleanupFinalizerName) {
+		kuberneteshelper.AddFinalizer(project, CleanupFinalizerName)
 		project, err = c.masterClusterProvider.kubermaticClient.KubermaticV1().Projects().Update(project)
```
mrIncompetent: According to the linter this is an ineffective assignment :/
Though i would expect it to be valid

nikhita: I think the linter complains here because we assign the updated value of the project to `project` but don't use it again.

mrIncompetent: We pass project into the function as pointer.
I would expect that the pointer points to the updated project after this call :thinking:

nikhita: Thought about this, but I can't understand why it complains here. `ineffassign` is known to be one of those projects that never gives any false positives so I'm very confused...but I guess it should be ok to look into this as a follow up and not block this PR based on it.

mrIncompetent: And https://github.com/kubermatic/kubermatic/pull/3492/files#diff-03e32103b3978aac1f57644cad0ce95bL90 does something very similar to this function...

@themue You have any idea ?

themue: Yip. You assign a new project to the parameter `project`. So you, inside of the scope of the function, exchange project and afterwards don't use it.

Like in `ensureProjectIsInActivePhase()` below you only can modify fields inside a passed project and it only lives outside due to the passed reference (which btw is just a copy).

As we have `Update(project)` and that hopefully modifies it the returned reference isn't needed. A

```
_, err = c.masterClusterProvider.kubermaticClient.KubermaticV1().Projects().Update(project)
```

should be enough.

themue (commented):
Just a small change for the linter needed.

api/pkg/controller/rbac/sync_project.go
mrIncompetent: > As we have Update(project) and that hopefully modifies it the returned reference isn't needed.

Nope. `Update` does not modify the `project` - it only marshals it.

Reading @themue comment correct, using:
```golang
_, err = c.masterClusterProvider.kubermaticClient.KubermaticV1().Projects().Update(project)
```
would at least result in the same logic as before?
If so, i would just change the code

themue: Yip, the assignment to `project` does not effect the project passed in the method by the caller.

mrIncompetent: Fixed with https://github.com/kubermatic/kubermatic/pull/3492/commits/97df042f6d5aede51ec36644b539708d3a10a07e

mrIncompetent:
@nikhita @zreigz PTAL

zreigz:
@mrIncompetent please rebase, then I will approve the commit

PR #3515, opened 2019-05-17
restore Elasticsearch compatibility with Kubermatic 2.9

xrstf:
In #3062 I updated the values.yaml structure to accomodate the new ES version, cerebro and stuff. This broke backwards compatibility with Kubermatic 2.9 and lead to the situation where our own cluster is now using the (smallish) default volume size of 10Gi instead of what we actually configured in our secrets repo.

This PR adds extra code to allow Kubermatic 2.9 configuration to still work, so we can then remove the compat stuff in Kubermatic 2.11 (when we also finally switch to ES 7.0).

Thankfully the renamed config keys were in fact [documented](https://docs.kubermatic.io/upgrading/2.9_to_2.10/#explicit-resource-requests-limits), but we should still use 2.10 as a transition period.

```release-note
NONE
```

### DIFF
--- a/config/logging/elasticsearch/templates/es-data-stateful.yaml
+++ b/config/logging/elasticsearch/templates/es-data-stateful.yaml
@@ -10,7 +10,7 @@ spec:
   updateStrategy:
     type: RollingUpdate
   serviceName: es-data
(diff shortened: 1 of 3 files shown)


kron4eg (approved):
/lgtm 
/approve

PR #3517, opened 2019-05-17
fix IAP generating invalid ingresses pointing to other namespaces

xrstf:
It turns out that Ingresses cannot have backends in other namespaces, which is sad for our attempt to passthrough paths directly to the services.

"This one weird trick" fixes it by creating an alias ExternalName service that points to the upstream service.

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve no-issue

PR #3526, opened 2019-05-21
update fluent-bit to 1.1.1

xrstf:
This PR updates fluent-bit to its latest stable version. I wish I could have also updated the pod labels and selector to the new format (especially since the cluster-service label is [deprecated](https://github.com/kubernetes/kubernetes/issues/72757)), but this would cause upgrade issues because the labels are basically immutable.

I did a canary deployment on dev and v1.1.1 seems to work just fine.

```release-note
update fluent-bit to 1.1.1
```


kron4eg (approved):
/lgtm 
/approve no-issue

PR #3527, opened 2019-05-22
update Velero to 1.0.0

xrstf:
This updates Velero to the new stable version. Since the manifests have been removed from the upstream distribution and been replaced with `velero install`, I added a tiny script to at least automatically extract the CRDs in the hopes of never having useless diffs because someone manually copies them over into our chart.

Upgrades from 0.11 are supposed to be easy and require no user interaction. As documented in Kubermatic 2.10, users are strongly supposed to replace Ark 0.x with Velero 0.11, so we have a good basis to update to Velero 1.0 in Kubermatic 2.11.

```release-note
update Velero to 1.0
```

### DIFF
--- a/config/backup/velero/Chart.yaml
+++ b/config/backup/velero/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: velero
-version: 1.0.0
-appVersion: v0.11.0
+version: 1.0.1
+appVersion: v1.0.0
 description: A Helm chart for Heptio Velero
 keywords:
 - kubermatic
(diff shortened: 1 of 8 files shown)


xrstf:
/hold

I have not yet actually deployed this to dev to test the upgrade path.

xrstf:
/hold cancel

Let's do this.

kron4eg (approved):
/lgtm
/approve

/hold

@xrstf please unhold it once you're ready

xrstf:
I deployed it on dev already and it didn't break stuff more than it was broken before. But I wanted to fix stuff before merging and then noticed that doing Prometheus properly (see Thanos PR) is a better way to fix [our] Velero setup.

So I would like to do Thanos first and Velero second.

xrstf:
/hold cancel

Seems to work just fine, let's get this deployed.

kron4eg (approved):
/approve no-issue

PR #3529, opened 2019-05-23
improve authentication layer stability

xrstf:
This PR attempts to make our authentication layer (Dex + Keycloak-Gatekeeper) more stable by

* defining PDBs for both components
* allowing to specify the replica count for Keycloak deployments
* raising the default replica count to 2
* setting affinities to prevent two pods to be scheduled on the same node

```release-note
NONE
```

### DIFF
--- a/config/iap/values.yaml
+++ b/config/iap/values.yaml
@@ -1,7 +1,7 @@
 iap:
   # replicas per deployment; you can set this explicitly per deployment
   # to override this
-  replicas: 1
+  replicas: 2
 
   deployments:
     # alertmanager:
--- a/config/iap/values.yaml
+++ b/config/iap/values.yaml
@@ -1,7 +1,12 @@
 iap:
+  # replicas per deployment; you can set this explicitly per deployment
+  # to override this
+  replicas: 1
+
   deployments:
     # alertmanager:
     #   name: alertmanager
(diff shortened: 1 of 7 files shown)


xrstf:
/hold

Need to check if we can actually define meaningful affinities in this chart.

config/iap/values.yaml
```diff
@@ -1,7 +1,12 @@
 iap:
+  # replicas per deployment; you can set this explicitly per deployment
+  # to override this
+  replicas: 1
+
   deployments:
     # alertmanager:
     #   name: alertmanager
+    #   replicas: 2 #
```
kron4eg, line 9: Is it really needed to have individual replicas?

xrstf: Yes.

kron4eg: But I don’t see how, could you please explain a bit more.

xrstf: There can be IAPs in front of services on preemtible nodes where one or more nodes die, whereas some other IAPs are running in front of apps that are on stable nodes, where only a single replica is enough. That was my motivation for adding this, given how quick and easy it was to do.

kron4eg (approved):
/lgtm 
/approve no-issue

PR #3531, opened 2019-05-25
Add optional Thanos Integration

xrstf:
This adds Thanos to our Prometheus chart. I have done this for at least two reasons:

* Our backups keep failing because they take too long. Removing the need to backup Prometheus itself and instead relying on object stores should remove this pain.
* Managing long-term data in Thanos is much easier than in Prometheus.

This PR is meant as a first step to get our feet wet and learn more about the resource requirements.

The Prometheus StatefulSet gets a new sidecar container that ships the blocks to the block storage *and* handles reloading the Prometheus configuration file. We use this to inject the pod name as a `replica` label which is needed for proper deduplication later. This is why the `config-reloader` container is not added when Thanos is enabled.

There could exist a race condition between Prometheus using the templated config from the sidecar before the sidecar has had time to create it. To prevent it from restarting and leaving a "restarts=1" behind in `k get pods`, I put a tiny waiting loop in front of Prometheus. Funny enough, the sidecar expects a running Prometheus before it templates, but it can deal with Prometheus not being ready and simply retries later and refreshes the config again.

In addition to the sidecars we will deploy two redundant Store Gateways in front of our chosen block storage. They will cache frequently used stuff up to the configurable limits.

For actually running Prometheus queries the Query deployment is used. This uses the headless `thanos-store` service to find all the pieces that provide the Store API (i.e. sidecars and store gateways).

First experiments have shown that we can transparently use Thanos instead of Prometheus inside of Grafana, given that we configure the Prometheus sources a bit more flexibly to include Thanos Query's HTTP port (10902).

After feedback at KubeCon I also removed the option to control the TSDB min/max block duration (with Thanos it's explicitly documented to disable compaction in Prometheus and set the min/max to 2h).

In a later PR I will set the resource constraints. Also, the user-cluster Prometheus installations would also need to be updated to include the sidecar.

```release-note
Added optional Thanos integration to Prometheus for better long-term metrics storage
```

### DIFF
--- a/config/monitoring/prometheus/rules/src/general/thanos.yaml
+++ b/config/monitoring/prometheus/rules/src/general/thanos.yaml
@@ -1,0 +1,71 @@
+groups:
+- name: thanos
+  rules:
+  - alert: ThanosSidecarDown
+    annotations:
+      message: The Thanos sidecar in `{{ $labels.namespace }}/{{ $labels.pod }}` is down.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-thanossidecardown
+    expr: thanos_sidecar_prometheus_up != 1
+    for: 5m
+    labels:
+      severity: warning
+
+  - alert: ThanosSidecarNoHeartbeat
+    annotations:
+      message: The Thanos sidecar in `{{ $labels.namespace }}/{{ $labels.pod }}` didn't send a heartbeat in {{ $value }} seconds.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-thanossidecardown
+    expr: time() - thanos_sidecar_last_heartbeat_success_time_seconds > 60
+    for: 3m
+    labels:
+      severity: warning
+
+  - alert: ThanosCompactorManyRetries
+    annotations:
+      message: The Thanos compactor in `{{ $labels.namespace }}` is experiencing a high retry rate.
+      runbook_url: https://docs.kubermatic.io/monitoring/runbook/#alert-thanoscompactormanyretries
+    expr: sum(rate(thanos_compactor_retries_total[5m])) > 0.01
+    for: 10m
+    labels:
+      severity: warning
+    runbook:
+      steps:
+      - Check the `thanos-compact` pod's logs.
+
+  - alert: ThanosShipperManyDirSyncFailures
+    annotations:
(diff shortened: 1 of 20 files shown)


config/monitoring/prometheus/rules/src/general/thanos.yaml
```diff
@@ -0,0 +1,71 @@
+groups:
```
kron4eg, line 1: shouldn't this somehow take into account that thanos could be disabled?

config/monitoring/prometheus/templates/statefulset.yaml
```diff
@@ -29,25 +34,45 @@ spec:
...
+          set -euo pipefail
+
+          echo "Cleaning up block compaction leftovers..."
+          find /var/prometheus/data -type d -name '*.tmp' -maxdepth 1 -print -exec rm -r {} \;
+
+          {{- if .Values.prometheus.thanos.enabled }}
+          config=/etc/prometheus-shared/prometheus.yaml
+
+          echo "Waiting for $config..."
+          until [ -f $config ]; do sleep 1; done
```
kron4eg: This is good opportunity to never see prometheus started :)

xrstf: How do you mean?

config/monitoring/prometheus/rules/src/general/thanos.yaml
xrstf: How so? The alerts should only be triggered if the metrics they are based on exist. No Thanos metrics = no Thanos alerts, even if the rules are there. At least that was my idea.

config/monitoring/prometheus/templates/statefulset.yaml
xrstf: Would it help to `exit 1` after a minute of waiting? Would that improve a broken a situation where Thanos Sidecar cannot start and cannot generate the prometheus.yaml?

kron4eg: Some emergency brake wouldn't be excessive.

xrstf: Prometheus should now wait for at most one minute and then give up with a more or less helpful error message.

kron4eg (approved):
/lgtm 
/approve

PR #3460, opened 2019-05-10
Improve events for initial NodeDeployment creation

alvaroaleman:
**What this PR does / why we need it**:

Fixes the events for the initial NodeDeployments to:
* Not contain the clusterName, that is implicit by the fact that its created on a Cluster
* Not contain the NodeDeploymentName because its empty

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
none
```

/assign @maciaszczykm

PR #3541, opened 2019-05-28
Improve event message for initial ND

floreks:
Small improvement of event message recorded during initial ND creation.

```release-note
NONE
```

### DIFF
--- a/api/pkg/handler/v1/cluster/cluster.go
+++ b/api/pkg/handler/v1/cluster/cluster.go
@@ -17,6 +17,7 @@ import (
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
+	"k8s.io/client-go/tools/record"
 
 	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
 	kubermaticapiv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
@@ -43,6 +44,7 @@ const (
 	nodeDeploymentCreationStart   NodeDeploymentEvent = "NodeDeploymentCreationStart"
 	nodeDeploymentCreationSuccess NodeDeploymentEvent = "NodeDeploymentCreationSuccess"
 	nodeDeploymentCreationFail    NodeDeploymentEvent = "NodeDeploymentCreationFail"
+	nodeDeploymentCreationRetry   NodeDeploymentEvent = "NodeDeploymentCreationRetry"
 )
 
 // clusterTypes holds a list of supported cluster types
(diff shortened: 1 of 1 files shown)


maciaszczykm:
@alvaroaleman removed it in https://github.com/kubermatic/kubermatic/pull/3460/files. Does it work?

alvaroaleman:
Yeah that name is empty because we use(d?) `GenerateName`

floreks:
@alvaroaleman hmm... `Name` is always empty then or only when the user uses generate name option and then `GenerateName` is filled?

alvaroaleman:
If the user passes in a name it should be filled. You could probably do something along the lines of:

```
ndName := ""
if nd.Name != "" {
    ndName = " "+nd.Name
}
```

and append that to the event message

maciaszczykm:
Either `GenerateName` or `Name` should always be filled. Filling both causes errors on the Kubernetes API side.

`GenerateName` is used when the user has not specified any name in his request.

floreks:
@alvaroaleman @maciaszczykm I have updated logic. PTAL

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -188,9 +191,18 @@ func createInitialNodeDeployment(nodeDeployment *apiv1.NodeDeployment, cluster *
...
+	eventRecorder.Eventf(cluster, corev1.EventTypeNormal, string(nodeDeploymentCreationStart), "started creation of initial node deployment %s", getMachineDeploymentName(md))
 	return client.Create(ctx, md)
 }
 
+func getMachineDeploymentName(md *v1alpha1.MachineDeployment) string {
+	if len(md.Name) != 0 {
+		return md.Name
+	}
+
+	return md.GenerateName
```
maciaszczykm: It will always return `worker-`. Should we indicate somehow that this is only a prefix? Kubernetes appends rest of the name when a resource is saved.

It makes sense when the `md.Name` is set.

floreks: Ahh... so we can't really access fully generated name. Can we return some generic name/info here? Any ideas?

alvaroaleman: You could GET the object from the apiserver, but other than that, no

maciaszczykm: Unfortunately, it is handled by Kubernetes, so we have no information. I'm okay with leaving it like this though. Or something like `md.Name...`, so the users will know that this is only the prefix.

floreks: We could return something like `worker-xxx` just as an indicator that the rest of the name has been generated or simply `worker`.

maciaszczykm: > You could GET the object from the apiserver, but other than that, no

Still, we will be not able to get the name when the generated name is set and the creation failed due to some reason.

maciaszczykm: Or just skip the name part if the generated name is used. If the name is not null then we can show it. I think it would be best, because it appears also in the UI.

floreks: Simply return empty string in here?

maciaszczykm: Yep.

maciaszczykm: ...but in case of error we will get additional space before `:` :P

PR #3554, opened 2019-05-31
Disable insecure versions v1.13.6 & v1.14.2

mrIncompetent:
Both versions are insecure: https://github.com/kubernetes/kubernetes/issues/78308

Keeping both versions commented out will avoid that users manually add those versions.

```release-note
Insecure Kubernetes versions v1.13.6 and v1.14.2 have been disabled.
```

### DIFF
--- a/config/kubermatic/static/master/versions.yaml
+++ b/config/kubermatic/static/master/versions.yaml
@@ -38,12 +38,16 @@ versions:
   default: false
 - version: "1.13.5"
   default: true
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@mrIncompetent Both of them are not in release/v2.10 so that cherrypick is not needed

/lgtm
/approve

mrIncompetent:
@alvaroaleman 
> Keeping both versions commented out will avoid that users manually add those versions.

PR #3593, opened 2019-06-07
do not create empty scrape targets

xrstf:
Prometheus does not like entirely empty scraping configs. When Thanos is disabled, we need to either not load the thanos-sidecar.yaml (via values.yaml, but it's tedious) or create a scraping config with just a name.

```release-note
NONE
```


kron4eg (approved):
/lgtm 
/approve no-issue

Issue #3562, opened 2019-06-03
Allow ICMP in all security group rules we create


alvaroaleman:
**User Story**

In order to be able to cope with an MTU < 1500 anywhere on the route between Kubelet and user cluster APIServer (Internet, Cloudprovider SDN, CNI) we need path MTU discovery. This uses ICMP underneath, which implies that we need to allow that in the security groups we create

**Acceptance criteria**

* All security groups we create at all cloudproviders allow ICMP inbound and outbound
* There is a migration in place, that adds an allow rule for ICMP inbound/outbound to all existing security groups that do not allow ICMP

kdomanski:
Let the migration run in series over the clusters in order to avoid perpetual rate limit issues due to many GETs.

kdomanski:
On GCP there is a default firewall rule that allows ICMP regardless of source and destination. However, this rule can be deleted. Does it make sense to add additional ICMP allow rule for each cluster just to be sure? cc @nikhita @mrIncompetent

alvaroaleman:
Yeah lets add it, our customers don't necesarily leave cloud defaults

PR #3594, opened 2019-06-07
GCP: add a firewall rule to allow ICMP per cluster

kdomanski:
No migration is required, since we haven't shipped GCP support yet.

Part of #3562 for GCP

/meow goggles

```release-note
NONE
```


nikhita (commented):
/lgtm

Will leave approval to Henrik

cc @kgroschoff for the API change

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -185,8 +202,23 @@ func createFirewallRule(spec kubermaticv1.GCPCloudSpec, clusterName string) (str
 		SourceTags: []string{tag},
 	}).Do()
 	if err != nil {
-		return "", err
+		return nil, err
```
mrIncompetent: What happens if the first rule gets created but the second creation fails?

Will the second sync/iteration fail because the first rule already exists?

PR #3606, opened 2019-06-11
Allow exposing userclusters via a LoadBalancer

### LINKED ISSUE #3344: Support dedicated LB per user cluster

alvaroaleman:

* Changes the nodeport proxy to make the annotation configurable
* Changes the nodeport proxy to allow it to work in a namespaced mode
* Extend the cluster CRD with a `.spec.exposeStrategy` field
* `.spec.ExposeStrategy` may be either NodePort or LoadBalancer, if the latter, an additional LoadBalancer service will be allocated and the Nodeport proxy will get deployed in namespaced mode to route the traffic from the LoadBalancer to the apiserver and openvpn server
* Adds a migration to default `.spec.exposeStrategy` to NodePort if unset
* Adds a config option to the api for the expose Strategy and set it on cluster creation

Fixes #3344

**Documentation**:

```release-note
It is now possible to expose userclusters via a distinct LoadBalancer per cluster by setting the `kubermatic.exposeStrategy` option in the Helm `values.yaml` to `LoadBalancer`
```

/assign @mrIncompetent

### DIFF
--- a/api/pkg/resources/address/address.go
+++ b/api/pkg/resources/address/address.go
@@ -39,14 +39,13 @@ func SyncClusterAddress(ctx context.Context,
 		if err := client.Get(ctx, nn, frontProxyLoadBalancerService); err != nil {
 			return nil, fmt.Errorf("failed to get the front-loadbalancer service: %v", err)
 		}
+		// Use this as default in case the implementation doesn't populate the status
+		frontProxyLoadBalancerServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
 		// Supposively there is only one if not..Good luck
 		for _, ingress := range frontProxyLoadBalancerService.Status.LoadBalancer.Ingress {
 			if ingress.IP != "" {
 				frontProxyLoadBalancerServiceIP = ingress.IP
 			}
-		}
-		if frontProxyLoadBalancerService.Spec.LoadBalancerIP != "" {
-			frontProxyLoadBalancerServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
 		}
 	}
 
--- a/api/pkg/resources/address/address.go
+++ b/api/pkg/resources/address/address.go
@@ -32,21 +32,28 @@ func SyncClusterAddress(ctx context.Context,
 		seedDCName = *nodeDc.SeedDNSOverwrite
 	}
 
-	service := &corev1.Service{}
-	serviceKey := types.NamespacedName{Namespace: cluster.Status.NamespaceName, Name: resources.ApiserverExternalServiceName}
-	if err := client.Get(ctx, serviceKey, service); err != nil {
-		return nil, err
-	}
-
-	// Something went terribly wrong
-	if service.Spec.Type != corev1.ServiceTypeLoadBalancer && service.Spec.Type != corev1.ServiceTypeNodePort {
-		return nil, fmt.Errorf("expected serviceType to be NodePort or LoadBalancer, was %q", service.Spec.Type)
+	frontProxyLoadBalancerServiceIP := ""
+	if cluster.Spec.ExposeStrategy == corev1.ServiceTypeLoadBalancer {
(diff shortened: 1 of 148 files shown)


api/pkg/resources/address/address.go
```diff
@@ -32,21 +32,28 @@ func SyncClusterAddress(ctx context.Context,
...
+		if err := client.Get(ctx, nn, frontProxyLoadBalancerService); err != nil {
+			return nil, fmt.Errorf("failed to get the front-loadbalancer service: %v", err)
+		}
+		// Supposively there is only one if not..Good luck
+		for _, ingress := range frontProxyLoadBalancerService.Status.LoadBalancer.Ingress {
+			if ingress.IP != "" {
+				frontProxyLoadBalancerServiceIP = ingress.IP
+			}
+		}
+		if frontProxyLoadBalancerService.Spec.LoadBalancerIP != "" {
```
mrIncompetent: Why does the spec field has a higher priority over the status field ?
What are the situations where only one of them is set ?

api/pkg/resources/nodeportproxy/nodeportproxy.go
```diff
@@ -0,0 +1,328 @@
...
+				APIGroup: "rbac.authorization.k8s.io",
+				Kind:     "Role",
+				Name:     name,
+			}
+			return r, nil
+		}
+	}
+}
+
+func deploymentEnvoy(image string, nppd nodePortProxyData) reconciling.NamedDeploymentCreatorGetter {
```
mrIncompetent: `nppd` -> `data`

api/pkg/resources/address/address.go
alvaroaleman: Because I thought the spec field has to be filled out by the implementation if empty and the status is optional. Turns out its more or less the other way round:

```
Some cloud providers allow you to specify the loadBalancerIP. In those cases, the load-balancer will be created with the user-specified loadBalancerIP. If the loadBalancerIP field is not specified, the loadBalancer will be set up with an ephemeral IP address. If you specify a loadBalancerIP but your cloud provider does not support the feature, the loadbalancerIP field that you set will be ignored.
```
Ref: https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer

Since I once wrote an LB implentation that doesn't populate the status I've assumed others may get that wrong as well and changed the whole thing to default to the spec field first, but overwrite it with the status if its set

alvaroaleman:
@zreigz Can you take a look?

zreigz:
> @zreigz Can you take a look?

@alvaroaleman  I will take a look tomorrow morning

PR #3607, opened 2019-06-11
Use foreground deletion for machinedeployments

alvaroaleman:

Currently when a MachineDeployment is deleted, it is instantly gone even thought there are still machines in the system whose deletion is ongoing. This creates a weird UX, where there are "hidden" machines after a MachineDeployment was deleted.

With this change we will use foregroundDeletion, which means that that MachineDeployment stays in the system until all dependants (machinesets or machines) are deleted.

Fixes #

**Documentation**:

```release-note
none
```

/assign @maciaszczykm


maciaszczykm:
Makes a lot of sense, had the same feeling about the UX.

/lgtm
/approve

PR #3615, opened 2019-06-12
pass through proxy & image settings from the datacenters.yaml

mrIncompetent:
Pass through proxy & image settings from the datacenters.yaml.

```release-note
Kubermatic now supports running in environments where the Internet can only be accessed via a http proxy
```

### DIFF
--- a/api/pkg/provider/datacenter.go
+++ b/api/pkg/provider/datacenter.go
@@ -122,6 +122,20 @@ type DatacenterSpec struct {
 	GCP          *GCPSpec          `yaml:"gcp,omitempty"`
 }
 
+// NodeSettings are node specific which can be configured on datacenter level
+type NodeSettings struct {
+	// If set, this proxy will be configured on all nodes.
+	HTTPProxy string `yaml:"http_proxy,omitempty"`
+	// If set this will be set as NO_PROXY on the node
+	NoProxy string `yaml:"no_proxy,omitempty"`
+	// If set, this image registry will be used for pulling all required images on the node
+	InsecureRegistries []string `yaml:"insecure_registries,omitempty"`
+	// Translates to --pod-infra-container-image on the kubelet. If not set, the kubelet will default it
+	PauseImage string `yaml:"pause_image,omitempty"`
+	// The hyperkube image to use. Currently only Container Linux uses it
+	HyperkubeImage string `yaml:"hyperkube_image,omitempty"`
+}
+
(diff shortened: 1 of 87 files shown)


api/pkg/provider/datacenter.go
```diff
@@ -122,6 +122,20 @@ type DatacenterSpec struct {
 	GCP          *GCPSpec          `yaml:"gcp,omitempty"`
 }
 
+// NodeSettings are node specific which can be configured on datacenter level
+type NodeSettings struct {
+	// If set, this proxy will be configured on all nodes.
+	HTTPProxy string `yaml:"http_proxy,omitempty"`
+	// If set this will be set as NO_PROXY on the node
+	NoProxy string `yaml:"no_proxy,omitempty"`
+	// If set, this image registry will be used for pulling all required images on the node
```
alvaroaleman: That is wrong, it will just be configured as insecure registry

api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -190,3 +186,30 @@ func getEnvVars(data machinecontrollerData) []corev1.EnvVar {
...
+		"-kubeconfig", "/etc/kubernetes/kubeconfig/kubeconfig",
+		"-logtostderr",
+		"-v", "4",
+		"-cluster-dns", clusterDNSIP,
+		"-internal-listen-address", "0.0.0.0:8085",
+	}
+	if len(nodeSettings.InsecureRegistries) > 0 {
+		flags = append(flags, "-node-insecure-registries", strings.Join(nodeSettings.InsecureRegistries, ","))
+	}
+	if nodeSettings.HTTPProxy != "" {
```
alvaroaleman, line 201: Do we even need all these conditions? If those settings are empty, they will be set to an empty string which makes the machine-controller not do anything

api/pkg/provider/datacenter.go
mrIncompetent: :man_facepalming: I just copied from the machine-controller code. Its wrong there too...
I'll fix it here & fix it later in the mc

api/pkg/resources/machinecontroller/deployment.go
mrIncompetent: For the proxy flags currently: No.
But i would like to stick do the defaults whenever we can - And avoid overriding it explicitly

alvaroaleman:
@mrIncompetent Please also update the import in `gopkt.toml`, other than that LGTM

PR #3614, opened 2019-06-12
update minio to RELEASE.2019-06-11T00-44-33Z

xrstf:
**What this PR does / why we need it**:
What the title says. This includes two Major Feature Releases with these major changes:

* Self-healing feature has been implemented. See (minio/minio#7604) for more details.
* Add admin API to send trace notifications to registered listeners. Support for MINIO_HTTP_TRACE environment variable has been removed. See (minio/minio#7128) for more details.
* MinIO server runs in high performance mode by turning off content-md5 sum calculation by default. See minio/minio#7609 for more details. To run MinIO server in strict S3 compatibility mode, please start server with --compat flag.
* Major performance improvements in listing. See (minio/minio#7611) for more details.
* Major performance improvements in bulk object deletes. See (minio/minio#7607 and minio/minio#7644) for more details.

I don't think anyone needs strict S3 compat, which is why I kept the "high performance mode" the default but added a flag to the values.yaml to re-enable strict compat when needed.

**Does this PR introduce a user-facing change?**:
```release-note
updated Minio to RELEASE.2019-06-11T00-44-33Z
```

PR #3616, opened 2019-06-12
Update Vault to 1.1.3

xrstf:
This updates Vault to the latest stable version across all Docker images. The Helm version has also been bumped because we use 2.13 everywhere else already.

From my simple tests, using 1.1.3 locally works perfectly fine when using vault.loodse.com as the server.

This PR also downgrades mc because the current Minio release does not (yet?) have a companion mc release and our Minio deployments keep failing because they cannot init their backup container. To prevent future fuckups, I added explicit calls to print each app's version.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/seed-proxy/resources.go
+++ b/api/pkg/controller/seed-proxy/resources.go
@@ -175,7 +175,7 @@ func masterDeploymentCreator(contextName string, secret *corev1.Secret) reconcil
 			d.Spec.Template.Spec.Containers = []corev1.Container{
 				{
 					Name:            "proxy",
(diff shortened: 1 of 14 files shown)


xrstf:
/hold

Hold until #3614 is merged, because that other PR would bump the util image to 1.0.0-3 first.

kron4eg:
why do we even need to update vault? Our vault is running at 0.11.4 version. So why bother here?

xrstf:
> why do we even need to update vault

Same reasons as for other updates we do. Bugfixes, improvements, ... -- and if we ever want to update the Vault server, it's good to know we have recent enough clients floating around.

PR #3620, opened 2019-06-12
Make every image in our charts configurable via the values.yaml

mrIncompetent:
This change does 2 things:
- Ensure we always specify the registry (Required by our retagging for offline setups)
- Ensures every image repository can be overwritten via the values.yaml (Required for offline setups)

All changes should be backwards compatible

```release-note
NONE
```

### DIFF
--- a/config/nodeport-proxy/values.yaml
+++ b/config/nodeport-proxy/values.yaml
@@ -1,3 +1,8 @@
+# Required to make the templating pass without a valid values.yaml
+# Unfortunately we reuse the kubermatic ImagePullSecret here - Changing it would be a breaking change
+# and would duplicate the ImagePullSecret in the "master" values.yaml.
+kubermatic:
+  imagePullSecretData: ""
 nodePortProxy:
   replicas: 3
   image:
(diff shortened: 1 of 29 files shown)


config/nodeport-proxy/values.yaml
```diff
@@ -1,11 +1,16 @@
+# Required to make the templating pass without a valid values.yaml
+# Unfortunately we reuse the kubermatic ImagePullSecret here - Changing it would be a breaking change
+# and would duplicate the ImagePullSecret in the "master" values.yaml.
+kubermatic:
+  imagePullSecretData: ""
```
xrstf: Have you checked if this breaks the sync-charts script? It was a few months ago tweaked to look at top level keys and ignore duplicates, I am not sure how well this works if you add `kubermatic` in this values.yaml as well.

mrIncompetent: Fixed with https://github.com/kubermatic/kubermatic/pull/3620/commits/59cbc617a19fcac5e6121c3b47b8381910de328a

Issue #3611, opened 2019-06-12
Investigate customer changes to openvpn mtu and canal mtu


alvaroaleman:
**User Story**

A customer of ours has found some issues with our current MTU settings for openVPN and canal and subsequently applied changes to them. We should investigate that and bring those changes back to our mainline, if applicable.

**Acceptance criteria**

* openVPN changes were investigated and if necesary, fixes were made
* Canal changes were investigated and if necesary, fixes were made

PR #3621, opened 2019-06-12
Stop setting openVPN link mtu and set mss on the underlying connection instead

alvaroaleman:

Stops setting the openVPN link MTU and instead sets mss for connections to the openVPN server to make sure we don't run into issues due to some MTU on our route.

Related to #3611

I manually checked that logs and exec still work, for the former with enough logs to make sure it doesn't fit in one packet

**Documentation**:

```release-note
none
```

CC @bashofmann 
/assign @thz

### DIFF
--- a/api/pkg/resources/openvpn/deployment.go
+++ b/api/pkg/resources/openvpn/deployment.go
@@ -162,7 +162,6 @@ iptables -A INPUT -i tun0 -j DROP
 				"--duplicate-cn",
 				"--client-config-dir", "/etc/openvpn/clients",
 				"--status", "/run/openvpn-status",
-				"--link-mtu", "1432",
 				"--cipher", "AES-256-GCM",
 				"--auth", "SHA1",
 				"--keysize", "256",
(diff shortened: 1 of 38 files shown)


api/pkg/resources/openvpn/deployment.go
```diff
@@ -225,13 +224,18 @@ iptables -A INPUT -i tun0 -j DROP
...
-					Name:    "ip-forward",
+					Name:    "ip-fixup",
 					Image:   data.ImageRegistry(resources.RegistryQuay) + "/kubermatic/openvpn:v0.5",
 					Command: []string{"/bin/bash"},
 					Args: []string{
 						"-c",
 						// Always set IP forwarding as a CNI plugin might reset this to 0 (Like Calico 3).
-						"while true; do sysctl -w net.ipv4.ip_forward=1; sleep 30; done",
+						`while true; do sysctl -w net.ipv4.ip_forward=1;
+  if ! iptables -t mangle -C POSTROUTING -p tcp --tcp-flags SYN,RST SYN --dport 1194 -j TCPMSS --set-mss 1300 &>/dev/null; then
```
thz: `mangle POSTROUTING` is not entered for packets entering a host (or netns) and have the same host as final destination. In this case the tcp:1194 packets are the outer (encapsulated) frames which are destined for the local tcp:1194 listening socket.
Either `--set-mss` is available in `mangle INPUT` or we need to find a different place/way.

thz (changes requested):

thz:
Unfortunately the mss setting is in effect. This needs some digging into. I suspect something between pod-exist and node-exit to reset the mss to the higher value.

/hold

thz:
I have to correct myself: mss changes are effective.

/unhold

thz (approved):

PR #3622, opened 2019-06-12
Correctly set the calico mtu to match the flannel interface

### LINKED ISSUE #3611: Investigate customer changes to openvpn mtu and canal mtu

alvaroaleman:
Also change the upgrade strategy to make sure this change gets rolled
out.

Ref: https://docs.projectcalico.org/v2.6/usage/configuration/mtu

Fixes https://github.com/kubermatic/kubermatic/issues/3611

**Documentation**:

```release-note
none
```

/assign @mrIncompetent 
CC @bashofmann

### DIFF
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -139,10 +139,15 @@ spec:
           - -c
           - |
             set -e
-            DEFAULT_INTERFACE_NAME=$(ip r s|grep default|cut -d ' ' -f5)
+            DEFAULT_INTERFACE_NAME=$(ip r s|grep ^default|cut -d ' ' -f5)
             DEFAULT_INTERFACE_MTU=$(cat /sys/class/net/$DEFAULT_INTERFACE_NAME/mtu)
             OUR_MTU=$((DEFAULT_INTERFACE_MTU - 50))
-            sed -i "s/__CALICO_MTU__/$OUR_MTU/g" /host/etc/cni/net.d/10-calico.conflist
+            if [[ "$OUR_MTU" -gt 0 ]]; then
+              sed -i "s/__CALICO_MTU__/$OUR_MTU/g" /host/etc/cni/net.d/10-calico.conflist
+            else
+              echo "MTU found on interface $DEFAULT_INTERFACE_MTU ($DEFAULT_INTERFACE_MTU) -50 was < 0"
+              exit 1
+            fi
           volumeMounts:
             - mountPath: /host/etc/cni/net.d
               name: cni-net-dir
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -130,7 +130,7 @@ spec:
             - mountPath: /host/etc/cni/net.d
               name: cni-net-dir
         - name: set-mtu
-          image: '{{Registry "docker.io" }}/alpine'
+          image: '{{Registry "docker.io" }}/alpine:3.9'
           # We can not use the MTU of the flannel interface directly, because
           # it may not exist yet. Running this in parallel to the other containers
           # is a bad idea as well, as they may come up with a wrong configuration
@@ -139,13 +139,13 @@ spec:
           - -c
           - |
             set -e
-            DEFAULT_INTERFACE_NAME=$(ip r s|grep ^default|cut -d ' ' -f5)
+            DEFAULT_INTERFACE_NAME=$(cat /proc/net/route |awk '{ print $1 " " $2 }'|grep '00000000'|cut -d' ' -f1)
             DEFAULT_INTERFACE_MTU=$(cat /sys/class/net/$DEFAULT_INTERFACE_NAME/mtu)
             OUR_MTU=$((DEFAULT_INTERFACE_MTU - 50))
             if [[ "$OUR_MTU" -gt 0 ]]; then
               sed -i "s/__CALICO_MTU__/$OUR_MTU/g" /host/etc/cni/net.d/10-calico.conflist
             else
-              echo "MTU found on interface $DEFAULT_INTERFACE_MTU ($DEFAULT_INTERFACE_MTU) -50 was < 0"
+              echo "MTU found on interface $DEFAULT_INTERFACE_NAME ($DEFAULT_INTERFACE_MTU) -50 was <= 0"
               exit 1
             fi
           volumeMounts:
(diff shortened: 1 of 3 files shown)


addons/canal/canal.yaml
```diff
@@ -99,6 +102,50 @@ spec:
...
+          image: '{{Registry "docker.io" }}/alpine'
+          # We can not use the MTU of the flannel interface directly, because
+          # it may not exist yet. Running this in parallel to the other containers
+          # is a bad idea as well, as they may come up with a wrong configuration
+          command:
+          - /bin/sh
+          - -c
+          - |
+            set -e
+            DEFAULT_INTERFACE_NAME=$(ip r s|grep default|cut -d ' ' -f5)
```
alvaroaleman: @thz Can you think of a nicer way to do that?

mrIncompetent:
/approve

letting @thz lgtm

addons/canal/canal.yaml
thz: So you use the default route for finding your "default interface". But actually you might have a more specific route for reaching your neighboring nodes via another interface. For nodes not connected to the public internet (air gap with proxy) you might even find yourself in a situation without without a default route. Would we know the node ip network already? In that case we can use `ip route get a.b.c.d` to find the effective interface being used.

alvaroaleman: In that case nothing works anyways :grimacing: https://github.com/kubermatic/kubermatic/pull/3622/files#diff-9b6947485e7f2083f4e0ade428ad67ecR16

thz: Wait what? You are searching for the interface of the default route to enter it into a variable which if left blank would default the the interface of the default route? That sounds like a complex no-op.

thz: Ahh no. We are always using the default route's interface anyways. And this change is only about deducing its MTU...

alvaroaleman: Yes, exactly. I am just slightly afraid the `ip` folkgs might decide to change the output order or something the like and have the whole thing break. Alternative would be to read from `/proc/net/route`? I guess that could be considered an API that will not be broken?

thz: In that case let's keep that piping but perhaps grep for `^default ` instead of `default` to prevent problems in case of very strange interface names ;-)

addons/canal/canal.yaml
```diff
@@ -99,6 +102,50 @@ spec:
...
+            DEFAULT_INTERFACE_MTU=$(cat /sys/class/net/$DEFAULT_INTERFACE_NAME/mtu)
+            OUR_MTU=$((DEFAULT_INTERFACE_MTU - 50))
+            sed -i "s/__CALICO_MTU__/$OUR_MTU/g" /host/etc/cni/net.d/10-calico.conflist
```
thz: I would add a safeguard around the `sed -i` like this:
```
if [ "$OUR_MTU" -gt 0 ]; then
  sed -i ...
else
  echo "complain here and then die"
  exit 1
fi
```
This catches empty variable and also empty DEFAULT_INTERFACE_MTU variable in one go and allows giving a better log entry... Just in case.

addons/canal/canal.yaml
```diff
@@ -99,6 +102,55 @@ spec:
...
+          # is a bad idea as well, as they may come up with a wrong configuration
+          command:
+          - /bin/sh
+          - -c
+          - |
+            set -e
+            DEFAULT_INTERFACE_NAME=$(ip r s|grep ^default|cut -d ' ' -f5)
+            DEFAULT_INTERFACE_MTU=$(cat /sys/class/net/$DEFAULT_INTERFACE_NAME/mtu)
+            OUR_MTU=$((DEFAULT_INTERFACE_MTU - 50))
+            if [[ "$OUR_MTU" -gt 0 ]]; then
```
thz, line 145: If you go with `[[` instead of `[` then please also select `/bin/bash` to not let some `sh` symlink/dash/ash/whatever randomness decide if it works or not.
I'd prefer to just stay with the standard conformant `[` (which is a shell built-in with bash too, but would also work if there is built-in)

addons/canal/canal.yaml
```diff
@@ -99,6 +102,55 @@ spec:
...
+              sed -i "s/__CALICO_MTU__/$OUR_MTU/g" /host/etc/cni/net.d/10-calico.conflist
+            else
+              echo "MTU found on interface $DEFAULT_INTERFACE_MTU ($DEFAULT_INTERFACE_MTU) -50 was < 0"
```
thz: you meant to escape one of the `$`, me thinks.

thz: ahh no. you wanted to use `$DEFAULT_INTERFACE_NAME` for the first variable

addons/canal/canal.yaml
alvaroaleman: I've pinned the image so we can be sure it will continue to work as long as it passes the tests

thz:
/lgtm
(g - does not stand for great or gorgeous, but good-enough ;-) )

addons/canal/canal.yaml
```diff
@@ -139,13 +139,13 @@ spec:
           - -c
           - |
             set -e
-            DEFAULT_INTERFACE_NAME=$(ip r s|grep ^default|cut -d ' ' -f5)
+            DEFAULT_INTERFACE_NAME=$(cat /proc/net/route |awk '{ print $1 " " $2 }'|grep '00000000'|cut -d' ' -f1)
```
thz, line 142: please go with this instead:
```
DEFAULT_INTERFACE_NAME=$(awk '($2 == 00000000) { print $1 }' /proc/net/route)
```

thz: Bahh. This got merged already because the LGTM was not removed after the code change it seems.

mrIncompetent: There have been no code changes after you commented with /lgtm :thinking:

thz: Yea - missed a reload and thus missed that change by 7 minutes ;-0

PR #3647, opened 2019-06-14
containers/util: Add iproute2, iptables, socat.

thz:
Version bump to: kubermatic/util-1.1.0

```release-note
NONE
```

### DIFF
--- a/containers/util/Dockerfile
+++ b/containers/util/Dockerfile
@@ -9,8 +9,11 @@ RUN apk add --no-cache -U \
     bash \
     ca-certificates \
     curl \
+    iproute2 \
+    iptables \
     jq \
     rsync \
+    socat \
     unzip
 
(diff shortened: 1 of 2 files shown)


xrstf:
What do we need these tools for?

thz:
> What do we need these tools for?

Deducing routes, setting iptables rules (NAT).

xrstf:
> Deducing routes, setting iptables rules (NAT).

And we need this for Kubermatic because .... ? ;-) That iptables is used for managing iptables is something the man page could have told me :-p I don't want to annoy just, but instead make sure that in half a year we still know where we needed iptables.

thz:
We need this for offline scenarios where the outbound connection attempts from Pods on workers to the ClusterIp of the Kubernetes Service (10.10.10.1) need to be translated (DNAT) towards an AWS VPC Endpoint to bridge the gap between workers in one VPC and the control-plane in another VPC. This is going to happen in a small controller running as a DaemonSet. 
We could alternatively create a dedicated util-net image, but we should try to keep maintenance needs low. The added tools are neither complex not huge, and pretty standard.

PR #3649, opened 2019-06-16
split deployment process into stacks: kubermatic, monitoring, logging

xrstf:
To make it easier to distinguish between core product and misc software stacks when deployments fail and we're getting notified, this PR splits the deployment into three distinct "stacks": kubermatic, monitoring and logging. By setting the `DEPLOY_STACK` variable (default: kubermatic) users can choose to install different charts.

Kubermatic is only built and pushed when the kubermatic stack is deployed.

PR that duplicates the Prow job follows.

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-deploy-cloud-asia.sh
+++ b/api/hack/ci/ci-deploy-cloud-asia.sh
@@ -1,23 +1,16 @@
 #!/usr/bin/env bash
 
+DEPLOY_STACK=${DEPLOY_STACK:-kubermatic}
+
 set -euo pipefail
 export GIT_HEAD_HASH="$(git rev-parse HEAD|tr -d '\n')"
 cd $(dirname $0)/../../..
 
 source ./api/hack/lib.sh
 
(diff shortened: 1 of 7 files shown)


api/hack/ci/ci-deploy-cloud-asia.sh
```diff
@@ -1,23 +1,15 @@
...
-
-echodate "Building binaries"
-time make -C api build
-echodate "Successfully finished building binaries"
-
-echodate "Building and pushing quay images"
-retry 5 ./api/hack/push_image.sh $GIT_HEAD_HASH $(git tag -l --points-at HEAD) "latest"
-echodate "Sucessfully finished building and pushing quay images"
+if [[ "${DEPLOY_STACK}" == "kubermatic" ]]; then
+  ./ci-push-images.sh
```
alvaroaleman: This needs to be a path releative to the root of the repository, as thats where we go in line 6

PR #3653, opened 2019-06-17
Add a migration to add ICMP allow rules to existing security groups

### LINKED ISSUE #3562: Allow ICMP in all security group rules we create

alvaroaleman:

Adds a migration for AWS, Azure and Openstack that adds ICMP allow rules if not present.

Fixes #3562 

**Special notes for your reviewer**

**Documentation**:

```release-note
none
```

/assign @kdomanski @mrIncompetent

### DIFF
--- a/api/pkg/crd/kubermatic/v1/cluster.go
+++ b/api/pkg/crd/kubermatic/v1/cluster.go
@@ -75,10 +75,6 @@ type ClusterList struct {
 
 // ClusterSpec specifies the data for a new cluster.
 type ClusterSpec struct {
-	// MigrationRevision describes the latest version of the migration that has been done
-	// It is used to avoid redundant and potentially costly migrations
-	MigrationRevision int `json:"migrationRevision"`
-
 	Cloud           CloudSpec                 `json:"cloud"`
 	ClusterNetwork  ClusterNetworkingConfig   `json:"clusterNetwork"`
 	MachineNetworks []MachineNetworkingConfig `json:"machineNetworks,omitempty"`
@@ -218,6 +214,10 @@ type ClusterStatus struct {
 	// Conditions contains conditions the cluster is in, its primary use case is status signaling between controllers or between
 	// controllers and the API
 	Conditions []ClusterCondition `json:"conditions,omitempty"`
+
+	// MigrationRevision describes the latest version of the migration that has been done
+	// It is used to avoid redundant and potentially costly migrations
+	MigrationRevision int `json:"migrationRevision"`
 }
 
 type ClusterStatusError string
--- a/api/pkg/crd/kubermatic/v1/cluster.go
+++ b/api/pkg/crd/kubermatic/v1/cluster.go
@@ -75,6 +75,10 @@ type ClusterList struct {
 
 // ClusterSpec specifies the data for a new cluster.
 type ClusterSpec struct {
+	// MigrationRevision describes the latest version of the migration that has been done
+	// It is used to avoid redundant and potentially costly migrations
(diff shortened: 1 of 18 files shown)


api/pkg/crd/migrations/migrations.go
```diff
@@ -543,3 +557,41 @@ func setExposeStrategyIfEmpty(cluster *kubermaticv1.Cluster, ctx *cleanupContext
 	}
 	return nil
 }
+
+func migrateCloudProvider(cluster *kubermaticv1.Cluster, ctx *cleanupContext) error {
+	if cluster.Spec.MigrationRevision >= currentMigrationRevision {
+		glog.Infof("Not migrating cluster %q, because its .spec.migrationRevision is > 0(%d)",
```
kdomanski: `>=`

api/pkg/provider/cloud/aws/provider.go
```diff
@@ -347,8 +426,8 @@ func createSecurityGroup(client *ec2.EC2, vpcID, clusterName string) (string, er
 				SetIpProtocol("icmpv6").
 				SetFromPort(-1). // any port
 				SetToPort(-1).   // any port
-				SetIpRanges([]*ec2.IpRange{
-					{CidrIp: aws.String("0.0.0.0/0")},
+				SetIpv6Ranges([]*ec2.Ipv6Range{
```
kdomanski, line 429: Nice catch.

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -75,6 +75,10 @@ type ClusterList struct {
 
 // ClusterSpec specifies the data for a new cluster.
 type ClusterSpec struct {
+	// MigrationRevision describes the latest version of the migration that has been done
+	// It is used to avoid redundant and potentially costly migrations
+	MigrationRevision int `json:"migrationRevision"`
```
mrIncompetent: Why is it part of the spec and not status ?

api/pkg/crd/migrations/migrations.go
kdomanski: When we add some migrations and `currentMigrationRevision` gets incremented, these will be run again. Don't you think it would be cleaner to wrap the ICMP migrations only in a check for  `luster.Spec.MigrationRevision >= 1` specifically?

alvaroaleman: The comment is actually right

alvaroaleman:
@mrIncompetent @kdomanski PTAL

api/pkg/provider/cloud/azure/provider.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/golang/glog"
```
mrIncompetent: All providers have been migrated to zap already

alvaroaleman: Nope, only Azure. Updated the PR, PTAL

mrIncompetent:
/lgtm

/hold 
@kdomanski PTAL

Issue #3588, opened 2019-06-07
Verify metrics scraping on Kubernetes v1.15


kdomanski:
**User Story**
We need to verify that on new Kubernetes v1.15 the metrics still work as expected. Apparently we've had issues before when moving on to a new release.

We must make sure that on v1.15:
- scraping succeeds
- all metrics we use for alerting are there as expected

xrstf:
Is this regarding user clusters, seeds or both?

kdomanski:
Both.

kdomanski:
Note that v1.15 is still in beta and it would probably make sense to wait for the .0 release.

xrstf:
After deploying Prometheus, node-exporter and kube-state-metrics into a user cluster running 1.15rc1, all alerts and dashboards kept functioning. I don't expect metrics related changes after rc1.

PR #3663, opened 2019-06-18
lower default Prometheus resource requirements

xrstf:
For smaller test setups and POCs it's nice if Prometheus by default does not tons of resources and might even be unschedulable. We override the resources for all our systems anyway and I expect most customers do as well (though we can note this in the release notes to prevent accidental downscaling).

The configurable storage class also helps for quick setups where no custom storage class is setup or required.

Relates to #3588

```release-note
NONE
```


kron4eg (approved):
/approve no-issue

PR #3664, opened 2019-06-18
use the canonical name for docker images inside the addon manifests

mrIncompetent:
This will replace the short names for docker images, like `alpine` with the canonical name: `docker.io/library/alpine`.
This ensures that we can reliably detect the registry, org & name.

```release-note
NONE
```


mrIncompetent:
@alvaroaleman Image tag has been updated

PR #3668, opened 2019-06-19
allow restic on all nodes

xrstf:
Since we moved Prometheus on the stable node pool, backups are failing because there are no local restic pods running on the stable pool by default. This PR adjusts the default tolerations to have Restic available everywhere.

```release-note
NONE
```


kron4eg (approved):
/approve no-issue

PR #3670, opened 2019-06-19
more realistic resource requests for Elasticsearch

xrstf:
Judging from our dev cluster, the default resource usage for Elasticsearch is higher than what we define. This leads to bad scheduling decisions.

Starting with Java 10, more container-centric options are available. The newest is `-XX:+UseContainerSupport`, which supersedes `-XX:+UseCGroupMemory*`. With this the JVM is smart enough to figure out the resources the container is assigned. The only thing we then set is the percentage of memory to use for the heap, which we set to 80% (determined using a 12-sided dice, thrown 17 times and then averaged).

```release-note
NONE
```

### DIFF
--- a/config/logging/elasticsearch/templates/es-data-stateful.yaml
+++ b/config/logging/elasticsearch/templates/es-data-stateful.yaml
@@ -78,14 +78,14 @@ spec:
         - name: NODE_DATA
           value: "true"
         - name: ES_JAVA_OPTS
-          value: "-Djava.net.preferIPv4Stack=true -Xms{{ .Values.logging.elasticsearch.data.heapSize }} -Xmx{{ .Values.logging.elasticsearch.data.heapSize }} {{ .Values.logging.elasticsearch.cluster.additionalJavaOpts }} {{ .Values.logging.elasticsearch.data.additionalJavaOpts }}"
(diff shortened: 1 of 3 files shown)


kron4eg:
Wha size of heap (-Xmx, -Xms and -xx:NewSize) are those images using? Is it possible to change them?

xrstf:
We are setting `-Xms=HEAP_SIZE -Xmx=HEAP_SIZE` and HEAP_SIZE is configured using our values.yaml (256MB for master nodes, 1024MB for data nodes). Any other JVM option would have to be set explicitly and we don't do that.

kron4eg:
@xrstf so... are you going to align `logging.elasticsearch.master.heapSize` and `logging.elasticsearch.data.heapSize` with those new `requests.memory`?

xrstf:
Our ES stack runs on OpenJDK 12, so container support is enabled by default. It seems that setting the heap size explicitly is not the optimal solution, we could rather just set a percentage of memory to use and let Java do its thing. It just requires that we have proper container limits set.

kron4eg:
oh wow, that's awesome

kron4eg (approved):
/approve no-issue

NOW I like it 👍 💯

PR #3671, opened 2019-06-19
lower dex resource requests

xrstf:
Dex does not need 128MB memory, let alone a limit of 200MB.

```release-note
NONE
```


kron4eg (approved):
/approve no-issue

PR #3681, opened 2019-06-20
override jvm.options for Elasticsearch

xrstf:
After the last PR (#3670) where we moved away from setting explicit memory constraints, we noticed that they were still present due to the jvm.options file. This PR adds a customized version to our ConfigMap and mounts that one instead of the default one.

Options that are changed:

* Do not set `Xms` and `Xmx`.
* Cache DNS results for 10 instead of 60 seconds to accomodate changing pod IPs quicker (`es.networkaddress.cache.ttl`).
* Re-enable `OmitStackTraceInFastThrow` because we do not care about stack traces.
* Disable `HeapDumpOnOutOfMemoryError` because we don't need the heap dump when ES crashes.

I ran the Rally benchmarks (`esrally --pipeline=benchmark-only --target-hosts es-data.es-pr.svc.cluster.local:9200`) on the old master branch, the current master branch and a couple of variants of this PR. All variants completed the benchmarks and all basically performed identically. IMHO it's safe to transition to MaxRAMPercentage.

This PR, while we're at it, also makes it possible to run a single-node ES "cluster" by setting the sum of data and master replicas to 1. This interferes with scaling the StatefulSet, but it's clearly documented and should be fine.

This PR also switches to the new GC because ES kept warning about the old one. There were no big gains to be expected and according to the benchmark the speedup in indexing new documents was barely noticeable. But as changing the GC did not break anything, I'd wager it's safe to switch the GC.

```release-note
NONE
```

### DIFF
--- a/config/logging/elasticsearch/Chart.yaml
+++ b/config/logging/elasticsearch/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: elasticsearch
-version: 1.0.4
+version: 1.0.5
 appVersion: 6.7.1
 description: Chart to deploy Elasticsearch master and data nodes.
 keywords:
--- a/config/logging/elasticsearch/templates/es-configmap.yaml
+++ b/config/logging/elasticsearch/templates/es-configmap.yaml
@@ -6,11 +6,6 @@ data:
   elasticsearch.yml: |-
     cluster.name: log-data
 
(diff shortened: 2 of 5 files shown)


kron4eg (approved):
/approve

kron4eg (approved):
/approve no-issue

Issue #3024, opened 2019-03-18
AWS: allow providing a CIDR with which to create a new VPC for a cluster


kdomanski:
**User Story**
On providers like Azure or OpenStack, we create a new virtual network for each cluster (if no specific network is provided in the config), ensuring that each cluster is isolated from other clusters' traffic. In our AWS provider, however, we use the default VPC instead, causing multiple clusters to share a network. This often cannot be the case.

**Acceptance criteria**
- [ ] If no VPC is provided, a new VPC will be created for the cluster on its creation
- [ ] If Kubermatic created a VPC for a cluster, then it will delete the VPC on the cluster's deletion

mrIncompetent:
@mrIncompetent Check for the priority

mrIncompetent:
@kdomanski This issue needs clarification around the subnet CIDR.
How does this affect multi AZ support?

kdomanski:
From my chat w/ Henrik on March 26th:
```
my proposal for the AWS VPC creation workflow:
- a new field in AWS provider config, `vpcCIDR`
- controller wil create a new VPC with that CIDR
- the CIDR will be split into 4 equal chunks
- chunks will be equally distributed among subnets, 1 for each availability zone
```

kdomanski:
Note: we're currently adding a default storage class in the addons, bound to the first AZ.

It seems like we have this one subnet/AZ binding, because at some point we created a VPC ourselves, and we wanted simplicity. But since that time, we removed the VPC creation
https://github.com/kubermatic/kubermatic/pull/282

kdomanski:
VPC creation was initially added in https://github.com/kubermatic/kubermatic/commit/8eb32a51ce8acaf0042694d5cf96c36829b80e11

mrIncompetent:
To sum it up, we create
- VPC
- Subnets (1 per zone)
- InternetGateway
- RouteTable
- Routes

And in the end we use the subnet, which fits to the zone defined in the datacenters.yaml?

mrIncompetent:
Does creating a VPC mean its public or private?
So do we create a InternetGateway or a NAT Gateway?

kdomanski:
IMO InternetGateway, because NAT gateways do not support IPv6 egress which might be an issue in the future.
@thz WDYT?

thz:
For v6 aws advises to use an egress-only-internet-gateway which is not the same thing as an igw.
[...]

Issue #2336, opened 2018-11-05
use secretKeyRef in user cluster k8s object


toschneck:
**User Story**
The current cluster object contains the provider secrets as plain values. For doing demo use cases (e.g. screen sharing) it's not great when we show the secrets as plain text, e.g:
```
kubectl get cluster dmmsjw8nnz -o yaml
address:
  adminToken: sg6nzz.q74dq2dr7vtnrkx2
  externalName: dmmsjw8nnz.europe-west3-c-1.cloud.kubermatic.io
  ip: 35.234.122.0
  url: https://dmmsjw8nnz.europe-west3-c-1.cloud.kubermatic.io:30952
apiVersion: kubermatic.k8s.io/v1
kind: Cluster
metadata:
  creationTimestamp: 2018-11-05T16:22:23Z
  finalizers:
  - kubermatic.io/cleanup-backups
  - kubermatic.io/delete-nodes
  generation: 1
  labels:
    project-id: p7fmvrx899
  name: dmmsjw8nnz
  resourceVersion: "213454149"
  selfLink: /apis/kubermatic.k8s.io/v1/clusters/dmmsjw8nnz
  uid: f9eacd97-e116-11e8-9fac-42010a9c00f1
spec:
  cloud:
    dc: do-fra1
    digitalocean:
      token: XXXXXXXXXXXXXXXXXXXXXXX
  clusterNetwork:
    dnsDomain: cluster.local
    pods:
      cidrBlocks:
      - 172.25.0.0/16
    services:
      cidrBlocks:
      - 10.10.10.0/24
  componentsOverride:
    apiserver: {}
    controllerManager: {}
    etcd: {}
    scheduler: {}
  humanReadableName: do-testa
  masterVersion: ""
  pause: falseUI 
  version: 1.12.1
```
**Solution:** use for all cloud provider `secretKeyRef` like here: https://github.com/kubermatic/machine-controller/blob/dbbc29796d291d9329780efeccea6659f11227d3/pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/hetzner.yaml

**Acceptance criteria**
- [ ] if a new cluster get created through the UI the command `kubectl get cluster <cluster-id> -o yaml` doesn't contain any secret.
- [ ] it is implemented at all cloud providers

mrIncompetent:
@toschneck has this been discussed with anyone from the development so far?

toschneck:
[...]

PR #3683, opened 2019-06-20
add a method to validate spec changes by cloud providers

kdomanski:
It allows a cloud provider to ensure that no breaking changes to the spec are applied by the API consumers, e.g. you can still update credentials but you cannot mess with network configuration of an existing cluster.
I need it to forbid changing of the CIDR mentioned in https://github.com/kubermatic/kubermatic/issues/3024 but I'm opening as a separate PR for clarity.

```release-note
NONE
```

/assign @mrIncompetent

### DIFF
--- a/api/pkg/provider/cloud/aws/provider.go
+++ b/api/pkg/provider/cloud/aws/provider.go
@@ -770,3 +770,8 @@ func isEntityAlreadyExists(err error) bool {
 	}
 	return aerr.Code() == "EntityAlreadyExists"
 }
+
+// ValidateCloudSpecUpdate verifies whether an update of cloud spec is valid and permitted
+func (a *AmazonEC2) ValidateCloudSpecUpdate(oldSpec kubermaticv1.CloudSpec, newSpec kubermaticv1.CloudSpec) error {
+	return nil
+}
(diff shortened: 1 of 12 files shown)


alvaroaleman:
An alternative approach to this would be to wait for https://github.com/kubermatic/kubermatic/issues/2336 which @nikhita  is currently working on.

That way we would just update the referenced secret and can continue to view the cluster spec as immutable

kdomanski:
The point is not being able to update it. Besides, the example I have (CIDR) is not really a secret.

alvaroaleman:
> The point is not being able to update it. Besides, the example I have (CIDR) is not really a secret.

The point is being able to update credentials but not being able to update anything else or did I understand that wrong?

kdomanski:
The point is being able to update some fields, but not others.

mrIncompetent:
I see the need for such an validation and i'm fine with merging it.

But: We should think of moving such validation into a webhook.
Despite its deployment complexity, it gives us a better separation of concerns (The CRD basically belongs to the cluster-lifecycle team) and does not leak possibilities to circumvent the validation (Like performing a `kubectl edit cluster xxxxxx` like now).

/lgtm
/approve no-issue

PR #3684, opened 2019-06-21
Allow specifying tokens using secrets

nikhita:
For https://github.com/kubermatic/kubermatic/issues/2336

Work in progress, but would love some initial feedback.

/cc @alvaroaleman 

```release-note
NONE
```


api/pkg/resources/data.go
```diff
@@ -273,3 +274,20 @@ func (d *TemplateData) KubermaticAPIImage() string {
 func (d *TemplateData) APIServerExposeStrategy() corev1.ServiceType {
 	return d.apiserverExposeStrategy
 }
+
+func (d *TemplateData) GetConfigVarStringValue(configVar providerconfig.GlobalSecretKeySelector) (string, error) {
+	// We need all three of these to fetch and use a secret
```
alvaroaleman, line 274: As this will most likely be needed in many places, it may make sense to generate the func via a factory and use that everywhere:

```
func ConfigVarStringGetterFactoy(ctx context.Context, client client.Client) func (configVar providerconfig.GlobalSecretKeySelector) (string, error)
```

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -355,9 +356,9 @@ type OpenstackCloudSpec struct {
 
 // PacketCloudSpec specifies access data to a Packet cloud.
 type PacketCloudSpec struct {
-	APIKey       string `json:"apiKey"`
```
alvaroaleman, line 345: I do not really feel comfortable changing the existing field. It will make migration code weird as we'd need to build a custom json unmarshaller that can cope with both the field being a string and it being a `GlobalSecretKeySelector`.

I'd prefer it if we instead added new field `OriginalFieldNameReference`, e.G. `APIKeyReference` here and add a comment to the old field that they are deprecated

nikhita:
One question I had around this was... should secrets be created using the credentials defined in `pkg/presets/manager.go` and then `secretRef` for these secrets be "copied" onto the cluster object?

I'm not entirely sure if this somehow conflicts with the credentials work from the API team? In other words: are the "presets" intended to be copied directly to the cluster object? cc @zreigz

zreigz:
@nikhita 
Instead of passing credentials or other parameters needed to create a cluster in UI, we prepared presets for this. User can specify presets with credentials and extended parameters and just use the preset name. In this part, we search for the values and inject them to cluster spec: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/cluster/cluster.go#L75
The presets manager converts the preset name to values which must be used to create the cluster object.

According to your question if secrets should be generated from the presets.
Presets are designed to add possibilities to use predefined credentials instead of passing all the time. When the presets are defined than our UI display credential name and use them by default. The custom credential feature was requested by big companies to be able to create clusters for individual employees. The admin will spin up the kubermatic with some provider credential which belongs to the company, not to the individual person.

When you start using presets for generating secrets then you enable this feature. It has some implications. Different users assigned to different projects can use the same credentials.

In my opinion, there are two scenarios for your issue:
First, you should create secrets from credentials passed from the user in UI and they stay until cluster exist.
Second, when presets are defined then you can use them to create secrets. In this case, we need some mechanism to check if the presets still exist or someone didn't change them to be able delete/update secrets

nikhita:
@zreigz thank you so much for that explanation! :) Updated with some new code for creating/deleting secrets as the cluster is created/deleted but I had some more questions.

> Second, when presets are defined 

- My understanding is that the current state of things (i.e. not the changes in this PR) is like: credentials will _always_ be supplied using presets and then these credential values are injected into the clusterSpec. However, is it possible that presets might not be defined? In this case, how are credentials being supplied - directly through the cluster spec?

- Should we be adding an option to selectively define when secrets should be used for credentials vs directly injecting the values into the cluster object? Without that, I'm confused how we can decide whether to directly copy the values or use a secretKeyRef...

> we need some mechanism to check if the presets still exist

- Would this mean adding something to check if the preset (endpoint) was updated? Or...?

> someone didn't change them to be able delete/update secrets

- If someone deleted them, would we just delete the secret and give an error (because of lack of credentials) or do we also take up follow-up actions (delete anything else?) :thinking:

api/pkg/api/v1/types.go
```diff
@@ -900,4 +900,6 @@ const (
 	InClusterPVCleanupFinalizer = "kubermatic.io/cleanup-in-cluster-pv"
 	// InClusterLBCleanupFinalizer indicates that the LBs still need cleanup
 	InClusterLBCleanupFinalizer = "kubermatic.io/cleanup-in-cluster-lb"
+	// CredentialsSecretsCleanupFinalizer indicates that secrets for credentials still need cleanup
+	CredentialsSecretsCleanupFinalizer = "kubermatic.io/cleanup-credentials-secrets"
```
alvaroaleman, line 911: Its easier to instead just put an OwnerRef pointing to the cluster CR on the secret and letting the garbage collector take care

api/pkg/controller/openshift/data.go
```diff
@@ -268,3 +270,20 @@ func (od *openshiftData) KubermaticAPIImage() string {
 	}
 	return od.ImageRegistry(registry) + "/" + imageWithoutRegistry
 }
+
+func (od *openshiftData) GetConfigVarStringValue(configVar providerconfig.GlobalSecretKeySelector) (string, error) {
```
alvaroaleman: Is this still needed?

api/pkg/handler/v1/cluster/credential_secrets.go
```diff
@@ -0,0 +1,81 @@
...
+
+func createPacketSecret(cluster *v1.Cluster, clusterProvider provider.ClusterProvider, userInfo *provider.UserInfo) error {
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	defer cancelCtx()
+
+	// create secret for storing credentials
+	packetSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("packet-%s", cluster.Name),
+			Namespace: "kubermatic-secrets",
```
alvaroaleman: Lets make this a const

api/pkg/provider/cloud/packet/provider.go
```diff
@@ -30,11 +30,19 @@ func (p *packet) DefaultCloudSpec(spec *kubermaticv1.CloudSpec) error {
 // ValidateCloudSpec validates the given CloudSpec.
 func (p *packet) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 	if spec.Packet.APIKey == "" {
-		return fmt.Errorf("apiKey cannot be empty")
+		apiKey := spec.Packet.APIKeyReference
+		if apiKey.Name == "" && apiKey.Namespace == "" && apiKey.Key == "" {
```
alvaroaleman: This should be ` ||` instead of `&&` as we need all three of them or not?

alvaroaleman: And I guess we will have a lot of these checks, so better DRY it out and put it into a func :)

api/pkg/resources/data.go
```diff
@@ -273,3 +274,40 @@ func (d *TemplateData) KubermaticAPIImage() string {
 func (d *TemplateData) APIServerExposeStrategy() corev1.ServiceType {
 	return d.apiserverExposeStrategy
 }
+
+func (d *TemplateData) GetConfigVarStringValue(configVar providerconfig.GlobalSecretKeySelector) (string, error) {
```
alvaroaleman: Can we rename this to something like "GetGlobalSecretKeySelectorValue`? :grimacing:

api/pkg/resources/data.go
```diff
@@ -273,3 +274,40 @@ func (d *TemplateData) KubermaticAPIImage() string {
...
+	// We need all three of these to fetch and use a secret
+	if configVar.Name != "" && configVar.Namespace != "" && configVar.Key != "" {
+		secret := &corev1.Secret{}
+		key := types.NamespacedName{Namespace: configVar.Namespace, Name: configVar.Name}
+		if err := d.client.Get(d.ctx, key, secret); err != nil {
+			return "", fmt.Errorf("error retrieving secret '%s' from namespace '%s': '%v'", configVar.Name, configVar.Namespace, err)
```
alvaroaleman: Nit: Using `%q` achieves almost the same as `'%s` but takes less space in the code

api/pkg/resources/data.go
```diff
@@ -273,3 +274,40 @@ func (d *TemplateData) KubermaticAPIImage() string {
...
+
+		if val, ok := secret.Data[configVar.Key]; ok {
+			return string(val), nil
+		}
+		return "", fmt.Errorf("secret '%s' in namespace '%s' has no key '%s'", configVar.Name, configVar.Namespace, configVar.Key)
+	}
+	return "", nil
+}
+
+// TODO: use this instead, and mostly move this to another file?
```
alvaroaleman: We'll need this at least in the openshift and the monitoring controller as well, so yeah, the factory makes sense :)

api/pkg/controller/cluster/resources.go
```diff
@@ -185,15 +185,15 @@ func (r *Reconciler) ensureServices(ctx context.Context, c *kubermaticv1.Cluster
 }
 
 // GetDeploymentCreators returns all DeploymentCreators that are currently in use
-func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuthentication bool) []reconciling.NamedDeploymentCreatorGetter {
+func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuthentication bool, getGlobalSecretKeySelectorValue resources.GetGlobalSecretKeySelectorValue) []reconciling.NamedDeploymentCreatorGetter {
```
alvaroaleman: Whats the reason for not making the `getGlobalSecretKeySelectorValue` part of `data`?

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -357,7 +358,11 @@ type OpenstackCloudSpec struct {
...
+	APIKeyReference    *providerconfig.GlobalSecretKeySelector `json:"apiKeyReference,omitempty"`
+	ProjectIDReference *providerconfig.GlobalSecretKeySelector `json:"projectIDReference,omitempty"`
+	// APIKey is deprecated. Please use APIKeyReference instead.
+	APIKey string `json:"apiKey"`
```
alvaroaleman: Can we haz `omitempty`? :)

api/pkg/handler/v1/cluster/credential_secrets.go
```diff
@@ -0,0 +1,83 @@
...
+func createPacketSecret(cluster *v1.Cluster, clusterProvider provider.ClusterProvider, userInfo *provider.UserInfo) error {
+	ctx, cancelCtx := context.WithCancel(context.Background())
+	defer cancelCtx()
+
+	// create secret for storing credentials
+	blockOwnerDeletion := true
+	packetSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("packet-%s", cluster.Name),
+			Namespace: KubermaticSecretsNamespace,
```
alvaroaleman: Can you add the creation of the NS to the Kubermatic Helm chart?

api/pkg/handler/v1/cluster/credential_secrets.go
```diff
@@ -0,0 +1,83 @@
...
+				},
+			},
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"apiKey":    []byte(cluster.Spec.Cloud.Packet.APIKey),
+			"projectID": []byte(cluster.Spec.Cloud.Packet.ProjectID),
+		},
+	}
+	client, err := clusterProvider.GetClientForCustomerCluster(userInfo, cluster)
```
alvaroaleman: This gives you a client for the usercluster but we want to store the secret in the seed cluster :)

api/pkg/validation/cluster.go
```diff
@@ -335,3 +339,10 @@ func validateFakeCloudSpec(spec *kubermaticv1.FakeCloudSpec) error {
 
 	return nil
 }
+
+func ValidateSecretKeySelector(selector *providerconfig.GlobalSecretKeySelector, key string) error {
+	if selector.Name == "" && selector.Namespace == "" && selector.Key == "" {
```
alvaroaleman: Can we also check if we can actually get the secret? :grimacing:

api/pkg/controller/cluster/resources.go
alvaroaleman: ?

nikhita: Doing that now.

nikhita: > Whats the reason for not making the getGlobalSecretKeySelectorValue part of data?

So I've tried a few things but now I'm confused what you meant :sweat_smile:

Wouldn't making it part of data essentially revert back to the old state i.e. without the `GlobalSecretKeySelectorValueGetterFactory`?

alvaroaleman: By that I meant extend the `templateData` struct with a `GetGlobalSecretKeySelectorValue` func

api/pkg/controller/openshift/resources/machinecontroller.go
```diff
@@ -9,8 +9,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func MachineController(osData openshiftData) reconciling.NamedDeploymentCreatorGetter {
-	name, creator := machinecontroller.DeploymentCreator(osData)()
+func MachineController(osData openshiftData, getGlobalSecretKeySelectorValue resources.GetGlobalSecretKeySelectorValue) reconciling.NamedDeploymentCreatorGetter {
```
alvaroaleman: I think the added argument is not needed anymore, as its part of `osData`

api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -48,7 +48,7 @@ type machinecontrollerData interface {
 }
 
 // DeploymentCreator returns the function to create and update the machine controller deployment
-func DeploymentCreator(data machinecontrollerData) reconciling.NamedDeploymentCreatorGetter {
+func DeploymentCreator(data machinecontrollerData, getGlobalSecretKeySelectorValue resources.GetGlobalSecretKeySelectorValue) reconciling.NamedDeploymentCreatorGetter {
```
alvaroaleman: Just extend the `machinecontrollerData` interface so the additional arg is not needed

alvaroaleman (commented):
Two small comments and can we also have a migration that creates the secret if no ref exists but the cluster has the string values on it? :grimacing: 

Thanks for all your work on this!

api/pkg/resources/machinecontroller/deployment.go
nikhita: > Just extend the machinecontrollerData interface so the additional arg is not needed

This would work here, but then we'll also need to add a new method which would do this for `TemplateData` so that it will then satisfy this interface....unless I'm missing something...

alvaroaleman: Yes. The purpose of the `templateData` struct is exactly that: Carry all Informationen needed by any resource around

nikhita:
/test pull-kubermatic-e2e-packet-coreos-1.14

:crossed_fingers: :pray:

nikhita:
/test pull-kubermatic-e2e-packet-coreos-1.14

:prayer_beads: :pray:

nikhita:
The last run almost worked :)

/test pull-kubermatic-e2e-packet-coreos-1.14

nikhita:
It worked. \o/

PTAL :)

api/pkg/clusterdeletion/deletion.go
```diff
@@ -203,6 +206,30 @@ func (d *Deletion) cleanupInClusterResources(ctx context.Context, cluster *kuber
...
 }
 
+func (d *Deletion) cleanUpCredentialsSecrets(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	// If no relevant finalizer exists, directly return
+	if !kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.CredentialsSecretsCleanupFinalizer) {
+		return nil
+	}
+
+	secretList := &corev1.SecretList{}
+	if err := d.seedClient.List(ctx, &controllerruntimeclient.ListOptions{Namespace: "kubermatic-secrets"}, secretList); err != nil {
```
alvaroaleman: Why do we delete _all_ secrets? Shouldn't we just delete the one that is named like our cluster?

Also we should tolerate a 404 as non error when we try to get our secret :)

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -106,6 +107,12 @@ func CreateEndpoint(sshKeyProvider provider.SSHKeyProvider, cloudProviders map[s
 				"kubermatic.io/openshift": "true",
 			}
 		}
+		// generate the name here so that it can be used in the secretName below
+		partialCluster.Name = rand.String(10)
```
alvaroaleman, line 113: Can we do a quick `GET` to ensure we get a 404 for that name? :grimacing:

api/cmd/conformance-tests/runner.go
```diff
@@ -262,6 +263,10 @@ func (r *testRunner) executeScenario(log *logrus.Entry, scenario testScenario) (
 		}
 		cluster.Finalizers = append(cluster.Finalizers, kubermaticapiv1.InClusterPVCleanupFinalizer, kubermaticapiv1.InClusterLBCleanupFinalizer)
 
+		if handlerv1.ShouldAddSecretsCleanupFinalizer(cluster) {
```
alvaroaleman: I think we can add this unconditionally and just tolerate a 404 on deletion

api/pkg/handler/v1/cluster/credential_secrets.go
```diff
@@ -0,0 +1,79 @@
...
+// It then adds secret key selectors to the cluster object referencing this secret.
+// Finally, it clears the credentials values from the cluster object.
+func CreateSecretsForCredentials(cluster *v1.Cluster, priviligedClusterProvider provider.PrivilegedClusterProvider) error {
+	if cluster.Spec.Cloud.Packet != nil {
+		return createPacketSecret(cluster, priviligedClusterProvider)
+	}
+	return nil
+}
+
+func createPacketSecret(cluster *v1.Cluster, priviligedClusterProvider provider.PrivilegedClusterProvider) error {
```
alvaroaleman: Just add the `context` as first arg and move the `TODO` into the caller if no context is available there?

api/pkg/controller/openshift/data.go
```diff
@@ -268,3 +270,20 @@ func (od *openshiftData) KubermaticAPIImage() string {
 	}
 	return od.ImageRegistry(registry) + "/" + imageWithoutRegistry
 }
+
+func (od *openshiftData) GetGlobalSecretKeySelectorValue(configVar providerconfig.GlobalSecretKeySelector) (string, error) {
```
alvaroaleman, line 273: Why not use the `GlobalSecretKeySelectorValueGetterFactory`?

api/cmd/projects-migrator/main.go
```diff
@@ -925,3 +932,42 @@ func doesUserOwnProjectAs(user kubermaticv1.User, ctx migrationContext, groupPre
...
+					},
+				},
+				Type: corev1.SecretTypeOpaque,
+				Data: map[string][]byte{
+					"apiKey":    []byte(packetSpec.APIKey),
+					"projectID": []byte(packetSpec.ProjectID),
+				},
+			}
+
+			if _, err := client.CoreV1().Secrets("kubermatic-secrets").Create(packetSecret); err != nil {
```
alvaroaleman: Use the const :grimacing:

alvaroaleman (commented):
almost there :))

api/pkg/controller/openshift/data.go
nikhita: Because would require the `GetGlobalSecretKeySelectorValue` to be added as a field to the `osData` struct, which doesn't gel well while using with DeploymentCreators because they expect a method instead of a field. :/

api/pkg/handler/v1/cluster/cluster.go
nikhita: Did that but tests were failing because of :disappointed: 

```
 [FAIL] - aws-coreos-1.13.5 : failed to create cluster: failed to create cluster via kubermatic api: "{\"Payload\":{\"error\":{\"details\":null,\"code\":500,\"message\":\"unable to generate unique cluster name even after 5 tries, error: clusters.kubermatic.k8s.io \\\"4hfq8jcw54\\\" is forbidden: User \\\"serviceaccount-2zqm6cfghz@ci.kubermatic.io\\\" cannot get resource \\\"clusters\\\" in API group \\\"kubermatic.k8s.io\\\" at the cluster scope\"}}}" 
```

api/pkg/controller/openshift/data.go
alvaroaleman: hm, I think its possible to do something like:

```
type openshiftData struct {
 globalSecretKeySelector func(args type)
}

func (os *openshiftData) GlobalSecretKeyselector(args type) {
return od.globalSecretKeySelector(args type)
}
```

But can be a follow-up

alvaroaleman:
Lets merge this if the packet presubmit passes
/approve

alvaroaleman:
@zreigz Can you have a look?

zreigz:
> @zreigz Can you have a look?

yes, I will review this today

api/pkg/resources/resources.go
```diff
@@ -238,6 +239,9 @@ const (
 	// MetricsServerPodDisruptionBudgetName is the name of the PDB for the metrics-server deployment
 	MetricsServerPodDisruptionBudgetName = "metrics-server"
 
+	// KubermaticSecretsNamespace is the namespace where secrets for credentials will be present
+	KubermaticSecretsNamespace = "kubermatic-secrets"
```
zreigz: Helm installer supports only one namespace: `kubermatic`. For SA we also create `Secrets` and we store them in `kubermatic` namespace otherwise operator must create the new namespace manually. It also has implications with e2e tests.

api/pkg/handler/v1/cluster/credential_secrets.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CreateSecretsForCredentials creates secrets containing cloud provider credentials.
+// It then adds secret key selectors to the cluster object referencing this secret.
+// Finally, it clears the credentials values from the cluster object.
+func CreateSecretsForCredentials(ctx context.Context, cluster *v1.Cluster, priviligedClusterProvider provider.PrivilegedClusterProvider) error {
```
zreigz: Instead of this part, please introduce CredentialSecretsProvider. The similar provider is here: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/provider/kubernetes/sa_token.go
We also create Secret for SA but using ImpersonatedClient. 
In addition, you need to extend RBAC controller for cluster secrets with your specific prefix: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/rbac/rbac_controller_aggregator.go#L99
I recommend using one prefix for all providers. I can handle the changes for RBAC because it's the hardest part and very difficult to test.
When you create this provider then you will be able to test cluster creation for the Packet.

zreigz (changes requested):

nikhita:
Updated. PTAL.

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,111 @@
...
+	kubernetesImpersonatedClient, err := createKubernetesImpersonationClientWrapperFromUserInfo(userInfo, p.kubernetesImpersonationClient)
+	if err != nil {
+		return nil, kerrors.NewInternalError(err)
+	}
+
+	createdSecret, err := kubernetesImpersonatedClient.CoreV1().Secrets(resources.KubermaticNamespace).Create(secret)
+	if err != nil {
+		return nil, err
+	}
+	createdSecret.Name = removeTokenPrefix(createdSecret.Name)
```
zreigz: you can remove this line

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,111 @@
...
+	if cluster.Spec.Cloud.Packet != nil {
+		return p.createPacketSecret(cluster, userInfo, projectID)
+	}
+	return nil, nil
+}
+
+func (p *CredentialsProvider) createPacketSecret(cluster *kubermaticv1.Cluster, userInfo *provider.UserInfo, projectID string) (*corev1.Secret, error) {
+	// create secret for storing credentials
+	secret := &corev1.Secret{}
+	name := fmt.Sprintf("%spacket-%s", CredentialPrefix, cluster.Name)
```
zreigz: please change the prefix to `credential` and patter to `%s-packet-%s`

api/pkg/provider/kubernetes/sa_token.go
```diff
@@ -79,7 +79,7 @@ func (p *ServiceAccountTokenProvider) Create(userInfo *provider.UserInfo, sa *ku
 		return nil, kerrors.NewInternalError(err)
 	}
 
-	createdToken, err := kubernetesImpersonatedClient.CoreV1().Secrets(kubermaticNamespace).Create(secret)
+	createdToken, err := kubernetesImpersonatedClient.CoreV1().Secrets(resources.KubermaticNamespace).Create(secret)
```
zreigz, line 82: thanks for this :)

zreigz (changes requested):

zreigz:
From the API perspective small changes. 
I've run it locally and I didn't notice any pod deployment for the cluster and I couldn't delete cluster because of the new finalizer. 
We can discuss it tomorrow

api/pkg/clusterdeletion/deletion.go
```diff
@@ -60,6 +62,9 @@ func (d *Deletion) cleanupCluster(ctx context.Context, cluster *kubermaticv1.Clu
 	if err := d.cleanupInClusterResources(ctx, cluster); err != nil {
 		return err
 	}
+	if err := d.cleanUpCredentialsSecrets(ctx, cluster); err != nil {
```
alvaroaleman: Lets do this in the very end, because after this we can not re-create the machine-controller deployment which we need to clean up the nodes which is currently done after cleaning up the `CredentialsSecret`

api/pkg/clusterdeletion/deletion.go
```diff
@@ -203,6 +208,41 @@ func (d *Deletion) cleanupInClusterResources(ctx context.Context, cluster *kuber
...
+
+	secretName := getSecretName(cluster)
+	if secretName == "" {
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	err := d.seedClient.Get(ctx, types.NamespacedName{Name: secretName, Namespace: resources.KubermaticNamespace}, secret)
+	if err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Secret %s/%s: %v", resources.KubermaticNamespace, secretName, err)
```
alvaroaleman, line 231: I wont suggest this very often, but maybe just nest it? As in:

```
if err != nil {
    if kerrors.IsNotfound(err) {
        return nil
    }
return fmt.Errorf(...)
```

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -107,9 +108,20 @@ func CreateEndpoint(sshKeyProvider provider.SSHKeyProvider, cloudProviders map[s
...
```
alvaroaleman, line 113: We don't need this anymore now that the cleanup is done via fa finalizer, right?

We can just create the secret using `generateName` and then use the name we got back. Only drawback is that if we fail creating the cluster, the secret will be kept dangling :grimacing:

api/pkg/provider/kubernetes/cluster.go
```diff
@@ -99,6 +99,10 @@ func (p *ClusterProvider) New(project *kubermaticv1.Project, userInfo *provider.
 	}
 
 	name := rand.String(10)
+	if cluster.Name != "" {
```
alvaroaleman: THis isnt needed, is it?

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,110 @@
...
+		secretLister:                  secretLister,
+	}
+}
+
+// CredentialsProvider manages secrets for credentials
+type CredentialsProvider struct {
+	// kubernetesImpersonationClient is used as a ground for impersonation
+	kubernetesImpersonationClient kubernetesImpersonationClient
+
+	secretLister kubev1.SecretLister
```
alvaroaleman: Can we not use any kind of listers but just the `ctrlruntime` client? Or do we have the problem that we can not get an impersonated version of it?

api/pkg/clusterdeletion/deletion.go
nikhita: Done.

api/pkg/clusterdeletion/deletion.go
nikhita: >     if kerrors.IsNotfound(err) {
>        return nil
>   }

In this case, we won't return nil but instead remove the finalizer in case it exists, right? :thinking:

api/pkg/handler/v1/cluster/cluster.go
nikhita: > Only drawback is that if we fail creating the cluster, the secret will be kept dangling

If we generate the name here, we can still use the cluster name while creating the secret and then if we aren't able to create the cluster, we can delete the secret. This won't keep the secret dangling. :)

Are there drawbacks to this approach though? :thinking:

api/pkg/provider/kubernetes/credentials.go
nikhita: > Or do we have the problem that we can not get an impersonated version of it?

This. :disappointed:

api/pkg/provider/kubernetes/cluster.go
nikhita: (wrote above https://github.com/kubermatic/kubermatic/pull/3684#discussion_r305743642)

api/pkg/provider/kubernetes/credentials.go
alvaroaleman: And actually fixing the controller-runtime client is not really possible is it? Because the lister-backed version of the client would basically have to replicate the whole RBAC logic of the apiserver

api/pkg/handler/v1/cluster/cluster.go
alvaroaleman: Well its obviously not transaction safe. I think we will have to write a simplistic controller that cleans up dangling secrets anyways (The delete-secret-after-cluster-creation-failed could fail). And at the point where we have to do that anyways, we can just use generateName and don't risk getting a 409 back on one of the creations. Or am I missing something?

api/pkg/clusterdeletion/deletion.go
alvaroaleman: Oh yeah indeed, good point

alvaroaleman: (We shouldn't go into the whole cleanup logic for the secret at all if the finalizer doesn't exist, so at this point we can safely assume that it does exist)

api/pkg/provider/kubernetes/credentials.go
alvaroaleman: But wait, the lister itself has the same problem, you can get any object off it, there is no impersonation, right? So using the lister here is basically the same as using a privileged controller-runtime client? Or is the problem that the API doesn't use controller-runtime yet?

nikhita:
Addressed comments. The last run almost worked.

> failed to create machine at cloudprover, due to failed to create server, due to POST `https://api.packet.net/projects/<redacted-project-id-here>/devices: 503` The facility ams1 has no provisionable t1.small.x86 servers matching your criteria

This is a new error :/

api/pkg/provider/kubernetes/credentials.go
nikhita: > And actually fixing the controller-runtime client is not really possible is it?

Hmm, iiuc this will need a field for the impersonated version in the `Client` itself, right? In that case, I'm not sure how we can do it...

nikhita:
/test pull-kubermatic-e2e-packet-coreos-1.14

The secret postsubmit ran successfully

api/pkg/handler/v1/cluster/cluster.go
nikhita: So I had to revert some changes here, PTAL.

Removed the deletion logic from here, but kept name generation here because we need to pass the partial cluster object to the credentials provider so that it can remove the secret values from there. and then we use the partial cluster to actually create the cluster object. 

Or else the fields on the cluster object would not be changed at all.

nikhita:
It's green! PTAL. :)

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -1035,3 +1054,13 @@ func DecodeGetClusterEvents(c context.Context, r *http.Request) (interface{}, er
 
 	return req, nil
 }
+
+func ShouldAddSecretsCleanupFinalizer(cluster *kubermaticv1.Cluster) bool {
```
alvaroaleman: Please just dont do this check and add the finalizer unconditionally, this simplifies the code. The cleanup already is a no-op if the provider is not packet.

api/cmd/projects-migrator/credentials.go
```diff
@@ -0,0 +1,47 @@
...
+func createSecretsForCredentials(cluster *kubermaticv1.Cluster, client kubernetes.Interface) error {
+	if cluster.Spec.Cloud.Packet != nil {
+		return createPacketSecret(cluster, client)
+	}
+	return nil
+}
+
+func createPacketSecret(cluster *kubermaticv1.Cluster, client kubernetes.Interface) error {
+	packetSpec := cluster.Spec.Cloud.Packet
+	referenceNotDefined := packetSpec.APIKeyReference == nil && packetSpec.ProjectIDReference == nil
```
alvaroaleman: please rename to `referenceDefined` and change code accordingly, the cognitive overhead of having to remember that a var describes an unmet condition is a lot higher then the one of negating a variable in an if statement

api/pkg/clusterdeletion/deletion.go
```diff
@@ -203,6 +209,44 @@ func (d *Deletion) cleanupInClusterResources(ctx context.Context, cluster *kuber
...
+
+	secretName := getSecretName(cluster)
+	if secretName == "" {
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	err := d.seedClient.Get(ctx, types.NamespacedName{Name: secretName, Namespace: resources.KubermaticNamespace}, secret)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
```
alvaroaleman, line 226: We need to remove the finalizer here

api/pkg/provider/kubernetes/cluster.go
alvaroaleman: name is already set below in line 109, there is no need for the `if name not empty then set it` you added

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	kubev1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+const CredentialPrefix = "credential"
+
+// NewCredentialsProvider returns a credentials provider
+func NewCredentialsProvider(kubernetesImpersonationClient kubernetesImpersonationClient, secretLister kubev1.SecretLister) (*CredentialsProvider, error) {
```
alvaroaleman, line 20: I don't understand why this doens't just get a `kubernetes.Interface` instead of the `kubernetesImpersonationClient`?

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return &CredentialsProvider{
+		SecretLister:               secretLister,
+		KubernetesClientPrivileged: kubernetesClient,
+	}, nil
+}
+
+// CredentialsProvider manages secrets for credentials
+type CredentialsProvider struct {
+	SecretLister               kubev1.SecretLister
```
alvaroaleman: Is there a reason to export the lister and the client?

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -0,0 +1,97 @@
...
+func (p *CredentialsProvider) Create(cluster *kubermaticv1.Cluster, projectID string) (*corev1.Secret, error) {
+	if cluster.Spec.Cloud.Packet != nil {
+		return p.createPacketSecret(cluster, projectID)
+	}
+	return nil, nil
+}
+
+func (p *CredentialsProvider) createPacketSecret(cluster *kubermaticv1.Cluster, projectID string) (*corev1.Secret, error) {
+	// create secret for storing credentials
+	secret := &corev1.Secret{}
```
alvaroaleman: Can we just directly set all properties?

api/pkg/provider/types.go
```diff
@@ -407,3 +407,9 @@ type EventRecorderProvider interface {
 	// referred by provided cluster config.
 	ClusterRecorderFor(client kubernetes.Interface) record.EventRecorder
 }
+
+// CredentialsProvider declares the set of methods for interacting with kubermatic credential secrets
+type CredentialsProvider interface {
```
alvaroaleman: Can we not define interfaces if we have exactly one implementation? This doesn't add any value in such a case

api/pkg/resources/machinecontroller/credentials.go
```diff
@@ -0,0 +1,47 @@
...
+	APIKey    string
+	ProjectID string
+}
+
+func GetPacketCredentials(data machinecontrollerData) (PacketCredentials, error) {
+	spec := data.Cluster().Spec.Cloud.Packet
+	packetCredentials := PacketCredentials{}
+	var err error
+
+	if spec.APIKey != "" {
```
alvaroaleman, line 17: Should probably a follow-up, but this should be part of `Data`. The cloud-controller needs the exact same code on providers that do have need resources to be created

alvaroaleman: (The interface you built in itself is perfect btw, just defined in the wrong place IMO)

alvaroaleman (commented):
Some more nits. Migration will be a follow-up?

api/pkg/provider/kubernetes/credentials.go
nikhita: > I don't understand why this doens't just get a kubernetes.Interface instead of the kubernetesImpersonationClient?

We can do that as well, but most other providers need the `kubernetesImpersonationClient` directly so it becomes easy and consistent to pass the same thing to all providers.

api/pkg/provider/types.go
nikhita: > Can we not define interfaces if we have exactly one implementation?

There are some other providers with single-implementations... should I changed them in this PR too? I feel like this is consistent but agree that it doesn't add value...

api/pkg/provider/helper.go
```diff
@@ -0,0 +1,13 @@
+package provider
+
+import (
+	"fmt"
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+)
+
+func ValidateSecretKeySelector(selector *providerconfig.GlobalSecretKeySelector, key string) error {
```
nikhita: I don't actually like this here, but I wasn't sure where to place this because it was leading to an import cycle otherwise (after rebasing).

nikhita (commented):
Addressed most comments, and left replies/questions for others.

api/pkg/provider/types.go
alvaroaleman: No, lets just do that for the new `CredentialsProvider`

api/pkg/provider/helper.go
alvaroaleman: Can we place it into the `kuberneteshelper` package (`pkg/kubernetes`)? If not, lets just keep it here

alvaroaleman (commented):
Lets remove the CredentialsProvider interface and then merge this

nikhita:
updated.

alvaroaleman:
/test test pull-kubermatic-e2e-packet-coreos-1.14

Lets merge it if that passes :)

nikhita:
Cleaned up some clusters that didn't get deleted because they were created before the kubermatic-ci postsubmit ran...this should hopefully work

/retest

PR #3692, opened 2019-06-25
more customized user cluster scheduler policy configuration

multi-io:

Set up more customized policy configuration for the user cluster schedulers. See e.g. https://docs.okd.io/latest/admin_guide/scheduling/scheduler.html. Depending on the cloud provider and potentially other attributes of the user cluster, we enable or disable needed/unneeded predicates (basically components of the scheduler's algorithm that determine whether a given to-be-scheduled pod may be scheduled on a given node) of the cluster's scheduler.

We need this to enable the MaxCinderVolumeCount predicate for Openstack clusters. The PR also disables the respective cloud-specific predicates if the cluster uses another cloud provider. For example we don't need the MaxEBSVolumeCount predicate if the cluster doesn't run on AWS.

For the MaxCinderVolumeCount predicate, we'll need a corresponding PR for machine-controller that will write the correct maximum volume count for a node into the cloud-config, from where kubelet will read it and write it into the node object, where the predicate will use it. We'd then also like to make this configurable via datacenters.yaml / machine settings, as discussed in the chat.

```release-note
User cluster scheduler configuration more optimized and configurable
```

### DIFF
--- a/api/pkg/resources/scheduler/configmap.go
+++ b/api/pkg/resources/scheduler/configmap.go
@@ -1,0 +1,162 @@
+package scheduler
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+	corev1 "k8s.io/api/core/v1"
+	"text/template"
+)
+
+type schedulerTplModel struct {
+	TemplateData         resources.TemplateData
+	PolicyConfigFileName string
+}
+
+const (
+	schedulerConfigTpl = `{{ if lt .TemplateData.Cluster.Spec.Version.Semver.Minor 12 -}}
+apiVersion: componentconfig/v1alpha1
+{{- else -}}
+apiVersion: kubescheduler.config.k8s.io/v1alpha1
+{{- end }}
+kind: KubeSchedulerConfiguration
+algorithmSource:
+  policy:
+    file:
+      path: /etc/kubernetes/scheduler/{{ .PolicyConfigFileName }}
+clientConnection:
+  kubeconfig: /etc/kubernetes/kubeconfig/kubeconfig
+`
+
+	schedulerPolicyTpl = `{
+    "apiVersion": "v1",
+    "kind": "Policy",
+    "predicates": [
+{{- if .TemplateData.Cluster.Spec.Cloud.Azure }}
+        {
+            "name": "MaxAzureDiskVolumeCount"
+        },
+{{- end }}
+        {
+            "name": "MatchInterPodAffinity"
+        },
+        {
+            "name": "GeneralPredicates"
+        },
+        {
+            "name": "CheckVolumeBinding"
+        },
+        {
+            "name": "CheckNodeUnschedulable"
+        },
+{{- if .TemplateData.Cluster.Spec.Cloud.AWS }}
+        {
+            "name": "MaxEBSVolumeCount"
+        },
+{{- end }}
+{{- if false }}
+        {
+            "name": "MaxGCEPDVolumeCount"
+        },
+{{- end }}
+        {
+            "name": "NoDiskConflict"
+        },
+        {
+            "name": "NoVolumeZoneConflict"
+        },
+        {
+            "name": "MaxCSIVolumeCountPred"
+        },
+{{- if and .TemplateData.Cluster.Spec.Cloud.Openstack (ge .TemplateData.Cluster.Spec.Version.Semver.Minor 14) }}
+        {
+            "name": "MaxCinderVolumeCount"
+        },
+{{- end }}
+        {
+            "name": "PodToleratesNodeTaints"
+        }
+    ],
+    "priorities": [
+        {
+            "name": "SelectorSpreadPriority",
+            "weight": 1
+        },
+        {
+            "name": "InterPodAffinityPriority",
+            "weight": 1
+        },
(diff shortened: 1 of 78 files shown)


api/pkg/resources/scheduler/configmap.go
```diff
@@ -0,0 +1,162 @@
...
+        },
+{{- end }}
+        {
+            "name": "MatchInterPodAffinity"
+        },
+        {
+            "name": "GeneralPredicates"
+        },
+        {
+            "name": "CheckVolumeBinding"
```
alvaroaleman, line 48: I must admit I am not happy about having to specifiy _all_ predicates here as opposed to just the ones we want on top of the default ones. This basically means we have to check on every release we add if the predicates we configured here match the upstream ones + the additional ones we add.

It there no way to just configure the additional predicates?

bashofmann: From reading the scheduler code (https://github.com/kubernetes/kubernetes/blob/00b0ab86af9427bf60a8ea58fd4e83e14d47a77a/pkg/scheduler/factory/factory.go#L366) it seems to be either it uses the default ones or the configured ones, but I don't see the option to use the default ones and add an additional one.

multi-io: > From reading the scheduler code (https://github.com/kubernetes/kubernetes/blob/00b0ab86af9427bf60a8ea58fd4e83e14d47a77a/pkg/scheduler/factory/factory.go#L366) it seems to be either it uses the default ones or the configured ones, but I don't see the option to use the default ones and add an additional one.

Correct, that's how things are currently. The set of predicates comes from either a predefined "profile" or a user-supplied config file, but not a combination.

alvaroaleman: Okay. Can we then deploy just only for Openstack and only for Kube >= 1.12? We'll basically have to check this on every release now

multi-io: OK, we'll do that. Yes you'd still have to check/adapt the list of predicates for every new k8s version to be supported (even though the list would only be used on Openstack clusters).

multi-io:
Adapted as discussed, rebased onto latest master.

api/pkg/resources/scheduler/configmap.go
```diff
@@ -0,0 +1,162 @@
...
+        },
+        {
+            "name": "CheckNodeUnschedulable"
```
alvaroaleman, line 51: This is missing:

* CheckNodeMemoryPressurePred
* CheckNodeDiskPressurePred
* CheckNodePIDPressurePred
* CheckNodeConditionPred

api/pkg/resources/scheduler/deployment.go
```diff
@@ -103,26 +103,36 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
...
+					MountPath: "/etc/kubernetes/kubeconfig",
+					ReadOnly:  true,
+				},
+				{
+					Name:      resources.CASecretName,
+					MountPath: "/etc/kubernetes/pki/ca",
+					ReadOnly:  true,
+				},
+			}
+			if data.Cluster().Spec.Version.Semver().Minor() >= 12 {
```
alvaroaleman, line 119: Actually I've changed my mind, lets please do this in Openstack only. The whole thing is likely to introduce failures if we dont remember to update the predicates if they change. If that was to happen, Id like to reduce the affected envs as much as possible

api/pkg/resources/scheduler/configmap.go
multi-io: I think those are removed by the TaintNodesByCondition feature gate, which is enabled by default. When running the scheduler with default settings in 1.14 or 1.15, it outputs this at startup:

```
I0722 14:37:15.330035   67910 factory.go:433] Creating scheduler with fit predicates 'map[CheckNodeUnschedulable:{} CheckVolumeBinding:{} GeneralPredicates:{} MatchInterPodAffinity:{} MaxAzureDiskVolumeCount:{} MaxCSIVolumeCountPred:{} MaxEBSVolumeCount:{} MaxGCEPDVolumeCount:{} NoDiskConflict:{} NoVolumeZoneConflict:{} PodToleratesNodeTaints:{}]' and priority functions 'map[BalancedResourceAllocation:{} ImageLocalityPriority:{} InterPodAffinityPriority:{} LeastRequestedPriority:{} NodeAffinityPriority:{} NodePreferAvoidPodsPriority:{} SelectorSpreadPriority:{} TaintTolerationPriority:{}]'
```

...which is exactly what we're configuring in that policy file, minus MaxCinderVolumeCount.

I've now changed this to only roll out the customized scheduler on K8s >=1.14 (the pred/prio sets are identical in 1.14 and 1.15). Didn't want to bother with pre-1.14 configuration.

api/pkg/resources/scheduler/deployment.go
multi-io: If you don't want to merge this, it's perfectly fine with us. It's a feature that we want to provide to our Openstack customers, so we'll use it. The machine-controller PR (https://github.com/kubermatic/machine-controller/pull/584) was important so we don't have to maintain a fork there. I can understand that merging this PR will mean additional work for you guys, and considering that it'll probably be obsoleted by CSI anyway, so it's not a big hassle for us if we maintain this in our fork only.

If you still want to merge it though, I'll gladly append the additional changes (see above) to the PR.

:)

PR #3702, opened 2019-06-28
Adjust CPU resources

xrstf:
We are experiencing timeouts for things that should normally not time out. Like TLS handshakes and node-exporter scrapes. It turns out that very low CPU limits actually work and really limit the CPU time we can spend. Who could have expected that?!

This PR increases the CPU shares for most components and allows more bursting. Likewise, this PR adds resource constraints for Thanos based on our current cloud-eu cluster.

```release-note
NONE
```

### DIFF
--- a/config/monitoring/prometheus/values.yaml
+++ b/config/monitoring/prometheus/values.yaml
@@ -231,7 +231,7 @@ prometheus:
     thanosQuery:
       resources:
         requests:
-          cpu: 500m
+          cpu: 50m
           memory: 64Mi
         limits:
           cpu: 1
--- a/config/monitoring/prometheus/values.yaml
+++ b/config/monitoring/prometheus/values.yaml
@@ -213,13 +213,37 @@ prometheus:
           cpu: 5m
           memory: 32Mi
     thanosSidecar:
-      resources: {}
+      resources:
+        requests:
+          cpu: 100m
+          memory: 32Mi
+        limits:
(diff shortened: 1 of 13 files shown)


config/monitoring/prometheus/values.yaml
```diff
@@ -213,13 +213,37 @@ prometheus:
...
+          memory: 32Mi
+        limits:
+          cpu: 300m
+          memory: 1Gi
     thanosStore:
-      resources: {}
+      resources:
+        requests:
+          cpu: 100m
+          memory: 1536Mi
```
kron4eg, line 227: What a strange number. Where does it come from?

config/monitoring/prometheus/values.yaml
```diff
@@ -213,13 +213,37 @@ prometheus:
...
+        limits:
+          cpu: 250m
+          memory: 3Gi
     thanosQuery:
-      resources: {}
+      resources:
+        requests:
+          cpu: 500m
```
kron4eg: Just to start 50% of vCPU?!

kron4eg (commented):
left couple of questions

config/monitoring/prometheus/values.yaml
xrstf: 3 * 512Mi

kron4eg (approved):
/approve no-issue

PR #3703, opened 2019-06-28
conformance tests: allow scenarios to add multiple NDs

kdomanski:
To run conformance scenarios with multiple availability zones.

```release-note
NONE
```

### DIFF
--- a/api/cmd/conformance-tests/runner.go
+++ b/api/cmd/conformance-tests/runner.go
@@ -27,6 +27,7 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/resources/machine"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -53,7 +54,8 @@ var (
 type testScenario interface {
 	Name() string
 	Cluster(secrets secrets) *kubermaticv1.Cluster
-	Nodes(num int, secrets secrets) *kubermaticapiv1.NodeDeployment
+	Nodes(num int, dc provider.DatacenterSpec, secrets secrets) []kubermaticapiv1.NodeDeployment
+	OS() kubermaticapiv1.OperatingSystemSpec
 }
 
 func newRunner(scenarios []testScenario, opts *Opts) *testRunner {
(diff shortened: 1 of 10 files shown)


api/cmd/conformance-tests/runner.go
```diff
@@ -53,7 +54,8 @@ var (
 type testScenario interface {
 	Name() string
 	Cluster(secrets secrets) *kubermaticv1.Cluster
-	Nodes(num int, secrets secrets) *kubermaticapiv1.NodeDeployment
+	Nodes(num int, dc provider.DatacenterSpec, secrets secrets) []kubermaticapiv1.NodeDeployment
```
alvaroaleman: Node introduced by you, but could we rename this to `NodeDeployments`?

api/cmd/conformance-tests/runner.go
```diff
@@ -537,65 +543,84 @@ func (r *testRunner) setupNodes(parentLog *logrus.Entry, scenarioName string, cl
...
-		return fmt.Errorf("failed to get MachineDeployment from NodeDeployment: %v", err)
-	}
 
-	log = log.WithField("machineDeployment", machineDeployment.Name)
-	if err := retryNAttempts(defaultAPIRetries, func(attempt int) error {
-		if err := client.Create(ctx, machineDeployment); err != nil {
-			if kerrors.IsAlreadyExists(err) {
-				return nil
+		// we try to subract the non-owned machines sort-of uniformly from all node deployments
+		for machinesWithoutOwner > 0 {
```
alvaroaleman: I think we can remove the whole code to handle machines without owner. It was needed for upgrade tests from a version of the conformance-tester that didn't use NodeDeployments yet. The one in 2.10 has it already, thought.

api/cmd/conformance-tests/runner.go
```diff
@@ -822,13 +847,18 @@ func (r *testRunner) getGinkgoRuns(
...
 	cluster *kubermaticv1.Cluster,
-	nd *kubermaticapiv1.NodeDeployment,
+	nd []kubermaticapiv1.NodeDeployment,
 	dc provider.DatacenterMeta,
 ) ([]*ginkgoRun, error) {
 	kubeconfigFilename = path.Clean(kubeconfigFilename)
 	repoRoot := path.Clean(r.repoRoot)
 	MajorMinor := fmt.Sprintf("%d.%d", cluster.Spec.Version.Major(), cluster.Spec.Version.Minor())
 
+	nodesTotal := int32(0)
```
alvaroaleman: Can we rename this to `nodNumberTotal`?

api/cmd/conformance-tests/runner.go
```diff
@@ -876,19 +906,31 @@ func (r *testRunner) getGinkgoRuns(
...
 		}
 
 		args = append(args, "--provider=local")
 
-		if nd.Spec.Template.OperatingSystem.Ubuntu != nil {
+		// verify that all operating system specs are identical
+		var osSpec *kubermaticapiv1.OperatingSystemSpec
+		for i, ndi := range nd {
+			if osSpec == nil {
+				osSpec = &ndi.Spec.Template.OperatingSystem
```
alvaroaleman, line 892: Why is this needed? Shouldn't it be enough to verify that the operating system is identical?

kdomanski: If I compare the whole OS spec then there's no need for comparing every OS's spec to nil in order to determine which OS is actually configured.

kdomanski:
/hold
Let's wait for the new release branch split before merging this.

PR #3706, opened 2019-07-01
fix Kibana index setup code

xrstf:
Since the botched Elasticsearch 7.0 upgrade we're using the new, shiny `ELASTICSEARCH_HOSTS` variable, which broke the setup index-pattern container. This PR fixes that and also explicitly checks the API response *before* attempting to pipe it into jq, so that when Kibana responds with "Kibana server is not ready yet" jq will not report a parse error.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #3714, opened 2019-07-01
conformance script: build both the base and target of upgrade test

kdomanski:
Some upgrade tests in pre-submits fail because they assume the images for base version to be present in the CI's registry. This commit ensures that the base has been built, just like it is already being ensured for the upgrade target.

https://prow.loodse.com/view/gcs/prow-data/pr-logs/pull/kubermatic_kubermatic/3712/pull-kubermatic-e2e-upgrade-2.10-2.11/1145711157407911936

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci/ci-run-minimal-conformance-tests.sh
@@ -120,24 +120,28 @@ vault kv get -field=datacenters.yaml \
 echodate "Successfully got secrets from Vault"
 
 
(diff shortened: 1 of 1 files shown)


kdomanski:
@alvaroaleman `pull-kubermatic-e2e-upgrade-2.10-2.11` is configured to only run on the `release/v2.11` branch

alvaroaleman:
Okay. We need this change at least in the release/v2.11 branch, right?

Issue #3538, opened 2019-05-27
Add a shellcheck presubmit to Kubermatic


alvaroaleman:
**User Story**

The amount of bashscripts we are using within Kubermatic for e2e tests is growing. To prevent issues, we should lint them with shellcheck

**Acceptance criteria**

* There is a presubmit on kubermatic/kubermatic that executes shellcheck for all scripts we have
* The presubmit does not fail

alvaroaleman:
This was done in kubermatic/infra#592

PR #3715, opened 2019-07-01
add shellcheck binary to the kubernetes-test-binaries image

kdomanski:
A step towards #3538 

```release-note
NONE
```

### DIFF
--- a/containers/conformance-tests/Dockerfile
+++ b/containers/conformance-tests/Dockerfile
@@ -12,7 +12,7 @@ FROM golang:1.11.4-stretch
 
 COPY --from=0 /opt/kube-test /opt/kube-test
 COPY start-docker.sh /usr/local/bin/start-docker.sh
-RUN apt update && apt install -y git-crypt curl apt-transport-https software-properties-common unzip jq && \
+RUN apt update && apt install -y git-crypt curl apt-transport-https software-properties-common unzip jq shellcheck && \
   curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - && \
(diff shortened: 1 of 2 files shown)


containers/conformance-tests/Dockerfile
```diff
@@ -12,7 +12,7 @@ FROM golang:1.11.4-stretch
 
 COPY --from=0 /opt/kube-test /opt/kube-test
 COPY start-docker.sh /usr/local/bin/start-docker.sh
-RUN apt update && apt install -y git-crypt curl apt-transport-https software-properties-common unzip jq && \
+RUN apt update && apt install -y git-crypt curl apt-transport-https software-properties-common unzip jq shellcheck && \
```
alvaroaleman, line 15: This image is only used for e2e tests and should not contain anything thats not required for those e2e tests, its already huge. We can use any docker image for the shellsheck presubmit

kdomanski: Alright, what docker image would you put it in?

alvaroaleman: What about just using https://hub.docker.com/r/koalaman/shellcheck/tags?

kdomanski: That image is just a static binary based on the scratch image and it process globs, so in order to use it we would have to specify every script file in the args. An alternative would be to rebuild it ourselves based on alpine, so that we can use `sh` to grab all files with a glob.

alvaroaleman: Sure, but either way it does not belong into the conformance-tests image

kdomanski:
closing in favor of #3834

PR #3719, opened 2019-07-02
mark Thanos support as experimental for upcoming 2.11 release

xrstf:
Considering we're not 100% sure that Thanos is our way to go and since VictoriaMetrics is a strong contender, we should not make Kubermatic customers jump on the Thanos bandwaggon just now. *If* we decide VM is better than Thanos, a migration would probably be not easy/impossible.

```release-note
NONE
```


kron4eg (approved):
/approve no-issue

PR #3721, opened 2019-07-02
fix resources-per-namespace-and-node dashboard

xrstf:
cAdvisor publishes metrics for each container *and then once per pod*. This leads to metrics like these:

screenshot-2019-07-02-115516

It's obvious that the non-container'ed metric is the sum of all containers. Without excluding these cumulative metrics, we would count everything twice.

Also, it turns out that we have other cadvisors in our cluster and this can interfere with the CPU metrics. This is why the PR now also explicitly restricts the queries to *our* cadvisor.

The same fixes are applied to a couple of dashboards. Finally some of the weird, broken charts begin to make sense :-)

```release-note
NONE
```

### DIFF
--- a/config/monitoring/grafana/Chart.yaml
+++ b/config/monitoring/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 1.0.7
+version: 1.0.8
 appVersion: 6.2.1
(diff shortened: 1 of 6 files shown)


kron4eg (approved):
/approve no-issue

kron4eg (approved):
/lgtm

Issue #3456, opened 2019-05-10
AWS: Multi AZ support for nodes in controllers


alvaroaleman:
**User Story**

As a user of Kubermatic I want to spread my nodes accross AZs in an AWS region.

**Further context:**

* https://github.com/kubermatic/kubermatic/issues/678
* https://github.com/kubermatic/kubermatic/issues/3139

**Acceptance criteria**

* The controllers support this use case
* The code still builds
* PVs and LBs still work
* Change the `zone_character string` in the  AWS datacenter spec to `zone_characters []string` (Potentially leave the old setting and just add it to `zone_characters` for compatibility reasons)
* The AWS cloudspec currently contains a `SubnetID` string. It is only used during machine creation. We must change this to a `map[string]string`, containing an AZ/SubnetID mapping
* We pass the Zone into the controller-manager and APIServer both via an env var `AvailabilityZone` and as part of the `cloud-config`. Remove this
* We set the `SubnetID` in the cloud-config, remove this
* Add a validation for `datacenters.yaml`, that in an AWS node dc `len(zone_characters) >= 1`
* Change all current references from the API to the datacenter specs `zone_character` to `zone_characters[0]`, until the API and UI fully support multiple zones
* Change all current references from the API to the `cluster.Spec.Cloud.AWS.SubnetID` to `cluster.Spec.Cloud.AWS.SubnetIDs[dc.zone_characters[0]]`, until the API fully supports multiple zones

mrIncompetent:
> We pass the Zone into the controller-manager and APIServer both via an env var AvailabilityZone and as part of the cloud-config. Remove this

Looking at the aws code its required: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1210

We can create a zone based on the region + the first entry in `zone_characters`

> We set the SubnetID in the cloud-config, remove this

Required as well, unless we switch to the RoleARN functionality.
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1272

We could take the subnet matching the first zone though.

alvaroaleman:
> Looking at the aws code its required: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1210

[...]

PR #3725, opened 2019-07-02
conformance tests: use multiple AWS AZs

kdomanski:
A step towards #3456 

```release-note
NONE
```

### DIFF
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -738,6 +738,10 @@ type AWSNodeSpec struct {
 	AMI string `json:"ami"`
 	// additional instance tags
 	Tags map[string]string `json:"tags"`
+	// Availiability zone in which to place the node. It is coupled with the subnet to which the node will belong.
+	AvailabilityZone string `json:"availabilityZone"`
+	// The VPC subnet to which the node shall be connected.
+	SubnetID string `json:"subnetID"`
 }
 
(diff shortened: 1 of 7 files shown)


api/pkg/api/v1/types.go
```diff
@@ -738,6 +738,10 @@ type AWSNodeSpec struct {
 	AMI string `json:"ami"`
 	// additional instance tags
 	Tags map[string]string `json:"tags"`
+	// Availiability zone in which to place the node. It is coupled with the subnet to which the node will belong.
+	AvailabilityZone string `json:"availabilityZone"`
+	// The VPC subnet to which the node shall be connected.
+	SubnetID string `json:"subnetID"`
```
mrIncompetent, line 744: Do we already have an idea on how to handle this from a UI perspective?

kdomanski: Yes, the list of AZ-Subnet pairs will be returned by the API and one will need to be picked fro the NodeDeployment in the UI.

mrIncompetent: How does the API load the pairs & does the user need to select them?
Does the API simply load all available subnets?

kdomanski: All subnets in the chosen VPC will be fetched from AWS and the user ~~chooses~~ will choose one.

Issue #2398, opened 2018-11-24
Migrate from docker hub to quay


scheeles:
**User Story**
Currently some party of our images are on docker hub and some are one quay. 

In a first step deploy all image additional to docker hub to quay and change all helm charts.

- [ ] Kubermatic api/controller
- [ ] Kubermatic dashboard

**Acceptance criteria**
With kubermatic 2.9 only quay is required and we can mask docker hub as deprecated.

scheeles:
@mrIncompetent: you add the images which are missing

kdomanski:
All images have been going to Quay and getting pulled only from there as well. We can start tearing down the parts of CI that push to Docker Hub.

mrIncompetent:
Kubermatic api/controller
 Kubermatic dashboard

Both are being deployed to quay only.

PR #3726, opened 2019-07-02
use machine controller image from quay.io instead of dockerhub

kdomanski:
It's not really in scope of #2398, but it's nice-to-have long term.

```release-note
NONE
```


alvaroaleman:
This cant work because we currently dont push the machine-controller to quay: https://prow.loodse.com/view/gcs/prow-data/logs/ci-push-machine-controller-image/1148579663019249664

Closing.

kdomanski:
Yeah, I was going to start a discussion about pushing to quay and then decided it's not worth the trouble.

PR #3745, opened 2019-07-04
Add new endpoint to get GCP zones for the datacenter

### LINKED ISSUE #3737: Implement endpoint to get GCP zones
**User Story**:
When creating a cluster or nodes on GCP, I'd like to be able to pick one of the available zones from a dropdown instead of having to know what they are.

**Acceptance criteria**:
Implement  `/providers/gcp/zones` endpoint

zreigz:
**What this PR does / why we need it**: Add new endpoint to get GCP zones for the datacenter

Fixes #3737 

```release-note
new endpoint for GCP zones:
GET /api/v1/providers/gcp/{dc}/zones
```

### DIFF
--- a/api/pkg/handler/v1/provider/gcp.go
+++ b/api/pkg/handler/v1/provider/gcp.go
@@ -2,28 +2,77 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/go-kit/kit/endpoint"
+	"github.com/gorilla/mux"
 	"google.golang.org/api/compute/v1"
 
 	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/handler/v1/common"
+	"github.com/kubermatic/kubermatic/api/pkg/handler/v1/dc"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/provider/cloud/gcp"
+	"github.com/kubermatic/kubermatic/api/pkg/util/errors"
 )
+
+// GCPZoneReq represent a request for GCP zones.
+// swagger:parameters listGCPZones
+type GCPZoneReq struct {
+	GCPCommonReq
+	// in: path
+	// required: true
+	DC string `json:"dc"`
+}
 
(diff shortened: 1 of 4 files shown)


api/pkg/handler/v1/provider/gcp.go
```diff
@@ -124,3 +173,60 @@ func listGCPSizes(ctx context.Context, sa string, zone string) (apiv1.GCPMachine
...
+				}
+			}
+		}
+
+		return listGCPZones(ctx, sa, req.DC, dcs)
+	}
+}
+
+func listGCPZones(ctx context.Context, sa, datacenterName string, dcs map[string]provider.DatacenterMeta) (apiv1.GCPZoneList, error) {
+	supportedZones := make(map[string]struct{})
```
alvaroaleman: Not sure if this makes sense,  IMHO we should remove the current suffix configuration, right now we basicallý only verify that the zone speicifed in the datacenters.yaml actually exists.

There is btw `k8s.io/apimachinery/pkg/util/sets` which is a bit nicer to use than a `map[string]struct{}`

zreigz: I can do something like this (filter by region):
```
func listGCPZones(ctx context.Context, sa, datacenterName string, dcs map[string]provider.DatacenterMeta) (apiv1.GCPZoneList, error) {
	zones := apiv1.GCPZoneList{}

	datacenter, err := dc.GetDatacenter(dcs, datacenterName)
	if err != nil {
		return nil, errors.NewBadRequest("%v", err)
	}

	if datacenter.Spec.GCP == nil {
		return nil, errors.NewBadRequest("the %s is not GCP datacenter", datacenterName)
	}
	
	computeService, project, err := gcp.ConnectToComputeService(sa)
	if err != nil {
		return nil, err
	}

	req := computeService.Zones.List(project)
	err = req.Pages(ctx, func(page *compute.ZoneList) error {
		for _, zone := range page.Items {

			if strings.HasPrefix(zone.Name, datacenter.Spec.GCP.Region) {
				apiZone := apiv1.GCPZone{Name: zone.Name}
				zones = append(zones, apiZone)
			}
		}
		return nil
	})

	return zones, err
}
```

zreigz: @alvaroaleman thanks for your support :+1:

PR #3751, opened 2019-07-05
Support setting oidc authentication settings on cluster

bashofmann:

Support setting oidc authentication settings on cluster entity and configuring the apiserver because of it

This also adds an endpoint to download a kubeconfig that has these settings included as well.

Background: We want to enable setting individual OpenID Connect issuers for a cluster, so that customers can authenticate against a cluster with their own identity provider.

```release-note
Support setting oidc authentication settings on cluster
```


alvaroaleman:
Hmhm, interesting idea. It may clash with the existing share cluster feature, but we can add a check that refuses setting the `OIDC config` via our api if the "share cluster" is active.

WDYT @zreigz ?

bashofmann:
The way I implemented it at the moment is, that if you set a oidc config over the cluster resource it has precedence over the oidc settings of the share cluster feature (https://github.com/kubermatic/kubermatic/pull/3751/files#diff-8884b018662d2bd252b7e4c1c56d98a7R258). But I'm definitely open doing it in another way if it works better for your use case.

zreigz:
It's a realy nice feature. For me looks fine. The new endpoint just gets master kubeconfig. Please rebase your code and then we will run the tests

zreigz:
@alvaroaleman you are right it clashes with share cluster feature https://docs.kubermatic.io/advanced/oidc_auth/
The `OIDC config` shouldn't be set when `OIDCKubeCfgEndpoint` feature is enabled
I can add this check after we merge this PR

alvaroaleman:
> I can add this check after we merge this PR

Great thanks! And just for my own clarification, the goal here is to eventually allow users that create clusters to set the OIDC config via UI, right?

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -99,6 +99,18 @@ type ClusterSpec struct {
...
+}
+
+type OIDCSettings struct {
+	IssuerURL     string `json:"issuerUrl,omitempty"`
+	ClientID      string `json:"clientId,omitempty"`
+	ClientSecret  string `json:"clientSecret,omitempty"`
+	UsernameClaim string `json:"usernameClaim,omitempty"`
+	GroupsClaim   string `json:"groupsClaim,omitempty"`
+	RequiredClaim string `json:"requiredClaim,omitempty"`
+	ExtraScopes   string `json:"extraScopes,omitempty"`
```
alvaroaleman, line 93: The `ExtraScopes` field is never used or am I missing it?

alvaroaleman: nvm, found it

bashofmann: I decided to add this field, even though the apiserver does not need it, so that users can easily download a kubeconfig that contains this field and is ready to use with tools like https://github.com/int128/kubelogin.

When authenticating with Google you need to set ExtraScopes for example so that Google also returns the E-Mail address.

bashofmann:
> the goal here is to eventually allow users that create clusters to set the OIDC config via UI, right?

Yes, definitely. We are a bit unsure ourselves though ourselves at the moment how to do this best in the UI. We have a proof-of-concept and are at the moment playing around with it a bit and getting internal feedback before actually creating a PR at the dashboard.

Here two, definitely not final, screenshots of the our current POC:

Bildschirmfoto 2019-07-10 um 10 41 56
Bildschirmfoto 2019-07-10 um 10 42 05

bashofmann:
Is there still something missing that I need to do to get it merged?

zreigz:
please update the swagger spec. I will approve it after your commit. Thanks again for your PR :+1:

bashofmann:
I can't see the details of why the job failed, but I rebased against master and updated the swagger speck.

PR #3753, opened 2019-07-05
Add addon: component-override-controller hack.

thz:

This PR adds a tiny controller which patches the spec.componentOverride.apiserver.replicas for clusters which do not have it set. Filtering by a label selector is support (e.g. for project-id).

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/addons/component-override-controller/coc-config.yaml
+++ b/addons/component-override-controller/coc-config.yaml
@@ -1,0 +1,9 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: component-override-controller-config
+  namespace: kubermatic
+data:
+  PROJECT_LIST: ""
+  DESIRED_REPLICAS: "3"
--- a/addons/component-override-controller/coc-deployment.yaml
+++ b/addons/component-override-controller/coc-deployment.yaml
@@ -1,0 +1,58 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: component-override-controller
+  name: component-override-controller
+  namespace: kubermatic
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: component-override-controller
+  strategy: {}
+  template:
+    metadata:
+      labels:
+        app: component-override-controller
+    spec:
+      serviceAccountName: kubermatic
+      containers:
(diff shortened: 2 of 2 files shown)


thz:
If we actually want to have this: should we put it in the addons or in some hack/ folder? I am not sure what the right place would be.

mrIncompetent:
I very much like the idea of having this as an example for customers so they can do custom changes if required.

Though i have no idea what the best place would be. Infra? kubermatic docs?

thz:
> I very much like the idea of having this as an example for customers so they can do custom changes if required.
> 
> Though i have no idea what the best place would be. Infra? kubermatic docs?

Would we be able to have it in the addons/ but not enabled by default?

mrIncompetent:
@thz 
It does not belong into the addons directory.
> In Kubermatic, Addons are manifests/applications we deploy into user clusters.
> That controller belongs into the seed cluster.

@toschneck suggested https://github.com/kubermatic/customer-projects which sounds like the best place - as this is a nice example for customers / POC's with special needs

thz:
> @thz
> It does not belong into the addons directory.
> 
> > In Kubermatic, Addons are manifests/applications we deploy into user clusters.
> > That controller belongs into the seed cluster.

Oh dang! Of course.

> 
> @toschneck suggested https://github.com/kubermatic/customer-projects which sounds like the best place - as this is a nice example for customers / POC's with special needs

Hmm. But customer-projects is not "delivered" to customers. Perhaps a documentation repository is an idea too?

toschneck:
@thz I would move the used implementation to the customer repo. 
In the docs it would be great to have a generic example, what could be based on this one. I wouldn't maintain real used code in the docs repository

mrIncompetent:
> @thz I would move the used implementation to the customer repo.
In the docs it would be great to have a generic example, what could be based on this one. I wouldn't maintain real used code in the docs repository

Sounds great!
@thz would you mind opening a PR in https://github.com/kubermatic/customer-projects ?
Also: Please open a PR for the docs - Or create an issue for the docs :)

thz:
Good to see that we are not mixing stuff for operating specific instances (`cloud` / `dev`) into the code repository of the product :-)

PR is over there:
https://github.com/kubermatic/customer-projects/pull/66

Issue #3626, opened 2019-06-13
Create Kubermatic Operator


xrstf:
This is a proposal to create a Kubermatic Operator to deploying and managing the various parts that make up a Kubermatic setup.

### Motivation

* Managing complexer migrations via Helm is complicated and brittle.
* [add more pls]

### Proposal

A new operator takes care of the Kubermatic stacks. What was previously the `values.yaml` is replaced with one or possibly multiple CRDs (i.e. one for kubermatic, monitoring, logging).

I would start by implementing the operator logic for a simple chart like the s3-exporter and once we are sure this is the way to go, would then go on with Kubermatic itself and then the rest.

#### How to retain configuration flexibility?

* The monitoring stack especially allows to freely mix and match how and where Grafana dashboards, Prometheus rules etc. are located.

### Acceptance criteria

* Helm isn't used anymore.
* All configuration is done via a new CRD.
* It should be much, much easier to configure the Kubermatic stack through the new CRD we introduce.
* A deployment of Kubermatic means to deploy only that operator and fill out the CR, the rest happens automagically.

alvaroaleman:
* Helm isn't used anymore
* All configuration is done via a new CRD
* A deployment of Kubermatic means to deploy only that operator and fill out the CR, the rest happens automagically

xrstf:
* It should be much, much easier to configure the Kubermatic stack through the new CRD we introduce.
* Think about maybe splitting the configuration into simple/extended or mandatory/optional or something along the lines.

alvaroaleman:
On a very high level I would consider one of the main goals with this: "Allow someone with a Kubernetes cluster and without knowledge about kubermatic architecture to use this operator and less than a page of docs to set up a functioning Kubermatic installation"

xrstf:
Most of the grunt work is done, the remaining stuff is tracked in https://github.com/kubermatic/kubermatic/issues/4723.

Issue #2113, opened 2018-10-01
Move datacenters into CRD


mrIncompetent:
Epic to group all actions required to move the `datacenters.yaml` into CRD's

**API**
- ~Implement endpoint to return SeedDatacenter's from CRD instead of file~
- ~Implement endpoint to return NodeLocation's from CRD instead of file~
- Implement a new endpoint that returns `Datacenters` only and not `Seeds`
- Make the new endpoint return `Datacenters` with a new type that doesn't contain any info about their seed
- Mark the existing endpoint that returns both `Seeds` and `Datacenters` as deprecated

**UI**
- ~Replace existing datacenters endpoints with new SeedDatacenter & NodeLocation endpoints~ Use  the new `Datacenter` endpoint that returns `datacenters` only and not `Seeds`

**Controller**
- ~[Add new CRD's](https://github.com/kubermatic/kubermatic/issues/2114)~
- ~[Add migration step to migrate the existing datacenters.yaml to CRD's](https://github.com/kubermatic/kubermatic/issues/2116)~
- [Remove handling of datacenters in all controllers by copying all fields to the cluster spec](https://github.com/kubermatic/kubermatic/issues/2118)
- [Implement a `MutatingAdmissionWebhook` to handle defaulting & validation for clusters](https://github.com/kubermatic/kubermatic/issues/1688)

**Docs**
- Document required migration steps
- Document how to create new SeedDatacenter's & NodeLocation's

alvaroaleman:
cc @kdomanski

PR #3767, opened 2019-07-08
Kubermatic Operator Proposal

xrstf:
This adds the proposal for the Kubermatic Operator. Formalises #3626.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/kubermatic-operator.md
+++ b/docs/proposals/kubermatic-operator.md
@@ -25,9 +25,9 @@ Let us create a Kubermatic Operator that takes care of setting up the various co
 required for running Kubermatic in production. It should be configured via one or multiple
 CRDs and over time replace Helm entirely.
 
-In this first iteration we would focus on installing *Kubermatic*. The logging and monitoring
+In this first iteration we will focus on installing *Kubermatic*. The logging and monitoring
 stacks are out-of-scope for now, as we need to decide if and how to migrate the existing
-configuration flexibility to a 3rd party operator or something we would write ourselves.
+configuration flexibility to a 3rd party operator or something we write ourselves.
 
 An example for a minimal starter configuration could look like this:
 
--- a/docs/proposals/kubermatic-operator.md
+++ b/docs/proposals/kubermatic-operator.md
@@ -50,7 +50,7 @@ spec:
   # You must configure credentials for quay.io and docker.io at the
   # very least.
   secrets:
-    dockerAuth: |
+    imagePullSecret: |
       {
         "auths": {
           "https://index.docker.io/v1/": {
--- a/docs/proposals/kubermatic-operator.md
+++ b/docs/proposals/kubermatic-operator.md
@@ -123,8 +123,6 @@ created by the cluster admin.
   flag for clusters).
 * We want to put the easy-configuration-experience into a (future) CLI that asks simple
   questions and generates one or more CRD manifests.
-* The operator is solely responsible for managing Kubermatic installations. It’s not a
-  toolset for consulting jobs to setup pretty monitoring stacks.
 
 ## Non-Goals
 
@@ -135,3 +133,5 @@ created by the cluster admin.
 * There is no need for a Big Bang release where we replace all charts with the operator
   at one point in time. Instead we will have a migration phase where both the operator and
   the charts are used.
+* The operator is solely responsible for managing Kubermatic installations. It’s not a
+  toolset for consulting jobs to setup pretty monitoring stacks.
--- a/docs/proposals/kubermatic-operator.md
+++ b/docs/proposals/kubermatic-operator.md
@@ -80,17 +80,13 @@ spec:
 
   # Feature gates are structs instead of simple booleans to make
   # extending individual features with additional fields easier.
-  api:
-    featureGates:
-      OIDCKubeCfgEndpoint:
-        enabled: true
-
-  controller:
-    featureGates:
-      OpenIDAuthPlugin:
-        enabled: true
-      VerticalPodAutoscaler:
-        enabled: true
+  featureGates:
+    OIDCKubeCfgEndpoint:
+      enabled: true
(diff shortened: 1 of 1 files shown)


docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+* The existing Kubermatic installer is barely maintained by a single person in seed-team
+  and quickly grows ouf-of-sync with actual Kubermatic.
+
+## Proposal
+
+Let us create a Kubermatic Operator that takes care of setting up the various components
+required for running Kubermatic in production. It should be configured via one or multiple
+CRDs and over time replace Helm entirely.
+
+In this first iteration we would focus on installing *Kubermatic*. The logging and monitoring
```
alvaroaleman: s/would/will

docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+  secrets:
+    dockerAuth: |
+      {
+        "auths": {
+          "https://index.docker.io/v1/": {
+            "auth": "[base64-encoded credentials here]",
+            "email": ""
+          },
+          "quay.io": {
+            "auth": "[base64-encoded credentials here]",
```
alvaroaleman, line 61: I would actually make this a kv par, key is the registry, value is the credential

docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Proposal
+
+Let us create a Kubermatic Operator that takes care of setting up the various components
+required for running Kubermatic in production. It should be configured via one or multiple
+CRDs and over time replace Helm entirely.
+
+In this first iteration we would focus on installing *Kubermatic*. The logging and monitoring
+stacks are out-of-scope for now, as we need to decide if and how to migrate the existing
+configuration flexibility to a 3rd party operator or something we would write ourselves.
```
alvaroaleman: s/would//

alvaroaleman (commented):
/approve

kron4eg:
Looks like Datacenter CR is missing...

alvaroaleman:
> Looks like Datacenter CR is missing...

That is not part of this, datacenters as CRDs is tracked in https://github.com/kubermatic/kubermatic/issues/2113

docs/proposals/kubermatic-operator.md
xrstf: But then we need to understand and assemble the docker config file ourselves *and* admins could not just simply copy their existing file into the Kubermatic configuration. For this I think not touching the data structure is a good idea.

docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+  # Whether or not this installation is a master, i.e. where
+  # the Kubermatic dashboard and API should be deployed.
+  isMaster: true
+
+  # The secrets are used to pull images from private Docker repositories;
+  # this is effectively a copy of the "auth" section in your ~/.docker/config.json.
+  # You must configure credentials for quay.io and docker.io at the
+  # very least.
+  secrets:
+    dockerAuth: |
```
mrIncompetent: nit: can we rename this to `imagePullSecret` ?

docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+
+  # Feature gates are structs instead of simple booleans to make
+  # extending individual features with additional fields easier.
+  api:
+    featureGates:
+      OIDCKubeCfgEndpoint:
+        enabled: true
+
+  controller:
+    featureGates:
```
mrIncompetent: Why not have one global "featureGates" setting?
A customer should not have to worry to which component a feature flag belongs

docs/proposals/kubermatic-operator.md
```diff
@@ -0,0 +1,137 @@
...
+  about Monitoring/Logging stacks.
+* Get rid of all Helm charts (long term).
+* Retain most of the flexibility of our charts, not necessarily using the same configuration
+  structure (from the Helm values.yaml).
+* Be able to opt-out of the operator managing certain pieces of the stacks (like the "pause"
+  flag for clusters).
+* We want to put the easy-configuration-experience into a (future) CLI that asks simple
+  questions and generates one or more CRD manifests.
+* The operator is solely responsible for managing Kubermatic installations. It’s not a
+  toolset for consulting jobs to setup pretty monitoring stacks.
```
mrIncompetent: > It’s not a toolset for consulting jobs to setup pretty monitoring stacks.

That belongs into the "Non-Goals" section

docs/proposals/kubermatic-operator.md
xrstf: Because I pitched a similar idea about a single storage section where you could configure all the various volume sizes and a single resources section where all resources are configured, but nobody seemed to like that idea. That's why i kept the original structure where we distinguish between API, UI, ...

alvaroaleman: Well, storage is a low-level sub-setting of individual parts, whereas a "Feature" currently is always part of Kubermatic as a whole

docs/proposals/kubermatic-operator.md
alvaroaleman: Which existing config? There is none. We hand out a key for quay and that has to be put into it.

xrstf: Do we not give customers exactly one JSON document for the pull secret? I thought customers receive a JSON blob and that's it, not that they have to manually assemble JSON and insert keys somewhere.

xrstf: We have decided (in Slack) that because sites like Quay hand out full JSON documents and not just keys, that we want to keep it that way and just passthrough the JSON to where it's needed, without having any deeper insight in its structure.

PR #3779, opened 2019-07-09
Update Prometheus to 2.11

xrstf:
This updates Prometheus to the latest stable. The new version supports compressing the WAL, which for us saves roughly 40-50% of disk space :-) As disabling compression is not supported without also wiping the WAL, we keep the compression disabled by default for now, but will disable it on our clusters later.

```release-note
update Prometheus to 2.11.0
```


xmudrii (approved):
/approve

PR #3783, opened 2019-07-10
disable unused seed-proxy controller

xrstf:
The seed-proxy controller is responsible for opening a communication channel from the master cluster to each channel. This is nice, but not tested well and not yet used for Prometheus/Alertmanager. This PR therefore disables the controller entirely in 2.11 so that we do not create resources that are not used at all.

```release-note
NONE
```


kron4eg:
linter is not happy

```
cmd/master-controller-manager/controllers.go:175:6: func `createSeedProxyController` is unused (megacheck.{unused,gosimple})
14
func createSeedProxyController(ctrlCtx *controllerContext) (runnerFn, error) {
15
     ^
16
```

kron4eg (approved):
/approve no-issue

Issue #3748, opened 2019-07-04
Limit concurrent cluster upgrades


mrIncompetent:
**User Story**
As a user of Kubermatic i need to limit the amount of clusters which can be updated concurrently to avoid bringing down the seed cluster.

Example:
A seed cluster with 100 clusters, where the Kubernetes version of each cluster gets updated due to a CVE. Currently this will update at least 300 deployments inside the seed cluster (basically the control plane). This means at least 300 new pods will be scheduled (Due to the RollingUpdate strategy). 

To avoid putting too much load on the seed cluster, each controller (Which manages Kubernetes resources), must set a "Condition" on the cluster object if any resource is still upgrading.
If there are already too much(defined by a flag) clusters which have the condition set to true, the controller will just enqueue the cluster object and skip processing.

**Edit**
After reconsidering this issue, we have decided instead of setting a condition in the cluster status which marks the cluster updates are being in progress, we set a condition on the cluster status to make that the updates have been rolled out successfully and if something goes wrong, this condition will be cleared from the status. 

Example:
The cluster controller updates the APIserver deployment. If the Deployment is still updating at the end of the sync, the cluster controller will set the `ClusterControllerUpdateInProgress` condition to `true` - otherwise `false`.

**Edit Example**
The cluster controller updates the APIserver deployment. If the Deployment has finished updating successfully at the end of the sync, the cluster controller will set the `ClusterControllerFinishedUpdatingSuccessfully` condition.

Important notes:
- To avoid having to wait until the deployment controller picks up the updated deployment (So we actually know that the deployment is in the upgrade process), we will reset the status on the deployment. This way we know that a deployment is not up-to-date.
- We have no conditions on the cluster object yet, thus we must define them (In a separate PR).
We must stick as close as possible to the Kubernetes conventions.

**Acceptance criteria**
- [ ] The cluster object has been extended by the conditions
- [ ] All controllers which update resources inside the seed cluster set a individual condition on the cluster object
- [ ] All controllers which update resources inside the seed cluster will only process a limited clusters concurrently.
- [ ] The User can specify the maximum number of concurrent cluster updates via a flag in the helm chart

mrIncompetent:
Reopening since https://github.com/kubermatic/kubermatic/pull/4115 only addressed a subset of this issue

PR #3784, opened 2019-07-10
extend cluster health endpoint for new status

### LINKED ISSUE #1485: Improve cluster's health endpoint
At the moment the data we return from `cluster/{cluster_name}/health` endpoint are very basic :
```
type NewClusterHealth struct {
	Apiserver         bool `json:"apiserver"`
	Scheduler         bool `json:"scheduler"`
	Controller        bool `json:"controller"`
	MachineController bool `json:"machineController"`
	Etcd              bool `json:"etcd"`
}
```

This issue helps to define the new metrics we would like to return to the users.

zreigz:
**What this PR does / why we need it**: Improve cluster's health endpoint

Fixes #1485

```release-note
Improve the cluster's health endpoint. Returns statuses: UP, DOWN and PROVISIONING
```


api/pkg/resources/health.go
```diff
@@ -3,34 +3,42 @@ package resources
...
 )
 
 // HealthyDeployment tells if the deployment has a minimum of minReady replicas in Ready status
-func HealthyDeployment(ctx context.Context, client client.Client, nn types.NamespacedName, minReady int32) (bool, error) {
+func HealthyDeployment(ctx context.Context, client client.Client, nn types.NamespacedName, minReady int32) (kubermaticv1.HealthStatus, error) {
 	deployment := &appsv1.Deployment{}
 	if err := client.Get(ctx, nn, deployment); err != nil {
 		if kerrors.IsNotFound(err) {
-			return false, nil
+			return kubermaticv1.DOWN, nil
```
alvaroaleman: This means it will in state `DOWN` during creation

api/pkg/resources/health.go
```diff
@@ -3,34 +3,42 @@ package resources
...
 		}
-		return false, err
+		return kubermaticv1.DOWN, err
 	}
 
-	return deployment.Status.ReadyReplicas >= minReady, nil
+	if deployment.Status.ReadyReplicas >= minReady {
+		return kubermaticv1.UP, nil
+	}
+	return kubermaticv1.PROVISIONING, nil
```
alvaroaleman: By default we run everything with one replica. This means we will never get into state PROVISIONING, becauase we are either DOWN or UP.

alvaroaleman (commented):
Thanks a lot for your work on this. I am not sure if this change will be useful for the default case, where user clusters have one replica for every thing.

@mrIncompetent and me had (for a different problem) the idea of extending the cluster CR with conditions, something along the lines of:

```
type condition struct {
    type conditionType
    message string
    conditionStatus bool
}
```

This could be better suited for the use case at hand, the UI could basically show all conditions. This also has the advantage of being extensible in the future, should we end up with more possible states, than up, down and provisioning. Maybe we should do a quick sync?

mrIncompetent:
I would favor a solution which uses conditions so we are more flexible, but i would be fine to merge this change so the UI can start integrating it & we don't block progressing.

api/pkg/resources/health.go
```diff
@@ -3,34 +3,42 @@ package resources
...
 		if kerrors.IsNotFound(err) {
-			return false, nil
+			return kubermaticv1.DOWN, nil
 		}
-		return false, err
+		return kubermaticv1.DOWN, err
 	}
 
-	return deployment.Status.ReadyReplicas >= minReady, nil
+	if deployment.Status.ReadyReplicas >= minReady {
```
alvaroaleman: Maybe change the whole thing to:

* Return down when `deployment.Status.ReadyReplicas < minReady`
* Return progressing when `deployment.Status.UpdatedReplicas != deployment.Spec.Replicas || deployment.Status.Replicas != deployment.Spec.Replicas`
* Return up when `deployment.Spec.Replicas == deployment.Status.Replicas == Deployment.Status.UpdatedReplicas`

This has the advantage, that it shows the thing as progressing during an upgrade

zreigz: yes sure, thanks

zreigz: @alvaroaleman I've made small modification because with this condition `deployment.Status.ReadyReplicas < minReady` the `PROVISIONING` status was never displayed
PTAL

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -381,6 +384,26 @@ type ClusterHealthStatus struct {
 	UserClusterControllerManager bool `json:"userClusterControllerManager"`
 }
 
+type HealthStatus int
+
+const (
+	DOWN         HealthStatus = iota
```
alvaroaleman: Can we rename those to `HealthStatusDown/Up/Provisioning`?

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -187,6 +187,9 @@ type ClusterStatus struct {
 	Phase ClusterPhase `json:"phase,omitempty"`
 	// Health exposes information about the current health state of the individual control plane components
 	Health ClusterHealth `json:"health,omitempty"`
```
alvaroaleman: Do we actually have to keep the currrent HealthStatus?

api/pkg/resources/health.go
```diff
@@ -3,34 +3,56 @@ package resources
...
-			return false, nil
+			return kubermaticv1.DOWN, nil
 		}
-		return false, err
+		return kubermaticv1.DOWN, err
 	}
 
-	return deployment.Status.ReadyReplicas >= minReady, nil
+	// component creation
+	if deployment.Status.UnavailableReplicas >= minReady {
```
alvaroaleman: In reality this can also mean that the node one of the replicas ran on went down. We currently have no way to distinguish "in creating" from "something broke"

api/pkg/crd/kubermatic/v1/cluster.go
zreigz: Yes I can change it

api/pkg/crd/kubermatic/v1/cluster.go
zreigz: We need this for the backward compatibility. Old clusters still have this `Health` structure. When I changed `ClusterHealth` from bool to new type `HealthStatus` the informers and listeners got crazy.
To be honest I am not sure what will happen when I remove `Health`. I will check if the controller will handle this

api/pkg/resources/health.go
zreigz: yes, it's not perfect but we would like to be able to display some intermediate state. The created deployment is always some step forward. In the futer, we can add some Error state

api/pkg/crd/kubermatic/v1/cluster.go
alvaroaleman: Okay yeah, the compatibility concern makes sense, lets then check if we still need the `Health` field

zreigz: I've checked both directions. From old code to new and vice versa. The controller can handle this situation and replaces cluster structure `Health -> ExtendedHealth` or `ExtendedHealth -> Health` and updates components status. For end user everything is transparent

zreigz:
@alvaroaleman PTAL

api/pkg/controller/addon/addon_controller.go
```diff
@@ -199,7 +199,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, addo
 	}
 
 	// When the apiserver is not healthy, we must skip it
-	if !cluster.Status.Health.Apiserver {
+	if kubermaticv1.HealthStatusUp != cluster.Status.ExtendedHealth.Apiserver {
```
alvaroaleman: I think the easiest is to return when `HealthStatusDown` because during provisioning, the apiserver is also available

api/pkg/resources/health.go
alvaroaleman: I am still not certain if this makes sense. The current definition of "provisioning" is the very same as unavailable. Maybe if the UI folks want that, they can just show "provisioning" as long as its unavailable?

I would really prefer to not have this in the API code because its just confusing.

zreigz: cc @floreks @maciaszczykm @kgroschoff WDYT?
when I change it you will never get `PROVISIONING` state during cluster creation only for cluster update.

alvaroaleman: We could probably also have some hackery in the frontend that shows it as "provisioning" for the first N (10?) minutes after creation if its unavailable.

And if this is only about the initial provisioning, then we don't even have to change anything in the API for that.

api/pkg/controller/addon/addon_controller.go
zreigz: done

api/pkg/resources/health.go
floreks: UI should not provide such hacky solutions. It should rely on the API to provide correct data at all times. In case someone else tries to use the API, he will encounter the same issues anyway. As the API is the single point of entry for all clients, it has to provide correct information.

alvaroaleman: > UI should not provide such hacky solutions. It should rely on the API to provide correct data at all times

I completely agree. The problem is this Pull doesn't do that. This pull makes the api return "provisioning" when the cluster is unhealthy. This happens to be correct during the initial creation but not afterwards.

The underlying problem of differentiating those is a complicated one also one that most likely the lifecycle team has to solve. We may end up doing that within https://github.com/kubermatic/kubermatic/issues/3748 for a different reason but right now its not solved and its not easy to do that.

My point was that _if_ we do hackery because we currently can not tell "provisioning" or not it should happen in the frontend because the fields the API uses are actually used by controllers which means that if we do hackery in the API we will also have to extend that hackery onto the controllers and that is something I am opposed to.

The alternate as mentioned is to wait for #3748

zreigz: Let's find a compromise here. The meaning of this PR is to change health status to have a better understanding of what is happening with the cluster. We could merge this with @alvaroaleman suggestion and wait for #3748 and then make the necessary upgrades.

zreigz: done

floreks: It's still an improvement compared to what we have now. The API logic can be eventually adjusted to better reflect the actual status of the cluster. Right now we can only display failed/ready which is imho even more confusing.

api/pkg/controller/addoninstaller/addoninstaller_controller.go
```diff
@@ -137,7 +137,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 
 	// Wait until the Apiserver is running to ensure the namespace exists at least.
 	// Just checking for cluster.status.namespaceName is not enough as it gets set before the namespace exists
-	if !cluster.Status.Health.Apiserver {
+	if kubermaticv1.HealthStatusUp != cluster.Status.ExtendedHealth.Apiserver {
```
alvaroaleman: This needs to be changed to ` kubermaticv1.HealthStatusDown == cluster.Status.ExtendedHealth.Apiserver`

api/pkg/controller/backup/backup_controller.go
```diff
@@ -249,7 +249,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 	}
 
 	// Wait until we have a running etcd
-	if !cluster.Status.Health.Etcd {
+	if kubermaticv1.HealthStatusUp != cluster.Status.ExtendedHealth.Etcd {
```
alvaroaleman: This also needs to be changed to check for down instead

api/pkg/controller/monitoring/monitoring_controller.go
```diff
@@ -190,7 +190,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 
 	// Wait until the UCCM is ready - otherwise we deploy with missing RBAC resources
-	if !cluster.Status.Health.UserClusterControllerManager {
+	if kubermaticv1.HealthStatusUp != cluster.Status.ExtendedHealth.UserClusterControllerManager {
```
alvaroaleman: Need to check for down instead

api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -272,7 +272,7 @@ func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.Cluste
 		return nil, fmt.Errorf("failed to sync health: %v", err)
 	}
 
-	if !osData.Cluster().Status.Health.Apiserver {
+	if kubermaticv1.HealthStatusUp != osData.Cluster().Status.ExtendedHealth.Apiserver {
```
alvaroaleman: Need to check for down instead

alvaroaleman (commented):
Some oversights regarding the check for down instead of up, other than that LGTM :+1:

api/pkg/controller/addoninstaller/addoninstaller_controller.go
zreigz: done

api/pkg/controller/openshift/openshift_controller.go
zreigz: done

api/pkg/controller/monitoring/monitoring_controller.go
zreigz: done

api/pkg/controller/backup/backup_controller.go
zreigz: done

PR #3794, opened 2019-07-11
Use yaml.UnmarshalStrict everywhere

alvaroaleman:

This PR makes us use `yaml.UnmarshalStrict` everywhere. This has the advantage that unknown fields cause an error instead of being silently dropped.

Fixes #

**Documentation**:

```release-note
[ACTION REQUIRED] Kubermatic now doesn't accept unknown keys in its config files anymore and will crash if an unknown key is present
[ACTION REQUIRED] BYO datacenters now need to be specific in the `datacenters.yaml` with a value of `{}`, e.G `bringyourown: {}`
```

/assign @zreigz 
/assign @nikhita

### DIFF
--- a/api/pkg/presets/manager.go
+++ b/api/pkg/presets/manager.go
@@ -6,7 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/ghodss/yaml"
+	"sigs.k8s.io/yaml"
 
 	"github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
@@ -163,7 +163,7 @@ func loadPresets(path string) (*Presets, error) {
 		Presets *Presets `json:"presets"`
 	}{}
 
-	err = yaml.Unmarshal(bytes, &s)
+	err = yaml.UnmarshalStrict(bytes, &s)
 	if err != nil {
 		return nil, err
 	}
--- a/api/pkg/provider/datacenter.go
+++ b/api/pkg/provider/datacenter.go
@@ -42,7 +42,7 @@ func LoadDatacentersMeta(path string) (map[string]DatacenterMeta, error) {
 	}
 
 	dcs := datacentersMeta{}
-	if err := yaml.Unmarshal(bytes, &dcs); err != nil {
+	if err := yaml.UnmarshalStrict(bytes, &dcs); err != nil {
 		return nil, err
 	}
 
--- a/api/pkg/provider/datacenter_test.go
+++ b/api/pkg/provider/datacenter_test.go
@@ -22,13 +22,9 @@ datacenters:
   europe-west3-c: #Master
     location: Frankfurt
     country: DE
-    provider: Loodse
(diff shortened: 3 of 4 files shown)


nikhita:
@alvaroaleman could this lead to compatibility issues with older versions if they persisted unknown fields?

alvaroaleman:
I _think_ we never removed fields so far, we only added stuff. Can you config @mrIncompetent ?

mrIncompetent:
@alvaroaleman 

> Can you config @mrIncompetent ?

How do you wan't me to be configured?

mrIncompetent:
We replaced `AMI` with `ImageList` within the AWS spec.
I assume there might be old files out there which might still contain the `AMI` field.

So we might want to think about something to ease the operators(The human, not the controller) life in case such an error gets triggered.
Maybe print an error linking to a docs page in case the unmarshal fails?

alvaroaleman:
> Maybe print an error linking to a docs page in case the unmarshal fails?

Not sure, I mean the error will contain a "field bla was not found", IMHO that is pretty self-explanatory.

IMHO compatibility is mainly an issue if we would want to allow ppl to use the same config files for multiple versions of Kubermatic, but that is not the case. The config files are supposed to be compatible with exactly one version of Kubermatic. And IMHO the burden of forcing everyone to clean up is not a bad thing at all, since it keeps everything tidy

*Goes to the secrets repo and removes `unknown field "provider"`*

zreigz:
the API can not start:
```
{"level":"fatal","time":"2019-07-11T15:01:29.330+0200","caller":"kubermatic-api/main.go:78","msg":"failed to create and initialize providers","error":"failed to load datacenter yaml \"/home/lukasz/go/src/github.com/kubermatic/secrets/seed-clusters/dev.kubermatic.io/datacenters.yaml\": error unmarshaling JSON: while decoding JSON: json: unknown field \"provider\""}
```

alvaroaleman:
Yes, I am fixing up our `datacenters.yaml`. Note even our unittests pass with this :grimacing:

mrIncompetent:
A nice customer experience would be if the customer knows ahead of time if the `datacenters.yaml` must be checked.
So far we never had a release note stating that the `datacenters.yaml` must be modified or kubermatic throws errors. Now we require changes on that file and those should be documented in upgrade docs + a release note.

alvaroaleman:
> A nice customer experience would be if the customer knows ahead of time if the datacenters.yaml must be checked.

Yes, I absolutely agree. Will add an [Action Needed] release note

zreigz:
there is another error:
```
{"level":"fatal","time":"2019-07-11T16:11:26.242+0200","caller":"kubermatic-api/main.go:78","msg":"failed to create and initialize providers","error":"failed to load datacenter yaml \"/home/lukasz/go/src/github.com/kubermatic/secrets/seed-clusters/dev.kubermatic.io/datacenters.yaml\": error unmarshaling JSON: while decoding JSON: json: unknown field \"region\""}
```

alvaroaleman:
Tests passed, ptal @mrIncompetent @zreigz

PR #3808, opened 2019-07-12
validate forbidden error for getting cluster

### LINKED ISSUE #3807: Validate the `forbidden` error during cluster creation
**User Story**
The method for getting a cluster uses the impersonated client. The UI can ask for the cluster before RBAC are set. In this case, get method returns `forbidden` error
**Acceptance criteria**
Use the privileged method when `forbidden` occurs and validate if the cluster is not ready yet or doesn't exist

zreigz:
**What this PR does / why we need it**: validate forbidden error for getting cluster

Fixes #3807

```release-note
NONE
```

### DIFF
--- a/api/pkg/handler/v1/cluster/cluster.go
+++ b/api/pkg/handler/v1/cluster/cluster.go
@@ -248,12 +248,8 @@ func GetEndpoint(projectProvider provider.ProjectProvider) endpoint.Endpoint {
 }
 
 func isStatus(err error, status int32) bool {
-	if kubernetesError, ok := err.(*kerrors.StatusError); ok {
-		if status == kubernetesError.Status().Code {
-			return true
-		}
-	}
-	return false
+	kubernetesError, ok := err.(*kerrors.StatusError)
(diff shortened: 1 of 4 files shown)


api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -221,20 +221,41 @@ func GetEndpoint(projectProvider provider.ProjectProvider) endpoint.Endpoint {
...
+			// Cluster is not ready yet, return original error
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
 		return convertInternalClusterToExternal(cluster), nil
 	}
 }
 
+func isStatus(err error, status int32) bool {
+	if kubernetesError, ok := err.(*kerrors.StatusError); ok {
```
maciaszczykm: I think this can do the same:

```go
kubernetesError, ok := err.(*kerrors.StatusError)
return ok && status == kubernetesError.Status().Code
```

zreigz: yes, thanks

PR #3816, opened 2019-07-15
Re-Build the nodeport proxy

alvaroaleman:
The current version of the envoy-manager is outputting **a lot** of

```
time="2019-07-15T11:28:13Z" level=error annotation=nodeport-proxy.k8s.io/expose
```

log lines for no apparent reason. To make sure the binary inside the Docker image is fresh and actually based on the code in the repo, this PR deploys a new, pristine version.

```release-note
none
```

/assign @xrstf

### DIFF
--- a/config/nodeport-proxy/Chart.yaml
+++ b/config/nodeport-proxy/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nodeport-proxy
-version: 1.2.0
+version: 1.2.1
 appVersion: v2.2.1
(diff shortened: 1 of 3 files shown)


config/nodeport-proxy/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: nodeport-proxy
 version: 1.2.0
```
xrstf: Please bump as well.

Issue #3740, opened 2019-07-04
Allow controllers to use CRs as basis for datacenter handling instead of a file


alvaroaleman:
**User Story**

Currently, we use a static file for datacenters, instead we want to use a custom resource.

**Acceptance criteria**

* All our controllers are able to use a CR as a source for info about datacenters
* The existing behaviour of using a file continues to exist
* If CRs or a file is used is controlled by a flag
* It is not possible to use both a file and a CR

alvaroaleman:
Done

PR #3824, opened 2019-07-15
Migrate code to use CRD data structures for DCs

alvaroaleman:

Moves towards fixing #3740

Moves our code over to use the new types for datacenters and explicitly distringuish between seeds and Datacenters
**Documentation**:

```release-note
none
```


alvaroaleman:
@zreigz @mrIncompetent Modulo unittests for the new code in `pkg/provider/conversions.go` this is done, can you have a look if this looks fine approachwise?

Semantically, this introduces two changes:
* It is not possible anymore to define one datacenter as both seedDC and nodeDC. This was done in one unittests, I hope no customers has this. IMHO this is not a valid usecase anyways, as those two are different things
* `ListDatacenters` doesn't filter out duplicate DCs. The only guarantee the api can give is that there will not be more than one seedDC with a given name and not more than one nodeDC per seed dc with a given name.

In practise multiple dcs with the same name can not get into the system currently, because we read everything from a file that is a flat map. Once we move to CRDs, this could happen. IMHO we should have validation that disallows setting up such a config.

zreigz:
I will take a look and test it today

zreigz:
Works and looks perfectly fine. I appreciate your huge effort :+1: 
Is it still WIP or can I approve it?

alvaroaleman:
> Is it still WIP or can I approve it?

Unittests are missing, I will ping you once they are done

alvaroaleman:
@zreigz Added unittests, PR is now done from my POV

api/cmd/master-controller-manager/main.go
```diff
@@ -102,9 +102,9 @@ func main() {
 	ctrlCtx.kubeMasterInformerFactory = kuberinformers.NewSharedInformerFactory(ctrlCtx.kubeMasterClient, informer.DefaultInformerResyncPeriod)
 	ctrlCtx.labelSelectorFunc = selector
 
-	ctrlCtx.datacenters, err = provider.LoadDatacentersMeta(ctrlCtx.runOptions.dcFile)
+	ctrlCtx.datacenters, err = provider.LoadDatacenters(ctrlCtx.runOptions.dcFile)
```
mrIncompetent: If we load the DC's during startup we won't be able to manage the datacenters during runtime - which is one of the goals of this whole migration.

alvaroaleman: Yes, this will be changed in a follow-up. This PR already changes 2.2k LOC, I didn't want to make it even bigger

api/cmd/master-controller-manager/main.go
```diff
@@ -102,9 +102,9 @@ func main() {
...
 	if err != nil {
-		sugarLog.Fatalw("Failed to read the datacenters definition", "error", err)
+		sugarLog.Fatalw("Failed toparse the datacenters definition", "error", err)
```
mrIncompetent: typo

alvaroaleman:
@mrIncompetent Typo fixed, PTAL

api/pkg/provider/conversions.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, seed := range seeds {
+		nodeLocation, exists := seed.Spec.NodeLocations[nodeLocationName]
+		if !exists {
+			continue
+		}
+
+		results = append(results, nodeLocation)
+	}
+
+	if n := len(results); n != 1 {
```
mrIncompetent, line 93:
```go
if len(results) != 1 {
```

api/cmd/conformance-tests/runner.go
```diff
@@ -280,9 +280,9 @@ func (r *testRunner) executeScenario(log *logrus.Entry, scenario testScenario) (
 		"version":        cluster.Spec.Version,
 	})
 
-	dc, found := r.dcs[cluster.Spec.Cloud.DatacenterName]
-	if !found {
-		return nil, fmt.Errorf("invalid cloud datacenter specified '%s'. Not found in datacenters list", cluster.Spec.Cloud.DatacenterName)
+	nodeDC, err := provider.NodeLocationFromSeedMap(r.dcs, cluster.Spec.Cloud.DatacenterName)
```
mrIncompetent: Why does the function name mention `NodeLocation` but the variable is called `nodeDC` ?
We  should have a consistent name

api/cmd/kubermatic-cli/README.md
```diff
@@ -1,112 +0,0 @@
-# Automate creation of kubermatic managed clusters
```
mrIncompetent, line 1: This was developed upon customer request & the intention of this PR has nothing todo with this cmd.
Revert please

mrIncompetent (changes requested):

mrIncompetent:
IMO
From a code quality standpoint this PR is a step backwards.
It mainly comes from the data model we use(Embedding the `NodeLocations` in the `SeedDatacenter`).
We now require every developer to know to only get the `NodeLocation` using `NodeLocationFromSeedMap`.
On top we now require components to know about `SeedDatacenters`, while the only data they need is the `NodeLocations`([LoD](https://en.wikipedia.org/wiki/Law_of_Demeter) is a good read here).

This is not meant to block this PR as this is just my personal opinion on the code quality & not the functionality itself.

api/cmd/kubermatic-cli/README.md
alvaroaleman: This is unused and doesn't use our API. It will get replaced with a CLI that actually use our API.

I am not going to fix up something we don't use and shouldn't have developed in the first place.

mrIncompetent: This code got used by customers. If you want to delete it, make sure they don't use it anymore.
@kron4eg Can you check if this still gets used?

alvaroaleman:
>On top we now require components to know about SeedDatacenters, while the only data they need is the NodeLocations(LoD is a good read here).

Yup, this is mainly due to the fact how the code is currently structured. We instantiate the cloud providers once with all data they could ever need. When making the thing actually dynamic, I will change their instantiation to only happen when they are needed and to only get the data they need for that specific invocation.

WRT the controllers, I could change them to have a `map[ string]NodeLocation` instead of the one element `map[string]SeedDatacenter`?

api/pkg/provider/conversions.go
alvaroaleman: That variable is being used in the error, hence the assignment

api/cmd/kubermatic-cli/README.md
kron4eg: Yeah, I think it can be removed. Actual usage never happened.

api/cmd/conformance-tests/main.go
```diff
@@ -247,6 +247,10 @@ func main() {
 	}
 	opts.kubermaticClient = apiclient.New(httptransport.New(kubermaticAPIServerAddress, "", []string{"http"}), nil)
 	opts.kubermaticAuthenticator = httptransport.BearerToken(kubermaticServiceaAccountToken)
+	seedName := os.Getenv("SEED_NAME")
```
mrIncompetent, line 254: If there is a reason why we use an environment variable instead of a flag, a comment should explain it :)

api/cmd/conformance-tests/runner.go
```diff
@@ -280,9 +281,9 @@ func (r *testRunner) executeScenario(log *logrus.Entry, scenario testScenario) (
 		"version":        cluster.Spec.Version,
 	})
 
-	dc, found := r.dcs[cluster.Spec.Cloud.DatacenterName]
-	if !found {
-		return nil, fmt.Errorf("invalid cloud datacenter specified '%s'. Not found in datacenters list", cluster.Spec.Cloud.DatacenterName)
+	datacenter, exists := r.seed.Spec.Datacenters[cluster.Spec.Cloud.DatacenterName]
+	if !exists {
+		return nil, fmt.Errorf("NodeLocation %q doesn't exist", cluster.Spec.Cloud.DatacenterName)
```
mrIncompetent: `NodeLocation` -> `Datacenter`

mrIncompetent (approved):

PR #3833, opened 2019-07-16
Prometheus alert inhibitions

xrstf:
This PR starts to make use of the "Alert Inhibitions" for the Prometheus Alertmanager. This allows us to not send alerts for consequential alerts, e.g. we do not want to alert about broken kube-state-metrics when at the same time the apiserver pod is already dead (causing kube-state-metrics to explode).

For now, most of the inhibitions are centered around user clusters, but I suspect we will expand this in the future. The workflow for figuring out the inhibition rules was as follows:

1. Nuke/remove/break a component inside the user cluster namespace.
2. Wait a few minutes.
3. Observe the alerts and inhibit all alerts that follow the alert for the component you nuked.
4. Goto 1.

So why is this PR so large? It needs to adjust a few things to make inhibitions easy/easier to define and add new alerts that we can use to inhibit others.

* The DNS resolver is now being scraped by Prometheus. A new alert for it has been defined (`DNSResolverDown`) to inhibit the other alerts.
* New alerts for missing cAdvisors, unscraped kubelets and usercluster-controllers have been defined.
* `KubeStateMetricsDown` has been reduced from critical to warning, as the user cluster itself should not be affected immediately.
* The *seed* Prometheus scraping rules have been extended to put a `cluster` label on each metric originating from a user-cluster namespace pod. This makes it much, much easier to define inhibitions.

```release-note
Alertmanager's inhibition feature is now used to hide consequential alerts.
```

### DIFF
--- a/api/pkg/resources/dns/dns.go
+++ b/api/pkg/resources/dns/dns.go
@@ -82,17 +82,18 @@ func DeploymentCreator(data deploymentCreatorData) reconciling.NamedDeploymentCr
 			volumes := getVolumes()
 			podLabels, err := data.GetPodTemplateLabels(resources.DNSResolverDeploymentName, volumes, nil)
 			if err != nil {
-				return nil, fmt.Errorf("failed to get podlabels: %v", err)
+				return nil, fmt.Errorf("failed to get pod labels: %v", err)
 			}
 
-			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
-				Labels: podLabels,
-				Annotations: map[string]string{
-					"prometheus.io/scrape": "true",
-					"prometheus.io/path":   "/metrics",
-					"prometheus.io/port":   "9253",
-				},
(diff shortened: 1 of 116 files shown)


kron4eg (approved):
/approve

api/pkg/resources/dns/dns.go
```diff
@@ -85,7 +85,15 @@ func DeploymentCreator(data deploymentCreatorData) reconciling.NamedDeploymentCr
 				return nil, fmt.Errorf("failed to get podlabels: %v", err)
 			}
 
-			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{Labels: podLabels}
+			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
+				Labels: podLabels,
+				Annotations: map[string]string{
```
alvaroaleman: Not introduced by you, but can we change this to not overwrite other annotations?

xrstf: We can.

PR #3836, opened 2019-07-17
Use secrets for credentials for AWS

### LINKED ISSUE #3792: Add secretKeySelector fields for all cloud providers
**User Story**:

When a cluster is created, the relevant fields holding the secret values should be cleared and instead actual secret objects should be created containing these values and the secretKeySelector for these secrets should be added to the cluster object. This should be done for all cloud providers.

**Acceptance criteria**:

- [ ] All cloud providers have relevant secretKeySelector fields. (packet is being done in https://github.com/kubermatic/kubermatic/pull/3684)

nikhita:
Fixes https://github.com/kubermatic/kubermatic/issues/3792

https://github.com/kubermatic/kubermatic/pull/3684 adds support for specifying credentials using secrets and adds support for Packet provider. This PR builds on top of https://github.com/kubermatic/kubermatic/pull/3684 and adds support for rest of the providers.

/hold
until https://github.com/kubermatic/kubermatic/pull/3684 is merged

```release-note
NONE
```


nikhita:
For vsphere, `username` and `password` are mentioned in two places:

https://github.com/kubermatic/kubermatic/blob/d89eb685b62d56f30101d02849be19d908b2c06f/api/pkg/crd/kubermatic/v1/cluster.go#L271-L286

Looking at the code, it looks like the ones in `VSphereCredentials` should be used for the secret? What does the other `Username` and `Password` do? :thinking:

api/cmd/conformance-tests/runner.go
```diff
@@ -552,7 +554,12 @@ func (r *testRunner) setupNodes(parentLog *logrus.Entry,
 		// is unset but need a deterministic name because we retry creation and dont
 		// want to accidentally create multiple MachineDeployments
 		nd.Name = fmt.Sprintf("md-%s-%d", scenarioName, ndIndex)
-		machineDeployment, err := machine.Deployment(cluster, &nd, dc.DeepCopy(), keys)
+
+		data := test.CredentialsData{
```
alvaroaleman: Ugh. Can you create a ticket to do machine creation via kubermatic API? That will improve the test coverage and remove the need for this

api/cmd/projects-migrator/credentials.go
```diff
@@ -15,12 +15,121 @@ import (
...
+	}
 	if cluster.Spec.Cloud.Packet != nil {
 		return createPacketSecret(cluster, client)
 	}
 	return nil
 }
 
+func createAWSSecret(cluster *kubermaticv1.Cluster, client kubernetes.Interface) error {
+	spec := cluster.Spec.Cloud.AWS
+	referenceDefined := spec.AccessKeyIDReference != nil && spec.SecretAccessKeyReference != nil
```
alvaroaleman: Err, how do we want to handle the case of both `referenceDefined` and `valuesDefine`? 
Also I cant seem to find the place that unsets the `value` on the Cluster CR?

api/pkg/provider/cloud/aws/provider.go
```diff
@@ -75,15 +75,15 @@ func (a *AmazonEC2) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 }
 
 // MigrateToMultiAZ migrates an AWS cluster from the old AZ-hardcoded spec to multi-AZ spec
-func (a *AmazonEC2) MigrateToMultiAZ(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) error {
+func (a *AmazonEC2) MigrateToMultiAZ(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater, secretKeySelector provider.SecretKeySelectorValueFunc) error {
```
alvaroaleman: Would it make sense to pass the `secretKeySelector` in when initializating the cloudprovider instead of dinstinctly on each func?

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -24,25 +24,199 @@ func NewCredentialsProvider(kubernetesImpersonationClient kubernetesImpersonatio
...
+
+	// remove credentials from cluster object
+	cluster.Spec.Cloud.Digitalocean.Token = ""
+
+	return createdSecret, nil
+}
+
+func (p *CredentialsProvider) createGCPSecret(cluster *kubermaticv1.Cluster, projectID string) (*corev1.Secret, error) {
+	// create secret for storing credentials
+	name := fmt.Sprintf("%s-gcp-%s", CredentialPrefix, cluster.Name)
```
alvaroaleman: Random thought: Can we maybe extend the `Cluster` object with a `GetSecretName`? This is needed at multiple places and should always yield the same result. Or do kube api conventions tell you to not add custom funcs to api types?

api/pkg/provider/types.go
```diff
@@ -40,8 +41,8 @@ const (
 
 // CloudProvider declares a set of methods for interacting with a cloud provider
 type CloudProvider interface {
-	InitializeCloudProvider(*kubermaticv1.Cluster, ClusterUpdater) (*kubermaticv1.Cluster, error)
-	CleanUpCloudProvider(*kubermaticv1.Cluster, ClusterUpdater) (*kubermaticv1.Cluster, error)
+	InitializeCloudProvider(*kubermaticv1.Cluster, ClusterUpdater, SecretKeySelectorValueFunc) (*kubermaticv1.Cluster, error)
```
alvaroaleman, line 44: Yeah as mentioned above, what about just passing both `ClusterUpdater` and `SecretKeySelector` in during initialization and storing it there?

api/pkg/resources/credentials.go
```diff
@@ -0,0 +1,170 @@
...
+	APIKey    string
+	ProjectID string
+}
+
+type CredentialsData interface {
+	Cluster() *kubermaticv1.Cluster
+	GetGlobalSecretKeySelectorValue(configVar *providerconfig.GlobalSecretKeySelector) (string, error)
+}
+
+func GetCredentials(data CredentialsData) (Credentials, error) {
```
alvaroaleman, line 28: Is there a reason to not use this everywhere?

api/cmd/conformance-tests/runner.go
nikhita: > Can you create a ticket to do machine creation via kubermatic API?

https://github.com/kubermatic/kubermatic/issues/3887

api/cmd/projects-migrator/credentials.go
nikhita: > how do we want to handle the case of both referenceDefined and valuesDefine

Updated to conditionally set/unset. PTAL.

api/pkg/resources/credentials.go
nikhita: > Is there a reason to not use this everywhere?

It's used for in controller-manager, machine-controller, etc deployments. Where else did you mean that this should be used?

nikhita:
Updated to include all cloud providers and tried to separate into logical commits, but the diff is a little huge :see_no_evil:

api/cmd/projects-migrator/credentials.go
```diff
@@ -38,10 +371,101 @@ func createPacketSecret(cluster *kubermaticv1.Cluster, client kubernetes.Interfa
...
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cluster.GetSecretName(),
+				Namespace: resources.KubermaticNamespace,
+			},
+			Type: corev1.SecretTypeOpaque,
+			Data: map[string][]byte{
+				"username":                    []byte(spec.Username),
+				"password":                    []byte(spec.Password),
+				"infraManagementUserUsername": []byte(spec.InfraManagementUser.Username),
```
alvaroaleman: I think this should be kept in the datacenter

api/cmd/projects-migrator/credentials.go
```diff
@@ -38,10 +371,101 @@ func createPacketSecret(cluster *kubermaticv1.Cluster, client kubernetes.Interfa
...
+		cluster.Spec.Cloud.VSphere.InfraManagementUser.PasswordReference = &providerconfig.GlobalSecretKeySelector{
+			ObjectReference: corev1.ObjectReference{
+				Name:      secret.Name,
+				Namespace: secret.Namespace,
+			},
+			Key: "infraManagementUserPassword",
+		}
 	}
+
+	if valuesDefined {
```
alvaroaleman: Maybe I am missing it but when do we update the cluster after changing its field here (and in all other `create$PROVIDERsecret` calls)?

api/cmd/projects-migrator/credentials.go
```diff
@@ -1,34 +1,367 @@
...
+		cluster.Spec.Cloud.Hetzner.Token = ""
+	}
+
+	return nil
+}
+
+func createOpenstackSecret(cluster *kubermaticv1.Cluster, client kubernetes.Interface) error {
+	spec := cluster.Spec.Cloud.Openstack
+	referenceDefined := spec.UsernameReference != nil && spec.PasswordReference != nil && spec.DomainReference != nil
```
alvaroaleman: Probably a follow-up but wouldn't it easier to just have a `credentialReference` or something on all providers that have more than one property? We can not and do not want to handle the case that there are different references defined for different properties, so I don't really see why we would create an API that allows that

api/pkg/resources/credentials.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+type VSphereCredentials struct {
+	Username                    string
+	Password                    string
+	InfraManagementUserUsername string
+	InfraManagementUserPassword string
+}
+
+type CredentialsData interface {
```
alvaroaleman, line 23: Why do we need to define this interface? There is only one implementation or did I miss it?

api/cmd/projects-migrator/credentials.go
nikhita: Here https://github.com/kubermatic/kubermatic/blob/secrets-handling-all/api/cmd/projects-migrator/main.go#L673-L678

api/pkg/resources/credentials.go
nikhita: It's used in `api/pkg/resources/machine/machinedeployment.go`, `api/pkg/resources/apiserver/deployment.go`, `api/pkg/controller/openshift/resources/apiserver_deployment.go` and other deployments

nikhita:
Looks like something broke after separating out providers... taking a look. :/

I have the changes for all other providers on top of this branch but I'll create PRs for those after this one is green or else we'll have lots of clusters in the CI to clean up :see_no_evil:

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -303,8 +306,11 @@ type BringYourOwnCloudSpec struct{}
 
 // AWSCloudSpec specifies access data to Amazon Web Services.
 type AWSCloudSpec struct {
-	AccessKeyID         string `json:"accessKeyId"`
-	SecretAccessKey     string `json:"secretAccessKey"`
+	AccessKeyIDReference     *providerconfig.GlobalSecretKeySelector `json:"accessKeyIdReference,omitempty"`
```
alvaroaleman: can we maybe instead do a `credentialsReference` field instead of the two current fields? We don't want to support the case of having different references so we shouldn't make that part of the API.

It means that we have to hardcode the fieldNames but I think thats okay

api/pkg/provider/cloud/azure/provider.go
```diff
@@ -512,7 +513,8 @@ func ensureRouteTable(ctx context.Context, cloud kubermaticv1.CloudSpec, locatio
 	return nil
 }
 
-func (a *Azure) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
+func (a *Azure) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater, secretKeySelector provider.SecretKeySelectorValueFunc) (*kubermaticv1.Cluster, error) {
```
alvaroaleman, line 516: Can we pass in the `secretKeySelector` when we initialize the cloudprovider in `pkg/provider/cloud/provider.go`?

nikhita: Tried that way, but we don't have the client here: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/provider/aws.go#L134 :/

alvaroaleman: :(

nikhita:
This is ready for another review

api/pkg/controller/cloud/cloud_controller.go
```diff
@@ -206,3 +207,19 @@ func (r *Reconciler) updateCluster(name string, modify func(*kubermaticv1.Cluste
...
+		if err := r.Get(context.Background(), namespacedName, secret); err != nil {
+			return "", fmt.Errorf("error retrieving secret %q from namespace %q: %v", configVar.Name, configVar.Namespace, err)
+		}
+
+		if val, ok := secret.Data[key]; ok {
+			return string(val), nil
+		}
+		return "", fmt.Errorf("secret %q in namespace %q has no key %q", configVar.Name, configVar.Namespace, key)
+	}
+	return "", nil
```
alvaroaleman: Nit: Calling this function doesn't make sense, so instead of only getting something if name/namespace/key is defined, can we check in the beginning if one of the three is empty and if yes, return an err?

alvaroaleman (commented):
one last nit

api/pkg/controller/cloud/cloud_controller.go
nikhita: Done.

nikhita:
/assign @zreigz
for approval

alvaroaleman:
/hold

This mus be clarified before we merge this: https://github.com/kubermatic/kubermatic/pull/3967#discussion_r310988253

PR #3840, opened 2019-07-17
Use a NLB for the NodePortProxy when running on AWS

mrIncompetent:
This PR will change the LB type for the NodePortProxy from a classic ELB to a NLB.
The NLB has 2 advantages:
- It has a static IP (We currently need logic to handle the changing IP's of a classic ELB)
- I is supported by VPC endpoints (We require this for a customer project)

```release-note
ACTION REQUIRED: On AWS, the nodeport-proxy will be recreated as NLB. DNS entries must be updated to point to the new LB.
```

/assign @xrstf


xrstf:
Please bump the chart version.

PR #3845, opened 2019-07-18
CRUD API for addons 

multi-io:

This adds endpoints to the API so users can C/R/U/D addons in their own clusters. We also added an option for passing a whitelist of the actually accessible addons to the apiserver, so you can prevent users from uninstalling canal or other vital addons.

Open issues:

- unit tests

- using runtime.RawExtension in the AddonSpec API type -- is this OK?

```release-note
added API endpoints for user to create/read/update/delete addons in the clusters
```

### DIFF
--- a/api/cmd/kubermatic-api/options.go
+++ b/api/cmd/kubermatic-api/options.go
@@ -3,12 +3,14 @@ package main
 import (
 	"flag"
 	"fmt"
+	"strings"
 
 	"github.com/kubermatic/kubermatic/api/pkg/features"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/provider/kubernetes"
 	"github.com/kubermatic/kubermatic/api/pkg/serviceaccount"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -29,6 +31,7 @@ type serverRunOptions struct {
 	exposeStrategy     corev1.ServiceType
 	dynamicDatacenters bool
 	log                kubermaticlog.Options
+	accessibleAddons   sets.String
 
 	// OIDC configuration
 	oidcURL                        string
@@ -51,6 +54,7 @@ func newServerRunOptions() (serverRunOptions, error) {
 	s := serverRunOptions{}
 	var rawFeatureGates string
 	var rawExposeStrategy string
+	var rawAccessibleAddons string
 
 	flag.StringVar(&s.listenAddress, "address", ":8080", "The address to listen on")
 	flag.StringVar(&s.kubeconfig, "kubeconfig", "", "Path to the kubeconfig.")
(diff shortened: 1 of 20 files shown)


api/cmd/kubermatic-api/options.go
```diff
@@ -90,6 +94,12 @@ func newServerRunOptions() (serverRunOptions, error) {
 	default:
 		return s, fmt.Errorf("--expose-strategy must be either `NodePort` or `LoadBalancer`, got %q", rawExposeStrategy)
 	}
+
+	s.accessibleAddons = map[string]bool{}
```
alvaroaleman: Nit: Using `k8s.io/apimachinery/pkg/util/sets.String` makes these things much more readable :)

config/kubermatic/values.yaml
```diff
@@ -128,6 +128,10 @@ kubermatic:
 
   api:
     replicas: 2
+    accessibleAddons:
+    - dashboard
```
maciaszczykm: I think we shouldn't list Dashboard at least until connect window show instructions how to connect to it. Disabling the addon will not make these intructions disappear and can confuse users.

Zrzut ekranu 2019-07-18 o 11 43 37

api/pkg/api/v1/types.go
```diff
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"encoding/json"
+	"k8s.io/apimachinery/pkg/runtime"
```
maciaszczykm: The new line is missing.

api/cmd/kubermatic-api/options.go
multi-io: You're right, didn't think of that. I'll change it.

multi-io:
I'll rebase, address the issues you guys brought up and add some tests too because we're actually using this in our fork currently only. Will report back early next week.

multi-io:
PR updated:
- unit tests for the crud operations
- map[string]bool replaced w/ string set
- use map[string]interface{} instead of RawExtension for the public-facing AddonSpec.Variables field
- empty default accessibleAddons list

maciaszczykm:
No further remarks from my side.

/approve

multi-io:
Rebased. Adapted to the dynamic-datacenters/CRDs stuff, should work but couldn't test it all since we don't currently have a stock Kubermatic installation (and our fork is still using the static datacenters.yaml/multi-seed kubeconfig). So basically if the providers initialization in main.go works, the CRUD apis themselves are covered by the unit tests.

alvaroaleman:
> Rebased. Adapted to the dynamic-datacenters/CRDs stuff, should work but couldn't test it all since we don't currently have a stock Kubermatic installation (and our fork is still using the static datacenters.yaml/multi-seed kubeconfig). So basically if the providers initialization in main.go works, the CRUD apis themselves are covered by the unit tests.

So far the old `datacenters.yaml` continues to work and will for some time, the feature is not yet ready. We just migrate it internally already to a new data structure that better represents the fact that seeds and datacenters are a complete distinct thing :)

/test all

api/pkg/controller/rbac/sync_resource.go
```diff
@@ -301,3 +310,130 @@ func shouldSkipRBACRoleBindingForNamedResource(projectName string, objectGVR sch
 	}
 	return false, generatedRole, nil
 }
+
+func (c *resourcesController) ensureRBACRoleForClusterAddons(projectName string, object metav1.Object, clusterProvider *ClusterProvider) error {
+	cluster, ok := object.(*kubermaticv1.Cluster)
+	if !ok {
+		panic("ensureRBACRoleBindingForClusterAddons called with non-cluster")
```
alvaroaleman: Err. No way.

api/pkg/controller/rbac/sync_resource.go
```diff
@@ -301,3 +310,130 @@ func shouldSkipRBACRoleBindingForNamedResource(projectName string, objectGVR sch
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *resourcesController) ensureRBACRoleBindingForClusterAddons(projectName string, object metav1.Object, clusterProvider *ClusterProvider) error {
+	cluster, ok := object.(*kubermaticv1.Cluster)
+	if !ok {
+		panic("ensureRBACRoleBindingForClusterAddons called with non-cluster")
```
alvaroaleman: same. Just return an error.

alvaroaleman: (the fact that this should never happen is absolutely not a reason to put a panic in production code)

multi-io: I'll change it. I'm pretty sure I just copied it from somewhere though.

multi-io:
Force-push  with just the two panics replaced w/ error returns.

api/pkg/controller/rbac/sync_resource.go
multi-io: https://github.com/kubermatic/kubermatic/blob/05b9ef7569ad1067678d165a4fe07c61f6760316/api/pkg/controller/rbac/informer_provider.go#L49 was the place I think.

alvaroaleman:
/test all
@zreigz can you lgtm if this looks good?

PR #3846, opened 2019-07-18
Define new KubermaticClusterPaused alert

xrstf:
This alert is triggered immediately when a cluster is paused. It's being used to inhibit all other alerts from within the cluster, assuming an operator or developer is investigating/developing.

The alert has severity=none because it's not actually an issue. Alertmanager needs to be configured to ignore these alerts accordingly.

```release-note
add new KubermaticClusterPaused alert with "none" severity for inhibiting alerts from paused clusters
```


kron4eg (approved):
/approve

PR #3850, opened 2019-07-19
[WIP] Kubermatic Operator

xrstf:
This implements the first parts of the new Kubermatic Operator, as per proposal (#3767).

```release-note
created Kubermatic Operator to manage deploying/maintaining Kubermatic
```


api/cmd/kubermatic-operator/controllers.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/oklog/run"
+	"go.uber.org/zap"
+
+	operatormaster "github.com/kubermatic/kubermatic/api/pkg/controller/operator-master"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type controllerCreator func(*operatorContext, *zap.SugaredLogger) (runnerFn, error)
+type runnerFn func(workerCount int, stopCh <-chan struct{}) error
```
alvaroaleman: Generally I am not sure if it makes sense to have something like the `controllerCreator`, because different controllers need diffent info in their initialization. But _if_ we decide to have such a thing, its should contain _all_ info, i.E. include the `workerCount`. That in turn would make the  `runerFN` type obsolete, as you can just use `manager.Runnable` instead (which is implemented by sigs.k8s.io/pkg/controller.Controller`

api/cmd/kubermatic-operator/controllers.go
```diff
@@ -0,0 +1,96 @@
...
+func noop(workerCount int, stopCh <-chan struct{}) error { <-stopCh; return nil }
+
+func controllerLog(log *zap.SugaredLogger, controller string) *zap.SugaredLogger {
+	return log.With("controller", controller)
+}
+
+// allControllers stores the list of all controllers that we want to run,
+// each entry holds the name of the controller and the corresponding
+// start function that will essentially run the controller
+var allControllers = map[string]controllerCreator{
```
alvaroaleman: I should take Henriks role over and mention that this is premature optimization, we only have one controller and shouldn't build infra for a problem we don't have :)

api/cmd/kubermatic-operator/controllers.go
```diff
@@ -0,0 +1,96 @@
...
+
+		// The controllers managed by the mgr don't have a dedicated runner
+		if controller != nil {
+			controllers[name] = controller
+		}
+	}
+	return controllers, nil
+}
+
+func runAllControllersAndCtrlManager(workerCnt int,
```
alvaroaleman: This doesn't make sense. The `manager` starts the controllers (or more specifically all `Runnables` that were passed into it via its `Add`. The reason we have something like this func here in Kubermatic is:
* Legacy cruft
* We cant use the `managers` leader election because we have to execute some migrations after aquiring the leader lease and before running the controllers

Generally speaking, everything that is long-running should implement the `Runnable` interface, be added to the manager and be started and managed by the `Manager`. Thankfully, `Mangager` itself also implements `Runnable` :)

api/cmd/kubermatic-operator/main.go
```diff
@@ -0,0 +1,166 @@
...
+	internalAddr string
+	log          kubermaticlog.Options
+	workerCount  int
+	workerName   string
+}
+
+type operatorContext struct {
+	runOptions                      controllerRunOptions
+	stopCh                          <-chan struct{}
+	kubeMasterClient                kubernetes.Interface
```
alvaroaleman: We don't need any of these. Just create a `manager` per seed and use its `GetClient` to get a lister-backed client. There is no reason at all to use the generated clientset or manage informers and listers ourselfes.

It also makes writing tests a lot easier, because you only need to replace one  `client`, instead of multiple clients that are used for different types and informers for them

api/cmd/kubermatic-operator/main.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		sugarLog.Fatalw("could not create all controllers", "error", err)
+	}
+
+	opCtx.kubermaticMasterInformerFactory.Start(opCtx.stopCh)
+	opCtx.kubeMasterInformerFactory.Start(opCtx.stopCh)
+	opCtx.kubermaticMasterInformerFactory.WaitForCacheSync(opCtx.stopCh)
+	opCtx.kubeMasterInformerFactory.WaitForCacheSync(opCtx.stopCh)
+
+	// This group is forever waiting in a goroutine for signals to stop
```
alvaroaleman: No need for waitgroup, the managers `Start` is blocking until the channel is closed

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// glog.V(4).Info("Reconciling Grafana provisioning...")
+	// if err := r.ensureMasterGrafanaProvisioning(ctx); err != nil {
+	// 	return reconcile.Result{}, fmt.Errorf("failed to reconcile Grafana: %v", err)
+	// }
+
+	return reconcile.Result{}, nil
+}
+
+/*
```
alvaroaleman: Why is all of this commented?

xrstf: Because the first commit was a braindump to ask Henrik a question.

xrstf:
I've now implemented the first reconciling parts. The deployment is not yet idempotent and causes and infinite reconcile loop, but the other parts seem to work fine already.

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return []reconcile.Request{
+		{
+			NamespacedName: name,
+		},
+	}
+}
+
+var namespaceNameRegex = regexp.MustCompile(`^\s*([^/]+)\s*/\s*([^/]+)\s*$`)
```
xrstf: Is there a ready-made package for this? I wanted to put the owning configuration name into a single annotation, but can just as well split it into two annotations or labels.

api/pkg/controller/operator-master/resources.go
```diff
@@ -0,0 +1,400 @@
...
+				"-address=0.0.0.0:8080",
+				"-internal-address=0.0.0.0:8085",
+				// "-datacenters=/opt/datacenter/datacenters.yaml",
+				"-kubeconfig=/opt/.kube/kubeconfig",
+				fmt.Sprintf("-oidc-url=%s", cfg.Spec.Auth.TokenIssuer),
+				fmt.Sprintf("-oidc-authenticator-client-id=%s", cfg.Spec.Auth.ClientID),
+				fmt.Sprintf("-oidc-skip-tls-verify=%v", cfg.Spec.Auth.SkipTokenIssuerTLSVerify),
+				fmt.Sprintf("-domain=%s", cfg.Spec.Domain),
+				fmt.Sprintf("-service-account-signing-key=%s", cfg.Spec.Auth.ServiceAccountKey),
+				// fmt.Sprintf("-feature-gates=%s", cfg.Spec.FeatureGates),
```
xrstf: This makes my original attempt to have fixed, static fields for each feature flag inside the CRD, harder to implement. Maybe we should go back to just having a `map[string]=bool` for the feature gates in the CRD?

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,223 @@
...
+		Reconciler: *r,
+		ctx:        ctx,
+		config:     config,
+		log:        logger,
+	}
+
+	return cfgReconciler.Reconcile()
+}
+
+type configReconciler struct {
```
xrstf: I got tired of handing the context, namespace and config object into each reconcile function, so I wrapped the context into this struct. `ns` is a field because I did not want to spread the "what is the effective namespace to work in?" logic into each named creator / reconciler function.

api/cmd/kubermatic-operator/main.go
```diff
@@ -0,0 +1,87 @@
...
+	config := clientcmd.NewNonInteractiveClientConfig(
+		*clientConfig,
+		clientConfig.CurrentContext,
+		&clientcmd.ConfigOverrides{CurrentContext: clientConfig.CurrentContext},
+		nil,
+	)
+
+	cfg, err := config.ClientConfig()
+	if err != nil {
+		sugarLog.Fatalw("Failed to create client", "error", err)
```
mrIncompetent: Can we rename `sugarLog` to `log` ?

api/cmd/kubermatic-operator/main.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	mgr, err := manager.New(cfg, manager.Options{MetricsBindAddress: opt.internalAddr})
+	if err != nil {
+		sugarLog.Fatalw("failed to create Controller Manager instance: %v", err)
+	}
+
+	if err := operatorv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		sugarLog.Fatalw("failed to register types in Scheme", "error", err)
```
mrIncompetent: `failed` -> `Failed`

api/cmd/kubermatic-operator/main.go
xrstf: We can and did.

api/cmd/kubermatic-operator/main.go
```diff
@@ -0,0 +1,87 @@
...
+	)
+
+	cfg, err := config.ClientConfig()
+	if err != nil {
+		sugarLog.Fatalw("Failed to create client", "error", err)
+	}
+
+	mgr, err := manager.New(cfg, manager.Options{MetricsBindAddress: opt.internalAddr})
+	if err != nil {
+		sugarLog.Fatalw("failed to create Controller Manager instance: %v", err)
```
mrIncompetent: `failed` -> `Failed`

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,185 @@
...
+	numWorkers int,
+	clientConfig *clientcmdapi.Config,
+	log *zap.SugaredLogger,
+	workerName string,
+) error {
+	reconciler := &Reconciler{
+		Client:       mgr.GetClient(),
+		recorder:     mgr.GetRecorder(ControllerName),
+		clientConfig: clientConfig,
+		log:          log,
```
mrIncompetent: `log: log.Named(ControllerName)`

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+
+	return nil
+}
+
+// eventHandler translates an incoming event into queue items, depending
+// on the affected object's type. The controller always puts the name
+// and namespace of the related KubermaticConfiguration on the queue.
+func eventHandler(a handler.MapObject) []reconcile.Request {
```
mrIncompetent: **Personal opinion**
I have to admit i'm not a huge fan of this handler as it has an entirely different logic based on the type.

It makes code less clear. I was a bit confused when i saw that the `KubermaticConfiguration` gets treated the same way as all other resources: https://github.com/kubermatic/kubermatic/pull/3850/files#diff-96bb0d5e8a5e69728cc12a27cd04c084R85

api/pkg/controller/operator-master/controller.go
mrIncompetent: https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L95

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return &types.NamespacedName{
+		Namespace: match[1],
+		Name:      match[2],
+	}
+}
+
+func joinNamespaceName(namespace string, name string) string {
+	return fmt.Sprintf("%s/%s", namespace, name)
```
mrIncompetent:
```go
name := NamespacedName{Name: "foo", Namespace: "bar"}
return name.String()
```

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,176 @@
...
+	// MasterController configures the controller-manager.
+	MasterController KubermaticMasterControllerConfiguration `json:"masterController,omitempty"`
+	// MasterFiles is a map of additional files to mount into each master component.
+	MasterFiles map[string]string `json:"masterFiles,omitempty"`
+	// ExposeStrategy is the strategy to expose the cluster with, either "NodePort" which creates a NodePort
+	// with a "nodeport-proxy.k8s.io/expose": "true" annotation to expose all clusters on one central Service
+	// of type LoadBalancer via the NodePort proxy or "LoadBalancer" to create a LoadBalancer service per cluster.
+	// Note: The `seed_dns_overwrite` setting of the `datacenters.yaml` doesn't have any effect if this is
+	// set to `LoadBalancer`.
+	ExposeStrategy string `json:"exposeStrategy,omitempty"`
```
mrIncompetent: Please define a custom type for `ExposeStrategy` + constants for the available strategies

api/pkg/resources/kubermatic/api.go
```diff
@@ -0,0 +1,256 @@
+package kubermatic
```
mrIncompetent: Move all resource packages belonging to the operator below the individual controller package.
Like: `/pkg/controller/operator-master/resources `

api/pkg/resources/kubermatic/api.go
```diff
@@ -0,0 +1,256 @@
...
+							corev1.ResourceCPU:    resource.MustParse("100m"),
+							corev1.ResourceMemory: resource.MustParse("64Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("250m"),
+							corev1.ResourceMemory: resource.MustParse("128Mi"),
+						},
+					},
+					TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+					TerminationMessagePath:   "/dev/termination-log",
```
mrIncompetent: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/container-linux/resources/deployment.go#L67

api/pkg/resources/kubermatic/master-controller-manager.go
```diff
@@ -0,0 +1,186 @@
...
+			pdb.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: masterControllerManagerPodLabels(),
+			}
+
+			return pdb, nil
+		}
+	}
+}
+
+func MasterControllerManagerServiceCreator(ns string, cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceCreatorGetter {
```
mrIncompetent: Please remove that.
We don't need a service for the master controller manager.

api/pkg/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,313 @@
...
+			pdb.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: seedControllerManagerPodLabels(),
+			}
+
+			return pdb, nil
+		}
+	}
+}
+
+func SeedControllerManagerServiceCreator(ns string, cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceCreatorGetter {
```
mrIncompetent: We don't need this service

api/pkg/resources/kubermatic/ui.go
```diff
@@ -0,0 +1,121 @@
...
+
+			return pdb, nil
+		}
+	}
+}
+
+func UIServiceCreator(ns string, cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		return uiServiceName, func(s *corev1.Service) (*corev1.Service, error) {
+			s.Spec.Type = corev1.ServiceTypeNodePort
```
mrIncompetent: ClusterIP

api/pkg/resources/kubermatic/master-controller-manager.go
```diff
@@ -0,0 +1,186 @@
...
+							corev1.ResourceCPU:    resource.MustParse("50m"),
+							corev1.ResourceMemory: resource.MustParse("128Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("100m"),
+							corev1.ResourceMemory: resource.MustParse("256Mi"),
+						},
+					},
+					TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+					TerminationMessagePath:   "/dev/termination-log",
```
mrIncompetent: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/container-linux/resources/deployment.go#L67

api/pkg/resources/kubermatic/master-controller-manager.go
```diff
@@ -0,0 +1,186 @@
...
+				},
+			}
+
+			return d, nil
+		}
+	}
+}
+
+func projectsMigratorContainer(cfg *operatorv1alpha1.KubermaticConfiguration) corev1.Container {
+	return corev1.Container{
```
mrIncompetent: Is missing default values like `TerminationMessagePolicy` & `TerminationMessagePath`

api/pkg/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,313 @@
...
+							corev1.ResourceCPU:    resource.MustParse("200m"),
+							corev1.ResourceMemory: resource.MustParse("512Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("500m"),
+							corev1.ResourceMemory: resource.MustParse("1Gi"),
+						},
+					},
+					TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+					TerminationMessagePath:   "/dev/termination-log",
```
mrIncompetent: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/container-linux/resources/deployment.go#L67

api/pkg/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,313 @@
...
+			{
+				MountPath: "/opt/addons",
+				Name:      "addons",
+			},
+		},
+	}
+}
+
+func seedControllerManagerCopyOpenshiftAddonsContainer(cfg operatorv1alpha1.KubermaticAddonConfiguration) corev1.Container {
+	return corev1.Container{
```
mrIncompetent: Is missing default values like TerminationMessagePolicy & TerminationMessagePath

api/pkg/resources/kubermatic/ui.go
```diff
@@ -0,0 +1,121 @@
...
+							corev1.ResourceCPU:    resource.MustParse("100m"),
+							corev1.ResourceMemory: resource.MustParse("64Mi"),
+						},
+						Limits: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("250m"),
+							corev1.ResourceMemory: resource.MustParse("128Mi"),
+						},
+					},
+					TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+					TerminationMessagePath:   "/dev/termination-log",
```
mrIncompetent: Is missing default values like TerminationMessagePolicy & TerminationMessagePath

api/pkg/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+}
+
+func SeedControllerManagerServiceCreator(ns string, cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		return seedControllerManagerServiceName, func(s *corev1.Service) (*corev1.Service, error) {
+			s.Spec.Type = corev1.ServiceTypeNodePort
+			s.Spec.Selector = seedControllerManagerPodLabels()
+
+			s.Spec.Ports = mergeServicePort(s.Spec.Ports, corev1.ServicePort{
```
mrIncompetent: Why do we want to merge here anything?
Why not forcefully overwriting?

mrIncompetent (changes requested):

api/pkg/resources/kubermatic/master-controller-manager.go
alvaroaleman: Not needed anymore, those are added when the deployment gets reconciled: https://github.com/kubermatic/kubermatic/blob/d80d2a87916c8e72dae799884bae3d45cf34be15/api/pkg/resources/reconciling/zz_generated_reconcile.go#L214

alvaroaleman: (which means its not needed anywhere so it can be removed again :grimacing: )

api/pkg/controller/operator-master/reconciler.go
alvaroaleman: Can we rename it to `configData` or something? This thing doesn't reconcile anything, its being used during reconciliation, which is a completely different thing. Also:
* The context is already part of the reconciler, why put it into a second struct?
* Why dont we take the namespace from the `KubermaticConfiguration`?

api/pkg/crd/client/clientset/versioned/clientset.go
```diff
@@ -4,6 +4,7 @@ package versioned
 
 import (
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned/typed/kubermatic/v1"
+	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned/typed/operator/v1alpha1"
```
alvaroaleman: Can we not generate the clientset? Else ppl will end up using it, we don't want that.

api/pkg/resources/kubermatic/api.go
```diff
@@ -0,0 +1,256 @@
...
+						Path:   "/api/v1/healthz",
+						Scheme: corev1.URISchemeHTTP,
+						Port:   intstr.FromInt(8080),
+					},
+				},
+			}
+
+			specLabels := apiPodLabels()
+
+			d.Spec.Replicas = i32ptr(2)
```
alvaroaleman: Please use the funcs in the `utilpointer "k8s.io/utils/pointer"` package for this

api/pkg/resources/kubermatic/api.go
```diff
@@ -0,0 +1,256 @@
...
+			d.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: specLabels,
+			}
+
+			d.Spec.Template.Labels = specLabels
```
alvaroaleman: If you initialize it like that, you can just set `d.Spec.Template.Labels = d.Spec.Selector.MatchLabes`. And at that point there is no reason to have the `specLabels` func anymore.

Also I should mention that I dislike funcs that have exactly one caller, unless there is a good reason, e.G. you want to test that specific codepath. It makes the whole thing hard to read, instead of going sequentally with the code, you have to jump to another location.

api/pkg/resources/kubermatic/api.go
```diff
@@ -0,0 +1,256 @@
...
+					fmt.Sprintf("-oidc-issuer-cookie-hash-key=%s", cfg.Spec.Auth.IssuerCookieKey),
+				)
+			}
+
+			volumes := []corev1.Volume{
+				{
+					Name: "kubeconfig",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							DefaultMode: i32ptr(420),
```
alvaroaleman: Utilpointer

api/pkg/resources/kubermatic/kubermatic.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func MasterFilesSecretCreator(ns string, cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return masterFilesSecretName, func(s *corev1.Secret) (*corev1.Secret, error) {
+			if s.Data == nil {
+				s.Data = make(map[string][]byte)
+			}
+
+			for name, content := range cfg.Spec.MasterFiles {
```
alvaroaleman: Makes me think if we should build a secretCreatorFactory that takes just name and and map[string][]byte and returns a creator lol

api/pkg/resources/kubermatic/master-controller-manager.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			args := []string{
+				"-v=4",
+				"-logtostderr",
+				"-internal-address=0.0.0.0:8085",
+				"-kubeconfig=/opt/.kube/kubeconfig",
+			}
+
+			volumes := []corev1.Volume{
```
alvaroaleman: Why not directly use `d.Volumes`? Same for args and everything else

api/pkg/resources/kubermatic/util.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+	"strings"
+
+	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// nameLabel is the recommended name for an identifying label.
```
alvaroaleman: "Recommended"? Either we use them in which case they are part of the desired state or we dont, in which we dont care

api/pkg/controller/operator-master/resources.go
xrstf: I have since writing this comment in fact replaced the FeatureGates with a simple map.

api/pkg/controller/operator-master/reconciler.go
xrstf: > Why dont we take the namespace from the KubermaticConfiguration?

We do?

```
func (r *configReconciler) Reconcile() error {
	r.applyDefaults() // sets the spec.Namespace to the object's actual namespace if none was given explicitly
	r.ns = r.config.Spec.Namespace
```

> The context is already part of the reconciler, why put it into a second struct?

Which context do you mean? The `ctx` context?

alvaroaleman: Lets please not do that. PPL that want the namespace should just access the `kubermaticConfiguration` object. There is no added value to move it to a different variable but it leads to confusion as to why there are two places where the NS is and how they are different.

Also it introduces the possibility of bugs, because some place in the code modifies the kubermaticConfiguration but not the `configReconciler` or the other way round. We should always have exactly one canonical source of truth which is the object itself.

alvaroaleman: Yeah, by context I mean `ctx`. And I must admit I don't see the added value of the `configReconciler` concept.

api/pkg/resources/kubermatic/master-controller-manager.go
mrIncompetent: Nice! Then we should issue a PR to remove them everywhere!

api/pkg/controller/operator-master/controller.go
alvaroaleman: Yeah I definitely agree, it also makes the relationship quite clear in the beginning, there is one `controllingObject` that uses `handler.EnqueueRequestForName` and a set of `controlled objects` that get a `enqOwnerViaCustomHandlingFunc`

xrstf: IMHO splitting the event handler into two will only create more code (two event handler functions, two .Watch() calls, two error handlings, two NamedspacedName struct creations, ...) with little to gain.

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+type predicateFunc func(runtime.Object, metav1.Object) bool
+
+// makeOwnerPredicate builds a new predicate using the given worker name. The
+// predicate checks whether the affected object has the proper labels.
+func makeOwnerPredicate(workerName string) predicate.Predicate {
+	return makePredicateFuncs(func(obj runtime.Object, meta metav1.Object) bool {
+		labels := meta.GetLabels()
+
+		// only reconcile configurations with our given worker name label
+		if _, ok := obj.(*operatorv1alpha1.KubermaticConfiguration); ok {
```
alvaroaleman: Can we simplify this by just doing a check in the beginning of the `Reconcile`? As you probably noticed, we can not really use the predicate concept as it checks the object itself, not the owner. But IMHO having a three line `if workerNameLabel != workerName { return nil }` is much simpler than this 12 line func that does type assertion

alvaroaleman: s/12 line func/one type and two funcs that occupy a total of 40 LOC/

api/pkg/controller/operator-master/reconciler.go
xrstf: > Also it introduces the possibility of bugs, because some place in the code modifies the kubermaticConfiguration but not the configReconciler or the other way round. We should always have exactly one canonical source of truth which is the object itself.

Agreed. Also, since the namespace does not need to be set in each creator function, this already reduces the repetitive access to the field in the spec anyway.

> Yeah, by context I mean ctx. And I must admit I don't see the added value of the configReconciler concept.

In the `Reconcile()` function we create three things: a context for cancellation, the kubermatic config and a logger with additional metadata for the kubermatic config. These three things would then need to be passed into all `reconcile<Object>` functions. This made my spider sense tingle because obviously these three fields are a common context for all the reconcile functions (in addition to the existing shared state in the `Reconciler` struct itself). I would have liked to spawn a new Reconciler for each incoming reconcile request, but it seemed the architecture of our code is built around the concept of a single reconciler. So I did the next best thing and created a new helper that wraps the original Reconciler and adds the new context. I don't neccessarily like its name either, but I liked the concept of having an obvious shared state for all reconcile functions.

api/pkg/resources/kubermatic/api.go
xrstf: > Also I should mention that I dislike funcs that have exactly one caller

The `apiPodLabels()` function is used for defining the PDB and Services as well.

api/pkg/resources/kubermatic/master-controller-manager.go
xrstf: I found shorter variables names increase readability and separating assembling these values from finally assembling the Deployment also helps my understand. It felt wrong to me to create a Deployment object and then start poking around and adding values here and there.

api/pkg/resources/kubermatic/seed-controller-manager.go
xrstf: When I forcefully set the values (`s.Spec.Ports = []....`), I ended up in an endless reconcile loop because the `nodePort` was added. That made me be cautious. The other approaches I found (manually doing `s.Spec.Ports[i].Name= ...`) looks brittle (if someone overwrote a port with theirs, they would lose their port after reconciling -- my code would *add* our port back).

api/pkg/resources/kubermatic/ui.go
xrstf: The UI service in our Helm chart says NodePort: https://github.com/kubermatic/kubermatic/blob/master/config/kubermatic/templates/kubermatic-ui-service.yaml

api/pkg/resources/kubermatic/util.go
xrstf: It's badly worded, I admit. I meant it's recommended by the Kubernetes docs to be used as a label. The comment was supposed to express that this is the label we use for naming because it is recommended.

api/pkg/controller/operator-master/controller.go
xrstf: That would certainly currently do the same, but you have now mixed up our splitting implementation with Kubernetes' joining logic. If either changes, the split+join combo would be broken. That's why I opted (in the now irrelevant code) to have my own join function.

api/pkg/controller/operator-master/controller.go
xrstf: The whole event handling stuff has been reworked.

api/pkg/controller/operator-master/reconciler.go
alvaroaleman: >  I don't neccessarily like its name either, but I liked the concept of having an obvious shared state for all reconcile functions.

Yeah we do that elsewhere with the `templateData` and `openshiftData`. Re the context, while we have some controllers that do create a context per reconciliation, I must admit that this is a pattern I introduced and which doesn't make any sense. The only case where this context should get canceled is for shutdowns of the whole controller.

api/pkg/crd/client/clientset/versioned/clientset.go
xrstf: I think I updated the codegen script accordingly now.

api/pkg/resources/kubermatic/master-controller-manager.go
alvaroaleman: It already is used in most places. What should be considered thought is that his is applied on a higher level and currently only implemented for Deployments and Statefulsets, so we should probably add it for DaemonSets as well :grimacing:

api/pkg/controller/operator-master/resources/kubermatic/util.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// versionLabel is the recommended name for a version label.
+	versionLabel = "app.kubernetes.io/version"
+)
+
+// mergeServicePort searches the src port inside the dst slice and if it
+// finds it, updates its properties. If the src port is not found yet,
+// it will be appended verbatim to the dst list. If dst is nil, a new
+// slice will be created.
+func mergeServicePort(dst []corev1.ServicePort, src corev1.ServicePort) []corev1.ServicePort {
```
alvaroaleman, line 23: Please remove, if such a func exists, ppl will think they are supposed to use it. We have to cases where wo do trickery with ports, both of them are on a per-usercluster level and in the responsibility of the kubermatic-controller-manager.

This operator should just achieve a desired state and not merge desired state together with current state on the vague assumption that current state somehow is part of the desired state just because it exists.

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	NodePortStrategy ExposeStrategy = "NodePort"
+	// LoadBalancerStrategy creates a LoadBalancer service per cluster.
+	LoadBalancerStrategy ExposeStrategy = "LoadBalancer"
+)
+
+// DockerImage describes a Docker image.
+type DockerImage struct {
+	Repository string            `json:"repository"`
+	Tag        string            `json:"tag"`
+	PullPolicy corev1.PullPolicy `json:"pullPolicy"`
```
alvaroaleman, line 25: Why does this type exist and why do we want to make the pullPolicy configurable?

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+
+	t := &operatorv1alpha1.KubermaticConfiguration{}
+	if err := c.Watch(&source.Kind{Type: t}, eventHandler, predicate); err != nil {
+		return fmt.Errorf("failed to create watcher for %T: %v", t, err)
+	}
+
+	// configure watches for all possibly controlled objects (filtering to only include those for the
+	// current worker name is done by the reconciler later)
+	predicate = newPredicateFuncs(func(meta metav1.Object) bool {
```
alvaroaleman, line 97: Can you explain why you don't want at the start of the `Reconcile` a simple check if the worker-name matches and instead use two predicate funcs?

3 LOC vs 6 LOC, also the possibility to add a debug log indicating to someone reading the logs that a given cluster is being ignored because its worker-name label doesn't match.

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+				NamespacedName: types.NamespacedName{
+					Namespace: a.Meta.GetNamespace(),
+					Name:      a.Meta.GetName(),
+				},
+			},
+		}
+	})
+
+	t := &operatorv1alpha1.KubermaticConfiguration{}
+	if err := c.Watch(&source.Kind{Type: t}, eventHandler, predicate); err != nil {
```
alvaroaleman, line 91: Thats obviously personal opinion, but assigning something to a variable that is complely not self-explanatory named makes the code for me pretty hard to read. Instead of just going doing and seeing `a source kind of type operatorv1alpha1.KubermaticConfiguration{` I see `a a source kind of type t is being watched, lets go one line back to see what type t is`

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	// configure watches for all possibly controlled objects (filtering to only include those for the
+	// current worker name is done by the reconciler later)
+	predicate = newPredicateFuncs(func(meta metav1.Object) bool {
+		return meta.GetLabels()[ManagedByLabel] == ControllerName
+	})
+
+	// put the owner onto the queue
+	eventHandler = newEventHandler(func(a handler.MapObject) []reconcile.Request {
```
alvaroaleman, line 102: Again personal opinion, but I dislike variables that have a generic name like `eventHandler` and get assigned a complete different value so depending on where I look at `eventHandler` it means complete different things.

something like `eventHandlerChildObject` and `eventHandlerMainObject` make the whole thing IMHO much more readable.

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	// reconcile for every KubermaticConfiguration labelled with the current worker name
+	predicate := newPredicateFuncs(func(meta metav1.Object) bool {
+		return meta.GetLabels()[WorkerNameLabel] == workerName
+	})
+
+	// put the config's identifier on the queue
+	eventHandler := newEventHandler(func(a handler.MapObject) []reconcile.Request {
```
alvaroaleman, line 79: This is a re-implementation of `sigs.k8s.io/controller-runtime/pkg/handler.EnqueueRequestForObject`

api/pkg/controller/operator-master/controller.go
```diff
@@ -0,0 +1,169 @@
...
+		&corev1.Secret{},
+		&corev1.Service{},
+		&corev1.ServiceAccount{},
+		&extensionsv1beta1.Ingress{},
+		&rbacv1.ClusterRole{},
+		&rbacv1.ClusterRoleBinding{},
+	}
+
+	for _, t := range typesToWatch {
+		if err := c.Watch(&source.Kind{Type: t}, eventHandler, predicate); err != nil {
```
alvaroaleman, line 132: (Here using `t` as a variable name is btw perfectly reasonable because it indicates that its a type that is different in every iteration of the loop)

api/pkg/controller/operator-master/controller.go
alvaroaleman: sorry correction: 3 LOC vs 30 LOC. And three  LOC that are completely obvious to reason about, the amount of funcs added to the code here just to get a predicate func is quite a lot.

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to determine string key for KubermaticConfiguration: %v", err)
+	}
+
+	logger := r.log.With("config", identifier)
+
+	return reconcile.Result{}, r.reconcile(ctx, config, logger)
+}
+
+func (r *Reconciler) fetchKubermaticConfiguration(ctx context.Context, identifier types.NamespacedName) (*operatorv1alpha1.KubermaticConfiguration, error) {
```
alvaroaleman, line 62: Why do we create a function that is being called exactly one and and very unlikely to need a dedicated unittest?

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+
+	if err := r.Get(ctx, identifier, cfg); err != nil {
+		return nil, fmt.Errorf("could not find KubermaticConfiguration %s", identifier)
+	}
+
+	return cfg, nil
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
+	// TODO: Implement garbage collection to remove orphaned resources.
```
alvaroaleman, line 73: Please use a dedicated controller for garbage collection, this one is complicated enough as-is.

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,387 @@
...
+		img.PullPolicy = pullPolicy
+		r.log.Debugf("Defaulting Docker repository for %s.pullPolicy to %s", key, pullPolicy)
+	}
+}
+
+// applyDefaultFields is generating a new ObjectModifier that wraps an
+// ObjectCreator and takes care of applying the default labels and
+// annotations from this operator. These are then used to establish
+// a weak ownership.
+func (r *configData) applyDefaultFields() reconciling.ObjectModifier {
```
alvaroaleman, line 168: Can we rename to `defaultMetadataModifier()`? One has to read the comment or the code to understand that this doesn't change `fields`, it always changes metadata. Also it doesnt apply anything, its a factory to get a func which applies something.

alvaroaleman:
@xrstf Can we please split this PR up? E.G. into:

* PR 0: Add types
* PR 1: Add a no-op reconciler that watches just the newly-added types
* PR 2: Add a single thing to reconcile to the reconciler and with it what should now be watched on top.

PRs of this size:
* Make your reviewers not want to review it because of the sheer size
* Have a very high probably of introducing bugs because reviewers dont catch them because of the sheer size of the PR
* Have a high chance of doing something wrong in multiple places

Please try to create PRs in a way that they change one thing at a time.

api/pkg/resources/kubermatic/master-controller-manager.go
xrstf: => #3891

api/pkg/crd/operator/v1alpha1/configuration.go
xrstf: It mirrors the exact same struct we have all over the place in our Helm charts. I assumed for offline clusters we still want the ability to overwrite repositories, need to overwrite tags for testing and the pull policy for testing as well. Am I wrong?

api/pkg/controller/operator-master/controller.go
xrstf: But then you need to replicate the same type into the error message just one line below this one and I have already seen cases in our codebase where the error message mentions a totally different object kind than was actually intended.

api/pkg/controller/operator-master/controller.go
xrstf: I was under the assumption that if there is a predicate feature, I am supposed to use it to check for relevant objects as soon as possible, most likely in an effort to improve controller performance in general.

xrstf:
This is being split up in #3896, #3898 and future PRs.

PR #3853, opened 2019-07-19
Add check for share kubeconfig and OIDC

### LINKED ISSUE #3852: Add check for share kubeconfig feature and OIDC settings during  claster creation
**User Story**
User can not set OIDC settings for the cluster when OIDCKubeCfgEndpoint is enabled

zreigz:
**What this PR does / why we need it**: Add check for share kubeconfig and OIDC

Fixes #3852

```release-note
NONE
```


api/pkg/provider/kubernetes/cluster.go
```diff
@@ -69,17 +72,23 @@ type ClusterProvider struct {
...
+	k8sClient            kubernetes.Interface
 }
 
 // New creates a brand new cluster that is bound to the given project
 func (p *ClusterProvider) New(project *kubermaticv1.Project, userInfo *provider.UserInfo, cluster *kubermaticv1.Cluster) (*kubermaticv1.Cluster, error) {
 	if project == nil || userInfo == nil || cluster == nil {
 		return nil, errors.New("project and/or userInfo and/or cluster is missing but required")
 	}
+	// share kubeconfig feature is contrary to cluster OIDC setting
+	if p.oidcKubeConfEndpoint && !reflect.DeepEqual(cluster.Spec.OIDC, kubermaticv1.OIDCSettings{}) {
```
alvaroaleman, line 88: Wouldn't it be easier to make `cluster.Spec.OIDC` a pointer and check ifs its nil?

zreigz: Yes, It's easier but I don't like using pointers if not needed. The main reason is debugging or printing values. Is hard to read pointer address. Sharing data via pointers will likely cause the data to be placed in the “heap.”  There is overhead to adding data to the heap and heap data can only be cleaned up by the garbage collector.

alvaroaleman: Err. I don't have an opinion towards heap or not and think it doesn't make much difference either way. But  structs always get serialized and pointers to structs dont

PR #3855, opened 2019-07-19
Dynamically load cloudproviders

alvaroaleman:

* Changes the CloudProviders to accept at maximum one Datacenter instead of a `map[string]*Seed`
* Changes all our code to initialize the cloudProviders when they are needed instead of at startup
* This is a requirement for #2113 

Fixes #

I had to change a lot of the testcode, because it ended up actually trying to connect to DO

**Documentation**:

```release-note
none
```


api/pkg/handler/v1/cluster/cluster_test.go
```diff
@@ -659,7 +659,7 @@ func TestCreateClusterEndpoint(t *testing.T) {
 		// scenario 1
 		{
 			Name:                   "scenario 1: a cluster with invalid spec is rejected",
-			Body:                   `{"cluster":{"name":"keen-snyder","spec":{"cloud":{"digitalocean":{"token":"dummy_token"},"dc":"us-central1"}, "version":""}}}`,
+			Body:                   `{"cluster":{"name":"keen-snyder","spec":{"cloud":{"fake":{"token":"dummy_token"},"dc":"fake-dc"}, "version":""}}}`,
```
alvaroaleman, line 662: So all the `CreateCluster` started to panic after the changes in this PR and it took me some time to figure out why. The reason is that:

Before:
* We find out the cloudproviders name from the cluster
* Pick an instance of it off a map using that name
* Inititialize it
* We never do any crosschecking if it matches the cloudprovider in the dc
* This is why these tests were able to create a cluster with a `fake` cloudprovider in a `digitalocean` datacenter (In reality, this will cause the api to panic with a NPD once you create/update a node/Nodedeployment )

After:
* We get the datacenter by its name on the cluster
* We get the right cloudprovider based on that datacenter
* These tests now initialize the `Digitalocean` cloudprovider instead of the `fake` one, as that is whats on the DC
* The DO `ValidateCloudSpec` panics with a NPD when trying to access its token

I fixed the production code by adding a crosscheck if the clusters cloudprovider matches the one on the DC. The tests were a bit more involved to fix thought, because:
* We can not use any cloudprovider other than `fake` or `bringyourown`, because we call the cloudproviders  `ValidateCloudSpec` in the  `CreateCluster` endpoint and that returns a 401 from the cp for everything except the two aforementioned
* There is a test that checks if applying `presets` works, since there are no presets for `bringyourown`, we cant use that
* This leaves us with adding a `fake` datacenter type which is what I did. I am not particulary fond of that, as parts of it will appear in our swagger spec.

The only other alternative I see is hackery to keep on getting a `fake` cloudprovider even though thats not on the DC, by e.G.:
* Returning the `fake` cp if the dc has no cp at all: IMHO hard to reason about and likely to create issues down the line
* Having a package-level boolean var in `pkg/provider/cloud` that makes `Provider` always a `fake` cp if set to true: Reasonable sure @mrIncompetent won't let that get it

api/pkg/resources/machine/common.go
```diff
@@ -97,6 +100,9 @@ func getAWSProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *ku
 }
 
 func getAzureProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*runtime.RawExtension, error) {
+	if dc.Spec.Azure == nil {
```
alvaroaleman: I am a bit torn wrt the question if we should keep these, I initially added them for debugging the tests. On the one hand, callees shouldn't have to check their args, on the other hand if we ever end up with a cluster that has a cloud provider that doesn't match the one on the DC, this will reliable make the API panic with a NPD once a node or nodeDeployment is created/updated.

It also opens up the wider question if we should have a `defer recover()` in our API handlers, so errors in the pointer handling dont result in a crash of the whole app

api/pkg/handler/test/helper.go
```diff
@@ -303,13 +298,11 @@ func buildSeeds() map[string]*kubermaticv1.Seed {
 							},
 						},
 					},
-					"us-central1": {
-						Location: "us-central",
-						Country:  "US",
+					"fake-dc": {
+						Location: "Henriks basement",
```
floreks, line 302: :joy:

maciaszczykm:
Codewise looks good to me. The issue with datacenter and the tests is a really tough point here. Finding workaround sounds a bit better than removing all impacted tests, but it cannot impact user/security. Unfortunately, @zreigz is on vacation for the next 2 weeks because it would be nice if he could check too.

alvaroaleman:
@maciaszczykm Re 
>  Finding workaround sounds a bit better than removing all impacted tests, but it cannot impact user/security. 

The only drawback of adding the `FakeDatacenter` is that it will appear in some of the swagger spec. Other than that, the approach is (IMHO) the cleanest possible and its the thing we should have in done in the first place, instead of using mixed providers in the cluster and datacenter.

api/pkg/handler/test/helper.go
mrIncompetent: Finally i have a tombstone :ok_hand: :joy:

maciaszczykm:
@alvaroaleman I agree, this makes the most sense. IIRC we do not have any other clients than dashboard, so having it in swagger is not that bad after all. If it will be well-documented with comments then I think it's the best solution.

api/pkg/handler/v1/cluster/cluster_test.go
mrIncompetent: > Reasonable sure @mrIncompetent won't let that get it 

:100: 

Having a `fake` datacenter & using that for unit tests seems like the most clean option here :+1:.

api/pkg/provider/cloud/provider.go
```diff
@@ -15,18 +17,54 @@ import (
...
+		return fake.NewCloudProvider()
+	}
+	return nil
+}
+
+func FakeProvider(datacenter *kubermaticv1.Datacenter) provider.CloudProvider {
+	return fake.NewCloudProvider()
+}
+
+func OpenstackProvider(datacenter *kubermaticv1.Datacenter) (*openstack.Provider, error) {
```
mrIncompetent: It feels easy to miss for a developer to use those functions instead of directly calling `openstack.NewCloudProvider(datacenter.Spec.Openstack)`.

Maybe move the check into the provider?
So `openstack.NewCloudProvider` actually checks if the passed datacenter is an OpenStack datacenter?

api/pkg/resources/machine/common.go
mrIncompetent: > callees shouldn't have to check their args

:100: 

> It also opens up the wider question if we should have a defer recover() in our API handlers, so errors in the pointer handling dont result in a crash of the whole app

I was expecting that we actually have that :thinking: 
But i could not find the corresponding option. 
@floreks Can you check on this?

api/pkg/provider/cloud/provider.go
alvaroaleman: Currently the cloud providers only get their own spec and not the whole DC spec. So should I make them check if the passed in `*DatacenterSpec$PROVIDER` is non-nil?

alvaroaleman: Oh I get what you mean. Sec

alvaroaleman:
@mrIncompetent @maciaszczykm Made the cloudprovider initialization always check if the corresponding spec field is non-nil, PTAL

PR #3857, opened 2019-07-21
Use wrapper instead of initContainer to check if apiserver is reachable

### LINKED ISSUE #3795: Move apiserver readiness-check out of init containers
**User Story**
After an apiserver (inside a user cluster control plane) has been unavailable for let's say an hour, all depending pods (scheduler, controller-manager, machine-controller, ...) will have entered a CrashLoop and have their backoff time increased to something like 10+ minutes. This means that when an admin fixes the apiserver, the other pods stay dead for quite some time.

We should reduce the max waiting time or find a smarter solution to check for the apiserver. Currently there are some init containers in place to do just that, but they do not work well when the application container itself (like the scheduler) die again and again.

We should maybe move away from init containers and instead tweak the actual container's entrypoint to include a wait-loop for the apiserver.

**Acceptance criteria**
- [ ] The time between the apiserver becoming available and the other pods noticing and restarting is reduced and never exceeds a sensible threshold like 3 minutes.

alvaroaleman:

A set of the controlplane components crashes if they can't reach the apiserver. To prevent them from going into crashloopbackoff while the apiserver is unavailable, we added an initContainer to check if the apiserver is reachable. The problem is that initcontainers get executed only once, meaning once the apiserver was reachable and the initcontainer executed, the initcontainer doesn't get executed again when the apiserver becomes unreachable a second time and the component may enter crashloopbackoff and consequently extending the duration until the controlplane is functioning properly again.

To fix this, this PR:

* Extends the `http-prober` with a new `command` arg
* If set, the `http-prober` will do a `EXECVE(2)` after its target became reachable, effectively replacing itself with the command in the `command` arg
* Adds a wrapper, that takes a `PodSpec` and a set of named containers, which will:
    * Add a new `emptyDir` volume
    * Add an `initContainer` that copies the `http-proper` onto that empty dir
    * Adds that `emptyDir` volume to all named containers
    * Replaces the command in all named containers with the `http-prober`
    * Adds the original command as `command` arg to the `http-prober`
* Replaces all existing occurences of the http-prober initContainer with the new wrapper

/assign @nikhita 
CC @thz to sanity check the syscall usage :)
CC @xrstf as you found the original issue
Fixes #3795

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/cmd/http-prober/Dockerfile
+++ b/api/cmd/http-prober/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.7
+FROM alpine:3.9
 LABEL maintainer "henrik@loodse.com"
 
 COPY ./http-prober /usr/local/bin/
--- a/api/cmd/http-prober/api/httpprober.go
+++ b/api/cmd/http-prober/api/httpprober.go
@@ -1,0 +1,6 @@
+package api
+
+type Command struct {
+	Command string   `json:"command"`
+	Args    []string `json:"args"`
+}
--- a/api/cmd/http-prober/main.go
+++ b/api/cmd/http-prober/main.go
@@ -2,21 +2,18 @@ package main
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"flag"
 	"log"
 	"net/http"
 	"net/http/httptrace"
 	"net/url"
 	"os"
+	"os/exec"
+	"syscall"
 	"time"
-)
 
-var (
-	endpoint  string
-	insecure  bool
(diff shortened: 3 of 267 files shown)


nikhita (approved):
/approve

>  CC @thz to sanity check the syscall usage :)

Lgtm but will leave it for @thz to add the label for sanity check :)

/hold
until dev0 suffix is removed from the version

nikhita:
err... github-approve-as-lgtm can be surprising :upside_down_face:

PR #3860, opened 2019-07-22
AWS multi-AZ: do not save the default subnet on cluster level

kdomanski:
Instead allow AWS to select it based on AZ.

A step towards #3456

The current behavior of determining the default subnet ourselves provides no benefit. Saving it to be later enforced on all MachineDeployments gets in the way of using multiple Availiability Zones.

```release-note
NONE
```


api/pkg/provider/cloud/aws/provider.go
```diff
@@ -56,44 +56,41 @@ func (a *AmazonEC2) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
...
+	}
 
-		_, err = getDefaultSubnet(client, *vpc.VpcId, datacenter.Spec.AWS.Region+datacenter.Spec.AWS.ZoneCharacter)
+	// ensure that there is a default subnet in every AZ
+	azs, err := a.GetAvailabilityZonesInRegion(spec, datacenter.Spec.AWS.Region)
+	if err != nil {
+		return err
+	}
+	for _, az := range azs {
+		_, err = getDefaultSubnet(client, vpcID, *az.ZoneName)
```
mrIncompetent: Why do we require a default subnet in each AZ ?

kdomanski: We require a default subnet in each AZ that we want to use, except now we don't know which one that would be, so we check in all of them instead of one hardcoded zone.

mrIncompetent: Before we did only required a default subnet if `subnetID` was not specified by the user.
Now we always require a default subnet in each AZ of a VPC? Whats the technical requirement here?

This breaks scenarios where people only have 1 subnet or even just normal subnets (Which are not marked default).

kdomanski: Fixed.

mrIncompetent: > Fixed.

What exactly? Looking at the code we still require a default subnet in each AZ

kdomanski: Oops, I forgot to `git add`.

PR #3883, opened 2019-07-24
update machine-controller to v1.5.1

kdomanski:
```release-note
Updated machine controller to `v1.5.1`
```

/assign @alvaroaleman


alvaroaleman:
@kdomanski You also need to update the const which we use for the machine-controllers docker image tag in `pkg/resources/machine-controller`

alvaroaleman:
This failed multiple times with all tests failing, this is almost guaranteed to not be a flake

alvaroaleman:
/lgtm cancel
Something is wrong, we now import multiple versions of the Azure stuff

kdomanski:
How do you figure that?

alvaroaleman:
> How do you figure that?

By looking at the `Gopkg.lock`, there are only additions for azure, nothing removed

kdomanski:
The updated Azure SDK for some reason removed certain API functions from the `2018-04-01` API, so I had to switch to the `2018-06-01` API in the cloud provider package. The machine controller didn't use any of the removed stuff, so it remained on `2018-04-01`, which is why both are in the dependency graph.

alvaroaleman:
So why dont we use the 2018-06-01 version everywhere?

kdomanski:
Because the update to the machine-controller came first and only later when updating kubermatic it turned out that some API calls were removed.

Note that this is API versions, not library versions.

alvaroaleman:
Perfectly fine but the API versions seem to pull in the lib versions and I'd prefer to not have two of them. Would you be ok with doing the bump to the version that actually works on the machine-controller and then updating this to have only one version imported?

PR #3891, opened 2019-07-25
remove explicit TerminationMessage and ImagePullPolicy fields

xrstf:
We have defaulting functions that take care of applying the default ImagePullPolicy and TerminationMessage stuff, so we do not need to set these explicitly everywhere. A few tests have been added to ensure the defaulting works.

```release-note
NONE
```

### DIFF
--- a/api/pkg/resources/reconciling/wrapper.go
+++ b/api/pkg/resources/reconciling/wrapper.go
@@ -66,21 +66,11 @@ func DefaultPodSpec(old, new corev1.PodSpec) (corev1.PodSpec, error) {
 	}
 
 	for idx, container := range new.InitContainers {
-		t := corev1.DefaultProcMount
-		if tt, ok := initContainerProcMountType[container.Name]; ok && tt != nil {
-			t = *tt
-		}
-
-		DefaultContainer(&new.InitContainers[idx], &t)
+		DefaultContainer(&new.InitContainers[idx], initContainerProcMountType[container.Name])
 	}
 
 	for idx, container := range new.Containers {
-		t := corev1.DefaultProcMount
-		if tt, ok := containerProcMountType[container.Name]; ok && tt != nil {
-			t = *tt
-		}
-
-		DefaultContainer(&new.Containers[idx], &t)
+		DefaultContainer(&new.Containers[idx], containerProcMountType[container.Name])
 	}
 
 	return new, nil
--- a/api/pkg/resources/reconciling/wrapper.go
+++ b/api/pkg/resources/reconciling/wrapper.go
@@ -2,6 +2,7 @@ package reconciling
 
 import (
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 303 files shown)


api/pkg/resources/reconciling/wrapper.go
```diff
@@ -48,34 +49,115 @@ func DefaultContainer(c *corev1.Container, procMountType *corev1.ProcMountType)
...
 }
 
-// DefaultDeployment defaults all Deployment attributes to the same values as they would et from the Kubernetes API
+// DefaultPodSpec defaults all Container attributes to the same values as they would get from the Kubernetes API
+func DefaultPodSpec(old, new corev1.PodSpec) (corev1.PodSpec, error) {
+	// make sure to keep the old procmount types in case a creator overrides the entire PodSpec
+	initContainerProcMountType := map[string]*corev1.ProcMountType{}
+	containerProcMountType := map[string]*corev1.ProcMountType{}
+	for _, container := range old.InitContainers {
+		if container.SecurityContext != nil {
```
alvaroaleman, line 58: Nit: The condition is not needed

xrstf: It should be easy to construct a unit test for this *public* function that would cause a panic if this check wasn't in place, or am I mistaken?

alvaroaleman: Oh wait, I thought we never dereference it but this check is about the SecurityContext, not the procMountType, my bad

api/pkg/resources/reconciling/wrapper.go
```diff
@@ -48,34 +49,115 @@ func DefaultContainer(c *corev1.Container, procMountType *corev1.ProcMountType)
...
+		if tt, ok := initContainerProcMountType[container.Name]; ok && tt != nil {
+			t = *tt
+		}
+
+		DefaultContainer(&new.InitContainers[idx], &t)
+	}
+
+	for idx, container := range new.Containers {
+		t := corev1.DefaultProcMount
+		if tt, ok := containerProcMountType[container.Name]; ok && tt != nil {
```
alvaroaleman: I think this is wrong, because we shouldn't actually default this property, we should just keep it if it was set. The current logic checks if the property is on the map and if yes, if its non nil and only if that is the case it dereferences it, assigns to `t`, then passes a pointer to `t` into `DefaultContainer`.

I think we can just pass whatever is on the map for the container in, its completely fine if its a nilpointer, that means it was a nilpointer before and and here we definitely do not dereference it anymore

xrstf: Changed.

xrstf: Hm doing this unconditionally means we would also restore nil values, effectively disabling the defaulting alltogether.  See the `Default values for security contexts are added` testcase, which now fails:

```
		{
			name: "Default values for security contexts are added",
			newObject: corev1.PodSpec{
				Containers: []corev1.Container{
					{
						SecurityContext: &corev1.SecurityContext{},
					},
				},
			},
			expectedObject: corev1.PodSpec{
				Containers: []corev1.Container{
					{
						ImagePullPolicy:          corev1.PullIfNotPresent,
						TerminationMessagePath:   corev1.TerminationMessagePathDefault,
						TerminationMessagePolicy: corev1.TerminationMessageReadFile,
						SecurityContext: &corev1.SecurityContext{
							ProcMount: &defaultProcMountType,
						},
					},
				},
			},
		},
```

xmudrii (approved):
/lgtm
/approve

PR #3892, opened 2019-07-26
Use seedKubeconfigGetter for seed-proxy-controller

alvaroaleman:

This PR changes the seed-proxy-controller to use a `seedKubeconfigGetter` to dynamically resolve the kubecofig for a given seed at runtime. This is required for for #2113

Fixes #

What should be mentioned is that the `seed-proxy-controller` doesn't work for unrelated reasons, namely it has a predicate that makes it only react to resources it created. However initially there will be no such resources, hence the controller never does anything. I tested the whole change by temporarily removing the predicate.
**Documentation**:

```release-note
none
```

/assign @xrstf

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -48,10 +48,11 @@ type controllerContext struct {
 	kubermaticMasterInformerFactory externalversions.SharedInformerFactory
 	kubeMasterInformerFactory       kuberinformers.SharedInformerFactory
 
-	mgr               manager.Manager
-	kubeconfig        *clientcmdapi.Config
-	seedsGetter       provider.SeedsGetter
-	labelSelectorFunc func(*metav1.ListOptions)
+	mgr                  manager.Manager
+	kubeconfig           *clientcmdapi.Config
+	seedsGetter          provider.SeedsGetter
+	seedKubeconfigGetter provider.SeedKubeconfigGetter
+	labelSelectorFunc    func(*metav1.ListOptions)
 }
 
 func main() {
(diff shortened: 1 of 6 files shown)


api/cmd/master-controller-manager/main.go
```diff
@@ -133,6 +134,11 @@ func main() {
 	if err != nil {
 		sugarLog.Fatalw("failed to get construct seedsGetter", "error", err)
 	}
+	ctrlCtx.seedKubeconfigGetter, err = provider.SeedKubeconfigGetterFactory(
+		ctx, mgr.GetClient(), ctrlCtx.runOptions.kubeconfig, ctrlCtx.runOptions.dynamicDatacenters)
+	if err != nil {
+		sugarLog.Fatalf("failed to construct seedKubeconfigGetter", "error", err)
```
xrstf: `Fatalw`

api/cmd/master-controller-manager/main.go
```diff
@@ -133,6 +134,11 @@ func main() {
 	if err != nil {
 		sugarLog.Fatalw("failed to get construct seedsGetter", "error", err)
```
xrstf: Can you also while you're at it remove the `get` here?

api/pkg/controller/seed-proxy/controller.go
```diff
@@ -116,13 +116,16 @@ func Add(
 	}
 
 	eventHandler := &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
-		requests := make([]reconcile.Request, 0)
+		seeds, err := seedsGetter()
+		if err != nil {
+			glog.Errorf("failed to get seeds: %v", err)
```
xrstf: `Failed` with a capital F.

PR #3894, opened 2019-07-26
Use opt-in rather than opt-out mechanism to determine if initial nd creation error is unrecoverable

alvaroaleman:

Currently we instanly stop retrying to create the inital nodedeploymen when we hit an unknown error. This PR changes that behaviour to instead retry unless we get an error that contains an admission webhook response, as this is the only case where we know for sure retrying wont help.

Fixes #

**Documentation**:

```release-note
A bug that sometimes resulted in the creation of the initial NodeDeployment failing was fixed
```

### DIFF
--- a/api/pkg/handler/v1/cluster/cluster.go
+++ b/api/pkg/handler/v1/cluster/cluster.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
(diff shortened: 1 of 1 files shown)


api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -167,24 +168,18 @@ func createInitialNodeDeploymentWithRetries(nodeDeployment *apiv1.NodeDeployment
...
-			//  Internal error occurred: failed calling webhook "machine-controller.kubermatic.io-machinedeployments"
-			//  Post REDACTED: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
-			fallthrough
-		case kerrors.IsServiceUnavailable(err):
-			fallthrough
-		case kerrors.IsServerTimeout(err):
+		if err != nil {
+			// unrecoverable
+			if strings.Contains(err.Error(), `admission webhook "machine-controller.kubermatic.io-machinedeployments" denied the request`) {
+				glog.V(4).Infof("giving up creating initial Node Deployments for cluster %s (%s) due to an unrecoverabl err %#v", cluster.Name, cluster.Spec.HumanReadableName, err)
```
floreks, line 174: Should this error be also recorded and visible in the UI?

alvaroaleman: It is, but from the layer above, see line 151

floreks: Perfect!

PR #3896, opened 2019-07-26
define KubermaticConfiguration CRD

xrstf:
This defines the new CRD used for the Kubermatic Operator. The diff has been extracted from #3850.

QQ: What API group do we want to use? I proposed `operator.kubermatic.io`, but our existing resources are in `kubermatic.k8s.io`, so maybe we should use `operator.kubermatic.k8s.io`? (nit: I don't think `kubermatic.k8s.io` is a great name, but alas, it exists)

Relates to #3626 

```release-note
NONE
```


api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	NodePortStrategy ExposeStrategy = "NodePort"
+	// LoadBalancerStrategy creates a LoadBalancer service per cluster.
+	LoadBalancerStrategy ExposeStrategy = "LoadBalancer"
+)
+
+// DockerImage describes a Docker image.
+type DockerImage struct {
+	Repository string            `json:"repository"`
+	Tag        string            `json:"tag"`
+	PullPolicy corev1.PullPolicy `json:"pullPolicy"`
```
alvaroaleman: Why do we need this? Is there any reason we want to make the `pullPolicy` configurable?

If not, I suggest using `github.com/docker/distribution/reference.Reference` which is already in our codebase and more powerful

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	IsMaster bool `json:"isMaster,omitempty"`
+	// Datacenters is a YAML string containing the available node datacenters. Note that
+	// this is deprecated and you should use explicit Datacenter CRDs instead.
+	Datacenters string `json:"datacenters,omitempty"`
+	// Secrets is a list of predefined credentials, like Docker registry authentication.
+	Secrets KubermaticSecretsConfiguration `json:"secrets,omitempty"`
+	// Auth defines keys and URLs for Dex.
+	Auth KubermaticAuthConfiguration `json:"auth"`
+	// FeatureGates are used to optionally enable certain features.
+	FeatureGates map[string]bool `json:"featureGates,omitempty"`
```
alvaroaleman: Please use a `k8s.io/apimachinery/pkg/util/sets.String` instead

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	MasterController KubermaticMasterControllerConfiguration `json:"masterController,omitempty"`
+	// MasterFiles is a map of additional files to mount into each master component.
+	MasterFiles map[string]string `json:"masterFiles,omitempty"`
+	// ExposeStrategy is the strategy to expose the cluster with.
+	// Note: The `seed_dns_overwrite` setting of the `datacenters.yaml` doesn't have any effect if this is set to LoadBalancerStrategy.
+	ExposeStrategy ExposeStrategy `json:"exposeStrategy,omitempty"`
+}
+
+// EnabledFeatureGates returns a sorted list of feature names that are currently enabled.
+func (spec *KubermaticConfigurationSpec) EnabledFeatureGates() []string {
```
alvaroaleman: Please remove and use `k8s.io/apimachinery/pkg/util/sets.String.List()` instead in the code

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	IssuerClientID           string `json:"issuerClientID,omitempty"`
+	IssuerClientSecret       string `json:"issuerClientSecret,omitempty"`
+	IssuerCookieKey          string `json:"issuerCookieKey,omitempty"`
+	CABundle                 string `json:"cABundle,omitempty"`
+	ServiceAccountKey        string `json:"serviceAccountKey,omitempty"`
+	SkipTokenIssuerTLSVerify bool   `json:"skipTokenIssuerTLSVerify,omitempty"`
+}
+
+// SimpleFeatureGate is a helper for features with no additional options.
+type SimpleFeatureGate struct {
```
alvaroaleman: This is unused

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -0,0 +1,184 @@
...
+	Namespace string `json:"namespace,omitempty"`
+	// Domain is the base domain where the dashboard shall be available.
+	Domain string `json:"domain"`
+	// IsMaster controls whether the dashboard and API will be deployed.
+	IsMaster bool `json:"isMaster,omitempty"`
+	// Datacenters is a YAML string containing the available node datacenters. Note that
+	// this is deprecated and you should use explicit Datacenter CRDs instead.
+	Datacenters string `json:"datacenters,omitempty"`
+	// Secrets is a list of predefined credentials, like Docker registry authentication.
+	Secrets KubermaticSecretsConfiguration `json:"secrets,omitempty"`
```
alvaroaleman: Can we just directly have the `ImagePullSecret` here with type `corev1.ObjecReference`?

alvaroaleman (commented):
Some nits, other than that looks good!

api/pkg/crd/operator/v1alpha1/configuration.go
alvaroaleman: Discussed offline, we will just inline the secrts in the object for now

api/pkg/crd/operator/v1alpha1/configuration.go
xrstf: Let's keep this as a string for now and parse it whenever we need it. If parsing becomes something we do often, we can create a custom type with its own unmarshalling logic.

PR #3898, opened 2019-07-26
Kubermatic Noperator

xrstf:
This is a bare bones operator that does nothing except watching KubermaticConfiguration objects.

Relates to #3626 

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/operator-master/reconciler.go
+++ b/api/pkg/controller/operator-master/reconciler.go
@@ -22,18 +22,21 @@ type Reconciler struct {
 	log          *zap.SugaredLogger
 	recorder     record.EventRecorder
 	workerName   string
+	ctx          context.Context
 }
 
 // Reconcile acts upon requests and will restore the state of resources
 // for the given namespace. Will return an error if any API operation
 // failed, otherwise will return an empty dummy Result struct.
 func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	ctx, cancel := context.WithCancel(context.Background())
+	var cancel context.CancelFunc
+
+	r.ctx, cancel = context.WithCancel(context.Background())
 	defer cancel()
 
 	// find the requested configuration
 	config := &operatorv1alpha1.KubermaticConfiguration{}
-	if err := r.Get(ctx, request.NamespacedName, config); err != nil {
+	if err := r.Get(r.ctx, request.NamespacedName, config); err != nil {
 		return reconcile.Result{}, fmt.Errorf("could not get KubermaticConfiguration %q: %v", request, err)
 	}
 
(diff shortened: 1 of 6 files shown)


api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,59 @@
...
+	log          *zap.SugaredLogger
+	recorder     record.EventRecorder
+	workerName   string
+}
+
+// Reconcile acts upon requests and will restore the state of resources
+// for the given namespace. Will return an error if any API operation
+// failed, otherwise will return an empty dummy Result struct.
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
alvaroaleman: IMHO this pattern of creating a context per request doesn't really make sense, bceause the context only gets canceled after the request was done in which case the canceled context doesn't affect the request anyways. Hence IMHO it makes more sense to store the context on the Reconciler. Thats just a nit thought, putting a hold, feel free to remove it if you dont care.

alvaroaleman (commented):
/lgtm
/approve
/hold

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,59 @@
...
+	defer cancel()
+
+	// find the requested configuration
+	config := &operatorv1alpha1.KubermaticConfiguration{}
+	if err := r.Get(ctx, request.NamespacedName, config); err != nil {
+		return reconcile.Result{}, fmt.Errorf("could not get KubermaticConfiguration %q: %v", request, err)
+	}
+
+	// silently ignore other worker names
+	if labels := config.GetLabels(); labels[WorkerNameLabel] != r.workerName {
```
alvaroaleman: you can btw also just do a `if conifg.Labels[WorkerNameLabel] != r. workerName`, accessing a `map[string]string` for reading is safe, even when its nil

alvaroaleman: These getter funcs like `GetLabels` exist, because `metav1.ObjectMeta` implements the `k8s.io/apimachinery/pkg/apis/meta/v1.Object` interface. There is no reason to use them thought if you know you have an ObjectMeta

api/pkg/controller/operator-master/reconciler.go
```diff
@@ -0,0 +1,59 @@
...
+	"go.uber.org/zap"
+	"k8s.io/client-go/tools/cache"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/tools/record"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Reconciler (re)stores all components required for proxying requests to
+// seed clusters. It also takes care of creating a nice ConfigMap for
```
alvaroaleman: I don't think it does that :P

api/pkg/controller/operator-master/reconciler.go
xrstf: I don't think putting the context anywhere at all is really helping here. If nobody from the outside can cancel it in a meaningful way anway..

alvaroaleman: I agree, but putting it on the `Reconciler` means you dont have to pass it around all the time

PR #3905, opened 2019-07-29
Kubermatic Operator deployment

xrstf:
This first and foremost removes the need to load a kubeconfig into the Kubermatic Operator. In the new world, Datacenters and Seeds will be linking to their kubeconfigs, which already exist as secrets. The operator therefore can easily find all the kubeconfigs it could ever need.

Apart from this, this PR adds static manifests and a new stack, `kubermatic-operator`, to the `deploy.sh` script. Unfortunately in this setup, the deployment will go into an ImagePull backoff until the `kubermatic` stack is done building and pushing the Docker images. We can solve this by simply doing the operator deployment as part of the `kubermatic` stack (I separated it because I wanted to be able to send prow errors to a custom Slack channel).

Relates to #3626 

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-operator/main.go
+++ b/api/cmd/kubermatic-operator/main.go
@@ -30,7 +30,7 @@ type controllerRunOptions struct {
 
 func main() {
 	opt := &controllerRunOptions{}
-	flag.StringVar(&opt.kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required out-of-cluster.")
+	flag.StringVar(&opt.kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if outside of cluster.")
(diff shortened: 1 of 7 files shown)


api/cmd/kubermatic-operator/main.go
```diff
@@ -30,7 +30,7 @@ type controllerRunOptions struct {
 
 func main() {
 	opt := &controllerRunOptions{}
-	flag.StringVar(&opt.kubeconfig, "kubeconfig", "", "Path to a kubeconfig.")
+	flag.StringVar(&opt.kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required out-of-cluster.")
```
alvaroaleman: `if out of cluster`

PR #3916, opened 2019-07-31
Split cluster deletion into multiple functions for readability

mrIncompetent:

Initially i just wanted to introduce structured logging here...
But adding the logging code made this code way more difficult to read than it already was.
Thus i decided to split the individual deletion logics into dedicated funcs to improve readability.
Also it adds some comments trying to explain some non-obvious behaviour.

The code does not change the behaviour of the deletion.

```release-note
NONE
```

/assign @moadqassem

### DIFF
--- a/api/pkg/clusterdeletion/lb.go
+++ b/api/pkg/clusterdeletion/lb.go
@@ -115,17 +115,10 @@ func (d *Deletion) checkIfAllLoadbalancersAreGone(ctx context.Context, cluster *
 	}
 
 	if err := d.updateCluster(ctx, cluster, func(cluster *kubermaticv1.Cluster) {
-		if deletedLoadBalancers.Len() > 0 {
-			cluster.Annotations[deletedLBAnnotationName] = strings.Join(deletedLoadBalancers.List(), ",")
-		} else {
-			delete(cluster.Annotations, deletedLBAnnotationName)
-		}
+		cluster.Annotations[deletedLBAnnotationName] = strings.Join(deletedLoadBalancers.List(), ",")
 	}); err != nil {
 		return false, fmt.Errorf("failed to update cluster: %v", err)
 	}
-	if deletedLoadBalancers.Len() > 0 {
-		return false, nil
-	}
 
-	return true, nil
+	return deletedLoadBalancers.Len() > 0, nil
 }
--- a/api/pkg/clusterdeletion/lb.go
+++ b/api/pkg/clusterdeletion/lb.go
@@ -16,6 +16,10 @@ import (
 	controllerruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	eventReasonDeletedLoadBalancer = "DeletedLoadBalancer"
+)
+
 func (d *Deletion) cleanupLBs(ctx context.Context, cluster *kubermaticv1.Cluster) (deletedSomeLBs bool, err error) {
 	serviceList := &corev1.ServiceList{}
 	if err := d.userClusterClient.List(ctx, &controllerruntimeclient.ListOptions{}, serviceList); err != nil {
@@ -104,7 +108,7 @@ func (d *Deletion) checkIfAllLoadbalancersAreGone(ctx context.Context, cluster *
 			return false, fmt.Errorf("failed to get service events: %v", err)
 		}
 		for _, event := range events.Items {
-			if event.Reason == "DeletedLoadBalancer" {
+			if event.Reason == eventReasonDeletedLoadBalancer {
 				deletedLoadBalancers.Delete(deletedLB)
 			}
 		}
--- a/api/pkg/clusterdeletion/lb.go
+++ b/api/pkg/clusterdeletion/lb.go
@@ -1,0 +1,127 @@
+package clusterdeletion
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	controllerruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (d *Deletion) cleanupLBs(ctx context.Context, cluster *kubermaticv1.Cluster) (deletedSomeLBs bool, err error) {
+	serviceList := &corev1.ServiceList{}
+	if err := d.userClusterClient.List(ctx, &controllerruntimeclient.ListOptions{}, serviceList); err != nil {
+		return false, fmt.Errorf("failed to list Service's from user cluster: %v", err)
+	}
+
+	for _, service := range serviceList.Items {
+		// Only LoadBalancer services create cost on cloud providers
+		if service.Spec.Type != corev1.ServiceTypeLoadBalancer {
+			continue
+		}
+
+		serviceName := fmt.Sprintf("%s/%s", service.Namespace, service.Name)
+		if err := d.cleanupLB(ctx, &service, cluster); err != nil {
+			return deletedSomeLBs, fmt.Errorf("failed to delete service %q inside user cluster: %v", serviceName, err)
+		}
+		deletedSomeLBs = true
+	}
+
+	return deletedSomeLBs, nil
+}
+
+func (d *Deletion) cleanupLB(ctx context.Context, service *corev1.Service, cluster *kubermaticv1.Cluster) error {
+	if err := d.userClusterClient.Delete(ctx, service); err != nil {
+		return fmt.Errorf("failed to delete service: %v", err)
+	}
+
+	// We store the deleted service UID's on the cluster so we can check on the next iteration if they are really gone.
+	// We only really know if a LoadBalancer(The cloud provider LB) is gone, until there has been an event stating that.
+	// The event contains the UID of the corresponding, deleted, service.
+	// Upstream changed that recently to use a finalizer - As soon as we only support Kubernetes versions above that, we can remove this
+	err := d.updateCluster(ctx, cluster, func(cluster *kubermaticv1.Cluster) {
+		if cluster.Annotations == nil {
+			cluster.Annotations = map[string]string{}
+		}
+		cluster.Annotations[deletedLBAnnotationName] = cluster.Annotations[deletedLBAnnotationName] + fmt.Sprintf(",%s", string(service.UID))
+	})
+	if err != nil {
+		return fmt.Errorf("failed to update cluster when trying to add UID of deleted LoadBalancer: %v", err)
+	}
+
+	// Wait for the update to appear in the lister as we use the data from the lister later on to verify if the LoadBalancers
+	// are gone
+	if err := wait.Poll(10*time.Millisecond, 5*time.Second, func() (bool, error) {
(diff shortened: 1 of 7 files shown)


api/pkg/clusterdeletion/lb.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	for _, service := range serviceList.Items {
+		// Only LoadBalancer services create cost on cloud providers
+		if service.Spec.Type != corev1.ServiceTypeLoadBalancer {
+			continue
+		}
+
+		serviceName := fmt.Sprintf("%s/%s", service.Namespace, service.Name)
+		if err := d.cleanupLB(ctx, &service, cluster); err != nil {
```
moadqassem, line 36: If the deletion of the LBs fails, will the cluster controller try to delete them later in a retry mechanism?

mrIncompetent: Yes, in case anything fails in this deletion process, the error will get passed up to the controller.
The controller will then handle the error by just requeuing the cluster.

As the `clusterdeletion` package is not below a controller package, i would add a comment to `CleanupCluster` to express that:
- `CleanupCluster` must be retried until all finalizers on the cluster object are gone.

@moadqassem WDYT?

moadqassem: Well, the reason why I am asking is, I saw that the deleted LBs are also listed somewhere in a map, where you check if they were deleted or not. Since you have this map in addition to the ability to get the LBs to be deleted in the cleanup method, you can start deleting those LBs asynchronously without having the feedback to the cluster controller directly by returning an error. The cluster controller can communicate with the deletion process via the context. 

Disclaimer: This is just an idea, not a change requirement.

api/pkg/clusterdeletion/lb.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	for deletedLB := range deletedLoadBalancers {
+		selector := fields.OneTermEqualSelector("involvedObject.uid", deletedLB)
+		events := &corev1.EventList{}
+		if err := d.userClusterClient.List(context.Background(), &controllerruntimeclient.ListOptions{FieldSelector: selector}, events); err != nil {
+			return false, fmt.Errorf("failed to get service events: %v", err)
+		}
+		for _, event := range events.Items {
+			if event.Reason == "DeletedLoadBalancer" {
```
moadqassem: Would be good to have `DeletedLoadBalancer` as a const for any refactoring or reuse in the future

api/pkg/clusterdeletion/lb.go
```diff
@@ -0,0 +1,127 @@
...
+	if err := d.updateCluster(ctx, cluster, func(cluster *kubermaticv1.Cluster) {
+		if deletedLoadBalancers.Len() > 0 {
+			cluster.Annotations[deletedLBAnnotationName] = strings.Join(deletedLoadBalancers.List(), ",")
+		} else {
+			delete(cluster.Annotations, deletedLBAnnotationName)
+		}
+	}); err != nil {
+		return false, fmt.Errorf("failed to update cluster: %v", err)
+	}
+	if deletedLoadBalancers.Len() > 0 {
```
moadqassem: The `modify` callback function can have an error in it's signature to handle any errors during the update: 
https://github.com/kubermatic/kubermatic/pull/3916/files#diff-4ff2a1fec00fb94e91ee96d3ad36b9c9R116
```
modify func(*kubermaticv1.Cluster) error 
// Apply modifications
if err := modify(cluster); err != nil {
   return err
}
```
Then you can simply check for err of the `updateCluster` instead of doing check on an interpreted variable and avoid any side effects

api/pkg/clusterdeletion/node.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("failed to list MachineDeployments: %v", err)
+	}
+	if len(machineDeploymentList.Items) > 0 {
+		// TODO: Use DeleteCollection once https://github.com/kubernetes-sigs/controller-runtime/issues/344 is resolved
+		for _, machineDeployment := range machineDeploymentList.Items {
+			if err := d.userClusterClient.Delete(ctx, &machineDeployment); err != nil {
+				return fmt.Errorf("failed to delete MachineDeployment %q: %v", machineDeployment.Name, err)
+			}
+		}
+		// Return here to make sure we don't attempt to delete MachineSets until the MachineDeployment is actually gone
```
moadqassem, line 65: I would assume that this method is gonna be called on multiple times and based on the machine Hierarchy, first deployment, machine sets, and machines have to be deleted respectively. Can't we instead of returning nil on each deletion process, just start a loop and delete each object respectively and asynchronously and have some waitgroups, to make sure we don't delete an object that shouldn't be deleted before another object(eg: don't delete MachineSet if MachineDeployment is not deleted)?

https://gist.github.com/moadqassem/e3966f14d07219f4e231aec27ede20ad

api/pkg/clusterdeletion/secret.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Secret %s/%s: %v", resources.KubermaticNamespace, secretName, err)
+	}
+
+	if err == nil {
+		if err := d.seedClient.Delete(ctx, secret); err != nil {
+			return fmt.Errorf("failed to delete Secret '%s/%s': %v", secret.Namespace, secret.Name, err)
+		}
+	}
+	return nil
```
moadqassem, line 56: What if the `err` was not `nil` and `!kerrors.IsNotFound(err)` returned false, then the error will be ignored. Is this desired by design?

api/pkg/clusterdeletion/volumes.go
```diff
@@ -0,0 +1,171 @@
...
+func (d *Deletion) cleanupPDBs(ctx context.Context, cluster *kubermaticv1.Cluster) (deletedPDBs bool, err error) {
+	pdbs := &policyv1beta1.PodDisruptionBudgetList{}
+	if err := d.userClusterClient.List(ctx, &controllerruntimeclient.ListOptions{}, pdbs); err != nil {
+		return false, fmt.Errorf("failed to list pdbs: %v", err)
+	}
+	for _, pdb := range pdbs.Items {
+		if err := d.userClusterClient.Delete(ctx, &pdb); err != nil {
+			return false, fmt.Errorf("failed to delete pdb '%s/%s': %v", pdb.Namespace, pdb.Name, err)
+		}
+		deletedPDBs = true
```
moadqassem, line 142: Why this is needed? if the cluster has no Pod Disruption Budget List, then it should be considered as true returned value, thus return `true` at line 138

api/pkg/clusterdeletion/lb.go
mrIncompetent: While i like the idea, i would like to keep functional changes out of this PR - I'm really just focusing readability

api/pkg/clusterdeletion/volumes.go
```diff
@@ -0,0 +1,171 @@
...
+		return false, fmt.Errorf("failed to disable future PV & PVC creation: %v", err)
+	}
+
+	deletedSomePDBs, err := d.cleanupPDBs(ctx, cluster)
+	if err != nil {
+		return false, fmt.Errorf("failed to cleanup PodDisruptionBudgets: %v", err)
+	}
+	// Make sure we don't continue until all PDBs are actually gone.
+	// TODO: Why?
+	if deletedSomePDBs {
```
moadqassem, line 37: Any idea what's going on here? why the code is exited after the PDBs are deleted? So what difference of PDBs are not existed and deleted? The conclusion should be the same I guess.

api/pkg/clusterdeletion/volumes.go
```diff
@@ -0,0 +1,171 @@
...
+
+	pvUsingPods := []*corev1.Pod{}
+	for _, pod := range podList.Items {
+		if podUsesPV(&pod) {
+			pvUsingPods = append(pvUsingPods, &pod)
+		}
+	}
+
+	for _, pod := range pvUsingPods {
+		namespace, name := pod.Namespace, pod.Name
```
moadqassem: why? just use the `pod.Namespace` and `pod.Name` please.

api/pkg/clusterdeletion/volumes.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func terminatingAdmissionWebhook() (string, *admissionregistrationv1beta1.ValidatingWebhookConfiguration) {
+	name := "kubernetes-cluster-cleanup"
+	failurePolicy := admissionregistrationv1beta1.Fail
+	return name, &admissionregistrationv1beta1.ValidatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Annotations: map[string]string{
+				"description": "This webhok configuration exists to prevent creation of any new stateful resources in a cluster that is currently being terminated",
```
moadqassem: typo s/webhok/webhook would also be nice to have `description` value as a const

api/pkg/clusterdeletion/node.go
mrIncompetent: I really don't want to change fundamental functionality in this PR

api/pkg/clusterdeletion/secret.go
mrIncompetent: The error handling wasn't that clear, i hope its more understandable with https://github.com/kubermatic/kubermatic/pull/3916/commits/ba85a1a4d585a71a42ae98ea3079bb3997613392

api/pkg/clusterdeletion/lb.go
mrIncompetent: Added in https://github.com/kubermatic/kubermatic/pull/3916/commits/57a14b0474ce83715dd0332107a43d88781a00aa

api/pkg/clusterdeletion/volumes.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3916/commits/1de41c8b6d39c88329c17a59098d127fef823b3b

api/pkg/clusterdeletion/volumes.go
mrIncompetent: I have no idea - I've added a TODO for the whole PDB handling.

We don't know why we delete PDBs at all & i don't want include the removal here

api/pkg/clusterdeletion/volumes.go
mrIncompetent: https://github.com/kubermatic/kubermatic/pull/3916#discussion_r309643815

api/pkg/clusterdeletion/lb.go
mrIncompetent: I've simplified the annotation handling: https://github.com/kubermatic/kubermatic/pull/3916/commits/2bead559e5036b8d8c42f04d59a2272e584f19d0

Changing the `modify` signature is something i would like to avoid - Is a function which gets retried & should only modify the cluster struct. It should never trigger an error

api/pkg/clusterdeletion/volumes.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3916/commits/87c0004a423af594e9095376e41ec31684cf0362

api/pkg/clusterdeletion/lb.go
moadqassem: Well, it's not about making it simple tbh, it's about the update of the cluster is already triggering an error if something is not expected. Anyway, I don't have a strong opinion on it.

api/pkg/clusterdeletion/volumes.go
moadqassem: Ok, since I don't know why this is implemented this way and the system is working as expected, we can leave it as it is.

PR #3921, opened 2019-08-01
add permissions for 1.14 kubeadm join configmap

kdomanski:
Currently kubeadm join with Kubernetes 1.14 doesn't work, because the necessary configmap is created, but it's inaccessible for the kubelets.

```release-note
NONE
```
/assign @alvaroaleman 
/cherry-pick release/v2.11
/cherry-pick release/v2.10


kdomanski:
/hold
Let's sort out the image versioning first.

PR #3923, opened 2019-08-01
Improve fatal message when creating the seed factory fails

mrIncompetent:
Improve fatal message when creating the seed factory fails.

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-controller-manager/main.go
+++ b/api/cmd/kubermatic-controller-manager/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/go-logr/zapr"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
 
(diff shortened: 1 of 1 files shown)


api/cmd/kubermatic-controller-manager/main.go
```diff
@@ -99,7 +99,7 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 	rootCtx, rootCancel := context.WithCancel(context.Background())
 	seedGetter, err := provider.SeedGetterFactory(rootCtx, mgr.GetClient(), options.dc, options.dcFile, options.dynamicDatacenters)
 	if err != nil {
-		log.Fatal(err)
```
alvaroaleman, line 102: Nit: Wouldn't it be nicer if this was a `log.With("error", err).Fatal(...)`?

mrIncompetent: Oh great idea, i'll update all fatals here with that

PR #3917, opened 2019-07-31
Use CRD datacenters during e2e tests

alvaroaleman:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
none
```

PR #3927, opened 2019-08-01
Dynamically restart the rbac controller on seed config changes

alvaroaleman:

Adds a new `seed-controller-lifecycle` controller (name can be improved) that hashes the current seed config and whenever it changes, restarts runnables constructed via a runnable factory.

This is needed to be able to react to changes in the seed config.

Fixes #

**Documentation**:

```release-note
none
```

/assign @mrIncompetent

### DIFF
--- a/api/cmd/master-controller-manager/controllers.go
+++ b/api/cmd/master-controller-manager/controllers.go
@@ -5,11 +5,15 @@ import (
 	"time"
 
 	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac"
+	seedcontrollerlifecycle "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-lifecycle"
 	seedproxy "github.com/kubermatic/kubermatic/api/pkg/controller/seed-proxy"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/service-account"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-project-binding"
 	kubermaticclientset "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned"
 	"github.com/kubermatic/kubermatic/api/pkg/crd/client/informers/externalversions"
+	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
+	"github.com/prometheus/client_golang/prometheus"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -18,8 +22,21 @@ import (
 )
 
 func createAllControllers(ctrlCtx *controllerContext) error {
-	if err := createRBACContoller(ctrlCtx); err != nil {
-		return fmt.Errorf("failed to create rbac controller: %v", err)
+	rbacControllerFactory := rbacControllerFactoryCreator(
+		ctrlCtx.mgr.GetConfig(),
+		ctrlCtx.seedsGetter,
+		ctrlCtx.seedKubeconfigGetter,
+		ctrlCtx.workerCount,
+		ctrlCtx.labelSelectorFunc)
+
+	if err := seedcontrollerlifecycle.Add(ctrlCtx.ctx,
+		kubermaticlog.Logger,
+		ctrlCtx.mgr,
+		ctrlCtx.seedsGetter,
+		ctrlCtx.seedKubeconfigGetter,
+		rbacControllerFactory); err != nil {
+		//TODO: Find a better name
+		return fmt.Errorf("failed to create seedcontrollerlifecycle: %v", err)
 	}
 	if err := userprojectbinding.Add(ctrlCtx.mgr); err != nil {
 		return fmt.Errorf("failed to create userprojectbinding controller: %v", err)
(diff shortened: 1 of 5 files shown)


api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -0,0 +1,203 @@
...
+	err := ci.controllers.Start(stopCh)
+	if err != nil {
+		ci.log.With("error", err).Error("controllers stopped with error")
+	} else {
+		ci.log.Info("controllers stopped")
+	}
+	ci.running = false
+
+	// Make sure a reconciliation happens
+	ci.enqueue()
```
alvaroaleman: The whole concept of injecting the `enqueue` func feels a bit weird, we basically create something that can then enqueue itself. But I couldn't think of a better way for this, any ideas?

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	controllers, err := r.controllerFactory()
+	if err != nil {
+		return fmt.Errorf("failed to construct controllers: %v", err)
+	}
+
+	if r.activeControllers != nil && r.activeControllers.running {
+		log.Info("Stopping old version of controllers")
+		close(r.activeControllers.stopChan)
```
alvaroaleman: I must admit I am a bit afraid that we may leak goroutines here and that I do not 100% trust the underlying informerFactories to properly kill off everything.

@moadqassem suggested to just use a different binary instead, WDYT?

api/cmd/master-controller-manager/controllers.go
```diff
@@ -33,47 +50,60 @@ func createAllControllers(ctrlCtx *controllerContext) error {
...
 
-	ctrl, err := rbac.New(rbac.NewMetrics(), allClusterProviders, ctrlCtx.workerCount)
-	if err != nil {
-		return err
-	}
+		for seedName := range seeds {
+			kubeConfig, err := seedKubeconfigGetter(seedName)
+			if err != nil {
+				kubermaticlog.Logger.With("error", err).With("seed", seedName).Error("error getting kubeconfig")
+				// Dont let a single broken kubeconfig break the whole controller creation
```
mrIncompetent, line 79: How about adding a metric we can alert on?
Like a broken seed counter

api/cmd/master-controller-manager/controllers.go
```diff
@@ -33,47 +50,60 @@ func createAllControllers(ctrlCtx *controllerContext) error {
...
+		allClusterProviders := []*rbac.ClusterProvider{masterClusterProvider}
 
-	ctrl, err := rbac.New(rbac.NewMetrics(), allClusterProviders, ctrlCtx.workerCount)
-	if err != nil {
-		return err
-	}
+		for seedName := range seeds {
+			kubeConfig, err := seedKubeconfigGetter(seedName)
+			if err != nil {
+				kubermaticlog.Logger.With("error", err).With("seed", seedName).Error("error getting kubeconfig")
```
mrIncompetent, line 78: nit: `zap.Error(err)` could ensure we never end up with different keys for `error`

api/cmd/master-controller-manager/controllers.go
```diff
@@ -33,47 +50,60 @@ func createAllControllers(ctrlCtx *controllerContext) error {
 	return nil
 }
 
-func createRBACContoller(ctrlCtx *controllerContext) error {
-	masterClusterProvider, err := rbacClusterProvider(ctrlCtx.mgr.GetConfig(), "master", true, ctrlCtx.labelSelectorFunc)
-	if err != nil {
-		return fmt.Errorf("failed to create master rbac provider: %v", err)
-	}
-	allClusterProviders := []*rbac.ClusterProvider{masterClusterProvider}
+func rbacControllerFactoryCreator(
```
mrIncompetent, line 53: It looks like this could actually be moved into the `seed_controller_lifecycle` as `NewRBACController(...)` ?

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -0,0 +1,203 @@
...
+	controllerFactory    func() (manager.Runnable, error)
+	enqueuer             *enqueuer
+	activeControllers    *controllersInstance
+}
+
+// controllersInstance represents an instance of a set of running controllers
+type controllersInstance struct {
+	log         *zap.SugaredLogger
+	configHash  string
+	controllers manager.Runnable
```
mrIncompetent: Why is this plural when its just a single `Runnable` ?

mrIncompetent: That applies for all the `controllers` occurences

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -0,0 +1,203 @@
...
+		seedKubeconfigMap[seedName] = *cfg
+	}
+
+	configHash, err := r.seedKubeconfigHashStable(seedKubeconfigMap)
+	if err != nil {
+		return fmt.Errorf("failed to calculate hash for seed kubeconfigs: %v", err)
+	}
+	log := r.log.With("config_hash", configHash)
+
+	if r.activeControllers != nil && r.activeControllers.configHash == configHash {
```
mrIncompetent: I think you could as well, store the map & perform a deepEqual.
That way you could get rid of the hashing func

api/cmd/master-controller-manager/controllers.go
mrIncompetent: Why is this function called "Factory" & "Creator" ?

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
mrIncompetent: If i read the code correctly the goal is to trigger a reconciling of `Reconciler` when `controllersInstance` stops?

This looks like it could be simpler/cleaner by having a channel.
- Define a channel on the reconciler
- Watch that channel using `source.Channel{}`
- Start the controllerInstance with a callback:
```go
go func() {
	r.activeControllers.Start(r.activeControllers.stopChan)
	r.mySuperAwesomeChannel <- "foo"
}()
```

api/cmd/master-controller-manager/controllers.go
alvaroaleman: Because it creates a factory

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
alvaroaleman: Yeah because initially I assumed it would be more than one controller, I've renamed that

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
alvaroaleman: Interestingly enough that works, I did assume that would create issues because the functions that are embedded in the `rest.Config` are pointers that differ (the address, not the value).

alvaroaleman:
@mrIncompetent PTAL, I think this works now.

Re metrics, I've looked into that but it will require some more changes as the master controller manager currently only exposes the metrics from the manager, all our own metrics that use the `prometheus.DefaultRegistry` are now shown at all. Hence I want to do that in a follow-up.

alvaroaleman:
Also, the amount of events that trigger this controller can be greatly optimized by filtering the Seeds and Secrets to our own namespace. The plumbing to pass the namespace around however was mostly done in #3917 already, copying it here will cause merge conflicts so I will just add that later

api/cmd/master-controller-manager/controllers.go
alvaroaleman: I didn't move it into the `seed_controller_lifecycle` because that would imply I have to pass through all the args needed for the rbac controller. This way I can just pass the func in and use it without any args

mrIncompetent: > Because it creates a factory

If it creates a factory it might be easier to understand if we name it `newControllerFactory` (or even `controllerFactory` , which might be "more" idiomatic go).
Calling something "creator" seems more appropriate for type definitions, not really functions.

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("failed to create watch for channelSource: %v", err)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) Reconcile(_ reconcile.Request) (reconcile.Result, error) {
+	err := r.reconcile()
+	if err != nil {
+		r.log.With("error", err).Error("reconciliation failed")
```
mrIncompetent: nit: `zap.Error(err)`

alvaroaleman:
@mrIncompetent updated

mrIncompetent:
/lgtm
/approve

/hold 
in case you wan't to wait for @moadqassem feedback

PR #3933, opened 2019-08-02
Start to use zap in the cluster controller

mrIncompetent:
This PR introduces zap in the cluster controller.
As migrating everything will create a huge PR - lets go with tiny PR's.

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -195,7 +195,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		log.Debugw(
 			"Skipping because the cluster has a different worker name set",
 			"cluster-worker-name", cluster.Labels[kubermaticv1.WorkerNameLabelKey],
-			"my-worker-name", r.workerName,
+			"worker-name", r.workerName,
 		)
 		return reconcile.Result{}, nil
 	}
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -195,7 +195,6 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		log.Debugw(
 			"Skipping because the cluster has a different worker name set",
 			"cluster-worker-name", cluster.Labels[kubermaticv1.WorkerNameLabelKey],
-			"worker-name", r.workerName,
 		)
 		return reconcile.Result{}, nil
 	}
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/golang/glog"
+	"go.uber.org/zap"
 
 	k8cuserclusterclient "github.com/kubermatic/kubermatic/api/pkg/cluster/client"
 	"github.com/kubermatic/kubermatic/api/pkg/clusterdeletion"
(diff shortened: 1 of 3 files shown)


api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -170,33 +173,41 @@ func Add(
 func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	log := r.log.With("request", request)
```
alvaroaleman, line 176: I must say I don't like putting the request inside, because for cluster scoped-resources it will contain `/cluster-id`. Or maybe we can put both the request and the  `ClusterName` as fields?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -170,33 +173,41 @@ func Add(
...
+	log.Debug("Processing")
 
 	cluster := &kubermaticv1.Cluster{}
 	if err := r.Get(ctx, request.NamespacedName, cluster); err != nil {
 		if kubeapierrors.IsNotFound(err) {
-			glog.V(4).Infof("Couldn't find cluster %q", request.NamespacedName.String())
+			log.Info("Could not find cluster")
```
alvaroaleman: IMHo thats debug log

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -170,33 +173,41 @@ func Add(
...
-		glog.V(4).Infof("skipping paused cluster %s", cluster.Name)
+		log.Debug("Skipping because the cluster is paused")
 		return reconcile.Result{}, nil
 	}
 
 	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
+		log.Debugw(
+			"Skipping because the cluster has a different worker name set",
+			"cluster-worker-name", cluster.Labels[kubermaticv1.WorkerNameLabelKey],
+			"my-worker-name", r.workerName,
```
alvaroaleman: Why `my-worker-name` and not just `worker-name`? We could btw add the worker name to the log already when we initialize the log in `Add` as it will never change

alvaroaleman: Probably even at the start of the binary

mrIncompetent: > Why my-worker-name and not just worker-name

mrIncompetent: > Why my-worker-name and not just worker-name ?

Renamed with https://github.com/kubermatic/kubermatic/pull/3933/commits/14af09178a61f93b569cf262929046f5d55c12e4

> We could btw add the worker name to the log already when we initialize the log in Add as it will never change

I though about this as well, but was not that sure if its useful to have it always.
It might become noise? But happy to change if you prefer

alvaroaleman: I _think_ there is no harm in an additional field ppl don't care about, except for a bit more indexing load on Elasticsearch? The advantage is that its always added in all controllers then

mrIncompetent: Added the field in https://github.com/kubermatic/kubermatic/pull/3933/commits/96cbb2f964ae6e3067d53f49fac61caf51510361

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3933/commits/496d52b6020bcc0b22ba025655ecc65c88dccef1

api/pkg/controller/cluster/cluster_controller.go
mrIncompetent: > Or maybe we can put both the request and the ClusterName as fields?

Done with https://github.com/kubermatic/kubermatic/pull/3933/commits/496d52b6020bcc0b22ba025655ecc65c88dccef1

PR #3939, opened 2019-08-02
Update azure dependency

mrIncompetent:
Update azure dependency. Now we use the same version as the machine-controller.

```release-note
NONE
```

/assign @moadqassem


mrIncompetent:
Seems i broke the vendor folder...

PR #3941, opened 2019-08-02
Prometheus metrics and alerts for addons

bashofmann:
Adds a Prometheus metrics collector for addons with the new metrics:
* kubermatic_addon_created
* kubermatic_addon_deleted

Adds an alert KubermaticAddonDeletionTakesTooLong which fires if an addon is stuck in deletion for more than 30 minutes.

The helps identifying problems when an addon fails to be deleted because of an error in the controller manager when cleaning up the addons manifests.

```release-note
New Prometheus metrics kubermatic_addon_created and kubermatic_addon_deleted
New alert KubermaticAddonDeletionTakesTooLong
```

### DIFF
--- a/api/pkg/collectors/addon.go
+++ b/api/pkg/collectors/addon.go
@@ -1,0 +1,93 @@
+package collectors
+
+import (
+	"context"
+	"fmt"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/prometheus/client_golang/prometheus"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	addonPrefix = "kubermatic_addon_"
+)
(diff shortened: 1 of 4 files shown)


xmudrii:
I'll leave it up to @xrstf for review :slightly_smiling_face: 

/unassign
/assign @xrstf

themue (commented):
/lgtm

Would like @xrstf to make approval.

bashofmann:
Since I can't see the test results: are they because of flaky CI? Locally at least the unit tests run through for me.

api/pkg/collectors/addon.go
```diff
@@ -0,0 +1,93 @@
+package collectors
```
alvaroaleman, line 1: @bashofmann Just for my understanding, the only reason we implement a collector ourselves is because Prometheus `GaugeVec` doesn't have a way to drop all metrics except the ones that we want to keep, right?

bashofmann: I don't know unfortunately, I just implemented it the same way the already existing cluster metrics are implemented.

PR #3942, opened 2019-08-02
Add kubevirt support to kubermatic

### LINKED ISSUE #3241: Add Kubevirt support to Kubermatic
**User Story**

As a user of Kubermatic I want to be able to use Kubevirt as backend for my Kubernetes clusters

**Acceptance criteria**

* It is possible to configure a Kubevirt node datacenter
* The conformance-tester binary can test Kubevirt clusters and they pass
* The kubevirt nodes use a PV-backed root disk
* It is possible to choose Kubevirt as provider from the UI

moadqassem:
Adding Kubevirt as Cloud Provider in Kubematic

**Which issue(s) this PR fixes** *
Fixes #3241

**Documentation**:

No
-->
```release-note
NONE
```


api/Gopkg.toml
```diff
@@ -121,11 +121,15 @@ required = [
...
+  revision = "4b1149e9e7805fbe850d0b829439903e8604850a"
+
+[[override]]
+  name = "kubevirt.io/containerized-data-importer"
+  source = "github.com/kubevirt/containerized-data-importer"
 
 [[override]]
   name = "sigs.k8s.io/controller-runtime"
-  version = "v0.1.10"
+  version = "v0.1.11"
```
alvaroaleman: Please do not change this

api/pkg/api/v1/types.go
```diff
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 
 	"github.com/Masterminds/semver"
-
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
```
alvaroaleman: Please keep our imports in their own group

api/pkg/api/v1/types.go
```diff
@@ -624,6 +630,17 @@ func newPublicGCPCloudSpec(internal *kubermaticv1.GCPCloudSpec) (public *PublicG
 	return &PublicGCPCloudSpec{}
 }
 
+// PublicKubevirtCloudSpec is a public counterpart of apiv1.KubevirtCloudSpec.
+type PublicKubevirtCloudSpec struct{}
+
+func newPublicKubevirtCloudSpec(internal *kubermaticv1.KubevirtCloudSpec) (public *PublicKubevirtCloudSpec) {
+	if internal == nil {
```
alvaroaleman, line 759: What purpose does this nil handling serve? funcs shouldn't have to verify their args and calles of this func expect a CloudSpec back

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-
 	"github.com/kubermatic/kubermatic/api/pkg/semver"
```
alvaroaleman: Please keep the import groups separate

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -364,6 +364,11 @@ type GCPCloudSpec struct {
 	Subnetwork     string `json:"subnetwork"`
 }
 
+// KubevirtCloudSpec specifies the access data to Kubevirt.
+type KubevirtCloudSpec struct {
+	Config string `json:"config"`
```
alvaroaleman: hmm, maybe even "kubeconfig" to make it easier to understand?

api/pkg/handler/routes_v1.go-e
```diff
@@ -0,0 +1,2510 @@
+// Kubermatic API.
```
alvaroaleman: Please remove this file from the PR

api/pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -0,0 +1,54 @@
...
+func (k *kubevirt) DefaultCloudSpec(spec *v1.CloudSpec) error {
+	return nil
+}
+
+func (k *kubevirt) ValidateCloudSpec(spec v1.CloudSpec) error {
+	if spec.Kubevirt.Config == "" {
+		return fmt.Errorf("config cannot be empty")
+	}
+
+	config, err := base64.StdEncoding.DecodeString(spec.Kubevirt.Config)
```
alvaroaleman: So we expect the end user to submit the kubeconfig base64 encoded already?

api/pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = clientcmd.RESTConfigFromKubeConfig(config)
+	if err != nil {
+		return err
+	}
+
+	spec.Kubevirt.Config = fmt.Sprintf("%s", config)
```
alvaroaleman: Why use fmt.Sprintf instead of just assigning `config` to `spec.Kubevirt.Config`?

api/pkg/resources/machine/machinedeployment.go
```diff
@@ -172,40 +179,52 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc *ku
...
 		config.OperatingSystem = providerconfig.OperatingSystemCentOS
 		osExt, err = getCentOSOperatingSystemSpec(nd.Spec.Template)
 		if err != nil {
-			return nil, err
+			return err
 		}
 	default:
-
-		return nil, errors.New("unknown OS")
+		return errors.New("unknown OS")
```
alvaroaleman: Maybe mention the os in the error to help the user?

api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -34,7 +34,7 @@ var (
 const (
 	Name = "machine-controller"
 
-	tag = "v1.5.1"
+	tag = "944bd35baca49ff53346c012ce33c2876a71ec4f"
```
alvaroaleman: Please ceate and use a tagged release of the machine-controller

api/cmd/conformance-tests/main.go
```diff
@@ -113,6 +113,9 @@ type secrets struct {
 		Subnetwork     string
 		Zone           string
 	}
+	Kubevirt struct {
```
alvaroaleman, line 115: Why is this a separate struct and how do we fill it?

api/cmd/conformance-tests/scenarios_kubevirt.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, v := range versions {
+		// Ubuntu
+		scenarios = append(scenarios, &kubevirtScenario{
+			version: v,
+			nodeOsSpec: &apimodels.OperatingSystemSpec{
+				Ubuntu: &apimodels.UbuntuSpec{},
+			},
+		})
+		// CoreOS
+		scenarios = append(scenarios, &kubevirtScenario{
```
alvaroaleman: Kubevirt doesn't support coreos

api/cmd/conformance-tests/scenarios_kubevirt.go
```diff
@@ -0,0 +1,98 @@
...
+	return []apimodels.NodeDeployment{
+		{
+			Spec: &apimodels.NodeDeploymentSpec{
+				Replicas: utilpointer.Int32Ptr(int32(num)),
+				Template: &apimodels.NodeSpec{
+					Cloud: &apimodels.NodeCloudSpec{
+						Kubevirt: &apimodels.KubevirtNodeSpec{
+							Memory:           utilpointer.StringPtr("1024M"),
+							Namespace:        utilpointer.StringPtr("kube-system"),
+							SourceURL:        utilpointer.StringPtr("http://10.109.79.210/<< OS_NAME >>.img"),
```
alvaroaleman: This doens't get replaced here so we have to actually put the OS in the `kubevirtScenario`

alvaroaleman (commented):
Some nits, but looks pretty good overall :)

api/pkg/api/v1/types.go
moadqassem: This is a copy of the other functions, so I am following how the other providers are implemented.

api/pkg/provider/cloud/kubevirt/provider.go
moadqassem: Yes

api/cmd/conformance-tests/main.go
moadqassem: I am following the same standards which this file is extending! Please have a look at the defined types of this struct.

api/pkg/provider/cloud/kubevirt/provider.go
moadqassem: This is needed to convert the config to string instead of pointer reference

api/pkg/resources/machine/machinedeployment.go
moadqassem: This is not possible, please take a look at how the switch declares the OS.  The OS is part of an enum which gets assigned to the node deployment beforehand!

api/pkg/handler/routes_v1.go-e
moadqassem: This is a generated file from either the swagger generator or the update-code gen, I will get rid of it.

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: This test scenario has been removed

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: Done

api/Gopkg.toml
moadqassem: Done

api/pkg/api/v1/types.go
moadqassem: Done

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: Done

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: I already thought of that, however, the machine controller uses the same name for the kubeconfig so I left it in as it is.

api/pkg/resources/machinecontroller/deployment.go
moadqassem: AFAIK the latest machine-controller release(`v1.5.1`) doesn't include the VM Templates functionality which we use for the PV. 
https://github.com/kubermatic/machine-controller/blob/v1.5.1/pkg/cloudprovider/provider/kubevirt/provider.go

alvaroaleman: Then please create a new one :)

api/cmd/conformance-tests/scenarios_kubevirt.go
```diff
@@ -0,0 +1,105 @@
...
+func (s *kubevirtScenario) NodeDeployments(num int, _ secrets) []apimodels.NodeDeployment {
+	var sourceURL string
+
+	switch {
+	case s.nodeOsSpec.Ubuntu != nil:
+		sourceURL = "http://10.109.79.210/ubuntu.img"
+	case s.nodeOsSpec.Centos != nil:
+		sourceURL = "http://10.109.79.210/centos.img"
+	default:
+		s.logger.Error("node deployment os is not supported, using ubuntu as a default image")
```
alvaroaleman: Please do include _what_ is not supported :P

api/cmd/conformance-tests/scenarios_kubevirt.go
```diff
@@ -0,0 +1,105 @@
...
+		sourceURL = "http://10.109.79.210/ubuntu.img"
```
alvaroaleman: This needs to go

api/pkg/api/v1/types.go
alvaroaleman: @maciaszczykm @zreigz Can you elaborate on why we do this?

api/pkg/crd/kubermatic/v1/cluster.go
alvaroaleman: True but the machine-controller field names wont be shown in the UI (and also there I'd prefer to use `kubeconfig`, maybe you can update that there as long as no one is using kubevirt yet?)

api/pkg/presets/manager.go
```diff
@@ -185,8 +196,10 @@ func NewWithPresets(presets *Presets) *Manager {
 
 // NewFromFile returns a instance of manager with the credentials loaded from the given paths
 func NewFromFile(credentialsFilename string) (*Manager, error) {
-	var presets *Presets
-	var err error
+	var (
```
alvaroaleman: For the sake of consistency I would btw prefer it if we stayed with not using blocks  to initialize variables inside funcs

api/pkg/provider/cloud/kubevirt/provider.go
alvaroaleman: I think it would make more sense to enter it raw and we base64 encode it in the api

api/pkg/provider/cloud/kubevirt/provider.go
alvaroaleman: so use `*config` to dereference the pointer?

api/pkg/provider/types.go
```diff
@@ -35,9 +35,9 @@ const (
 	HetznerCloudProvider      = "hetzner"
 	VSphereCloudProvider      = "vsphere"
 	GCPCloudProvider          = "gcp"
-
```
alvaroaleman, line 39: Can we leave this split?

api/pkg/resources/machine/machinedeployment.go
alvaroaleman: Okay but this error actually means there is no os, not that we have an unknown os, right? Admittedly you didn't introduce that

moadqassem: I actually didn't, this was introduced in a previous commit and I just moved the code from a function into a new function. Anyhow I will change that into, `no machine os was provided`

api/pkg/provider/cloud/kubevirt/provider.go
moadqassem: what? dude this is a slice of bytes! if you don't like the `fmt.Sprintf()` I can't use casting `string(config)`

api/pkg/provider/types.go
moadqassem: Sure

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: Done

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: Done

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: Made the message very specific to CoreOS instead of a general error

api/cmd/conformance-tests/scenarios_kubevirt.go
moadqassem: Done

api/pkg/presets/manager.go
moadqassem: Returned as their original state

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: Done

api/pkg/api/v1/types.go
maciaszczykm: It is used to return `nil` when the internal object is `nil`. If we would return structure like `&PublicKubevirtCloudSpec{}` in all cases (filled with properties or not) it would cause errors in the frontend. Frontend looks for a non-null object to recognize which provider is used (only one of them can be non-null in the `PublicCloudSpec`).

PR #3943, opened 2019-08-02
improve error handling and logging in s3-storeuploader

xrstf:
This replaces glog with zap in the s3-storeuploader and generally improves its error handling. The most important fix is that bad endpoints (bad URL, HTTP failures, network failures) are no longer hidden behind a bogus "object name cannot be empty" error, but rather displayed to the user:

```
./s3-storeuploader delete-all --endpoint=foo --prefix=foo
2019-08-02T22:14:46.143+0200    fatal   s3-storeuploader/main.go:153    Failed to run command   {"err": "Get http://foo/kubermatic-backups/?location=: dial tcp: lookup foo: no such host"}
```

Likewise, empty prefixes are no longer allowed.

I made the two new logging CLI flags global because otherwise I could not access them in my global Before() handler. Alternatively we could (maybe?) use a Before() handler for each action, so that the CLI flags can be per-command. Given urfave's parsing, users now **must** put the logging flags before the subcommand name: `./s3-storeuploader --log-debug delete-all --endpoint=...`.

```release-note
NONE
```

### DIFF
--- a/api/cmd/s3-storeuploader/main.go
+++ b/api/cmd/s3-storeuploader/main.go
@@ -76,7 +76,7 @@ func main() {
 	logFormatFlag := cli.GenericFlag{
 		Name:  "log-format",
 		Value: &defaultLogFormat,
-		Usage: fmt.Sprintf("Use either %s or %s to change the log output", log.FormatConsole, log.FormatJSON),
+		Usage: fmt.Sprintf("Use one of [%v] to change the log output", log.AvailableFormats),
 	}
 
 	app.Flags = []cli.Flag{
--- a/api/cmd/s3-storeuploader/main.go
+++ b/api/cmd/s3-storeuploader/main.go
@@ -150,7 +150,7 @@ func main() {
 	// we know it's been a CLI parsing failure and the cli package
 	// has already output the error and printed the usage hints.
 	if err != nil && logger != nil {
-		logger.Fatalw("Failed to run command", "err", err)
+		logger.Fatal("Failed to run command", zap.Error(err))
 	}
 }
 
--- a/api/cmd/s3-storeuploader/main.go
+++ b/api/cmd/s3-storeuploader/main.go
@@ -1,13 +1,18 @@
 package main
 
 import (
-	"log"
+	"fmt"
 	"os"
 
-	"github.com/golang/glog"
+	"github.com/go-logr/zapr"
+	"github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/storeuploader"
 	"github.com/urfave/cli"
+	"go.uber.org/zap"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
+
+var logger *zap.SugaredLogger
 
 func main() {
 	app := cli.NewApp()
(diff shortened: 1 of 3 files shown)


api/cmd/s3-storeuploader/main.go
```diff
@@ -108,27 +128,53 @@ func main() {
...
+	}()
+
 	err := app.Run(os.Args)
-	if err != nil {
-		log.Fatal(err)
+	// Only log failures when the logger has been setup, otherwise
+	// we know it's been a CLI parsing failure and the cli package
+	// has already output the error and printed the usage hints.
+	if err != nil && logger != nil {
+		logger.Fatalw("Failed to run command", "err", err)
```
mrIncompetent: nit: use `zap.Error(err)` instead. That ensures we have a consistent error key

api/pkg/storeuploader/storeuploader.go
```diff
@@ -21,54 +23,76 @@ const prefixSeparator = "storeuploader"
...
+		if object.Err != nil {
+			return object.Err
+		}
 		existingObjects = append(existingObjects, object)
 	}
 
+	u.logger.Debugw("Done listing bucket", "bucket", bucket, "prefix", prefix, "objects", len(existingObjects))
+
 	for _, object := range u.getObjectsToDelete(existingObjects, revisionsToKeep) {
+		u.logger.Debugw("Removing object", "bucket", bucket, "object", object.Key)
```
mrIncompetent: Info

api/pkg/storeuploader/storeuploader.go
```diff
@@ -77,17 +101,29 @@ func (u *StoreUploader) DeleteOldBackups(file, bucket, prefix string, revisionsT
...
+		if object.Err != nil {
+			return object.Err
+		}
 		existingObjects = append(existingObjects, object)
 	}
 
+	u.logger.Debugw("Done listing bucket", "bucket", bucket, "prefix", prefix, "objects", len(existingObjects))
+
 	for _, object := range existingObjects {
+		u.logger.Debugw("Removing object", "bucket", bucket, "object", object.Key)
```
mrIncompetent: Info

api/pkg/storeuploader/storeuploader.go
```diff
@@ -21,54 +23,76 @@ const prefixSeparator = "storeuploader"
...
+// DeleteOldBackups deletes revisions of all files of the given prefix which are older than max-revisions
+func (u *StoreUploader) DeleteOldBackups(bucket, prefix string, revisionsToKeep int) error {
+	if len(prefix) == 0 {
+		return errors.New("prefix cannot be empty")
+	}
+
 	doneCh := make(chan struct{})
 	defer close(doneCh)
 
+	u.logger.Debugw("Listing existing objects", "bucket", bucket, "prefix", prefix)
```
mrIncompetent: Create a sub logger here with bucket & prefix. That way you avoid setting those fields for the following calls.
This also reduces the chances for typos in keys

mrIncompetent: Oh and please add `revisions-to-keep`

api/pkg/storeuploader/storeuploader.go
```diff
@@ -77,17 +101,29 @@ func (u *StoreUploader) DeleteOldBackups(file, bucket, prefix string, revisionsT
...
+// DeleteAll deletes all revisions of all files matching the given prefix
+func (u *StoreUploader) DeleteAll(bucket, prefix string) error {
+	if len(prefix) == 0 {
+		return errors.New("prefix cannot be empty")
+	}
+
 	doneCh := make(chan struct{})
 	defer close(doneCh)
 
+	u.logger.Debugw("Listing existing objects", "bucket", bucket, "prefix", prefix)
```
mrIncompetent: https://github.com/kubermatic/kubermatic/pull/3943/files#r310285769

api/pkg/storeuploader/storeuploader.go
xrstf: I called it `keep`.

PR #3967, opened 2019-08-06
Cluster cleanup: Only create usercluster client if its actually needed

alvaroaleman:
We currently always require a usercluster client to even attempt to
cleanup a cluster. This means cleanup for clusters that were deleted too
quick can not work.

This PR fixes that by defering the creation of the usercluster client
until its actually being used.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
none
```

/assign @nikhita @mrIncompetent

PR #3969, opened 2019-08-06
fix deleting cluster when not started

### LINKED ISSUE #3968: Fix deleting cluster when not started yet
**User Story**
Deleting cluster after creation causes the problem. Deleting hangs because of finalizers: CredentialsSecretsCleanupFinalizer, InClusterLBCleanupFinalizer, InClusterPVCleanupFinalizer which are set and can not be deleted.

zreigz:
**What this PR does / why we need it**: fix deleting cluster when not started yet

Fixes #3968

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -7,10 +7,12 @@ import (
 
 	"go.uber.org/zap"
 
+	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
 	k8cuserclusterclient "github.com/kubermatic/kubermatic/api/pkg/cluster/client"
 	"github.com/kubermatic/kubermatic/api/pkg/clusterdeletion"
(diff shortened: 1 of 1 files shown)


api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -215,6 +217,28 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 		log.Debug("Cleaning up cluster")
 		userClusterClient, err := r.userClusterConnProvider.GetClient(cluster)
 		if err != nil {
+			if kuberneteshelper.HasFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer) {
```
alvaroaleman, line 220: I do not think it is safe to assume that any error we get creating the usercluster client implies that the cluster was never up. We were thinking to do something like this: https://github.com/kubermatic/kubermatic/pull/3967

and add code to the API to not add the LB+PV cleanup finalizer if the node cleanup finalizer doesn't exist, as it will only get added when the cluster was up

PR #3972, opened 2019-08-06
fix namespace handling in s3-exporter

xrstf:
The S3 exporter chart has the `kube-system` namespace hardcoded, even though on some of our clusters it's running in an `s3-exporter` namespace. This PR attempts to normalise this by making the chart more flexible (like all our other charts) and putting it into the `kubermatic` namespace by default (because without Kubermatic, the exporter is useless). It does not matter in the slightest into which namespace the exporter is installed.

I expect some ruckus and manual intervention in cases where the Chart is currently weirdly broken (i.e. our cloud clusters), but this is non-critical (at worst there will be a few non-essential metrics missing while the admin fixes the chart).

**NOTE**
This has been partially reverted by #3984 because of the location of the s3-credentials secret.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #3973, opened 2019-08-06
Only do LB and PV cleanup when the cluster was up

alvaroaleman:
This PR changes the API do not add the LB and PV cleanup finalizer if
the usercluster APIServer was never up. We use the
node-deletion-finalizer to determine this, it gets only added after the
API came up. This prevents deadlocks in cases where the usercluster API never came up successfully.

Also it
* Changes the api finalizer testcases to only contain the attributes that actually change
* Removes the `kubermaticapiv1` import in which was a second alias in `api/pkg/handler/test/helper.go` for the `github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1` package which was very confusing because we usually use that import alias for a different package
* Extends the `GenCluster` used for testing to accept an optional `...func(*kubermaticv1.Cluster)` so the cluster can be manipulated which was required in order to update the existing tests

Fixes #

**Documentation**:

```release-note
Fixed an issue that kept clusters stuck if their creation didn't succeed and they got deleted with LB and/or PV cleanup enabled
```

/assign @mrIncompetent @zreigz

### DIFF
--- a/api/pkg/handler/test/helper.go
+++ b/api/pkg/handler/test/helper.go
@@ -29,7 +29,6 @@ import (
 	kubermaticfakeclentset "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned/fake"
 	kubermaticclientv1 "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned/typed/kubermatic/v1"
 	kubermaticinformers "github.com/kubermatic/kubermatic/api/pkg/crd/client/informers/externalversions"
-	kubermaticapiv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/handler/auth"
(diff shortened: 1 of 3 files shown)


floreks:
What about `kubermatic.io/cleanup-credentials-secrets`? Right now it is the one always blocking deletion of the cluster. If you create the cluster and immediately delete it then this finalizer prevents it from being deleted.

alvaroaleman:
> What about kubermatic.io/cleanup-credentials-secrets? Right now it is the one always blocking deletion of the cluster. If you create the cluster and immediately delete it then this finalizer prevents it from being deleted.

The underlying problem is that we always create a client for the usercluster before even attempting to do any kind of cleanup. The `kubermatic.io/cleanup-credentials-secrets` made this surface, because it doesn't actually need that client, but the code that cleans it up wont get executed when the usercluster client creation failed.

I have sent a different PR to fix that: https://github.com/kubermatic/kubermatic/pull/3967

floreks:
> The underlying problem is that we always create a client for the usercluster before even attempting to do any kind of cleanup. The kubermatic.io/cleanup-credentials-secrets made this surface, because it doesn't actually need that client, but the code that cleans it up wont get executed when the usercluster client creation failed.
>
> I have sent a different PR to fix that: #3967

Awesome. Hopefully, this gets merged soon.

PR #3980, opened 2019-08-06
Fix seed-proxy controller not actually talking to seeds

alvaroaleman:
The current deployment creates pods Kubernetes wont create, resulting in
the Deployments failing:

```
$ k describe pod seed-proxy-europe-west3-c-1-579fd55967-dg5j5|grep Error
      Reason:       CreateContainerError
  Warning  Failed     3m36s (x12 over 5m54s)  kubelet, gke-kubermatic-cloud-euro-preemtible2-8982be0c-z2gs  Error: Error response from daemon: Duplicate mount point: /var/run/secrets/kubernetes.io/serviceaccount
```

Fixes #

**Documentation**:

```release-note
none
```

/assign @xrstf 
/approve

### DIFF
--- a/api/pkg/controller/seed-proxy/reconciler.go
+++ b/api/pkg/controller/seed-proxy/reconciler.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


xrstf:
I have extended the hotfix to include a real fix for the underlying issue. The mount path was just plain wrong, but funny enough it did not explode in clusters other than dev. Likewise, because I tested this on dev only, I never noticed that the SA token I was injecting into the seed-proxy pod had no cluster URL, so the seed-proxy was always only talking to the local cluster (the master), never the remote cluster (the seed).

The changes include converting the CA+token into a nice, handy kubeconfig. This is so that using kuebctl is easier inside the pod. I also removed the explicit setters for names and namespaces, as I have now learned that the reconciling system already takes care of these fields.

PR #3986, opened 2019-08-07
AWS: remove the remains of hardcoded AZ from datacenter and cluster configuration

kdomanski:
It removes useless code and data field related to how AWS clusters behaved before we had multiple AZ support.

```release-note
ACTION REQUIRED: the `zone_character` field must be removed from all AWS datacenters in `datacenters.yaml`
```

### DIFF
--- a/api/pkg/crd/kubermatic/v1/cluster.go
+++ b/api/pkg/crd/kubermatic/v1/cluster.go
@@ -316,8 +316,6 @@ type AWSCloudSpec struct {
 	RouteTableID        string `json:"routeTableId"`
 	InstanceProfileName string `json:"instanceProfileName"`
 	SecurityGroupID     string `json:"securityGroupID"`
-
-	AvailabilityZone string `json:"availabilityZone"`
 }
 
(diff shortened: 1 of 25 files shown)


api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -316,8 +316,6 @@ type AWSCloudSpec struct {
 	RouteTableID        string `json:"routeTableId"`
 	InstanceProfileName string `json:"instanceProfileName"`
 	SecurityGroupID     string `json:"securityGroupID"`
-
-	AvailabilityZone string `json:"availabilityZone"`
```
alvaroaleman, line 320: What happens to pre-existing AWS clusters? If someone updates an existing NodeDeployment in an existing AWS cluster, will everything stay the same except for the given update? What about a new NodeDeployment, are additional settings required?

kdomanski: Existing NodeDeployments should not be affected, since the AZ from the cluster used to get simply copied into the MD object. New NodeDeployments have their AZ assigned by the UI.

PR #3991, opened 2019-08-07
Use Seed CRD for the conformance-tester

alvaroaleman:

Makes the conformance tester use the seed crd instead of the `datacenters.yaml` file

Fixes #

**Documentation**:

```release-note
none
```

/assign @kdomanski

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,9 +23,7 @@ linters:
 
 issues:
   exclude:
-  - cyclomatic complexity 32 of func `main` is high
-  - cyclomatic complexity 33 of func `main` is high
-  - cyclomatic complexity 43 of func `main` is high
+  - cyclomatic complexity [0-9]+ of func `main` is high
   - cyclomatic complexity 37 of func `migrateToProject` is high
   - cyclomatic complexity 33 of func `ValidateCloudSpec` is high
   - cyclomatic complexity 38 of func `migrateToProject` is high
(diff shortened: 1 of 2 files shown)


kdomanski:
`cmd/conformance-tests/main.go:134:1: cyclomatic complexity 35 of func `main` is high (> 30) (gocyclo)`

You could extract everything related to manipulating `opts` into a separate function.

.golangci.yml
```diff
@@ -23,9 +23,7 @@ linters:
 
 issues:
   exclude:
-  - cyclomatic complexity 32 of func `main` is high
-  - cyclomatic complexity 33 of func `main` is high
-  - cyclomatic complexity 43 of func `main` is high
+  - cyclomatic complexity [0-9]+ of func `main` is high
```
kdomanski, line 26: The way to cure a fever is not to throw out the thermometer. And the sheer audacity of calling the commit "fix" is mind-boggling.

kdomanski:
cc @mrIncompetent

.golangci.yml
alvaroaleman: Its not worse than having five exclude rules  with different numbers that will gladly match any `main` func

kdomanski: An proper alternative would be to split the over-sized function into smaller ones for better readability.

mrIncompetent: We had that discussion on Slack at some point.
I think the main outcome was:
- Ignoring the linter in mains is acceptable
- Some dislike the linter, some like it

alvaroaleman: I will not write shitty factored code just to satisfy a linter. A func that does a lot of top-level conditional logic without nesting it is a lot easier to read than factoring that stuff out into a distinct func that to call that distinct func once

PR #3992, opened 2019-08-07
cloud-controller: replace glog with zap

mrIncompetent:
This PR just replaces glog with zap so we have structured logging.

```release-note
NONE
```

/assign @moadqassem

### DIFF
--- a/api/pkg/controller/cloud/cloud_controller.go
+++ b/api/pkg/controller/cloud/cloud_controller.go
@@ -159,19 +159,19 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 }
 
 func (r *Reconciler) migrateICMP(ctx context.Context, log *zap.SugaredLogger, cluster *kubermaticv1.Cluster, cloudProvider provider.CloudProvider) error {
-	switch provider := cloudProvider.(type) {
+	switch prov := cloudProvider.(type) {
 	case *aws.AmazonEC2:
-		if err := provider.AddICMPRulesIfRequired(cluster); err != nil {
+		if err := prov.AddICMPRulesIfRequired(cluster); err != nil {
 			return fmt.Errorf("failed to ensure ICMP rules for cluster %q: %v", cluster.Name, err)
 		}
 		log.Info("Successfully ensured ICMP rules in security group of cluster")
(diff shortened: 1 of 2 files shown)


api/pkg/controller/cloud/cloud_controller.go
```diff
@@ -148,23 +158,23 @@ func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.Cluste
 	return nil, err
 }
 
-func (r *Reconciler) migrateICMP(ctx context.Context, cluster *kubermaticv1.Cluster, cloudProvider provider.CloudProvider) error {
+func (r *Reconciler) migrateICMP(ctx context.Context, log *zap.SugaredLogger, cluster *kubermaticv1.Cluster, cloudProvider provider.CloudProvider) error {
 	switch provider := cloudProvider.(type) {
```
moadqassem: I know this wasn't changed in your PR, however, this var here collides with an imported package: 
`"github.com/kubermatic/kubermatic/api/pkg/provider"`

api/pkg/controller/cloud/cloud_controller.go
```diff
@@ -81,15 +90,21 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		}
 		return reconcile.Result{}, err
 	}
+	log = log.With("cluster", cluster.Name)
+
+	if cluster.Spec.Pause {
+		log.Debug("Skipping because the cluster is paused")
```
moadqassem, line 96: would be good to have the cluster name in the log message

mrIncompetent: We're already enriching the logger on line 93 with the cluster name :)
So it will definitely appear in the log message

api/pkg/controller/cloud/cloud_controller.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3992/commits/b1e575078c4110cff3373bcff886672419c038f1

api/pkg/controller/cloud/cloud_controller.go
moadqassem: I see, cool

PR #3995, opened 2019-08-07
Implement different roles for worker nodes & the control plane

### LINKED ISSUE #3716: Improve AWS cluster separation
Based on: https://github.com/kubermatic/customer-projects/issues/36

**User Story**
The current "bring-your-own" account approach is solving the separation of AWS resources. We now facing to the requirement, that access needed to be separated on cluster level:

- In the same account end user of cluster A could not access resources of cluster B
- This includes resources created by the CCM (Volumes, Loadbalancer, ...)

**Identified Problems**
- Created AWS worker nodes (kubelet) get the IAM instance profile assigned -> trough host path mounting or root container the instance profile is accessible from the inside of the cluster
- The default IAM policy Kubermatic deploys is not limited to the cluster resources  -> access to other cluster resources is possible
- More details can be found here: https://github.com/kubermatic/customer-projects/issues/36

**Potential solution**
A potential solution approach have been discovered by @kron4eg. For detail see: https://github.com/kubermatic/customer-projects/issues/36#issuecomment-507447343

**Acceptance criteria**
- [ ] A end user of a cluster can only access the AWS resource they belong
- [ ] A end user cannot read out AWS credentials or IAM policies at all. AWS credentials are only visible in the Kubermatic operator scope

mrIncompetent:
This PR will make sure that we:
- Create a very limited role for the worker nodes
- Create a limited roles for the control plane (Uses assume-role)
- Both roles will only(If the API action allows it) have access to resources which have the cluster specific Kubernetes tag(Kubernetes sets the tag when creating new resources). This ensures that we have a fairly strong isolation between AWS clusters using the same account.

I decided against creating new API credentials as that would not make any improvement, since the control plane does a assume-role.

Fixes #3716 

```release-note
Reduce the AWS role permissions for the control-plane & worker role to the minimum
```


api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -308,16 +308,21 @@ type BringYourOwnCloudSpec struct{}
...
 	RouteTableID        string `json:"routeTableId"`
 	InstanceProfileName string `json:"instanceProfileName"`
 	SecurityGroupID     string `json:"securityGroupID"`
 
 	AvailabilityZone string `json:"availabilityZone"`
+
+	// DEPRECATED. Don't care for the role name. We only require the ControlPlaneRoleARN to be set so the control plane
+	// can perform the assume-role.
+	// We keep it for backwards compatibility (We use this name for cleanup purpose).
+	RoleName string `json:"roleName"`
```
alvaroaleman: Maybe add `,omitempty`?

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -308,16 +308,21 @@ type BringYourOwnCloudSpec struct{}
...
 
-	AccessKeyID         string `json:"accessKeyId,omitempty"`
-	SecretAccessKey     string `json:"secretAccessKey,omitempty"`
-	VPCID               string `json:"vpcId"`
-	RoleName            string `json:"roleName"`
-	RoleARN             string `json:"roleARN"`
+	AccessKeyID     string `json:"accessKeyId"`
+	SecretAccessKey string `json:"secretAccessKey"`
+	VPCID           string `json:"vpcId"`
+	// The IAM role, the control plane will use. The control plane will perform an assume-rule
```
alvaroaleman: `s/rule/role`

api/pkg/provider/cloud/aws/provider.go
```diff
@@ -549,21 +450,35 @@ func (a *AmazonEC2) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
...
-	if cluster.Spec.Cloud.AWS.RoleName == "" && cluster.Spec.Cloud.AWS.InstanceProfileName == "" {
-		svcIAM, err := a.getIAMClient(cluster.Spec.Cloud)
+	iamClient, err := a.getIAMClient(cluster.Spec.Cloud)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get IAM client: %v", err)
+	}
+
+	// We create a dedicated role for the control plane
+	if cluster.Spec.Cloud.AWS.ControlPlaneRoleARN == "" {
+		controlPlaneRole, err := createControlPlaneRole(iamClient, cluster.Name)
```
alvaroaleman, line 451: Since this can not be set via the API we need a lot of permisisons, right?

api/pkg/provider/cloud/aws/provider.go
```diff
@@ -669,49 +584,45 @@ func (a *AmazonEC2) CleanUpCloudProvider(cluster *kubermaticv1.Cluster, updater
...
-		_, err = iamClient.RemoveRoleFromInstanceProfile(&iam.RemoveRoleFromInstanceProfileInput{
-			RoleName:            aws.String(cluster.Spec.Cloud.AWS.RoleName),
-			InstanceProfileName: aws.String(cluster.Spec.Cloud.AWS.InstanceProfileName),
-		})
-		if err != nil {
-			if err.(awserr.Error).Code() != iam.ErrCodeNoSuchEntityException {
-				return nil, fmt.Errorf("failed to remove role %s from instance profile %s: %s", cluster.Spec.Cloud.AWS.RoleName, cluster.Spec.Cloud.AWS.InstanceProfileName, err.(awserr.Error).Message())
-			}
+		rolesToDelete := []string{
+			workerRoleName(cluster.Name),
```
alvaroaleman, line 589: Can we cope with the case that this doesn't exist because the credential lacked permissions to create it?

api/pkg/provider/cloud/aws/provider.go
```diff
@@ -741,6 +652,15 @@ func isEntityAlreadyExists(err error) bool {
 	return aerr.Code() == "EntityAlreadyExists"
 }
 
+func isNoSuchEntity(err error) bool {
```
alvaroaleman: Could we rename this to `isNotFound` for readability?

api/pkg/provider/cloud/aws/policy.go
```diff
@@ -0,0 +1,181 @@
+package aws
+
+import (
+	"bytes"
+	"text/template"
+)
+
+var (
+	// This allows instances to perform the assume role
+	assumeRolePolicy = `{
```
alvaroaleman, line 10: Why are these `var` and not `const`?

api/pkg/provider/cloud/aws/instance-profile_test.go
```diff
@@ -0,0 +1,203 @@
...
+				}, nil
+			},
+			// Make the func return an error so we get an error when it gets called
+			addRoleToInstanceProfile: func(input *iam.AddRoleToInstanceProfileInput) (*iam.AddRoleToInstanceProfileOutput, error) {
+				return nil, awserr.New("SomethingBadHappened", "test", errors.New("test"))
+			},
+		},
+	}
+
+	for _, test := range tests {
```
alvaroaleman, line 178: This is pretty awesome :)

api/pkg/provider/cloud/aws/provider.go
mrIncompetent: > Since this can not be set via the API 

The `CloudSpec` actually gets directly used by the API - so it can be set.

Regarding permissions:
We require a similar set of permissions like before.
Before we created a Role with full Route53 & EC2 access, now a role with limited EC2 & ELB access.

api/pkg/provider/cloud/aws/provider.go
mrIncompetent: If the role could not be created the finalizer won't exist - So we won't reach this code.
The finalizer will only be set if the Role&Instance profile could be created successfully.

api/pkg/provider/cloud/aws/policy.go
mrIncompetent: Only to avoid the visual "break" by introducing a separate `const` block.

We require that `var` block here because we parse the control-plane template already (Which is not possible in a const).

If you prefer, i can move the fixed string variable to a const block

api/pkg/provider/cloud/aws/instance-profile_test.go
mrIncompetent: I wasn't even aware that this is possible. 
The AWS SDK had a great example on how to test the IAM part - Their docs are pretty good!

api/pkg/crd/kubermatic/v1/cluster.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3995/commits/89fd0569fa8f11a0cea254cd74fe548907c64c6d

api/pkg/crd/kubermatic/v1/cluster.go
mrIncompetent: Fixed with https://github.com/kubermatic/kubermatic/pull/3995/commits/29d472eb7aee5afb02a42d8293ce7257ec59b9e5

api/pkg/provider/cloud/aws/provider.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/3995/commits/43ba204968d6f39aa35151daee5349e84f57303a

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -308,14 +308,21 @@ type BringYourOwnCloudSpec struct{}
...
+	AccessKeyID     string `json:"accessKeyId"`
+	SecretAccessKey string `json:"secretAccessKey"`
+	VPCID           string `json:"vpcId"`
+	// The IAM role, the control plane will use. The control plane will perform an assume-role
+	ControlPlaneRoleARN string `json:"roleARN"`
 	RouteTableID        string `json:"routeTableId"`
 	InstanceProfileName string `json:"instanceProfileName"`
 	SecurityGroupID     string `json:"securityGroupID"`
+
+	AvailabilityZone string `json:"availabilityZone"`
```
alvaroaleman: Why did you re-add this? @kdomanski deliberately remvoved it in #3986

mrIncompetent: I screwed up the merge :man_facepalming:

mrIncompetent: Gone with https://github.com/kubermatic/kubermatic/pull/3995/commits/564566eab7eab8933aab27e641ca9bf885da3124

PR #4008, opened 2019-08-08
Add ValidatingWebhook for seed CRDs

### LINKED ISSUE #3888: Add validatingAdmissionWebhook for Seed CRD
**User Story**

As an operator of Kubermatic, I want validation that prevents me from accidentally bringing my Kubermatic installation in a defunct state.

**Acceptance criteria**

* There is a ValidatingAdmissionWebhook for the `Seed` CRD
* It gets deployed in both seeds and master clusters
* It gets automatically set up within the Kubermatic chart
* It validates that no Datacenter can be deleted or updated as long as there are clusters that use it
* It validates that Datacenter names are unique

alvaroaleman:

Adds a ValidatingWebhook for seed CustomResources that prevents removal of Datacenters that are being in use or Creation of duplicate Datacenters
Fixes #3888 

**Documentation**:

```release-note
none
```


alvaroaleman:
Unfortunately I had to change the signature of the `SeedKubeconfigGetter` to take a `*kubermaticv1.Seed` as argument instead of a `seedName string`, because otherwise CREATE requests can never succeed, as the webhook can not get the kubeconfig as the seed does not exist yet.

This is why the PR resulted in so many changes.

Other than the included unit tests I also manually verified that its not possible to add a second Seed that contains DCs that are already being used and to delete a seed that is being used.

As an aside, it would probably make sense to DRY out the whole restmapper story and offer a higher-level type that has a simple `func (rc *RestMapperCache) GetRestMapper(cfg *rest.Config, cacheKey string)(*RESTMapper, error)`, as we now have three places where its being used and in the future there may come even more.

/assign @mrIncompetent

api/pkg/validation/seed/server.go
```diff
@@ -0,0 +1,131 @@
...
+	if admissionRequest != nil {
+		uid = admissionRequest.UID
+	}
+	response := &admissionv1beta1.AdmissionReview{
+		Request: admissionRequest,
+		Response: &admissionv1beta1.AdmissionResponse{
+			UID:     uid,
+			Allowed: validationErr == nil,
+			Result: &metav1.Status{
+				Message: fmt.Sprintf("%v", validationErr),
```
mrIncompetent, line 92: validationErr.Error()

api/pkg/validation/seed/server.go
```diff
@@ -0,0 +1,131 @@
...
+	serializedAdmissionResponse, err := json.Marshal(response)
+	if err != nil {
+		s.log.Errorw("failed to serialize admission response", zap.Error(err))
+		http.Error(resp, "failed to serialize response", http.StatusInternalServerError)
+		return
+	}
+	resp.WriteHeader(http.StatusOK)
+	if _, err := resp.Write(serializedAdmissionResponse); err != nil {
+		s.log.Errorw("failed to write response body", zap.Error(err))
+	}
```
mrIncompetent, line 106: This is missing a return

api/pkg/validation/seed/server.go
```diff
@@ -0,0 +1,131 @@
...
+	s.log.Info("Successfully validated seed")
```
mrIncompetent: Rather debug?

api/pkg/validation/seed/server.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	seed := &kubermaticv1.Seed{}
+	if err := json.Unmarshal(admissionReview.Request.Object.Raw, seed); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal object from request into a Seed: %v", err)
+	}
+
+	validationErr := s.validator.Validate(seed, admissionReview.Request.Operation == admissionv1beta1.Delete)
+	if validationErr != nil {
+		s.log.With("seed", seed.Name).Errorw("seed failed validation", "validationError", validationErr.Error())
```
mrIncompetent: Chose one way to append fields

mrIncompetent: The `validationErr` might also be caused by some transient error, so it would be great to also set `zap.Error`

api/pkg/validation/seed/server.go
alvaroaleman: This was done delibaretely, `validationErr` may be nil

api/pkg/validation/seed/server.go
alvaroaleman: Maybe, but that is not possible to tell for us, it can also mean the user submitted a broken seed which is why I deliberately went without `zap.Error`

alvaroaleman:
@mrIncompetent updated, ptal

Issue #3987, opened 2019-08-07
Kubermatic e2e tests: Include cluster deletion


alvaroaleman:
**User Story**

Currently we dont include cluster deletion in e2e tests, allowing bugs there to not make the tests fail

**Acceptance criteria**

* The kubermatic e2e tests delete the cluster
* They wait until the cluster is gone
* They fail if the cluster doesn't disappear within a given timeout

kdomanski:
kubermatic/kubermatic#2739

alvaroaleman:
Taking this over so we have something here asap.
/assign

PR #4011, opened 2019-08-08
conformance-tests: use zap logger

nikhita:
To ensure that we delete the cluster in conformance tests (https://github.com/kubermatic/kubermatic/issues/3987), we'll need to use `CleanupCluster`, which needs the zap logger:

https://github.com/kubermatic/kubermatic/blob/15598e6f69881c52327993a5313498aeddec4018/api/pkg/clusterdeletion/deletion.go#L35

```release-note
NONE
```

/cc @alvaroaleman @mrIncompetent

### DIFF
--- a/api/cmd/conformance-tests/main.go
+++ b/api/cmd/conformance-tests/main.go
@@ -19,11 +19,12 @@ import (
 
 	"github.com/go-openapi/runtime"
 	httptransport "github.com/go-openapi/runtime/client"
-	"github.com/sirupsen/logrus"
+	"go.uber.org/zap"
 	"golang.org/x/crypto/ssh"
 
 	clusterclient "github.com/kubermatic/kubermatic/api/pkg/cluster/client"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/semver"
 	kubermaticsignals "github.com/kubermatic/kubermatic/api/pkg/signals"
@@ -62,7 +63,6 @@ type Opts struct {
 	workerName                   string
 	homeDir                      string
 	versions                     []*semver.Semver
-	log                          *logrus.Entry
 	excludeSelector              excludeSelector
 	excludeSelectorRaw           string
 	existingClusterLabel         string
@@ -127,14 +127,10 @@ var defaultTimeout = 10 * time.Minute
 var (
 	providers  string
 	pubKeyPath string
-	debug      bool
 	sversions  string
 )
 
 func main() {
-	mainLog := logrus.New()
-	mainLog.SetLevel(logrus.InfoLevel)
-
 	opts := Opts{
 		providers:  sets.NewString(),
 		publicKeys: [][]byte{},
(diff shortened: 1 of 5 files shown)


api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -372,7 +372,8 @@ elif [[ $provider == "vsphere" ]]; then
 fi
 
 timeout -s 9 90m ./conformance-tests $EXTRA_ARGS \
-  -debug \
+  -log-debug=true \
```
alvaroaleman: We can unfortunately not change flags for the conformance tester, because we use the same CLI args for older versions of it when we run upgrade tests. So the existing `-debug` flag must stay for now, else we lose the debug log on older conformance-tester versions and we must not add new flags.

IMHO we should just hardcode it to debug log enabled and json log, so we can properly filter in ES. WDYT?

mrIncompetent:
This PR makes me happy :)

alvaroaleman:
That was me, I am sorry :grimacing: 
/retest

api/cmd/conformance-tests/main.go
```diff
@@ -143,9 +140,10 @@ func main() {
 
 	defaultTimeoutMinutes := 10
 
+	mainLog := kubermaticlog.New(false, kubermaticlog.FormatConsole)
```
alvaroaleman: Can we change this to enable debug and print json? :grimacing:

api/cmd/conformance-tests/main.go
```diff
@@ -171,6 +168,7 @@ func main() {
 	flag.BoolVar(&opts.openshift, "openshift", false, "Whether to create an openshift cluster")
 	flag.BoolVar(&opts.printGinkoLogs, "print-ginkgo-logs", false, "Whether to print ginkgo logs when ginkgo encountered failures")
 	flag.BoolVar(&opts.onlyTestCreation, "only-test-creation", false, "Only test if nodes become ready. Does not perform any extended checks like conformance tests")
+	flag.BoolVar(&opts.log.Debug, "debug", false, "Enable debug logging")
```
alvaroaleman: Can we change this to `_ = flag.Bool(...)` so ppl don't think it has an effect?

api/cmd/conformance-tests/runner.go
```diff
@@ -76,7 +80,7 @@ func newRunner(scenarios []testScenario, opts *Opts) *testRunner {
 		workerName:                   opts.workerName,
 		homeDir:                      opts.homeDir,
 		seedClusterClient:            opts.seedClusterClient,
-		log:                          opts.log,
+		log:                          log,
```
alvaroaleman, line 79: Why don't we continue to pass the log down?

api/cmd/conformance-tests/main.go
nikhita: It's being used below in line 198 though... alternatively, I can always enable debug to `true` too

api/cmd/conformance-tests/runner.go
nikhita: Because it is easier to invoke `r.log.Info` if log is a `zap.SugaredLogger` rather than `kubermaticlog.Options` :grimacing:

api/cmd/conformance-tests/main.go
alvaroaleman: Yeah I think that makes sense :+1: 

Also you are constructing a log on line 143 and then again on 198? I think one is enough :P

api/cmd/conformance-tests/runner.go
alvaroaleman: I thought more about passing the actual `mainLog` down :P Doesn't have to be put in the `Opts` sctruct thought

api/cmd/conformance-tests/main.go
nikhita: It was a little annoying because we needed to log before we parsed the flags... but changed it to hardcode it so this is fixed now.

api/cmd/conformance-tests/runner.go
nikhita: Done :)

PR #4016, opened 2019-08-08
Return SSH key when assigning it to cluster

kgroschoff:
Return SSH key when calling `PUT ​/api​/v1​/projects​/{project_id}​/dc​/{dc}​/clusters​/{cluster_id}​/sshkeys​/{key_id}`.

Related to https://github.com/kubermatic/dashboard-v2/issues/1443 

@zreigz `GET` request for SSH keys that belong to a cluster already returns full object. Therefor I think it's okay to also do that in `PUT` request.

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -2768,7 +2768,10 @@
         ],
         "responses": {
           "201": {
-            "$ref": "#/responses/empty"
+            "description": "SSHKey",
+            "schema": {
+              "$ref": "#/definitions/SSHKey"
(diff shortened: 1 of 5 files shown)


zreigz (changes requested):
I think this part also should be changed: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/cluster/cluster.go#L531
Please adopt tests for the changes and add a new test: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/v1/cluster/cluster_test.go#L527

PR #4020, opened 2019-08-08
Prevent that removed owners come back

### LINKED ISSUE #4018: Deleted user appears in the project again
**User Story**
When I delete a user from a project after a short time (approx 10min) he appears again

**Acceptance criteria**

mrIncompetent:
The intention here was to keep the PR as small as possible to reduce the potential for undesired side effects. We still require a cleanup of this PR.

What happens when we add a `Owner` to a `Project`:
- The API creates a `UserProjectBinding` which binds the user to the `Project`
- The `kubermatic_sync_projectbinding_controller` will set all users which are `Owner` as  `OwnerReference` on the `Project`

What happens when we remove a owner from a `Project`:
- The API deletes the `UserProjectBinding`

The caveat here is that we have a project-controller which will create `UserProjectBindings` for `OwnerReferences` of a `Project`.
Since we do not remove users from a project `OwnerReferences`, there is not way to remove a `Owner` from a `Project`.

This fix simply skips the `UserProjectBinding` creation for entries in a projects OwnerReferences.

A goal longterm would be to completely remove this second way of establishing a relation between `Users` & `Projects`. 
While the initial idea of having a automated cleanup(When the `Owner` `User` gets deletes) seems nice, it can have catastrophic results.
Example:
- A project owner leaves the company & no new owner got set
- That project hosts production clusters
- Some day an operator goes ahead and deletes all "dead" users
- We suddenly save a lot of energy.

Fixes #4018

```release-note
Fixed an issue where deleted project owners would come back after a while
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -44,3 +44,4 @@ issues:
   - "`osData` can be `github.com/kubermatic/kubermatic/api/pkg/resources.RoleBindingDataProvider`"
   - "cyclomatic complexity 31 of func `main` is high"
   - "`currentMigrationRevision` is unused"
+  - "unreachable code"
--- a/api/pkg/controller/rbac/sync_project.go
+++ b/api/pkg/controller/rbac/sync_project.go
@@ -98,6 +98,11 @@ func (c *projectController) ensureProjectIsInActivePhase(project *kubermaticv1.P
 
 // ensureProjectOwner makes sure that the owner of the project is assign to "owners" group
 func (c *projectController) ensureProjectOwner(project *kubermaticv1.Project) error {
+	// TODO: Cleanup
+	// This is a hotfix to prevent that owners that got removed from a project get a new UserProjectBinding.
+	// We already create & delete the appropriate UserProjectBinding in the API when adding/removing users, so this should not be needed.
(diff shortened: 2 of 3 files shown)


mrIncompetent:
Regarding testing:
I did a quick test locally using the worker-name flag:
- Run all controllers, API & Dashboard locally (Probably using the hack scripts)
- Create a new project
- Add the `worker-name` label to the project
- Invite someone as `Owner` to the project
- Remove the other person
- See that the person does not get re-added

mrIncompetent:
Fixed linting with a temporary exception rule: https://github.com/kubermatic/kubermatic/pull/4020/commits/6a83d6a628f38194f963413685de062eeae5c0b3

zreigz:
restarting the master controller causes this problem. Did you restart the master controller?

> Regarding testing:
> I did a quick test locally using the worker-name flag:
> 
> * Run all controllers, API & Dashboard locally (Probably using the hack scripts)
> * Create a new project
> * Add the `worker-name` label to the project
> * Invite someone as `Owner` to the project
> * Remove the other person
> * See that the person does not get re-added

zreigz:
I've tested the fix. Work fine :)

PR #4029, opened 2019-08-09
drop EOL Kubernetes 1.11

kdomanski:
```release-note
Kubernetes 1.11 which is end-of-life has been removed.
```
/assign @alvaroaleman


config/kubermatic/static/master/updates.yaml
```diff
@@ -1,42 +1,9 @@
 updates:
-# ======= 1.8 =======
-# Allow to next minor release
```
alvaroaleman, line 3: If we remove this, ppl can not even upgrade from older versions anymore and are basically stuck on it

PR #4033, opened 2019-08-10
FluentBit: parse the kubelet glog output

mrIncompetent:
This makes FluentBit parse the kubelet logs when `collectSystemd` is enabled.

```release-note
NONE
```

### DIFF
--- a/config/logging/fluentbit/templates/configmap.yaml
+++ b/config/logging/fluentbit/templates/configmap.yaml
@@ -22,6 +22,7 @@ data:
     @INCLUDE kubernetes.conf
     {{- if .Values.logging.fluentbit.configuration.collectSystemd }}
     @INCLUDE systemd.conf
+    @INCLUDE kubelet.conf
     {{- end }}
     {{- if .Values.logging.fluentbit.configuration.collectKernelMessages }}
     @INCLUDE kmesg.conf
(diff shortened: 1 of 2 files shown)


config/logging/fluentbit/templates/configmap.yaml
```diff
@@ -258,7 +278,5 @@ data:
     [PARSER]
         Name        glog
         Format      regex
-        Regex       ^(?<level>[A-Z])(?<time>[0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]+)\s+(?<thread>[0-9]+)\s+(?<file>[^:]+):(?<line>[0-9]+)\]\s*(?<message>.*)$
-        Time_Key    time
```
mrIncompetent, line 262: Taking the time from the log line breaks the forwarding to ES somehow.

Question is: Do we actually want to take the time from the log?
We already have a time key when the log comes through systemd/journald or containers.

xrstf: For containers I found that the time I could see in Kibana was "chunked" and seemingly it was the time when fluentbit shipped the log, not when the line was actually output. That was the reason for parsing the time and calling it `time`.

mrIncompetent: I'll see if i can do some testing on this tonight.
So far i only tested this with containerd, which has a different logging format than Docker :/

PR #4043, opened 2019-08-12
Use a distinct and public image for the dnat controller

alvaroaleman:

Puts the dnat controller into a distinct image that is public. We need this because with openshift, we have one pod that contains both the dnat controller and some openshift image. We need a registry secret for openshift and its only possible to use one registry secret per pod and registry.

Fixes #

**Documentation**:

```release-note
none
```


api/Dockerfile
```diff
@@ -1,14 +1,10 @@
-FROM debian:buster
+FROM alpine:3.10
```
alvaroaleman, line 1: We might have to switch back to something glibc-based for openshift later on, but right now its not required

PR #4054, opened 2019-08-13
Get VSphere Folders

### LINKED ISSUE #4004: vsphere: create an endpoint to list available folders
**User Story**
Similarly to the existing `/providers/vsphere/networks` endpoint which allows the UI to offer a choice of existing networks, we need an endpoint that will list the available folders in which to place a VM.

**Acceptance criteria**
- [ ] and endpoint `/providers/vsphere/folders` exists that lists available VM folders
- [ ] and endpoint `/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/providers/vsphere/folders` exists that lists available VM folders
- [ ] both endpoints also list nested folders and do not obfuscate the hierarchy

kgroschoff:
Get VSphere Folders from api

Fixes #4004

/

```release-note
Fetch VSphere Folders from api.
```


api/pkg/handler/v1/provider/vsphere.go
```diff
@@ -98,6 +99,91 @@ func getVsphereNetworks(seedsGetter provider.SeedsGetter, username, password, da
...
+					break
+				}
+			}
+		}
+
+		return getVsphereFolders(seedsGetter, username, password, req.DatacenterName)
+	}
+}
+
+func VsphereFoldersNoCredentialsEndpoint(projectProvider provider.ProjectProvider, seedsGetter provider.SeedsGetter) endpoint.Endpoint {
```
alvaroaleman, line 125: I don't think we should have this. This happens to work on certain circumstances (infraManagementUser is configured) but IMHO we shouldn't expose info to users that haven't entered valid credentials

api/pkg/handler/v1/provider/vsphere.go
```diff
@@ -98,6 +99,91 @@ func getVsphereNetworks(seedsGetter provider.SeedsGetter, username, password, da
...
+	datacenter, err := provider.DatacenterFromSeedMap(seeds, datacenterName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find Datacenter %q: %v", datacenterName, err)
+	}
+	vsProvider, err := vsphere.NewCloudProvider(datacenter)
+	if err != nil {
+		return nil, err
+	}
+
+	networks, err := vsProvider.GetFolders(kubermaticv1.CloudSpec{
```
alvaroaleman: copy paste variable name

api/pkg/handler/v1/provider/vsphere.go
```diff
@@ -98,6 +99,91 @@ func getVsphereNetworks(seedsGetter provider.SeedsGetter, username, password, da
...
+				Username: username,
+				Password: password,
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var filterVM = regexp.MustCompile(`(^\/[\w\d\-]+\/vm)`)
```
alvaroaleman: Isn't the logic here just a simple `strings.HasPrefix(folder, "/vm")`?

api/pkg/provider/cloud/vsphere/provider.go
```diff
@@ -181,6 +186,45 @@ func (v *Provider) GetNetworks(cloud kubermaticv1.CloudSpec) ([]Network, error)
...
+	finder.SetDatacenter(vsphereDC)
+
+	// finder is relative to datacenter, so * is fine for us.
+	folderRefs, err := finder.FolderList(ctx, "*")
+	if err != nil {
+		return nil, fmt.Errorf("couldn't retrieve folder list: %v", err)
+	}
+
+	var folders []Folder
+	for _, netRef := range folderRefs {
```
alvaroaleman: copy paste variable name

api/pkg/api/v1/types.go
```diff
@@ -410,6 +410,13 @@ type VSphereNetwork struct {
 	Name string `json:"name"`
 }
 
+// VSphereFolder is the object representing a vsphere folder.
+// swagger:model VSphereFolder
+type VSphereFolder struct {
+	// Path is the path of the folder
+	Path string `json:"path"`
```
alvaroaleman, line 417: It is the fully qualified path, right?

api/pkg/handler/v1/provider/vsphere.go
```diff
@@ -98,6 +99,91 @@ func getVsphereNetworks(seedsGetter provider.SeedsGetter, username, password, da
...
+		DatacenterName: datacenterName,
+		VSphere: &kubermaticv1.VSphereCloudSpec{
+			InfraManagementUser: kubermaticv1.VSphereCredentials{
+				Username: username,
+				Password: password,
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
```
alvaroaleman: Can we be a bit nice to users and wrap it with a `failed to get folders: %v`?

alvaroaleman (commented):
Thanks! Did you check if this works for both top-level and nested folders? If not we should leave the ticket open to verify, Nico mentioned we have some issues around that

api/pkg/handler/v1/provider/vsphere.go
kgroschoff: I think the name is a bit confusing. 🤔 
`[...]NoCredentialsEndpoint` is used, if user enters his credentials seperately instead of using the presets option.

To be honest, I just reused name, as we're already using it for all providers. Maybe we could find a better name? WDYT @zreigz ?

api/pkg/handler/v1/provider/vsphere.go
kgroschoff: Done

api/pkg/provider/cloud/vsphere/provider.go
kgroschoff: Done

api/pkg/handler/v1/provider/vsphere.go
kgroschoff: Tried it. Eather I did something wrong, or it just returned a empty array. 🤔 
Paths start with the datacenter name and then `/vm` (or `/network` or whatever), e.g. `/dc-1/vm/kubermatic`

api/pkg/handler/v1/provider/vsphere.go
kgroschoff: Done

kgroschoff:
Thanks @alvaroaleman , added most of you're comments. EP should also contain nested folders.

api/pkg/handler/v1/provider/vsphere.go
alvaroaleman: Oh, that is definitely very confusing. Could we probably rename it to `${Thing}WithPresetEndpoint` or something the like?

api/pkg/handler/v1/provider/vsphere.go
alvaroaleman: Oh but we do know the datacenter name or not? So maybe a strings.HasPrefix(folder, `/$datacenterName/vm/`?

api/pkg/handler/v1/provider/vsphere.go
alvaroaleman: (Not part of this PR thought I guess)

api/pkg/handler/v1/provider/vsphere.go
alvaroaleman: (I just find the current regexp hard to read and what happens if there is something like `/datacenter/networks/vm/my-folder`?)

api/pkg/handler/v1/provider/vsphere.go
zreigz: we use this convention for other endpoints too. We can change this later during Friday refactor

zreigz (approved):

api/pkg/handler/v1/provider/vsphere.go
zreigz: > Oh, that is definitely very confusing. Could we probably rename it to `${Thing}WithPresetEndpoint` or something the like?

I've created a task for this: https://app.zenhub.com/workspaces/all-kubermatic-58863654794153994b62af76/issues/kubermatic/kubermatic/4061

api/pkg/handler/v1/provider/vsphere.go
kgroschoff: datacenter name: yes & no. 
We only know the name that is specified in `\dc` endpoint, e.g. `vsphere-hamburg`. But folder - in our case - uses `dc-1`. So there is a "internal" name as well. We don't know the last one in `api/pkg/handler/v1/provider/vsphere.go`, but we get name in here `api/pkg/provider/cloud/vsphere/provider.go`.
Not sure, if it makes sense to also return dc name from `provider.go` just to have it in `vsphere.go` 🤔 

Regarding your nested example: I already asked before I started, where vms could be created. So far, everyone told me, that it's only possible under `/vm`. Nevertheless good point, as your example also contains `/vm`. Will do some more research on this topic, to find out, if this would be possible. 👍

mrIncompetent: I had to spend some time on that matter today as well.

I've added the filtering by `/datacenter/vm` using `string.HasPrefix` in https://github.com/kubermatic/kubermatic/pull/4068

I also added a simple E2E test to cover the filtering.

alvaroaleman: Okay, would you mind to wait for #4068 to merge and then use the logic in there for filtering?

kgroschoff: Yes, that was the plan 👍

mrIncompetent: So this filtering can now be removed :) 
@kgroschoff should i or do you wan't to?

kgroschoff: I could do it after lunch :)

zreigz:
/retest
the `pull-kubermatic-verify ` is already fixed but your PR is not rebased so `/retest` should help

alvaroaleman:
> the pull-kubermatic-verify is already fixed but your PR is not rebased so /retest should help

For a change that was actually a legitimate failure here :P

/lgtm

PR #4056, opened 2019-08-13
Verify deployment postsubmit in a presubmit

alvaroaleman:

The purpose of this PR combined with an additional presubmit is to have CI notify us about mistakes in the deployment scripting. It also adds some `TEST_NAME` variables so we get nice Junit output in Prow telling us which steps where executed, their state and how long it took.

Fixes #

**Documentation**:

```release-note
none
```

/assign @xrstf @mrIncompetent @nikhita

### DIFF
--- a/api/hack/deploy.sh
+++ b/api/hack/deploy.sh
@@ -39,7 +39,7 @@ function deploy {
 
   TEST_NAME="[Helm] Deploy chart ${name}"
 
-	inital_revision="$(retry 5 helm list --tiller-namespace=${TILLER_NAMESPACE} ${name} --output=json|jq '.Releases[0].Revision')"
+  inital_revision="$(retry 5 helm list --tiller-namespace=${TILLER_NAMESPACE} ${name} --output=json|jq '.Releases[0].Revision')"
 
   echodate "Upgrading ${name}..."
   retry 5 helm --tiller-namespace ${TILLER_NAMESPACE} upgrade --install --atomic --timeout $timeout ${MASTER_FLAG} ${HELM_EXTRA_ARGS} --values ${VALUES_FILE} --namespace ${namespace} ${name} ${path}
--- a/api/hack/deploy.sh
+++ b/api/hack/deploy.sh
@@ -103,7 +103,13 @@ case "${DEPLOY_STACK}" in
       deploy "certs" "default" ./config/certs/
       deploy "oauth" "oauth" ./config/oauth/
       # We might have not configured IAP which results in nothing being deployed. This triggers https://github.com/helm/helm/issues/4295 and marks this as failed
-      deploy "iap" "iap" ./config/iap/ || true
+      # We hack around this by grepping for a string that is mandatory in the values file of IAP
+      # to determine if its configured
+      if grep -q discovery_url ${VALUES_FILE}; then
+        deploy "iap" "iap" ./config/iap/
+      else
+        echodate "Skipping IAP deployment because discovery_url is unset in values file"
+      fi
     fi
 
     # CI has its own Minio deployment as a proxy for GCS, so we do not install the default Helm chart here.
--- a/api/hack/deploy.sh
+++ b/api/hack/deploy.sh
@@ -37,18 +37,31 @@ function deploy {
   local path=$3
   local timeout=${4:-300}
 
+  TEST_NAME="[Helm] Deploy chart ${name}"
+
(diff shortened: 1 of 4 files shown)


alvaroaleman:
/hold
Until https://github.com/kubermatic/infra/pull/527 merged and we verified that it works with the changes here

api/hack/deploy.sh
```diff
@@ -37,18 +37,31 @@ function deploy {
...
 }
 
 function initTiller() {
+  TEST_NAME="[Helm] Init Tiller"
   echodate "Initializing Tiller in namespace ${TILLER_NAMESPACE}"
-  # In clusters which have not been initialized yet, this will fail
-  helm version --tiller-namespace ${TILLER_NAMESPACE} || true
-  kubectl create serviceaccount -n ${TILLER_NAMESPACE} tiller-sa || true
-  kubectl create clusterrolebinding tiller-cluster-role --clusterrole=cluster-admin --serviceaccount=${TILLER_NAMESPACE}:tiller-sa  || true
+  helm version --client --tiller-namespace ${TILLER_NAMESPACE}
```
xrstf: No need to specify `--tiller-namespace` when only asking for the client version.

api/hack/deploy.sh
```diff
@@ -37,18 +37,31 @@ function deploy {
   local path=$3
   local timeout=${4:-300}
 
+  TEST_NAME="[Helm] Deploy chart ${name}"
+
+	inital_revision="$(retry 5 helm list --tiller-namespace=${TILLER_NAMESPACE} ${name} --output=json|jq '.Releases[0].Revision')"
```
alvaroaleman: Will this work as expected, aka the `retry` will execute both the helm command and the `|jq` or will the `|jq` be executed over the result of all helm tries?

xrstf: i would guess the latter, so you would have to do a

```
initial_revision="$(retry 5 sh -c "helm list ... | jq ....")"
```

Shell FTW, eh?

thz: It actually makes no difference as the semantics are the same.
`retry 5 foo | bar` is like `retry 5 sh -c 'foo | bar'` unless your failing tries output anything.

So I'd say we avoid the double double quoting.

xrstf:
@thz  you saw https://github.com/kubermatic/kubermatic/pull/4056#discussion_r313316993 ?

api/hack/deploy.sh
xrstf: Okay then.

alvaroaleman:
/hold
To make sure the `pull-kubermatic-canary-deployment-ci-kubermatic-io ` actually passed before we merge this

api/hack/deploy.sh
thz: This still assumes that the helm command does not write to stdout on failure. So giving json but exiting non-zero would break this.

api/hack/push_image.sh
```diff
@@ -13,6 +13,10 @@ fi
 
 cd "$(dirname "$0")/../"
 
+# Needed because docker-in-docker results in a reduced MTU and fastly, the hoster of the APK
+# mirros, blocks ICMP. See https://github.com/gliderlabs/docker-alpine/issues/307#issuecomment-427246497
+iptables -t mangle -A POSTROUTING -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
```
thz: But when ICMP is blocked then the pmtu result is not available. Shouldn't you clamp on some explicit value then? 

Also: shouldn't that clamping be in the responsibility of the element causing the restriction? So this script is run in the dind already, right?

alvaroaleman: No, this script runs docker in a docker container managed by k8s. Good point about the missing pmtu

thz: So the pmtu can get good values even with blocked icmp. So the value is sane.
The location should be the start-docker.sh script which is setting up the dind.

api/hack/deploy.sh
```diff
@@ -90,7 +103,13 @@ case "${DEPLOY_STACK}" in
       deploy "certs" "default" ./config/certs/
       deploy "oauth" "oauth" ./config/oauth/
       # We might have not configured IAP which results in nothing being deployed. This triggers https://github.com/helm/helm/issues/4295 and marks this as failed
-      deploy "iap" "iap" ./config/iap/ || true
+      # We hack around this by grepping for a string that is mandatory in the values file of IAP
+      # to determine if its configured
```
xrstf: please add `We need to do this because deploying an empty chart leads to Helm doing weird things.`

PR #4058, opened 2019-08-13
add script to verify generated api client

### LINKED ISSUE #4057: Add script to verify generated API client code
**User Story**
Script checks if the user made changes and there is a need to regenerate the API client.

zreigz:
**What this PR does / why we need it**: add the script to verify generated API client

Fixes #4057

```release-note
NONE
```

### DIFF
--- a/api/hack/verify-api-client.sh
+++ b/api/hack/verify-api-client.sh
@@ -3,7 +3,6 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x
 
 function cleanup() {
     if [[ -n "${TMP_DIFFROOT:-}" ]]; then
@@ -20,12 +19,13 @@ TMP_DIFFROOT=$(mktemp -d)
 
 cp -a "${DIFFROOT}"/* "${TMP_DIFFROOT}"
 
-"${SCRIPT_ROOT}/hack/gen-api-client.sh"
+"${SCRIPT_ROOT}/hack/gen-api-client.sh" &>/dev/null
 
 echo "diffing ${DIFFROOT} against freshly generated api client"
 ret=0
 diff -Naupr "${DIFFROOT}" "${TMP_DIFFROOT}" || ret=$?
-cp -a "${TMP_DIFFROOT}"/* "${DIFFROOT}"
+cp -a "${TMP_DIFFROOT}"/client "${DIFFROOT}"
+cp -a "${TMP_DIFFROOT}"/models "${DIFFROOT}"
 
 if [[ $ret -eq 0 ]]
 then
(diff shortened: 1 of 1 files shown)


api/hack/verify-api-client.sh
```diff
@@ -0,0 +1,36 @@
...
+TMP_DIFFROOT=$(mktemp -d)
+
+cp -a "${DIFFROOT}"/* "${TMP_DIFFROOT}"
+
+"${SCRIPT_ROOT}/hack/gen-api-client.sh"
+
+echo "diffing ${DIFFROOT} against freshly generated api client"
+ret=0
+diff -Naupr "${DIFFROOT}" "${TMP_DIFFROOT}" || ret=$?
+cp -a "${TMP_DIFFROOT}"/* "${DIFFROOT}"
```
alvaroaleman: I wouldn't expect a `verify` script to overwrite anything

alvaroaleman: Also you can just remove the `|| ret=$?` and setting it to `0` on line 26 and instead just do a `ret=$?` directly after the diff command

alvaroaleman (commented):
Some nitpicks, I do realize this is copied from the existing script, would you mind to update the existing script to also not do changes and verify only?

works fine thought, thanks for doing this!

api/hack/verify-api-client.sh
```diff
@@ -0,0 +1,36 @@
...
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
+API_DIR="$(go env GOPATH)/src/github.com/kubermatic/kubermatic/api"
+DIFFROOT="${API_DIR}/pkg/test/e2e/api/utils/apiclient"
+
+TMP_DIFFROOT=$(mktemp -d)
+
+cp -a "${DIFFROOT}"/* "${TMP_DIFFROOT}"
+
+"${SCRIPT_ROOT}/hack/gen-api-client.sh"
```
alvaroaleman: Somehow my other comment got lost: Currently the codegen generates an awful lot of output that makes seeing the actual error hard, would you mind to:
* Remove the `set -x`
* Suffix line 23 with a `&>/dev/null` to discard the output of the client generation?

api/hack/verify-api-client.sh
zreigz: when files are different the pure command `diff -Naupr "${DIFFROOT}" "${TMP_DIFFROOT}"` exit with `1` and next lines are never executed.

api/hack/verify-api-client.sh
zreigz: done

zreigz:
@alvaroaleman PTAL

Issue #4053, opened 2019-08-13
Enable Pod Security Policies


scheeles:
**User Story**
customer cross ref: https://github.com/kubermatic/customer-projects/issues/155

For many customers, it would be preferable to have Kubernetes PSP enabled. The feature is available as beta in Kubernetes 1.15 and requires an admission controller to be configured. 

We can not enable this for all cluster especially for existing cluster because this can break running workload.

As a first step it would be great if we can activate it on a per cluster basis and afterwards customers can decide if they want to enable it per default.

**Acceptance criteria**

- [ ] per cluster the admission controller can be enabled
- [ ] It can be configured if the default for the admission controller is enabled or disabled

scheeles:
1. Enhance the cluster resource with a flag for  enable or disable the admission controller 
2. Create a config to enable/disable the PSP per Datacenter

@mrIncompetent @alvaroaleman @kdomanski Please add your thoughts

alvaroaleman:
Acceptance criteria:

* There is an option in the CREATE and UPDATE cluster endpoint to enable/disable PSPs
* On the Cluster CRD there is a field that indicates whether PSPs are enabled
* If PSPs are enabled, the relevant admission controller on the APIServer is set
* There is an E2E tests that runs with PSPs enabled and passes
* There is an E2E tests that runs with PSPs disabled and passes
* It is possible to enforce one of enabled/disabled on a per-Datacenter level

scheeles:
@alvaroaleman 
I would add
- There is an E2E tests that runs with PSPs disabled and passes

bashofmann:
We are currently implementing this in our fork. If you want I can create a PR for this later today.

alvaroaleman:
> We are currently implementing this in our fork. If you want I can create a PR for this later today.

That would be super awesome!

bashofmann:
I submitted https://github.com/kubermatic/kubermatic/pull/4062 to make the admission plugins configurable through the api. What is still missing:

* Dashboard support to easily activate and deactivate PodSecurityPolicy (we can provide a PR for this as well)
* Adding PodSecurityPolicies and corresponding RBAC bindings for all addons
* e2e tests

toschneck:
> Dashboard support to easily activate and deactivate PodSecurityPolicy (we can provide a PR for this as well)

@bashofmann this would be awesome as well :heart_eyes:

alvaroaleman:
[...]

PR #4062, opened 2019-08-13
Allow enabling PSPs

bashofmann:
Make additional admission plugins configurable per cluster. This can be used to for example enable Pod Security Policies on a per cluster level. See #4053.

What is still missing to make this complete:
* Dashboard support to easily activate and deactivate PodSecurityPolicy (we can provide a PR for this as well)
* Adding PodSecurityPolicies and corresponding RBAC bindings for all addons
* e2e tests

```release-note
Pod Security Policies can now be enabled
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -5534,6 +5534,11 @@
         "oidc": {
           "$ref": "#/definitions/OIDCSettings"
         },
+        "usePodSecurityPolicyAdmissionPlugin": {
+          "description": "If active the PodSecurityPolicy admission plugin is configured at the apiserver",
+          "type": "boolean",
+          "x-go-name": "UsePodSecurityPolicyAdmissionPlugin"
+        },
         "version": {
           "$ref": "#/definitions/Semver"
         }
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -565,16 +565,20 @@ type ClusterSpec struct {
 
 	// OIDC settings
 	OIDC kubermaticv1.OIDCSettings `json:"oidc,omitempty"`
+
+	// If active the PodSecurityPolicy admission plugin is configured at the apiserver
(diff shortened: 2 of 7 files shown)


alvaroaleman:
@bashofmann What is the reason you went for allowing any additional admission plugin instead of just PSPs?

bashofmann:
> @bashofmann What is the reason you went for allowing any additional admission plugin instead of just PSPs?

@alvaroaleman I was thinking that it's better to make this API more generic, so that we don't have to create a new dedicated field every time we want to make another admission plugin configurable, but I'm also fine with doing it differently, however you prefer.

alvaroaleman:
> @alvaroaleman I was thinking that it's better to make this API more generic, so that we don't have to create a new dedicated field every time we want to make another admission plugin configurable, but I'm also fine with doing it differently, however you prefer.

Then we also need an extended validation that it isn't already in the list. I'd tend to prefer just making this a binary option for now, we can still make it extensible later on if we actually have that usecase or WDYT?

bashofmann:
What about `UsePodSecurityPolicyAdmissionPlugin`?

alvaroaleman:
/lgtm
/approve

@bashofmann Do you already have the necesary changes for the addons?

bashofmann:
I'm not allowed to see the test results. Is the failing `pull-kubermatic-canary-deployment-ci-kubermatic-io` something I need to fix?

alvaroaleman:
No, the job currently has issues, we'll trigger a re-run once thats fixed

bashofmann:
> @bashofmann Do you already have the necesary changes for the addons?

Yes, I'll try to make a PR for this (as well as for the UI integration) later today or on Monday.

alvaroaleman:
/hold
Until the test is fixed

alvaroaleman:
/retest
/hold cancel

> Yes, I'll try to make a PR for this (as well as for the UI integration) later today or on Monday.

Perfect! Lets add and verify a presubmit that runs with PSPs enabled on that PR then :)

PR #4065, opened 2019-08-13
drop EOL Kubernetes 1.12

kdomanski:

```release-note
Kubernetes 1.12 which is end-of-life has been removed.
```


thz:
/lgtm
/approve
In a follow-up we need to remove more vulnerable versions.

PR #4068, opened 2019-08-13
vSphere folder handling: restrict by rootPath + Add E2E tests

mrIncompetent:
This PR will make sure we filter VM folders by the configured root path & it adds a E2E test.

```release-note
NONE
```


api/pkg/provider/cloud/vsphere/provider_test.go
```diff
@@ -0,0 +1,87 @@
+// +build e2e
```
mrIncompetent: @alvaroaleman Objections against this approach?
It felt like a very simple way to test the vCenter API without having to do the full cluster creation dance

mrIncompetent: I'll add a presubmit if we agree that this is acceptable :)

alvaroaleman: This is perfect, that we currently only have unit or full-blown e2e tests is subpar.

Would you mind using an `integration` build tag instead thought? While it arguably e2e tests the particular functionality, in the scope of the whole product this is not an e2e test and I would like to be able to differentiate this kind of tests from tests that actually create a cluster

alvaroaleman: Oh and also just to be sure, would you mind extending the `test` maketarget to also build this?

mrIncompetent: > Would you mind using an integration build tag instead thought?

:+1: 

> would you mind extending the test maketarget to also build this?

:+1:

mrIncompetent: Done in https://github.com/kubermatic/kubermatic/pull/4068/commits/817ef96b949421519f6cd17dd5e12ea60b7584bb

mrIncompetent:
/hold

Let me fixup the env names so they match what we use in prow

mrIncompetent:
@alvaroaleman PTAL

infra PR incoming

alvaroaleman:
/lgtm
/hold

Can we actually merge the presubmit with `optional: true` and verify on this PR that it works?

mrIncompetent:
@alvaroaleman PTAL

PR #4071, opened 2019-08-14
fix flaky api client verification

zreigz:
**What this PR does / why we need it**: fix flaky API client verification. Stop overriding default values for schemes and host

Fixes #

```release-note
NONE
```


zreigz:
@alvaroaleman the flaky verification fixed. This change doesn't have an impact on tests.

PR #4081, opened 2019-08-15
test vSphere folder creation & adapt code to match common structure

mrIncompetent:
This PR will introduce a integration test for the vSphere folder creation. 
Also it will move some vSphere SDK specific code out of the `provider.go` to keep it smaller/more readable.

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/cloud/vsphere/error.go
+++ b/api/pkg/provider/cloud/vsphere/error.go
@@ -1,0 +1,10 @@
+package vsphere
+
+import (
+	"github.com/vmware/govmomi/find"
+)
+
+func isNotFound(err error) bool {
+	_, ok := err.(*find.NotFoundError)
+	return ok
+}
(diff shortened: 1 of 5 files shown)


alvaroaleman:
Can we run the Vsphere presubmit when we change something there?

As an aside we could write integration tests that just call the cloud providers `initialize` and `cleanup`

mrIncompetent:
> As an aside we could write integration tests that just call the cloud providers initialize and cleanup

Like a single test, that tests every provider "happy-path" ? 
That would be super dope

mrIncompetent:
Testing if i can trigger a test that has `run_if_changed` configured & does not match here.

/test pull-kubermatic-user-cluster-prometheus-config-validation

mrIncompetent:
@alvaroaleman Objections in merging this one?

PR #4096, opened 2019-08-16
Conformance tester: Write junit regardless of test result

alvaroaleman:

Currently the junit report created by the conformance tester will only be written if all tests were completed successfully because otherwise we do not reach the relevant code part.

This PR changes that behaviour to always write out the junit, regardless of the result.

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/cmd/conformance-tests/runner.go
+++ b/api/cmd/conformance-tests/runner.go
@@ -375,11 +375,13 @@ func (r *testRunner) testCluster(
 	var err error
 	totalStart := time.Now()
 	log.Info("Starting to test cluster...")
-	defer log.Infof("Finished testing cluster after %s", time.Since(totalStart))
+
+	// We need the closure to defer the evaluation of the time.Since(totalStart) call
+	defer func() { log.Infof("Finished testing cluster after %s", time.Since(totalStart)) }()
 
 	// Always write junit to disk
-	defer func(totalDurationSeconds float64) {
-		report.Time = totalDurationSeconds
+	defer func() {
+		report.Time = time.Since(totalStart).Seconds()
 		b, err := xml.Marshal(report)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


api/cmd/conformance-tests/runner.go
```diff
@@ -377,6 +377,19 @@ func (r *testRunner) testCluster(
...
+		report.Time = totalDurationSeconds
+		b, err := xml.Marshal(report)
+		if err != nil {
+			log.Errorw("failed to marshal junit", zap.Error(err))
+			return
+		}
+		if err := ioutil.WriteFile(path.Join(r.reportsRoot, fmt.Sprintf("junit.%s.xml", scenarioName)), b, 0644); err != nil {
+			log.Errorw("failed to write junit", zap.Error(err))
+		}
+	}(time.Since(totalStart).Seconds())
```
kdomanski: I think `time.Since(totalStart).Seconds()` will get evaluated instantly instead of at the end. I suggest you only pass `totalStart` and calculate the interval within the function.

xrstf: I don't think this does what you want it to do: https://play.golang.org/p/nFVIVv9FuWk

api/cmd/conformance-tests/runner.go
```diff
@@ -377,6 +377,19 @@ func (r *testRunner) testCluster(
 	log.Info("Starting to test cluster...")
 	defer log.Infof("Finished testing cluster after %s", time.Since(totalStart))
```
xrstf: Likewise, this does not work as expected. It's the same error as we had in the preemtible-manager.

api/cmd/conformance-tests/runner.go
alvaroaleman: I somehow thought it was the other way round

PR #4099, opened 2019-08-17
Extend Seed CR Validation

### LINKED ISSUE #3888: Add validatingAdmissionWebhook for Seed CRD

xrstf:
This implements the remaining validation features for seed CRs. It also converts the existing `provider.validateDatacenters` function to work based on seeds, i.e. it is now called *after* the conversion has happened. _This also means that datacenters that belong to no seed are not validated anymore,_ as they are dropped anyway.

By accident I also found a typo for a CLI flag and included the fix here because why not.

Part of #2113, fixes #3888

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/datacenter.go
+++ b/api/pkg/provider/datacenter.go
@@ -136,7 +136,7 @@ func validateDatacenters(datacenters map[string]DatacenterMeta) error {
 			continue
 		}
 		if errs := validation.IsDNS1123Subdomain(datacenterName); errs != nil {
-			return fmt.Errorf("Datacentername %q is not a valid DNS name: %v", datacenterName, errs)
+			return fmt.Errorf("Datacenter name %q is not a valid DNS name: %v", datacenterName, errs)
 		}
 	}
 
--- a/api/pkg/provider/datacenter.go
+++ b/api/pkg/provider/datacenter.go
@@ -126,6 +126,8 @@ func ValidateSeed(seed *kubermaticv1.Seed) error {
 		}
 	}
 
+	// invalid DNS overwrites can happen when a seed was freshly converted from
+	// the datacenters.yaml and has not yet been validated
 	if seed.Spec.SeedDNSOverwrite != nil && *seed.Spec.SeedDNSOverwrite != "" {
 		if errs := validation.IsDNS1123Subdomain(*seed.Spec.SeedDNSOverwrite); errs != nil {
 			return fmt.Errorf("DNS overwrite %q is not a valid DNS name: %v", *seed.Spec.SeedDNSOverwrite, errs)
--- a/api/pkg/provider/datacenter.go
+++ b/api/pkg/provider/datacenter.go
@@ -74,16 +74,18 @@ func loadSeeds(path string) (map[string]*kubermaticv1.Seed, error) {
 		return nil, fmt.Errorf("failed to parse datacenters.yaml: %v", err)
 	}
 
-	if err := validateDatacenters(dcMetas.Datacenters); err != nil {
-		return nil, fmt.Errorf("failed to validate datacenters.yaml: %v", err)
-	}
-
-	dcs, err := DatacenterMetasToSeeds(dcMetas.Datacenters)
+	seeds, err := DatacenterMetasToSeeds(dcMetas.Datacenters)
 	if err != nil {
 		return nil, fmt.Errorf("failed to convert datacenters.yaml: %v", err)
 	}
 
-	return dcs, nil
+	for seedName, seed := range seeds {
+		if err := ValidateSeed(seed); err != nil {
+			return nil, fmt.Errorf("failed to validate datacenters.yaml: seed %q is invalid: %v", seedName, err)
+		}
+	}
+
+	return seeds, nil
 }
 
 func LoadSeed(path, datacenterName string) (*kubermaticv1.Seed, error) {
@@ -110,8 +112,8 @@ func validateImageList(images kubermaticv1.ImageList) error {
 	return nil
 }
 
-func validateDatacenters(datacenters map[string]DatacenterMeta) error {
-	for name, dc := range datacenters {
+func ValidateSeed(seed *kubermaticv1.Seed) error {
+	for name, dc := range seed.Spec.Datacenters {
 		if dc.Spec.VSphere != nil {
 			if err := validateImageList(dc.Spec.VSphere.Templates); err != nil {
 				return fmt.Errorf("invalid datacenter defined '%s': %v", name, err)
@@ -124,19 +126,13 @@ func validateDatacenters(datacenters map[string]DatacenterMeta) error {
 		}
 	}
 
-	for datacenterName, datacenter := range datacenters {
-		if !datacenter.IsSeed {
-			continue
+	if seed.Spec.SeedDNSOverwrite != nil && *seed.Spec.SeedDNSOverwrite != "" {
+		if errs := validation.IsDNS1123Subdomain(*seed.Spec.SeedDNSOverwrite); errs != nil {
+			return fmt.Errorf("DNS overwrite %q is not a valid DNS name: %v", *seed.Spec.SeedDNSOverwrite, errs)
 		}
-		if datacenter.SeedDNSOverwrite != nil && *datacenter.SeedDNSOverwrite != "" {
-			if errs := validation.IsDNS1123Subdomain(*datacenter.SeedDNSOverwrite); errs != nil {
-				return fmt.Errorf("SeedDNS overwrite %q of datacenter %q is not a valid DNS name: %v",
-					*datacenter.SeedDNSOverwrite, datacenterName, errs)
-			}
-			continue
-		}
-		if errs := validation.IsDNS1123Subdomain(datacenterName); errs != nil {
-			return fmt.Errorf("Datacenter name %q is not a valid DNS name: %v", datacenterName, errs)
+	} else {
+		if errs := validation.IsDNS1123Subdomain(seed.Name); errs != nil {
+			return fmt.Errorf("seed name %q is not a valid DNS name: %v", seed.Name, errs)
 		}
 	}
 
--- a/api/pkg/provider/datacenter_test.go
+++ b/api/pkg/provider/datacenter_test.go
@@ -116,47 +116,43 @@ datacenters:
 	assert.Equal(t, expectedSeeds, resultDatacenters)
 }
 
-func TestValidateSeed(t *testing.T) {
+func TestMigrateDatacenters(t *testing.T) {
 	testCases := []struct {
 		name        string
-		seed        *kubermaticv1.Seed
+		datacenters map[string]DatacenterMeta
 		errExpected bool
 	}{
 		{
 			name: "Invalid name, error",
-			seed: &kubermaticv1.Seed{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "&invalid",
+			datacenters: map[string]DatacenterMeta{
+				"&invalid": {
+					IsSeed: true,
 				},
 			},
 			errExpected: true,
 		},
 		{
 			name: "Valid name succeeds",
-			seed: &kubermaticv1.Seed{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "valid-name",
+			datacenters: map[string]DatacenterMeta{
+				"valid": {
+					IsSeed: true,
 				},
 			},
 		},
 		{
 			name: "Invalid name, valid seed dns override",
-			seed: &kubermaticv1.Seed{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "&invalid",
-				},
-				Spec: kubermaticv1.SeedSpec{
+			datacenters: map[string]DatacenterMeta{
+				"&invalid": {
+					IsSeed:           true,
 					SeedDNSOverwrite: utilpointer.StringPtr("valid"),
 				},
 			},
 		},
 		{
 			name: "Valid name, invalid seed dns override",
-			seed: &kubermaticv1.Seed{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "valid-name",
-				},
-				Spec: kubermaticv1.SeedSpec{
+			datacenters: map[string]DatacenterMeta{
+				"valid": {
+					IsSeed:           true,
 					SeedDNSOverwrite: utilpointer.StringPtr("&invalid"),
 				},
 			},
@@ -166,8 +162,15 @@ func TestValidateSeed(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			if err := ValidateSeed(tc.seed); (err != nil) != tc.errExpected {
-				t.Fatalf("Expected err: %t, but got err: %v", tc.errExpected, err)
+			seeds, err := DatacenterMetasToSeeds(tc.datacenters)
+			if err != nil {
+				t.Fatalf("Failed to convert datacenters to seeds: %v", err)
+			}
+
+			for _, seed := range seeds {
+				if err := ValidateSeed(seed); (err != nil) != tc.errExpected {
+					t.Fatalf("Expected err: %t, but got err: %v", tc.errExpected, err)
+				}
 			}
 		})
 	}
--- a/api/pkg/provider/datacenter_test.go
+++ b/api/pkg/provider/datacenter_test.go
@@ -158,6 +158,12 @@ func TestMigrateDatacenters(t *testing.T) {
 			},
 			errExpected: true,
 		},
+		{
+			name: "Invalid name, but is not a seed",
+			datacenters: map[string]DatacenterMeta{
+				"&invalid": {},
+			},
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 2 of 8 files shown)


api/pkg/validation/seed/server.go
```diff
@@ -140,6 +142,11 @@ func (s *Server) handle(req *http.Request) (*admissionv1beta1.AdmissionRequest,
 		}
 	}
 
+	// ignore requests for non-matcher worker names
+	if seed.Labels[kubermaticv1.WorkerNameLabelKey] != s.workerName {
```
alvaroaleman: This shouldn't be needed, the webhookconfig itself contains a labelselector

api/pkg/provider/datacenter.go
```diff
@@ -120,19 +122,13 @@ func validateDatacenters(datacenters map[string]DatacenterMeta) error {
...
-			}
-			continue
+	if seed.Spec.SeedDNSOverwrite != nil && *seed.Spec.SeedDNSOverwrite != "" {
+		if errs := validation.IsDNS1123Subdomain(*seed.Spec.SeedDNSOverwrite); errs != nil {
+			return fmt.Errorf("DNS overwrite %q is not a valid DNS name: %v", *seed.Spec.SeedDNSOverwrite, errs)
 		}
-		if errs := validation.IsDNS1123Subdomain(datacenterName); errs != nil {
-			return fmt.Errorf("Datacentername %q is not a valid DNS name: %v", datacenterName, errs)
+	} else {
+		if errs := validation.IsDNS1123Subdomain(seed.Name); errs != nil {
```
alvaroaleman, line 136: Could you add a comment explaining that this is only needed because the seed may come from a file and not from a CRD in which case ppl could give it a non-dns123subdomain conformant name?

api/pkg/provider/datacenter.go
```diff
@@ -120,19 +122,13 @@ func validateDatacenters(datacenters map[string]DatacenterMeta) error {
...
-			continue
-		}
-		if datacenter.SeedDNSOverwrite != nil && *datacenter.SeedDNSOverwrite != "" {
-			if errs := validation.IsDNS1123Subdomain(*datacenter.SeedDNSOverwrite); errs != nil {
-				return fmt.Errorf("SeedDNS overwrite %q of datacenter %q is not a valid DNS name: %v",
-					*datacenter.SeedDNSOverwrite, datacenterName, errs)
-			}
-			continue
+	if seed.Spec.SeedDNSOverwrite != nil && *seed.Spec.SeedDNSOverwrite != "" {
+		if errs := validation.IsDNS1123Subdomain(*seed.Spec.SeedDNSOverwrite); errs != nil {
```
alvaroaleman, line 132: I am btw fine renaming this to DNSOverwrite, just the name and espcially the json marker in the old `DatacenterMeta` type must ofc stay as-is

api/pkg/validation/seed/seed.go
```diff
@@ -58,30 +59,68 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
...
 
-func (sv *seedValidator) validate(seed *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, existingSeeds map[string]*kubermaticv1.Seed, isDelete bool) error {
+func (sv *seedValidator) validate(seed *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, seeds map[string]*kubermaticv1.Seed, isDelete bool) error {
+	existingSeed := seeds[seed.Name]
 
-	newDatacenters := sets.NewString()
+	// remove the seed itself from the list, so uniqueness checks won't fail
+	delete(seeds, seed.Name)
+
+	affectedDatacenters := sets.NewString()
```
alvaroaleman: what about `ourDatacenters` and `allDatacenters`?

api/pkg/validation/seed/seed.go
```diff
@@ -58,30 +59,68 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
...
+	for dcName, dc := range seed.Spec.Datacenters {
+		providerName, err := provider.DatacenterCloudProviderName(&dc.Spec)
+		if err != nil {
+			return fmt.Errorf("datacenter %q is invalid: %v", dcName, err)
+		}
+		if providerName == "" {
+			return fmt.Errorf("datacenter %q has no provider defined", dcName)
+		}
+
+		if existingSeed != nil {
```
alvaroaleman: Can we make this a bit less of an arrow code and do it something like this:
```
if existingSeed == nil {
  continue
}
if _, alreadyExists := existingSeed.Spec.Datacenters[dcName]; !alreadyExists {
  continue
}
if existingProvider, _ := provider.DatacenterCloudProviderName(&existingDC.Spec); existingProvider != provider {
  retun err
}
```
?

api/pkg/validation/seed/seed_test.go
```diff
@@ -22,60 +22,311 @@ func init() {
...
-						"in-use": {},
+						"dc1": {
+							Spec: fakeProviderSpec,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Adding new datacenter should be possible",
```
alvaroaleman, line 86: The test name is misleading, this is an update of an existing datacenter or not?

api/pkg/validation/seed/seed_test.go
```diff
@@ -22,60 +22,311 @@ func init() {
...
+						},
+						"dc2": {
+							Spec: fakeProviderSpec,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Should be able to remove unused datacenters from a seed",
```
alvaroaleman, line 118: This adds a new seed without datacenters instead of updating an existing seed or not?

api/pkg/validation/seed/seed_test.go
```diff
@@ -22,60 +22,311 @@ func init() {
...
+					},
+				},
+			},
+			seedToValidate: &kubermaticv1.Seed{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "existing-seed",
+				},
+				Spec: kubermaticv1.SeedSpec{
+					Datacenters: map[string]kubermaticv1.Datacenter{
+						"dc1": {},
```
alvaroaleman: this could also fail because there is no provider anymore, can we actually add a different provider?

api/pkg/validation/seed/seed_test.go
```diff
@@ -85,12 +336,14 @@ func TestValidate(t *testing.T) {
 				listOpts: &ctrlruntimeclient.ListOptions{},
 			}
 
+			existingClusters := tc.existingClusters[tc.seedToValidate.Name]
```
alvaroaleman: Why do we make the clusters a map and only consider the ones that have a matching seed? This is confusing and since we always only consider exactly one element of the map, we can just keep it as a slice

alvaroaleman (commented):
> This also means that datacenters that belong to no seed are not validated anymore, as they are dropped anyway.

Can we add a validation when reading the `datacenters.yaml` that fails in such a case?

api/pkg/validation/seed/server.go
xrstf: Are you sure? Right now I cannot create seed CRs, because the webhook will always react to my new seed (even if the seed has a worker-label). It's true that is does have a worker-label handling, but it's not enough.

api/pkg/validation/seed/seed.go
xrstf: I don't think this makes it any easier to read or comprehend. Quite the opposite, now I would have returns *and* continues in my loop. But alas...

api/pkg/validation/seed/seed_test.go
xrstf: No, this is adding a new datacenter, `dc2`, to an existing seed, not modifying the existing `dc1` datacenter. It's adding a new datacenter to a seed.

api/pkg/validation/seed/seed_test.go
xrstf: Good catch, that's a typo.

api/pkg/validation/seed/seed_test.go
xrstf: Because otherwise we would break the code. The validator assumes that the seedGetter() is returning it a connection to the seed that is being validated. If the connection however returns effectively *all* clusters in *all* seeds, then the validation is broken, because it tries to answer the "what clusters are still remaining in this seed?" question.

api/pkg/validation/seed/seed.go
xrstf: `affected` is clearer than `our`, IMHO.

api/pkg/validation/seed/seed.go
xrstf: It really feels like I just did converted the left logic to the right, just for the sake of "not having arrow code".

Untitled Diagram-1

api/pkg/validation/seed/server.go
alvaroaleman: Then the  webhook config is broken and should be fixed. This should absolutely not get filtered at the webhook level but at the apiserver level to prevent a single defunct webhook from causing all seed operations to fail

xrstf: /shrug

api/pkg/validation/seed/seed.go
alvaroaleman: affected by what? A disease? I literally had to rename that var in my had to be able to follow up on the code. I am open to more suggestions but the current name is just confusing.

xrstf: Affected by the change that we are supposed to validate. Who is "our"? Is "our" the data in etcd or is "us" the seed that is being validated? Am I the seed or the validator?

api/pkg/validation/seed/seed.go
alvaroaleman: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

Your code may not be error handling but exclude handling, but the same principles apply IMHO

api/pkg/validation/seed/seed_test.go
alvaroaleman: My point being: This is super confusing for both ppl writing the tests and ppl reading the tests. The tests should not have weird and hard to reason about code just to be similiar to the implementation they test.

xrstf: The test does what the real world does. But fine, I can change it.

xrstf:
Tests now fail. Please advise on how to fix the validation logic to handle "fetching clusters in seed A can yield clusters from seed B".

xrstf:
> Can we add a validation when reading the datacenters.yaml that fails in such a case?

What do you mean exactly? Failing if the are datacenters in the datacenters.yaml that are unused? That sounds harsh.

alvaroaleman:
> Tests now fail. Please advise on how to fix the validation logic to handle "fetching clusters in seed A can yield clusters from seed B".

I do not follow. The client we pass into the `validate` is the client for the seed we are currently validating. Why do the tests define clusters that are in a different seed? They are irrelevant for our validation and should obviously not get returned by the client for the current seed or am I completely missing something?

> What do you mean exactly? Failing if the are datacenters in the datacenters.yaml that are unused? That sounds harsh.

Why is it harsh? It tells the operator that something is misconfigured, IMHO providing feedback is better than silently not working as expected.

Edit: We could probably add a release note mentioning that the validation is now stricter

alvaroaleman:
@xrstf can we resolve the conflicts and try to get this merged? :)

xrstf:
> I do not follow. The client we pass into the validate is the client for the seed we are currently validating. Why do the tests define clusters that are in a different seed?

You're right, that was a brainfart.

api/pkg/provider/datacenter_test.go
```diff
@@ -116,59 +116,57 @@ datacenters:
 	assert.Equal(t, expectedSeeds, resultDatacenters)
 }
 
-func TestValidateDataCenters(t *testing.T) {
+func TestValidateSeed(t *testing.T) {
```
alvaroaleman: Would it be okay to leave the tests to contain the `map[string]DatacenterMeta` and run the conversion on it before validating? I think that would make me feel a bit more confident

api/pkg/validation/seed/seed.go
```diff
@@ -54,30 +55,73 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
...
-	for _, existingSeed := range existingSeeds {
-		for existingDatacenter := range existingSeed.Spec.Datacenters {
-			if newDatacenters.Has(existingDatacenter) {
-				return fmt.Errorf("datacenter %q already exists in seed %q, can only have one datacenter with a given name", existingDatacenter, existingSeed.Name)
+	if !isDelete {
+		allDatacenters = allDatacenters.Union(ourDatacenters)
+	}
+
+	// check if all the DCs in the seed are unique
+	for _, s := range seeds {
```
alvaroaleman: Would you mind renaming the `seeds` var to something that reflects that the current seed we validate is not in it? `existingSeeds` or `currentSeeds` or sth the like?

api/pkg/validation/seed/seed.go
```diff
@@ -54,30 +55,73 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
...
+		}
+
+		existingDC, ok := existingSeed.Spec.Datacenters[dcName]
+		if !ok {
+			continue
+		}
+
+		existingProvider, _ := provider.DatacenterCloudProviderName(&existingDC.Spec)
+		if providerName != existingProvider {
+			return fmt.Errorf("cannot change datacenter %q provider from %q", dcName, existingProvider)
```
alvaroaleman: Can add the name of the new provider to the error?

api/pkg/validation/seed/seed_test.go
```diff
@@ -32,50 +35,284 @@ func TestValidate(t *testing.T) {
...
+			name: "Should be able to remove unused datacenters from a seed",
 			existingSeeds: map[string]*kubermaticv1.Seed{
 				"existing-seed": {
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "existing-seed",
 					},
 					Spec: kubermaticv1.SeedSpec{
 						Datacenters: map[string]kubermaticv1.Datacenter{
-							"in-use": {},
+							"in-use": {
```
alvaroaleman: Nit: Can we name this from `in-use` to anything else? Because its not actually in use which is a bit confusing to read

api/pkg/validation/seed/server.go
```diff
@@ -58,6 +58,7 @@ func (opts *WebhookOpts) Server(
 		listenAddress: opts.ListenAddress,
 		certFile:      opts.CertFile,
 		keyFile:       opts.KeyFile,
+		workerName:    workerName,
```
alvaroaleman: this var is unused

api/pkg/validation/seed/seed.go
xrstf: `existingSeeds` it is.

api/pkg/validation/seed/server.go
xrstf: Good catch, fixed.

api/pkg/validation/seed/seed_test.go
xrstf: We can, should and did rename it.

api/pkg/validation/seed/seed.go
xrstf: Been there, done that.

api/pkg/provider/datacenter_test.go
xrstf: It feels like mixing the migration and the validation in one test, but fine by me.

api/pkg/provider/datacenter_test.go
```diff
@@ -158,18 +158,19 @@ func TestValidateDataCenters(t *testing.T) {
 			},
 			errExpected: true,
 		},
-		{
-			name: "Invalid name, but is not a seed",
```
alvaroaleman, line 162: Why is this testcase removed?

alvaroaleman (commented):
one question left

api/pkg/provider/datacenter_test.go
xrstf: By accident. When I reworked the tests to be based on seeds, this testcase didn't make any sense. When restoring the datacenters test I only translated the remaining tests back to their datacenters form.

xrstf: Tests fail now (doh!) because we require datacenters to form a sensible set of seeds and node datacenters, where previously we only validated the DCs on their own. What direction do you want to go with this? `errExpected: true` or remove the test?

xrstf:
This took longer than expected because my IDE and me were locked in a trench war. And I did not expect that we had tests checking specific error strings ^^

alvaroaleman:
The cat pics completely justify delays here.
/lgtm

PR #4100, opened 2019-08-17
Implement Seed-Sync Controller

### LINKED ISSUE #3889: Add controller that synchronizes Seed CRs from the master cluster to the seed
**User Story**

As an operator of Kubermatic I want to create my `Seed` customResources in one kubernetes cluster only.

**Acceptance criteria**

* There is a controller that sync the `seed` CR from the master cluster onto the corresponding seed

xrstf:
This is a new controller that runs inside the master cluster and copies the seed CRs into their respective clusters.

Part of #2113, fixes #3889

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/controllers.go
+++ b/api/cmd/master-controller-manager/controllers.go
@@ -15,7 +15,6 @@ import (
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/prometheus/client_golang/prometheus"
-	"go.uber.org/zap"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -24,7 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-func createAllControllers(ctrlCtx *controllerContext, log *zap.SugaredLogger) error {
+func createAllControllers(ctrlCtx *controllerContext) error {
 	rbacControllerFactory := rbacControllerFactoryCreator(
 		ctrlCtx.mgr.GetConfig(),
 		ctrlCtx.seedsGetter,
@@ -48,7 +47,7 @@ func createAllControllers(ctrlCtx *controllerContext, log *zap.SugaredLogger) er
 	if err := serviceaccount.Add(ctrlCtx.mgr); err != nil {
 		return fmt.Errorf("failed to create serviceaccount controller: %v", err)
 	}
-	if err := seedsync.Add(ctrlCtx.mgr, 1, log, ctrlCtx.namespace, ctrlCtx.seedKubeconfigGetter); err != nil {
+	if err := seedsync.Add(ctrlCtx.mgr, 1, ctrlCtx.log, ctrlCtx.namespace, ctrlCtx.seedKubeconfigGetter); err != nil {
 		return fmt.Errorf("failed to create seedsync controller: %v", err)
 	}
 	if err := seedproxy.Add(ctrlCtx.mgr, 1, ctrlCtx.seedsGetter, ctrlCtx.seedKubeconfigGetter); err != nil {
--- a/api/cmd/master-controller-manager/controllers.go
+++ b/api/cmd/master-controller-manager/controllers.go
@@ -7,13 +7,15 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac"
 	seedcontrollerlifecycle "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-lifecycle"
 	seedproxy "github.com/kubermatic/kubermatic/api/pkg/controller/seed-proxy"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/service-account"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-project-binding"
+	seedsync "github.com/kubermatic/kubermatic/api/pkg/controller/seed-sync"
+	serviceaccount "github.com/kubermatic/kubermatic/api/pkg/controller/service-account"
+	userprojectbinding "github.com/kubermatic/kubermatic/api/pkg/controller/user-project-binding"
 	kubermaticclientset "github.com/kubermatic/kubermatic/api/pkg/crd/client/clientset/versioned"
 	"github.com/kubermatic/kubermatic/api/pkg/crd/client/informers/externalversions"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -22,7 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-func createAllControllers(ctrlCtx *controllerContext) error {
+func createAllControllers(ctrlCtx *controllerContext, log *zap.SugaredLogger) error {
 	rbacControllerFactory := rbacControllerFactoryCreator(
 		ctrlCtx.mgr.GetConfig(),
 		ctrlCtx.seedsGetter,
@@ -45,6 +47,9 @@ func createAllControllers(ctrlCtx *controllerContext) error {
 	}
 	if err := serviceaccount.Add(ctrlCtx.mgr); err != nil {
 		return fmt.Errorf("failed to create serviceaccount controller: %v", err)
+	}
+	if err := seedsync.Add(ctrlCtx.mgr, 1, log, ctrlCtx.namespace, ctrlCtx.seedKubeconfigGetter); err != nil {
+		return fmt.Errorf("failed to create seedsync controller: %v", err)
 	}
 	if err := seedproxy.Add(ctrlCtx.mgr, 1, ctrlCtx.seedsGetter, ctrlCtx.seedKubeconfigGetter); err != nil {
 		return fmt.Errorf("failed to create seedproxy controller: %v", err)
--- a/api/pkg/controller/seed-sync/controller.go
+++ b/api/pkg/controller/seed-sync/controller.go
@@ -1,0 +1,67 @@
+package seedsync
+
+import (
+	"fmt"
+
+	"go.uber.org/zap"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// ControllerName is the name of this very controller.
+	ControllerName = "seed-sync-controller"
+)
+
+// Add creates a new Seed-Sync controller and sets up Watches
+func Add(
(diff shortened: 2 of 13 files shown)


api/pkg/controller/seed-sync/controller.go
```diff
@@ -0,0 +1,67 @@
...
+
+	ctrlOptions := controller.Options{Reconciler: reconciler, MaxConcurrentReconciles: numWorkers}
+	c, err := controller.New(ControllerName, mgr, ctrlOptions)
+	if err != nil {
+		return err
+	}
+
+	eventHandler := &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+			if a.Meta.GetNamespace() != namespace {
```
alvaroaleman: err while this works I find this a bit confusing, the filtering mechanism for objects are predicates. Is the reason you built it like that that you need less code?

We could probably build a `PredicateFactory(func(m metav1.Object, r runtime.Object) bool)` to make the construction less tedious?

api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	logger := r.log.With("seed", request.Name)
+	logger.Info("Reconciling seed")
+
+	seed := &kubermaticv1.Seed{}
+	if err := r.Get(r.ctx, request.NamespacedName, seed); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get seed: %v", err)
+	}
+
+	if err := r.reconcile(seed, logger); err != nil {
```
alvaroaleman, line 40: It would be nice to also record an event on the seed :)

api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, logger *zap.SugaredLogger) error {
+	kubeconfig, err := r.seedKubeconfigGetter(seed)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve kubeconfig: %v", err)
+	}
+
+	client, err := ctrlruntimeclient.New(kubeconfig, ctrlruntimeclient.Options{})
```
alvaroaleman: We should cache the restMapper, if the seed is geographically far away its construction is very expensive, e.G. constructing the one for asia-dev from europe-dev takes 10-20 seconds.

See https://github.com/kubermatic/kubermatic/blob/f07ffdd01a2309bc2d63f9c91076097fc567c803/api/cmd/kubermatic-api/main.go#L386-L405 , https://github.com/kubermatic/kubermatic/blob/f07ffdd01a2309bc2d63f9c91076097fc567c803/api/pkg/validation/seed/seed.go#L101-L119 and https://github.com/kubermatic/kubermatic/blob/f07ffdd01a2309bc2d63f9c91076097fc567c803/api/pkg/cluster/client/client.go#L59-L75

Also since that would be the fourth time we implement the exact same thing, we could probably add a RestMapperCache type that has a simple stupid `GetMapper(cfg *rest.Config)(meta.RestMapper, error)` and use it in the three aforementioned places and here?

Would probably fit here: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/util/restmapper/dynamic.go

api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create client for seed: %v", err)
+	}
+
+	namespaceCreators := []reconciling.NamedNamespaceCreatorGetter{
+		namespaceCreator(seed.Namespace),
```
alvaroaleman: Err, wouldn't the assumption be that the namesapce already exists, either because of helm or because of the kubermatic operator?

api/pkg/controller/seed-sync/controller.go
xrstf: > err while this works I find this a bit confusing, the filtering mechanism for objects are predicates. Is the reason you built it like that that you need less code?

I build it like this because last time I was told that predicates are extra wasted lines and a simple if inside the event handler would be the better approach.

api/pkg/controller/seed-sync/reconciler.go
xrstf: Inside the seeds? Isn't the seed possibly completely empty?

api/pkg/controller/seed-sync/controller.go
xrstf: cf. https://github.com/kubermatic/kubermatic/pull/3850/commits/f65bc7c031db8cd142ae8442c7e19ef053c285b8#r306857552

api/pkg/controller/seed-sync/reconciler.go
xrstf: Good idea! :)

api/pkg/controller/seed-sync/reconciler.go
alvaroaleman: No, the kubermatic operator or helm is responsible for deploying kubermatic in it. This controller only runs after kubermatic got deployed in the master, its responsibility is to sync changes to the seed into the seed object in the seed cluster, not to install kubermatic there

xrstf: Fair enough. Counterpoint: What's the harm in ensuring that the namespace exists?

alvaroaleman: Masking errors on a different layer, maintaining unneeded code

xrstf: I very strongly disagree that adding these few lines creates a maintenance burden. What I can understand however is the argument that without Kubermatic already running inside the seed, the Seed CRD is not present and the reconciling would fail anyway. ;-) I still think that ensuring the namespace exists is a cheap and safe thing to do, both in terms of coding, maintaining and executing the code.

api/pkg/controller/seed-sync/reconciler.go
xrstf: I've added an event for cases where reconciles failed.

api/pkg/controller/seed-sync/controller.go
xrstf: I've intoduced a new `pkg.controller.util/PredicateFactory()` function and reworked the existing WorkerName predicate helper.

api/pkg/controller/seed-sync/reconciler.go
alvaroaleman: Sidenote: I implemented this in a different PR already

xrstf: Ooouuh, having something like `provider.SeedClientGetterFactory` is what immediately came to my mind as well. Nice. :)

(Nit: I would have not called it this way, because "Factory" does not tell me in any way that the secret sauce of this function is the *caching* aspect. And it's the only function in this whole file that has anything to do with caching. More like `CachedSeedClientGetter()`.)

alvaroaleman: Well the caching is part of it, but the caching is not built specifically for that because its used in other places as well and as long as we dont have a non-caching `SeedClientGetterFactory` I think its a bit confusing

api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, logger *zap.SugaredLogger) error {
+	client, err := r.seedClientGetter(seed)
+	if err != nil {
+		return fmt.Errorf("failed to create client for seed: %v", err)
+	}
+
+	seedCreators := []reconciling.NamedSeedCreatorGetter{
+		seedCreator(seed),
```
alvaroaleman, line 56: Err did you test this? I would except this to error because the seed has some fields set that are readonly and may only be set by the apiserver, like UID, selflink and such

api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if err := reconciling.ReconcileSeeds(r.ctx, seedCreators, "", client); err != nil {
```
alvaroaleman: I think this should not be `""` but the namespace of the seed or not?

api/pkg/controller/util/predicate/predicate.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// Factory returns a predicate func that applies the given filter function
+// on CREATE, UPDATE and DELETE events. For UPDATE events, the the filter is applied
```
alvaroaleman: `the the`

api/pkg/controller/seed-sync/reconciler.go
alvaroaleman: *in a multi seed setup

xrstf: I was actually never able to test this because I could not find a way to create Seed resources anywhere. Maybe the new validation logic that we finally now got merged lets me create seeds now, have to check.

xrstf: You were right, I adjusted the reconciling code.

xrstf:
Added a unit test for the reconciling. Wasn't sure if I should add the boilerplate for table driven tests when we only have one test, but I did it anyway.

api/cmd/master-controller-manager/controllers.go
```diff
@@ -22,7 +24,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-func createAllControllers(ctrlCtx *controllerContext) error {
+func createAllControllers(ctrlCtx *controllerContext, log *zap.SugaredLogger) error {
```
alvaroaleman: nit: The `controllerContext` is specifically made to transport everything you need to initialize the controllers, so just add the `log` there?

api/pkg/controller/seed-sync/reconciler_test.go
```diff
@@ -0,0 +1,140 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlruntimefake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+type nopEventRecorder struct{}
```
alvaroaleman: Is there no fakeEventRecorder somewhere?

api/pkg/controller/util/predicate/predicate.go
```diff
@@ -0,0 +1,32 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// Factory returns a predicate func that applies the given filter function
+// on CREATE, UPDATE and DELETE events. For UPDATE events, the filter is applied
+// to both the old and new object and OR's the result.
+func Factory(filter func(m metav1.Object, r runtime.Object) bool) predicate.Funcs {
```
alvaroaleman, line 13: This is really nice, might even PR it upstream

api/cmd/master-controller-manager/controllers.go
xrstf: Done.

api/pkg/controller/seed-sync/reconciler_test.go
xrstf: There is, but to be fair, I had to look one file over to find it. And the file was named `fake.go`. Nobody could have seen that coming.

alvaroaleman (commented):
/lgtm
/approve

PR #4111, opened 2019-08-19
make the GetNetworks return all possible network adapters within the dc

mrIncompetent:
Make the GetNetworks return all possible network adapters within the DC.
I also took the time and align the client handling with the machine-controller code.

cc @kgroschoff 

/hold 
I need to update the machine-controller before we can test this PR.

```release-note
NONE
```


api/pkg/provider/cloud/vsphere/network.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"path"
+
+	"github.com/vmware/govmomi/object"
+	"k8s.io/apimachinery/pkg/util/runtime"
+)
+
+type NetworkInfo struct {
+	Name string
+	Path string
```
mrIncompetent: @kgroschoff `Path` must be used when creating new MachineDeployments.
Its the absolute path to the network

alvaroaleman: Can we then rename it to `AbsoluePath` :grimacing: ?

mrIncompetent: Done in https://github.com/kubermatic/kubermatic/pull/4111/commits/33518df19ddbc1cfe4da1251fe6902855d1f4d18

mrIncompetent:
@alvaroaleman PTAL

api/pkg/api/v1/types.go
```diff
@@ -434,12 +434,18 @@ type OpenstackSecurityGroup struct {
...
+	RelativePath string `json:"relativePath"`
+	// Type defines the type of network
+	Type string `json:"type"`
 }
 
 // VSphereFolder is the object representing a vsphere folder.
 // swagger:model VSphereFolder
 type VSphereFolder struct {
-	// Path is the path of the folder
+	// RelativePath is the path of the folder
```
moadqassem: The comment here is different than the fields name

api/pkg/provider/cloud/vsphere/network.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	for _, network := range networks {
+		if _, err := network.EthernetCardBackingInfo(ctx); err != nil {
+			// Some network devices cannot be used by VM's.
+			if err == object.ErrNotSupported {
+				continue
+			}
+
+			// Just log the error. We do not want a single weird network to break the entire listing.
+			runtime.HandleError(fmt.Errorf("Failed to get network backing info for %q: %v", network.Reference().String(), err))
```
moadqassem: Can you please lowercase the first word in the error message: s/Failed/failed.

mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/4111/commits/36c32bc047a158bd26f9c00919fbb5fe899dfc59

api/pkg/api/v1/types.go
mrIncompetent: Done with https://github.com/kubermatic/kubermatic/pull/4111/commits/36c32bc047a158bd26f9c00919fbb5fe899dfc59

api/pkg/provider/cloud/vsphere/provider.go
```diff
@@ -118,92 +135,35 @@ func (v *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update
...
-			// ignore virtual networks
-			continue
-		}
-
-		network := Network{Name: netBacking.DeviceName}
-		networks = append(networks, network)
-	}
-
-	return networks, nil
+	return getPossibleVMNetworks(ctx, session)
```
moadqassem, line 150: This method here tries to get all the possible networks and it ignores all the errors which are reported in the process. Now, if all the tries failed the handler will handle the empty slice however it would be nice to have a general log message or potentially an error here, that none of the tries has succeeded.  
```
if info, err := getPossibleVMNetworks(ctx, session); err != nil || len(info) < 1 {
    return nil, errors.New("no possible networks were found")
}
```
Or simply add this check in the `getPossibleVMNetworks` method.

api/pkg/provider/cloud/vsphere/provider.go
```diff
@@ -118,92 +135,35 @@ func (v *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update
...
 }
 
 // GetVMFolders returns a slice of VSphereFolders of the datacenter from the passed cloudspec.
 func (v *Provider) GetVMFolders(cloud kubermaticv1.CloudSpec) ([]Folder, error) {
-	if v.dc.InfraManagementUser != nil {
-		cloud.VSphere.InfraManagementUser.Username = v.dc.InfraManagementUser.Username
-		cloud.VSphere.InfraManagementUser.Password = v.dc.InfraManagementUser.Password
-	}
+	ctx := context.Background()
```
moadqassem: This actually should be `ctx.TODO()` because the context starts in the endpoint, not in this function.

api/pkg/provider/cloud/vsphere/provider.go
```diff
@@ -242,11 +202,11 @@ func (v *Provider) DefaultCloudSpec(cloud *kubermaticv1.CloudSpec) error {
 
 // ValidateCloudSpec validates whether a vsphere client can be constructued for the passed cloudspec.
 func (v *Provider) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
-	client, err := v.getClient(spec)
+	session, err := v.newSession(context.Background(), spec)
```
moadqassem: Same here, the `context.TODO()` because the context of this process starts in the cluster `Spec` which corresponds to the CreateEndpoint call.

api/pkg/provider/cloud/vsphere/provider.go
```diff
@@ -255,18 +215,16 @@ func (v *Provider) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 // We also remove the finalizer if either the folder is not present or we successfully deleted it
 func (v *Provider) CleanUpCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater, _ provider.SecretKeySelectorValueFunc) (*kubermaticv1.Cluster, error) {
 	v.clusterUpdater = update
+	ctx := context.Background()
```
moadqassem: Same goes here, please omit this line here and call `context.TODO()` in the `v.newSession()` method

api/pkg/provider/cloud/vsphere/provider.go
mrIncompetent: We directly return the error here - So in case something fails the caller will now ?

moadqassem: Not really you ignore the errors here: 
https://github.com/kubermatic/kubermatic/pull/4111/files#diff-7f41c6031dfdb241d336cde37220601dR32-R42

mrIncompetent: In case we decide to ignore an error there, we log it - which should be enough.
What would an additional logging help here?

moadqassem: If there was no network available([]NetworkInfo length is Zero) do you consider this an error or not? In your comment you say:

`// Just log the error. We do not want a single weird network to break the entire listing.`


No single network is available this means the entire network listing is broken. So instead of going to the logs and start counting how many broken networks we have and associating them with list NetworksList from the session Finder.

mrIncompetent: > If there was no network available([]NetworkInfo length is Zero) do you consider this an error or not?

An empty list is fine.

> No single network is available this means the entire network listing is broken. 

No it does not. An empty slice means, there are no VM networks available.
I'll update that comment to make that more clear

PR #4129, opened 2019-08-20
update kube-state-metrics to 1.7.2

xrstf:
This might cause trouble, let's see. The Addon-Resizer had only 1.8.5 released between 1.8.4 and 2.3 and the only changes seem to be harmless for us.

```release-note
update kube-state-metrics to 1.7.2
```


kron4eg (approved):
/approve

let's see :)

PR #4143, opened 2019-08-21
Initialize CloudProviders with a SecretKeyGetter

alvaroaleman:

This PR changes the cloud provider interface in a away that they get initialized with a SecretKeyGetter, ensuring they are always able to get credentials. This fixes various AWS endpoints.

The whole thing is done by:

* Creating a factory for a `SecretKeySelectorValueFunc` in the provider pkg
* Asserting the `ClusterProvider` to its implementation `*kubernetesprovider.ClusterProvider` then using its `GetSeedClusterAdminRuntimeClient` to call the factory
* For all endpoints that do not have a cluster already, the `ClusterProvider` can not be uses as its creation needs a DC which we don't have. Hence a `SeedClientGetter` was added to the `Routing`, it is an abstraction on top of the `KubeconfigGetter` with an integrated restMapper cache. It needs the seed thought, which is why `DatacenterFromSeedMap` had to be extended to also return the seed

Fixes #

What is really bad about all of this is that a lot of changes had to be done in order to have a `SecretKeySelector` in endpoints that do not operate on existing clusters _even thought it will never be used_ down the line in the AWS provider.

IMHO we should refactor the VPC/Subnet/Zone endpoints to not require an initialized AWS cloud provider but instead be standalone and just take the credentials as args.

What I will do as follow-ups are:
* DRY out restmapper cache usage and use the `*restmapper.Cache` everywhere
* DRY out `SecretKeySelectorValueFunc` creation and use the factory in the provider package everywhere

**Documentation**:

```release-note
none
```

/assign @mrIncompetent @zreigz @thz

### DIFF
--- a/api/pkg/util/restmapper/cache.go
+++ b/api/pkg/util/restmapper/cache.go
@@ -1,0 +1,47 @@
+package restmapper
+
+import (
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/rest"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// New returns a new Cache
+func New() *Cache {
+	return &Cache{cache: &sync.Map{}}
+}
+
+// RestMapperCache is used to dynamically create controllerruntimeClients whilst caching the RestMapper. It uses properties of the
+// *cfg as cache key
+type Cache struct {
+	cache *sync.Map
+}
+
+// Client returns a brand new controllerruntime.Client, using a cache for the restMapping to avoid doing discovery during startup.
+// It uses properties of the *cfg as cache Key
+func (c *Cache) Client(cfg *rest.Config) (ctrlruntimeclient.Client, error) {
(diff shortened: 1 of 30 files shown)


api/pkg/util/restmapper/cache.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/rest"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// New returns a new Cache
+func New() *Cache {
```
mrIncompetent, line 13: This should also be used in `api/pkg/cluster/client/client.go` otherwise we have 2 different cache mechanisms

alvaroaleman: Yeah I mentioned in the PR description that I want to DRY out the restmapper caching everywhere, but as a follow-up, to unbreak master asap

mrIncompetent: Sorry, missed reading the description :man_facepalming:

PR #4117, opened 2019-08-20
update cert-manager to 0.9.1

xrstf:
**What this PR does / why we need it**:
Latest and Greatest. Because of changes to the API version and CRDs, customers must manually uninstall the chart, remove the CRDs and then re-install the chart. :-( See https://docs.cert-manager.io/en/latest/tasks/upgrading/upgrading-0.8-0.9.html

**Documentation**:
https://github.com/kubermatic/docs/pull/156

**Does this PR introduce a user-facing change?**:
```release-note
update cert-manager to 0.9.1
```

PR #4144, opened 2019-08-21
Revert "update cert-manager to 0.9.1 (#4117)"

xrstf:
This reverts commit 00bdbef02467cb0bcf298ca8ef22a9d27cba4d7d.

The migration turns out to be not just a simple "purge + reinstall", so in order to get our clusters working again, this reverts #4117 until we figured out the actual migration path. We cannot simply pin the version in the cluster's values.yaml because the 0.9.1 update included CRD changes and I don't know if CM 0.8 can handle certs with CRD 0.9.1.

```release-note
NONE
```

### DIFF
--- a/config/cert-manager/Chart.yaml
+++ b/config/cert-manager/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cert-manager
-version: 1.1.0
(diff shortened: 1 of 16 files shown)


kron4eg (approved):
/approve

PR #4146, opened 2019-08-21
3748 Limit concurrent cluster upgrades(in the cluster controller)

### LINKED ISSUE #3748: Limit concurrent cluster upgrades

moadqassem:
To avoid putting too much load on the seed cluster, each controller (Which manages Kubernetes resources), must set a "Condition" on the cluster object if any resource is still updated.
If there are already too much(defined by a flag) clusters which have the condition set to true, the controller will just enqueue the cluster object and skip processing.

Fixes #3748 

At the moment, this PR only implements the limit concurrent cluster updates in the cluster controller, once the approach is accepted it should be also reflected on the controllers which update the cluster crd.

No

```release-note
NONE
```


moadqassem:
@mrIncompetent please have a look on the last two commits. 
https://github.com/kubermatic/kubermatic/pull/4146/commits/5d084e560f4d27c44927c9f24f187b620862bf99

https://github.com/kubermatic/kubermatic/pull/4146/commits/b8bf4b04b6bf24c626241d107f219d281f3d0a1a

api/pkg/crd/kubermatic/v1/cluster_status.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ClusterConditionType string
+
+const (
+	ClusterControllerUpdateInProgressCondition ClusterConditionType = "ClusterControllerUpdateInProgress"
```
alvaroaleman: Can we rename this to `ClusterConditionClusterController...`? The reason is that when we use `ClusterCondition` as prefix, ppl have a chance to actually find it via the completion of their IDE

api/pkg/crd/kubermatic/v1/cluster_status.go
```diff
@@ -0,0 +1,134 @@
...
+
+	ClusterUpdateInProgressReason = "Current Cluster is updating its resources"
+)
+
+type ClusterCondition struct {
```
alvaroaleman, line 20: One of the goals of the whole thing is to use it within tests to figure out of a current version of Kubermatic finished reconiling a given cluster, so I think it would be good to add the Kubermatic version that created the condition here. We have the `KUBERMATICCOMMIT` var in `pkg/resources` that contains this info

api/pkg/controller/util/util.go
```diff
@@ -47,3 +49,68 @@ func EnqueueConst(queueKey string) *handler.EnqueueRequestsFromMapFunc {
...
+		deployments  = &v1.DeploymentList{}
+	)
+
+	if err := client.List(ctx, opts, statefulSets); err != nil {
+		return fmt.Errorf("failed to list statefulSets: %v", err)
+	}
+
+	for _, statefulSet := range statefulSets.Items {
+		if statefulSet.Status.Replicas != statefulSet.Status.ReadyReplicas {
+			cluster.Status.SetClusterUpdateInProgressConditionTrue(fmt.Sprintf("cluster %v resources updates are in progress",
```
alvaroaleman, line 96: Rather than setting a `ClusterUpdateInProgress` I think a `ReconciliationFinishedSuccessfully` makes more sense, because the "I am doing work" condition requires the controller to be up, its absence can also mean the controller is just down, whereas the `I successfully finished work` condition can only be set when that is actually the case.

Also, we must check for `.Status.UpdatedReplicas` as well

api/pkg/crd/kubermatic/v1/cluster_status.go
mrIncompetent: Can we rely on `lastTransitionTime` within the tests?

alvaroaleman: If the new Kubermatic version does nothing, we wont have a transition

api/pkg/crd/kubermatic/v1/cluster_status.go
moadqassem: Yeah SGTM

api/pkg/crd/kubermatic/v1/cluster_status.go
moadqassem: Done

api/pkg/controller/util/util.go
moadqassem: This is not the job of the method, the method job is setting a condition not checking if the work has been done successfully. Or you are talking about the message? Regarding the 

`.Status.UpdateReplicas` yes I totally agree on that one.

moadqassem:
Duplicate https://github.com/kubermatic/kubermatic/pull/4193

PR #4153, opened 2019-08-22
 Make openstack rootDiskSize configurable over API

### LINKED ISSUE #3251: Make the root disk size of Openstack instances configurable
**User Story**

As a user I want to configure the size of the rootdisk of my Openstack instances

**Acceptance criteria**

* It is possible to configure the root disk size of Openstack instances via Kubermatic

bashofmann:
This adds support to the API to make the root disk size of OpenStack nodes configurable. Support was already added to machine-controller in https://github.com/kubermatic/machine-controller/pull/579.

Fixes #3251

```release-note
Disk size of OpenStack nodes is now configurable
```

### DIFF
--- a/api/pkg/api/v1/types.go
+++ b/api/pkg/api/v1/types.go
@@ -911,7 +911,7 @@ type OpenstackNodeSpec struct {
 	UseFloatingIP bool `json:"useFloatingIP,omitempty"`
 	// if set, the rootDisk will be a volume. If not, the rootDisk will be on ephemeral storage and its size will be derived from the flavor
 	// required: false
-	RootDiskSizeGB int64 `json:"diskSize"`
+	RootDiskSizeGB *int `json:"diskSize"`
 }
 
 // AWSNodeSpec aws specific node settings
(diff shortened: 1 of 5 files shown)


api/pkg/api/v1/types.go
```diff
@@ -909,6 +909,9 @@ type OpenstackNodeSpec struct {
 	// Defines whether floating ip should be used
 	// required: false
 	UseFloatingIP bool `json:"useFloatingIP,omitempty"`
+	// if set, the rootDisk will be a volume. If not, the rootDisk will be on ephemeral storage and its size will be derived from the flavor
+	// required: false
+	RootDiskSizeGB int64 `json:"diskSize"`
```
alvaroaleman: Wouldn't it be nicer to make this a `*int` so we can directly put it onto the machine without further checks?

bashofmann: I made it the same type as the AWS disk size, but I guess from the json side it does not matter and it makes things easier in the code. I'll change it.

PR #4188, opened 2019-08-27
conformance tests: dump logs for all control plane pods

kdomanski:
**Why we need it**:
I'm debugging a failed test and according to the log, the kubermatic api server returns an empty error response. I'd like to be able to easily see what it has to say in its logs.

In addition, the existing code that dumps the controller-manager logs assumes that there is only one replica. I generalized it to always iterate over all pods, to avoid misleading the user.

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci/ci-run-minimal-conformance-tests.sh
@@ -68,7 +68,18 @@ function cleanup {
     fi
 
     # Controller manager logs
-    kubectl logs -n $NAMESPACE  $(kubectl get pod -n $NAMESPACE -l role=controller-manager |tail -n 1|awk '{print $1}')
+    echodate "Dumping controller-manager logs"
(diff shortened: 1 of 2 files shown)


api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -68,7 +68,18 @@ function cleanup {
     fi
 
     # Controller manager logs
-    kubectl logs -n $NAMESPACE  $(kubectl get pod -n $NAMESPACE -l role=controller-manager |tail -n 1|awk '{print $1}')
+    echodate "Dumping controller-manager logs"
+    for pod in $(kubectl get pods -n $NAMESPACE -l role=controller-manager -o go-template='{{ range .items }}{{.metadata.name}}{{ "\n" }}{{end}}'); do
```
alvaroaleman: What about just dumping all logs of all pods in the ns?

kdomanski: Sure, even better.

PR #4193, opened 2019-08-28
Limit concurrent cluster upgrades in the cluster controller

moadqassem:
To avoid putting too much load on the seed cluster, each controller (Which manages Kubernetes resources), must set a "Condition" on the cluster object if any resource is still updated.
If there are already too much(defined by a flag) clusters which have the condition set to true, the controller will just enqueue the cluster object and skip processing.

Towards fixing #3748

At the moment, this PR only implements the limit concurrent cluster updates in the cluster controller, once the approach is accepted it should be also reflected on the controllers which update the cluster crd.

No

```release-note
NONE
```


api/pkg/provider/kubernetes/cluster.go
```diff
@@ -112,6 +113,7 @@ func (p *ClusterProvider) New(project *kubermaticv1.Project, userInfo *provider.
 			UserEmail:              userInfo.Email,
 			NamespaceName:          NamespaceName(name),
 			CloudMigrationRevision: cloud.CurrentMigrationRevision,
+			KubermaticVersion:      resources.KUBERMATICCOMMIT,
```
alvaroaleman, line 116: Why is this put in the cluster status?

api/pkg/resources/reconciling/ensure.go
```diff
@@ -92,6 +93,13 @@ func EnsureNamedObject(ctx context.Context, namespacedName types.NamespacedName,
 	}
 
 	if !requiresRecreate {
+		switch v := obj.(type) {
```
alvaroaleman, line 98: What is the purpose of this?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -173,6 +176,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	log := r.log.With("request", request)
 	log.Debug("Processing")
 
+	if process, err := controllerutil.LimitConcurrentUpdates(ctx, r, r.concurrentClusterUpdates); !process && err != nil {
```
alvaroaleman: This is wrong, if `process` is false, we must return regardless of `err` being nil or not

api/pkg/controller/util/util.go
```diff
@@ -47,3 +49,70 @@ func EnqueueConst(queueKey string) *handler.EnqueueRequestsFromMapFunc {
...
+	}
+
+	if err := client.List(ctx, opts, deployments); err != nil {
+		return fmt.Errorf("failed to list deployments: %v", err)
+	}
+
+	for _, deployment := range deployments.Items {
+		if deployment.Status.Replicas != deployment.Status.UpdatedReplicas ||
+			deployment.Status.Replicas != deployment.Status.ReadyReplicas {
+			cluster.Status.SetClusterUpdateInProgressConditionTrue(fmt.Sprintf("cluster %v resources updates are in progress",
```
alvaroaleman: As mentioned in the previous PR, please do not set a condition "update in progress", set a condition "update finished" once that is the case. If something is broken, nothing will set the condition.

Also, both the name and the godoc of this func is plain wrong, this doesn't check anything, this sets something. It also doesn't return a bool.

Please also make sure we do this with one update call only.

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: Also, can we rename the func to `ConcurrencyLimitReached`? The function doesnt do any limiting, it just returns the info if you are limited

api/pkg/controller/util/util.go
```diff
@@ -47,3 +49,70 @@ func EnqueueConst(queueKey string) *handler.EnqueueRequestsFromMapFunc {
...
+			return client.Update(ctx, cluster)
+		}
+	}
+
+	if err := client.List(ctx, opts, deployments); err != nil {
+		return fmt.Errorf("failed to list deployments: %v", err)
+	}
+
+	for _, deployment := range deployments.Items {
+		if deployment.Status.Replicas != deployment.Status.UpdatedReplicas ||
```
alvaroaleman: This is missing the check if `.Status.CurrentReplicas` matches Replicas. Also, is `.Status.Replicas` always equal to `.Spec.Replicas?

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -123,6 +124,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
 	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
 	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
+	flag.IntVar(&c.concurrentClusterUpdate, "concurrent-cluster-update", 10, "The default number of resources updates per cluster")
```
alvaroaleman: I am not completely happy with the name, it seems to imply that this is about updating clusters, when its about reconciling them which can happen for a number of reasons. What about `max-parallel-reconcile`?

api/pkg/provider/kubernetes/cluster.go
moadqassem: It was a quick fix for the import cycle also the methods that could use this field are actually attached on the Cluster Status struct.

api/pkg/controller/cluster/cluster_controller.go
moadqassem: I actually changed how the code works and forgot to refactor the condition respectively. Regarding the name we can change it to `ConcurrencyLimitReached`.

api/pkg/resources/reconciling/ensure.go
moadqassem: This is type assertion! obj is a generic variable and the only way to access its properties and methods is by using a type assertion.

alvaroaleman: Sorry, badly phrased, I know what a type assertion is ;) The question was why you reset the Status?

moadqassem: The reason why we keep resetting the status is to avoid working on any outdated object. Imagine if something goes wrong with the Kubernetes Deployment Controller(maybe it's down). we never have a guarantee that those objects are ready or not because the status won't be updated.

api/cmd/kubermatic-controller-manager/options.go
moadqassem: We can use this name instead.

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -173,6 +176,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	log := r.log.With("request", request)
 	log.Debug("Processing")
 
+	if process, err := controllerutil.ConcurrencyLimitReached(ctx, r, r.concurrentClusterUpdates); process {
```
alvaroaleman: err the logic here is the wrong way round or not? If `ConcurrencyLimitReached` returns `true` you can continue, currently you continue when it returns `false`

Also, we should return if err != nil as well, regardless of the value of the boolean

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -207,9 +216,15 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		log.Errorw("Reconciling failed", zap.Error(err))
 		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError", "%v", err)
 	}
+
 	if result == nil {
 		result = &reconcile.Result{}
 	}
+
+	if err := controllerutil.SetClusterUpdatedSuccessfullyCondition(ctx, cluster, r); err != nil {
```
alvaroaleman: `s/err :=/err =/` so the error gets returned

api/pkg/controller/util/util.go
```diff
@@ -70,3 +71,78 @@ func SupportsFailureDomainZoneAntiAffinity(ctx context.Context, client ctrlrunti
...
+// is more than the limit, then it will return true, which means the concurrency limit has reached.
+func ConcurrencyLimitReached(ctx context.Context, client ctrlruntimeclient.Client, limit int) (bool, error) {
+	clusters := &kubermaticv1.ClusterList{}
+	if err := client.List(ctx, nil, clusters); err != nil {
+		return true, fmt.Errorf("failed to list clusters: %v", err)
+	}
+
+	finishedUpdatingClustersCount := 0
+	for _, cluster := range clusters.Items {
+		for _, condition := range cluster.Status.Conditions {
```
alvaroaleman: Can we move everything inside the inner for loop in a dedicated func for readability, something like `hasConditionValue(c *kubermaticv1.Cluster, condition kubermaticv1.ClusterCondition, value bool`)?

Also, in the future `ConcurrencyLimitReached` must be able to cope with different condition names as we have multiple controllers that needs this. That can be done in a follow-up thought, if you prefer.

api/pkg/controller/util/util.go
```diff
@@ -70,3 +71,78 @@ func SupportsFailureDomainZoneAntiAffinity(ctx context.Context, client ctrlrunti
...
+
+	if err := client.List(ctx, opts, statefulSets); err != nil {
+		return fmt.Errorf("failed to list statefulSets: %v", err)
+	}
+
+	for _, statefulSet := range statefulSets.Items {
+		if *statefulSet.Spec.Replicas == statefulSet.Status.UpdatedReplicas &&
+			*statefulSet.Spec.Replicas == statefulSet.Status.CurrentReplicas &&
+			*statefulSet.Spec.Replicas == statefulSet.Status.ReadyReplicas {
+			updatedStatefulSetCount++
```
alvaroaleman: I think this can be simplified by just using a boolean:

```
var  hasUnfinishedUpgrades bool
if *statefulSet.Spec.Replicas != statefulSet.Status.UpdatedReplicas || [...remaining conditions here ] {
hasUnfinishedUpgrades bool
}
```

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -94,6 +95,134 @@ type ClusterSpec struct {
...
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+
+func newClusterCondition(condType ClusterConditionType, status corev1.ConditionStatus, reason, message string) *ClusterCondition {
+	now := metav1.Time{
+		Time: time.Now(),
+	}
+
+	return &ClusterCondition{
```
alvaroaleman, line 132: This is missing the Kubermatic version

api/Gopkg.lock
```diff
@@ -1972,9 +1972,12 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:03a96603922fc1f6895ae083e1e16d943b55ef0656b56965351bd87e7d90485f"
+  digest = "1:0728020defbd5d26e88ec5f6a9e844b6f2982998bad1356b649ccc9469e0aa89"
   name = "k8s.io/kube-openapi"
-  packages = ["pkg/util/proto"]
+  packages = [
+    "pkg/common",
```
alvaroaleman: Why the change in vendor, I cant find the import of `k8s.io/kube-openapi/pkg/common`?

api/pkg/controller/cluster/cluster_controller.go
moadqassem: The method name is LimitConcurrentUpdates, and as the name suggests it should limit any concurrent updates if it's true, please have a look on the implementation of the method.

Regarding the error, it returned anyway whether it's nil or not and it will be handled by the caller . Please have a look on line number 182(https://github.com/kubermatic/kubermatic/pull/4193/files/18282f2e33a6c6b629824881f5c6c6a869766c81#diff-8def4b8292bbf4e35c590acf915a4dcbR182)  and on the processNextItem implementation https://github.com/kubermatic/kubermatic/blob/master/api/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L215

api/pkg/controller/cluster/cluster_controller.go
moadqassem: Done

api/pkg/controller/util/util.go
moadqassem: Not really! we are checking multiple values here not one, in other words if the first two statefulSet were Not ready and the last one was then the value will be true and that's wrong because not all resources are ready that's.

api/Gopkg.lock
moadqassem: `dep check` was failing so I ran `dep ensure` thus I think this was responsible for that change. I will take care of it

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: It's already in the status, why do we need to add Kubermatic version in both the status and the condition if conditions are fileds in the status?

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: What? This Pull request has no func named `LimitConcurrentUpdates`. Do you have local changes you didn't push?

moadqassem: Oh sorry that was the old name.  ConcurencyLimitReached is what I meant but because I was fixing an old branch(where the name is LimitConcurrentUpdates the name was kept in my head. Anyway the(ConcurencyLimitReached) also has the same name reflecting. If the concurrency limit reached then it should return true. I will change the name of bool var from `process` to `limitReached` to make it clearer.

api/pkg/controller/util/util.go
moadqassem: Sure we can move it in a private func in the same file.

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: okay makes sense, but can you rename the bool var from `process` to `limitReached` or sth? While the code is correct the var name makes it hard to understand.

Re the error, that is not enough. If the error is not nil, we must return. Setting a different return var to specific value because you happen to know that the caller only considers the error if the other return var has this specific value is hacky at best and extremely likely to break in the future.

api/pkg/controller/util/util.go
alvaroaleman: Thats why the boolean is `hasUnfinishedUpgrades` not `hasFinishedUpgrades`. Its false by default and once you encounter the first unfinished upgrade, you can set it to true, nothing will set it back to false.

api/pkg/controller/cluster/cluster_controller.go
moadqassem: It's returned! if the error is not nil then the limit reached is true and then it will return.

api/pkg/crd/kubermatic/v1/cluster.go
alvaroaleman: Consider the following scenario:

* Controller a does something, finishes successfully, sets the Status and its condition
* Controller b is still in the process of reconciling and didn't yet set anything, however there is still a Condition from an older kubermatic version

=> Observer will think all controllers finished even thought only controller a finished

api/pkg/controller/util/util.go
moadqassem: When I am setting it to false or true I am not exiting the loop so it will be assigned again with next run. Now this can be fixed however with adding a `break`.

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: Please always check errors. Ppl that adjust `ConcurrencyLimitReached` in the future are unlikely to know that it must return true when the error is non-nil.

api/pkg/controller/util/util.go
alvaroaleman: the bool is `false` be default, so there is no reason to ever set it to `false`

moadqassem: This thing here `hasUnfinishedUpgrades bool`inside the if statement is kinda misled me, you meant `hasUnfinishedUpgrades = true` right ?

alvaroaleman: Yes, sorry for the confusing comment, my bad

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: Done

api/Gopkg.lock
moadqassem: Actually I got those changes from the master

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -123,6 +124,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
 	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
 	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
+	flag.IntVar(&c.concurrentClusterUpdate, "max-parallel-reconcile", 10, "The default number of resources updates per cluster")
```
alvaroaleman, line 127: Can you also please put this in the helm chart at REPOROOT/config/kubermatic/values.yaml respective REPOROOT/.config/kubermatic/templates/kubermatic-controller-manager-dep.yaml?

api/pkg/provider/kubernetes/cluster.go
alvaroaleman: This isn't enough, this will only set the KubermaticVersion at creation time but never update it when Kubermatic gets updated

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -94,6 +95,136 @@ type ClusterSpec struct {
...
+func (cs *ClusterStatus) SetClusterFinishedUpdatingSuccessfullyCondition(message string) {
+	condition := newClusterCondition(ClusterConditionControllerFinishedUpdatingSuccessfully, corev1.ConditionTrue,
+		ClusterUpdateInProgressReason, message, cs.KubermaticVersion)
+	cs.setClusterCondition(*condition)
+}
+
+func (cs *ClusterStatus) setClusterCondition(c ClusterCondition) {
+	pos, clusterCondition := cs.getClusterCondition(c.Type)
+	if clusterCondition != nil &&
+		clusterCondition.Status == c.Status && clusterCondition.Reason == c.Reason && clusterCondition.Message == c.Message {
```
alvaroaleman, line 193: This also needs to check the Kubermatic version

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -94,6 +95,136 @@ type ClusterSpec struct {
...
+	CloudMigrationRevision int `json:"cloudMigrationRevision"`
+}
+
+func (cs *ClusterStatus) SetClusterFinishedUpdatingSuccessfullyCondition(message string) {
+	condition := newClusterCondition(ClusterConditionControllerFinishedUpdatingSuccessfully, corev1.ConditionTrue,
+		ClusterUpdateInProgressReason, message, cs.KubermaticVersion)
+	cs.setClusterCondition(*condition)
+}
+
+func (cs *ClusterStatus) setClusterCondition(c ClusterCondition) {
```
alvaroaleman, line 189: Can we have a unittest that verirfies this correctly adds or updates depending on prior existance?

api/cmd/kubermatic-controller-manager/options.go
moadqassem: Done

api/pkg/crd/kubermatic/v1/cluster.go
moadqassem: Done

api/pkg/provider/kubernetes/cluster.go
moadqassem: Done

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -207,9 +216,15 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		log.Errorw("Reconciling failed", zap.Error(err))
 		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError", "%v", err)
 	}
+
 	if result == nil {
 		result = &reconcile.Result{}
 	}
+
+	if err = controllerutil.SetClusterUpdatedSuccessfullyCondition(ctx, cluster, r); err != nil {
```
alvaroaleman: Wait, this is wrong, it gets executed even if the error from `r.reconcile` is non-nil.

api/pkg/crd/kubermatic/v1/cluster.go
alvaroaleman: Where? Please dont comment on something as done before you've pushed the changes.

moadqassem: Yeah sorry about that, I usually do this right before pushing, but since a test has failed I am trying to fix it before.

api/pkg/controller/cluster/cluster_controller.go
moadqassem: That's true, the logic should be evaluated by the AC, in other words, from a logic point of view, if an error occurred during the update do we ignore this condition update? if yes, we should then clear the condition when an error occurred, and what if the error did influence the number of replicas then the condition might fail and the condition will be cleared anyway. but if it didn't(if that's possible) do we still clear the condition?

alvaroaleman: There is one case in which we set the condition to true: No error and everything is up-to-date.

For everything else, we unset it/set it to false.

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -206,10 +215,16 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if err != nil {
 		log.Errorw("Reconciling failed", zap.Error(err))
 		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError", "%v", err)
+	} else {
+		if err = controllerutil.SetClusterUpdatedSuccessfullyCondition(ctx, cluster, r); err != nil {
```
alvaroaleman: Actually looking at it again, `SetClusterUpdatedSuccessfullyCondition` both sets and resets the condition, depending on the status. Maybe it would be better to pass in a boolean `reconciliationSuccessful` and call it unconditionally?

api/pkg/controller/util/util.go
```diff
@@ -7,6 +7,7 @@ import (
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
 
+	"k8s.io/api/apps/v1"
```
alvaroaleman: Nit: Can you alias this import as `appv1`? `v1` as import name is very confusing, because there are a lot of `v1` apis in Kubernetes. In most contexts I've seen, ppl always alias the kube api imports as `apinameapiversion`

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,245 @@
...
+	var (
+		ctx    = context.Background()
+		client = &mockClient{
+			errorOccurred: true,
+		}
+	)
+
+	reached, err := ConcurrencyLimitReached(ctx, client, testSetup.limit)
+	if err == nil || !reached {
+		t.Fatal("error should not be nil and reached should be true")
```
alvaroaleman: If an error is non-nil, the value of all other returned vars is irrelevant

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,245 @@
...
+					UpdatedReplicas:   1,
+					ReadyReplicas:     2,
+					AvailableReplicas: 2,
+				},
+			},
+		},
+	}
+)
+
+func TestConcurrencyLimitReached(t *testing.T) {
```
alvaroaleman: I find this very hard to read tbh. What about doing some table driven tests that clearly show the input and the expected outcome, e.G.:

```
testCases := []struct{
 name string
 objects []runtime.Object
 expectedConcurrencyLimitReached bool
}{
  name: "Concurrency limit reached",
  objects: []runtime.Object{{a-cluster}, {some-object-that-updates}},
  expectedConcurrencyLimitReached: true,
 }.
{
  name: "Concurrency limit not reached",
  objects: []runtime.Object{{a-cluster}, {some-object-that-is-up-to-date}},
  expectedConcurrencyLimitReached: false,
 },
}
// Check ConcurrencyLimitReached with limit == 1
```

Also I find it a bit overkill to test that the func returns an error when the client returns an error, we have a linter that verifies error handling.

What may be helpful for your testing is to use the fakeclient in `sigs.k8s.io/controller-runtime/pkg/client/fake`

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	err := SetClusterUpdatedSuccessfullyCondition(ctx, &testCluster, client)
+	if err != nil {
+		t.Fatal("error has occurred: ", err)
+	}
+
+	if ok := hasConditionValue(testCluster,
+		kubermaticv1.ClusterConditionControllerFinishedUpdatingSuccessfully,
+		corev1.ConditionTrue); ok {
+		t.Fatal(fmt.Sprintf("resources are still updating, condition %v shouldn't be there",
```
alvaroaleman: There is `t.Fatalf` for this

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !reached {
+		t.Fatal("max concurrent updates should have been reached")
+	}
+}
+
+func TestSetClusterUpdatedSuccessfullyCondition(t *testing.T) {
```
alvaroaleman, line 62: Basically the same as for the other test, please write table-driven tests that clearly show the input and expected outcome. Here it would be something like:

```
testcases := []struct{
name string
existingCondition corev1.ConditionType
conditionToSet corev1.ConditionType
}{
{
  name: "Condition successful set",
  existingCondition: "",
  conditionToSet: corev1.ConditionTrue",
},
}

 t.Run(tc.name, func(t *testing.T {
   if existingCondition != "" {
       cluster.Status.Conditions := []ClusterCondition{existingCondition}
   }
   setClusterCondition{tc.ConditionToSet}
   if len(cluster.Status.Conditions ) != 1 {
    t.Errorf(...)
   }
   if cluster.Spec.Conditions[0].Value != tc.ConditionToSet {
      t.Errorf(...)
   }
  }
}
```

In the end there should be tests for:
* Condition successfully set without pre-existing conditions
* Condition successfully updated with no change
* Condition successfully updated with a change
* Condition successfully cleared

It is also very crucial that you verify there is always exactly one condition to make sure we don't have duplicates.

api/pkg/controller/cluster/cluster_controller.go
moadqassem: Honestly, I would leave it as it is. Why make an extra call or extra check for that.

api/pkg/controller/util/util_test.go
moadqassem: So? This is a test case and I am controlling the test flow with flags.

alvaroaleman: If someone changes the func and makes it return `false, errors.New("Some error")` this testcase will fail even thought it shouldn't. When something returns a non-nil errors, the value of all other returns are 100% irrelevant, which is why we should not verify them.

api/pkg/controller/cluster/cluster_controller.go
alvaroaleman: Because the current behaviour of this is:
* If ! reconcile error && everything up to date: Set successful
* if ! reconcile error && ! everything up to date: Unset condition
* if reconcile error: Leave the condition at whatever state it was althought we can't possibly know if everything is in a good state or not

api/pkg/controller/util/util_test.go
moadqassem: Oh now I understand what you mean, you are talking about one var which is reached. You said all vars without explicitly mentioning which so I thought you meant all the vars in the function. If the case is just about the `reached` yeah this can be ignored

api/pkg/controller/cluster/cluster_controller.go
moadqassem: We can add this call where it was before and as you mentioned we have a param that we set to false if `err != nil` and if it was false then we clear the condition right away and skip processing the checks in the  `SetClusterUpdatedSuccessfullyCondition`

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -201,15 +210,23 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, nil
 	}
 
+	successfullyRecunciled := true
 	// Add a wrapping here so we can emit an event on error
 	result, err := r.reconcile(ctx, log, cluster)
 	if err != nil {
+		successfullyRecunciled = false
```
alvaroaleman: There is a typo in the variable name and its probably more expressive to do a `succesfullyReconciled := err == nil`?

api/pkg/controller/util/util.go
alvaroaleman: This one is still open

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,354 @@
...
+		expects struct {
+			limitReached bool
+			err          error
+		}
+	}{
+		{
+			name: "concurrency limit has not reached",
+			wants: struct {
+				ctx    context.Context
+				client *mockClient
```
alvaroaleman: I don't really understand why this `mockClient` exists and why we don't use the fake client in `sigs.k8s.io/controller-runtime/pkg/client/fake`. You can just pass an `inProgressCluster`/`readyCluster` into the `...runtime.Object` you initialize the fake client with.

The upstream fake client is very powerful, IMHO we should just use it instead of rebuilding it.

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,354 @@
...
+			name: "statefulSet resources are not yet updated",
+			wants: struct {
+				ctx                    context.Context
+				client                 *mockClient
+				cluster                kubermaticv1.Cluster
+				successfullyReconciled bool
+			}{
+				ctx: context.Background(),
+				client: &mockClient{
+					inProgressStatefulSet: true,
```
alvaroaleman: similiarily like with the clusters, istead of building your own client that does stuff based on boolean, why not use the upstream fake client and build a simple `func getInProgressStatefulSet()`, then pass that in at initialization type?

Thats much less magic and much easier to reason about.

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,354 @@
...
+				limitReached: true,
+				err:          nil,
+			},
+		},
+	}
+
+	setClusterUpdatedSuccessfullyConditionTestCases = []struct {
+		name  string
+		wants struct {
+			ctx                    context.Context
```
alvaroaleman: I don't think the `context` should be part of the testcase, just generate one at test execution time

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -201,15 +210,23 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		successfullyReconciled = false
 		log.Errorw("Reconciling failed", zap.Error(err))
 		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError", "%v", err)
 	}
+
 	if result == nil {
 		result = &reconcile.Result{}
 	}
+
+	if err = controllerutil.SetClusterUpdatedSuccessfullyCondition(ctx, cluster, r, successfullyReconciled); err != nil {
```
alvaroaleman: This should be `err := [...`, because we must return the `err` from the reconcile

alvaroaleman: Maybe even rename it to `reconcileErr` to make that more obvious?

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,334 @@
...
+				err          error
+			}{
+
+				limitReached: true,
+				err:          nil,
+			},
+		},
+	}
+
+	for _, testCase := range concurrencyLimitReachedTestCases {
```
alvaroaleman, line 47: The test output is easier to read if you call `t.Run` in the loop

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,334 @@
...
+		reached, err := ConcurrencyLimitReached(context.Background(), testCase.wants.client, testCase.wants.limit)
+
+		if err != testCase.expects.err {
+			t.Fatalf("failed to run test: %v with error: %v", testCase.name, err)
```
alvaroaleman: There is no testcase that expects an `err` so might as well remove that here and in the `testCase` struct :upside_down_face:

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,334 @@
...
+			t.Fatalf("failed to run test: %v, expects: %v, got: %v", testCase.name, testCase.expects.limitReached, reached)
+		}
+
+	}
+}
+
+func TestSetClusterUpdatedSuccessfullyCondition(t *testing.T) {
+	setClusterUpdatedSuccessfullyConditionTestCases := []struct {
+		name  string
+		wants struct {
```
alvaroaleman: so it _might_ be easier and less tedious if you didn't use anonymous substructs, just naming the params e.G. `expectHasConditionValue` and `expectErr` should be enough. Also I think there is no testcase where an error is expected or did I miss it? If not, that arg isn't needed here

api/pkg/controller/util/util_test.go
```diff
@@ -0,0 +1,334 @@
...
+				hasConditionValue bool
+				err               error
+			}{
+				hasConditionValue: false,
+				err:               nil,
+			},
+		},
+	}
+
+	for _, testCase := range setClusterUpdatedSuccessfullyConditionTestCases {
```
alvaroaleman, line 180: same here, much easier to read when you use `t.Run`

api/pkg/resources/reconciling/ensure.go
alvaroaleman: Can we get the comment here explainging that we reset it to make sure the status is up-to-date? :)

api/pkg/controller/util/util_test.go
moadqassem: The error actually detects if any of the parameters in the `NewFackeClinet` are corrects, this was the only way how I figured out that I am missing some resources that I need to initialize so I need to check the error to give more clarity about why the test has failed when some params are not there. I can remove the err from the struct but the check must be there

alvaroaleman: I am not questioning the error checking, I am questioning having `testCase.expects.err` because its `false` for each and every test

api/pkg/controller/util/util_test.go
moadqassem: I can use private types here instead of the anonymous structs. Regarding the error please refer to: 
https://github.com/kubermatic/kubermatic/pull/4193#discussion_r320791980

edit: 
What do you mean about `e.G. expectHasConditionValue and expectErr should be enough` ? We actually need those sturtcs for the fake client or it wouldn't work.

alvaroaleman: The anonymous struct is fine, but the anonymous substruct makes the whole thing very tedious.

api/pkg/resources/reconciling/ensure.go
moadqassem: Strange!! I actully remember that I left a comment here reagrding that change :/

api/pkg/controller/util/util_test.go
moadqassem: For the expects yes I can change that but for the `wants` I would like to keep that because it's more compact.

alvaroaleman: How is redefining an anymous struct for the wants part in every single testcase more compact than having one anymous struct for the whole testcase only that never needs to be redefined?

moadqassem: Because I am isolating each test case with its own testing data as simple as that!

alvaroaleman: But you already do that by putting each testcase into its own instance of the testcase struct :)
Those testcase become very long and harder to read if all of them must contain a struct definition. Check out e.g.https://github.com/kubermatic/kubermatic/blob/master/api/pkg/validation/seed/seed_test.go#L26 for a sample of a set of nice to read table-driven tests

moadqassem: What is happening here, I would like to have the input of the method that I am testing which is the `SetClusterUpdatedSuccessfullyCondition` wrapped, that's why I have this struct. If the value of the input, for example, was of the same type then I could have used a slice but they are of different types. In go table-driven model that's normal please refer to this wiki which shows how the testing data is classified into In and Out(wants and expects):

https://github.com/golang/go/wiki/TableDrivenTests

If it's hard to read I can simply put those vars that I using in the structs as local variable however I still would like to have them wrapped in this struct!

moadqassem: In order to increase readability, I have used vars instead of the definition of the inline types.

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -201,15 +210,23 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+		successfullyReconciled = false
 		log.Errorw("Reconciling failed", zap.Error(err))
 		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError", "%v", err)
 	}
+
 	if result == nil {
 		result = &reconcile.Result{}
 	}
+
+	if reconcileErr := controllerutil.SetClusterUpdatedSuccessfullyCondition(ctx, cluster, r, successfullyReconciled); err != nil {
```
alvaroaleman: What I meant with the comment about `reconcileErr` was to put the result of `r.reconile` in it

alvaroaleman (commented):
one last remark, other than that lgtm

PR #4202, opened 2019-08-29
[release/v2.11] Cherrpick PV cleanup: Do not attempt to delete pods when there are no PVs/PVCs #4079

alvaroaleman:

Makes the cleanup not attempt to delete any pods if there are no PVs and PVCs, also makes all of the Pod/PV/PVC deletion tolerate a 404 and not treat it as an error.

Cherrypicking required a bit more shuffling code around as the code in master has diverged quite a bit.

Fixes #

**Documentation**:

```release-note
Fixed a bug that could result in the clusterdeletion sometimes getting stuck
```

/assign @kdomanski


alvaroaleman:
The lint exception isn't needed on master anymore as the code got refactored there, but I'd prefer not to backport the refactorings and instead just stick with the exception

PR #4213, opened 2019-08-30
Nodeport proxy: Dont rotate nodeport when target services port changed

### LINKED ISSUE #4210: nodeport-proxy lb-updater updates twice
The lb-updater of the nodeport-proxy takes care of adding ports to the `nodeport-lb` service so that the cloud-provider will add the respective load balancing targets for user-clusters services. The lb-updater uses a specifically formatted port name `namespace-servicename-serviceport-nodeport` to implement detection of user-cluster service changes which need an update in the `nodeport-lb` service.
Each user-cluster `apiserver-external` service starts with a serviceport `443` and updates that to match the nodeport a bit later. The lb-updater acts on the initial `cluster-xxxxx-apiserver-external-443-nodeport` port name by updating the ports and thus requesting a nodeport for the `nodeport-lb` service. When the `apiserver-external` service gets updated to match its serviceport to its nodeport (`cluster-xxxxx-apiserver-external-nodeport-nodeport`) the lb-updater acts again and sends an update which removes the assigned `nodeport-lb`'s nodeport which will cause a new nodeport getting assigned.
In K8s 1.15.3 the aws cloud-provider is quick enough to already act on the initial nodeport and create an aws targetgroup. The following nodeport change triggers the aws cloud-provider to update the targetgroup which then fails with:
```
I0829 11:42:24.595463       1 event.go:258] Event(v1.ObjectReference{Kind:"Service", Namespace:"nodeport-proxy", Name:"nodeport-lb", UID:"633b0031-a9ef-4085-8dee-fcd1c97941db", APIVersion:"v1", ResourceVersion:"39728", FieldPath:""}): type: 'Warning' reason: 'SyncLoadBalancerFailed' Error syncing load balancer: failed to ensure load balancer: Error creating load balancer target group: "DuplicateTargetGroupName: A target group with the same name 'k8s-nodeport-nodeport-df5051f821' exists, but with different settings\n\tstatus code: 400, request id: 11b98275-ca52-11e9-b4c7-c9beaed8d7c4"
```
So this seems to be an upstream issue (unable to update targetgroup), but it is triggered by an unnecessary nodeport change on our lb-updater side.

alvaroaleman:

Fixes the nodeport proxy to not rotate nodeports

Fixes #4210

**Documentation**:

```release-note
none
```

/assign @thz 
/assign @xrstf 

Please review thoroughly.

### DIFF
--- a/api/cmd/nodeport-proxy/lb-updater/main.go
+++ b/api/cmd/nodeport-proxy/lb-updater/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/go-test/deep"
 	"github.com/golang/glog"
 
 	controllerutil "github.com/kubermatic/kubermatic/api/pkg/controller/util"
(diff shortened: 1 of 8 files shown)


api/cmd/nodeport-proxy/lb-updater/main.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/go-test/deep"
```
xrstf: If we care about sorting imports, this should be moved down.

PR #3946, opened 2019-08-05
Use secrets for credentials for Digitalocean

nikhita:
Follow up of https://github.com/kubermatic/kubermatic/pull/3836
/hold

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #4215, opened 2019-08-30
fix getting sizes for DO

### LINKED ISSUE #4214: Fix getting sizes for DO during adding new ND
**User Story**
Get DO token from secret. Right now it takes token from cloud spec

zreigz:
**What this PR does / why we need it**: fix getting sizes for DO

Fixes #4214

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/kubernetes/credentials.go
+++ b/api/pkg/provider/kubernetes/credentials.go
@@ -12,6 +12,14 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func GetCredentialSecret(ctx context.Context, seedClient ctrlruntimeclient.Client, secretName string) (*corev1.Secret, error) {
+	secret := &corev1.Secret{}
+	if err := seedClient.Get(ctx, ctrlruntimeclient.ObjectKey{Name: secretName, Namespace: resources.KubermaticNamespace}, secret); err != nil {
+		return nil, err
+	}
+	return secret, nil
+}
+
 // CreateCredentialSecretForCluster creates a new secret for a credential
 func CreateCredentialSecretForCluster(ctx context.Context, seedClient ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster, projectID string) error {
 	if cluster.Spec.Cloud.AWS != nil {
--- a/api/pkg/provider/kubernetes/credentials.go
+++ b/api/pkg/provider/kubernetes/credentials.go
@@ -12,14 +12,6 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


nikhita (approved):
/lgtm
/approve

alvaroaleman:
This breaks the endpoint for all clusters that do not have a secret yet, which is any cluster that was created before Nikhitas change, as we don't yet have a migration.

This must use the same Code the provider itself uses to get the token and not  duplicate anything. This must also be able to cope with both the secretRef and the token being directly on the cluster CR:

alvaroaleman:
Can we please just revert https://github.com/kubermatic/kubermatic/pull/3946 instead?

zreigz:
what about something like this:
```
		doToken := cluster.Spec.Cloud.Digitalocean.Token
		if len(doToken) == 0 {
			secret, err := kubernetesprovider.GetCredentialSecret(ctx, privilegedClusterProvider.GetSeedClusterAdminRuntimeClient(), cluster.Spec.Cloud.Digitalocean.CredentialsReference.Name)
			if err != nil {
			}

			doTokenBytes, ok := secret.Data[resources.DigitaloceanToken]
			if !ok {
				return nil, errors.NewNotFound("token for ", req.ClusterID)
			}
			doToken = string(doTokenBytes)
		}

		return digitaloceanSize(ctx, doToken)
```

zreigz:
it will make the endpoint backward compatible

alvaroaleman:
There is `resources.GetDigitalOceanCredentials` for this. Another problem I found is that the `Validate` of the DO provider just does a
```
static := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: spec.Digitalocean.Token})
```

I am not sure why this works (probably because its called before we unset the Token on the cluster?) but it needs to be fixed.

zreigz:
> There is `resources.GetDigitalOceanCredentials` for this. Another problem I found is that the `Validate` of the DO provider just does a
> 
> ```
> static := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: spec.Digitalocean.Token})
> ```
> 
> I am not sure why this works (probably because its called before we unset the Token on the cluster?) but it needs to be fixed.

This part is called before secret creation and cluster creation when cloud spec still contains credentials fields, so it's perfectly fine

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -12,6 +12,14 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func GetCredentialSecret(ctx context.Context, seedClient ctrlruntimeclient.Client, secretName string) (*corev1.Secret, error) {
```
alvaroaleman: Is this a leftover? I can't find it referenced anywhere

PR #4216, opened 2019-08-30
OpenVPN client: Explicitly enable ip forwarding

alvaroaleman:

Makes sure the openvpn client pod has ip_forwarding enabled, as otherwise nating of the node addresses wont work.

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/addons/openvpn/openvpn-client-dep.yaml
+++ b/addons/openvpn/openvpn-client-dep.yaml
@@ -73,6 +73,7 @@ spec:
           # access node list via kubectl and generate iptables rules
           # for DNAT from it.
           while sleep 5; do
+            echo 1 > /proc/sys/net/ipv4/ip_forward
             iptables -t nat -F host_dnat_inactive
             # query nodes and create DNAT rules in inactive chain
             kubectl get nodes -o json \
@@ -92,8 +93,9 @@ spec:
           echo "Terminating."
 
         securityContext:
-          capabilities:
-            add: ["NET_ADMIN"]
+          # Needed to be able to enable ip_forward, there is no capability that mounts
+          # /proc with RW
+          privileged: true
         resources:
           requests:
             cpu: 25m
--- a/openshift_addons/openvpn/openvpn-client-dep.yaml
+++ b/openshift_addons/openvpn/openvpn-client-dep.yaml
@@ -73,6 +73,7 @@ spec:
           # access node list via kubectl and generate iptables rules
           # for DNAT from it.
           while sleep 5; do
+            echo 1 > /proc/sys/net/ipv4/ip_forward
             iptables -t nat -F host_dnat_inactive
             # query nodes and create DNAT rules in inactive chain
             kubectl get nodes -o json \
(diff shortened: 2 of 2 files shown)


alvaroaleman:
cc @multi-io

addons/openvpn/openvpn-client-dep.yaml
```diff
@@ -73,6 +73,7 @@ spec:
           # access node list via kubectl and generate iptables rules
           # for DNAT from it.
           while sleep 5; do
+            echo 1 > /proc/sys/net/ipv4/ip_forward
```
multi-io, line 76: Are we sure that fs is writable here?

alvaroaleman: I just checked by setting to zero via `kubectl exec`. My `kubectl` froze, I consider that a success

multi-io: I checked in the broken cluster too, it's RO there. Have to figure out why then.

(The initial manual fix I did while nsenter-ed into the pod's NW namespace on the host -- /proc/sys is writable then)

alvaroaleman: Oh okay, interesting, althought I'd say this change is still useful then because it would give you an error (that you can not easely get but well, better than going through the whole chain like we did).

thz: It will be read-only for unprivileged pods (/proc mounted r/o). The right way to set ip_forward would be by sysctl annotations which would also need that sysctl be added to the allowed unsafe ones. reference: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

the nsenter is of course a workaround.

multi-io: The script terminates at that point (because of bash -e), so it might actually make things worse :)

Patched the addon in our dev system, can reproduce that the FS is RO. Weird. May be related to docker runtime, looking into it.

multi-io: > It will be read-only for unprivileged pods (/proc mounted r/o). The right way to set ip_forward would be by sysctl annotations which would also need that sysctl be added to the allowed unsafe ones. reference: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

I saw that document too, but didn't really want to enable this on the kubelet if it can be helped. Or is there another way?

alvaroaleman: Okay interesting. Thinking about it I probably executed the thing from the openvpn container which has `privileged: True`.

If this is a global setting, wouldn't it be simpler to just make the dnat controller container privileged as well?

multi-io: > It will be read-only for unprivileged pods (/proc mounted r/o)

> If this is a global setting, wouldn't it be simpler to just make the dnat controller container privileged as well?

Thanks. I tried adding CAP_SYS_ADMIN, that didn't make a difference. Setting privileged makes it work here.

thz: Yes - unfortunately there is no specific cap for that, as it is the /proc mount option only...

PR #4219, opened 2019-08-30
fix data race condition

zreigz:
**What this PR does / why we need it**: fix data race condition for unit tests. 

Fixes #

```release-note
NONE
```


alvaroaleman:
Just for my understanding, why was this issue happening and why does this change fix it?

zreigz:
I was able to reproduce this data race condition every time and output pointed out those two places all the time.
I didn't discover the root cause. I discovered that adding `time.Local = time.UTC` in any init function cause the problem

PR #4225, opened 2019-09-01
AWS: Decouple Get{Zone,Subnet,VPC} from cloud provider implementation

alvaroaleman:

Currently, the `GetZone`, `GetSubnet` and `GetVPC` funcs in the AWS cloudprovider package require an initialized cloudprovider.

Since we do not want to save secrets on the Cluster CR anymore, we require cloudproviders to contain a `provider.SecretKeySelectorValueFunc` in order to be able to resolve the secrets for their cluster.

However, the `GetZone` and `GetSubnet` are not necesarily used with a cluster and `GetVPC` is never.

This forced us to pass around a `SeedClientGetter` to be used when one of the aforementioned clusters does not have a Cluster available, because by default such a seed client is obtained by using a cluster to satisfy the cloudproviders `New`, even thought it will never be used because its only needed to resolve secrets from a cluster.

This PR decouples the GetZone, GetSubnet and GetVPC in `pkg/cloudprovider/provider/aws` funcs from the cloudprovider implementation. The cloudprovider implementation has basically nothing to do with those  and the state the cloudproviuder requires is not required for them, so there is no reason at all why they should be coupled.

We could even think about moving them from `pkg/cloudprovider/provider/aws` to `pkg/handler/v1/provider/`.

Fixes #

I've manually tested that the endpoints still work, both before a cluster exists and afterwards.

**Documentation**:

```release-note
none
```

/assign @thz @zreigz

### DIFF
--- a/api/cmd/kubermatic-api/main.go
+++ b/api/cmd/kubermatic-api/main.go
@@ -225,7 +225,6 @@ func createInitProviders(options serverRunOptions) (providers, error) {
 		eventRecorderProvider:                 eventRecorderProvider,
 		clusterProviderGetter:                 clusterProviderGetter,
 		seedsGetter:                           seedsGetter,
-		seedClientGetter:                      seedClientGetter,
 		addons:                                addonProviderGetter}, nil
 }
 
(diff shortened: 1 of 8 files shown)


zreigz:
works fine, thanks for this.
/approve
/lgtm

PR #4227, opened 2019-09-01
Increase default service cidr from 10.10.10.0/24 to 10.240.16.0/20

alvaroaleman:

Chages the default service cidr for new clusters from 10.10.10.0/24 to 10.240.16.0/20.  A maximum of 254 services is very small, because of that we even have a guide on how change this: https://docs.kubermatic.io/operation/control-plane/increase-service-pod-cidr/

A maximum of 4k services should hopefully cover almost all users.

Also fixes the IP calculation for the kubernetes.default and the kube-dns service. It uses the IP returned from `net.ParseCIDR()` which unfortunately has no netmask set, which means the result is invalid if something other than the network address is passed.

Fixes #

**Documentation**:

```release-note
The default service CIDR for new clusters was increased and changed from 10.10.10.0/24 to 10.240.16.0/20
```

/assign @thz 
@thz can you think of anything this collides with? I chose is pretty randomly. Also it seems we basically don't have any kind of sanity checking for CIDR, so if there are colissions, users will notice it only because stuff doesn't work :grimacing:

### DIFF
--- a/api/cmd/kubeletdnat-controller/main.go
+++ b/api/cmd/kubeletdnat-controller/main.go
@@ -29,10 +29,11 @@ func main() {
 
 	log.SetLogger(log.ZapLogger(false))
 
-	nodeAccessNetwork, _, err := net.ParseCIDR(*networkFlag)
+	_, network, err := net.ParseCIDR(*networkFlag)
 	if err != nil {
 		glog.Fatalf("node-access-network invalid or missing: %v", err)
 	}
+	nodeAccessNetwork := network.IP
 
 	config, err := clientcmd.BuildConfigFromFlags(*master, *kubeconfigFlag)
 	if err != nil {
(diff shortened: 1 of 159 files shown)


alvaroaleman:
Lets verify this doesn't break any of our e2e tests:

/test pull-kubermatic-e2e-do-centos-1.15
/test pull-kubermatic-e2e-gcp-coreos-1.15
/test pull-kubermatic-e2e-azure-coreos-1.15
/test pull-kubermatic-e2e-packet-coreos-1.15
/test pull-kubermatic-e2e-kubevirt-centos-1.15
/test pull-kubermatic-e2e-hetzner-ubuntu-1.15
/test pull-kubermatic-e2e-openstack-coreos-1.15
/test pull-kubermatic-e2e-vsphere-coreos-1.15

thz:
> /assign @thz
> @thz can you think of anything this collides with? I chose is pretty randomly. Also it seems we basically don't have any kind of sanity checking for CIDR, so if there are colissions, users will notice it only because stuff doesn't work 😬

Well there is no way anyways to avoid collisions in global address assignment. The IANA tried but then they ran out of 2^32 public address space. So people are fleeing into private address spaces by willingly overlapping with others. And here we are back at the start of that cycle ;)

PR #4230, opened 2019-09-02
chart-sync: automatically plug in the latest dashboard tag from a matching release branch

kdomanski:
```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-sync-charts.sh
+++ b/api/hack/ci/ci-sync-charts.sh
@@ -29,7 +29,7 @@ fi
 # create fresh clone of the dashboard repository
 cd "$(mktemp -d)"
 git clone git@github.com:kubermatic/kubermatic-installer.git .
-MINOR_VERSION="$(sed 's:release/::' <<< "$INSTALLER_BRANCH")"
+MINOR_VERSION="${INSTALLER_BRANCH##release/}"
 if ! FOUND_TAGS="$(git for-each-ref refs/tags --sort=-authordate --format='%(refname)' | grep $MINOR_VERSION)"; then
   echo "Error, no Dashboard tags contain $MINOR_VERSION"
   exit 1
--- a/api/hack/ci/ci-sync-charts.sh
+++ b/api/hack/ci/ci-sync-charts.sh
@@ -26,18 +26,8 @@ if [[ -z ${INSTALLER_BRANCH} ]]; then
   exit 1
 fi
 
-# create fresh clone of the dashboard repository
-cd "$(mktemp -d)"
-git clone git@github.com:kubermatic/kubermatic-installer.git .
-MINOR_VERSION="${INSTALLER_BRANCH##release/}"
-FOUND_TAG="$(git for-each-ref "refs/tags/$MINOR_VERSION*" --sort=-authordate --format='%(refname)' --count=1)"
-if [ -z "$FOUND_TAG" ]; then
-  echo "Error, no Dashboard tags contain $MINOR_VERSION"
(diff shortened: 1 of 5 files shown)


api/hack/ci/ci-sync-charts.sh
```diff
@@ -26,6 +26,19 @@ if [[ -z ${INSTALLER_BRANCH} ]]; then
   exit 1
 fi
 
+# create fresh clone of the dashboard repository
+cd "$(mktemp -d)"
+git clone git@github.com:kubermatic/kubermatic-installer.git .
+MINOR_VERSION="$(sed 's:release/::' <<< "$INSTALLER_BRANCH")"
```
thz: You are nesting the double quotes here. I propose to use the following instead:
```
MINOR_VERSION="${INSTALLER_BRANCH##release/}"
```

thz (changes requested):
Just a minor substitution nit.

thz (approved):

thz:
It just got so much better ;-)
/lgtm

config/kubermatic/values.yaml
```diff
@@ -183,7 +183,7 @@ kubermatic:
     replicas: 2
     image:
       repository: "quay.io/kubermatic/ui-v2"
-      tag: "v1.3.0"
+      tag: "__DASHBOARD_TAG__"
```
alvaroaleman, line 186: This breaks the deployment to all of our envs. It must stay consistent on master branch and it must be valid.

The way this works is:

* Helm doesn't do anything about the UI if nothing changes in the helm manifests and values
* There is a postsubmit in the dashboard repo that updates the tag on every push to master

/hold

kdomanski: Yes, it's marked as [WIP].

api/hack/lib.sh
```diff
@@ -61,3 +61,26 @@ write_junit() {
...
 EOF
 }
+
+get_latest_dashboard_tag() {
+  FOR_BRANCH="$1"
+
+  git config --global core.sshCommand 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /ssh/id_rsa'
+  local DASHBOARD_URL="git@github.com:kubermatic/dashboard-v2.git"
+
+  if [[ "$FOR_BRANCH" == "master" ]]; then
```
alvaroaleman: Rather than introducing some magic exception to not get a tag but a sha in a func named `get_latest_tag`, can we just not use this func in `deploy.sh` and instead do a simple ` git ls-remote "$DASHBOARD_URL" refs/heads/$PULL_BASE_REF | awk '{ print $1 }'` there?

In contrast to the current implementation, that would also work properly when we are in a release branch.

kdomanski: Wouldn't removing the exception break PRs opened against master?

alvaroaleman: The most correct way of doing this would be to have `hack/ci/ci-run-minimal-conformance-tests.sh` also figure out the latest hash of the corresponding base branch rather than the latest tag.

In reality this has very little relevance thought, as the only thing these tests test wrt UI is that it doens't crashloop. So as long as `get_latest_dashboard_tag` returns _any_ tag, that should be fine.

So basically, whatever you prefer, both what is currently done and changing it would work.

alvaroaleman:
@kdomanski can you not override tests unless you know for sure they are broken? The correct approach is to fix the test, not to override it. The particular test you are overriding here got fixed yesterday in #4259

PR #4231, opened 2019-09-02
Openshift: Deploy dns operator

### LINKED ISSUE #4170: Set up DNS for openshift 4
**User Story**

As a user of Openshift 4 I want DNS that behaves the same as with upstream

**Acceptance criteria**

Openshift four in Kubermatic uses the same dns images and config as upstream Openshift

alvaroaleman:

Fixes #4170

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/openshift/resources/dns_operator.go
+++ b/api/pkg/controller/openshift/resources/dns_operator.go
@@ -1,0 +1,124 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/apiserver"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	utilpointer "k8s.io/utils/pointer"
+)
+
+const openshiftDNSOperatorDeploymentName = "openshift-dns-operator"
+
+var (
+	openshiftDNSOperatorResourceRequirements = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("50Mi"),
+			corev1.ResourceCPU:    resource.MustParse("10m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("200Mi"),
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+		},
+	}
+)
+
+func OpenshiftDNSOperatorFactory(data openshiftData) reconciling.NamedDeploymentCreatorGetter {
+	return func() (string, reconciling.DeploymentCreator) {
+		return openshiftDNSOperatorDeploymentName, func(d *appsv1.Deployment) (*appsv1.Deployment, error) {
+
+			image, err := openshiftDNSOperatorImage(data.Cluster().Spec.Version.String())
+			if err != nil {
+				return nil, err
+			}
+
+			env, err := openshiftDNSOperatorEnv(data.Cluster().Spec.Version.String())
+			if err != nil {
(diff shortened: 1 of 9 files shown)


api/pkg/controller/openshift/resources/dns_operator.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+			d.Spec.Template.Spec.AutomountServiceAccountToken = utilpointer.BoolPtr(false)
+			// Configure user cluster DNS resolver for this pod.
+			d.Spec.Template.Spec.DNSPolicy, d.Spec.Template.Spec.DNSConfig, err = resources.UserClusterDNSPolicyAndConfig(data)
+			if err != nil {
+				return nil, err
+			}
+
+			d.Spec.Template.Spec.Containers = []corev1.Container{{
+				Name:    "dns-operator",
```
moadqassem: Can you please add a const for the value `dns-operator` and use it in each occurrence?

api/pkg/controller/openshift/resources/dns_operator.go
```diff
@@ -0,0 +1,124 @@
...
+			d.Spec.Template.Spec = *wrappedPodSpec
+
+			return d, nil
+		}
+	}
+}
+
+func openshiftDNSOperatorEnv(openshiftVersion string) ([]corev1.EnvVar, error) {
+	switch openshiftVersion {
+	case "4.1.9":
```
moadqassem: Same here, please create a const for this value. This value here is occurring +10 time in different files in this package. Thus I would suggest you create a file and add all the constants that are shared and used among the files in this package.

api/pkg/controller/openshift/resources/dns_operator.go
```diff
@@ -0,0 +1,124 @@
...
+					},
+				},
+			}
+
+			d.Spec.Template.Labels, err = data.GetPodTemplateLabels(openshiftDNSOperatorDeploymentName, d.Spec.Template.Spec.Volumes, nil)
+			if err != nil {
+				return nil, err
+			}
+
+			d.Spec.Template.Spec.InitContainers = nil
```
moadqassem, line 97: Why this is needed?

api/pkg/controller/openshift/resources/dns_operator.go
alvaroaleman: In this example I disagree, because while its important that this is the same everywhere, its at least as important that it has a value of `4.1.9`. The latter implies the former and recognizing the latter is easier when there is no additional indirection.

alvaroaleman:
@moadqassem addressed all comments, PTAL

api/pkg/controller/openshift/resources/dns_operator.go
moadqassem: Sorry, I don't quite follow the point here. Usually when the code is heavily dependent on a version and the code will change dramatically if the version changes, we create a package that has the same version name and lay the code for that version in there. However, in your code you are just adjusting and selecting properties(e.g: images) and to be honest when we will have multiple versions, for example, this will get even nastier to refactor and maintain.

alvaroaleman: But the code doesn't change dramatically, some properties of the Deployment change which is why there are func that switch env vars or image based on the version. What would we gain by putting the func into a package with the version name? And why is it hard to maintain with multiple versions, we just have to add a new version to the switch statement or am I missing something?

alvaroaleman: If you have a concrete example on how you imagine that to look like, that would be helpful.

moadqassem: I know that it doesn't change dramatically and that's why I said the last sentence which starts with However :).

alvaroaleman: The case this would help a lot with is if we remove support for a given version, because then all places that reference it result in a compilation error.

moadqassem: You missed adding the const in those lines:
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/openshift/openshift_controller_test.go#L116

https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/openshift/resources/openshift_network_operator.go#L123

https://github.com/kubermatic/kubermatic/pull/4231/files#diff-2f1f7e99f0195657f01e0dc9250235a5R113

PR #4236, opened 2019-09-02
Addon controller: Retry when kubeconfig wasnt found

alvaroaleman:

Makes the addon controller gracefully cope with an absent `admin-kubeconfig` secret. This is needed for Openshift, which doesn't support token auth, so in order to have a revocable kubeconfig, we create a serviceaccount inside the user cluster. That however takes a bit more time, which made the addon controller always throw some initial errors.

/assign @xrstf 

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/addon/addon_controller.go
+++ b/api/pkg/controller/addon/addon_controller.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


api/pkg/controller/addon/addon_controller.go
```diff
@@ -155,75 +156,88 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 		log.Debug("Skipping because the API server is not running")
-		return nil
+		return nil, nil
+	}
+
+	// Openshift needs some time to create this, so avoid getting into the backoff
+	// while the admin-kubeconfig secret doesn't exist yet
+	if _, err := r.KubeconfigProvider.GetAdminKubeconfig(cluster); err != nil {
+		if kerrors.IsNotFound(err) {
+			log.Debug("Skipping because the kubeconfig wasn't found")
```
xrstf: Can you reword this to `Retrying later because ...`? Compared to the other conditions where we "skip" a reconcile request, this case is special.

PR #4241, opened 2019-09-03
Cleanup fluent-bit Chart

xrstf:
First of all, this looks much worse than it really is.

I had the urge to understand our fluentbit chart better and improve the parsing rules for containers that emit multiple styles of logging. I also investigated how easy it would be to define automated tests for our fluent-bit configuration (spoiler: not easily possible). After restructuring the config to fit my local experimental setup, I applied the same changes to our chart.

These are the significant changes:

* The configuration is now done in dedicated files instead of a giant ConfigMap template. The `.conf` files are treated as templates, so we do not lose flexibility. This change caused a gigantic diff, even though most of the configuration is identical.
* The chart has been, like many others, prepared to be installed multiple times. This makes it much, much easier to test changes because I can simply deploy a second DaemonSet and pump the new logs into a new Elasticsearch index. However, this leads to at least one BC break: the offset database (`/var/lib/flb_kube.db`) now has a generic name depending on the release name (`/var/lib/flb_{{ .Release.Name }}.db`). This means that after deploying the new chart version, it will re-scan older logs and re-submit them. Sad, but also nice because this also makes use of the new parsing rules.
* New parsers have been defined:
  * nginx_ingress_controller
  * kibana (this one is important, because Kibana's logs contain an `@timestamp` field and that conflicts with Elasticsearch; we need to rename/parse this field explicitly to avoid this conflict)
  * elasticsearch_json (for ES 7.0+ with JSON logging, i.e. a future PR)
* The `docker` parser has been refactored. Docker's JSON logs contain a `time` and a `log` field and setting `Keep_Time` or the field decoding stuff was pointless and had no effect on how the metrics were parsed (this might be due to the many JSON-related fixes in FB 1.1 and 1.2).
* The parsers for Apache logs have been removed. We don't use them and if someone needs them, see the following point:
* It's now possible to define new parsers via values.yaml, just like you can override the outputs. I chose not to make all parsers overridable, but just offer an option to add _new_ parsers.
* The chart now uses the modern label names (`app.kubernetes.io/name` etc.). This makes it impossible to just upgrade an existing release, but since we changed other stuff as well I think it's worth it. FB has no real state (except for the offset DB, as discussed earlier), so purging and re-installing the chart should not be too bad.
* The environment variables for configuring the ES output are gone. Why have the output be variable in the values.yaml and then hardcode the ES service name and port in the DaemonSet template? Just put the damn service name and port into your values.yaml and you're done.

Multiple log formats are now supposed to be handled by having FB filters that try to apply a series of parsers. So the nginx-ingress-controller uses the special nginx_ingress_controller parser (that is capable of parsing the access.log-like lines) and then FB will apply all other parser filters and one of them is glog and hopefully able to parse the remaining lines.

---

And now that we all agree that the chart needs to be re-installed, I took the opportunity to finally rename it to `fluent-bit`. We use this spelling in many other places, it matches the Github repo name, the FB binary name, etc.

---

I also added a tiny dashboard for 4 basic Fluent Bit metrics.

```release-note
overhauled fluent-bit chart (see release notes for migration steps)
```


config/logging/fluentbit/values.yaml
```diff
@@ -8,13 +8,14 @@ logging:
...
       collectKernelMessages: false
+      parsers: []
       outputs:
       - |
         Name            es
         Alias           elasticsearch
         Match           *
-        Host            ${FLUENT_ELASTICSEARCH_HOST}
-        Port            ${FLUENT_ELASTICSEARCH_PORT}
+        Host            es-data
```
xmudrii, line 17: Is having the host and port defined statically an expected change?

xrstf: This is now less static than before. Before we had explicit environment variables on the DaemonSet template that you could not change. Now you can configure your ES endpoint where it actually makes sense: at the same point where you define the whole output section.

xrstf:
From what I've seen on the CI cluster, the new config works pretty well and this should be ready to be reviewed.

xrstf:
Closed in favor of just updating the ruleset in #4395. But I'm hopeful that someday in the future we will fix the naming issue once and for all.

PR #4242, opened 2019-09-03
Correctly return openstack.useFloatingIP in NodeDeployment API

bashofmann:
Correctly return openstack.useFloatingIP in NodeDeployment API. 

This fixes editing a NodeDeployment which uses floating IPs when it's not required by the datacenter.

```release-note
NONE
```

### DIFF
--- a/api/pkg/machine/convert.go
+++ b/api/pkg/machine/convert.go
@@ -118,10 +118,10 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 			Image:  config.Image.Value,
 			Tags:   config.Tags,
 		}
+		cloudSpec.Openstack.UseFloatingIP = config.FloatingIPPool.Value != ""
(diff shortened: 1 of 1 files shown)


api/pkg/machine/convert.go
```diff
@@ -121,6 +121,7 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 		if config.RootDiskSizeGB != nil && *config.RootDiskSizeGB > 0 {
 			cloudSpec.Openstack.RootDiskSizeGB = config.RootDiskSizeGB
 		}
+		cloudSpec.Openstack.UseFloatingIP = len(config.FloatingIPPool.Value) > 0
```
alvaroaleman: Nit: Could you change the right side to `config.FloatingIPPool.Value != ""`? I was actually wondering why `FloatingIpPool.Value` is a slice when looking at the code

ALso maybe move it up to line 119?

PR #4256, opened 2019-09-05
sync-charts: point to the correct ssh key location in line with prow preset

kdomanski:
```release-note
NONE
```


alvaroaleman:
err are there more places where we might need this? So far the ssh key has been in a default location. Why change that?

kdomanski:
A dozen other jobs had it in `/ssh/id_rsa` due to `api/hack/run-kubermatic-upgrade-test.sh:14`

PR #4258, opened 2019-09-05
Add option to auto update nodes

alvaroaleman:

Adds the option to configure `automaticNodeUpdate` in the updates file which will then automatically update the nodes.

Fixes #4257 

**Documentation**:

```release-note
It is now possible to also configure automatic node updates by setting `automaticNodeUpdate: true` in the `updates.yaml`. This option implies `automatic: true` as node versions must not be newer than the version of the corresponding controlplane.
```

/hold
Until I have tried out if this actually works
/assign @thz


api/pkg/version/manager_test.go
```diff
@@ -0,0 +1,71 @@
...
+			controlPlaneVersion: "1.6.0",
+			updates: []*Update{{
+				From:                "1.5.0",
+				To:                  "1.6.0",
+				AutomaticNodeUpdate: true,
+			}},
+			expectedVersion: &Version{Version: semver.MustParse("1.6.0")},
+		},
+		{
+			name:                "Node compatibity check fails, error",
```
xrstf: `compatibility`

api/pkg/controller/update/update_controller.go
```diff
@@ -75,40 +75,96 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+
+func (r *Reconciler) nodeUpdate(ctx context.Context, cluster *kubermaticv1.Cluster, clusterType string) error {
+	client, err := r.userClusterConnectionProvider.GetClient(cluster)
+	if err != nil {
+		return fmt.Errorf("failed to get usercluster client: %v", err)
+	}
+
+	machineDeployments := &clusterv1alpha1.MachineDeploymentList{}
+	// Kubermatic only creates MachineDeployments in the kube-system namespace, everything else is essentially unsupported
+	listOpts := &ctrlruntimeclient.ListOptions{Namespace: "kube-system"}
```
xrstf, line 137: Is it worth using `metav1.NamespaceSystem` instead of the string literal here?

alvaroaleman: I hate those constants, they make stuff harder to read without providing any added value I can see

api/pkg/version/load.go
```diff
@@ -21,19 +21,26 @@ func LoadUpdates(path string) ([]*MasterUpdate, error) {
...
 
 	err = yaml.UnmarshalStrict(bytes, &s)
 	if err != nil {
 		return nil, err
 	}
+	for _, update := range s.Updates {
+		// AutomaticNodeUpdate implies automatic update, because nodes
+		// must not have a newer version than the control plane
+		if update.AutomaticNodeUpdate && !update.Automatic {
+			update.Automatic = true
```
xrstf, line 35: `update.Automatic |= update.AutomaticNodeUpdate` ? At the very least, the `&& !update.Automatic` is redundant, no?

api/pkg/version/load_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	defer file.Close()
+	defer os.Remove(file.Name())
+
+	if _, err := file.Write(fileContent); err != nil {
+		t.Fatalf("failed to write to tempfile: %v", err)
+	}
+
+	updates, err := LoadUpdates(file.Name())
+	if err != nil {
```
xrstf, line 28: Can you add a `|| len(updates) == 0`? Otherwise the test might panic if loading didn't fail, but also didn't succeed.

api/pkg/version/manager.go
```diff
@@ -18,27 +19,28 @@ var (
...
 type Manager struct {
-	versions []*MasterVersion
-	updates  []*MasterUpdate
+	versions []*Version
+	updates  []*Update
 }
 
-// MasterVersion is the object representing a Kubernetes Master version.
-type MasterVersion struct {
+// Version is the object representing a Kubernetes Master version.
```
xrstf: is it really representing a *master version*, or just a regular version?

api/pkg/version/manager.go
```diff
@@ -115,34 +117,65 @@ func (m *Manager) GetMasterVersions(clusterType string) ([]*MasterVersion, error
 	return masterVersions, nil
 }
 
-// AutomaticUpdate returns a version if an automatic update can be found for version sfrom
```
xrstf, line 118: Here and a couple of lines below that you make still references to `sfrom`.

api/pkg/version/manager_test.go
alvaroaleman: @xrstf Would you mind extending the existing lint presubmit/adding a new one that checks spelling?

api/pkg/version/load.go
alvaroaleman: I've removed the `&& !update.Automatic`.

`update.Automatic != update.AutomaticNodeUpdate` would be wrong, because then we would disable `update.Automatic` when ` update.AutomaticNodeUpdate` is `false`

api/pkg/version/load_test.go
alvaroaleman: Done

api/pkg/version/manager.go
alvaroaleman: Updated

api/pkg/version/manager.go
alvaroaleman: Updated

PR #4260, opened 2019-09-05
Defensive Seed Proxy Reconciling

xrstf:
This makes sure that the seed-proxy controller does nothing if the required namespaces do not exist. So if there is no `monitoring-master` namespace, it will not attempt to put Grafana stuff in there.

This PR also replaces glog with Zap, uses the new predicate factory and makes use of the caching SeedGetterFactory.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/seed-proxy/reconciler.go
+++ b/api/pkg/controller/seed-proxy/reconciler.go
@@ -44,54 +44,64 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	log := r.log.With("seed", request.NamespacedName.String())
 	log.Debugw("reconciling seed")
 
+	return reconcile.Result{}, r.reconcile(request.Name, log)
+}
+
+func (r *Reconciler) reconcile(seedName string, log *zap.SugaredLogger) error {
 	seeds, err := r.seedsGetter()
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get seeds: %v", err)
+		return fmt.Errorf("failed to get seeds: %v", err)
 	}
 
 	log.Debug("garbage-collecting orphaned resources...")
 	if err := r.garbageCollect(seeds, log); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to garbage collect: %v", err)
+		return fmt.Errorf("failed to garbage collect: %v", err)
 	}
 
-	seed, found := seeds[request.Name]
+	seed, found := seeds[seedName]
 	if !found {
-		return reconcile.Result{}, fmt.Errorf("didn't find seed %q", request.Name)
+		return fmt.Errorf("didn't find seed %q", seedName)
 	}
 
 	// both of these namespaces must exist or else we cannot establish
 	// the service account token sharing
-	if !r.namespaceExists(r, MasterTargetNamespace, log) {
+	err = r.Get(r.ctx, types.NamespacedName{Name: MasterTargetNamespace}, &corev1.Namespace{})
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check for namespace %s: %v", MasterTargetNamespace, err)
+		}
+
 		log.Debug("skipping because target master namespace does not exist", "namespace", MasterTargetNamespace)
-		return reconcile.Result{}, nil
+		return nil
 	}
 
 	client, err := r.seedClientGetter(seed)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -278,3 +302,16 @@ func (r *Reconciler) ensureMasterGrafanaProvisioning(seeds map[string]*kubermati
 
 	return nil
 }
+
+func (r *Reconciler) namespaceExists(client ctrlruntimeclient.Client, namespace string, log *zap.SugaredLogger) bool {
+	ns := corev1.Namespace{}
+	if err := client.Get(r.ctx, types.NamespacedName{Name: namespace}, &ns); err != nil {
+		if !kerrors.IsNotFound(err) {
```
alvaroaleman: Wouldn't it make more sense to return an error to differentiate `namespaceDoesNotExist` from `some random error occured, retry later`?

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -78,114 +106,110 @@ func (r *Reconciler) garbageCollect(seeds map[string]*kubermaticv1.Seed) error {
 	}
 
 	if err := r.List(r.ctx, options, list); err != nil {
-		return fmt.Errorf("failed to list Secrets: %v", err)
+		return fmt.Errorf("failed to list secrets: %v", err)
```
alvaroaleman: This one I like upercase, it makes clear that we are talking about the `kind: Secret`

api/pkg/controller/seed-proxy/reconciler.go
xrstf: I can't see right now how that would make the code easier. Can you help me? Wouldn't a caller then need to check for three different states (no error, notFoundErr and randomOtherErr)?

alvaroaleman: Its actually not even about being easier, its about being correct.

If we encounter an error here, the caller will assume that the namespace doesn't exist, return and not do anything.

If we return the error, the callers knows there was an error, it gets propagated up and will result in the reconciliation being retried after the backoff period.

xrstf:
@alvaroaleman you can review this again

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -37,39 +41,73 @@ type Reconciler struct {
 // an error if any API operation failed, otherwise will return an empty
 // dummy Result struct.
 func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	glog.V(4).Infof("Reconciling seed %q", request.NamespacedName.String())
+	log := r.log.With("seed", request.NamespacedName.String())
+	log.Debugw("reconciling seed")
```
alvaroaleman: s/Debugw/Debug

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -37,39 +41,73 @@ type Reconciler struct {
...
+		return fmt.Errorf("failed to get seed client: %v", err)
+	}
+
+	err = client.Get(r.ctx, types.NamespacedName{Name: SeedServiceAccountNamespace}, &corev1.Namespace{})
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check for namespace %s: %v", SeedServiceAccountNamespace, err)
+		}
+
+		log.Debug("skipping because seed namespace does not exist", "namespace", SeedServiceAccountNamespace)
```
alvaroaleman, line 89: s/Debug/Debugw

api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -95,49 +133,49 @@ func (r *Reconciler) garbageCollect(seeds map[string]*kubermaticv1.Seed) error {
...
+	log.Debug("reconciling ServiceAccounts...")
+	if err := r.reconcileSeedServiceAccounts(client, log); err != nil {
+		return fmt.Errorf("failed to ensure ServiceAccount: %v", err)
 	}
 
-	glog.V(4).Info("Reconciling roles...")
-	if err := r.ensureSeedRoles(client); err != nil {
-		return fmt.Errorf("failed to ensure role: %v", err)
+	err = client.Get(r.ctx, types.NamespacedName{Name: SeedPrometheusNamespace}, &corev1.Namespace{})
+	if err != nil && !kerrors.IsNotFound(err) {
```
alvaroaleman: Maybe do one condition and add debug log for why it is being skipped if so?

PR #4261, opened 2019-09-06
Reading kubevirt kubeconfig from credentials

moadqassem:
This PR creates credentials for the Kubevirt cluster to read the Kubeconfig from the secretes instead of reading them from the cluster object directly

No

```release-note
NONE
```


alvaroaleman:
/retitle Reading kubevirt kubeconfig from credentials

This will be the commit title one this PR is merged, the leading number does not really have to be in the title.

moadqassem:
@alvaroaleman Yes sorry about that I opened the PR via IntelliJ and it took the name from the commit message I guess.

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -394,6 +394,8 @@ type GCPCloudSpec struct {
 
 // KubevirtCloudSpec specifies the access data to Kubevirt.
 type KubevirtCloudSpec struct {
+	CredentialsReference *providerconfig.GlobalSecretKeySelector `json:"credentialsReference,omitempty"`
+
 	Kubeconfig string `json:"kubeconfig"`
```
alvaroaleman: Could you add an `,omitempty` to the json tag so this field is only marshalled if it has a value?

alvaroaleman (commented):
One point and lets make sure the kubevirt e2e tests still pass, other than that, LGTM

PR #4262, opened 2019-09-06
filter presets

### LINKED ISSUE #3806: Filter predefined environments
**User Story**
Due to the customer feedback: https://github.com/kubermatic/customer-projects/issues/76 

A operator of Kubermatic want to set a filter for the predefined environments. The reason is that different user groups will access our platform in a self services. To be able to divide different parties in a company we need  a possibility to filter different environments, e.g.:
*  `user@company-A.com` -> see `company A environment` in the selection of environments
*  `user@company-B.com` -> see `company B environment` in the selection of environments  

Possible solutions could be to filter based on:
- regex on the user e-mail address: `.*@company-A.com` match to preset `customer-A` -> so only the company A user would see this preset.
- similar logic as above but matching some OIDC group names

The filter should be added as config parameter to the backend config. We should go first the way with the smallest effort. 

**Acceptance criteria**
- [ ] user's of different groups/company will see different predefined environments
- [ ] the usage of the filters is document for Kubermatic operators

zreigz:
**What this PR does / why we need it**: Filter presets. First step to move presets to CRD.
An operator of Kubermatic wants to set a filter for the presets. The reason is that different user groups will access our platform. To be able to divide different parties in a company we need a possibility to filter different environments, e.g.:
user@company-A.com -> see company A presets in the selection of environments
user@company-B.com -> see company B presets in the selection of environments

Fixes #3806

Depends on: https://github.com/kubermatic/secrets/pull/363

```release-note
Filter presets
[ACTION REQUIRED] change presets.yaml:
presets:
  items:
    - metadata:
        name: example
      spec:
        requiredEmailDomain: example.com
        aws:
          accessKeyId: 
          secretAccessKey: 
          vpcId: 
        azure:
          tenantId: 
          subscriptionId: 
          clientId: 
          clientSecret: 
        digitalocean:
          token: 
        gcp:
          serviceAccount: 
        hetzner:
          token: 
        openstack:
          username: 
          password: 
          tenant: 
          domain:
          floatingIpPool: 
        packet:
          apiKey: 
          projectId: 
        vsphere:
          username: 
          password: 

```


zreigz:
this PR should break the API tests. The new test image is required with correct presets.yaml. To merge this PR we need to disable the `pull-kubermatic-api-e2e`, fix it, rebuild the image and enable again

alvaroaleman:
@zreigz This PR is huge. Can we please divide it into "Filter presets" and "Move Presets to CRD types"?

alvaroaleman:
The reason for that is that the filtering in itself is a delicate piece of business logic we should get right, whereas the type move is more of a refactoring than anything else

zreigz:
@alvaroaleman the filtering uses a new Preset structure and this is the main reason why I did this. Otherwise, I could touch the same code twice. It would be good to have those two things in one.

zreigz:
If it's problematic for you we can sit together on Wednesday and review the changes and debug this

api/pkg/presets/manager.go
```diff
@@ -206,57 +62,56 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+
 	}
 	return &Manager{presets: presets}, nil
 }
 
-func (m *Manager) GetPresets() *Presets {
-	return m.presets
+func (m *Manager) GetPreset(userInfo provider.UserInfo) *kubermaticv1.Preset {
+	for _, preset := range m.presets.Items {
+		emialDomain, ok := preset.Labels[kubermaticv1.PresetEmailDomainLabel]
```
alvaroaleman: Can we not do this via a label but instead add fields to the presets `.Spec`?

Something like `requiredEmailDomain` for the current case, which would then be an optional parameter that _if set_ requires users to have the correct email domain.

Generally we shouldn't be using labels for this, because we want to document both within our code and externally what configuration settings exist and what they are supposed to do, this is very hard if we don't properly type them but use labels which at the end of the day are just part of a `map[string]string`

zreigz: good point, thanks

api/pkg/handler/v1/presets/credentials.go
```diff
@@ -45,22 +48,28 @@ func CredentialEndpoint(credentialManager common.PresetsManager) endpoint.Endpoi
 			return nil, errors.NewBadRequest(err.Error())
 		}
 
+		userInfo := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
```
alvaroaleman: Can we please check this type assertion?

Also IMHO it would be much, much better if  we did the `UserInfoExtraction` inside the `CredentialsEndpoint` because with the current approach, we only see at runtime if we forgot it somewhere. Passing it into the endpoint and calling it everywhere where we need that info would allow the go compiler to tell us if we forgot it somewhere.

api/pkg/presets/manager.go
```diff
@@ -206,57 +62,56 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+
 	}
 	return &Manager{presets: presets}, nil
 }
 
-func (m *Manager) GetPresets() *Presets {
-	return m.presets
+func (m *Manager) GetPreset(userInfo provider.UserInfo) *kubermaticv1.Preset {
+	for _, preset := range m.presets.Items {
+		emialDomain := preset.Spec.RequiredEmailDomain
```
alvaroaleman: Typo and do we really want to make the `RequiredEmailDomain` a mandatory field and not allow ppl to leave it empty in which case everyone can use all presets?

api/pkg/handler/v1/presets/credentials.go
```diff
@@ -45,22 +48,28 @@ func CredentialEndpoint(credentialManager common.PresetsManager) endpoint.Endpoi
...
-				item := credentialItems.Index(i)
-				if item.Kind() == reflect.Struct {
-					v := reflect.Indirect(item)
-					if _, ok := v.Type().FieldByName("Name"); ok {
-						rawName := v.FieldByName("Name").Interface()
-						name := rawName.(string)
-						names = append(names, name)
+		preset := credentialManager.GetPreset(*userInfo)
+		providersRaw := reflect.ValueOf(preset.Spec)
+		if providersRaw.Kind() == reflect.Struct {
```
alvaroaleman: Can you add a comment here explaining what this does? IMHO the code is currently pretty hard to understand and it seems it is basically a workaround for the lack of generics in Go that allows you not to write a loop for each and every provider?

An alternate approach could be btw to make all the Preset types implement a
```
type Namer interface {
Name() string
}
```

Which would allow you to implement a 
```
func getNames(namers []Namer) result []string {
  for _, namer range namers {
    result = append(result, namer.Name())
  }
  return result
}
func (spec PresetSpec) GetPresetNamesForNamedProvider(providerName string) [] string {
switch providerName {
case "DigitalOcean":
    return getNames(spec.DigitalOcean)
[...]
```

api/pkg/presets/manager_test.go
```diff
@@ -3,10 +3,11 @@ package presets_test
 import (
```
alvaroaleman, line 3: Could we also have some tests that test _just_ `GetPreset` and specifically its filtering? The current endpoint tests are very verbose

api/pkg/test/e2e/api/credentials_test.go
```diff
@@ -1,4 +1,4 @@
-// +build e2e
+// +build none
```
alvaroaleman: Why is this disabled?

api/pkg/handler/v1/presets/credentials.go
zreigz: I've added type assertion check and added a new issue to improve UserInfo extraction: #4271

api/pkg/presets/manager.go
zreigz: yes, I've changed this

api/pkg/handler/v1/presets/credentials.go
zreigz: comments added. It's really generic and will work for the new providers

api/pkg/presets/manager_test.go
zreigz: done

api/pkg/test/e2e/api/credentials_test.go
zreigz: fixed

api/pkg/presets/manager.go
```diff
@@ -206,57 +62,66 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+
 	}
 	return &Manager{presets: presets}, nil
 }
 
-func (m *Manager) GetPresets() *Presets {
-	return m.presets
+// GetPreset returns preset which belong to the specific email group.
+// If there is no preset for the email domain then returns preset for all user (the RequiredEmailDomain is empty).
+func (m *Manager) GetPreset(userInfo provider.UserInfo) *kubermaticv1.Preset {
```
alvaroaleman: This confuses me a bit. Why does this only return a single preset and not multiple?

api/pkg/presets/manager.go
```diff
@@ -206,57 +62,66 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+	// find preset for specific email domain
+	for _, preset := range m.presets.Items {
```
alvaroaleman: This can be simplified and done in a single loop:
```
userEmailDomainRaw := strings.Split(userInfo.Email, "@")
var userEmailDomain string
if len(userEmailDomainRaw) == 2 {
  userEmailDomain = userEmailDomainRaw[1]
}
for _, preset := range m.presets.Items {
  if preset.Spec.RequiredEmailDomain != "" && !strings.EqualFold(preset.Spec.RequiredEmailDomain, userEmailDomain) {
    continue
  }
  return preset
}
```

api/pkg/presets/manager.go
zreigz: the preset contains a credential list for providers. It would be difficult to merge credentials from different presets because as a final result user gets the names of the credentials. Different presets can have the same credential names. It would be impossible to distinguish credentials from different presets.

api/pkg/presets/manager.go
zreigz: depends on the priority. If we want to get first preset for a specific group it won't work.
In your code we will get preset for all users first. Preset assigned to specific group won't be returned.

api/pkg/presets/manager.go
alvaroaleman: Okay, but if there are two credentials with the same name in different presets, the user is unable to access the second one which is confusing and hard to explain.

Wouldn'it it make sense to have exactly _one_ credential per preset to enforce globally unique name and prevent this situation?

zreigz: we wanted a list of credentials to have the possibility to pick the correct one depends on the datacenter. But I think if there such need we can extend Preset for another field for the filter. Making unique credential for preset is better.

zreigz:
@alvaroaleman PTAL

api/pkg/presets/manager.go
alvaroaleman: Okay but then we still need to provide a name or not? Currently we assume there is maximum one preset a user can use

api/pkg/crd/kubermatic/v1/preset.go
```diff
@@ -0,0 +1,167 @@
...
+type Preset struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec PresetSpec `json:"spec"`
+}
+
+// Presets specifies default presets for supported providers
+type PresetSpec struct {
+	Digitalocean Digitalocean `json:"digitalocean,omitempty"`
```
alvaroaleman: What about making these pointers to be able to check if a secret for a given provider was populated? Also, that is required for gos marshaller to actually omit that field

api/pkg/crd/kubermatic/v1/preset.go
```diff
@@ -0,0 +1,167 @@
...
+	Packet       Packet       `json:"packet,omitempty"`
+	GCP          GCP          `json:"gcp,omitempty"`
+	Kubevirt     Kubevirt     `json:"kubevirt,omitempty"`
+
+	Fake                Fake   `json:"fake,omitempty"`
+	RequiredEmailDomain string `json:"requiredEmailDomain,omitempty"`
+}
+
+type Digitalocean struct {
+	Credentials DigitaloceanPresetCredentials `json:"credentials,omitempty"`
```
alvaroaleman: Why not directly use the `$PROVIDERPresetCredentials` types and instead have a surronding struct with one field for each?

api/pkg/presets/manager.go
```diff
@@ -206,240 +63,254 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
-func (m *Manager) GetPresets() *Presets {
-	return m.presets
+// GetPresets returns presets which belong to the specific email group and for all users
+func (m *Manager) GetPresets(userInfo provider.UserInfo) []kubermaticv1.Preset {
+	presetList := make([]kubermaticv1.Preset, 0)
+
+	for _, preset := range m.presets.Items {
+		requiredEmailDomain := preset.Spec.RequiredEmailDomain
+		// find preset for specific email domain
+		if len(requiredEmailDomain) > 0 {
```
alvaroaleman: nit: Replace the `len(requiredEmailDomain)` with a `requiredEmailDomain != ""` to make it more obvious this is a string?

api/pkg/presets/manager.go
```diff
@@ -206,240 +63,254 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+			// find preset for "all" without RequiredEmailDomain field
+			presetList = append(presetList, preset)
+		}
+	}
+
+	return presetList
+}
+
+// GetPreset returns presets which belong to the specific email group and for all users
+func (m *Manager) GetPreset(name string) *kubermaticv1.Preset {
```
alvaroaleman: I think we need to filter for the email domain here as well, otherwise ppl can just specify random preset names during cluster creation and will be allowed to use them.

I think it also makes sense to factor out the filtering to make sure the exact same filtering gets applied everywhere

zreigz: OK I will change it

api/pkg/crd/kubermatic/v1/preset.go
zreigz: We already had a discussion about not using pointers in Go: debug, performance. Me and other UI team members trying to avoid using pointers. This is our good practice and we would like to follow this.

api/pkg/crd/kubermatic/v1/preset.go
zreigz: I need a common field name to use reflection to get credentials for different providers. The main idea for the preset was to be able to extend it in the future and credentials are some kind of the presets. This is why I would like to group them.

api/pkg/crd/kubermatic/v1/preset.go
alvaroaleman: Well yeah, the problem with that is:
* It is not obvious for the reader of the code that the field in question is optional
* Nilchecking is very akward and slow, you have to `DeepEqual` with an empty struct of the same type
* We pollute the APIServer with empty fields

Also, Kubernetes api conventions explicitly tell you to use pointers for optional types: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

api/pkg/crd/kubermatic/v1/preset.go
alvaroaleman: I think you could achieve that by just naming the field like the provider

api/pkg/crd/kubermatic/v1/preset.go
zreigz: thanks for the explanation, I've changed it

api/pkg/crd/kubermatic/v1/preset.go
zreigz: done

api/pkg/presets/manager.go
```diff
@@ -206,240 +62,279 @@ func NewFromFile(credentialsFilename string) (*Manager, error) {
...
+	}
+
+	return nil, fmt.Errorf("missing preset '%s' for the user '%s'", name, userInfo.Email)
+}
+
+func filterOutPresets(userInfo *provider.UserInfo, list *kubermaticv1.PresetList) ([]kubermaticv1.Preset, error) {
+	if list == nil {
+		return nil, fmt.Errorf("the preset list can not be nil")
+	}
+	presetList := make([]kubermaticv1.Preset, 0)
```
alvaroaleman: Nothing that needs fixing here, but if you just want to initialize a slice you can do one of:

* ` var presetList []kubermaticv1.Preset`: Nil slice, but can be used for append and range
* `presetList := []kubermaticv1.Preset{}`: Empty but initialized slice

I am just mentioning this here, because ppl usually only use `make` if they know the length of the slice in advance, which is not the case here

api/pkg/crd/kubermatic/v1/preset.go
```diff
@@ -0,0 +1,120 @@
...
+	Packet       *Packet       `json:"packet,omitempty"`
+	GCP          *GCP          `json:"gcp,omitempty"`
+	Kubevirt     *Kubevirt     `json:"kubevirt,omitempty"`
+
+	Fake                *Fake  `json:"fake,omitempty"`
+	RequiredEmailDomain string `json:"requiredEmailDomain,omitempty"`
+}
+
+type Digitalocean struct {
+	Token string `json:"token"` // Token is used to authenticate with the DigitalOcean API.
```
alvaroaleman: Nit: Doc is usually done above the field, not after, e.G here:
```
type Digitalocean struct {
  // Token is used to authenticate with the DigitalOcean API.
  Token string `json:"token"`
}
```

api/pkg/presets/manager_test.go
```diff
@@ -3,171 +3,504 @@ package presets_test
...
+						},
+					},
+				})
 				return manager
 			}(),
 			cloudSpec:     kubermaticv1.CloudSpec{},
 			expectedError: "can not find provider to set credentials",
 		},
+		{
+			name:       "test 12: set credentials for Kubevirt provider",
```
alvaroaleman, line 657: Can we also have one test `using credential with wrong email domain returns error`?

api/pkg/presets/manager_test.go
```diff
@@ -3,171 +3,504 @@ package presets_test
...
-
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/presets"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func TestGetPresets(t *testing.T) {
```
alvaroaleman, line 197: Please also add a test for `GetPreset` (without the trailing `s`) where you verify that the user gets an error when trying to access a preset with the wrong email domain

alvaroaleman (commented):
Two small things regarding tests, other than that LGTM.

Thanks for all your work on this!

api/pkg/presets/manager.go
zreigz: thanks

api/pkg/presets/manager_test.go
zreigz: done

api/pkg/presets/manager_test.go
zreigz: done

alvaroaleman:
/lgtm
/approve

Thanks again for all your work on this!

zreigz:
@alvaroaleman thanks for your patience and time :1st_place_medal:

PR #4263, opened 2019-09-06
Including the limit concurrent cluster updates in openshift and…

moadqassem:
In another PR(https://github.com/kubermatic/kubermatic/pull/4193), we have introduced new functionality to limit any cluster resources updates in parallel if some ongoing updates are carrying on and those updates have reached a predefined limit. With this PR we include this functionality in other controllers(openshift and monitoring controllers).

No

```release-note
NONE
```


alvaroaleman (approved):
/lgtm

PR #4264, opened 2019-09-06
Documenter of the default resource usage on a customer cluster worker node

### LINKED ISSUE #3493: Document default resource usage on a customer cluster worker node
**User Story**
We must document how much resources a minimal node must have to be able to host all default addons.

Since we set resource requests & limits for all addons, documenting it becomes mandatory.
Otherwise customers add nodes to their clusters and the only thing which runs on those nodes are the addons.

**Acceptance criteria**
- [ ] Minimum resource requirements are documented in the docs repo

themue:

Documentation of the default resource usage on a customer cluster worker node for usage in Hugo CMS.

Fixes #3493 

So far just the tool. Integration into process for Hugo is next.

**Documentation**:

```release-note
NONE
```


alvaroaleman (commented):
@themue Wouldn't it make more sense to build this in a schema-aware way? Granted, we have to parse the `yamls` as template before we can do that, but for that, we could just insert a dummy `Cluster` object into them.

themue:
Cool approach, yes. Too quick after struggled with the templates had the idea to simply to a text parse on my own. Will check your approach.

themue:
@alvaroaleman Would you mind taking another look?

api/cmd/node-resource-documenter/main.go
```diff
@@ -0,0 +1,198 @@
...
+	funcs["Registry"] = func(registry string) string {
+		return registry
+	}
+	return funcs
+}
+
+// readYAML reads a YAML file and unmarshals it.
+func readYAML(filepath string) ([]kv, error) {
+	log.Printf("Parsing YAML file %q ...", filepath)
+	f, err := os.Open(filepath)
```
alvaroaleman: What about using `ioutil.ReadFile` instead?

alvaroaleman (commented):
What I meant with schema aware the other day was to to:
* Read all files
* Generate yaml documents out of them by splitting at `---`
* Try to unmarshal them into one of:
    * DaemonSet
    * Statefulset
    * Deployment

Alternatively, but I am not completely sure if that works, you could also unmarshal them into `k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.Unstructured`, then use `GetKind()` to find out the kind and depending on the result of that, unmarshal them into the correct type.

Also, please add some unittests that demonstrate that this works and a readme.

api/cmd/node-resource-documenter/main.go
themue: Ah, sure. That's a left-over from the line-by-line scanning approach. But now I can read a file in total, yes.

themue:
> What I meant with schema aware the other day was to to:
> 
> * Read all files
> * Generate yaml documents out of them by splitting at `---`
> * Try to unmarshal them into one of:
>   
>   * DaemonSet
>   * Statefulset
>   * Deployment
> 
> Alternatively, but I am not completely sure if that works, you could also unmarshal them into `k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.Unstructured`, then use `GetKind()` to find out the kind and depending on the result of that, unmarshal them into the correct type.

Yeah, surely the most safe way because it also verifies the schema of the yaml files. It's only a bit different from the initially with Henrik discussed approach of "just grep the data and generate the docs (`cat addons/*/*.yaml | grep -A 2 limits`)". 

But I see the benefit and will change it.

api/cmd/node-resource-documenter/documenter_test.go
```diff
@@ -0,0 +1,271 @@
...
+        hostPath:
+          path: /sys
+      - name: root
+        hostPath:
+          path: /
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534`
+
+	blockZ = `apiVersion: apps/v1
```
alvaroaleman: Nit: Naming these after their content, i.E. `ServiceAccountManifest`, `DeploymentManifest`, `DefunctDeploymentManifest` instead of `blockN` would make this more readable :)

api/cmd/node-resource-documenter/documenter_test.go
```diff
@@ -0,0 +1,271 @@
...
+		t.Logf("#%d: %s", i, test.name)
+		docr := newDocumenter(test.path, []byte(test.content))
+		if test.shallFail {
+			err := docr.scanAll()
+			if err == nil || !strings.Contains(err.Error(), test.expected) {
+				t.Errorf("did not fail like expected: %v", err)
+			}
+			continue
+		}
+		err := docr.scanAll()
```
alvaroaleman, line 251: Nit: This and the block above can be consolidated:
```
err := docr.ScanAll()
If (err != nil) != test.shallFail() {
t.Fatalf("Expected error %t but got %v", test.shallFail(), err)
```

alvaroaleman (commented):
/lgtm
/approve
/hold

If you have the time you can fix the two `nit`s , else feel free to `/hold cancel` . Thanks for your work on this!

api/cmd/node-resource-documenter/documenter_test.go
themue: Thanks, yes, will also rename it from block to manifest in code.

api/cmd/node-resource-documenter/documenter_test.go
themue: As I compare the error it still looks a bit different but more clear now.

PR #4270, opened 2019-09-09
Implement zap logger in the cluster life cycle controllers

### LINKED ISSUE #3500: Implement the zap logger in the cluster-lifecycle controllers
**User Story**
We switched to structured logging using the zap logger in https://github.com/kubermatic/kubermatic/pull/3296.

So far we only implemented it in the addon controller (PR for the addoninstaller & backup controller has been created already: https://github.com/kubermatic/kubermatic/pull/3499)

This issue aims to replace all glog calls with the zap logger in all our remaining controllers:
- [x] Cluster
- [x] OpenShift
- [x] Cloud
- [x] Container-Linux
- [x] IPAM
- [x] KubeletDNAT
- [x] Monitoring
- [x] NodeCSRApprover
- [x] Update
- [x] UserCluster

As we have share code with the API & other controllers which don't use the zap logger yet, we can potentially skip updated those code places.
Plumbing in a zap logger in the API might be acceptable to avoid that.

**Acceptance criteria**
- [ ] All those controllers use a passed through logger
- [ ] None of our controllers import glog directly

moadqassem:
Using the zap logger instead of glog, in order to switch to structured logging.

Fixes #3500

Some of the controllers are still using glog, the controllers which are using zap are the controller that are mentioned in the ticket.

No 

```release-note
NONE
```


api/cmd/kubeletdnat-controller/main.go
```diff
@@ -25,19 +25,33 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	rawLog := kubermaticlog.New(logOptions.Debug, kubermaticlog.Format(logOptions.Format))
+	log := rawLog.Sugar()
 
 	_, network, err := net.ParseCIDR(*networkFlag)
 	if err != nil {
-		glog.Fatalf("node-access-network invalid or missing: %v", err)
+		log.Fatalf("node-access-network invalid or missing: %v", err)
```
xrstf: With Zap we're using the `log.Fatalw("....", zap.Error(err))` pattern to include errors in the output, IIRC. :-)

api/pkg/controller/cluster/launching.go
```diff
@@ -19,7 +17,7 @@ func (r *Reconciler) clusterIsReachable(ctx context.Context, c *kubermaticv1.Clu
 	}
 
 	if err := client.List(ctx, &ctrlruntimeclient.ListOptions{}, &corev1.NamespaceList{}); err != nil {
-		glog.V(4).Infof("Cluster %q not yet reachable: %v", c.Name, err)
+		r.log.Infow("Cluster not yet reachable", "cluster", c.Name, "error", err)
```
xrstf: IIRC V(4) is the equivalent to Zap's debug level, even though we called an `Info..` function.

api/pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -63,7 +63,7 @@ var allowedUsages = []certificatesv1beta1.KeyUsage{certificatesv1beta1.UsageDigi
 	certificatesv1beta1.UsageServerAuth}
 
 func (r *reconciler) reconcile(ctx context.Context, request reconcile.Request) error {
-	glog.V(4).Infof("Reconciling csr %q", request.NamespacedName.String())
+	r.log.Infof("Reconciling csr %q", request.NamespacedName.String())
```
xrstf: Just as a sample, but can you put the CSR name into the logger's context, so we don't have to repeat it again and again?

```
log = r.log.With("csr", request.NamespacedName.String())
log.Debug("reconciling")
```

api/pkg/controller/ipam/ipam_controller.go
```diff
@@ -9,10 +9,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/glog"
-
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
 
+	"go.uber.org/zap"
```
xrstf, line 12: Please keep non-Kubermatic 3rd party imports in between standard library and kubermatic imports, as per https://github.com/kubermatic/kubermatic/blob/master/docs/code-style.md#imports The same oops happened in some other files of this PR :-)

api/pkg/controller/usercluster/controller.go
```diff
@@ -117,12 +120,14 @@ type reconciler struct {
...
 	reconciledSuccessfullyOnce bool
+
+	log *zap.SugaredLogger
 }
 
 // Reconcile makes changes in response to objects in the user cluster.
 func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	if err := r.reconcile(context.TODO()); err != nil {
-		glog.Errorf("Reconciling failed: %v", err)
+		r.log.Errorf("Reconciling failed: %v", err)
```
xrstf: I think this log statement can go entirely. We return the error anyway, so it will be logged by someone else.

api/cmd/kubeletdnat-controller/main.go
moadqassem: Yeah after I rebased I saw the new `log.Fatalw` statements and I am refactoring them ;)

api/pkg/controller/update/update_controller.go
```diff
@@ -127,15 +130,15 @@ func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.Cluste
 }
 
 func (r *Reconciler) nodeUpdate(ctx context.Context, cluster *kubermaticv1.Cluster, clusterType string) error {
-	client, err := r.userClusterConnectionProvider.GetClient(cluster)
+	c, err := r.userClusterConnectionProvider.GetClient(cluster)
```
xrstf, line 133: /undo delete-question

May I ask why you renamed this variable?

moadqassem: Because the `client` collides with the package name. 
https://github.com/kubermatic/kubermatic/pull/4270/files/0ff1941c0dc4445409cc0c3e4e9d766be3cbf093#diff-c0591e603181d59e8b2f9c52d6443141R11

api/pkg/controller/usercluster/controller.go
alvaroaleman: Are you sure?

xrstf: All my controllers that I got recently merged certainly just return errors and only log progress/success statements.

alvaroaleman: Then you probably need to fix your controllers

api/cmd/kubeletdnat-controller/main.go
```diff
@@ -25,10 +27,13 @@ func main() {
 	networkFlag := flag.String("node-access-network", "", "The network in CIDR notation to translate to.")
 	chainNameFlag := flag.String("chain-name", "node-access-dnat", "Name of the chain in nat table.")
 	vpnInterfaceFlag := flag.String("vpn-interface", "tun0", "Name of the vpn interface.")
+	logDebugFlag := flag.Bool("log-debug", false, "Enables debug logging")
+	logFormatFlag := flag.String("log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
```
xrstf, line 31: The `log.Format` type implements the `flag.Value` interface, so you can let the flag package handle validation for you: https://github.com/kubermatic/kubermatic/blob/master/api/cmd/s3-exporter/main.go#L29

This is relatively new and IIRC we only used this in the s3-exporter for now. But maybe it's time to use it more often or even extend the `log.Options` type so that it can define the possible CLI flags itself, like

```
opt := kubermaticlog.Options{}
opt.ApplyFlags() // in a perfect world we would give the function a FlagSet instead of letting it operate on global state
```

@alvaroaleman WDYT?

api/cmd/kubeletdnat-controller/main.go
```diff
@@ -68,19 +73,19 @@ func main() {
...
 	})
 	if err != nil {
-		log.Fatalf("Failed waiting for the API server to be alive")
+		log.Fatalw("Failed waiting for the API server to be alive", zap.Error(err))
 	}
 
 	mgr, err := manager.New(config, manager.Options{})
 	if err != nil {
-		log.Fatalf("failed to create mgr: %v", err)
+		log.Fatalw("failed to create mgr: %v", zap.Error(err))
```
xrstf: No need for a `: %v` anymore, likewise 4 lines down.

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/heptiolabs/healthcheck"
 	"github.com/oklog/run"
```
xrstf: Please move the zap import one line up.

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -93,42 +95,42 @@ func main() {
...
-		log.Fatalf("Failed to parse openVPN CA file: %v", err)
+		log.Fatalw("Failed to parse openVPN CA file", zap.Error(err))
 	}
 	if certsLen := len(openVPNCACerts); certsLen != 1 {
 		log.Fatalf("Did not find exactly one but %v certificates in the openVPN CA file", certsLen)
 	}
 	openVPNCAKeyBytes, err := ioutil.ReadFile(runOp.openvpnCAKeyFilePath)
 	if err != nil {
-		log.Fatalf("Failed to read openvpn-ca-key-file: %v", err)
+		log.Fatalw("Failed to read openvpn-ca-key-file: %v", zap.Error(err))
```
xrstf: `: %v`

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -159,15 +161,15 @@ func main() {
...
 	})
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalw("Failed creating user cluster controller", zap.Error(err))
 	}
 
 	log.Info("registering components")
 	if err := apiextensionv1beta1.AddToScheme(mgr.GetScheme()); err != nil {
-		log.Fatal(err)
+		log.Fatalw("Failed to register scheme", zap.Stringer("api", apiextensionv1beta1.SchemeGroupVersion), zap.Error(err))
```
xrstf, line 168: This looks interesting. Care to explain [to me] what the Stringer does and how a log line would look once it's printed? :-)

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -184,31 +200,31 @@ func main() {
...
-			glog.Fatalf("Failed to add nodecsrapprover controller: %v", err)
+		if err := nodecsrapprover.Add(mgr, 4, cfg, log); err != nil {
+			log.Fatalw("Failed to add nodecsrapprover controller", zap.Error(err))
 		}
 		if err := openshiftmasternodelabeler.Add(context.Background(), kubermaticlog.Logger, mgr); err != nil {
-			glog.Fatalf("Failed to add openshiftmasternodelabeler controller: %v", err)
+			log.Fatalw("Failed to add openshiftmasternodelabeler controller", zap.Error(err))
 		}
-		glog.Infof("Registered nodecsrapprover controller")
+		log.Infof("Registered nodecsrapprover controller")
```
xrstf: `Info` instead of `Infof`

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -231,7 +247,7 @@ func main() {
 			// Start the Cmd
 			return mgr.Start(done)
 		}, func(err error) {
-			glog.Infof("stopping user cluster controller manager, err = %v", err)
+			log.Infof("stopping user cluster controller manager, err = %v", err)
```
xrstf: This could also use `log.Infow`, no?

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -243,13 +259,13 @@ func main() {
 			defer cancel()
 
 			if err := h.Shutdown(shutdownCtx); err != nil {
-				glog.Errorf("Healthcheck handler terminated with an error: %v", err)
+				log.Errorf("Healthcheck handler terminated with an error: %v", err)
```
xrstf: `Errorw`

api/pkg/controller/cluster/launching.go
```diff
@@ -19,7 +19,7 @@ func (r *Reconciler) clusterIsReachable(ctx context.Context, c *kubermaticv1.Clu
 	}
 
 	if err := client.List(ctx, &ctrlruntimeclient.ListOptions{}, &corev1.NamespaceList{}); err != nil {
-		glog.V(4).Infof("Cluster %q not yet reachable: %v", c.Name, err)
+		r.log.Debugf("Cluster %q not yet reachable", c.Name, zap.Error(err))
```
xrstf: You mixed formatting-based and structured approaches here. What about `r.log.Debugw("cluster not yet reachable", "cluster", c.Name, zap.Error(err))`? Same for the code in `pending.go`.

api/pkg/controller/ipam/ipam_controller.go
```diff
@@ -88,7 +90,7 @@ func (r *reconciler) reconcile(ctx context.Context, machine *clusterv1alpha1.Mac
 	}
 
 	if !strings.Contains(machine.Annotations[annotationMachineUninitialized], annotationValue) {
-		glog.V(4).Infof("Machine %s doesn't need initialization", machine.Name)
+		r.log.Debugf("Machine %s doesn't need initialization", machine.Name)
```
xrstf: `r.log.Debugw(....)`

api/cmd/kubeletdnat-controller/main.go
moadqassem: We actually use this approach in other controllers and use this struct to initialize and validate the input(which is actually part of the package so we are not writing a new custom validator).    

https://github.com/kubermatic/kubermatic/blob/master/api/cmd/kubermatic-controller-manager/options.go#L220-L222

The two fields (log.Debug and log.Format) belong to the same struct and I am using this struct explicitly to initialize the logger and it's a natural thing to validate those values as an extra check.

api/pkg/controller/kubeletdnat/kubeletdnat_controller.go
```diff
@@ -11,10 +11,10 @@ import (
 	"strings"
 
 	"github.com/go-test/deep"
-	"github.com/golang/glog"
-
 	"github.com/kubermatic/kubermatic/api/pkg/provider"
 
+	"go.uber.org/zap"
```
xrstf, line 15: Please group imports.

api/pkg/controller/kubeletdnat/kubeletdnat_controller.go
```diff
@@ -115,7 +119,7 @@ func (r *Reconciler) getDesiredRules(nodes []corev1.Node) []string {
 	for _, node := range nodes {
 		nodeRules, err := r.getRulesForNode(node)
 		if err != nil {
-			glog.Errorf("could not generate rules for node %s: %v - skipping", node.Name, err)
+			r.log.Errorf("could not generate rules for node %s: %v - skipping", node.Name, err)
```
xrstf: `r.log.Errorw("could not...", "node", node.Name, zap.Error(err))` please :-)

api/cmd/user-cluster-controller-manager/main.go
moadqassem: ??

api/cmd/kubeletdnat-controller/main.go
moadqassem: This is strange, I actually got rid of all the `%v` but I guess something went wrong during rebasing 🤔

api/cmd/user-cluster-controller-manager/main.go
moadqassem: The Stringer a field with the given key and the output of the value's String() method. The output would be similar to any `key, value` print similar to ` "api": "123"  ` however instead of calling the method `String()` explicitly, you use the Stringer which calls the String() method implicitly. And since the type `SchemeGroupVersion` implements the method thing will work out of the box. 
Here are some examples of live code:
https://github.com/kubermatic/kubermatic/blob/master/api/cmd/kubermatic-controller-manager/main.go#L80-L88

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -66,59 +70,70 @@ func main() {
...
+		log.Fatalw("Failed to read openvpn-ca-cert-file", zap.Error(err))
 	}
 	openVPNCACerts, err := certutil.ParseCertsPEM(openVPNCACertBytes)
 	if err != nil {
-		glog.Fatalf("Failed to parse openVPN CA file: %v", err)
+		log.Fatalw("Failed to parse openVPN CA file", zap.Error(err))
 	}
 	if certsLen := len(openVPNCACerts); certsLen != 1 {
-		glog.Fatalf("Did not find exactly one but %v certificates in the openVPN CA file", certsLen)
+		log.Fatalw("Did not find exactly one but certificate in the openVPN CA file", "certificates-count", certsLen)
```
xrstf: Maybe remove the `but` so it reads nicer.

api/pkg/controller/cluster/pending.go
```diff
@@ -63,7 +62,7 @@ func (r *Reconciler) reconcileCluster(ctx context.Context, cluster *kubermaticv1
 	}
 
 	if !cluster.Status.ExtendedHealth.AllHealthy() {
-		glog.V(4).Infof("Cluster %q not yet healthy: %+v", cluster.Name, cluster.Status.ExtendedHealth)
+		r.log.Debugf("Cluster %q not yet healthy: %+v", cluster.Name, cluster.Status.ExtendedHealth)
```
xrstf, line 65: Forgot to apply Debugw here? The `launching.go` got it right :-)

moadqassem: Not really I am leaving the formating because of the use of `%+v` formatter

api/pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -73,30 +75,30 @@ func (r *reconciler) reconcile(ctx context.Context, request reconcile.Request) e
...
 		if condition.Type == certificatesv1beta1.CertificateApproved {
-			glog.V(4).Infof("CSR %q already approved, skipping", csr.Name)
+			log.Debug("already approved, skipping reconciling")
 			return nil
 		}
 	}
 
 	if !sets.NewString(csr.Spec.Groups...).Has("system:nodes") {
-		glog.V(4).Infof("Skipping CSR %q because it 'system:nodes' is not in its groups", csr.Name)
+		log.Debug("Skipping reconciling because its 'system:nodes' is not in its groups")
```
xrstf: You fixed the typo, but maybe `because 'system:nodes' is not in its groups` reads even better?

api/pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -107,7 +109,7 @@ func (r *reconciler) reconcile(ctx context.Context, request reconcile.Request) e
 		return fmt.Errorf("failed to update approval for CSR %q: %v", csr.Name, err)
 	}
 
-	glog.V(2).Infof("Successfully approved CSR %q", csr.Name)
+	r.log.Infof("Successfully approved")
```
xrstf: Shouldn't this be `log` instead of `r.log`?

api/pkg/controller/cluster/pending.go
xrstf: How about `log.Debugf("Cluster not healthy: %+v", cluster.Status.ExtendedHealth)`?

xrstf: Nevermind, I got confused with another file.

moadqassem: This is a debug message, listing the cluster that has a failing health condition within the message can be helpful. The cluster name was in the original message so the person who got it in there might have had a scenario where he/she left there for a reason.

PR #4272, opened 2019-09-09
increase Prometheus and Thanos resources

xrstf:
We have adjusted Prometheus and Thanos resources pretty much in every deployment. It makes sense to increase the defaults to fit a realistic Kubermatic setup. Small POCs use hopefully machines large enough to hold the monitoring stack when/if customers decide to try out the stack.

```release-note
NONE
```


xmudrii (approved):
/approve
/hold
It looks reasonable to me but holding if there's anyone else who should take a look. If not, just remove the hold

PR #4295, opened 2019-09-12
extend seed-proxy to cover multiple monitoring services

xrstf:
We want to proxy not just Prometheus, but also the Alertmanager. This PR therefore extends the seed-proxy controller to not manage a proxy-pod just for Prometheus, but instead use one proxy pod for all services inside a given namespace (the monitoring namespace in this case).
This PR also fixes the outdated naming scheme for services. `prometheus-kubermatic` is now just `prometheus`.

```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #4299, opened 2019-09-12
add Karma dashboard Chart

xrstf:
This PR adds a Helm Chart for https://github.com/prymitive/karma to our monitoring stack. It's meant to be deployed into master clusters and then aggregate all the alerts from the various seeds by simply using the seed-proxy pods.

This solves not only the visibility issue, but it can also be used to *silence* alerts inside the seeds. And since it's only a dashboard it's not a single point of failure (if the dashboard dies, the Alertmanagers will continue to work).

The very simple init container approach will not handle seeds being added or deleted, so admins will have to restart the Karma pod when they modify seeds. For now this is IMHO acceptable until we have more experience with this and can build something more sophisticated.

```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #4307, opened 2019-09-17
Add leader election to master controller manager

xrstf:
This PR adds leader election to the master controller manager, because the regular controller manager does the same. It also renames `sugarLog` to our more common `log` variable and removes one seemingly pointless variable (`done := ctx.Done()`).

```release-note
master-controller-manager can now be deployed with multiple replicas
```

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
+	"github.com/kubermatic/kubermatic/api/pkg/leaderelection"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/metrics"
 	metricserver "github.com/kubermatic/kubermatic/api/pkg/metrics/server"
@@ -19,9 +20,16 @@ import (
 	seedvalidation "github.com/kubermatic/kubermatic/api/pkg/validation/seed"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	kubeleaderelection "k8s.io/client-go/tools/leaderelection"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+const (
+	controllerName = "kubermatic-master-controller-manager"
 )
 
 type controllerRunOptions struct {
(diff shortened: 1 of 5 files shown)


api/cmd/master-controller-manager/main.go
```diff
@@ -64,35 +72,37 @@ func main() {
 	flag.StringVar(&runOpts.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
 	flag.Parse()
 
-	log.SetLogger(log.ZapLogger(false))
```
xrstf, line 67: I would like to put this entire stanza to setup the logging (up until the `kubermaticlog.Logger = ...` line) into a util somewhere. In the `api/pkg/log` package or in a real utility at `api/pkg/controller/util`?

alvaroaleman: If that was possible Henrik would've done it, the problem is we need the `defer` statement in the `main`

api/cmd/master-controller-manager/main.go
```diff
@@ -113,65 +123,97 @@ func main() {
...
 
-	// This group is running all controllers
+	// This group is running the actual controller logic
 	{
+		leaderCtx, stopLeaderElection := context.WithCancel(ctx)
+		defer stopLeaderElection()
+
 		g.Add(func() error {
-			return mgr.Start(ctx.Done())
+			leaderElectionClient, err := kubernetes.NewForConfig(rest.AddUserAgent(cfg, "kubermatic-master-controller-manager-leader-election"))
```
alvaroaleman: Would it make sense to write an abstraction for all of this that just takes a restconfig, a `func(context.Context)` and the name of the lock?

api/cmd/master-controller-manager/main.go
xrstf: SadFace

api/cmd/master-controller-manager/main.go
xrstf: Created `leaderelection.RunAsLeader()` helper as discussed and used it in the regular controller-manager.

xrstf:
@alvaroaleman WDYT about changing

```go
OnStartedLeading: func(ctx context.Context) {
	log.Info("acquired the leader lease")
	callback(ctx)
},
```

to allow `callback` to return an error. In this case the helper could already cancel the context and writing callbacks becomes easier:

```go
OnStartedLeading: func(ctx context.Context) {
	log.Info("acquired the leader lease")
	if err := callback(ctx); err != nil {
		log.Error(zap.Error(err)) // not 100% sure this would work
		cancel()
	}
}
```

and a callback could be reduced from

```go
func(ctx context.Context) {
   log.Info("Executing migrations...")
   if err := migrations.RunAll(ctrlCtx.mgr.GetConfig(), options.workerName, runOpts.dcFile); err != nil {
      log.Errorw("failed to run migrations", zap.Error(err))
      stopLeaderElection()
      return
   }
   log.Info("Migrations executed successfully")

   log.Info("starting the master-controller-manager...")
   if err := mgr.Start(ctx.Done()); err != nil {
      log.Errorw("the controller-manager stopped with an error", zap.Error(err))
      stopLeaderElection()
   }
}
```

to

```go
func(ctx context.Context) error {
   log.Info("Executing migrations...")
   if err := migrations.RunAll(ctrlCtx.mgr.GetConfig(), options.workerName, runOpts.dcFile); err != nil {
      return fmt.Errorf("failed to run migrations: %v", err)
   }
   log.Info("Migrations executed successfully")

   log.Info("starting the master-controller-manager...")
   if err := mgr.Start(ctx.Done()); err != nil {
      return fmt.Errorf("the controller-manager stopped with an error: %v", err)
   }

   return nil
}
```

api/pkg/leaderelection/leaderelection.go
```diff
@@ -48,3 +53,36 @@ func New(name string, leaderElectionClient kubernetes.Interface, recorder record
...
+	}
+
+	leaderElectionCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	callbacks := kubeleaderelection.LeaderCallbacks{
+		OnStartedLeading: func(ctx context.Context) {
+			log.Info("acquired the leader lease")
+			if err := callback(ctx); err != nil {
+				log.Error(zap.Error(err))
```
alvaroaleman: Nit: Put the `leaderName` as field into the log? :)

alvaroaleman (approved):
/lgtm
/approve

xrstf:
Sorry had to push another commit because the error printing code didn't do what I hoped it would do.

alvaroaleman (commented):
/lgtm

Issue #4171, opened 2019-08-26
Deploy openshift 4 console


alvaroaleman:
**User Story**

As a user of openshift four I want to use its console

**Acceptance criteria**

* The openshift four console gets deployed
* There is an API endpoint in the Kubermatic API over which it can be reached

PR #4309, opened 2019-09-17
Deploy the openshift console

alvaroaleman:

With the changes in this PR, one can visit the openshift console and login by adding `console.openshift.seed.tld` as alias for `localhost` to their hosts file and port-fowarding port 8443 to an openshift console container. Login then works by using the data in the `openshift-bootstrap-password` secret and `kubeadmin` as user.

Follow-Ups will be to put this behind our API and make the password usable for normal users that do not have access to the seed.

Fixes #

**Documentation**:

```release-note
none
```


api/pkg/controller/openshift/resources/console.go
```diff
@@ -0,0 +1,316 @@
...
+			// Create oauthClient object in the usercluster first, as it can not be reset otherwise
+			// because end-users do not have access to the seed
+			name := types.NamespacedName{Name: consoleOauthClientObjectName}
+			if err := client.Get(context.Background(), name, oauthClientObject); err != nil {
+				if !kerrors.IsNotFound(err) {
+					return nil, fmt.Errorf("failed to get OauthClient %q from usercluster: %v", consoleOauthClientObjectName, err)
+				}
+				secret, err := generateNewSecret()
+				if err != nil {
+					return nil, fmt.Errorf("failed to generate oauth client secret: %v", err)
```
xrstf: There are `OauthClient`, `OAuthClient` and `oauth client` in this function alone. If we want to capitalize kind names, please use `OAuthClient` consistently :-)

api/pkg/controller/openshift/resources/console.go
```diff
@@ -0,0 +1,81 @@
...
+				}
+				oauthClientObject.Object["grantMethod"] = "auto"
+				oauthClientObject.SetName(consoleOauthClientObjectName)
+				if err := client.Create(context.Background(), oauthClientObject); err != nil {
+					return nil, fmt.Errorf("failed to create OauthClient object in user cluster: %v", err)
+				}
+			}
+
+			stringVal, ok := oauthClientObject.Object["secret"].(string)
+			if !ok {
```
xrstf, line 243: I don't think it's likely that it's not a string (though I don't really but), *but* I love the safety check anyway 👍

api/pkg/controller/openshift/resources/kube_apiserver_deployment.go
```diff
@@ -33,9 +33,52 @@ var (
...
+	apiServerOauthMetadataConfigMapName = "openshift-oauth-metadata"
+	apiServerOauthMetadataConfigMapKey  = "oauthMetadata"
 )
 
+func APIServerOauthMetadataConfigMapCreator(data openshiftData) reconciling.NamedConfigMapCreatorGetter {
+	return func() (string, reconciling.ConfigMapCreator) {
+		return apiServerOauthMetadataConfigMapName, func(cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+			oauthPort, err := data.GetOauthExternalNodePort()
+			if err != nil {
+				return nil, fmt.Errorf("failed to get exeternal port for oauth service: %v", err)
```
xrstf: `exeternal` should be `external`.

api/pkg/controller/openshift/resources/kube_apiserver_deployment.go
```diff
@@ -33,9 +33,52 @@ var (
...
+			}
+			oauthAddress := fmt.Sprintf("https://%s:%d", data.Cluster().Address.ExternalName, oauthPort)
+			if cm.Data == nil {
+				cm.Data = map[string]string{}
+			}
+			cm.Data["oauthMetadata"] = fmt.Sprintf(`{
```
xrstf: This is how we assembled XML back in the PHP4 days. :trollface: I assume we will improve this at some point, because we have been bitten by the "printf-style YAML/JSON construction" before.

api/pkg/controller/openshift/openshift_controller.go
```diff
@@ -348,6 +348,13 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 		}
 	}
 
+	// We must put the hash for this into the usercluster and the raw value to expose in the
+	// seed. The usercluster secret must be created max 1h after crreation timestamp of the
```
xrstf: `crreation` -> `creation`

api/pkg/controller/openshift/resources/kube_apiserver_deployment.go
alvaroaleman: Not better or worse than all the templated yaml in this package ;) The whole thing is a bit tricky, because we can not simply import the openshift api as we will need different api versions for different openshfit versions

alvaroaleman:
Tracking ticket for the feature is btw #4171 

@xrstf PTAL

PR #4311, opened 2019-09-17
chart-sync: fix listing getting latest release tag

kdomanski:
We've agreed to use `ls-remote` instead of `for-each-ref` to avoid cloning a whole repo, however it turns out that the former does not work like the latter.

We cannot sort by `authordate`, since it is contained within the object and `ls-remote` cannot know it.
The `--count` parameter is unsupported entirely.

This PR switches to even better sorting, by `version:refname` which is refname parsed as semver. It also switches to using `head` command instead of the unsupported `--count` parameter.

```release-note
NONE
```

### DIFF
--- a/api/hack/lib.sh
+++ b/api/hack/lib.sh
@@ -69,7 +69,7 @@ get_latest_dashboard_tag() {
   local DASHBOARD_URL="git@github.com:kubermatic/dashboard-v2.git"
 
   MINOR_VERSION="${FOR_BRANCH##release/}"
-  FOUND_TAG="$(retry 5 git ls-remote "$DASHBOARD_URL" "refs/tags/$MINOR_VERSION*" --sort=-authordate --count=1 | awk '{print $2}')"
+  FOUND_TAG="$(retry 5 git ls-remote --sort=-version:refname "$DASHBOARD_URL" "refs/tags/$MINOR_VERSION*" --count=1 | awk '{print $2}')"
   if [ -z "$FOUND_TAG" ]; then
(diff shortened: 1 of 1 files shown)


api/hack/lib.sh
```diff
@@ -69,7 +69,7 @@ get_latest_dashboard_tag() {
   local DASHBOARD_URL="git@github.com:kubermatic/dashboard-v2.git"
 
   MINOR_VERSION="${FOR_BRANCH##release/}"
-  FOUND_TAG="$(retry 5 git ls-remote "$DASHBOARD_URL" "refs/tags/$MINOR_VERSION*" --sort=-authordate --count=1 | awk '{print $2}')"
+  FOUND_TAG="$(retry 5 git ls-remote --sort=-version:refname "$DASHBOARD_URL" "refs/tags/$MINOR_VERSION*" | head -n1 | awk '{print $2}')"
```
alvaroaleman, line 72: How is this supposed to work? `refs/tags/X` are individual tags. Could you show an example on how you are getting from the compeltely different named branches in Kubermatic to the tags here?

kdomanski: `$MINOR_VERSION` is build by parsing the branch name and is already in the form of e.g. `v2.12`

alvaroaleman:
/lgtm
/approve

However I think we must not backport it to 2.11, as the tag names are different there hence it doesn't work there, e.G.  `git ls-remote --sort=-version:refname git@github.com:kubermatic/dashboard-v2.git refs/tags/v2.11*` yields nothing

PR #4316, opened 2019-09-18
Add proposal for user cluster RBACs

### LINKED ISSUE #4302: Proposal for user cluster RBAC
**User Story**
Prepare a proposal for user cluster RBAC

zreigz:
**What this PR does / why we need it**: Proposal for user cluster RBACs

Fixes #4302

```release-note
NONE
```

### DIFF
--- a/docs/proposals/user-cluster-rbac.md
+++ b/docs/proposals/user-cluster-rbac.md
@@ -1,0 +1,58 @@
+# User Cluster RBACs
+**Author**: Lukasz Zajaczkowski(@zreigz)
+
+**Status**: Draft proposal.
+
+## Introduction
+Share cluster feature allows to share access to the user cluster with other users.
+The owner of the cluster must grant other user some permissions. To do so, configure `kubectl` to point to the cluster and
+create a `rolebinding` or `clusterrolebinding`, using the email address of the user the kubeconfig was shared.
+We would like simplify this process.
+
+## Core concept
+The cluster owner will be able to create roles which could restrict the user from accessing specific resources.
+For example, a role can be created which allows the user to view and update clusters but does not allow deleting them.
+Roles can be created in cluster scope (impacts all namespaces) or namespace scope (impacts a specific namespace in which it is created).
+In order to link the user to a role, the cluster owner has to create a binding.
+We expose API endpoints for all the operation which are needed for managing Role/ClusterRole and bindings.
+End-user will get UI interface to interact with RBACs on the cluster detail page.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/user-cluster-rbac.md
```diff
@@ -0,0 +1,114 @@
+# User management
```
alvaroaleman: @scheeles suggested that we do not have a CRD at all, but instead just build a set of APIs that directly talk to the usercluster only:

* An endpoint that lists all (cluster-)roles that have an associated binding for a User (as opposed to ServiceAccount) and the names of all associated users
* Endpoints to Create/Update/Delete (cluster-) roles
* An endpoint to Create/Update/Delete (cluster-) role bindings for users

WDYT?

zreigz: It's not a problem. Sounds more clear and simple for me. More things must be done on the API side and we have to design nice and shiny UI for it.
I have only one remark. I am not sure if the customer will be willing to create every time the same roles for each cluster. Maybe it would be good to create `roles` on the project level and create bindings on the cluster page. I am open to any suggestions.

alvaroaleman: > I have only one remark. I am not sure if the customer will be willing to create every time the same roles for each cluster. Maybe it would be good to create roles on the project level and create bindings on the cluster page. I am open to any suggestions.

That is one option, another one would be to use the addons controller to make sure there is a set of roles that exists in _all_ clusters

zreigz: For sure using only API calls in cluster scope is really simple and performs atomic operations. We don't need to create a separate controller and new CRD. This is the biggest advantage. If it OK for the Customer to specify roles for each cluster we can go this direction.

alvaroaleman: >  If it OK for the Customer to specify roles for each cluster we can go this direction.

Well, we can probably provide some default roles via our controllers or an addon. If the customer wants more, they can always set up an additional addon for that.

In a far away future we could even build API and UI to create a RBAC addon and e.G. assign that to all clusters in a project

zreigz: OK, thanks for the explanation. I am starting with new API endpoints. In parallel, I will change this proposal for the new approach.

alvaroaleman:
Abusing this PR to test something, sorry :grimacing: 
/test pull-kubermatic-kubermatic-validate-prow-yaml

PR #4338, opened 2019-09-20
Update cert-manager to 0.10

xrstf:
According to the documentation there are no breaking changes and doing a test deployment of 0.9 and then upgrading to 0.10 worked without a hitch.

```release-note
update cert-manager to 0.10.0
```


config/cert-manager/templates/webhook-apiservice.yaml
```diff
@@ -1,15 +1,15 @@
...
 metadata:
-  name: v1beta1.admission.certmanager.k8s.io
+  name: v1beta1.webhook.certmanager.k8s.io
   labels:
     app: webhook
     app.kubernetes.io/name: webhook
     app.kubernetes.io/instance:  {{ .Release.Name }}
   annotations:
-    certmanager.k8s.io/inject-ca-from: '{{ .Release.Namespace }}/webhook-tls'
+    certmanager.k8s.io/inject-ca-from-secret: '{{ .Release.Namespace }}/webhook-tls'
```
kron4eg, line 10: is this place we once had a problem with?

config/cert-manager/templates/webhook-mutating-webhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+{{- if .Values.certManager.injectAPIServerCA }}
+    certmanager.k8s.io/inject-apiserver-ca: 'true'
+{{- end }}
+webhooks:
+  - name: webhook.certmanager.k8s.io
+    rules:
+      - apiGroups:
+          - "certmanager.k8s.io"
+        apiVersions:
+          - v1alpha1
```
kron4eg, line 19: is it really `v1alpha1` but not `v1beta1`? I thought CM moving away from v1alpha1

config/cert-manager/templates/webhook-apiservice.yaml
xrstf: Yes because this change was in 0.10 and not in 0.9, but at the time of 0.9 I looked art the master branch by accident.

https://github.com/jetstack/cert-manager/blob/v0.10.0/deploy/charts/cert-manager/templates/webhook-apiservice.yaml#L13

config/cert-manager/templates/webhook-mutating-webhook.yaml
xrstf: Apparently it's still v1alpha1: https://github.com/jetstack/cert-manager/blob/v0.10.0/deploy/charts/cert-manager/templates/webhook-mutating-webhook.yaml#L22

PR #4344, opened 2019-09-20
fix json name for securityGroupId

### LINKED ISSUE #4337: AWS: custom security group is not applied
I've been trying to create an AWS cluster with a specific security group assigned. (An "allow all" group is needed for the extended conformance-certification test suite.)
The cluster didn't end up with my custom security group assigned, but instead had a new group created as usual.

image

According to the chrome dev console, the request sent to the API is correct and includes the SG ID I used. However, the created cluster doesn't have that group assigned.
I've also tried filling out all of the advanced settings instead, in assumption that they are silently dropped if only some are provided. The results was all of these custom values applied, except for the security group ID.

**Environment**
- UI Version: kubermatic/dashboard-v2@40538eae00da20201d1600ab216db3ecc418aef9
- API Version: 0aead8820fa30fbe4377f2a06aad9883a420d195
- Domain: cloud

zreigz:
**What this PR does / why we need it**: fix json name for securityGroupId

Fixes #4337 

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -4873,7 +4873,7 @@
           "type": "string",
           "x-go-name": "SecretAccessKey"
         },
-        "securityGroupID": {
+        "securityGroupId": {
           "type": "string",
           "x-go-name": "SecurityGroupID"
         },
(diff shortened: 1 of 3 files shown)


zreigz:
@kgroschoff can you fix Dashboard and change the json name from `securityGroup` to `securityGroupId`. Right now setting this flag doesn't work for AWS

PR #4348, opened 2019-09-21
E2E tests: Provide more granular feedback during cluster creation

alvaroaleman:

Provides more granular feedback when pods inside usercluster do not get ready in time by splitting it into three distinct checks:
* Do machines result in a node joining the cluster
* Do all nodes get ready
* Do all pods inside the cluster get ready

It is also the better test overall, as the current test succeeds as soon as one node gets ready and is able to host the openVPN pod. If the two other machines never joined the cluster, we wouldn't notice that.

Fixes #

**Documentation**:

```release-note
none
```


alvaroaleman:
/assign @xrstf 

Ironically openshift tests pass instantly now

PR #4350, opened 2019-09-21
Revert "fix json name for securityGroupId (#4344)"

alvaroaleman:
This reverts commit 7a264ae13cfaf964edb979c612a95662f841937f.

This change caused us to drop the existing `securityGroupID` field on all AWS
clusters because the tag got renamed, causing us to not recognize the existing field.

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/cmd/kubermatic-api/swagger.json
+++ b/api/cmd/kubermatic-api/swagger.json
@@ -4873,7 +4873,7 @@
           "type": "string",
           "x-go-name": "SecretAccessKey"
         },
-        "securityGroupId": {
+        "securityGroupID": {
           "type": "string",
           "x-go-name": "SecurityGroupID"
         },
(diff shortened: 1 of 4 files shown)


alvaroaleman:
/assign @zreigz @kgroschoff @kdomanski 
CC @thz , I hope this didn't rolled out anywhere other than on cloud, where this change results us in now not being able to delete any AWS cluster, as we are unable to cleanup the security group as we don't know the name anymore.

Please _never rename json tags_ on the Kubermatic cluster CRD (or any CRD for that matter). This causes us to lose information.

Doing this on the API types is okay because those do not get persistet. Its a completely different story for the CRDs, as they are used to persist data.

If you must rename a field for whatever reason, please add a new field and a migration that copies the value of the old field to the new one. You may then remove the old field in the Kubermatic release after the Kubermatic release with the migration.

alvaroaleman:
Also as a general rule of thumb: We do not do incompatible API changes just because the UI does not use the correct name for a field. Changes like this break every client generated from our swagger spec for no good reason.

zreigz:
Good point, sorry for that. We will change only the UI part and it will work.

PR #4360, opened 2019-09-23
Datacenters-to-Seeds Migration

### LINKED ISSUE #2116: Add migration step for `datacenters.yaml` -> CRD
**User Story**
To make the switch from `datacenters.yaml` -> CRD's as easy as possible, a migration step should be implemented which will convert the passed `datacenters.yaml` to the relevant CRD (SeedDatacenter/NodeLocation).
The migration step should be part of our existing migration cmd: https://github.com/kubermatic/kubermatic/blob/master/api/cmd/kubermatic-controller-manager-cleanup/main.go

Flow:
- Check if `datacenters.yaml` has been passed via flags
  - If not -> Skip the migration step (If its a new system, it probably has no `datacenters.yaml`)
- Parse `datacenters.yaml`
- Iterate over `datacenters.yaml->datacenters`
  - Take the key as `name`
  - Determine if it is a SeedDatacenter or NodeLocation (Via `is_seed: true`)
  - Check if the CRD already exists
    - Yes: Skip (Do not make any modifications)
    - No: Create a new CRD based on the given data

**Acceptance criteria**
- [ ] A migration step has been added to https://github.com/kubermatic/kubermatic/blob/master/api/cmd/kubermatic-controller-manager-cleanup/main.go which implements the above mentioned flow.
- [ ] A unit test exists which covers the basic logic of the above mentioned flow

xrstf:
This implements a basic migration for datacenters. For safety reasons the migration will not do anything if a Seed CR already exists.

Fixes #2116

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -11,7 +11,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
-	"github.com/kubermatic/kubermatic/api/pkg/crd/migrations/master"
+	mastermigrations "github.com/kubermatic/kubermatic/api/pkg/crd/migrations/master"
 	"github.com/kubermatic/kubermatic/api/pkg/leaderelection"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/metrics"
@@ -64,14 +64,14 @@ func main() {
 	runOpts := controllerRunOptions{}
 	runOpts.seedvalidationHook.AddFlags(flag.CommandLine)
 	flag.StringVar(&runOpts.kubeconfig, "kubeconfig", "", "Path to a kubeconfig.")
-	flag.StringVar(&runOpts.dcFile, "datacenters", "", "The datacenters.yaml file path")
+	flag.StringVar(&runOpts.dcFile, "datacenters", "", "The datacenters.yaml file path.")
 	flag.StringVar(&runOpts.masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&runOpts.workerName, "worker-name", "", "The name of the worker that will only processes resources with label=worker-name.")
 	flag.IntVar(&ctrlCtx.workerCount, "worker-count", 4, "Number of workers which process the clusters in parallel.")
-	flag.StringVar(&runOpts.internalAddr, "internal-address", "127.0.0.1:8085", "The address on which the /metrics endpoint will be served")
-	flag.BoolVar(&runOpts.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters. Enabling this and defining the datcenters flag will enable the seed migration")
-	flag.StringVar(&ctrlCtx.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources")
-	flag.BoolVar(&runOpts.log.Debug, "log-debug", false, "Enables debug logging")
+	flag.StringVar(&runOpts.internalAddr, "internal-address", "127.0.0.1:8085", "The address on which the /metrics endpoint will be served.")
+	flag.BoolVar(&runOpts.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters. Enabling this and defining the datcenters flag will enable the migration of the datacenters defined in datancenters.yaml to Seed custom resources.")
+	flag.StringVar(&ctrlCtx.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources.")
+	flag.BoolVar(&runOpts.log.Debug, "log-debug", false, "Enables debug logging.")
 	flag.StringVar(&runOpts.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
 	flag.Parse()
 
@@ -88,7 +88,7 @@ func main() {
 
 	selector, err := workerlabel.LabelSelector(runOpts.workerName)
 	if err != nil {
-		log.Fatalw("Failed to create the label selector for the given worker", "workerName", runOpts.workerName, zap.Error(err))
+		log.Fatalw("failed to create the label selector for the given worker", "workerName", runOpts.workerName, zap.Error(err))
 	}
 
 	// register the global error metric. Ensures that runtime.HandleError() increases the error metric
@@ -103,7 +103,7 @@ func main() {
 	ctrlCtx.ctx = ctx
 
 	// prepare migration options
-	migrationOptions := master.MigrationOptions{
+	migrationOptions := mastermigrations.MigrationOptions{
 		DatacentersFile:    runOpts.dcFile,
 		DynamicDatacenters: runOpts.dynamicDatacenters,
 	}
@@ -111,7 +111,7 @@ func main() {
 	// load kubeconfig and create API client
 	kubeconfig, err := clientcmd.LoadFromFile(runOpts.kubeconfig)
 	if err != nil {
-		log.Fatalw("Failed to read the kubeconfig", zap.Error(err))
+		log.Fatalw("failed to read the kubeconfig", zap.Error(err))
 	}
 
 	config := clientcmd.NewNonInteractiveClientConfig(
@@ -123,7 +123,7 @@ func main() {
 
 	cfg, err := config.ClientConfig()
 	if err != nil {
-		log.Fatalw("Failed to create client", zap.Error(err))
+		log.Fatalw("failed to create client", zap.Error(err))
 	}
 
 	ctrlCtx.labelSelectorFunc = func(listOpts *metav1.ListOptions) {
@@ -215,11 +215,11 @@ func main() {
 
 				// wait for the webhook to be ready
 				if migrationOptions.SeedMigrationEnabled() {
-					deadline := 30 * time.Second
+					timeout := 30 * time.Second
 					endpoint := types.NamespacedName{Namespace: ctrlCtx.namespace, Name: "seed-webhook"}
 
-					log.Infow("waiting for webhook to be ready...", "webhook", endpoint, "deadline", deadline)
-					if err := wait.Poll(500*time.Millisecond, deadline, func() (bool, error) {
+					log.Infow("waiting for webhook to be ready...", "webhook", endpoint, "timeout", timeout)
+					if err := wait.Poll(500*time.Millisecond, timeout, func() (bool, error) {
 						endpoints := &corev1.Endpoints{}
 						if err := client.Get(ctx, endpoint, endpoints); err != nil {
 							return false, err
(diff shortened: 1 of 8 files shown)


api/pkg/crd/migrations/master/migrations.go
```diff
@@ -0,0 +1,85 @@
...
+		log:                 log,
+		client:              client,
+		workerName:          workerName,
+		kubermaticNamespace: kubermaticNamespace,
+	}
+
+	if opt.DatacentersFile != "" && opt.DynamicDatacenters {
+		log.Info("datacenters given and dynamic datacenters enabled, attempting to migrate datacenters to Seeds")
+		if err := migrateDatacenters(migrationCtx, opt.DatacentersFile); err != nil {
+			return fmt.Errorf("failed to migrate datacenters.yaml: %v", err)
```
alvaroaleman, line 51: Can we also add a success log message here?

api/pkg/crd/migrations/master/migrations.go
```diff
@@ -0,0 +1,85 @@
...
+	existingSeeds := kubermaticv1.SeedList{}
+	if err := ctx.client.List(ctx.ctx, &ctrlruntimeclient.ListOptions{
+		Namespace:     ctx.kubermaticNamespace,
+		LabelSelector: labelSelector,
+	}, &existingSeeds); err != nil {
+		return fmt.Errorf("failed to list existing Seeds in %s: %v", ctx.kubermaticNamespace, err)
+	}
+
+	if len(existingSeeds.Items) > 0 {
+		ctx.log.Warn("migration enabled, but existing Seed CRs found; refusing to migrate again")
```
alvaroaleman: Maybe return that as an error?

api/pkg/crd/migrations/master/migrations.go
```diff
@@ -0,0 +1,85 @@
...
+	seeds, err := provider.LoadSeeds(dcFile)
+	if err != nil {
+		return fmt.Errorf("failed to load %s: %v", dcFile, err)
+	}
+
+	for name, seed := range seeds {
+		seed.Namespace = ctx.kubermaticNamespace
+		if err := ctx.client.Create(ctx.ctx, seed); err != nil {
+			return fmt.Errorf("failed to create seed %s: %v", name, seed)
+		}
```
alvaroaleman, line 165: Please also add a success log message for each seed creation, so that if this breaks, ppl know what was done already

api/pkg/crd/migrations/master/migrations.go
xrstf: My intention was that this is a "harmless" error (especially given our clusters, where controllers randomly restart). Returning an error would make the master controller crashloop, right?

alvaroaleman: I was just wondering what we do in case this breaks in the middle because the webhook will prevent us from deleting the seed crds as there is likely a cluster using them.

xrstf: Stupid idea, just so only the good ideas are left: Can we relax the webhook when the "migration mode" is enabled?

alvaroaleman: Generally this doesn't feel like a situation  where a good solution exists :thinking: 

How exactly would you relax the webhook, what should it allow and from whom?

Another idea: Add another flag `--force-migration` that makes us skip this check? Should rarely if ever needed, only if the migration broke.

xrstf: > How exactly would you relax the webhook, what should it allow and from whom?

While migration is enabled (both CLI flags defined), it's allowed to delete seeds. This exception would be made just for this very case (that a migration went kaputt). We should be able to verify that a seed that is being deleted still exists inside the datacenters.yaml.

xrstf: I've pushed a commit to disable modifying seeds while the migration is enabled. We now just need to figure out how to make the webhook available so the migration can actually create seeds.

xrstf: I've pushed another commit that puts the validation webhook into its own run group and makes the leader wait for the webhook to be ready before running the migrations.

The migrations now carefully create missing seeds, as that's the only option that is allowed, so there is no verification that seeds have not been modified, **but this also means that changes in the datacenters.yaml are not reflected in the Seeds** once a Seed has been migrated.

We should really plan on making the migration a quick thing and maybe not leave it enabled for a whole release cycle.

api/cmd/master-controller-manager/main.go
```diff
@@ -63,7 +69,7 @@ func main() {
 	flag.StringVar(&runOpts.workerName, "worker-name", "", "The name of the worker that will only processes resources with label=worker-name.")
 	flag.IntVar(&ctrlCtx.workerCount, "worker-count", 4, "Number of workers which process the clusters in parallel.")
 	flag.StringVar(&runOpts.internalAddr, "internal-address", "127.0.0.1:8085", "The address on which the /metrics endpoint will be served")
-	flag.BoolVar(&runOpts.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters")
+	flag.BoolVar(&runOpts.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters. Enabling this and defining the datcenters flag will enable the seed migration")
```
alvaroaleman: What about `...will enable the migration of the datacenters defined in datancenters.yaml to the Seed CustomResource`?

api/cmd/master-controller-manager/main.go
```diff
@@ -181,6 +206,37 @@ func main() {
...
+				if err != nil {
+					return fmt.Errorf("failed to create kube client: %v", err)
+				}
+
+				// wait for the webhook to be ready
+				if migrationOptions.SeedMigrationEnabled() {
+					deadline := 30 * time.Second
+					endpoint := types.NamespacedName{Namespace: ctrlCtx.namespace, Name: "seed-webhook"}
+
+					log.Infow("waiting for webhook to be ready...", "webhook", endpoint, "deadline", deadline)
```
alvaroaleman: maybe rename to `timeout`? `deadline` sounds like its a point in time, not an interval

api/cmd/master-controller-manager/main.go
```diff
@@ -181,6 +206,37 @@ func main() {
...
+					if err := wait.Poll(500*time.Millisecond, deadline, func() (bool, error) {
+						endpoints := &corev1.Endpoints{}
+						if err := client.Get(ctx, endpoint, endpoints); err != nil {
+							return false, err
+						}
+						return len(endpoints.Subsets) > 0, nil
+					}); err != nil {
+						return fmt.Errorf("failed to wait for webhook: %v", err)
+					}
+					log.Info("webhook is ready")
```
alvaroaleman: I thought log has to start uppercase? :P

api/cmd/master-controller-manager/main.go
```diff
@@ -181,6 +206,37 @@ func main() {
...
+				}
+
+				log.Info("executing migrations...")
+				if err := master.RunAll(ctx, log, client, ctrlCtx.namespace, migrationOptions); err != nil {
```
alvaroaleman: Could we alias the impart as `mastermigrations` or something the like for clarity?

api/pkg/crd/migrations/master/migrations.go
```diff
@@ -0,0 +1,70 @@
...
+
+		log.Info("checking for Seed existence...")
+		existingSeed := kubermaticv1.Seed{}
+		err = client.Get(ctx, key, &existingSeed)
+		if err != nil {
+			if !kerrors.IsNotFound(err) {
+				return fmt.Errorf("failed to get Seed %s: %v", name, err)
+			}
+
+			log.Info("creating Seed CR...")
```
alvaroaleman, line 119: Can we add a comment somewhere here explaining that we do not update them because we know they can't be modified because the webhook prevents that?

api/pkg/crd/migrations/master/migrations.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := migrateDatacenters(ctx, log, client, kubermaticNamespace, opt.DatacentersFile); err != nil {
+			return fmt.Errorf("failed to migrate datacenters.yaml: %v", err)
+		}
+		log.Info("migration completed successfully")
+	}
+
+	return nil
+}
+
+func migrateDatacenters(ctx context.Context, log *zap.SugaredLogger, client ctrlruntimeclient.Client, kubermaticNamespace string, dcFile string) error {
```
alvaroaleman: What I realized yesterday is that we completely missed an important step: We also need to migrate the content of the kubeconfig to a secret that is referenced by the Seed object

api/pkg/validation/seed/server.go
```diff
@@ -32,13 +33,16 @@ func (opts *WebhookOpts) AddFlags(fs *flag.FlagSet) {
...
+// When migrationMode is enabled, only creating new seeds is allowed, not
+// changing or deleting existing.
 func (opts *WebhookOpts) Server(
 	ctx context.Context,
 	log *zap.SugaredLogger,
 	workerName string,
 	seedsGetter provider.SeedsGetter,
-	seedClientGetter provider.SeedClientGetter) (*Server, error) {
+	seedClientGetter provider.SeedClientGetter,
+	migrationMode bool) (*Server, error) {
```
alvaroaleman: Nit: Rename `migationMode` to `migrationModeEnabled`? I thought the var would be a string or sth describing the mode

api/pkg/validation/seed/server.go
```diff
@@ -122,6 +128,12 @@ func (s *Server) handle(req *http.Request) (*admissionv1beta1.AdmissionRequest,
 		return nil, fmt.Errorf("failed to unmarshal request body: %v", err)
 	}
 
+	// during datacenters->seed migration we reject any changes, so the seeds can never get
+	// out of sync with the datacenters.yaml
+	if s.migrationMode && admissionReview.Request.Operation != admissionv1beta1.Create {
+		return admissionReview.Request, errors.New("migration is enabled, changes to Seed resources are forbidden")
```
alvaroaleman: Can we add the info on how to get out of `migrationModeEnabled`?

api/cmd/master-controller-manager/main.go
xrstf: All logs within the main.go and migrations are lowercased, so it's at least consistent here. :-)

api/pkg/crd/migrations/master/migrations.go
xrstf: I've added some code to create kubeconfigs and Secrets for each seed.

api/pkg/crd/migrations/master/migrations.go
xrstf: Added a comment to the function.

api/cmd/master-controller-manager/main.go
```diff
@@ -181,6 +207,37 @@ func main() {
 			}
 
 			return leaderelection.RunAsLeader(leaderCtx, log, cfg, mgr.GetRecorder(controllerName), electionName, func(ctx context.Context) error {
+				// create a dedicated client because the one from the manager is not yet
+				// initialized and returns 404's for everything
+				client, err := ctrlruntimeclient.New(cfg, ctrlruntimeclient.Options{})
```
alvaroaleman: What about just using `mgr.GetAPIReader()`?

alvaroaleman: Also we can put this after the check if the migration is enabled, right? Constructing the client requires doing the REST discovery and that has some associated costs, so we should only do it if we actually need to

xrstf: I can't find a `GetAPIReader` function anywhere in our repo. :thinking:

alvaroaleman: Seems this was added in controller-runtime 0.2. The comment about only creation the client when the seed migration is enabled still applies, thought.

xrstf: Very true. I moved the wait loop to the migration code and introduced a more generic `MigrationEnabled()` for the case when we have multiple migrations.

alvaroaleman (approved):
/lgtm
/approve

PR #4363, opened 2019-09-23
Add endpoints for Openshift console

### LINKED ISSUE #4171: Deploy openshift 4 console

alvaroaleman:

Adds endpoints for the Openshift console:
* `/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/openshift/console/login` which loggs the user into the openshift oauth server
* `/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/openshift/console/proxy"` which proxies the request into an openshift console pod

We could also probably re-use some of the logic here for the Kubernetes dashboard :)

Fixes #4171 

**Documentation**:

```release-note
none
```

/assign @xrstf @zreigz


api/pkg/handler/v1/openshift/openshift.go
```diff
@@ -0,0 +1,468 @@
...
+		// The endpoint the middleware is called with is the innermost one, hence we must
+		// define it as closure and pass it to the middleware() call below.
+		endpoint := func(ctx context.Context, request interface{}) (interface{}, error) {
+			cluster, clusterProvider, err := getClusterProviderFromRequest(ctx, request, projectProvider)
+			if err != nil {
+				writeHTTPError(log, w, err)
+				return nil, nil
+			}
+			log = log.With("cluster", cluster.Name)
+			// TODO: Verify the user is owner, else they should not be allowed to do this
```
alvaroaleman: @zreigz How can we check if the user is cluster owner here? The console gives the user admin permissions for the cluster

zreigz:
```
			userInfo, ok := ctx.Value(middleware.UserInfoContextKey).(*provider.UserInfo)
			if !ok {
				writeHTTPError(log, w, kubermaticerrors.New(http.StatusInternalServerError, err.Error()))
			}
			if userInfo.Group == "owners" {
				consoleLogin(ctx, log, w, cluster, clusterProvider.GetSeedClusterAdminRuntimeClient(), r)
			}
```

api/pkg/handler/routes_v1.go
```diff
@@ -2739,3 +2748,58 @@ func (r Routing) createClusterRole() http.Handler {
...
+//       default: empty
+func (r Routing) openshiftConsoleProxy() http.Handler {
+	return openshift.ConsoleProxyEndpoint(
+		r.log,
+		middleware.TokenExtractor(r.tokenExtractors),
+		r.projectProvider,
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers),
+			middleware.UserSaver(r.userProvider),
+			// TODO: Instead of using an admin client to talk to the seed, we should provide a seed
```
zreigz, line 2827: What do you want to exactly achieve here? If you need some special privileges for the logged user you can extend Role for the user cluster, like for example for the owners: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/rbac-user-cluster/mapper.go#L72

alvaroaleman: But that is for inside the usercluster, right?

What I meant is to extend the clusterroles we already generate in the seed to allow access to the clusternamespace, then use impersonation. Thats more of a cleanup task thought.

zreigz: yes, this is Role which is binded to the user group inside the user cluster.
Ok, now it's clear.

api/pkg/handler/routes_v1.go
```diff
@@ -2739,3 +2748,58 @@ func (r Routing) createClusterRole() http.Handler {
...
+			// client that allows access to the cluster namespace only
+			middleware.SetPrivilegedClusterProvider(r.clusterProviderGetter, r.seedsGetter),
+			middleware.UserInfoExtractor(r.userProjectMapper),
```
zreigz, line 2830: I can't find usage `UserInfo` in this endpoint, you can remove it if not used.

zreigz:
Could you make a small demo during the Dev-Team meeting? I am curious how it works

alvaroaleman:
> Could you make a small demo during the Dev-Team meeting? I am curious how it works

Yes, definitely

alvaroaleman:
@zreigz Updated the PR, PTAL. Unfortunately I can not test it anymore as serviceaccounts can not have the `Owner` role but I think it still works.

The missing ability to have an owner role is also the reason this doesn't have any tests.

api/cmd/kubermatic-api/main.go
```diff
@@ -309,7 +310,8 @@ func createAPIHandler(options serverRunOptions, prov providers, oidcIssuerVerifi
 	registerMetrics()
 
 	mainRouter := mux.NewRouter()
-	mainRouter.Use(setSecureHeaders)
+	// This breaks the Openshift console proxying.
+	//mainRouter.Use(setSecureHeaders)
```
alvaroaleman: @zreigz Any idea why and what we can do to fix this?

zreigz: This is only my guess but you can try to disable this header: https://github.com/kubermatic/kubermatic/blob/master/api/cmd/kubermatic-api/main.go#L382

If won't help you can disable all and enable one by one and you will find which header blocks you and later we can think about solving it.

alvaroaleman: `script-src 'unsafe-inline' and style-src 'unsafe-inline'` were required. I have no idea what that means. Is it safe to allow that? If not, how can we get the Openshift console to work without?
CC @xrstf

api/pkg/handler/routes_v1.go
alvaroaleman: I is needed to get the ClusterProvider

api/cmd/kubermatic-api/main.go
xrstf: These attributes basically mean "allow `eval()` and allow inline styles (like in `<p style="....">`)", the most relaxed CSP possible. That's normal if a web application was not specifically built to *not* require these techniques. Unless the Openshift console documents other CSP rules, it IMHO makes sense to not try to come up with our own but instead let the policy be as open as it can be.

You should be able to override the CSP header per URL, so if you find a way around this setSecureHeaders middleware, you should be golden.

alvaroaleman: I see. I changed the code to just set the `Content-Security-Policy` header in the `console/proxy` endpoint, seems to work fine :+1:

alvaroaleman:
Hm an e2e test is still not possible, because the generated oauth settings require the user to use the domain configured in `external-url`. We will need the kind-based testing for that. I'll create tickets for both.

Other than that I think this is ready for another round, PTAL @zreigz @xrstf

PR #4372, opened 2019-09-24
Fix missing CA injection in cert manager

xrstf:
When updating to 0.10, the default value for the injectAPIServiceCA flag was changed to true. We also had a slgiht error in our template, which would have made the flag ineffective.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #4381, opened 2019-09-25
Add Kubernetes Dashboard helm config

floreks:
This PR adds Kubernetes Dashboard with the auth reverse proxy that will inject authorization header taken from OIDC or from the config if OIDC is not configured for the K8S hosting the Kubermatic. This way access to the Dashboard is restricted.

Current demo deployed based on this config can be found at https://dashboard.dev.kubermatic.io
Only users with email in domain `loodse.com` are allowed to access it.

Wait until https://github.com/kubermatic/secrets/pull/371 gets merged.
/hold

N/A

```release-note
NONE
```


config/kubernetes-dashboard/templates/dashboard-certs-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard-certs
+data:
+  "tls.crt": ""
```
alvaroaleman: Please make these variables

config/kubernetes-dashboard/templates/dashboard-csrf-secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard-csrf
+type: Opaque
+data:
+  csrf: ""
```
alvaroaleman: Please make this a variable

config/kubernetes-dashboard/templates/dashboard-key-holder-secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    k8s-app: kubernetes-dashboard
+  name: kubernetes-dashboard-key-holder
+type: Opaque
```
alvaroaleman, line 10: This is not supposed to contain any data?

config/kubernetes-dashboard/templates/oauth-proxy-deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+            - --redirect-url=https://{{ .Values.dashboard.domain }}/{{ .Values.dashboard.oauth.ingress.path }}/callback
+            - --oidc-issuer-url={{ .Values.dashboard.oidc.issuerUrl }}
+            - --cookie-secure=false
+            - --upstream=file:///dev/null
+            - --http-address=0.0.0.0:4180
+            - --ssl-insecure-skip-verify
+            - --set-authorization-header=true
+          env:
+            - name: OAUTH2_PROXY_CLIENT_ID
+              value: {{ .Values.dashboard.oidc.clientID }}
```
alvaroaleman: Please put these into a secret and reference the secret here

alvaroaleman (commented):
CC @xrstf

config/kubernetes-dashboard/templates/dashboard-certs-secret.yaml
floreks: This will not be used in our case as we start Dashboard on HTTP. I can actually remove it from the config.

config/kubernetes-dashboard/templates/dashboard-key-holder-secret.yaml
floreks: This acts as a storage and will be used by Dashboard.

config/kubernetes-dashboard/templates/dashboard-csrf-secret.yaml
floreks: This one we can export.

config/kubernetes-dashboard/templates/dashboard-certs-secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    k8s-app: kubernetes-dashboard
```
xrstf: Please do not use these outdated labels anymore, but use the app and instance labels from https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ -- set the name label to "kubernetes-dashboard" and the instance label to `{{ .Release.Name }}`.

Please also ensure that the chart can be installed multiple times by not hard-coding `kubernetes-dashboard` everywhere. Take a look at the labels and names used in https://github.com/kubermatic/kubermatic/blob/master/config/monitoring/karma/templates/configmap.yaml and the common helper template in https://github.com/kubermatic/kubermatic/blob/master/config/monitoring/karma/templates/_helpers.tpl

config/kubernetes-dashboard/templates/dashboard-deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+    matchLabels:
+      k8s-app: kubernetes-dashboard
+  template:
+    metadata:
+      labels:
+        k8s-app: kubernetes-dashboard
+    spec:
+      containers:
+        - name: kubernetes-dashboard
+          image: kubernetesui/dashboard:v2.0.0-beta4
```
xrstf: The image should be configurable and start with `docker.io/....`, just like the imagePullPolicy should be configurable. Please use a structure with repository, tag and pullPolicy like in https://github.com/kubermatic/kubermatic/blob/master/config/monitoring/karma/values.yaml#L24-L27

config/kubernetes-dashboard/templates/oauth-proxy-deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+kind: Deployment
+metadata:
+  labels:
+    k8s-app: oauth-proxy
+  name: oauth-proxy
+spec:
+  selector:
+    matchLabels:
+      k8s-app: oauth-proxy
+  replicas: 1
```
xrstf: Can we increase this to two just so the proxy doesn't go down when a node is preemted? Same for all other deployments where is makes sense.

config/kubernetes-dashboard/templates/scraper-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            runAsUser: 1001
+            runAsGroup: 2001
+      serviceAccountName: kubernetes-dashboard
+      nodeSelector:
+        "beta.kubernetes.io/os": linux
+      # Comment the following tolerations if Dashboard must not be deployed on master
+      tolerations:
```
xrstf, line 56: Node selectors, tolerations and affinities should be configurable like in https://github.com/kubermatic/kubermatic/blob/master/config/monitoring/karma/values.yaml#L40-L50 (note how the affinity is a **template**, so it can handle dynamic label names). If it makes sense, have separate sets of selector+toleration+affinity per deployment. If it doesn't make sense, a single set used for all deployments would be fine as well.

The affinity template can be used like this in a Deployment template: https://github.com/kubermatic/kubermatic/blob/master/config/monitoring/karma/templates/deployment.yaml#L71

config/kubernetes-dashboard/templates/dashboard-certs-secret.yaml
floreks: I am wondering if it even makes sense to install Dashboard multiple times in the same cluster. It is supposed to monitor the cluster it is installed in, so I don't see why we would need to install it multiple times, especially in the same namespace.

floreks:
@xrstf @alvaroaleman PTAL

config/kubernetes-dashboard/templates/dashboard-clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: kubernetes-dashboard
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: '{{ .Chart.Version }}'
+    app.kubernetes.io/managed-by: helm
+  name: kubernetes-dashboard
```
xrstf: This name and most others are still hardcoded, but should be `{{ template "name" . }}` to allow multiple installations.

floreks: Yes, because I am not sure if we should allow multiple installations. I don't see a reason why we would need multiple Dashboards inside a single cluster. It is supposed to monitor the cluster it is installed in, so installing it many times will not give us any advantage.

xrstf: It makes testing chart updates much, much easier if Prow is not constantly overwriting your changes.

floreks: It might make sense to install it in a different namespace (usercluster) if we want to connect it to an outside cluster, but then name conflicts do not apply.

xrstf: They still do, because you install cluster-wide resources like ClusterRoles.

floreks: There might be a problem with using generated names for all resources as Dashboard expects i.e. configmap with name `kubernetes-dashboard-settings` in the same namespace. It also requires some secrets. Those names right now are not configurable. We could use `.Release.Name` for others though.

floreks:
I have used a release name where it was possible. Other resources have to stay as-is.

xrstf:
Does it make sense to add an OWNERS file to this chart so the UI team owns this?

floreks:
I think it makes sense.

floreks:
Done

config/kubernetes-dashboard/OWNERS
```diff
@@ -0,0 +1,13 @@
...
+  - team-ui
+
+reviewers:
+  - team-ui
+
+labels:
+  - team/ui
+
+options:
+  no_parent_owners: true
```
xrstf, line 13: This would exclude the seed team from maintaining the chart. Is this really what we want?

floreks: No idea. If someone else than us will want to modify it, we can extend it.

floreks:
Let's see if it works :slightly_smiling_face: 

/hold cancel

PR #4387, opened 2019-09-26
Add migrations for credential secrets

nikhita:
For https://github.com/kubermatic/kubermatic/issues/3792

```release-note
NONE
```

### DIFF
--- a/api/pkg/crd/migrations/credentials.go
+++ b/api/pkg/crd/migrations/credentials.go
@@ -1,0 +1,654 @@
+package migrations
+
+import (
+	"fmt"
+	"github.com/golang/glog"
+
+	apiv1 "github.com/kubermatic/kubermatic/api/pkg/api/v1"
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	kuberneteshelper "github.com/kubermatic/kubermatic/api/pkg/kubernetes"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func createSecretsForCredentials(cluster *kubermaticv1.Cluster, ctx *cleanupContext) error {
+	var err error
+	if cluster.Spec.Cloud.AWS != nil {
+		err = createAWSSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Azure != nil {
+		err = createAzureSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Digitalocean != nil {
+		err = createDigitaloceanSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.GCP != nil {
+		err = createGCPSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Hetzner != nil {
+		err = createHetznerSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Openstack != nil {
+		err = createOpenstackSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Packet != nil {
+		err = createPacketSecret(cluster, ctx)
+	}
+	if cluster.Spec.Cloud.Kubevirt != nil {
+		err = createKubevirtSecret(cluster, ctx)
+	}
+	if err != nil {
+		return err
+	}
+
+	kuberneteshelper.AddFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer)
+	updatedCluster, err := ctx.kubermaticClient.KubermaticV1().Clusters().Update(cluster)
+	if err != nil {
+		return fmt.Errorf("failed to update cluster %q: %v", cluster.Name, err)
+	}
+	*cluster = *updatedCluster
+	return nil
+}
+
+func createAWSSecret(cluster *kubermaticv1.Cluster, ctx *cleanupContext) error {
+	spec := cluster.Spec.Cloud.AWS
+	referenceDefined := spec.CredentialsReference != nil && spec.CredentialsReference.Name != ""
+	valuesDefined := spec.AccessKeyID != "" || spec.SecretAccessKey != ""
+
+	var secret *corev1.Secret
+	if !referenceDefined && valuesDefined {
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cluster.GetSecretName(),
+				Namespace: resources.KubermaticNamespace,
+			},
+			Type: corev1.SecretTypeOpaque,
+			Data: map[string][]byte{
+				resources.AWSAccessKeyID:     []byte(spec.AccessKeyID),
+				resources.AWSSecretAccessKey: []byte(spec.SecretAccessKey),
+			},
+		}
+
+		if _, err := ctx.kubeClient.CoreV1().Secrets(resources.KubermaticNamespace).Create(secret); err != nil {
+			return err
+		}
+		glog.Infof("the secret = %s was created for cluster = %s resource", cluster.GetSecretName(), cluster.Name)
+	}
+
+	if !referenceDefined {
+		cluster.Spec.Cloud.AWS.CredentialsReference = &providerconfig.GlobalSecretKeySelector{
+			ObjectReference: corev1.ObjectReference{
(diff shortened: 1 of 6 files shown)


api/pkg/crd/migrations/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+		if _, err := ctx.kubeClient.CoreV1().Secrets(resources.KubermaticNamespace).Create(secret); err != nil {
+			return err
+		}
+		glog.Infof("the secret = %s was created for cluster = %s resource", secret.Name, cluster.Name)
+	}
+
+	if !referenceDefined {
+		cluster.Spec.Cloud.AWS.CredentialsReference = &providerconfig.GlobalSecretKeySelector{
+			ObjectReference: corev1.ObjectReference{
+				Name:      secret.Name,
```
alvaroaleman: I think this will result in a NPD when `!referenceDefined && !valuesDefined`, maybe just use `cluster.GetSecretName` and `resources.KubermaticNamespace`?

api/pkg/crd/migrations/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+				Namespace: secret.Namespace,
+			},
+		}
+	}
+
+	if valuesDefined {
+		cluster.Spec.Cloud.AWS.AccessKeyID = ""
+		cluster.Spec.Cloud.AWS.SecretAccessKey = ""
+	}
+
```
alvaroaleman, line 126: I think we will need to update the secret here if one of `!referenceDefined` or `valuesDefined` was true

api/pkg/crd/migrations/credentials.go
```diff
@@ -0,0 +1,235 @@
...
+				Namespace: secret.Namespace,
+			},
+		}
+	}
+
+	if valuesDefined {
+		cluster.Spec.Cloud.Digitalocean.Token = ""
+	}
+
+	return nil
```
alvaroaleman, line 283: Basically the same comments as for AWS. Also I know this is annoying but can we have unittests for all of this? Small mistakes here will unfortunately cause pretty big problems and I do not trust myself to be able to catch everything here.

api/pkg/crd/migrations/credentials.go
nikhita: > I think this will result in a NPD when !referenceDefined && !valuesDefined

Won't this result in a validation error early on, so we don't need to consider that case here?

Edit: ignore this comment :)

nikhita:
Updated. PTAL.

alvaroaleman:
So code looks good to me but I have to admit I am still a bit hesitant to merge this :grimacing: 
@zreigz Are we sure all our endpoints can cope with this?

zreigz:
It looks good, let's see what will happen

alvaroaleman:
> It looks good, let's see what will happen

In particular, I remember @kgroschoff mentioning we can update cluster credentials. What endpoint is that, PatchCluster? Can it cope with the credential ref?

zreigz:
> > It looks good, let's see what will happen
> 
> In particular, I remember @kgroschoff mentioning we can update cluster credentials. What endpoint is that, PatchCluster? Can it cope with the credential ref?

Yes, with PatchCLuster you can change the cluster fields also credential ref.

alvaroaleman:
> Yes, with PatchCLuster you can change the cluster fields also credential ref.

Sorry for unclear phrasing: Is the user still able to change the credentials? That change does not change the credentialRef, it changes the values in the referenced secret.

zreigz:
The API allows changing all fields in the cluster. In the UI with edit options, you can change the credentials field, not the credentialRef.

alvaroaleman:
> The API allows changing all fields in the cluster. In the UI with edit options, you can change the credentials field, not the credentialRef.

So do we need a new "change credentials" endpoint here?

nikhita:
Updated. PTAL.

api/pkg/provider/kubernetes/update_credentials.go
```diff
@@ -0,0 +1,378 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UpdateCredentialSecretForCluster checks if credentials are directly specified on the cluster object.
+// If yes, it compares this value to the value in the referenced secret. If the values are not equal,
+// it updates the secret and clears the credentials field.
+func UpdateCredentialSecretForCluster(ctx context.Context, seedClient ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster, projectID string) error {
```
alvaroaleman: What about re-using the `CreateNNNSecret` code here? We can use our [`reconcileSecrets`](https://github.com/kubermatic/kubermatic/blob/8b1406f52a81d1e3805d1b62021e203dee430bab/api/pkg/resources/reconciling/zz_generated_reconcile.go#L111) code to get an upsert for the secret

nikhita:
Hmm, the test is still running but prow says it has failed.

alvaroaleman:
> Hmm, the test is still running but prow says it has failed.

Not sure whats going on but that test is definitely not running anymore.

/retest

alvaroaleman:
> Not sure whats going on but that test is definitely not running anymore.

So apparently spyglass considers the test as running until it has a `finished.json` in gcs which this test run doesn't have for some reason, even thought all other files exist there. The ProwJob CR correctly has a status of `failure`.

api/pkg/crd/migrations/credentials.go
```diff
@@ -0,0 +1,654 @@
...
+	kuberneteshelper.AddFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer)
+	updatedCluster, err := ctx.kubermaticClient.KubermaticV1().Clusters().Update(cluster)
+	if err != nil {
+		return fmt.Errorf("failed to update cluster %q: %v", cluster.Name, err)
+	}
+	*cluster = *updatedCluster
+	return nil
+}
+
+func createAWSSecret(cluster *kubermaticv1.Cluster, ctx *cleanupContext) error {
```
alvaroaleman, line 56: Why do we not re-use the `CreateOrUpdateCredentialSecretForCluster` func here and instead duplicate everything?

nikhita: Mainly because the migration code uses the `Clientset` and `CreateOrUpdateCredentialSecretForCluster` takes in a `ctrlruntimeclient.Client`...

alvaroaleman: Okay, so what about extending the `cleanupContext` to contain a `ctrlruntimeclient.Client`? `RunAll` already gets a `*rest.Config` so shouldn't be hard. I would really like to avoid having the two diverge accidentally :grimacing:

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -9,335 +9,628 @@ import (
...
-			Namespace: resources.KubermaticNamespace,
-			Labels: map[string]string{
-				kubermaticv1.ProjectIDLabelKey: projectID,
-				"name":                         name,
+func createOrUpdateAWSSecret(ctx context.Context, seedClient ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster, projectID string) error {
+	spec := cluster.Spec.Cloud.AWS
+	referenceDefined := spec.CredentialsReference != nil && spec.CredentialsReference.Name != ""
+	valuesDefined := spec.AccessKeyID != "" || spec.SecretAccessKey != ""
+
+	if !referenceDefined && valuesDefined {
```
alvaroaleman, line 50: Could you build these as a `NamedSecretCreatorGetter` and use `reconcile.Secrets` so we have a proper upsert?

* Sample of a NamedSecretCreatorGetter: https://github.com/kubermatic/kubermatic/blob/8f25c22ee1ce5d61ef32901de1024f5d683c0152/api/pkg/resources/apiserver/token-users.go#L14
* Sample ussage of `reconcileSecrets`: 
https://github.com/kubermatic/kubermatic/blob/8f25c22ee1ce5d61ef32901de1024f5d683c0152/api/pkg/controller/cluster/resources.go#L276

We also have some docs for this: https://github.com/kubermatic/kubermatic/blob/master/docs/resource-handling.md

Ping me if you need more info!

nikhita: I tried this approach as well, but this leads to duplicated code (again) for migrations and the normal operation...because this approach can't be used for migration.

Let me know if you'd still prefer this way though and I can add it again.

Currently, the code is de-duplicated.

nikhita:
Closing in favour of https://github.com/kubermatic/kubermatic/pull/4616

PR #4389, opened 2019-09-26
fix client api generation for golang v1.13

zreigz:
**What this PR does / why we need it**: Fix client API generation for golang v1.13

Fixes #

According to this bug https://github.com/go-swagger/go-swagger/issues/2063 the schema validation was skipped. It happens for go v1.13. Our CI uses go v1.12 and everything works fine. There is still a problem with `verify-swagger.sh` which use the command `swagger validate` and cannot be skipped for this step.

```release-note
NONE
```


zreigz:
I also tried update go-swagger for master branch (and new version) but it hangs for `swagger generate spec` command.
Update all dependencies helps but our code can not be compiled. So we have to wait until they fix it for new golang version.

alvaroaleman:
Okay, so this effectively blocks us from moving to go 1.13?

zreigz:
yes, `verify-swagger.sh` won't work. My PR fix only the problem to generate API client locally to go forward with PRs.

alvaroaleman:
> yes, verify-swagger.sh won't work. My PR fix only the problem to generate API client locally to go forward with PRs.

Okay, thanks for the heads-up.
/lgtm
/approve

PR #4390, opened 2019-09-26
Add credentials-cleanup finalizer after secret creation

### LINKED ISSUE #4388: Credentials secrets cleanup is broken
When a cluster is created, corresponding secrets for storing credentials are created in the `kubermatic` namespace. Example: for AWS, `credential-aws-<cluster-name>` is the secret created in the `kubermatic` namespace.

When the cluster is deleted, it gets the `kubermatic.io/cleanup-credentials-secrets` finalizer. Once the secret is deleted, the finalizer should be removed and then the cluster should get deleted.

However, the current behaviour is that the finalizer gets added, the secret _does not_ get deleted, but the finalizer gets removed (and the cluster gets deleted), which means that we end up with orphaned secrets in the `kubermatic` namespace.

nikhita:
Fixes https://github.com/kubermatic/kubermatic/issues/4388

Current behaviour is to add the finalizer while deleting the cluster
in the `DeleteEndpoint` method. This works fine if we are deleting
the cluster using the API/dashboard.

But if someone manually issues a `kubectl delete cluster` command, it
does not hit the `DeleteEndpoint` method, due to which the cleanup
finalizer is not added and the secret gets orphaned.

To prevent this scenario, we can directly add the finalizer when we
create the secret. This would also help in cases when `kubectl` is
directly used.

```release-note
NONE
```

/kind bug

### DIFF
--- a/api/pkg/handler/v1/cluster/cluster.go
+++ b/api/pkg/handler/v1/cluster/cluster.go
@@ -127,6 +127,7 @@ func CreateEndpoint(sshKeyProvider provider.SSHKeyProvider, projectProvider prov
 		if err := kubernetesprovider.CreateCredentialSecretForCluster(ctx, privilegedClusterProvider.GetSeedClusterAdminRuntimeClient(), partialCluster, req.ProjectID); err != nil {
 			return nil, err
 		}
+		kuberneteshelper.AddFinalizer(partialCluster, apiv1.CredentialsSecretsCleanupFinalizer)
 
 		newCluster, err := clusterProvider.New(project, userInfo, partialCluster)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


api/pkg/provider/kubernetes/credentials.go
```diff
@@ -66,6 +68,7 @@ func createAWSSecret(ctx context.Context, seedClient ctrlruntimeclient.Client, c
 	// remove credentials from cluster object
 	cluster.Spec.Cloud.AWS.AccessKeyID = ""
 	cluster.Spec.Cloud.AWS.SecretAccessKey = ""
+	kuberneteshelper.AddFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer)
```
alvaroaleman: This can be added unconditionally on creation, as the finalizer cleanup is a no-op when there is no secret, since it returns without an error if either the secret name is empty or a `get` on the secret returns a 404

alvaroaleman (commented):
Are you sure this fixes the underlying problem? From your issue description:

> However, the current behaviour is that the finalizer gets added, the secret does not get deleted, but the finalizer gets removed (and the cluster gets deleted), which means that we end up with orphaned secrets in the kubermatic namespace.

The only case where I see this PR making a difference is when someone uses `kubectl` for cluster deletion

nikhita:
/assign @zreigz 
for approval

PR #4393, opened 2019-09-26
 Use secrets for credentials for Openstack

nikhita:
For https://github.com/kubermatic/kubermatic/issues/3792 
/hold

```release-note
NONE
```


api/pkg/handler/v1/provider/openstack.go
```diff
@@ -67,13 +67,25 @@ func OpenstackSizeWithClusterCredentialsEndpoint(projectProvider provider.Projec
...
+		}
+
+		return getOpenstackSizes(creds.Username, creds.Password, creds.Tenant, creds.TenantID, creds.Domain, datacenterName, datacenter)
 	}
 }
 
 func getOpenstackSizes(username, passowrd, tenant, tenantID, domain, datacenterName string, datacenter *kubermaticv1.Datacenter) ([]apiv1.OpenstackSize, error) {
 
-	provider, err := openstack.NewCloudProvider(datacenter)
+	provider, err := openstack.NewCloudProvider(datacenter, nil)
```
alvaroaleman: Hey, please do not pass nil dependencies in. Can we instead just make the `GetFlavors` standalone instead of it requiring an initialized provider?

api/pkg/handler/v1/provider/openstack.go
```diff
@@ -543,7 +599,7 @@ func getOpenstackCloudProvider(seeds map[string]*kubermaticv1.Seed, datacenterNa
 	if err != nil {
 		return nil, fmt.Errorf("failed to find datacenter %q: %v", datacenterName, err)
 	}
-	osProvider, err := openstack.NewCloudProvider(dc)
+	osProvider, err := openstack.NewCloudProvider(dc, nil)
```
alvaroaleman: Same here wrt to the nil dependency. Please just make the corresponding funcs in the openstack cloud provider package not require an initialized provider

api/pkg/resources/credentials.go
```diff
@@ -39,6 +40,14 @@ type HetznerCredentials struct {
 	Token string
 }
 
+type OpenstackCredentials struct {
```
alvaroaleman, line 43: Nit: Why is it needed to define this struct twice, here and in the provider/cloud/openstack package?

api/pkg/validation/cluster.go
```diff
@@ -261,17 +261,32 @@ func ValidateCloudSpec(spec kubermaticv1.CloudSpec, dc *kubermaticv1.Datacenter)
...
+		}
 	}
 	if spec.Password == "" {
-		return errors.New("no password specified")
+		if err := kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackPassword); err != nil {
+			return err
+		}
+	}
+
+	var err1, err2 error
```
alvaroaleman: err :D Can we just make this a `var errors []error` instead? :P

nikhita:
Openstack test is green: https://prow.loodse.com/view/gcs/prow-data/pr-logs/pull/kubermatic_kubermatic/4393/pull-kubermatic-e2e-openstack-coreos-1.15/1177514758010572802

Re-pushed to fix lint error

api/pkg/validation/cluster.go
```diff
@@ -261,17 +261,32 @@ func ValidateCloudSpec(spec kubermaticv1.CloudSpec, dc *kubermaticv1.Datacenter)
...
 	if spec.Password == "" {
-		return errors.New("no password specified")
+		if err := kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackPassword); err != nil {
+			return err
+		}
+	}
+
+	var errs []error
+	if spec.Tenant == "" && spec.CredentialsReference != nil && spec.CredentialsReference.Name != "" {
+		errs = append(errs, kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackTenant))
```
alvaroaleman, line 282: Wait, we just append the returned error here. This means we end up with two `nil` errors in the `errs` slice, which means its lenght won't be zero, right?

What you can do thought is to use `utilerror "k8s.io/apimachinery/pkg/util/errors".NewAggregate`, the returned error implementation is nil if either the lenght of the passed in err slice is zero or all errors in the slice are nil

nikhita: Done

nikhita:
/hold

test is broken, so retesting won't help

PR #4400, opened 2019-09-29
Update nginx-ingress-controller to 0.26.1

xrstf:
This has a fix for the issue plaguing our offline cluster. It was a smooth update on dev.

This PR also adds the `allowPrivilegeEscalation` field to the security context, which was set since 0.22 by https://github.com/kubernetes/ingress-nginx/pull/3342 but we never noticed because we don't have policies forbidding this behaviour by default.

The memory limit was increased because the release notes for 0.26 mentioned higher resource usage while connections are drained.

```release-note
Update nginx-ingress-controller to 0.26.1
```

### DIFF
--- a/config/nginx-ingress-controller/Chart.yaml
+++ b/config/nginx-ingress-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: nginx-ingress-controller
-version: 1.0.2
-appVersion: 0.25.1
+version: 1.0.3
+appVersion: 0.26.0
 description: nginx-ingress-controller
 keywords:
 - kubermatic
--- a/config/nginx-ingress-controller/Chart.yaml
+++ b/config/nginx-ingress-controller/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: nginx-ingress-controller
 version: 1.0.3
-appVersion: 0.26.0
+appVersion: 0.26.1
 description: nginx-ingress-controller
 keywords:
 - kubermatic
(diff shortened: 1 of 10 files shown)


config/nginx-ingress-controller/values.yaml
```diff
@@ -4,16 +4,18 @@ nginx:
   replicas: 3
   image:
     repository: quay.io/kubernetes-ingress-controller/nginx-ingress-controller
-    tag: 0.25.1
+    tag: 0.26.0
   config: {}
+  extraArgs:
+  - '--default-ssl-certificate=default/kubermatic-tls-certificates'
```
kron4eg, line 10: What’s the reason behind this move? It used to be a strictly defined integration point between ingress controller and certmanager and now it became a possible to be overwritten and forgotten configuration option.

xrstf: It felt weird to read some Kubermatic identifier within the nginx template, and a good opportunity to add the extraArgs stuff. But I don't feel strongly about it.

config/nginx-ingress-controller/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: nginx-ingress-controller
-version: 1.0.2
-appVersion: 0.25.1
+version: 1.0.3
+appVersion: 0.26.0
```
kron4eg: Bugfix 0.26.1 is available

xrstf: :trollface:

config/nginx-ingress-controller/values.yaml
xrstf: So.... keep it or remove it?

kron4eg (approved):
/lgtm

Issue #4254, opened 2019-09-05
Investigate registry usage for openshift


alvaroaleman:
**User Story**

Openshift comes with an integrated registry by default, we should figure out why, how important it is and think about if and how we want to do the same on

**Acceptance criteria**

* We know what for the integrated registry in openshift is used
* We know what openshift functionalities we break by not deploying it
* We wrote down possible approaches to deploy it on different providers (i.E., what do we want to use as storage backend?

Issue #2948, opened 2019-03-07
Model resource deployment as a DAG


alvaroaleman:
**User Story**

Right now we deploy the control plane components ordered by type and then ordered by what the developer of the code did (e.G secrets first, then ConfigMaps, then Deployments). This is only a very rough approximation of the actual hirarchy and requires developers and reviewers to be very careful, to e.G. always make sure that CA secrets get created before certificates that use that CA get created, otherwise we run into a dead lock.

Ideally it would be possible that every resource can specify its requirements (other resources but also e.G. something that is managed by another controller like the cloud-controller) and we calculate a DAG from that and just traverse that DAG.

**Acceptance criteria**
* All resources managed by the cluster-controller or openshift-controller can specify requirements they have
* These requirements may either be other resources or something that can be derived from the cluster CR
* Based on those requirements we calculate a DAG
* The resource deployment is done by traversing the DAG instead of by deploying the resources ordered by type

alvaroaleman:
An alternate approach (suggested by @moadqassem ) would be to let all kube objects we deploy  implement a

```
type resourceDeployment interface {
  // Blocks until all dependencies are met
  // Uses a kube client to poll the API
  WaitForDependencies() error
  // Deplos the object
  Deploy() error
}
```

Then have in the cluster/openshift controller a simple
```
resources := []resourceDeployment{...}
errChan := make(chan error)
for _, resource {
  go func() { 
    if err := resource.WaitForDependencies(); err != nil {
         errChan <- err
         return
     }
    if err := resource.Deploy(); err != nil {
      errChan <- err
    }
  }()
}
```

(waitgroup and timeout omitted for simplicity)

irozzo-1A:
Re-opening as this approach could be a good fit for #6225 and #5949.

irozzo-1A:
I'm also evaluating [expectations](https://github.com/kubernetes/kubernetes/blob/1889a6ef52eb18b08e24843577c5b9d3b9a65daa/pkg/controller/controller_utils.go) used in some k8s controllers as an alternative approach (thx @mfranczy for the suggestion).

PR #4404, opened 2019-09-30
Openshift: Cloud credentials operator

alvaroaleman:

Fixes #

**Documentation**:

```release-note
none
```

The thing almost certainly does't actually work as it does not have credentials to talk to the cloudprovider, I will update that when working further on the registry in #4254 

/assign @xrstf


api/pkg/controller/openshift/resources/testdata/kube-config.golden.yaml
```diff
@@ -10,6 +10,10 @@ data:
       - 'true'
       cert-dir:
       - /var/run/kubernetes
+      cloud-provider:
+       - fake-cloud-provider
```
xrstf: This and line 16 are indented other than the others (like cert-dir's value). It works, but it's an invitation to inconsistency and will sooner or later break.

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -340,3 +345,36 @@ func (r *reconciler) reconcileNamespaces(ctx context.Context) error {
...
+	if !r.openshift {
+		return nil
+	}
+
+	creators := []reconciling.NamedUnstructuredCreatorGetter{
+		openshift.InfrastructureCreatorGetter(r.platform),
+	}
+	r.log.Debug("Reconciling unstructured")
+	// The delegatingReader from the `mgr` always redirects request for unstructured.Unstructured
+	// to the API even thought the cache-backed reader is perfectly well capable of creating watches
```
xrstf: `even though` without the `t`. `perfectly capable` without `well`.

api/pkg/resources/reconciling/unstructured.go
```diff
@@ -0,0 +1,53 @@
...
+		emptyObject.SetKind(kind)
+		emptyObject.SetAPIVersion(apiVersion)
+
+		createObject := UnstructuredObjectWrapper(create, emptyObject)
+		for _, objectModifier := range objectModifiers {
+			createObject = objectModifier(createObject)
+		}
+
+		if err := EnsureNamedObject(ctx, types.NamespacedName{Namespace: namespace, Name: name}, createObject, client, emptyObject, false); err != nil {
+			return fmt.Errorf("failed to ensure Unstructured %s/%s: %v", namespace, name, err)
```
xrstf: How about including kind and apiVersion in the error message?

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -340,3 +345,36 @@ func (r *reconciler) reconcileNamespaces(ctx context.Context) error {
...
+	// own client that uses the cache as reader.
+	client := ctrlruntimeclientClient{
+		Reader:       r.cache,
+		Writer:       r.Client,
+		StatusClient: r.Client,
+	}
+	if err := reconciling.ReconcileUnstructureds(ctx, creators, "", client); err != nil {
+		return fmt.Errorf("failed to reconcile unstructureds: %v", err)
+	}
+	r.log.Debug("Finished reconciling unstructured")
```
xrstf, line 371: I get where this is coming from and that it makes sense for everyone who has the Kubermatic code, but "reconciling unstructured[ things]" is not really a helpful log message. Maybe we can one time find a better term for this. :-)

alvaroaleman: Its the only info available at this point and its debug log for a reason :)

xrstf: True and true, but maybe we shouldn't group the resources by the method by which they are reconciled (concrete or unstructured), but instead by "topic", so we can have a more meaningful log message in the future ;-)

alvaroaleman: That is a current shortcoming of the implementation (among others) which is tracked here: https://github.com/kubermatic/kubermatic/issues/2948

PR #4409, opened 2019-10-01
Use single timeout window in e2e tests

nikhita:
Uses a single timeout window of 10 minute for e2e tests, but keeps the granual feedback added in https://github.com/kubermatic/kubermatic/pull/4348.

```release-note
NONE
```

### DIFF
--- a/api/cmd/conformance-tests/runner.go
+++ b/api/cmd/conformance-tests/runner.go
@@ -343,37 +343,38 @@ func (r *testRunner) executeScenario(log *zap.SugaredLogger, scenario testScenar
 		return report, fmt.Errorf("failed to setup nodes: %v", err)
 	}
 
-	var overallTimeout time.Duration = 10 * time.Minute
+	var overallTimeout = 10 * time.Minute
+	var timeoutLeft time.Duration
 	if cluster.Annotations["kubermatic.io/openshift"] == "true" {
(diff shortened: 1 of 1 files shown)


api/cmd/conformance-tests/runner.go
```diff
@@ -343,38 +343,43 @@ func (r *testRunner) executeScenario(log *zap.SugaredLogger, scenario testScenar
...
 		"[Kubermatic] Wait for machines to get a node",
 		report,
 		func() error {
-			return waitForMachinesToJoinCluster(log, userClusterClient, nodeJoinTimeout)
+			return waitForMachinesToJoinCluster(log, userClusterClient, overallTimeout)
 		},
 	); err != nil {
 		return report, fmt.Errorf("failed to wait for machines to get a node: %v", err)
 	}
+	timeoutLeft := time.Duration(overallTimeout.Minutes() - time.Since(startTime).Minutes())
```
alvaroaleman: I _think_ the issue is that `time.Duration` is an int64 representing nanoseconds. What you do here is taking the number of minutes left and interpreting them as nanoseconds.

You could probably factor this out into a function in order to be able to write a unittest for it.

nikhita:
Tests passing and ready for review. :tada:

PR #4414, opened 2019-10-01
Extend controllers to deploy Kubernetes Dashboard 

floreks:
Dashboard will be now installed by the controller in every user cluster, together with the metrics scraper. The main Dashboard binary will run on the seed cluster. Other required components will be installed in the user cluster.

I think we can delete Dashboard as an addon. Are then any additional steps required to remove it othen than deleting from `addons` directory?

**TODO**
- [x] Add comments in-place of todo's
- [ ] ~Remove addon~ (postponed)
- [x] Fix tests

Related to https://github.com/kubermatic/dashboard-v2/issues/1413

```release-note
NONE
```


.gitignore
```diff
@@ -4,6 +4,8 @@
 .env
 .idea
 *.iml*
+go.mod
```
floreks, line 7: I am using go modules to run everything locally and it creates those files, so I added them to the gitignore.

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -111,6 +117,15 @@ func (r *reconciler) reconcileServiceAcconts(ctx context.Context) error {
 	if err := reconciling.ReconcileServiceAccounts(ctx, creators, metav1.NamespaceSystem, r.Client); err != nil {
 		return fmt.Errorf("failed to reconcile ServiceAccounts in the namespace %s: %v", metav1.NamespaceSystem, err)
 	}
+
+	// Kubernetes Dashboard and related resources
+	creators = []reconciling.NamedServiceAccountCreatorGetter{
+		metricsscraper.ServiceAccountCreator(),
```
alvaroaleman: You can just append the Creator you added to the existing list in line 113.

More importantly though: What keeps you from using a cert for this user? Serviceaccounts have the unfortunate implication that we have to copy the serviceaccountsecret from the usercluster into the seed

api/pkg/controller/usercluster/resources/kubernetes-dashboard/namespace.go
```diff
@@ -0,0 +1,14 @@
+package kubernetesdashboard
+
+import (
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NamespaceCreator TODO(floreks)
+func NamespaceCreator() (string, reconciling.NamespaceCreator) {
```
alvaroaleman, line 10: Why do you need a namespace in the usercluster for deploying the dashboard into the seed?

api/pkg/controller/usercluster/resources/kubernetes-dashboard/secret.go
```diff
@@ -0,0 +1,36 @@
+package kubernetesdashboard
```
alvaroaleman, line 1: Would it be possible to create the secrets here in the seed in the usercluster namespace and pass them onto the dashboard via a flag or something? The end user doesn't really need to see them IMHO

api/pkg/controller/usercluster/resources/metrics-scraper/clusterrole.go
```diff
@@ -0,0 +1,33 @@
+package metricsscraper
```
alvaroaleman: Can we somehow make it more obvious all of this is needed for the dashboard, by either renaming this package to something that contains the word "dashboard" or consolidating the two packages?

api/pkg/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,139 @@
...
+				},
+			},
+			Ports: []corev1.ContainerPort{
+				{
+					ContainerPort: 9090,
+					Protocol:      corev1.ProtocolTCP,
+				},
+			},
+			SecurityContext: &corev1.SecurityContext{
+				RunAsUser:                &[]int64{1001}[0],
```
alvaroaleman: Do I understand this correctly that you basically create a one element slice in order to be able to get a pointer to a int64 without a helper? If yes, please use the helpers from `utilpointer "k8s.io/utils/pointer"` instead, because this is not easy to read :)

alvaroaleman (commented):
some small comments, but looks very good overall

.gitignore
alvaroaleman: You are obviously free to do whatever works for you, but I think the result you get when using go modules may be different from what CI and others see, because they will user whatever version you specified via a constraint or latest, instead of what is in `/vendor`

api/pkg/controller/usercluster/reconciler.go
floreks: I am creating this SA in a different namespace than the above. It is intentional. The thing is that this SA is for metrics scraper deployment that has to run inside the user cluster. That's why I am creating SA and granting it privileges directly in the user cluster. Not on the seed.

api/pkg/controller/usercluster/resources/kubernetes-dashboard/namespace.go
floreks: The main Dashboard deployment can run inside the seed cluster, but it requires secrets/configmap to be created inside the cluster it monitors. It can only connect to a single apiserver. That's why the binary can run somewhere else, but required resources have to exist inside the cluster.

api/pkg/controller/usercluster/resources/kubernetes-dashboard/secret.go
floreks: We can't do that. Dashboard can't connect to 2 different clusters at the same time. It acts as a proxy for a single apiserver.

api/pkg/resources/kubernetes-dashboard/deployment.go
floreks: I wasn't aware we have some utils for that. Good tip. Thanks.

api/pkg/controller/usercluster/resources/metrics-scraper/clusterrole.go
floreks: I guess we can merge them into a single package.

.gitignore
floreks: Sure, there might be some cases where dependency version differences might cause some unexpected behavior of the application. It's just easier to configure IntelliJ to run everything this way as it does it by default now. I am too lazy to reconfigure it :slightly_smiling_face:

api/pkg/controller/usercluster/reconciler.go
alvaroaleman: Where is this metrics scraper deployment and why does it have to run in the usercluster? All usercluster apis, kubelet endpoint etc work perfectly well from the seed

floreks: Metrics scraper is used by the Dashboard to gather metrics. It scrapes the metrics server and stores metrics from last X minutes. This way we can show graphs with actual resource consumption change over time, not only current usage. As Dashboard monitors the user cluster, it looks for the service that points to the metrics scraper there, thus it has to be deployed inside the cluster.

alvaroaleman: I still don't understand the reasoning, what is the problem with deploying it into the seed, it can still look into the usercluster to find the metrics?

The reason that I find this isubpar is that the metrics-server actually runs in the seed, so for dashboard metrics the datapath is now:

Dashboard in seed -> VPN into usercluster onto the metrics scraper -> VPN from the usercluster back into the seed onto the metrics server

floreks: The thing is that Dashboard automatically discovers metrics scraper inside the cluster it monitors and uses it to get metrics. Metrics scraper periodically gathers and stores metrics inside its own database. This way there is async communication between both dashboard -> metrics scraper and metrics scraper -> metrics server. Request to the scraper does not trigger request back to the metrics server.

We can think about supporting metrics scraper to run outside of the cluster it monitors and allowing Dashboard to connect to it, right now it doesn't really work.

floreks:
@alvaroaleman can we simply remove dashboard addon now? Are there any additional steps to consider here?

alvaroaleman:
> @alvaroaleman can we simply remove dashboard addon now? Are there any additional steps to consider here?

Aside from deleting it you also have to remove it from the default addons in cmd/kubermatic-controller-manager and config/kubermatic/values.yaml

But AFAIK we have a button in the dashboard that assumes the addon exists?

floreks:
> But AFAIK we have a button in the dashboard that assumes the addon exists?

We'll do the switch once it is up and running then.

PS. I am starting to think that I haven't updated something as coreos tests seem to be working fine in other PRs. Any ideas?

alvaroaleman:
Yeah pretty sure there is an actual issue, as there are four tests now on the same commit that failed on the same step. Looks a bit like somehow the dashboard pods don't get ready

api/pkg/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,140 @@
...
+
+			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
+				Labels: podLabels,
+			}
+
+			dep.Spec.Template.Spec.Volumes = volumes
+			dep.Spec.Template.Spec.Containers = getContainers(data)
+			dep.Spec.Template.Spec.Affinity = resources.HostnameAntiAffinity(name, data.Cluster().Name)
+
+			wrappedPodSpec, err := apiserver.IsRunningWrapper(data, dep.Spec.Template.Spec, sets.NewString(name), "", "")
```
alvaroaleman: The two trailing empty string args are wrong (the signature of `IsRunningWrapper` was changed in the meantime in a from the compilers POV compatible but actually not compatible way :grimacing: ). Just remove them, and tests should pass. right now this initcontainer makes the dashboard fail: `{"level":"fatal","time":"2019-10-02T14:22:15.825Z","logger":"http-prober","caller":"http-prober/main.go:65","msg":"--crd-to-wait-for was set, but KUBECONFIG env var was not"}`

floreks: Nice catch. Thanks 🙂

floreks:
PTAL

api/pkg/controller/usercluster/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,100 @@
...
+	return func() (string, reconciling.DeploymentCreator) {
+		return scraperName, func(dep *appsv1.Deployment) (*appsv1.Deployment, error) {
+			dep.Name = scraperName
+			dep.Labels = resources.BaseAppLabel(scraperName, nil)
+
+			dep.Spec.Replicas = resources.Int32(2)
+			dep.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: resources.BaseAppLabel(scraperName, nil),
+			}
+			dep.Spec.Template.Spec.ImagePullSecrets = []corev1.LocalObjectReference{{Name: resources.ImagePullSecretName}}
```
alvaroaleman: I don't think we need the image pull secret here and it doesn't even exist in the usercluster, so the pods here will never start

api/pkg/controller/usercluster/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+)
+
+const (
+	scraperName      = resources.MetricsScraperDeploymentName
+	scraperImageName = "kubernetesui/metrics-scraper"
+	scraperTag       = "v1.0.1"
+)
+
+// DeploymentCreator returns the function to create and update the metrics server deployment
```
alvaroaleman: `metrics scraper deployment` not `metrics-server` or not?

api/pkg/controller/usercluster/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,100 @@
...
+			dep.Spec.Template.ObjectMeta = metav1.ObjectMeta{
+				Labels: resources.BaseAppLabel(scraperName, nil),
+			}
+
+			volumes := getVolumes()
+			dep.Spec.Template.Spec.Volumes = volumes
+			dep.Spec.Template.Spec.Containers = getContainers()
+			dep.Spec.Template.Spec.ServiceAccountName = scraperName
+
+			return dep, nil
```
alvaroaleman, line 55: Would it make sense to add anti affinity so it gets spread out if possible to avoid the whole thing going down if one node goes down?

api/pkg/resources/kubernetes-dashboard/deployment.go
```diff
@@ -0,0 +1,140 @@
...
+	return func() (string, reconciling.DeploymentCreator) {
+		return name, func(dep *appsv1.Deployment) (*appsv1.Deployment, error) {
+			dep.Name = name
+			dep.Labels = resources.BaseAppLabel(name, nil)
+
+			dep.Spec.Replicas = resources.Int32(2)
+			dep.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: resources.BaseAppLabel(name, nil),
+			}
+			dep.Spec.Template.Spec.ImagePullSecrets = []corev1.LocalObjectReference{{Name: resources.ImagePullSecretName}}
```
alvaroaleman: I don't think this is needed, the image is public or not?

api/pkg/controller/usercluster/resources/kubernetes-dashboard/deployment.go
floreks: Scraper is not really a critical resource for the dashboard. It's not required to run the whole time. If it goes down for a while metrics will not be shown. When it goes up again they will reappear. We can add it, but it's not that important.

api/pkg/resources/kubernetes-dashboard/deployment.go
floreks: Images are public

api/pkg/controller/usercluster/resources/kubernetes-dashboard/deployment.go
floreks: We have to either keep that or remove this check https://github.com/kubermatic/kubermatic/blob/5bf0b79a82c5dc2e814c66ec95fcb03de1f99690/api/pkg/resources/test/load_files_test.go#L569-L572

floreks:
@alvaroaleman PTAL

PR #4418, opened 2019-10-02
Use own cache to use ctrlruntime's client before the manager was started

xrstf:
Since we changed the master-controller to only start the manager once the migrations are completed, the validating webhook for seeds was (because of the migrations) still started on its own, **before** the manager is started. This leads to the situation that the webhook does not find anything (unstarted managers have clients that return 404's for everything), rejecting any Seeds because it cannot find the kubeconfig secret a Seed must reference.

Since the webhook runs in all pods (even non-leaders), moving the initialization to a different place is confusing (see early commits in this PR). That's why the PR implements a hack based on creating a dedicated cache, warming it up and then injecting it into the manager (while making sure that the manager cannot start the cache *again*). The custom cache makes it possible to use the manager's client before the manager has been started, simplifying a bunch of code.

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -23,7 +23,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"sigs.k8s.io/controller-runtime/pkg/cache"
 	ctrlruntimecache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/lgtm
/approve

api/cmd/master-controller-manager/main.go
```diff
@@ -21,8 +21,10 @@ import (
 	seedvalidation "github.com/kubermatic/kubermatic/api/pkg/validation/seed"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
```
alvaroaleman: Via import twice?

api/cmd/master-controller-manager/main.go
```diff
@@ -120,12 +122,36 @@ func main() {
...
+		log.Fatalw("failed to create cache", zap.Error(err))
+	}
+
+	go func() {
+		if err := cache.Start(ctx.Done()); err != nil {
+			log.Fatalw("failed to start cache", zap.Error(err))
+		}
+	}()
+
+	cache.WaitForCacheSync(ctx.Done())
```
alvaroaleman: Please check the returned boolean, it indicates success or not

api/cmd/master-controller-manager/main.go
alvaroaleman: Why*

xrstf: I added some code, noticed that `cache` was also used as a local variable, changed the import to use an alias, then commented the new code out, VScode removed the new aliased-import, later I uncommented the code again, VScode added an import back and I fucked up somewhere.

PR #4422, opened 2019-10-02
Finish cloud credentials operator

alvaroaleman:

Finishes the work on the cloudcredentials operator to actually work on AWS. We need it to create a bucket for us which is used by the registry.

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/cmd/user-cluster-controller-manager/main.go
+++ b/api/cmd/user-cluster-controller-manager/main.go
@@ -77,7 +77,7 @@ func main() {
 	flag.BoolVar(&runOp.log.Debug, "log-debug", false, "Enables debug logging")
 	flag.StringVar(&runOp.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
 	flag.StringVar(&runOp.cloudProviderName, "cloud-provider-name", "", "Name of the cloudprovider")
-	flag.StringVar(&runOp.cloudCredentialSecretTemplate, "cloud-credential-secret-template", "", "A serialized Kubernetes secret whose Name and Data fields will be used to create a secret for the openshift cloud credentials opreator.")
+	flag.StringVar(&runOp.cloudCredentialSecretTemplate, "cloud-credential-secret-template", "", "A serialized Kubernetes secret whose Name and Data fields will be used to create a secret for the openshift cloud credentials operator.")
 
 	flag.Parse()
 
--- a/api/cmd/user-cluster-controller-manager/main.go
+++ b/api/cmd/user-cluster-controller-manager/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/ecdsa"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -29,6 +30,7 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
 
+	corev1 "k8s.io/api/core/v1"
 	apiextensionv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	certutil "k8s.io/client-go/util/cert"
@@ -41,20 +43,21 @@ import (
 )
 
 type controllerRunOptions struct {
-	metricsListenAddr     string
-	healthListenAddr      string
-	openshift             bool
-	version               string
-	networks              networkFlags
-	namespace             string
-	caPath                string
-	clusterURL            string
-	openvpnServerPort     int
-	openvpnCACertFilePath string
-	openvpnCAKeyFilePath  string
-	overwriteRegistry     string
-	cloudProviderName     string
-	log                   kubermaticlog.Options
+	metricsListenAddr             string
+	healthListenAddr              string
+	openshift                     bool
+	version                       string
+	networks                      networkFlags
+	namespace                     string
+	caPath                        string
+	clusterURL                    string
+	openvpnServerPort             int
+	openvpnCACertFilePath         string
+	openvpnCAKeyFilePath          string
+	overwriteRegistry             string
+	cloudProviderName             string
+	cloudCredentialSecretTemplate string
+	log                           kubermaticlog.Options
 }
 
 func main() {
@@ -74,6 +77,7 @@ func main() {
 	flag.BoolVar(&runOp.log.Debug, "log-debug", false, "Enables debug logging")
 	flag.StringVar(&runOp.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
 	flag.StringVar(&runOp.cloudProviderName, "cloud-provider-name", "", "Name of the cloudprovider")
+	flag.StringVar(&runOp.cloudCredentialSecretTemplate, "cloud-credential-secret-template", "", "A serialized Kubernetes secret whose Name and Data fields will be used to create a secret for the openshift cloud credentials opreator.")
 
 	flag.Parse()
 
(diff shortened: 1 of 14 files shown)


api/pkg/controller/openshift/data.go
```diff
@@ -343,3 +344,31 @@ func (od *openshiftData) SeedName() string {
 func (od *openshiftData) GetKubernetesCloudProviderName() string {
 	return kubernetesresources.GetKubernetesCloudProviderName(od.Cluster())
 }
+
+func (od *openshiftData) CloudCredentialSecretTemplate() ([]byte, error) {
```
alvaroaleman, line 348: So this means we put the AWS credentials inside the cluster which has the obvious implication that it can be read from there. In the Kubernetes context, this was explicitly not wanted. However here it is needed so the cloud-credentials-opreator can work. Opinions?

xrstf: I don't understand the issue enough to have a valid opinion.

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -74,6 +77,7 @@ func main() {
 	flag.BoolVar(&runOp.log.Debug, "log-debug", false, "Enables debug logging")
 	flag.StringVar(&runOp.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
 	flag.StringVar(&runOp.cloudProviderName, "cloud-provider-name", "", "Name of the cloudprovider")
+	flag.StringVar(&runOp.cloudCredentialSecretTemplate, "cloud-credential-secret-template", "", "A serialized Kubernetes secret whose Name and Data fields will be used to create a secret for the openshift cloud credentials opreator.")
```
thz: typo "opreator"

api/pkg/api/v1/types.go
```diff
@@ -1236,6 +1236,9 @@ const (
 	InClusterPVCleanupFinalizer = "kubermatic.io/cleanup-in-cluster-pv"
 	// InClusterLBCleanupFinalizer indicates that the LBs still need cleanup
 	InClusterLBCleanupFinalizer = "kubermatic.io/cleanup-in-cluster-lb"
+	// InClusterCredentialsRequestsCleanupFinalizer indicates that CredentialsRequests still need cleanup. This
+	// CRD only exists on Openshift and is a no-op for Kubernetes.
+	InClusterCredentialsRequestsCleanupFinalizer = "kubermatic-io/cleanup-credentials-requests"
```
thz: probably meant to be a dot in "kubermatic-io"?

api/pkg/clusterdeletion/credentialsrequests.go
```diff
@@ -0,0 +1,51 @@
...
+	"go.uber.org/zap"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (d *Deletion) cleanupCredentialsRequest(ctx context.Context, log *zap.SugaredLogger, cluster *kubermaticv1.Cluster) (deletedSomething bool, err error) {
```
thz: why are you using named return values and then passing values to every return? also some `err` here are shadowing the named return, right?

api/pkg/clusterdeletion/credentialsrequests.go
```diff
@@ -0,0 +1,51 @@
...
+	credentialRequests := &unstructured.UnstructuredList{}
+	credentialRequests.SetAPIVersion("cloudcredential.openshift.io/v1")
+	credentialRequests.SetKind("CredentialsRequest")
+
+	if err := userClusterClient.List(ctx, &ctrlruntimeclient.ListOptions{}, credentialRequests); err != nil {
+		if _, matches := err.(*meta.NoKindMatchError); matches {
+			log.Debug("Got a NoKindMatchError when listing CredentialsRequests, skipping their cleanup")
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to list CredentialsRequests: %v", err)
```
thz, line 33: All 4 possible variants of plural/singular for credential(s)requests(s) are present in this func. I do not get the semantics of every such distinction ;-)

api/pkg/clusterdeletion/deletion.go
```diff
@@ -66,10 +66,11 @@ func (d *Deletion) cleanupInClusterResources(ctx context.Context, log *zap.Sugar
 
 	shouldDeleteLBs := kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.InClusterLBCleanupFinalizer)
 	shouldDeletePVs := kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.InClusterPVCleanupFinalizer)
+	shouldDeleteCredentialsRequests := kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.InClusterCredentialsRequestsCleanupFinalizer)
 
 	// If no relevant finalizer exists, directly return
-	if !shouldDeleteLBs && !shouldDeletePVs {
-		log.Debug("Skipping in-cluster-resources deletion. Neither the LB nor the PV finalizers is set")
+	if !shouldDeleteLBs && !shouldDeletePVs && !shouldDeleteCredentialsRequests {
+		log.Debug("Skipping in-cluster-resources deletion. None iof the in-cluster cleanup finalizers is set.")
```
thz: typo "iof"

api/pkg/controller/usercluster/reconciler.go
```diff
@@ -322,6 +322,11 @@ func (r *reconciler) reconcileSecrets(ctx context.Context) error {
 	creators := []reconciling.NamedSecretCreatorGetter{
 		openvpn.ClientCertificate(r.openVPNCA),
 	}
+	if r.openshift {
+		if r.cloudCredentialSecretTemplate != nil {
+			creators = append(creators, openshift.CloudCredentialSecretCreator(*r.cloudCredentialSecretTemplate))
+		}
```
thz, line 328: So there are valid cases without a credential secret with openshift?
(Otherwise I would expect at least a warning here.)

api/pkg/controller/usercluster/resources/openshift/namespace.go
```diff
@@ -82,3 +82,23 @@ func NetworkOperatorNSGetter() (string, reconciling.NamespaceCreator) {
 		return ns, nil
 	}
 }
+
+// CloudCredentialOperatorNSGetter creates the namesapce in which all credentialsrequests end up
```
thz: typo "namesapce"

api/pkg/resources/usercluster/deployment.go
```diff
@@ -96,26 +97,36 @@ func DeploymentCreator(data userclusterControllerData, openshift bool) reconcili
...
+				fmt.Sprintf("-openvpn-ca-cert-file=%s/%s", openvpnCAMountDir, resources.OpenVPNCACertKey),
+				fmt.Sprintf("-openvpn-ca-key-file=%s/%s", openvpnCAMountDir, resources.OpenVPNCAKeyKey),
+			}, getNetworkArgs(data)...)
+
+			cloudCredentialSecretTemplate, err := data.CloudCredentialSecretTemplate()
+			if err != nil {
+				return nil, fmt.Errorf("failed to get cloud-credential-secret-template: %v", err)
+			}
+			if cloudCredentialSecretTemplate != nil {
+				args = append(args, "--cloud-credential-secret-template", string(cloudCredentialSecretTemplate))
```
thz: Is the inconsistency of having a double-dash option intended here? (The other flags have a single dash.)

api/pkg/clusterdeletion/credentialsrequests.go
alvaroaleman: The only purpose of the named return is to clarify the meaning of the returned boolean

api/pkg/controller/usercluster/reconciler.go
alvaroaleman: Yeah, the whole CredentialsOperator is only implemented for AWS in 4.1. This effectively means that the integrated registry and with that all the build features don't work on other platforms which is pretty subpar.

The only other backend supported is PV with RWX mode, which basically doesn't exist anywhere. We may have to discuss if we either deploy an RWX provisioner or maybe a minio to fake the S3 backend on other platforms.

I am not very fond of either of the two approaches. For now I just want to get the thing running on AWS, afterwards we can think about other platforms.

api/pkg/clusterdeletion/credentialsrequests.go
alvaroaleman: `CredentialsRequest` is the name of the `kind` and I used `CredentialsRequests` to refer to a plural of them.

thz: So all CredentialRequest(s) are unintended typos?

PR #4437, opened 2019-10-07
Synchronize usersshkeys into the userclusters

moadqassem:
This PR creates secrets out of the UserSSHKeys and inside the user cluster namespace and syncs these secrets inside the usercluster. The reason behind it is, we would like to add&remove SSH keys to a cluster & directly access all nodes via the newly added SSH key. Right now I'm required to rotate all nodes which impose a risk to my workload.

This PR is a subset of the required changes to fix this ticket: https://app.zenhub.com/workspaces/all-kubermatic-58863654794153994b62af76/issues/kubermatic/kubermatic/3573

No
 
```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
+++ b/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
@@ -1,0 +1,37 @@
+package usersshkeys
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CreateUserSSHKeysSecrets creates a secret in the usercluster from the user ssh keys.
+func CreateUserSSHKeysSecrets(path string) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return resources.UserSSHKeys, createSecretsFromUserSSHDirPath(path)
+	}
+}
+
+func createSecretsFromUserSSHDirPath(path string) reconciling.SecretCreator {
+	return func(existing *corev1.Secret) (secret *corev1.Secret, e error) {
+		files, err := ioutil.ReadDir(path)
+		if err != nil {
+			return nil, err
+		}
+
+		existing.Data = map[string][]byte{}
+		for _, file := range files {
+			data, err := ioutil.ReadFile(fmt.Sprintf("%v/%v", path, file.Name()))
+			if err != nil {
+				return nil, err
+			}
+			existing.Data[file.Name()] = data
+		}
+		return existing, nil
+	}
+}
--- a/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
+++ b/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
@@ -24,11 +24,18 @@ func createSecretsFromUserSSHDirPath(path string) reconciling.SecretCreator {
 			return nil, err
 		}
 
-		existing.Data = map[string][]byte{}
+		if existing.Data == nil {
+			existing.Data = map[string][]byte{}
+		}
+
 		for _, file := range files {
+			if _, ok := existing.Data[file.Name()]; ok {
+				continue
+			}
+
 			data, err := ioutil.ReadFile(fmt.Sprintf("%v/%v", path, file.Name()))
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("failed to read file %v during secret creation", file.Name())
 			}
 			existing.Data[file.Name()] = data
 		}
--- a/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
+++ b/api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
@@ -1,57 +1,25 @@
 package usersshkeys
 
 import (
-	"fmt"
-	"io/ioutil"
-	"os"
-
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
 	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
 
 	corev1 "k8s.io/api/core/v1"
 )
 
-const (
-	secretsDirLink = "..data"
-)
-
 // CreateUserSSHKeysSecrets creates a secret in the usercluster from the user ssh keys.
-func CreateUserSSHKeysSecrets(path string) reconciling.NamedSecretCreatorGetter {
+func CreateUserSSHKeysSecrets(userSSHKeys map[string][]byte) reconciling.NamedSecretCreatorGetter {
 	return func() (string, reconciling.SecretCreator) {
-		return resources.UserSSHKeys, createSecretsFromUserSSHDirPath(path)
+		return resources.UserSSHKeys, createSecretsFromUserSSHDirPath(userSSHKeys)
 	}
 }
 
-func createSecretsFromUserSSHDirPath(path string) reconciling.SecretCreator {
+func createSecretsFromUserSSHDirPath(userSSHKeys map[string][]byte) reconciling.SecretCreator {
 	return func(existing *corev1.Secret) (secret *corev1.Secret, e error) {
-		secretsDir, err := os.Readlink(fmt.Sprintf("%v/%v", path, secretsDirLink))
-		if err != nil {
-			return nil, err
-		}
+		existing.Data = map[string][]byte{}
 
-		files, err := ioutil.ReadDir(fmt.Sprintf("%v/%v", path, secretsDir))
-		if err != nil {
-			return nil, err
-		}
-
-		if existing.Data == nil {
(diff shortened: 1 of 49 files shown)


api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
```diff
@@ -0,0 +1,37 @@
...
+func createSecretsFromUserSSHDirPath(path string) reconciling.SecretCreator {
+	return func(existing *corev1.Secret) (secret *corev1.Secret, e error) {
+		files, err := ioutil.ReadDir(path)
+		if err != nil {
+			return nil, err
+		}
+
+		existing.Data = map[string][]byte{}
+		for _, file := range files {
+			data, err := ioutil.ReadFile(fmt.Sprintf("%v/%v", path, file.Name()))
```
alvaroaleman: Since we re-create the userclustercontroller pod whenever this is updated, we should just read it once during start up (and consider an error there fatal)

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	if err := r.Get(ctx, request.NamespacedName, cluster); err != nil {
+		if kubeapierrors.IsNotFound(err) {
+			log.Debug("Could not find cluster")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != "" &&
+		cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
```
alvaroaleman: the first half of the condition is not correct, what you are currently doing results in the controller always reconciling clusters that have no worker-name label.

Just doing `cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName` works, becuase `r.workerName` will be `""` in case the arg was unset

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (r *Reconciler) reconcileCluster(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	if err := r.ensureUserSSHKeySecretCreation(ctx, cluster); err != nil {
+		return err
+	}
+
+	userSSHKeys := &kubermaticv1.UserSSHKeyList{}
+	if err := r.Client.List(ctx, &ctrlruntimeclient.ListOptions{}, userSSHKeys); err != nil {
+		if kubeapierrors.IsNotFound(err) {
```
alvaroaleman: This shouldn't happen, if there are no `userSSHKeys` you will simply get back a liste with zero `Items` but no error

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,188 @@
...
+			if clusterName == clusterID {
+				clusterKeys = append(clusterKeys, item)
+			}
+		}
+	}
+
+	return clusterKeys
+}
+
+func (r *Reconciler) ensureUserSSHKeySecretCreation(ctx context.Context, cluster *kubermaticv1.Cluster) error {
```
alvaroaleman: as mentioned when we talked, please use `github.com/kubermatic/kubermatic/api/pkg/resources/reconciling.ReconcileSecrets` to get a proper cachesafe upsert and avoid code duplication

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	c, err := controller.New(ControllerName, mgr, controller.Options{Reconciler: reconciler, MaxConcurrentReconciles: numWorkers})
+	if err != nil {
+		return err
+	}
+
+	if err := c.Watch(&source.Kind{Type: &kubermaticv1.UserSSHKey{}}, enqueueAllClusters(mgr.GetClient())); err != nil {
+		return fmt.Errorf("failed to create watcher for userSSHKey: %v", err)
+	}
```
alvaroaleman, line 61: You also need to watch `Secrets` using `controllerutil "github.com/kubermatic/kubermatic/api/pkg/controller/util.EnqueueClusterForNamespacedObject`

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,192 @@
...
+				clusterKeys = append(clusterKeys, item)
+			}
+		}
+	}
+
+	return clusterKeys
+}
+
+func (r *Reconciler) ensureUserSSHKeySecretCreation(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	secreteGetter := createUserSSHKeysSecrets()
```
alvaroaleman: Please also use the `createUserSSHKeysSecrets` function to fill the secret with data, instead of having one function that creates the second and then another one that gets and updated it

alvaroaleman: Right now you have:
* `ensureUserSSHKeySecretCreation` to create the secret
* then you get the suerclusterssh keys
* Then you update the secret in `reconcileClustersUserSSHKeys

Instead, please get the keys first, change `ensureUserSSHKeySecretCreation` to also fill in the data in the secret and remove `reconcileClustersUserSSHKeys`

moadqassem: addressed in this commit https://github.com/kubermatic/kubermatic/pull/4437/commits/75a47d7bd8fc1331e0f613133ac90f796ff4fdc7

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,160 @@
...
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, request.NamespacedName, cluster); err != nil {
+		if kubeapierrors.IsNotFound(err) {
+			log.Debug("Could not find cluster")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
```
alvaroaleman, line 85: This also needs to respect the `.Spec.Pause` setting

api/pkg/controller/usersshkeys/controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+		return clustersRequests
+	})}
+}
+
+// updateUserSSHKeysSecrets creates a secret in the seed cluster from the user ssh keys.
+func updateUserSSHKeysSecrets(list []kubermaticv1.UserSSHKey) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return resources.UserSSHKeys, func(existing *corev1.Secret) (secret *corev1.Secret, e error) {
+			if existing.Data == nil {
```
alvaroaleman: The `.Data` fields needs to be set to an empty `map[string][]byte{}` unconditionally, else we wont delete secrets that were removed from the cluster

api/pkg/handler/v1/cluster/cluster.go
```diff
@@ -574,7 +574,7 @@ func AssignSSHKeyEndpoint(sshKeyProvider provider.SSHKeyProvider, projectProvide
 				}
 			}
 			if !found {
-				return nil, fmt.Errorf("the given ssh key %s does not belong to the given project %s (%s)", req.KeyID, project.Spec.Name, project.Name)
+				return nil, fmt.Errorf(fmt.Sprintf("the given ssh key %s does not belong to the given project %s (%s)", req.KeyID, project.Spec.Name, project.Name))
```
alvaroaleman: What is this, can you remove it?

alvaroaleman (commented):
Mostly looks good, except for some details.

What is conceptionally wrong is the assumption, that there is exactly one cluster that is both a master and a seed, but I suggest we fix that in a follow-up.

api/pkg/handler/v1/cluster/cluster.go
moadqassem: Oh that's a leftover from refactoring this code because I changed the error from `fmt.Errorf()` to custom error `IsNotFound` which should returns 404 instead of 500. but then I reverted because there was some tests depending on 500 thus I reverted since it is not my cup of tea :)

api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
alvaroaleman: @moadqassem this is still open

api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	secretsDirLink = "..data"
```
alvaroaleman: The linter will complain about this I think, as its used

alvaroaleman:
/lgtm
/approve

Thanks for the work!

PR #4440, opened 2019-10-07
Debug log webhook

xrstf:
This adds a nil-pointer check (that should only ever be triggered if someone manually sends crap to the webhook) and some debug logging so we can hopefully figure out what's causing the current issues.

I have not tested the log line. :crossed_fingers: 

```release-note
NONE
```

### DIFF
--- a/config/kubermatic/templates/master-controller-manager-dep.yaml
+++ b/config/kubermatic/templates/master-controller-manager-dep.yaml
@@ -29,6 +29,7 @@ spec:
         - -internal-address=0.0.0.0:8085
         - -v=2
         - -logtostderr
+        - -log-debug=true
         - -kubeconfig=/opt/.kube/kubeconfig
         {{- if .Values.kubermatic.datacenters }}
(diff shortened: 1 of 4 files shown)


config/kubermatic/templates/master-controller-manager-dep.yaml
```diff
@@ -29,6 +29,7 @@ spec:
         - -internal-address=0.0.0.0:8085
         - -v=2
         - -logtostderr
+        - -log-debug=true
```
alvaroaleman: Are you sure this should be the default? This doesn't only affect the one log you added

xrstf: My thought was to let the debug log in place and disable the debug-logging later, but we can also do it on info level and change it to debug later.

alvaroaleman: I would more suggest to make it configurable in the chart, default to info and configure it to debug for us

PR #4446, opened 2019-10-08
Fix openshift controller cluster creation when running externally

alvaroaleman:

Since https://github.com/kubermatic/kubermatic/pull/4443 was merged, the openshift controller can not create the `admin-kubeconfig` secrtet anymore when running out of cluster, because in that case it requires the secret to exist to create it, which is a non-fulfillable dependency.

Fixes #

**Documentation**:

```release-note
none
```


xrstf:
Code looks good, except for the ugly error message checking. You wanted to improve this in the future, so

/lgtm

PR #4449, opened 2019-10-08
handler tests: properly fill ObjectMeta for test user objects

kdomanski:
Previously, converting API users to Kubernetes objects in our tests
did not fill out the `ObjectMeta` struct - therefore as far as
the fake client's object tracker is concerened, every user
was named "". As a consequence, a test couldn't add multiple user objects.

```release-note
NONE
```

/assign @maciaszczykm


kdomanski:
I need this fixed in order to be able to test datacenter filtering.

PR #4458, opened 2019-10-09
Keep reconciling unready clusters regardless of the concurrency limit

### LINKED ISSUE #4455: Concurrency limiting: Continue reconciling clusters that do not have the ClusterControllerFinishedUpdatingSuccessfully condition
**User Story**

In #4193 we introduced a feature that limits the number of concurrent reconciles in order to keep the load on the Seed cluster within a predictable window, especially during Kubermatic upgrades.

The way this is implemented is by using a condition `ClusterControllerFinishedUpdatingSuccessfully` on the cluster that gets set to true whenever all pods Statefulsets and Deployments in the cluster namespace have `.Spec.Replicas == .Status.UpdatedReplicas == .Status.CurrentReplicas == .Status.ReadyReplicas`.

We then check if a threshold of clusters do not have this condition set to true and if that is the case, we do not reconcile anything anymore.

The problem with this approach is that we don't get to the point of reconciling clusters that do not have this condition or updating the condition, effectively deadlocking ourselves.

**Acceptance criteria**

* Even when the concurrency limit is reached, all clusters that do not have the `ClusterControllerFinishedUpdatingSuccessfully` condition set continue to be reconciled

xrstf:
The concurrency limit for controllers lead to cases where more than N clusters were unready, which made the controllers simply stop working entirely. This meant that the controllers also would never add the condition to clusters that they are ready, so the situation never fixes itself.

This PR changes the logic so that unready clusters are not considered for the concurrency limit, so they will always be reconciled. We only want to limit *new* reconciles (e.g. when a cluster update is performed, we don't want to start updating all clusters at the same time). A new `ClusterAvailableForReconciling` helper takes care of checking clusters.

I also renamed the `ClusterConditionControllerFinishedUpdatingSuccessfully` constant to `SeedResourcesUpToDate`, because

* the constant is used not just by the cluster controller,
* the new name reflects that the constant is only representing *in-cluster* resources (control plane, Prometheus, ...), but not cloud provider resources.

Fixes #4455.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/util/util.go
+++ b/api/pkg/controller/util/util.go
@@ -72,9 +72,22 @@ func SupportsFailureDomainZoneAntiAffinity(ctx context.Context, client ctrlrunti
 	return len(nodeList.Items) != 0, nil
 }
 
-// ConcurrencyLimitReached checks all the clusters inside the seed cluster and checks for
-// ClusterConditionControllerFinishedUpdatingSuccessfully. If the count of clusters which doesn't have that condition
-// is more than the limit, then it will return true, which means the concurrency limit has reached.
+// ClusterAvailableForReconciling returns true if the given cluster can be reconciled. This is true if
+// the cluster does not yet have the SeedResourcesUpToDate condition or if the concurrency limit of the
+// controller is not yet reached. This ensures that not too many cluster updates are running at the same
+// time, but also makes sure that un-UpToDate clusters will continue to be reconciled.
+func ClusterAvailableForReconciling(ctx context.Context, client ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster, concurrencyLimit int) (bool, error) {
+	if !cluster.Status.HasConditionValue(kubermaticv1.SeedResourcesUpToDate, corev1.ConditionTrue) {
+		return true, nil
+	}
+
+	limitReached, err := ConcurrencyLimitReached(ctx, client, concurrencyLimit)
+	return !limitReached, err
+}
+
+// ConcurrencyLimitReached checks all the clusters inside the seed cluster and checks for the
+// SeedResourcesUpToDate condition. Returns true if the number of clusters without this condition
+// is equal or larger than the given limit.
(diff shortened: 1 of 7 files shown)


api/pkg/controller/util/util.go
```diff
@@ -72,9 +72,22 @@ func SupportsFailureDomainZoneAntiAffinity(ctx context.Context, client ctrlrunti
...
 }
 
-// ConcurrencyLimitReached checks all the clusters inside the seed cluster and checks for
-// ClusterConditionControllerFinishedUpdatingSuccessfully. If the count of clusters which doesn't have that condition
-// is more than the limit, then it will return true, which means the concurrency limit has reached.
+// ClusterAvailableForReconciling returns true if the given cluster can be reconciled. This is true if
+// the cluster does not yet have the SeedResourcesUpToDate condition or if the concurrency limit of the
+// controller is not yet reached. This ensures that not too many cluster updates are running at the same
+// time, but also makes sure that un-UpToDate clusters will continue to be reconciled.
+func ClusterAvailableForReconciling(ctx context.Context, client ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster, concurrencyLimit int) (bool, error) {
```
alvaroaleman, line 79: This is beautiful

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -180,8 +183,22 @@ type ClusterStatus struct {
 	CloudMigrationRevision int `json:"cloudMigrationRevision"`
 }
 
+// HasConditionValue returns true if the cluster status has the given condition with the given status.
+// It does not verify that the condition has been set by a certain Kubermatic version, it just checks
+// the existence.
+func (cs *ClusterStatus) HasConditionValue(conditionType ClusterConditionType, conditionStatus corev1.ConditionStatus) bool {
+	for _, clusterCondition := range cs.Conditions {
+		if clusterCondition.Type == conditionType &&
```
alvaroaleman: Nit: This results in some useless looping if the condition exists but has a different value, this can be avoided by doing a:

```
if clusterCondition.Type == conditionType {
  return clusterCondition.value == conditionStatus
}
```

PR #4459, opened 2019-10-09
Configure HTTP proxy for components running in the seed

### LINKED ISSUE #4399: configure proxy for control plane components in a seed
**User Story**
Components of the control plane, like the machine-controller or the controller-manager might need to issue requests to external destinations, e.g. cloud provider API endpoints. However, if they are located in a DC that doesn't normally have Internet access, then such requests must go over an HTTP proxy. For that purpose, the proxy must be configurable for this DC.

**Acceptance criteria**
- [ ] a new top-level field `ProxySettings` is added to the `Datacenter` struct and contains settings for both "https proxy" and "no proxy", analogous to the settings in `NodeSettings`
- [ ] if set, then they are applied as env vars (or command line options, if applicable) to the following control plane components:
  - machine-controller
  - controller-manager
  - usercluster-controller
  - apiserver
- [ ] they are also used as fallback for the node proxy settings, if the relevant values in `NodeSettings` are not specified

alvaroaleman:

* Introduces the `ProxySettings` struct into the api which holds proxy settings
* Extends the `SeedSpec` with an optional `ProxySettings` property
* Replaces the proxy settings in the `NodeSettings` with an embedded `ProxySettings`
* Adds a `SetDefaults` func for the seed that sets the `ProxySettings` of all Datacenters to whats configured on the `SeedSpec` if unset
* Calls the seeds `SetDefaults` in both the `SeedGetter` and `SeedsGetter`
* Uses the `ProxySettings` from the seed to set to populate proxy env vars in
    * Kube apiserver
    * Kube controller-manager
    * machine-controller
    * machine-controller-webhook
    * openshift kube apiserver
    * openshift kube controller-manager

Fixes #4399

**Documentation**:

```release-note
It is now possible to configure an http proxy on a Seed. This will result in the proxy being used for
all control plane pods in that seed that talk to a cloudprovider and for all machines in that Seed, unless its overriden on Datacenter level.
```


api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -169,12 +185,20 @@ type DatacenterSpecFake struct {
...
 
+// ProxySettings allow configuring a HTTP proxy for the controlplanes
+// and nodes
+type ProxySettings struct {
+	// If set, this proxy will be configured for both http and https.
+	HTTPProxy *string `json:"http_proxy,omitempty"`
+	// If set this will be set as NO_PROXY on the node
+	NoProxy *string `json:"no_proxy,omitempty"`
+}
+
```
xrstf, line 196: Should we extend the seed validation hook to prevent both values being set at the same time?

api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -32,6 +32,20 @@ type Seed struct {
 	Spec SeedSpec `json:"spec"`
 }
 
+func (s *Seed) SetDefaults() {
```
xrstf, line 35: Will we move this to the validation hook at some point?

api/pkg/provider/datacenter_test.go
```diff
@@ -161,7 +163,9 @@ func TestMigrateDatacenters(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
xrstf: Can we never do this and instead rely on the (IMHO) more common

```
for idx := range slice {
    slice[idx].DoStuff()
}
```

pattern? This seemingly useless `a := a` line is triggering me so hard...

api/pkg/provider/datacenter_test.go
```diff
@@ -175,3 +179,69 @@ func TestMigrateDatacenters(t *testing.T) {
...
+		t.Fatalf("failed calling seedGetter: %v", err)
+	}
+	if seed.Spec.Datacenters["a"].Node.ProxySettings.HTTPProxy == nil ||
+		*seed.Spec.Datacenters["a"].Node.ProxySettings.HTTPProxy != "seed-proxy" {
+		t.Errorf("expected the datacenters http proxy setting to get set but was %v",
+			seed.Spec.Datacenters["a"].Node.ProxySettings.HTTPProxy)
+	}
+}
+
+func TestSeedSGetterFactorySetsDefaults(t *testing.T) {
```
xrstf: TestSeedSGetter should probably have a lowercase s

api/pkg/resources/controllermanager/deployment.go
```diff
@@ -343,6 +348,7 @@ func GetEnvVars(data resources.CredentialsData) ([]corev1.EnvVar, error) {
 	if cluster.Spec.Cloud.GCP != nil {
 		vars = append(vars, corev1.EnvVar{Name: "GOOGLE_APPLICATION_CREDENTIALS", Value: "/etc/gcp/serviceAccount"})
 	}
+	vars = append(vars, resources.GetHTTPPRoxyEnvVarsFromSeed(data.Seed(), data.Cluster().Address.InternalName)...)
```
xrstf: "Get HTTPP Roxy" should probably have a lowercase r (`HTTPProxy`).

api/pkg/crd/kubermatic/v1/datacenter.go
xrstf: Nevermind, I understood the variables wrong.

api/pkg/crd/kubermatic/v1/datacenter.go
alvaroaleman: ValidationHooks can only act on writing requests, not on reading ones. So no, which is not quite ideal

We could copy it into the `Datacenters` with a webhook on create/update, but then the Datacenters proxy settings don't get updated when the Seed proxy settings are changed, which is also not great.

api/pkg/crd/kubermatic/v1/datacenter.go
alvaroaleman: Why should that not be allowed?

api/pkg/resources/resources.go
```diff
@@ -857,3 +857,40 @@ func GlobalSecretKeySelectorValueGetterFactory(ctx context.Context, client ctrlr
...
+			{
+				Name:  "https_proxy",
+				Value: *seed.Spec.ProxySettings.HTTPProxy,
+			},
+		}
+	}
+	noProxyValue := inClusterAPIServerURL + ","
+	if seed.Spec.ProxySettings.NoProxy != nil {
+		noProxyValue += *seed.Spec.ProxySettings.NoProxy
+	}
```
xrstf: Can you move the comma into the if statement? (`noProxyValue += "," + *seed....`)

api/pkg/crd/kubermatic/v1/datacenter.go
xrstf: Agreed.

PR #4461, opened 2019-10-09
get kubeconfig for the viewers

### LINKED ISSUE #2922: update "...{cluster_id}/kubeconfig" endpoint
**Description** At the moment the endpoint always generates the admin configuration file for the given cluster. We need to change the EP so that:

1. only `owners` and `editors` are allowed to generate the admin config.
   There are no RBAC as we don't store config files in the database, thus we need to simply provide a check in the code (if group == onwers)

zreigz:
**What this PR does / why we need it**: restrict master kubeconfig for owners and editors

Fixes #2922

```release-note
NONE
```


alvaroaleman:
This means that `Viewers` can not connect to the usercluster at all anymore. Is this intended? Shouldn't we instead create a second kubeconfig for the `Vieweres` that has less permissions?

CC @kdomanski

zreigz:
> This means that `Viewers` can not connect to the usercluster at all anymore. Is this intended? Shouldn't we instead create a second kubeconfig for the `Vieweres` that has less permissions?
> 
> CC @kdomanski

I got feedback from @toschneck about this. Clients complain that everyone can get access to the cluster right now. I agree with the solution to return new kubeconfig for the viewers only. It sounds better than restrict it at all.
@kdomanski and @toschneck WDYT?

kdomanski:
Seems like rushing into the topic that's supposed to be covered by https://docs.google.com/document/d/1cUNBFH7K2pkUlrLmljKr6SfYULopPM9eugPLdWtrI5Q

api/pkg/controller/cluster/resources.go
```diff
@@ -119,6 +123,22 @@ func (r *Reconciler) getClusterTemplateData(ctx context.Context, cluster *kuberm
 		return nil, err
 	}
 
+	viewerTokenSecret := &corev1.Secret{}
```
alvaroaleman: Instead of having a complete random secret created here, can we:

* Have a `reconciling.NamedSecretCreatorGetter` for the viewer
* make it check if the token is already set and only if not create a new one
* Call it _before_ we call the `TokenUsersCreator`
* change `TemplateData.GetViewerToken` to get the secret and read the value
* Inside the `TokenUserCreator` just read the secret for the viewer token and copy its value

api/pkg/handler/v1/cluster/kubeconfig.go
```diff
@@ -44,11 +46,23 @@ func GetAdminKubeconfigEndpoint(projectProvider provider.ProjectProvider) endpoi
...
+		filePrefix := "admin"
+		var adminClientCfg *clientcmdapi.Config
+		if strings.HasPrefix(userInfo.Group, "viewers") {
+			filePrefix = "viewer"
+			adminClientCfg, err = clusterProvider.GetViewerKubeconfigForCustomerCluster(cluster)
+			if err != nil {
+				return nil, common.KubernetesErrorToHTTPError(err)
+			}
+		} else {
+			adminClientCfg, err = clusterProvider.GetAdminKubeconfigForCustomerCluster(cluster)
```
alvaroaleman, line 56: Nit: You can de-duplicate the error handling by moving it out of the `if-else`

api/pkg/cluster/client/client.go
```diff
@@ -73,6 +75,34 @@ func (p *provider) GetAdminKubeconfig(c *kubermaticv1.Cluster) ([]byte, error) {
 	return d, nil
 }
 
+// GetViewerKubeconfig returns the viewer kubeconfig for the given cluster
+func (p *provider) GetViewerKubeconfig(c *kubermaticv1.Cluster) ([]byte, error) {
```
alvaroaleman, line 80: Can we make this return an `error not implemented` for openshift clusters?

I didn't manage to get the whole `--static-token` auth stuff working with openshift but still have some hope It is possible. In the meantime thought its IMHO better to just straight away tell the user that this can't work

api/pkg/cluster/client/client.go
```diff
@@ -73,6 +75,34 @@ func (p *provider) GetAdminKubeconfig(c *kubermaticv1.Cluster) ([]byte, error) {
...
+}
+
+// RevokeViewerKubeconfig deletes viewer token to deploy new one and regenerate viewer-kubeconfig
+func (p *provider) RevokeViewerKubeconfig(c *kubermaticv1.Cluster) error {
+	s := &corev1.Secret{}
+
+	if err := p.seedClient.Get(context.Background(), types.NamespacedName{Namespace: c.Status.NamespaceName, Name: resources.ViewerTokenSecretName}, s); err != nil {
+		return err
+	}
+	if err := p.seedClient.Delete(context.Background(), s); err != nil {
```
alvaroaleman, line 103: Nit: You can just put the namespace and the name into the secret and directly call Delete without the Get. Also, maybe make this tolerate `kerrors.IsNotFound`?

api/pkg/controller/cluster/resources.go
zreigz: done

api/pkg/handler/v1/cluster/kubeconfig.go
zreigz: done

api/pkg/cluster/client/client.go
zreigz: done

api/pkg/cluster/client/client.go
zreigz: done

api/pkg/handler/routes_v1.go
```diff
@@ -313,6 +313,10 @@ func (r Routing) RegisterV1(mux *mux.Router, metrics common.ServerMetrics) {
 		Path("/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/token").
 		Handler(r.revokeClusterAdminToken())
 
+	mux.Methods(http.MethodPut).
+		Path("/projects/{project_id}/dc/{dc}/clusters/{cluster_id}/viewertoken").
```
alvaroaleman, line 317: Does it make sense from an UX perspective to have a distinct endpoint for this as opposed to infer which kubeconfig to revoke based on the user group?

api/pkg/resources/apiserver/token-users.go
```diff
@@ -20,17 +22,44 @@ func TokenUsersCreator(data *resources.TemplateData) reconciling.NamedSecretCrea
...
+
+// TokenViewerCreator returns a secret containing the viewer token
+func TokenViewerCreator(data *resources.TemplateData) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return resources.ViewerTokenSecretName, func(se *corev1.Secret) (*corev1.Secret, error) {
+			if se.Data == nil {
+				se.Data = map[string][]byte{}
+			}
 
+			viewerToken, err := data.GetViewerToken()
```
alvaroaleman: While this works its a bit confusing, `GetViewerToken` just gets the very same secret this func gets as `se` arg. Just checking if `se.Data[ViewerTokenSecretKey]` is set would achieve the very same result and be much easier to understand

api/pkg/resources/data.go
```diff
@@ -94,6 +94,20 @@ func NewTemplateData(
...
+func (d *TemplateData) GetViewerToken() (string, error) {
+	viewerTokenSecret := &corev1.Secret{}
+	if err := d.client.Get(d.ctx, ctrlruntimeclient.ObjectKey{Name: ViewerTokenSecretName, Namespace: d.cluster.Status.NamespaceName}, viewerTokenSecret); err != nil {
+		return "", err
+	}
+	return string(viewerTokenSecret.Data[ViewerTokenSecretKey]), nil
+}
+
+// GetClient returns the client
+func (d *TemplateData) GetClient() ctrlruntimeclient.Client {
```
alvaroaleman: This isn't needed

api/pkg/resources/kubeconfig.go
```diff
@@ -58,6 +58,46 @@ func AdminKubeconfigCreator(data adminKubeconfigCreatorData, modifier ...func(*c
 	}
 }
 
+// ViewerKubeconfigCreator returns a function to create/update the secret with the viewer kubeconfig
+func ViewerKubeconfigCreator(data *TemplateData, modifier ...func(*clientcmdapi.Config)) reconciling.NamedSecretCreatorGetter {
```
alvaroaleman: The variadic `modifier` arg is a remainder or not? Its not used anywhere

api/pkg/provider/kubernetes/cluster.go
```diff
@@ -229,6 +231,21 @@ func (p *ClusterProvider) GetAdminKubeconfigForCustomerCluster(c *kubermaticv1.C
 	return clientcmd.Load(b)
 }
 
+// GetViewerKubeconfigForCustomerCluster returns the viewer kubeconfig for the given cluster
+func (p *ClusterProvider) GetViewerKubeconfigForCustomerCluster(c *kubermaticv1.Cluster) (*clientcmdapi.Config, error) {
+	b, err := p.userClusterConnProvider.GetViewerKubeconfig(c)
```
alvaroaleman, line 240: Can we return an ErrNotImplemented here for openshift clusters as this doesn't work for Openshift?

alvaroaleman (commented):
Mostly looks good, when I tried to try it out though the UI showed the "connect" button grayed out, do you know why?

api/pkg/handler/routes_v1.go
zreigz: it doesn't make sense when we enable it for the viewers, right now it's disabled. Only owners/editors can make some operations on the cluster. We can enable it for all groups in the UI  and stay with one endpoint. WDYT?

api/pkg/resources/apiserver/token-users.go
zreigz: yes, you are right, thanks

api/pkg/resources/data.go
zreigz: deleted

api/pkg/resources/kubeconfig.go
zreigz: removed

api/pkg/provider/kubernetes/cluster.go
zreigz: done

api/pkg/handler/routes_v1.go
alvaroaleman: I am not sure, lets keep it as-is for now

alvaroaleman (commented):
> Mostly looks good, when I tried to try it out though the UI showed the "connect" button grayed out, do you know why?

Is the UI just disabling the button? Code looks good from my side but I cant test if it actually works

zreigz:
@alvaroaleman I've tested everything locally and works fine

zreigz:
@alvaroaleman thanks a lot for the review :+1:

PR #4466, opened 2019-10-10
Proposal: Allow multiple idp for share cluster

chrisz100:
Proposal to allow multiple identity providers for share cluster functionality. 

Release notes:
```release-note
NONE
```

### DIFF
--- a/docs/proposals/multi-idp.md
+++ b/docs/proposals/multi-idp.md
@@ -1,0 +1,54 @@
+# Multi-IDP for share cluster
+
+Multi-tenancy with third party identity providers for kubermatic managed clusters.
+
+## Requirements
+
+- [https://docs.kubermatic.io/advanced/oidc_auth/]()
+- Multi-Tenant setup
+
+## Case
+
+When using kubermatic to provide a multi-customer environment with clusters
+they usually do have their own IDP that should be connected to kubernetes. 
+
+As of now the only two  authentication options in kubermatic managed clusters are:
+
+- admin token
+- central idp / oidc provider
+
+These two options are great for single-tenant installations but provide additional
+risk and complexity in multi-tenant environments. 
+
+As of now every user in kubermatic would be able to authenticate against any cluster
+if endpoints are known and a custom kubeconfig is created. They would not be able to
+do things as auth-z is not configured for the user.
+
+From a security standpoint this works - but as soon as customer start working with groups
+there is intermediate to high risk of people authorizing people from unrelated companies.
+Group names like `administrator` is just to common in current IT. 
+
+## Proposal
+
+When creating a cluster users are able to also provide endpoints, secrets and
+attribute information. This will then be configured in the cluster. 
+
+Additionally this will be used for the share cluster URL the oidc auth functionality
+already provides.
+
+This way people have the option to bring their own IDP, hosting companies are
+able of providing multiple authentication realms for different corporate customers
+but still keeping setups maintainable and in an unpaused state.
+
+## Implementation
+
+OIDC providers already can be configured for customer clusters in kubermatic.
+This is just to be used as share cluster is already doing.
+
+Share cluster functionality needs to be extended to allow kubermatic users to set
+custom oidc provider settings.
+
+Share cluster functionality needs to be extended to authenticate against third party IDP.
+
+Share cluster functionality needs to be extended to allow setting properties like username
+and group fields from the to be retrieved token.


docs/proposals/multi-idp.md
```diff
@@ -0,0 +1,54 @@
...
+These two options are great for single-tenant installations but provide additional
+risk and complexity in multi-tenant environments. 
+
+As of now every user in kubermatic would be able to authenticate against any cluster
+if endpoints are known and a custom kubeconfig is created. They would not be able to
+do things as auth-z is not configured for the user.
+
+From a security standpoint this works - but as soon as customer start working with groups
+there is intermediate to high risk of people authorizing people from unrelated companies.
+Group names like `administrator` is just to common in current IT.
```
alvaroaleman, line 29: Not sure I follow the example, can you elaborate?

docs/proposals/multi-idp.md
```diff
@@ -0,0 +1,54 @@
...
+do things as auth-z is not configured for the user.
+
+From a security standpoint this works - but as soon as customer start working with groups
+there is intermediate to high risk of people authorizing people from unrelated companies.
+Group names like `administrator` is just to common in current IT. 
+
+## Proposal
+
+When creating a cluster users are able to also provide endpoints, secrets and
+attribute information. This will then be configured in the cluster.
```
alvaroaleman, line 34: For an OIDC provider or what do you mean by `endpoints, secrets and attribute information`?

docs/proposals/multi-idp.md
```diff
@@ -0,0 +1,54 @@
...
+there is intermediate to high risk of people authorizing people from unrelated companies.
+Group names like `administrator` is just to common in current IT. 
+
+## Proposal
+
+When creating a cluster users are able to also provide endpoints, secrets and
+attribute information. This will then be configured in the cluster. 
+
+Additionally this will be used for the share cluster URL the oidc auth functionality
+already provides.
```
alvaroaleman, line 37: So the way this currently works is that our API identifies the user via the global IDP and then uses this global IDP to issue an OIDC-based kubeconfig. IIUC, the proposal is to basically make the kubermatic API identify the user against an arbitrary, user-provided IDP and use that to generate the kubeconfig?

alvaroaleman (commented):
CC @kdomanski

docs/proposals/multi-idp.md
chrisz100: Correct. Even if it’s in the same keycloak, we want to give each customer their own realm they can use to manage their own identities. Thus the kubeconfig needs to map against different keycloak realms or even oidc providers for different clusters

docs/proposals/multi-idp.md
chrisz100: Oidc endpoint, oidc client secret and what attribute contains the username and which contains a list of groups 

Flags: 
--oidc-username-claim
--oidc-groups-claim

docs/proposals/multi-idp.md
chrisz100: Let’s say we have 2 customers: 
- megacorp 
- evilcorp 

Both of them have groups in their system and map users to groups. 

Both have a group called admin 

Users in group admin from evilcorp should not be able to get access to clusters from megacorp if megacorp decides to create a clusterrolebinding to group „admin“

alvaroaleman: This is true going by the assumption that there is an identity broker configured as IDP for kubermatic and via the share cluster feature then on the clusters and both megacorp and evilcorp can add their IDP to the identity broker used for Kubermatic. Do you (plan to) do this? 

Wouldn't it be simpler to just have the customers configure their IDP directly on the cluster and not having a broker involved?

chrisz100: As of now this would simply happen as you can only configure one global idp if you want to use the share cluster feature. As Lukasz wrote on slack - right now you have to decide to either use the same idp as kubermatic uses and be able to use share cluster or configure your own and be unable to use share cluster 

We need the share cluster as it needs to be easy and also usable with our webterminal in Imke. Adding their own IDP directly would disallow us from using this feature. 

This proposal basically is: 
Merging share cluster and configure a custom idp for the cluster

dionb:
I think the title for this PR may have been misleading. "Merging share cluster and configure a custom IDP for the cluster" is the best way to put what we need. We are trying to make it more secure, compliant, and user-friendly to grant access to a cluster to a group of people from a separate company.

If our customers are wanting to delegate IDP access for their contractors we would federate it through an IDP we set up (or an existing one the customer has) so that we can control group claims from the contractors' identity solution. We don't need multiple IDPs per cluster, we need the two existing(incoming) features to work together. And the thought of it being 'multiple IDPs' is that we want kubermatic to use one IDP and each cluster to be able to have it's own (which you currently support), but the 'share cluster' button needs to be aware of the cluster-specific IDP and to use that one rather than the one that is configured globally. This way we can configure either a separate IDP per cluster, or the same IDP, but with a different realm depending on what our customers need. 
But it should be noted that we want the 'download kubeconfig' to continue working as-is so that our customers can still access the clusters in the same way when delegating the OIDC access to a contractor's IDP.

Does this help to clarify things?

chrisz100:
Is there any input on this from loodse side?

alvaroaleman:
> Is there any input on this from loodse side?

Please reword the title and content of the proposal to reflect that this is about re-using the per-usercluster IDP for the share cluster functionality, this is not completely obvious currently.

Other than that we have to check what implications it has to make our API call into abitrary user-provided IDPs. CC @zreigz

xrstf:
No discussion for over a year, closing until this might get on the agenda again.

PR #4467, opened 2019-10-10
fix a panic on missing DigitalOcean credentials

kdomanski:
```release-note
NONE
```


api/pkg/validation/cluster.go
```diff
@@ -386,6 +386,10 @@ func validateAzureCloudSpec(spec *kubermaticv1.AzureCloudSpec) error {
 
 func validateDigitaloceanCloudSpec(spec *kubermaticv1.DigitaloceanCloudSpec) error {
 	if spec.Token == "" {
```
alvaroaleman, line 388: Wouldn't a `if spec.Token == "" && spec.CredentialsReference == nil {` make more sense?

kdomanski: It's perfectly equivalent.

alvaroaleman (commented):
/lgtm
/approve

Issue #1782, opened 2018-08-16
[Grafana/Prom Customizing] Allow the use of grafana plugins


pkavajin:
It would be good to have the possibility to install grafana plugins, which could be maybe done with such an init container for example:

```
      {{- if .Values.grafana.plugins }}
      initContainers:
      - name: install-plugins
        image: {{ .Values.grafana.image.repository }}:{{ .Values.grafana.image.tag }}
        env:
        - name: GF_PLUGIN_DIR
          value: /data/plugins
        command:
          - /bin/sh
          - -c
          - |
            mkdir -p /data/plugins
            {{- range .Values.grafana.plugins }}
            ./bin/grafana-cli plugins install {{ . }}
            {{- end }}
        volumeMounts:
        - mountPath: /data
          name: grafana-storage
          readOnly: false
      {{- end }}
```

Issue extracted from https://github.com/kubermatic/kubermatic/pull/1775#issuecomment-413459153

pkavajin:
Blocked by https://github.com/grafana/grafana/issues/11409
Depending on the customers use case we could build some workarounds, but thats TBD.

PR #4469, opened 2019-10-10
WIP - Precompress Grafana dashboards to reduce Tiller's ConfigMap size

xrstf:
Helm stores not just the rendered templates, but also the entire build context in a ConfigMap per revision. So not only would our rendered dashboard ConfigMap be included, the raw dashboard files are also stored in Tiller's history.
Because of this we are now randomly experiencing issues updating the chart because of Kubernetes 1MB limit for ConfigMaps.

This PR introduces a pre-compressed .tar.bz2 file, containing all the dashboards. This archive is then extracted in an initContainer. This still means that we have the same tar.bz2 file twice in Tiller's history, but it's "only" 37K x 2 = roughly 80K.
Pre-compressing the dashboards unfortunately means that we cannot template them during the Helm installation anymore. I think this is something rarely anyone used ever and we can safely remove it. This will get rid of the horrible templating engine (that I invented myself) :-)

A script in `api/hack` is meant to be a tiny Prow job in the future, ensuring that the tar.bz2 is always in-sync with the actual dashboards.

This PR also includes an option to dynamically install plugins for Grafana. This is something we wanted to have for aaaaaaaaages (#1782), but now it's apparently possible to just clone the plugin somewhere and then start Grafana. Neat! As plugins would also blow up the chart size, we will most likely ship a predefined set of plugins as a dedicated Docker image, but this is TBD. For now the installation is dynamic and optional so we can play around with some super fancy plugins before we actually ship them to our customers.

The most confusing change of this PR is that some ConfigMaps and volume mounts have been slightly renamed. Don't despair, I've deployed the new chart and it all seems to work just fine. :-D

The plugin stuff is still considered beta, so I would suggest adding a release note once it's actually usable for customers.

```release-note
NONE
```


kron4eg (approved):
/lgtm

config/monitoring/grafana/templates/configmaps.yaml
```diff
@@ -39,10 +49,11 @@ data:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: grafana-dashboard-definitions
-data:
-  # See the README.md for more details on how dashboards are treated as templates.
-  {{- range $file, $content := (.Files.Glob "dashboards/**") }}
-  {{ $file | replace "dashboards/" "" | replace "/" "-" }}: |
-{{ ((tpl (($.Files.Get $file) | replace "{{" "_{_{" | replace "}}" "_}_}" | replace "[[ " "{{ .Values.grafana.dashboards." | replace "]]" "}}" | replace "\"<< " "{{ .Values.grafana.dashboards." | replace ">>\"" "}}") $) | replace "_{_{" "{{" | replace "_}_}" "}}") | fromJson | toJson | indent 4 }}
```
xrstf, line 47: Good riddance.

xrstf:
This has been split into #4475 and #4480. The dashboard compression has been removed entirely because it was wrongly triggered by me having a Grafana plugin cloned into my local Grafana chart. Tiller's ConfigMap on cloud is 110K, on dev it was 100K (after my patches). Saving 10K is not enough to justify the added complexity (and when we have roughly 900K to spare).

PR #4475, opened 2019-10-11
Verify Grafana dashboards

xrstf:
This removes the ugly templating engine we had in our dashboards, as nobody was really using it and it made things much more complicated. This PR also moves the cleanup script to the global hack directory, under the motivation that devs should not search for scripts at random places in our repository.
A new verification script has been added that will hopefully soon be run as a Prow job, ensuring that our dashboards are always properly formatted.

```release-note
Removed Helm-based templating in Grafana dashboards
```


kron4eg (approved):
/lgtm

PR #4477, opened 2019-10-11
Fix concurrency in TestMigrateDatacenters

alvaroaleman:

Fixes `TestMigrateDatacenters` to properly run concurrently.

Fixes #

**Documentation**:

```release-note
none
```

/assign @xrstf

### DIFF
--- a/api/pkg/provider/datacenter_test.go
+++ b/api/pkg/provider/datacenter_test.go
@@ -162,17 +162,18 @@ func TestMigrateDatacenters(t *testing.T) {
 		},
 	}
 
-	for idx := range testCases {
-		t.Run(testCases[idx].name, func(t *testing.T) {
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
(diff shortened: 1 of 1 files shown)


api/pkg/provider/datacenter_test.go
```diff
@@ -162,17 +162,18 @@ func TestMigrateDatacenters(t *testing.T) {
 		},
 	}
 
-	for idx := range testCases {
+	for _, tc := range testCases {
+		tc := tc
```
alvaroaleman, line 166: the `idx` var has the same issue as the other one: It may have a different value by the time the goroutine is started

xrstf: Sad Panda

api/pkg/provider/datacenter_test.go
```diff
@@ -162,17 +162,18 @@ func TestMigrateDatacenters(t *testing.T) {
...
 		t.Run(testCases[idx].name, func(t *testing.T) {
```
xrstf: This needs to be `tc.name` now.

alvaroaleman: Oups. Fixed

PR #4480, opened 2019-10-11
Grafana Plugins & Cluster Dashboard

### LINKED ISSUE #1782: [Grafana/Prom Customizing] Allow the use of grafana plugins

xrstf:
This PR introduces a new public Docker image for carrying Grafana plugins, `quay.io/kubermatic/grafana-plugins`. We use some of the new plugins in the new, fancy User Cluster dashboard already:

screenshot-2019-10-11-142255

This PR also moves the NGINX dashboard to a better place and makes the dashboard verification job stricter by not just requiring properly formatted JSON, but also that all the other normalizations have been applied.

Fixes #1782 (mostly; not all plugins can be installed this way, but this is an 80/20 solution to the problem and customers can easily create their own Docker image with plugins).

```release-note
NONE
```

### DIFF
--- a/containers/grafana-plugins/Dockerfile
+++ b/containers/grafana-plugins/Dockerfile
@@ -1,0 +1,16 @@
+FROM alpine:3.10
+
+RUN mkdir -p /plugins
+WORKDIR /plugins
+
+RUN wget -O- https://grafana.com/api/plugins/grafana-piechart-panel/versions/1.3.9/download | unzip - && \
+    mv grafana-piechart-panel-* grafana-piechart-panel
+
+RUN wget -O- https://grafana.com/api/plugins/farski-blendstat-panel/versions/1.0.1/download | unzip - && \
+    mv farski-blendstat-grafana-* farski-blendstat-panel
+
+RUN wget -O- https://grafana.com/api/plugins/michaeldmoore-multistat-panel/versions/1.2.3/download | unzip - && \
+    mv michaeldmoore-michaeldmoore-multistat-panel-* michaeldmoore-multistat-panel
+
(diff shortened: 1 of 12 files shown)


containers/grafana-plugins/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.10
+
+RUN mkdir -p /plugins
+WORKDIR /plugins
+
+RUN wget -O- https://grafana.com/api/plugins/grafana-piechart-panel/versions/1.3.9/download | unzip - && \
```
kron4eg, line 6: I'd parametrize version, with `ARG` instruction.

xrstf: I hate Docker args when building images. :-D I also don't see an advantage. Either we manage simple, working URLs inside the Dockerfile or split the stuff into two or more places.

kron4eg (commented):
/lgtm

PR #4482, opened 2019-10-11
Remove pointless backups for Prometheus/Elasticsearch/Minio

xrstf:
This PR removes the predefined backup schedules from Velero:

* Prometheus/Thanos is already redundant and using S3 as its storage. Creating copies from blocks in S3 and then storing them in the same S3 does not make sense, but requires a lot of storage and a lot of time.
* Same for Elasticsearch. It's already a clustering system with sharding. Also, backing up plain volumes via restic is probably never going to create consistent backups. Having a dedicated elasticsearch-exporter, that dumps indexes as JSON, is a much better solution for backing up ES.
* Backing up Minio into Minio is ... yeah. (If you're using Minio in the first place, you likely don't have an external object storage; so all your backups are probably also being stored in your local Minio.)

```release-note
ACTION REQUIRED: Velero does not backup Prometheus, Elasticsearch and Minio by default anymore.
```


kron4eg (approved):
/lgtm

PR #4485, opened 2019-10-14
Create example YAMLs from Go structs

xrstf:
This uses a [patched version](https://github.com/kubernetes/test-infra/pull/14751) from genyaml to generate a nice, commented YAML version for our Seed CRs. I made a couple of choices in the validation logic, which should hopefully make sense even when ignoring all the reflect gibberish.

```release-note
NONE
```


docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
+# Country of the seed. For informational purposes only
+country: foo
+
+# The possible datacenters for the nodes
+datacenters:
+    exampleseed1:
+        # Node holds node-specific settings, like e.g. HTTP proxy, insecure registries and the like
+        node:
+            http_proxy: external.com
```
alvaroaleman: Lets please for all settings whot their default, i.E. here an empty string

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+            no_proxy: localhost,internal.example.com
+        spec:
+            aws:
+                images:
+                    coreos: foo
```
alvaroaleman: We need this for all possible OSes, not only coreos

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+                region: ""
+            azure:
+                location: ""
+            bringyourown: {}
+            digitalocean:
+                region: ""
+            fake: {}
```
alvaroaleman: I don't think customers should see this

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+                # still use the credentials passed in via the frontend/api
+                infra_management_user: {}
+
+                # Optional root path for cluster specific VM folders.
+                # Each cluster gets a own folder below the root folder.
+                # Must be the FQDN (example: /datacenter-1/vm/all-kubermatic-vms-in-here )
+                # Defaults to the root VM folder: /datacenter-1/vm
+                root_path: ""
+                templates:
+                    coreos: foo
```
alvaroaleman: Same as on AWS; needs also Ubuntu and CentOS

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+# A reference to the Kubeconfig of this cluster
+kubeconfig: {}
```
alvaroaleman: This needs its properties

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+# Optional: Configure a http proxy for this seed
+proxy_settings:
+    # If set, this proxy will be configured for both http and https.
+    http_proxy: external.com
```
alvaroaleman: Same here, please show the default which is an empty string

docs/seed-cr.generated.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+    # If set this will be set as NO_PROXY on the node
+    no_proxy: localhost,internal.example.com
+
+# Optional: Overwrite the DNS domain for this seed
+seed_dns_overwrite: foo.example.com
```
alvaroaleman: Same here, please show the default which is an empty string

api/hack/verify-crd-examples.sh
```diff
@@ -0,0 +1,24 @@
...
+  rm -rf "$tmpdir"
+}
+trap "cleanup" EXIT SIGINT
+
+cleanup
+
+echodate "Recreating example CRs..."
+
+cp -ar docs $tmpdir
+rm -f docs/*.generated.yaml
```
alvaroaleman: Nit: We usually name our generated files `zz_generated_whatever.fileending`

api/codegen/seed-yaml/main.go
```diff
@@ -0,0 +1,178 @@
+package main
```
alvaroaleman: Not sure if this is the right location, as this doesn't generate any code

xrstf: I'm open for suggestions :)

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
+# Country of the seed. For informational purposes only
```
xrstf: Should we output a full resource manifest, i.e. with apiVersion, kind, metatdata etc.?

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
...
+      openstack:
+        auth_url: ""
+        availability_zone: ""
+
+        # Used for automatic network creation
+        dns_servers: []
+        enforce_floating_ip: false
+        ignore_volume_az: false
+        images:
+          coreos: foo
```
xrstf: Should we rely on the map being documented properly or do we want to keep my validation rule that every map must have at least one sample entry? We could make the same exception as for slices, i.e. only require one sample element if the map value type is a complex type (i.e. empty `map[string]string` are fine, but `map[string]Datacenter` would need at least one sample).

api/codegen/seed-yaml/main.go
alvaroaleman: I'd just put it unter `cmd/seed-yaml-generator` or something the like

api/codegen/seed-yaml/main.go
```diff
@@ -0,0 +1,182 @@
...
+	// find all .go files in kubermatic/v1, just in case a definition is split
+	// into multiple files
+	files, err := filepath.Glob("pkg/crd/kubermatic/v1/*.go")
+	if err != nil {
+		log.Fatalf("Failed to find go files: %v", err)
+	}
+
+	files = append(
+		files,
+		"vendor/k8s.io/api/core/v1/types.go",
```
alvaroaleman: Making sure this conforms to PEP8? :)

docs/zz_generated.seed-cr.yaml
alvaroaleman: Yeah, I think that would be great

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
...
+      # If set, this image registry will be configured as insecure on the container runtime.
+      insecure_registries: []
+      no_proxy: null
+
+      # Translates to --pod-infra-container-image on the kubelet. If not set, the kubelet will default it
+      pause_image: ""
+    spec:
+      aws:
+        images:
+          coreos: foo
```
alvaroaleman: Please do not put non-default values in this

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
...
+        region: ""
+      azure:
+        location: ""
+      bringyourown: {}
+      digitalocean:
+        region: ""
+      fake:
```
alvaroaleman: Can we get this out?

docs/zz_generated.seed-cr.yaml
alvaroaleman: IMHO all the distribution fields should be filled with all possible keys (ubuntu, centos, coreos) but no value

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
...
+        dns_servers: []
+        enforce_floating_ip: false
+        ignore_volume_az: false
+        images:
+          coreos: foo
+
+        # Gets mapped to the "manage-security-groups" setting in the cloud config.
+        # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
+        # To make this change backwards compatible, this will default to true.
+        manage_security_groups: null
```
alvaroaleman: This needs to be filed with a `false` bool ptr I guess

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,144 @@
+# Country of the seed. For informational purposes only
+country: DE
+
+# The possible datacenters for the nodes
+datacenters:
+  exampleseed1:
```
alvaroaleman: Can we put something in here that makes it obvious the `exampleseed1` should be replaced by the user, i.E. `<< my seed >>` (is that valid yaml)?

alvaroaleman (commented):
Some nits, mostlly looks good. We obviously need more comments on the types, but that is probably out of scope for this PR

docs/zz_generated.seed-cr.yaml
xrstf: That's quite a hefty requirement to find the map type, resolve it and find all constants for the given type, to make sure none are missed. And I would also avoid hardcoding this rule for imagelists only, if possible.

I guess we stay with the current logic (always require at least one element) and I will just add the currently existing values.

api/codegen/seed-yaml/main.go
xrstf: I don't quite understand? It might have been a been premature to start listing files in their own lines, or am I overlooking something? ^^

alvaroaleman: a tiny bit, yes

alvaroaleman: Pep8 is famous for its 80 char line length limit, I think it was designed to smoothing the experience on smarthphones :)

docs/zz_generated.seed-cr.yaml
xrstf: I can only make it be `fake: null`, but to completely remove it we would have to be smart about trimming the `,omitempty` tag and not remove that tag for the fake provider.

xrstf: Would it be enough to just add a comment in the Go code that the Fake provider is only meant for internal use?

docs/zz_generated.cluster.yaml
```diff
@@ -0,0 +1,128 @@
+address:
```
xrstf: We don't ever want to output a sample YAML with address and status fields, right?

api/cmd/example-yaml-generator/main.go
```diff
@@ -0,0 +1,244 @@
...
+func createExampleSeed() *kubermaticv1.Seed {
+	seed := &kubermaticv1.Seed{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "kubermatic.k8s.io/v1",
+			Kind:       "Seed",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "<<exampleseed>>",
+			Namespace: "kubermatic",
+		},
```
xrstf, line 90: I wish I didn't have to specify the TypeMeta, but without it there is no meta in the YAML output.

api/codegen/seed-yaml/main.go
xrstf: Now that the line contains a further function call, I'd call it justified (no pun intended).

docs/zz_generated.seed-cr.yaml
alvaroaleman: Can we maybe have a special sed added just for the fake provider that re-adds the omitempty?

api/cmd/example-yaml-generator/main.go
alvaroaleman: If you want to be fancy you could use the scheme to set it

api/Gopkg.toml
```diff
@@ -212,3 +212,7 @@ required = [
 [[constraint]]
   name = "github.com/docker/distribution"
   version = "v2.7.1"
+
+[[constraint]]
```
alvaroaleman: I think this constraint is pretty pointless :D

api/vendor/github.com/kubernetes/test-infra/gubernator/github/secrets.py
```diff
@@ -0,0 +1 @@
+../secrets.py
```
alvaroaleman: why  does dep add this?

api/vendor/github.com/kubernetes/test-infra/prow/cmd/tide/pr-authors.md
```diff
@@ -0,0 +1,38 @@
+# PR Author's Guide to Tide
```
alvaroaleman: As a stopgap, can we maybe `.gitignore` md and py files in vendor? Dep shouldn't be adding these

docs/zz_generated.seed-cr.yaml
```diff
@@ -0,0 +1,121 @@
+# Country of the seed. For informational purposes only
```
alvaroaleman: I think this neeeds to be deleted now

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,125 @@
...
+  # The possible datacenters for the nodes
+  datacenters:
+    <<exampledc>>:
+      # Country of the seed. For informational purposes only
+      country: ""
+      # Detailed location of the cluster. For informational purposes only
+      location: ""
+      # Node holds node-specific settings, like e.g. HTTP proxy, insecure registries and the like
+      node:
+        http_proxy: null
```
alvaroaleman: Hm, this is problematic, because internally this is a struct pointer that if set sets the proxy. If we show it here even when it contains empty strings it will be set thought..Any idea  how to cope with that?

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,125 @@
...
+          # Must be the FQDN (example: /datacenter-1/vm/all-kubermatic-vms-in-here )
+          # Defaults to the root VM folder: /datacenter-1/vm
+          root_path: ""
+          templates:
+            coreos: foo
+  # A reference to the Kubeconfig of this cluster
+  kubeconfig:
+    # API version of the referent.
+    apiVersion: ""
+    # If referring to a piece of an object instead of an entire object, this string
```
alvaroaleman, line 158: Upstream seems to have a tiny bit more godocs

docs/zz_generated.seed-cr.yaml
xrstf: 3r9wef

Can do.

api/vendor/github.com/kubernetes/test-infra/gubernator/github/secrets.py
xrstf: Because I am bad and don't know how to use it. I did `dep ensure -add github.com/kubernetes/test-infra/pkg/genyaml` and dep said it was pointless because nobody uses the package. Pls halp.

docs/zz_generated.cluster.yaml
alvaroaleman: No but do we need this for clusters? They are not really a public api

api/vendor/github.com/kubernetes/test-infra/gubernator/github/secrets.py
alvaroaleman: You need to import it as `k8s.io/test-infra/pkg/genyaml`

docs/zz_generated.seed.yaml
xrstf: What do you mean? This matches

```
type ObjectReference struct {
	// Kind of the referent.
	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds
	// +optional
	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
	// Namespace of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/
	// +optional
	Namespace string `json:"namespace,omitempty" protobuf:"bytes,2,opt,name=namespace"`
	// Name of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,3,opt,name=name"`
	// UID of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
	// +optional
	UID types.UID `json:"uid,omitempty" protobuf:"bytes,4,opt,name=uid,casttype=k8s.io/apimachinery/pkg/types.UID"`
	// API version of the referent.
	// +optional
	APIVersion string `json:"apiVersion,omitempty" protobuf:"bytes,5,opt,name=apiVersion"`
	// Specific resourceVersion to which this reference is made, if any.
	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#concurrency-control-and-consistency
	// +optional
	ResourceVersion string `json:"resourceVersion,omitempty" protobuf:"bytes,6,opt,name=resourceVersion"`

	// If referring to a piece of an object instead of an entire object, this string
	// should contain a valid JSON/Go field access statement, such as desiredState.manifest.containers[2].
	// For example, if the object reference is to a container within a pod, this would take on a value like:
	// "spec.containers{name}" (where "name" refers to the name of the container that triggered
	// the event) or if no container name is specified "spec.containers[2]" (container with
	// index 2 in this pod). This syntax is chosen only to have some well-defined way of
	// referencing a part of an object.
	// TODO: this design is not final and this field is subject to change in the future.
	// +optional
	FieldPath string `json:"fieldPath,omitempty" protobuf:"bytes,7,opt,name=fieldPath"`
}
```

api/cmd/example-yaml-generator/main.go
xrstf: Nah, not worth it.

docs/zz_generated.seed.yaml
alvaroaleman: yeah it was just an obvservation, no critique :)

xrstf: Aaaah I get it.

docs/zz_generated.seed-cr.yaml
xrstf: I think I found a pretty neat solution for this. We will simply annotate fields we do not want to have in genyaml with ... `omitgenyaml`. Our sed magic then replaces that with `omitempty` and we're done:

```bash
# remove omitempty tags from structs so that genyaml will not skip fields
sed -i 's/,omitempty//g' pkg/crd/kubermatic/v1/*.go
sed -i 's/,omitempty//g' vendor/k8s.io/api/core/v1/*.go

# there are some fields that we do actually want to ignore
sed -i 's/omitgenyaml/omitempty/g' pkg/crd/kubermatic/v1/*.go
```
Go code:

```go
Fake *DatacenterSpecFake `json:"fake,omitempty,omitgenyaml"` // omitgenyaml is used by the example-yaml-generator
```

api/pkg/resources/resources.go
```diff
@@ -864,11 +864,16 @@ func GlobalSecretKeySelectorValueGetterFactory(ctx context.Context, client ctrlr
...
 		return nil
 	}
 	var envVars []corev1.EnvVar
 
+	// TODO: How do we replicate the "is nil or empty string" check here?
+	//       a) We could turn the HTTPPRoxy from *string into a custom type that has an IsEmpty() function
+	//       b) We could splatter the logic for determining if this is set throughout the codebase (i.e. copy-paste)
+	//       c) When defaulting the Seed at runtime, we ensure that empty values are set to nil,
+	//          so the rest of the codebase can rely on a "clean" ProxySettings struct.
 	if seed.Spec.ProxySettings.HTTPProxy != nil {
```
xrstf: @alvaroaleman

docs/zz_generated.seed-cr.yaml
alvaroaleman: That sounds awesome!

docs/zz_generated.seed.yaml
xrstf: I've begun using a new `Empty()` function on ProxySettings structs. See https://github.com/kubermatic/kubermatic/pull/4485#discussion_r335355827 for the question on how we want to handle this throughout Kubermatic.

api/Gopkg.toml
alvaroaleman: How is this resolved?

api/pkg/resources/resources.go
alvaroaleman: SGTM

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,184 @@
...
+  proxy_settings:
+    # If set, this proxy will be configured for both HTTP and HTTPS.
+    http_proxy: ""
+    # If set this will be set as NO_PROXY environment variable on the node;
+    # The value must be a comma-separated list of domains for which no proxy
+    # should be used, e.g. "*.example.com,internal.dev"
+    no_proxy: ""
+  # SeedDNSOverwrite can be used to override the DNS name used for this seed.
+  # This field is optional and defaults to the seed name.
+  seed_dns_overwrite: null
```
alvaroaleman: We can actually change the internal check to check for an empty string here I think, because an empty string as value doesn't make any sense

xrstf: Either that or we set an empty string as the example value. I also did this for the proxy stuff, because "null" doesn't tell the user what data type is expected (string? list? object?).

api/Gopkg.toml
xrstf: I though this was the old constraint. I now finally deleted it entirely.

docs/zz_generated.seed.yaml
xrstf: I changed the type of the field to `string` both in the `kubermatic/v1` package as well as in the `DatacenterMeta` inside the `provider` package.

api/pkg/resources/resources.go
xrstf: I introduced a new `ProxyValue` type with an `Empty() bool` and a `String() string` function.

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,183 @@
...
+          enforce_floating_ip: false
+          ignore_volume_az: false
+          images:
+            centos: ""
+            coreos: ""
+            ubuntu: ""
+          # Gets mapped to the "manage-security-groups" setting in the cloud config.
+          # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
+          # To make this change backwards compatible, this will default to true.
+          manage_security_groups: false
```
xrstf: So is this correct? What does `will default` mean? Is it not yet implemented? Does defaulting happen later?

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,183 @@
...
+          node_size_requirements:
+            # MinimumMemory is the minimum required amount of memory, measured in MB
+            minimum_memory: 0
+            # VCPUs is the minimum required amount of (virtual) CPUs
+            minimum_vcpus: 0
+          region: ""
+          # Gets mapped to the "trust-device-path" setting in the cloud config.
+          # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
+          trust_device_path: null
```
xrstf: Is null really the default we want to show here?

alvaroaleman: Yeah I guess so. We should probably add the info that this is a string.

docs/zz_generated.seed.yaml
alvaroaleman: I don't know but I have trust in your code-reading skills :)

api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -33,48 +33,62 @@ type Seed struct {
...
+	// For informational purposes in the Kubermatic dashboard only.
 	Country string `json:"country,omitempty"`
-	// Detailed location of the cluster. For informational purposes only
+	// Detailed location of the cluster, like "Hamburg" or "Datacenter 7".
+	// For informational purposes in the Kubermatic dashboard only.
 	Location string `json:"location,omitempty"`
-	// Node holds node-specific settings, like e.g. HTTP proxy, insecure registries and the like
-	Node NodeSettings   `json:"node"`
+	// Node holds node-specific settings, like e.g. HTTP proxy and Docker
+	// registries and the like. Proxy settings are inherited from the seed if
```
alvaroaleman, line 78: Nit: Use a comma instead of more than one `and`

api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -132,33 +157,46 @@ type OpenstackNodeSizeRequirements struct {
...
 	Datacenter string `json:"datacenter"`
-	Cluster    string `json:"cluster"`
-	// Optional root path for cluster specific VM folders.
-	// Each cluster gets a own folder below the root folder.
-	// Must be the FQDN (example: /datacenter-1/vm/all-kubermatic-vms-in-here )
-	// Defaults to the root VM folder: /datacenter-1/vm
-	RootPath  string    `json:"root_path"`
+	// The name of the cluster to use.
+	Cluster string `json:"cluster"`
+	// Optional root path for cluster specific VM folders. Each cluster gets its own
```
alvaroaleman: Not introduced by you, but can we replace `cluster` with `Kubernetes cluster`, as vsphere also has the concept of a cluster which may lead to confusion?

api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -132,33 +157,46 @@ type OpenstackNodeSizeRequirements struct {
...
 
 // DatacenterSpecAWS describes an AWS datacenter
 type DatacenterSpecAWS struct {
-	Region string    `json:"region"`
+	// The AWS region to use, e.g. "us-east-1". For a list of available regions, see
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html
+	Region string `json:"region"`
+
+	// List of AMIs to use for a given operating system.
+	// TODO: Explain how the default for this is assembled.
```
alvaroaleman: This gets defaulted by querying for the latest AMI for the given distribution when machines are created.

docs/zz_generated.seed.yaml
xrstf: It *is* already defaulting to true, so I changed the commit slightly and updated the example YAML.

docs/zz_generated.seed.yaml
xrstf: Gotcha! It's actually a bool and not a string. As it defaults to false, I changed the example to display `false` instead of `null`.

api/pkg/crd/kubermatic/v1/datacenter.go
xrstf: Do you mean the comment or the name of the struct field?

alvaroaleman: The comment

xrstf:
From my side, this is finally done.

api/vendor/k8s.io/test-infra/gubernator/github/secrets.py
```diff
@@ -0,0 +1 @@
+../secrets.py
```
alvaroaleman, line 1: Can we try to get rid from `vendor/k8s.io/test-infra/{prow,gubernator}`? Not sure how `dep` ended up importing this in the first place

docs/zz_generated.seed.yaml
```diff
@@ -0,0 +1,194 @@
...
+        # to this value.
+        no_proxy: ""
+        # Optional: Translates to --pod-infra-container-image on the kubelet.
+        # If not set, the kubelet will default it.
+        pause_image: ""
+      # Spec describes the cloud provider settings used to manage resources
+      # in this datacenter. Exactly one cloud provider must be defined.
+      spec:
+        aws:
+          # List of AMIs to use for a given operating system.
```
alvaroaleman, line 46: Could we add a comment indicating that this shouldn't be set in most cases because it gets defaulted?

alvaroaleman (commented):
Two nits, other than that lgtm. Really great work!

api/vendor/k8s.io/test-infra/gubernator/github/secrets.py
xrstf: Maybe dep gets confused because this is a symlink?

alvaroaleman (approved):
/lgtm
/approve

api/vendor/k8s.io/test-infra/gubernator/github/secrets.py
xrstf: Seems like dep is just buggy: https://github.com/golang/dep/issues/1564

PR #4486, opened 2019-10-14
Add endpoint listing accessible addons

maciaszczykm:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Related to https://github.com/kubermatic/dashboard-v2/issues/1469.

**Documentation**:

```release-note
Added endpoint listing accessible addons.
```


api/pkg/handler/routes_v1.go
```diff
@@ -2666,6 +2673,34 @@ func (r Routing) deleteNodeDeployment() http.Handler {
 	)
 }
 
+// swagger:route POST /api/v1/addons addon
```
multi-io, line 2676: should be GET, not POST

multi-io:
As a suggestion -- probably for a separate PR though -- we might at some point want different users / projects / groups to have different lists of accessible addons.

multi-io:
(sorry for the late post-merge review btw)

api/pkg/handler/routes_v1.go
maciaszczykm: True, my bad.

maciaszczykm:
@multi-io Thanks for the input. I think that might be a good idea, but that should be easy to add even when the frontend will be already done. We will be always able to switch the endpoint.

Addditionaly this endpoint might be able to return the list of addons that are not in use for a given cluster. So if somebody already deployed some addon it will not be returned as accessible anymore.

PR #4496, opened 2019-10-15
Fix disabling of apiserver endpoints reconciler.

thz:
* Actually reconcile for APIServerSettings.EndpointReconcilingDisabled
* Adds test for componentsOverride

```release-note
none
```

### DIFF
--- a/api/cmd/kubermatic-controller-manager/options.go
+++ b/api/cmd/kubermatic-controller-manager/options.go
@@ -123,7 +123,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.BoolVar(&c.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters")
 	flag.StringVar(&c.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources")
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
-	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-endpoint-reconciling-disabled", false, "Whether to disable reconciling for the apiserver endpoints")
+	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-endpoint-reconciling-disabled", false, "Whether to disable reconciling for the apiserver endpoints by default")
 	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
 	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
 	flag.IntVar(&c.concurrentClusterUpdate, "max-parallel-reconcile", 10, "The default number of resources updates per cluster")
--- a/api/cmd/kubermatic-controller-manager/options.go
+++ b/api/cmd/kubermatic-controller-manager/options.go
@@ -123,7 +123,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.BoolVar(&c.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters")
 	flag.StringVar(&c.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources")
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
-	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-endpoint-reconciling-disabled", false, "Whether to disable reconciling for the apiserver endpoints by default")
+	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-reconciling-disabled-by-default", false, "Whether to disable reconciling for the apiserver endpoints by default")
 	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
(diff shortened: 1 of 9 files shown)


api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter_test.go
```diff
@@ -0,0 +1,148 @@
...
+				if val := cluster.Spec.ComponentsOverride.Apiserver.EndpointReconcilingDisabled; val != nil {
+					return fmt.Errorf("expected EndpointReconcilingDisabled to be nil, was %v", val)
+				}
+				if cluster.Spec.ComponentsOverride.Apiserver.Replicas == nil ||
+					cluster.Spec.ComponentsOverride.Scheduler.Replicas == nil ||
+					cluster.Spec.ComponentsOverride.ControllerManager.Replicas == nil ||
+					*cluster.Spec.ComponentsOverride.Apiserver.Replicas != 1 ||
+					*cluster.Spec.ComponentsOverride.Scheduler.Replicas != 2 ||
+					*cluster.Spec.ComponentsOverride.ControllerManager.Replicas != 3 {
+					return fmt.Errorf("expected unmodified Replica counts, at least one was modified")
```
alvaroaleman: You could use `github.com/go-test/deep.Equal` to make this shorter and the error easier to understand

alvaroaleman: Same comment for basically all the comparisons

alvaroaleman (commented):
You forgot to rename the cmd arg

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter.go
```diff
@@ -93,6 +93,13 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, cluster *kubermaticv1.Clu
 	if targetComponentsOverride.Apiserver.Resources == nil {
 		targetComponentsOverride.Apiserver.Resources = r.defaults.Apiserver.Resources
 	}
+	if targetComponentsOverride.Apiserver.EndpointReconcilingDisabled == nil {
+		if r.defaults.Apiserver.EndpointReconcilingDisabled != nil {
```
alvaroaleman: I don't think you need this check, in the log will in doubt just display `nil`

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+func TestReconciliation(t *testing.T) {
+	testCases := []struct {
+		name     string
+		cluster  kubermaticv1.Cluster
+		override kubermaticv1.ComponentSettings
+		verify   func(error, *kubermaticv1.Cluster) error
+	}{
+		{
+			name:    "Defaulting without EndpointReconcilingDisabled",
```
alvaroaleman: I think you will never hit this case, the `r.defaults.Apiserver.EndpointReconcilingDisabled` is always non-nil

api/cmd/kubermatic-controller-manager/options.go
```diff
@@ -123,7 +123,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.BoolVar(&c.dynamicDatacenters, "dynamic-datacenters", false, "Whether to enable dynamic datacenters")
 	flag.StringVar(&c.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources")
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
-	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-endpoint-reconciling-disabled", false, "Whether to disable reconciling for the apiserver endpoints")
+	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-endpoint-reconciling-disabled", false, "Whether to disable reconciling for the apiserver endpoints by default")
```
alvaroaleman: maybe `apiserver-reconciling-disabled-by-default`?

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter.go
thz: Without that check (target == nil) the default would be applied even if the cluster already has a value set, right?

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter_test.go
thz: True. With current defaults it is false or true depending on the command line flag. But shouldn't that be changed so that it is nil unless the command line flag is true?
Otherwise we would always have true or false in the cluster.spec.componentsOverride and never "unspecified".
How does that work with the replicas, is that specifically omitted on replicas=1?

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter.go
alvaroaleman: I was referring to the `r.defaults.Apiserver.EndpointReconcilingDisabled != nil` check, the `if targetComponentsOverride.Apiserver.EndpointReconcilingDisabled == nil` is needed, yes

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter_test.go
alvaroaleman: IIRC we always set it via the controller if its nil. The check mainly serves the purpose of avoiding to override a change a user did.

IMHO that is fine, having any value in the fields makes them easier to discover

api/pkg/controller/clustercomponentdefaulter/clustercomponentdefaulter.go
thz: The problem is the pointer deref on nil... But I think I can remove the whole log here as it does not make much sense for me to log just for this one.
(More a left-over from when i tested it.)

Issue #3417, opened 2019-05-06
Upgrade test: Verify reconciliation was done prior to running conformance tests


alvaroaleman:
**User Story**

Currently, our upgrade tests:

1. Deploy old Kubermatic
1. Create a Cluster
1. Run conformance tests
1. Upgrade conformance tests
1. Deploy new Kubermatic

This has the issue that we don't wait for the new Kubermatic to actually reconcile the cluster. If it fails, we will never notice and basically just run the tests again against the same, unchanged cluster.

**Acceptance criteria**

* Upgrade tests wait for all Kubermatic controller to successfully reconcile the cluster before running the conformance tests the second time
* To achieve this, all our controllers must emit a metric, telling the version with which they last successfully reconciled: `lastSuccessfullSyncTimestamp{controller_name: string, reqAfter: bool, kubermatic_version: string, cluster_name: string}`
* This includes all controllers in the Kubermatic controller manager and in the usercluster controller manager
* After upgrading Kubermatic, the conformance tester waits for this metric to appear for all controllers
* After this metric appeared, the conformance tester waits for all Deployments and Statefulsets to have `replicas == .status.updatedReplicas` and for all Pods to be `Ready` to ensure all Pods were successfully updated

mrIncompetent:
@alvaroaleman @mrIncompetent Please refine

PR #4505, opened 2019-10-17
Refactor cluster condition handling

alvaroaleman:

This PR refactors the cluster condition handling.

User-visible changes:

* We use a combination of the tag and the commit to show the kubermatic version to make the whole thing more readable
* We derive the kubermatic version from the vars in the resource package rather than the cluster status, to make sure its correct in the condition regardless of the question if the one controller that sets it in the status (cluster controller) did that already
* Instead of removing the condition when its false, we set its status to false
* We set the `LastTransitionTime` only when we can actually say for sure if it changed, rather than assuming absent -> present means it changed

Code changes:

* SetClusterCondition is exported to make it re-usable
* GetClusterCondition is exported to make it re-usable
* SetClusterCondition and GetClusterCondition are moved into their own package to avoid a recursive import of the `resources` package
* SetClusterCondition and GetClusterCondition got tests
* `SeedResourcesUpToDate` was renamed to `ClusterConditionSeedResourcesUpToDate` so ppl have a chance to find it via their IDEs completion
* `ClusterUpdateSuccesfulReason` was renamed to `ReasonClusterUpdateSuccessful` so ppl have a chance to find it via their IDE
* `SetSeedResourcesUpToDateCondition` retries cluster updates to avoid having it randomly fail due to concurrency issues
* The checks if SeedResources are up to date where factored out into their own function to allow evaluating them lazily instead of always evaluation everything and make it easier to follow the code

Fixes #

I am wondering if it wouldn't make more sense to have a small controller that sets the `SeedResourcesUpToDate` condition, because we currently have three controllers that do this but using the same underlying data which seems pretty pointless. What is more, they use their own reconciliation success to determine its state, so if one of them has an error it will fight with the others and the condition will keep flapping.

This will most likely entail having having a reconciliation success condition for each controller (which we need anyways, #3417) and use that plus the state of the resources to set the `SeedResourcesUpToDate` condition.

**Documentation**:

```release-note
none
```

/assign @xrstf 
/cc @moadqassem

### DIFF
--- a/api/pkg/controller/cluster/cluster_controller.go
+++ b/api/pkg/controller/cluster/cluster_controller.go
@@ -22,6 +22,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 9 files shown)


alvaroaleman:
> I am wondering if it wouldn't make more sense to have a small controller that sets the SeedResourcesUpToDate condition, because we currently have three controllers that do this but using the same underlying data which seems pretty pointless. What is more, they use their own reconciliation success to determine its state, so if one of them has an error it will fight with the others and the condition will keep flapping.

Then again this implies that on Kubermatic upgrade, all clusters will be available for reconciliation which defeats the original purpose of that change. Maybe it makes more sense to simply ignore the `reconciliationSuccess` of the controllers for that?

api/pkg/controller/cluster/cluster_controller.go
```diff
@@ -214,23 +215,25 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 
-	if err := controllerutil.SetSeedResourcesUpToDateCondition(ctx, cluster, r, successfullyReconciled); err != nil {
+	if err := controllerutil.SetSeedResourcesUpToDateCondition(ctx, cluster, r.Client, successfullyReconciled); err != nil {
 		log.Errorw("failed to update clusters status conditions", zap.Error(err))
-		reconcileErr = fmt.Errorf("failed to set cluster status: %v after reconciliation was done with err=%v", err, reconcileErr)
+		errs = append(errs, err)
 	}
 
-	return *result, reconcileErr
+	return *result, utilerrors.NewAggregate(errs)
```
xrstf, line 236: This is a really nice refactoring :+1:

api/pkg/controller/util/util.go
```diff
@@ -106,56 +108,84 @@ func ConcurrencyLimitReached(ctx context.Context, client ctrlruntimeclient.Clien
...
-					statefulSetHasUnfinishedUpdates = true
-				}
-			}
-		}
+	return updateCluster(ctx, client, cluster.Name, func(c *kubermaticv1.Cluster) {
+		kubermaticv1helper.SetClusterCondition(c,
+			kubermaticv1.ClusterConditionSeedResourcesUpToDate,
+			corev1.ConditionTrue,
+			kubermaticv1.ReasonClusterUpdateSuccessful,
+			"Some controlplane componends did not finish updating")
```
xrstf: componenTs

api/pkg/controller/util/util.go
```diff
@@ -106,56 +108,84 @@ func ConcurrencyLimitReached(ctx context.Context, client ctrlruntimeclient.Clien
...
+		//Get latest version
+		cluster := &kubermaticv1.Cluster{}
+		if err := client.Get(ctx, types.NamespacedName{Name: clusterName}, cluster); err != nil {
+			return err
+		}
+		// Apply modifications
+		modify(cluster)
+		// Update the cluster
+		return client.Update(ctx, cluster)
+	})
```
xrstf, line 190: I've seen so many of these updateXXX functions lately that I wonder how they stand in relation to the reconciling system and if we should extract them somewhere. But then we would basically re-invent the codegen we already have for reconciling things.

Or am I overlooking something dramatically different between this and the reconciling package? Besides being a bit simpler to use, of course.

api/pkg/crd/kubermatic/v1/helper/helper_test.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+
+	for idx := range testCases {
+		tc := testCases[idx]
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			index, condition := GetClusterCondition(tc.cluster, tc.conditionType)
+			if index != tc.expectedIndex {
+				t.Errorf("expected index %d, got index %d", index, tc.expectedIndex)
```
xrstf: You accidentally flipped the two variables here.

api/pkg/controller/util/util.go
alvaroaleman: We do have a lot of them and yes, they all do the same. They are different from what is done in the reconciling package, as they are just about updating things, not upsert.

The main reason we have them so often is:
* We need them once per type as they are typed
* convenience

Generally, we could move all of them in typed factories but I thought this is easier to read and they are hard to get wrong anyways.

xrstf: > they are hard to get wrong

Challenge accepted! :trollface:

PR #4507, opened 2019-10-18
Mark default addons

maciaszczykm:
**What this PR does / why we need it**: We need it to be able to tell in the UI if the addon is default or not. User should be aware that he is trying to delete default addon that will be recreated automatically. We could even block such an operation then.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Related to https://github.com/kubermatic/dashboard-v2/pull/1683.

**Documentation**:

```release-note
Added isDefault field for addons.
```

### DIFF
--- a/api/pkg/controller/addon/addon_controller.go
+++ b/api/pkg/controller/addon/addon_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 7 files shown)


api/pkg/controller/addon/addon_controller.go
```diff
@@ -53,6 +53,8 @@ type Reconciler struct {
 	log                *zap.SugaredLogger
 	workerName         string
 	addonVariables     map[string]interface{}
+	kubernetesAddons   []string
```
alvaroaleman: Could we rename those variables to `default{Kubernetes,Openshift}Addons`?

api/pkg/controller/addon/addon_controller.go
```diff
@@ -70,8 +72,10 @@ func Add(
 	numWorkers int,
 	workerName string,
 	addonCtxVariables map[string]interface{},
-	kubernetesAddonDir string,
-	openshiftAddonDir string,
+	kubernetesAddons,
+	openshiftAddons []string,
```
alvaroaleman: Nit: We could probably use `k8s.io/apimachinery/pkg/util/sets.String` instead of a slice here, then you can just do an `if defaultAddons.Has(addon.Name) != addon.Spec.IsDefault {updateAddon]` in `markDefaultAddons`

maciaszczykm:
@alvaroaleman Done, PTAL.

PR #4519, opened 2019-10-21
Kubermatic controller manager controller: Set seed reconcile success condition

alvaroaleman:

Makes all the controllers in the kubermatic controller manager set a `reconcileSuccess` condition, which we need for #3417 

I think I'll create follow-ups for:
* Adding a variadic arg `predicates ..func(*cluster)(*reconcile.Result, error)` to the `ClusterReconcileWrapper`  which we can use for controllers that have extra conditions, e.G. openshift/kubernetes or the `ClusterAvailableForReconciling`
* Move the whole `SetSeedResourcesUpToDateCondition` into its own controller and remove the `successfullyReconciled` bit from it, as this is a condition that spans multiple controller and can currently cause racing if one controller fails to deploy some parts

Fixes #

**Documentation**:

```release-note
none
```
/assign @xrstf

### DIFF
--- a/api/pkg/controller/addon/addon_controller.go
+++ b/api/pkg/controller/addon/addon_controller.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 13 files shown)


api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -107,6 +107,21 @@ const (
...
 	ClusterConditionSeedResourcesUpToDate ClusterConditionType = "SeedResourcesUpToDate"
 
+	// ClusterConditionClusterControllerReconcilingSuccess indicates whether the cluster
+	// controller successfully finished reconciling this cluster.
+	ClusterConditionClusterControllerReconcilingSuccess ClusterConditionType = "ClusterControllerReconciledSuccessfully"
+
+	// ClusterConditionAddonControllerReconcilingSuccess indicates whether the addon controller
+	// succeeded in reconciling the cluster.
+	ClusterConditionAddonControllerReconcilingSuccess          ClusterConditionType = "AddonControllerReconciledSuccessfully"
+	ClusterConditionAddonInstallerControllerReconcilingSuccess ClusterConditionType = "AddonInstallerControllerReconciledSuccessfully"
```
xrstf, line 112: And at this line you just gave up adding comments? :-D I don't mind missing comments on these self-explanatory constants, just wondering about the inconsistent use.

alvaroaleman: Yeah, pretty much. Do you want all with or all without comment?

api/pkg/controller/addon/addon_controller.go
```diff
@@ -149,42 +150,44 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		reconcilingError := err
-		//Get the cluster so we can report an event to it
-		cluster := &kubermaticv1.Cluster{}
-		if err := r.Get(ctx, types.NamespacedName{Name: addon.Spec.Cluster.Name}, cluster); err != nil {
-			log.Errorw("failed to get cluster for reporting error onto it", zap.Error(err))
-		} else {
-			r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ReconcilingError",
-				"failed to reconcile Addon %q: %v", addon.Name, reconcilingError)
-		}
+		r.recorder.Event(addon, corev1.EventTypeWarning, "ReconcilingError", err.Error())
```
xrstf, line 175: WDYT about including the cluster in this event, like we include the addon name in the event below?

api/pkg/crd/kubermatic/v1/cluster.go
xrstf: Without. The comments add nothing and auto-generated comments are pointless.

api/pkg/controller/addon/addon_controller.go
alvaroaleman: Addons are namespaced and created per cluster, so that is implicit

PR #4528, opened 2019-10-22
Disable cluster status

xrstf:
Setting cluster status makes e2e tests fail. This temporarily disables them.

This PR also renames the status name to something that fits better with the rest of Kubermatic and Kubernetes.

```release-note
NONE
```


xrstf:
This did not have the desired effect. Closing.

PR #4531, opened 2019-10-22
Sync SSH keys to authorized_keys files on nodes

### LINKED ISSUE #3573: Synchronize cluster SSH keys to machines during runtime
**User Story**
As a user i would like to add&remove SSH keys to a cluster & directly access all nodes via the newly added SSH key.

Right now i'm required to rotate all nodes to archive which imposes a risk to my workload.

A potential fix would an agent which runs on each node & updates the `authorized_keys` file on the host.

Open questions:
- [ ] How to handle keys which got manually added ?

**Acceptance criteria**
- [ ] We have a solution which automatically synchronizes the authorized keys

moadqassem:
Signed-off-by: Moath Qasim <moad.qassem@gmail.com>

Adding the daemonset in the usercluster in order to sync the usersshkeys secret with the authorized_keys file on the node.

Fixes #3573

No

```release-note
The `authorized_keys` files on nodes are now updated whenever the SSH keys for a cluster are changed
```


api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,128 @@
...
+		log.Fatalw("Failed to get users directories", zap.Error(err))
+	}
+	if err := usersshkeys.Add(mgr, log, users); err != nil {
+		log.Fatalw("Failed registering user ssh key controller", zap.Error(err))
+	}
+	var g run.Group
+
+	// This group is forever waiting in a goroutine for signals to stop
+	{
+		g.Add(func() error {
```
alvaroaleman: You don't need this rungroup, you need to instead call `mgr.Start()`, its blocking :)

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
+package usersshkeys
```
alvaroaleman: Please move this into its own package `pkg/controller/user-ssh-keys-agent`, the current structure is very confusing, its a subpackage of a controller that doesn't run this controller

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+		Client: mgr.GetClient(),
+		log:    log,
+	}
+
+	c, err := controller.New(operatorName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return err
+	}
+
+	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
alvaroaleman: Nit: We should add a `predicate` so we don't trigger reconciliation on any secret but only the one we care about

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+	var keys = make(map[string][]string, len(paths))
+	for _, path := range paths {
+		file, err := os.Open(path)
+		if err != nil {
+			return nil, err
+		}
+
+		defer file.Close()
+
+		// At the moment we need to read the first line of the authorized_keys file. The reason for that is
```
alvaroaleman: The secret is supposed to be the canonical source of the whole `authorized_keys` file, so we should overwrite everything in it

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+
+	return r.updateAuthorizedKeysFile(existingKeys)
+}
+
+// TODO(MQ): Sync the secrets once the authorized_keys file has been updated.
+func (r *Reconciler) updateAuthorizedKeysFile(data map[string][]string) error {
+	for path, keys := range data {
+		file, err := os.Create(path)
```
alvaroaleman: What if it already exists?

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+		var sshKeys string
+
+		r.log.Debugw("Number of keys per path", len(keys))
```
alvaroaleman: This now gets logged once per `authorized_keys` file rather than once, even thought the number is the same

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+		return err
+	}
+
+	for _, s := range secret.Data {
+		for k := range existingKeys {
+			existingKeys[k] = append(existingKeys[k], string(s))
+		}
+	}
+
+	return r.updateAuthorizedKeysFile(existingKeys)
```
alvaroaleman: Instead of unconditionally writing the file, can we compare it to the desired state and only write if the two differ?

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+	c, err := controller.New(operatorName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return err
+	}
+
+	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("failed to create watcher for secrets: %v", err)
+	}
+
+	return nil
```
alvaroaleman: This also needs to be notified when the file on disk changes. You can e.G. use something like https://github.com/fsnotify/fsnotify + a `source.Channel` for that

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
alvaroaleman: Please add some tests for this

api/pkg/controller/usercluster/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,114 @@
...
+							Path: "/root/.ssh/authorized_keys",
+							Type: &hostPathType,
+						},
+					},
+				},
+				{
+					Name: "core",
+					VolumeSource: corev1.VolumeSource{
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/core/.ssh/authorized_keys",
```
alvaroaleman: This is wrong, its at `/home/core/.ssh/authorized_keys`

api/pkg/controller/usercluster/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,114 @@
...
+							Type: &hostPathType,
+						},
+					},
+				},
+				{
+					Name: "centos",
+					VolumeSource: corev1.VolumeSource{
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/centos/.ssh/authorized_keys",
```
alvaroaleman: This is wrong, its at `/home/centos/.ssh/authorized_keys`

api/pkg/controller/usercluster/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,114 @@
...
+							Type: &hostPathType,
+						},
+					},
+				},
+				{
+					Name: "ubuntu",
+					VolumeSource: corev1.VolumeSource{
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/ubuntu/.ssh/authorized_keys",
```
alvaroaleman: This is wrong, its at `/home/ubuntu/.ssh/authorized_keys`

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
moadqassem: Yeah I know, I added a todo for that 
https://github.com/kubermatic/kubermatic/pull/4531/files/626da65c2742a825b3a55a998261f9d5e083339a#diff-099352596adf63713292d1e6e52d6698R93

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
moadqassem: The file will be recreated again with the new content. The first SSH Key which is added by default will be included in the `data`map

api/pkg/controller/usercluster/resources/usersshkeys/usersshkeys_operator.go
moadqassem: I agree, I was thinking of adding this usersshkey which is added once the user cluster is created to the added secrets. so it will be like(user added keys + this default key) what do you think?

api/Gopkg.toml
```diff
@@ -181,6 +181,10 @@ required = [
   name = "github.com/gogo/protobuf"
   version = "1.1.1"
 
+[[constraint]]
```
alvaroaleman: We have an override for this on line 172

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,98 @@
...
+type controllerRunOptions struct {
+	namespace string
+	log       kubermaticlog.Options
+}
+
+func main() {
+	runOp := controllerRunOptions{}
+	flag.BoolVar(&runOp.log.Debug, "log-debug", true, "Enables debug logging")
+	flag.StringVar(&runOp.log.Format, "log-format", string(kubermaticlog.FormatJSON), "Log format. Available are: "+kubermaticlog.AvailableFormats.String())
+	flag.StringVar(&runOp.namespace, "namespace", metav1.NamespaceSystem, "Namespace in which the cluster is running in")
```
alvaroaleman: The `namespace` flag is unused

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return err
+	}
+
+	namePredicate := predicateutil.ByName(resources.UserSSHKeys)
+	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, namePredicate); err != nil {
+		return fmt.Errorf("failed to create watcher for secrets: %v", err)
+	}
+
+	if err := reconciler.watchAuthorizedKeys(context.TODO(), authorizedKeysPaths); err != nil {
```
alvaroaleman: Please create a watch on the controller for a  `sigs.k8s.io/controller-runtime/pkg/source.Channel` based on the inotify events instead, so this goes through the workqeue and is properly retried with a backoff on errors

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,224 @@
...
+	if err := r.reconcileUserSSHKeys(secret); err != nil {
+		log.Errorw("Failed reconciling user ssh key secret", zap.Error(err))
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *Reconciler) reconcileUserSSHKeys(secret *corev1.Secret) error {
+	for _, path := range r.authorizedKeysPath {
```
alvaroaleman: Plase simplify all of this by just doing a:

```
expectedContent := bytes.Buffer{}
for _, secretData range secret.Data {
  if err := expectedContent.WriteByte(secretData); err != nil {
    return err  
 }
}
for _, path := range r.authorizedKeysPath {
  if _, err := os.Stat(path); if err != nil {
    if !os.IsNotExist(err){
       return err
    }
    if err := ioutil.WriteFile(path, expectedContent.Bytes(), 0600); err != nil {
      return err 
    }
  }
   actualContent, err := ioutil.ReadFile(path)
   if err != nil {
      return err
    }
    if actualContent != expectedContent.Byte() {
       if err := ioutil.WriteFile(path, expectedContent.Byte(), 0600); err != nil {
          return err
       }
    }
  }
}
```

There is no reason to use a set here and the current code is super hard to reason about.

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+var (
+	tmpFiles []string
+	log      = kubermaticlog.New(true, kubermaticlog.FormatConsole).Sugar()
+)
+
+func TestMain(m *testing.M) {
+	m.Run()
+	if err := cleanupFiles(); err != nil {
```
alvaroaleman: Please do the teardown within the tests instead of in a `TestMain` as described in the [testing package](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) so you can properly use `testing.T` to indicate failures instead of having to use `os.Exit`, the latter is ridicolously complicated to debug if it ever goes south

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
```diff
@@ -0,0 +1,154 @@
...
+				if key != tc.expectedSSHKey {
+					t.Fatal("usersshkey secret and authorized_keys file don't match")
+				}
+			}
+		})
+	}
+}
+
+func updateAuthorizedKeysFile(path string) error {
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY, os.ModePerm)
```
alvaroaleman: Please use `io/ioutil.WriteFile` to write files

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	defer file.Close()
+	if _, err := file.WriteString("test_update_ssh_keys"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func readAuthorizedKeysFile(path string) (string, error) {
+	file, err := os.Open(path)
```
alvaroaleman: Plase use `io/ioutil.ReadFile` to read files

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
```diff
@@ -0,0 +1,154 @@
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		text += scanner.Text()
+	}
+	return text, nil
+}
+
+func cleanupFiles() error {
+	for _, tmpFile := range tmpFiles {
+		if _, err := os.Stat(tmpFile); err != nil {
```
alvaroaleman: Please use `os.RemoveAll` for this

api/pkg/resources/resources.go
```diff
@@ -438,7 +438,8 @@ const (
 	VsphereInfraManagementUserUsername = "infraManagementUserUsername"
 	VsphereInfraManagementUserPassword = "infraManagementUserPassword"
 
-	UserSSHKeys = "usersshkeys"
+	UserSSHKeys        = "usersshkeys"
+	AuthorizedKeysPath = "/etc/authorized-keys/"
```
alvaroaleman: Please do not put trailing slashes in paths and instead add the slash once you append something

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,224 @@
...
+		log:                log,
+		authorizedKeysPath: authorizedKeysPaths,
+	}
+
+	c, err := controller.New(operatorName, mgr, controller.Options{Reconciler: reconciler})
+	if err != nil {
+		return err
+	}
+
+	namePredicate := predicateutil.ByName(resources.UserSSHKeys)
```
alvaroaleman: This is not enough, it should filter on both the namespace and the name

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
moadqassem: I totally agree on using the buffer which I can add, however, I need this logic to compare what is saved in the secret and what is saved in the authorized_keys file: 
https://github.com/kubermatic/kubermatic/pull/4531#discussion_r338558210

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
moadqassem: Have you read what is described in the link you are referring to? `TestMain` is used mainly, If you have some tests which need some setup or teardown, and that's exactly what I am doing. For more info please refer to this link:
https://golang.org/pkg/testing/#hdr-Main

About the "the latter is ridiculously complicated to debug if it ever goes south". If something goes wrong I am already logging this error and it will be logged so the debugger would know what went wrong :)

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
moadqassem: Good idea 👍

api/pkg/resources/resources.go
moadqassem: Why is that? Just would like to know out of curiosity :)

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
alvaroaleman: Thats what the code above does:
```
   actualContent, err := ioutil.ReadFile(path)
   if err != nil {
      return err
    }
    if actualContent != expectedContent.Byte() {
       if err := ioutil.WriteFile(path, expectedContent.Byte(), 0600); err != nil {
          return err
       }
    }
```

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
alvaroaleman: >  It also provides a way to share common setup and tear-down cod

moadqassem: Ok your reasoning is absolutely valid, but one thing that I like about `TestMain` is, regardless of any outcomes that we have from the test it will execute what it should be executed. Take this scenario: If one of the test cases fails, then the files which got created won't be deleted because simply  `t.Fatal` will interrupt the execution of the test. Having a cleanup job before each `t.Fatal` makes no sense. TestMain gets executed in the main Thread which makes sure we clean any files after all tests have been executed regardless of the outcome.

alvaroaleman: `t.Fatal` makes the concrete test fail, not all tests in the testsuite. And even if you weren't to use table driven tests a simple `defer` would do the trick.

moadqassem: Who said anything about the whole test suite? 
 
> If one of the test cases fails

api/pkg/resources/resources.go
alvaroaleman: Because that allows to use it as a folder or not and keeping it equal everywhere avoids surprises and having to check for every path

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
alvaroaleman: Unaddressed

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
alvaroaleman: Unadressed

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	if err := mgr.Start(done); err != nil {
+		log.Fatalw("error occurred while running the controller manager", zap.Error(err))
+	}
+}
+
+func availableUsersPaths() ([]string, error) {
+	var paths []string
+	for _, user := range []string{"root", "core", "ubuntu", "centos"} {
```
alvaroaleman, line 74: This is wrong, the paths are:
```
/root/.ssh/authorized_keys
/home/{core,ubuntu,centos}/.ssh/authorized_keys
```

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
alvaroaleman: Unaddressed

api/pkg/resources/resources.go
alvaroaleman: Unaddressed

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
moadqassem: Removed the user ssh keys syncing mechanism from here it trigger the reconcile one an event is detected.
https://github.com/kubermatic/kubermatic/pull/4531/files#diff-d368a8c54741bdc5e868f7ea2d434e2aR113-R128

api/pkg/resources/resources.go
moadqassem: Splitting it into two variables here:
https://github.com/kubermatic/kubermatic/pull/4531/files#diff-f47a57c49bd14a8592f280aa061d94dbR442-R443

moadqassem:
@alvaroaleman PTAL

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,121 @@
...
+			ds.Spec.Selector = &metav1.LabelSelector{MatchLabels: labels}
+			ds.Spec.Template.ObjectMeta.Labels = labels
+
+			ds.Spec.Template.Spec.ServiceAccountName = serviceAccountName
+
+			ds.Spec.Template.Spec.Containers = []corev1.Container{
+				{
+					Name:            daemonSetName,
+					ImagePullPolicy: corev1.PullAlways,
+					Image:           resources.RegistryQuay + "/" + mainRepoName + "/" + daemonSetName + ":" + tag,
```
alvaroaleman: Please do not use `latest`, add some scripting so this image gets built together with kubermatic by using the `resources.KUBERMATICCOMMIT` ldvar as tag and rename the consts in a way that its clear they refer to parts of the image.

Right now we can't tell if the DS that ends up being deployed is whats visible in the diff here.

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,121 @@
...
+						},
+						{
+							Name:      "ubuntu",
+							MountPath: "/home/ubuntu/.ssh/authorized_keys",
+						},
+					},
+				},
+			}
+
+			ds.Spec.Template.Spec.Tolerations = []corev1.Toleration{
```
alvaroaleman: Why did you add the tolerations?

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	daemonSetName = "user-ssh-keys-agent"
+	mainRepoName  = "kubermatic"
+	tag           = "latest"
+)
+
+var (
+	daemonSetMaxUnavailable = intstr.FromInt(1)
+	hostPathType            = corev1.HostPathUnset
```
alvaroaleman: This needs to be updated to fileOrCreate

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	ctx, ctxDone := context.WithCancel(context.Background())
+	defer ctxDone()
+
+	// Create Context
+	done := ctx.Done()
+	ctrlruntimelog.Log = ctrlruntimelog.NewDelegatingLogger(zapr.NewLogger(rawLog).WithName("controller_runtime"))
+
+	mgr, err := manager.New(cfg, manager.Options{})
```
alvaroaleman: Can we limit this to the `kube-system` namespace? The current configuraton won't work as it will try to watch all namespaces which it isn't allowed to

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	log := r.log.With("request", request)
+	log.Debug("Processing")
+
+	secret, err := r.fetchUserSSHKeySecret(ctx, request.NamespacedName.Namespace)
+	if err != nil || secret == nil {
+		return reconcile.Result{}, err
```
alvaroaleman: Generally, please always wrap errors via `fmt.Errorf` before returning them, otherwise debugging errors may become super tricky, as there is no strack trace associated

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,196 @@
...
+		return err
+	}
+
+	namePredicate := predicateutil.ByName(resources.UserSSHKeys)
+	namespacePredicate := predicateutil.ByNamespace(metav1.NamespaceSystem)
+	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, namePredicate, namespacePredicate); err != nil {
+		return fmt.Errorf("failed to create watcher for secrets: %v", err)
+	}
+
+	if err := reconciler.watchAuthorizedKeys(context.TODO(), authorizedKeysPaths); err != nil {
```
alvaroaleman: Wouldn't it be more logical to make `watchAuthorizedKeys` return a channel instead setting it on the `Reconciler`?

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,196 @@
...
+		for {
+			select {
+			case e, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				if e.Op&fsnotify.Write == fsnotify.Write {
+					r.events <- event.GenericEvent{}
+				}
+			case _, ok := <-watcher.Errors:
```
alvaroaleman: Would it maybe make sense to log these? Oherwise errors on the `fsnotify` are completely invisible

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
moadqassem: This was copied and pasted from other daemonset thus I thought of having the same tolerations to this one too since it doesn't have a direct effect it anyway. 

I can get rid of them.

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
moadqassem: This wouldn't work. changing this would result in an error while mounting. This is current error `MountVolume.SetUp failed for volume \"root\" : open /root/.ssh/authorized_keys: no such file or directory"` and was fixed when reverted it to this `corev1.HostPathUnset`

api/cmd/user-ssh-keys-agent/main.go
moadqassem: I have tested this agent and with the same configuration and it didn't result any error while watching the secret resources. Plus I am watching only the secrets in the `kube-system` namespaces https://github.com/kubermatic/kubermatic/pull/4531/files#diff-d368a8c54741bdc5e868f7ea2d434e2aR59-R62

Isn't that enough or I am missing something?

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
moadqassem: Yeah that makes total sense actually

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
moadqassem: Generally, I think having the channel being part of the reconciler has also other benefits like operating on the channel during the life-cycle of the controller(e.g: closing the channel when stoping the controller or for testing by trying to invoke events for testing).

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
alvaroaleman: The question though is if the mode the `.ssh` folder is created with breaks it or not, I am not sure what the default permissions are

api/cmd/user-ssh-keys-agent/main.go
alvaroaleman: So I would have to test this out to give you a detailed `why` but the current configuration results in the `mgr` watching all namespaces (and thus needing the corresponding permissions). It does not dispatch events in a different namespace, but it gets them nonetheless.

moadqassem: I see, well watching the secret in the `kube-system` ns is all that it does. Thus no other namespaces are required. Will adjust it accordingly

api/cmd/user-ssh-keys-agent/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.10
+
+COPY ./user-ssh-keys-agent /usr/local/bin/user-ssh-keys-agent
+
+ENTRYPOINT ["/usr/local/bin/user-ssh-keys-agent"]
```
alvaroaleman: Please set up your IDE to add trailing newlines, otherwise this is really annoying to use :)

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,93 @@
...
+func availableUsersPaths() ([]string, error) {
+	var paths []string
+	for _, user := range []string{"root", "core", "ubuntu", "centos"} {
+		path := fmt.Sprintf("/%v/.ssh/authorized_keys", user)
+		if user != "root" {
+			path = fmt.Sprintf("/home%v", path)
+		}
+		file, err := os.Stat(path)
+		if err != nil {
+			if os.IsNotExist(err) {
```
alvaroaleman, line 81: This will always exist because the kubelet mounting creates it or not?

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,197 @@
...
+				if !ok {
+					return
+				}
+				r.log.Errorw("Error occurred during watching authorized_keys file", zap.Error(err))
```
alvaroaleman: Unless I am misreading this you are logging the error from the `fsnotify.NewWatcher` call here, as you drop the one from `watcher.Errors`

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,197 @@
...
+		r.log.Errorw("Cannot get secret", "secret", resources.UserSSHKeys)
+		return nil, err
+	}
+
+	return secret, nil
+}
+
+func (r *Reconciler) updateAuthorizedKeys(sshKeys map[string][]byte) error {
+	expectedUserSSHKeys := bytes.Buffer{}
+	for _, secretData := range sshKeys {
```
alvaroaleman: since its a map I would think we have to sort it first to not end up updating it with a different oder on each iteration

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	}
+
+	for _, path := range r.authorizedKeysPath {
+		if _, err := os.Stat(path); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+			if err := ioutil.WriteFile(path, expectedUserSSHKeys.Bytes(), 0600); err != nil {
+				return err
```
alvaroaleman: Please wrap this error and all other errors so if we end up falling on our toes we know where it happened

alvaroaleman:
Some small comments, other than that I'd also like to test it out before merging this to make sure this works correctly, but that will have to wait till next week. Thanks for iterating on this!

api/cmd/user-ssh-keys-agent/main.go
moadqassem: You are right. I will get rid of it.

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
moadqassem: Yep, that was the reason why I made it as set in the previous implementation by reversing the keys/values. Will sort it

api/cmd/user-ssh-keys-agent/Makefile
```diff
@@ -0,0 +1,7 @@
+build:
+	CGO_ENABLED=0 go build -o user-ssh-keys-agent
+
+docker: build
+	docker build -t quay.io/kubermatic/user-ssh-keys-agent:$(TAG) .
+
+.PHONY: build docker
```
alvaroaleman: Please add a trailing newline here as well

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/serviceaccount.go
```diff
@@ -0,0 +1,62 @@
...
+	return func() (string, reconciling.RoleBindingCreator) {
+		return roleBindingName, func(rb *rbacv1.RoleBinding) (*rbacv1.RoleBinding, error) {
+			rb.RoleRef = rbacv1.RoleRef{
+				Name:     roleName,
+				Kind:     "Role",
+				APIGroup: rbacv1.GroupName,
+			}
+			rb.Subjects = []rbacv1.Subject{
+				{
+					Kind:     rbacv1.UserKind,
```
alvaroaleman: Must be an empty string for `APIGroup` and `ServiceAccount` for `Kind`

api/pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/daemonset.go
alvaroaleman: Yeah so as mentioned, this does not work:

```
{"level":"error","time":"2019-11-22T17:22:43.852Z","caller":"usersshkeysagent/usersshkeys_operator.go:98","msg":"Failed reconciling user ssh key secret","request":"kube-system/usersshkeys","error":"failed reading authorized key file in path /root/.ssh/authorized_keys: read /root/.ssh/authorized_keys: is a directory"}
```

And thinking about this, we can not really have a mount for each user, because there is no case in which all of the users exist. 

This basically means that we have to mount `/root` which always exists and `/home` which also always exists. Then the controller needs to:

* Iterate over all users
* If userhome does not exist (/root or /home/$user): continue
* make sure `$userhome/.ssh` exists and is a dir
* Make sure `$userhome/.ssh` has a mode of `0600`
* Make sure `$userhome/.ssh` has the correct owner and group, we can find them out by looking at `$userhome`
* Make sure `$userhome/.ssh/authorized_keys` exists, is a file, has a mode of 0600 and is owned by the correct user and group
* Make sure the contents of `$userhome/.ssh/authorized_keys` are as we expect them to be

moadqassem: Just to be clear about the error message(since the comment is actually on the `hostPathType`). The authorized_keys file was created here as a directory, I think that's because the path was empty or? To fix this issue we can simply ignore any file which is a Dir.  Btw which `userhost` you had for this test? was the `authorized_keys` file in the root or in `/home/$user` ?

alvaroaleman: The problem is that we do not know which users exist, so for any user other than root, we can not assume they do, hence we can not mount their home or anything below that

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return paths, nil
+}
+
+func createDirIfNotExist(path string, uid, gid int) error {
+	_, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			if err := os.Mkdir(path, 0600); err != nil {
```
alvaroaleman: Must be `0700` as its a directory, otherwise it can't be listed. Also, please write left-aligned happy path code:

```
_, err := os.Stat(path)
if err == nil {
  return nil
}
if !os.IsNotExist(err) {
return fmt.Errorf("failed describing file info: %v", err)
}
// Create dir
```

The current style of handling the error eleven lines after it occured is super annoying to read.

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+		return fmt.Errorf("failed to create watch for channelSource: %v", err)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	log := r.log.With("request", request)
```
alvaroaleman: Adding the request here is pretty pointless, as its a const

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	for _, path := range r.authorizedKeysPath {
+		if _, err := os.Stat(path); err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("faild describing the authorized_keys file in path %s: %v", path, err)
+			}
+			if err := ioutil.WriteFile(path, expectedUserSSHKeys.Bytes(), 0600); err != nil {
+				return fmt.Errorf("failed writing a new authorzied_keys file in path %s: %v", path, err)
+			}
```
alvaroaleman: Can we add a success log here?

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+
+		actualUserSSHKeys, err := ioutil.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("failed reading authorized key file in path %s: %v", path, err)
+		}
+
+		if !bytes.Equal(actualUserSSHKeys, expectedUserSSHKeys.Bytes()) {
+			if err := ioutil.WriteFile(path, expectedUserSSHKeys.Bytes(), 0600); err != nil {
+				return fmt.Errorf("failed to overwrite authorized_keys file in path %s: %v", path, err)
+			}
```
alvaroaleman: Can we add a success log here?

moadqassem:
The reason why the tests have failed is due to the job `pull-kubermatic-canary-deployment-ci-kubermatic-io`  didn't start

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
+package usersshkeysagent
```
alvaroaleman: Pleae remove the `_operator` from the filename, this is not an operator, this is a controller. The former is a subcategory of the later focusing on managing an application which is not what we do here: https://coreos.com/operators/

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+					return
+				}
+				r.log.Errorw("Error occurred during watching authorized_keys file", zap.Error(err))
+				r.events <- event.GenericEvent{}
+			}
+		}
+	}()
+
+	for _, path := range paths {
+		err = watcher.Add(path)
```
alvaroaleman: Nit: This style as opposed to `if err := watched.Add(path); err != nil {...}` makes it look like you need the error outside of the error handling and keeps the err var longer in scope than necesary

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			return fmt.Errorf("failed reading authorized key file in path %s: %v", path, err)
+		}
+
+		if !bytes.Equal(actualUserSSHKeys, expectedUserSSHKeys.Bytes()) {
+			if err := ioutil.WriteFile(path, expectedUserSSHKeys.Bytes(), 0600); err != nil {
+				return fmt.Errorf("failed to overwrite authorized_keys file in path %s: %v", path, err)
+			}
+		}
+		r.log.Infow("File authorized_keys has been updated successfully", "file", path)
```
alvaroaleman: This is placed wrong, it always always logs that the file has been updated even if it wasn't updated at all because it already had the correct content.

Also, please remove the word `authorized_keys` from all log messages that contain the `path` as a field, it is redundant.

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
```diff
@@ -0,0 +1,254 @@
...
+	return buffer, nil
+}
+
+func updateOwnAndPermissions(path string) error {
+	if err := os.Chmod(path, os.FileMode(0600)); err != nil {
+		return fmt.Errorf("failed to change permission on file: %v", err)
+	}
+
+	sshPath := strings.TrimSuffix(path, "/authorized_keys")
+	if err := os.Chmod(sshPath, os.FileMode(0600)); err != nil {
```
alvaroaleman: This must be `0700`, as its the directory again

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func TestReconcileUserSSHKeys(t *testing.T) {
+	testCases := []struct {
+		name           string
+		reconciler     Reconciler
+		expectedSSHKey string
+	}{
+		{
+			name: "Test updating authorized_keys file from reconcile",
```
alvaroaleman: Since that now went wrong multiple times, please add a testcase for "`.ssh` dir has 0700 permissions, `.ssh/authorized_key` file has 0600 permissions"

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type controllerRunOptions struct {
+	log kubermaticlog.Options
+}
+
+func main() {
+	runOp := controllerRunOptions{}
+	flag.BoolVar(&runOp.log.Debug, "log-debug", true, "Enables debug logging")
```
alvaroaleman: Please set this to false, it shouldn't be running with debug log by default

alvaroaleman:
Also, please add `api/cmd/user-ssh-keys-agent/user-ssh-keys-agent` to `.gitignore`, else someone will check it in sooner or later

api/pkg/controller/usersshkeysagent/usersshkeys_operator_test.go
moadqassem: Good idea

api/pkg/controller/usersshkeysagent/usersshkeys_operator.go
alvaroaleman: Yeah, that comment was wrong. It would help if you would just test if this actually works rather than doing changes, then waiting for me to test it, that would be much quicker.

moadqassem: But that's the thing, I tested it worked. That's why I was confused because I thought you comment was correct! That's why I wasn't sure why it failed when you tested it and it failed.

Anyway will adjust it accordingly

alvaroaleman (approved):
/lgtm
/approve

PR #4533, opened 2019-10-22
Re-Add: Controller-Manager controllers: Set reconcile success condition

alvaroaleman:
…concile success condition (#4519)" (#4529)"

This reverts commit 7dd565ff5af3dee9eb280d2682bc658b4e1aecef.

Fixes #

**Documentation**:

```release-note
none
```


api/pkg/controller/monitoring/monitoring_controller.go
```diff
@@ -217,10 +221,10 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	if err := controllerutil.SetSeedResourcesUpToDateCondition(ctx, cluster, r, successfullyReconciled); err != nil {
```
xrstf: This is now being called for *every* cluster, even paused and foreign worker-name clusters.

alvaroaleman: Hmm that is indeed wrong. But it doesn't explain why this change fails tests, only how tests would interrupt each other after it was merged?

alvaroaleman:
@xrstf PTAL, no conflicts whatsoever :)

api/pkg/controller/monitoring/monitoring_controller.go
alvaroaleman: Fixed

alvaroaleman: This is pretty ugly, I want to do a follow-up that moves all of the `SeedResourcesUpToDate` handling into a distinct controller

PR #4537, opened 2019-10-23
Simplify usercluster SSH key handling code a bit

xrstf:
This removes some leftover residue from the usercluster SSH key syncing PR.

```release-note
NONE
```


api/pkg/controller/usercluster/resources/usersshkeys/secrets.go
```diff
@@ -7,20 +7,16 @@ import (
...
 	return func() (string, reconciling.SecretCreator) {
-		return resources.UserSSHKeys, createSecretsFromUserSSHDirPath(userSSHKeys)
-	}
-}
-
-func createSecretsFromUserSSHDirPath(userSSHKeys map[string][]byte) reconciling.SecretCreator {
-	return func(existing *corev1.Secret) (secret *corev1.Secret, e error) {
-		existing.Data = map[string][]byte{}
+		return resources.UserSSHKeys, func(sec *corev1.Secret) (*corev1.Secret, error) {
+			sec.Data = map[string][]byte{}
```
moadqassem, line 14: What about defining also a size for the map :) 
`
sec.Data = make(map[string][]byte, len(userSSHKeys))
`

xrstf: Could do that, but would it really have a measurable difference? It's not like a slice where every item needs to be copied when the slice grows.

moadqassem: The benefit of this is, no map resize is required with each use of the map which is a good practice.

xrstf: Sooo.... what do? ^^

PR #4544, opened 2019-10-23
Use dedicated kibana parser for fluentbit

xrstf:
Kibana outputs `@timestamp`, which conflicts with Elasticsearch's internal field. This PR configures Fluentbit to use a dedicated parser that can handle the duplicated field.

```release-note
Fix parsing Kibana's logs in Fluent-Bit
```


kron4eg (approved):
/lgtm

Issue #4415, opened 2019-10-01
Upgrade to Go 1.13


xrstf:
### User Story
We want to use the most recent Go version for building and testing Kubermatic.

#### Go-Swagger
With the current set of dependencies, running `make verify` breaks:

```
$ docker run --rm -it -v $(realpath .):/go/src/github.com/kubermatic/kubermatic -e GO111MODULE=off -w /go/src/github.com/kubermatic/kubermatic golang:1.13.1 make -C api verify
make: Entering directory '/go/src/github.com/kubermatic/kubermatic/api'
.......
./hack/verify-swagger.sh
panic: Invalid schema provided to SchemaValidator: object has no key "default"

goroutine 1 [running]:
github.com/kubermatic/kubermatic/api/vendor/github.com/go-openapi/validate.NewSchemaValidator(0xc000b5ed80, 0xc4e5c0, 0xc000b5ed80, 0x0, 0x0, 0xde0560, 0xc0004420c0, 0xc00093fb90)
        /go/src/github.com/kubermatic/kubermatic/api/vendor/github.com/go-openapi/validate/schema.go:71 +0xf83
github.com/kubermatic/kubermatic/api/vendor/github.com/go-openapi/validate.(*SpecValidator).Validate(0xc000ce4940, 0xc080c0, 0xc000af1770, 0xc000d02fd0, 0xc0000ba000)
        /go/src/github.com/kubermatic/kubermatic/api/vendor/github.com/go-openapi/validate/spec.go:92 +0x208
github.com/kubermatic/kubermatic/api/vendor/github.com/go-swagger/go-swagger/cmd/swagger/commands.(*ValidateSpec).Execute(0xc00038b858, 0xc0003d6200, 0x1, 0x2, 0xc00038b858, 0x1)
        /go/src/github.com/kubermatic/kubermatic/api/vendor/github.com/go-swagger/go-swagger/cmd/swagger/commands/validate.go:60 +0x14e
github.com/kubermatic/kubermatic/api/vendor/github.com/jessevdk/go-flags.(*Parser).ParseArgs(0xc0004c7200, 0xc0000b0010, 0x2, 0x2, 0x0, 0xcabd5c, 0x2a, 0xb2e880, 0xc00027fe30)
        /go/src/github.com/kubermatic/kubermatic/api/vendor/github.com/jessevdk/go-flags/parser.go:316 +0x8dc
[...]

PR #4551, opened 2019-10-25
Upgrade binaries in kubernetes-test-binaries image

kron4eg:
Binaries grew old in our testing image, and especially Go itself.

```release-note
NONE
```


xrstf:
You need to adjust the release.sh as well.

alvaroaleman:
This is blocked by https://github.com/kubermatic/kubermatic/issues/4415, we do not want our tests to use a different version of go than the binaries we release.
/hold

kron4eg:
kinda superseded by #4587

PR #4560, opened 2019-10-26
Conformance tester: Do not delete kubernetes-dashboard NS

alvaroaleman:

Prevents the NS cleanup that is run before each ginkgo run to delete the kubernetes-dashboard NS, to not cause errors in the userclustercontroller manager, which will try to create content in it.
Fixes #

**Documentation**:

```release-note
none
```

/assign @xrstf

### DIFF
--- a/api/cmd/conformance-tests/cleanup.go
+++ b/api/cmd/conformance-tests/cleanup.go
@@ -13,7 +13,12 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


xrstf:
Proposal: Can we make the `namespace` const in `pkg/resources/kubernetes-dashboard/deployment.go` public and use that instead of the string literal? I don't have strong opinions here.

alvaroaleman:
> Proposal: Can we make the namespace const in pkg/resources/kubernetes-dashboard/deployment.go public and use that instead of the string literal? I don't have strong opinions here.

Done

xrstf:
Prow does not understand Fakadian (Fake Canadian). :-(

/lgtm

PR #4561, opened 2019-10-26
Conformance tests: Fix bogus nil error logging

alvaroaleman:

Fixes #

**Documentation**:

```release-note
none
```
/assign @xrstf


xrstf:
I think the old version, while being unfortunately phrased, still served as a nice ".Delete() call in the code is done" marker. Can we add a log line if `err == nil`? And if we can, I think we would agree that one log line with an optional err that can be nil would be sufficient and we end up with the code we already had.

Thoughts? :-)

alvaroaleman:
@xrstf done

xrstf:
Beauty!

/lgtm

PR #4589, opened 2019-10-29
Operator: Unify naming of kubermatic resources

xrstf:
This PR takes the opportunity of the operator and tries to fix some longstanding naming inconsistencies:

* <del>The `-manager` suffix has been dropped, as all other controllers from other projects I've seen always just call it `acme-controller`, even if there are multiple controllers running inside.</del>
* The version suffixes have been dropped, as it's unrealistic to run two different APIs or UIs in the same namespace.
* The seed controller manager will not be called `kubermatic-controller-manager`, but `kubermatic-seed-controller-manager`.

I could also live with removing the `kubermatic-` prefix entirely, but having an "api" deployment can be confusing. On the other hand, if we keep the `kubermatic-` prefix, shouldn't our secrets et al. also be prefixed? /shrug

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/operator-master/resources/kubermatic/api.go
+++ b/api/pkg/controller/operator-master/resources/kubermatic/api.go
@@ -17,8 +17,7 @@ import (
 
 func apiPodLabels() map[string]string {
 	return map[string]string{
-		nameLabel:    "kubermatic-api",
-		versionLabel: "v1",
+		nameLabel: "kubermatic-api",
 	}
 }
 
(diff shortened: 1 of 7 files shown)


alvaroaleman:
> The -manager suffix has been dropped, as all other controllers from other projects I've seen always just call it acme-controller, even if there are multiple controllers running inside.

Not true, its called `kubernetes-controller-_manager_`. Please keep the `manager` naming, as all of them run many controllers and not one.

xrstf:
Fixed.

api/pkg/controller/operator-master/resources/kubermatic/api.go
```diff
@@ -170,7 +169,7 @@ func APIDeploymentCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconci
 }
 
 func APIPDBCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedPodDisruptionBudgetCreatorGetter {
-	name := "kubermatic-api-v1"
```
alvaroaleman, line 173: I bet the `const` linter will now complain :P

PR #4594, opened 2019-10-29
Support non-namespaced predicates in seed-lifecycle controller

xrstf:
The seed-operator will soon need to have a lifecycle controller for seeds across all namespaces. To prevent naming conflicts, this PR changes the internal key for seeds to be `namespace/name` and to allow arbitrary predicates. It would have been nice to make predicates a variadic parameter too, but AFAIK there can be only one(tm).

NB: This does not yet solve the question as to how the SeedsGetter itself would handle naming conflicts.

```release-note
NONE
```


alvaroaleman:
Closing, since we wont be needing this anymore

PR #4597, opened 2019-10-30
Make the operator-master controller run namespaced

xrstf:
After learning about all the various troubles we get into when doing seed-cluster reconciles from a single operator pod, we decided that it would make more sense to

- have the master-operator only reconcile its own namespace (or whatever is specified via `-namespace=...`), and
- deploy the seed-operator into the Kubermatic namespace, i.e. have one seed-operator pod per Kubermatic installation. This will significantly simplify the seed-operator logic.

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-operator/main.go
+++ b/api/cmd/kubermatic-operator/main.go
@@ -23,6 +23,7 @@ import (
 
 type controllerRunOptions struct {
 	kubeconfig   string
+	namespace    string
 	internalAddr string
 	log          kubermaticlog.Options
 	workerCount  int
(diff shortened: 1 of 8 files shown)


api/cmd/kubermatic-operator/main.go
```diff
@@ -71,7 +77,7 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	if err := operatormaster.Add(ctx, mgr, 1, log, opt.workerName); err != nil {
+	if err := operatormaster.Add(ctx, mgr, log, opt.namespace, 1, opt.workerName); err != nil {
```
alvaroaleman: If this is hardcoded to `1`, we do not need to pass it in

xrstf: Fixed. It's now using the `-worker-count` CLI flag.

PR #4601, opened 2019-10-30
NOP seed-operator controller

xrstf:
This PR bootstraps a new controller that will be responsible for reconciling seed resources for Kubermatic. At this point the controller does nothing.

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-operator/main.go
+++ b/api/cmd/kubermatic-operator/main.go
@@ -9,8 +9,11 @@ import (
 	"go.uber.org/zap"
 
 	operatormaster "github.com/kubermatic/kubermatic/api/pkg/controller/operator-master"
+	operatorseed "github.com/kubermatic/kubermatic/api/pkg/controller/operator-seed"
+	seedcontrollerlifecycle "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-lifecycle"
 	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
+	"github.com/kubermatic/kubermatic/api/pkg/provider"
 	"github.com/kubermatic/kubermatic/api/pkg/signals"
 
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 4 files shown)


api/cmd/kubermatic-operator/main.go
```diff
@@ -77,11 +80,95 @@ func main() {
...
+	namespace            string
+	seedsGetter          provider.SeedsGetter
+	seedKubeconfigGetter provider.SeedKubeconfigGetter
+	workerCount          int
+	workerName           string
+}
+
+func seedOperatorControllerFactoryCreator(ctrlCtx seedOperatorContext) seedcontrollerlifecycle.ControllerFactory {
+	factory := func(mgr manager.Manager) error {
+		flog := ctrlCtx.log.Named("operator-seed-controller-factory")
```
alvaroaleman: Why `flog` and not just `log`?

api/pkg/controller/operator-seed/controller.go
```diff
@@ -0,0 +1,155 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// ControllerName is the name of this very controller.
+	ControllerName = "kubermatic-seed-operator"
```
alvaroaleman: `s/operator/controller`

api/cmd/kubermatic-operator/main.go
xrstf: Originally I had an explicit `log` parameter on the `seedOperatorControllerFactoryCreator`, which then conflicted at the end. So I used a temporary helper, `flog` means `factory log`. Can clean up if you insist, though IMHO we would trade one name for another and both are equally bad.

alvaroaleman: Well, for `log` I know what it is, for `flog` I didnt't

api/pkg/controller/operator-seed/controller.go
xrstf: The master is called `kubermatic-master-operator`. Naming it `kubermatic-master-controller` will 100% confuse people because of our kubermatic-master-controller-manager.

I found the distinction between something that manages Kubermatic itself (`-operator`) and other controllers, managing non-Kubermatic stuff (`-controller`) rather nice.

api/cmd/kubermatic-operator/main.go
xrstf: `sed`ed it.

PR #4610, opened 2019-11-01
Add external cloud provider support for OpenStack.

ainmosni:
This PR adds external cloud provider support for OpenStack, to accomplish this, the following changes and additions have been made:
    
    * Add a node labeller operator to add distribution labels to nodes.
    * Add Cluster.Spec.Features to the API.
    * Add the FeatureNameExternalCloudProvider feature flag.
    * Create the external cloud provider deployment to the user cluster
      namespace.
    * Add the CSI addon to the addons.
    * Mount the cloud-config into the usercluster controller.
    * Inject the cloud-config into the user cluster.
    * Add a MajorMinor function to semver.

```release-note
Add support for the external cloud provider for OpenStack.
```

### DIFF
--- a/addons/default-storage-class/storage-class.yaml
+++ b/addons/default-storage-class/storage-class.yaml
@@ -44,6 +44,18 @@ parameters:
 {{ end }}
 
 {{ if .Cluster.Spec.Cloud.Openstack }}
+{{ if (.Cluster.Spec.Features.externalCloudProvider) }}
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  annotations:
+    storageclass.beta.kubernetes.io/is-default-class: "true"
+  labels:
+    kubernetes.io/cluster-service: "true"
+  name: cinder-csi
+provisioner: cinder.csi.openstack.org
+volumeBindingMode: WaitForFirstConsumer
+{{ else }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
@@ -53,6 +65,7 @@ metadata:
     kubernetes.io/cluster-service: "true"
   name: standard
 provisioner: kubernetes.io/cinder
+{{ end }}
 {{ end }}
 
 {{ if .Cluster.Spec.Cloud.GCP }}
--- a/api/cmd/user-cluster-controller-manager/main.go
+++ b/api/cmd/user-cluster-controller-manager/main.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -22,10 +23,11 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/ipam"
 	nodelabeler "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/node-labeler"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/nodecsrapprover"
+	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/nodelabel"
 	openshiftmasternodelabeler "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/openshift-master-node-labeler"
 	openshiftseedsyncer "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/openshift-seed-syncer"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/rbac"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources"
+	rbacusercluster "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/rbac"
+	usercluster "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources"
 	machinecontrolerresources "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/machine-controller"
 	kubermaticlog "github.com/kubermatic/kubermatic/api/pkg/log"
 	"github.com/kubermatic/kubermatic/api/pkg/pprof"
@@ -57,6 +59,7 @@ type controllerRunOptions struct {
 	overwriteRegistry             string
 	cloudProviderName             string
 	cloudCredentialSecretTemplate string
+	cloudConfig                   string
 	nodelabels                    string
 	seedKubeconfig                string
 	openshiftConsoleCallbackURI   string
@@ -84,6 +87,7 @@ func main() {
 	flag.StringVar(&runOp.nodelabels, "node-labels", "", "A json-encoded map of node labels. If set, those labels will be enforced on all nodes.")
 	flag.StringVar(&runOp.seedKubeconfig, "seed-kubeconfig", "", "Path to the seed kubeconfig. In-Cluster config will be used if unset")
 	flag.StringVar(&runOp.openshiftConsoleCallbackURI, "openshift-console-callback-uri", "", "The callback uri for the openshift console")
+	flag.StringVar(&runOp.cloudConfig, "cloud-config", "", "Path to cloud config")
 
 	flag.Parse()
 
@@ -107,6 +111,14 @@ func main() {
 	}
 	if runOp.openvpnServerPort == 0 {
 		log.Fatal("-openvpn-server-port must be set")
+	}
+
+	var cloudConfig []byte
+	if runOp.cloudConfig != "" {
+		cloudConfig, err = ioutil.ReadFile(runOp.cloudConfig)
+		if err != nil {
+			log.Fatalf("Failed to read cloud-config: %v", err)
+		}
 	}
 
 	var cloudCredentialSecretTemplate *corev1.Secret
@@ -197,6 +209,7 @@ func main() {
 		healthHandler.AddReadinessCheck,
 		cloudCredentialSecretTemplate,
 		runOp.openshiftConsoleCallbackURI,
+		cloudConfig,
 		log,
 	); err != nil {
 		log.Fatalw("Failed to register user cluster controller", zap.Error(err))
@@ -259,6 +272,10 @@ func main() {
 		log.Fatalw("Failed to register clusterrolelabeler controller", zap.Error(err))
 	}
 	log.Info("Registered clusterrolelabeler controller")
+
+	if err := nodelabel.Add(log, mgr); err != nil {
+		log.Fatalf("Failed to register the node label controller: %v", err)
+	}
 
 	// This group is forever waiting in a goroutine for signals to stop
 	{
--- a/api/pkg/addon/template.go
+++ b/api/pkg/addon/template.go
@@ -33,13 +33,14 @@ func txtFuncMap(overwriteRegistry string) template.FuncMap {
 }
 
 type TemplateData struct {
-	Addon        *kubermaticv1.Addon
-	Kubeconfig   string
-	Cluster      *kubermaticv1.Cluster
-	Credentials  resources.Credentials
-	Variables    map[string]interface{}
-	DNSClusterIP string
-	ClusterCIDR  string
+	Addon             *kubermaticv1.Addon
+	Kubeconfig        string
+	MajorMinorVersion string
+	Cluster           *kubermaticv1.Cluster
+	Credentials       resources.Credentials
+	Variables         map[string]interface{}
+	DNSClusterIP      string
+	ClusterCIDR       string
 }
 
 func ParseFromFolder(log *zap.SugaredLogger, overwriteRegistry string, manifestPath string, data *TemplateData) ([]runtime.RawExtension, error) {
--- a/api/pkg/controller/user-cluster-controller-manager/nodelabel/node_label_controller.go
+++ b/api/pkg/controller/user-cluster-controller-manager/nodelabel/node_label_controller.go
@@ -1,0 +1,140 @@
+package nodelabel
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"go.uber.org/zap"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/nodelabel/resources"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// ControllerName is the name of the controller.
+	ControllerName = "kubermatic_node_label_controller"
+)
+
+// Reconciler is the node label
+type Reconciler struct {
+	log *zap.SugaredLogger
+	ctrlruntimeclient.Client
+}
+
+// Add adds the reconciler to the controller.
+func Add(log *zap.SugaredLogger, mgr manager.Manager) error {
+	reconciler := &Reconciler{
+		log:    log,
+		Client: mgr.GetClient(),
+	}
+
+	ctrlOpts := controller.Options{
+		Reconciler:              reconciler,
+		MaxConcurrentReconciles: 1,
+	}
+
+	c, err := controller.New(ControllerName, mgr, ctrlOpts)
+	if err != nil {
+		return err
+	}
+
+	predicates := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// TODO: Add proper func when they exist.
+			return true
+		},
+	}
+
+	return c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}, predicates)
+}
+
+// Reconcile is the main reconcilliation loop for the controller.
+func (r *Reconciler) Reconcile(_ reconcile.Request) (reconcile.Result, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err := r.labelAllNodes(ctx)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to ensure all nodes have the %s label: %v", resources.DistributionLabelKey, err)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *Reconciler) labelAllNodes(ctx context.Context) error {
+	nodeList := &corev1.NodeList{}
+	if err := r.List(ctx, nodeList); err != nil {
+		return fmt.Errorf("failed to list nodes: %v", err)
+	}
+
+	for _, node := range nodeList.Items {
+		label, err := detectLabel(&node)
+		if err != nil {
+			return fmt.Errorf("could not detect label on node %q: %v", node.Name, err)
+		}
+		hasLabel := hasCorrectDistributionLabel(&node, label)
+
+		if !hasLabel {
+			err := r.applyLabel(ctx, &node, label)
+			if err != nil {
+				return fmt.Errorf("could not apply label on node %q: %v", node.Name, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyLabel(ctx context.Context, node *corev1.Node, label string) error {
+	oldNode := node.DeepCopy()
+	if node.Labels == nil {
+		node.Labels = map[string]string{}
+	}
+	if node.Labels[resources.DistributionLabelKey] != label {
+		r.log.Debugf("Changing label %q from value %q to value %q", resources.DistributionLabelKey, node.Labels[resources.DistributionLabelKey], label)
+		node.Labels[resources.DistributionLabelKey] = label
+		if err := r.Client.Patch(ctx, node, ctrlruntimeclient.MergeFrom(oldNode)); err != nil {
+			return fmt.Errorf("failed to update node: %v", err)
+		}
+	}
+
+	r.log.Debug("No label changes, not updating node.")
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		newNode := &corev1.Node{}
+		if err := r.Get(ctx, types.NamespacedName{Name: node.Name}, newNode); err != nil {
+			return err
+		}
+
+		return r.Client.Update(ctx, newNode)
+	}); err != nil {
+		return fmt.Errorf("failed to update node %q: %v", node.Name, err)
+	}
+
+	return nil
+}
+
+func detectLabel(node *corev1.Node) (string, error) {
+	osImage := strings.ToLower(node.Status.NodeInfo.OSImage)
+	for k, v := range resources.OSLabelMatchValues {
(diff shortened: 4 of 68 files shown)


ainmosni:
Note that this is untested as of this writing, but I'm submitting it now so that it can be looked at while I test.

api/pkg/controller/addon/addon_controller.go
```diff
@@ -294,6 +294,7 @@ func (r *Reconciler) getAddonManifests(log *zap.SugaredLogger, addon *kubermatic
 		Variables:    make(map[string]interface{}),
 		Cluster:      cluster,
 		Credentials:  credentials,
+		MinorVersion: cluster.Spec.Version.MajorMinor(),
```
alvaroaleman: I guess this should be the minor version?

ainmosni: No, I should rename that. I need the combo of MajorMinor. :)

api/pkg/controller/nodelabel/resources/label.go
```diff
@@ -0,0 +1,23 @@
...
+	CentOSLabelValue = "centos"
+
+	// UbuntuLabelValue is the value of the label for Ubuntu
+	UbuntuLabelValue = "ubuntu"
+
+	// ContainerLinuxLabelValue is the value of the label for Container Linux
+	ContainerLinuxLabelValue = "container-linux"
+)
+
+// OSLabelMatchValues is a mapping between OS labels and the strins to match on in OSImage.
```
alvaroaleman: nit: `s/strins/strings`

api/pkg/controller/nodelabel/resources/label.go
```diff
@@ -0,0 +1,23 @@
+package resources
```
alvaroaleman: Could you move the controller below `pkg/controller/user-cluster-controller-manager`? We want to arrange the controllers by binary they run in, because the current structure of having one folder with all controller confuses ppl

api/pkg/controller/openshift/testdata/Kubermatic_API_image_is_overwritten-usercluster-controller.golden.yaml
```diff
@@ -41,7 +42,7 @@ spec:
         - -timeout
         - "1"
         - -command
-        - '{"command":"/usr/local/bin/user-cluster-controller-manager","args":["-kubeconfig","/etc/kubernetes/kubeconfig/kubeconfig","-metrics-listen-address","0.0.0.0:8085","-health-listen-address","0.0.0.0:8086","-namespace","$(NAMESPACE)","-ca-cert","/etc/kubernetes/pki/ca/ca.crt","-ca-key","/etc/kubernetes/pki/ca/ca.key","-cluster-url","https://test-cluster.alias-europe-west3-c.dev.kubermatic.io:0","-openvpn-server-port","0","-overwrite-registry","","-openshift=true","-version","4.1.9","-cloud-provider-name","","-openvpn-ca-cert-file=/etc/kubernetes/pki/openvpn/ca.crt","-openvpn-ca-key-file=/etc/kubernetes/pki/openvpn/ca.key","-user-ssh-keys-dir-path=/etc/kubernetes/usersshkeys"]}'
+        - '{"command":"/usr/local/bin/user-cluster-controller-manager"}'
```
alvaroaleman: That doesn't look right :thinking:

api/pkg/resources/usercluster/deployment.go
```diff
@@ -138,7 +139,6 @@ func DeploymentCreator(data userclusterControllerData, openshift bool) reconcili
 					Name:    name,
 					Image:   data.KubermaticAPIImage() + ":" + resources.KUBERMATICCOMMIT,
 					Command: []string{"/usr/local/bin/user-cluster-controller-manager"},
-					Args:    args,
```
alvaroaleman: This must stay :)

api/pkg/controller/nodelabel/node_label_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		Reconciler:              reconciler,
+		MaxConcurrentReconciles: 1,
+	}
+
+	c, err := controller.New(ControllerName, mgr, ctrlOpts)
+	if err != nil {
+		return err
+	}
+
+	predicates := predicate.Funcs{
```
alvaroaleman: You could probably use the same predicate as in the other node-label controller in `controller/user-cluster-controller-manager/node-labeler/node_labeleler.go`. And maybe just combine the two as they are very similiar?

api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/kubermatic/kubermatic/api/pkg/controller/usercluster/resources/cloudcontroller"
+
```
alvaroaleman: Nit: Please put it into the same inport group as the other kubermatic resources

api/pkg/controller/usercluster/resources/cloudcontroller/clusterrolebinding.go
```diff
@@ -0,0 +1,32 @@
...
+func ClusterRoleBindingCreator() reconciling.NamedClusterRoleBindingCreatorGetter {
+	return func() (string, reconciling.ClusterRoleBindingCreator) {
+		return resources.CloudControllerManagerRoleBindingName, func(crb *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, error) {
+			crb.Labels = resources.BaseAppLabel(resources.CloudControllerManagerRoleBindingName, nil)
+
+			crb.RoleRef = rbacv1.RoleRef{
+				// Can probably be tightened up a bit but for now I'm following the documentation.
+				// https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/
+				Name:     "cluster-admin",
+				Kind:     "ClusterROle",
```
alvaroaleman: `s/ClusterROle/ClusterRole`

api/pkg/controller/usercluster/resources/cloudcontroller/secret.go
```diff
@@ -0,0 +1,22 @@
+package cloudcontroller
+
+import (
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CloudConfig generates the
```
alvaroaleman: Please finish the sentence or completely remove it

api/pkg/resources/apiserver/deployment.go
```diff
@@ -333,6 +333,26 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 	return flags, nil
 }
 
+func GetKubernetesCloudProviderName(cluster *kubermaticv1.Cluster) string {
```
alvaroaleman: Please extend `pkg/resources/data.go#GetKubernetesCloudProviderName`  instead

api/pkg/resources/cloudcontroller/deployment.go
```diff
@@ -0,0 +1,22 @@
...
+
+// DeploymentCreator returns the function to create and update the external cloud provider deployment.
+func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeploymentCreatorGetter {
+	if data.Cluster().Spec.Cloud.Openstack != nil {
+		return openStackDeploymentCreator(data)
+	}
+
+	return func() (name string, create reconciling.DeploymentCreator) {
+		return osName, func(dep *appsv1.Deployment) (*appsv1.Deployment, error) {
+			return &appsv1.Deployment{}, fmt.Errorf("unsupported external cloud controller")
```
alvaroaleman: Please just return nil for the Deployment and use `errors.New` if there is no format string

api/pkg/resources/cloudcontroller/openstack.go
```diff
@@ -0,0 +1,161 @@
...
+	flags := []string{
+		"--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig",
+		"--v=1",
+		"--cloud-config=/etc/kubernetes/cloud/config",
+		"--cloud-provider=openstack",
+	}
+	return flags
+}
+
+func getOSVersion(data *resources.TemplateData) (string, error) {
```
alvaroaleman, line 147: If this is not supported for all versions, can we maybe export a `ExternalCloudControllerSupported(c *kubermaticv1.Cluster) bool` and only if both that func returns true and the feature gate is set enable this?

api/pkg/resources/controllermanager/deployment.go
```diff
@@ -224,7 +224,11 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
 		flags = append(flags, "--cloud-config", "/etc/kubernetes/cloud/config")
 	}
 	if cluster.Spec.Cloud.Openstack != nil {
-		flags = append(flags, "--cloud-provider", "openstack")
+		cloudProvider := "openstack"
```
alvaroaleman: Can we maybe re-use `data.GetKubernetesCloudProviderName` here to not have this in multiple places?

alvaroaleman (commented):
Should we automatically set the feature gate for clusters that support it on create requests?

api/pkg/controller/nodelabel/resources/label.go
ainmosni: Yeah no problem. :)

api/pkg/controller/openshift/testdata/Kubermatic_API_image_is_overwritten-usercluster-controller.golden.yaml
ainmosni: Nice catch, will look into it.

api/pkg/resources/usercluster/deployment.go
ainmosni: Indeed, I think I made a mistake during rebasing.

api/pkg/controller/nodelabel/node_label_controller.go
ainmosni: Yeah when I saw that there was another node labeler, I was thinking of combining them but it was a yak I didn't want to shave today.

api/pkg/controller/usercluster/resources/cloudcontroller/secret.go
ainmosni: Oops my bad.

api/pkg/resources/apiserver/deployment.go
ainmosni: 👍

api/pkg/resources/cloudcontroller/deployment.go
ainmosni: 👍

api/pkg/resources/cloudcontroller/openstack.go
ainmosni: Good idea.

api/pkg/resources/controllermanager/deployment.go
ainmosni: I was thinking that, but that function was still private and in a different package when I worked on this.

api/cmd/user-cluster-controller-manager/main.go
```diff
@@ -142,6 +145,14 @@ func main() {
 	}
 	caCert := &triple.KeyPair{Cert: certs[0], Key: rsaCAKey}
 
+	var cloudConfig []byte
+	if runOp.cloudConfig != "" {
+		cloudConfig, err = ioutil.ReadFile(runOp.cloudConfig)
+		if err != nil {
+			glog.Fatalf("Failed to read cloud-config: %v", err)
```
alvaroaleman: This will fail, we don't use `glog` anymore but `zap`, there should be a `log` variable you can use

api/pkg/addon/template.go
```diff
@@ -35,6 +35,7 @@ func txtFuncMap(overwriteRegistry string) template.FuncMap {
 type TemplateData struct {
 	Addon        *kubermaticv1.Addon
 	Kubeconfig   string
+	MinorVersion string
```
alvaroaleman: Can we rename this here as well?

api/pkg/controller/cluster/resources.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/kubermatic/kubermatic/api/pkg/resources/cloudcontroller"
```
alvaroaleman: Nit: Please group this with the other kuberamatic imports

api/pkg/controller/user-cluster-controller-manager/nodelabel/node_label_controller.go
```diff
@@ -0,0 +1,126 @@
...
+				return fmt.Errorf("could not apply label on node %q: %v", node.Name, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyLabel(ctx context.Context, node *corev1.Node, label string) error {
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
alvaroaleman, line 114: Coukd you use Patch instead?  That imposes less load on the APIserver and is pretty much what we want here. We couldn't use it before because our controller-runtime version didn't support it but thats now the case, check e.G. pkg/controller/user-cluster-controller-manager/node-labeler/node_labeleler.go:115

api/pkg/resources/cloudcontroller/openstack.go
```diff
@@ -0,0 +1,161 @@
...
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: resources.CloudControllerManagerKubeconfigSecretName,
+				},
+			},
+		},
+	}
+}
+
+func getOSPodAnnotations(data *resources.TemplateData) map[string]string {
```
alvaroaleman: I think we can just remove this

api/pkg/resources/controllermanager/deployment.go
```diff
@@ -211,59 +211,41 @@ func getFlags(cluster *kubermaticv1.Cluster) ([]string, error) {
...
-		flags = append(flags, "--cloud-provider", "vsphere")
-		flags = append(flags, "--cloud-config", "/etc/kubernetes/cloud/config")
-	}
-	if cluster.Spec.Cloud.Azure != nil {
-		flags = append(flags, "--cloud-provider", "azure")
-		flags = append(flags, "--cloud-config", "/etc/kubernetes/cloud/config")
-		if cluster.Spec.Version.Semver().Minor() >= 15 {
-			// Required so multiple clusters using the same resource group can
-			// allocate public IPs. Ref: https://github.com/kubernetes/kubernetes/pull/77630
-			flags = append(flags, "--cluster-name", cluster.Name)
```
alvaroaleman, line 240: This got lost or am I missing it?

alvaroaleman (commented):
Some nits, but looks mostly good. Do you want to add anything to enable this by default on new clusters that support it?

api/cmd/user-cluster-controller-manager/main.go
ainmosni: I do use the `log` facility, wonder how you still saw that. :)

api/pkg/addon/template.go
ainmosni: Already renamed, looks like you were reviewing an older version.

api/pkg/controller/cluster/resources.go
ainmosni: Serves me right for letting VSCode import for me and forgetting to organise it. Fixed locally, will push it in a bit.

api/pkg/controller/user-cluster-controller-manager/nodelabel/node_label_controller.go
ainmosni: Sure, would it replace the whole `RetryOnConflict` or just the `r.Client.Update` ?

alvaroaleman: The whole `RetryOnConflict`, just one attempt on patching.

`RetryOnConflict` with the modifiers is essentially a poor persons patching :grimacing:

api/pkg/resources/controllermanager/deployment.go
ainmosni: I shouldn't do big deletions when super tired. :)

api/pkg/controller/user-cluster-controller-manager/nodelabel/node_label_controller.go
ainmosni: Thanks for the tip, will keep that in mind. :)

alvaroaleman:
/lgtm
/approve

Since this won't be used by default I am fine merging this as it is

alvaroaleman:
The unittests fail with a:
```
--- FAIL: TestRetagImageForAllVersions (0.60s)
    main_test.go:31: Error calling getImagesForVersion: failed to get images from addons: failed to get images for addon cinder-csi-plugin: failed to parse addon templates in ../../../addons/cinder-csi-plugin: failed to parse file ../../../addons/cinder-csi-plugin/nodeplugin.yaml: template: nodeplugin.yaml:61: undefined variable "$version"
    main_test.go:31: Error calling getImagesForVersion: failed to get images from addons: failed to get images for addon cinder-csi-plugin: failed to parse addon templates in ../../../addons/cinder-csi-plugin: failed to parse file ../../../addons/cinder-csi-plugin/nodeplugin.yaml: template: nodeplugin.yaml:61: undefined variable "$version"
    main_test.go:31: Error calling getImagesForVersion: failed to get images from addons: failed to get images for addon cinder-csi-plugin: failed to parse addon templates in ../../../addons/cinder-csi-plugin: failed to parse file ../../../addons/cinder-csi-plugin/nodeplugin.yaml: template: nodeplugin.yaml:61: undefined variable "$version"
    main_test.go:31: Error calling getImagesForVersion: failed to get images from addons: failed to get images for addon cinder-csi-plugin: failed to parse addon templates in ../../../addons/cinder-csi-plugin: failed to parse file ../../../addons/cinder-csi-plugin/nodeplugin.yaml: template: nodeplugin.yaml:61: undefined variable "$version"
    main_test.go:31: Error calling getImagesForVersion: failed to get images from addons: failed to get images for addon cinder-csi-plugin: failed to parse addon templates in ../../../addons/cinder-csi-plugin: failed to parse file ../../../addons/cinder-csi-plugin/nodeplugin.yaml: template: nodeplugin.yaml:61: undefined variable "$version"
```

addons/default-storage-class/storage-class.yaml
```diff
@@ -52,8 +52,13 @@ metadata:
   labels:
     kubernetes.io/cluster-service: "true"
   name: standard
+{{ if (.Cluster.Spec.Features.externalCloudProvider) }}
```
alvaroaleman: Storageclasses are immutable, so can we use complete different storage classes?

api/pkg/resources/cloudcontroller/openstack.go
```diff
@@ -0,0 +1,167 @@
...
+		"--v=1",
+		"--cloud-config=/etc/kubernetes/cloud/config",
+		"--cloud-provider=openstack",
+	}
+	return flags
+}
+
+func getOSVersion(data *resources.TemplateData) (string, error) {
+	x := data.Cluster().Spec.Version.String()
+	switch {
```
alvaroaleman: I think we can just use `data.Cluster().Spec.Version.Minor()`, no reason to do string prefix comparisons

addons/cinder-csi-plugin/controllerplugin-rbac.yaml
```diff
@@ -0,0 +1,212 @@
+{{ if (.Cluster.Spec.Features.externalCloudProvider) }}
```
alvaroaleman: Can we move this into the `default-storage-class` addon? With the current approach we'd have to activate this addon by default which will be confusing for setups where Openstack is unsed

api/pkg/resources/machinecontroller/deployment.go
```diff
@@ -114,12 +114,17 @@ func DeploymentCreatorWithoutInitWrapper(data machinecontrollerData) reconciling
 				return nil, err
 			}
 
+			externalCloudProvider := false
+			if x := data.Cluster().Spec.Features[resources.FeatureNameExternalCloudProvider]; x {
```
alvaroaleman: Can be simplified as `externalCloudProvider := data.Cluster().Spec.Features[resources.FeatureNameExternalCloudProvider]`

addons/default-storage-class/storage-class.yaml
ainmosni: We could, but as new clusters with the external CCM enabled should only use the CSI storage, mutating the standard class seemed more logical.

api/pkg/resources/cloudcontroller/openstack.go
ainmosni: We could, until K8s 2 comes around. :)

addons/cinder-csi-plugin/controllerplugin-rbac.yaml
ainmosni: Yeah I thought about that, I thought that adding it to another addon would make it less clear what the addon did. Of course, we could rename `default-storage-class` to just `storage` and it would be more clear.

api/pkg/resources/machinecontroller/deployment.go
ainmosni: Done, confused Go's behaviour with other languages where you'd get a "unknown key" error. :)

addons/default-storage-class/storage-class.yaml
alvaroaleman: True, but we eventually want to have a migration and at that point in time we will have to rename it. And if we keep the name now that would mean that even clusters that never used the in-tree provider will end up with two identical but differently named storageclasses, which is IMHO confusing

api/pkg/resources/cloudcontroller/openstack.go
alvaroaleman: Right, but that is not going to happen anytime soon, so currently this is solving a problem we don't have. Once a 2.X exists (for which upstream doesn't even have plans currently) we will have to update a variety of places anyways

addons/cinder-csi-plugin/controllerplugin-rbac.yaml
alvaroaleman: I do agree that the name would be clearer, but if we did that, we would re-create all resources in it for existing clusters so that is not an option I am afraid :grimacing: Compatibility can be annoying sometimes

ainmosni: I could also just rename `cinder-csi-plugin` to just `csi` or just `storage`, would look a lot less jarring, even if it is a noop on non-OpenStack clusters for now.

api/pkg/resources/cloudcontroller/openstack.go
ainmosni: True, at the very least I could use the `MajorMinor` function that I created after I wrote this code so I can do plain equality instead of prefix comparison.

addons/default-storage-class/storage-class.yaml
ainmosni: Fair enough, I'll change this.

addons/cinder-csi-plugin/controllerplugin-rbac.yaml
alvaroaleman: Sgtm

addons/default-storage-class/storage-class.yaml
ainmosni: Changed to create a storage class named `cinder-csi` when the external ccm is enabled.

api/pkg/resources/cloudcontroller/openstack.go
ainmosni: Adapted it to use the `MajorMinor` function, which cleans it up a bit.

ainmosni:
@alvaroaleman It's kinda a shame that I can't see the test results myself, then I could see what's breaking that one test.

alvaroaleman:
> Chart kubermatic was modified but its version (1.0.18) was not changed. Please adjust config/kubermatic/Chart.yaml.

/test all

ainmosni:
Ok, bumped the chart. :)

alvaroaleman (approved):
/lgtm

PR #4615, opened 2019-11-01
sync-charts: configure proper version sorting in get_latest_dashboard_tag

kdomanski:
By default git treats any version with a suffix as being newer (greater)
than a version without, e.g. 1.0 < 1.0-alpha.1
This commit will tell git to treat the -rc and -alpha suffix version as older.

chart-sync is broken without it. The latest RC is selected instead of the public releases.

```release-note
NONE
```


alvaroaleman:
As an aside, it feels a little bit like instead of using git to return the right version it might be easier to get all tags and use a small go app to do it.

PR #4623, opened 2019-11-04
Improve error handling in cloud controller / openstack provider

xrstf:
The cloud controller and openstack provider was not very clear about the errors it can encounter. When no external network was defined, for example, you ended up with a simple `"not found"` error and nothing else.

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/cloud/openstack/helper.go
+++ b/api/pkg/provider/cloud/openstack/helper.go
@@ -84,9 +84,9 @@ func getNetworkByName(netClient *gophercloud.ServiceClient, name string, isExter
 	case 1:
 		return candidates[0], nil
 	case 0:
-		return nil, errNotFound
+		return nil, fmt.Errorf("no network named '%s' with external=%v found", name, isExternal)
 	default:
-		return nil, fmt.Errorf("found %d external networks for name '%s', expected one at most", len(candidates), name)
+		return nil, fmt.Errorf("found %d networks for name '%s' (external=%v), expected exactly one", len(candidates), name, isExternal)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


api/pkg/provider/cloud/openstack/helper.go
```diff
@@ -31,10 +31,6 @@ const (
 	resourceNamePrefix = "kubernetes-"
 )
 
-var (
-	errNotFound = errors.New("not found")
```
alvaroaleman, line 35: Do we do any kind of matching with this error or why did we define it like that?

api/pkg/provider/cloud/openstack/provider.go
```diff
@@ -207,37 +207,37 @@ func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
...
-				cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
-					cluster.Spec.Cloud.Openstack.RouterID = router.ID
-					kubernetes.AddFinalizer(cluster, RouterCleanupFinalizer)
-				})
-				if err != nil {
-					return nil, err
-				}
-			} else {
-				return nil, fmt.Errorf("failed to verify that the subnet '%s' has a router attached: %v", cluster.Spec.Cloud.Openstack.SubnetID, err)
+			return nil, fmt.Errorf("failed to verify that the subnet '%s' has a router attached: %v", cluster.Spec.Cloud.Openstack.SubnetID, err)
```
alvaroaleman, line 218: I don't think we can do that, we need to still be able to identify not found. What about defining it as a `type` with a variable message and than asserting the type rather than equality?

api/pkg/provider/cloud/openstack/helper.go
xrstf: Yes we did when searching for the router ID. That's why I replaced the code with my second commit. I was also assuming that since the error variable is not made public, nobody outside of this package asserts on it.

api/pkg/provider/cloud/openstack/provider.go
xrstf: NotFound should now be an empty string, i.e. `err = nil` in this case.

PR #4629, opened 2019-11-05
Add seed-operator controller

xrstf:
This implements more of the seed-cluster reconciliation.

```release-note
NONE
```


api/pkg/controller/operator/seed/controller.go
```diff
@@ -98,6 +106,8 @@ func Add(
 
 	// watch all resources we manage inside all configured seeds
 	for key, manager := range seedManagers {
+		reconciler.seedClients[key] = manager.GetClient()
```
xrstf, line 103: Not having done this before was simply a glaring bug. There's no new logic in this change.

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -15,22 +27,199 @@ import (
...
+	// into the seed cluster.
+	seedCopy := &kubermaticv1.Seed{}
+	name := types.NamespacedName{
+		Name:      seedName,
+		Namespace: r.namespace,
+	}
+
+	if err := client.Get(r.ctx, name, seedCopy); err != nil {
+		if kerrors.IsNotFound(err) {
+			return errors.New("seed cluster has not yet been provisioned and contains no Seed CR yet")
```
xrstf, line 108: Do we want to rely on exponential backoff here or instead return a fixed RequeueAfter?

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,290 @@
...
+				// {{- end }}
+				// {{- if .Values.kubermatic.clusterNamespacePrometheus.disableDefaultScrapingConfigs }}
+				// - -in-cluster-prometheus-disable-default-scraping-configs=true
+				// {{- end }}
+				// {{- if .Values.kubermatic.clusterNamespacePrometheus.scrapingConfigs }}
+				// - -in-cluster-prometheus-scraping-configs-file=/opt/incluster-prometheus-scraping-configs/_custom-scraping-configs.yaml
+				// {{- end }}
+				// {{- if .Values.kubermatic.monitoringScrapeAnnotationPrefix }}
+				// - -monitoring-scrape-annotation-prefix={{ .Values.kubermatic.monitoringScrapeAnnotationPrefix }}
+				// {{- end }}
```
xrstf: I want to defer extending the KubermaticConfiguration with new fields in a future PR. The whole Prometheus integration in master and seed operators is still TODO.

api/pkg/controller/operator/common/resources.go
```diff
@@ -49,17 +59,82 @@ func MasterFilesSecretCreator(cfg *operatorv1alpha1.KubermaticConfiguration) rec
 	}
 }
 
-func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedConfigMapCreatorGetter {
-	return func() (string, reconciling.ConfigMapCreator) {
-		return BackupContainersConfigMapName, func(c *corev1.ConfigMap) (*corev1.ConfigMap, error) {
-			if c.Data == nil {
-				c.Data = make(map[string]string)
+func SeedWebhookServingCertSecretCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedSecretCreatorGetter {
```
alvaroaleman: Would using `pkg/resources/certificates/servingcerthelper.ServingCertSecretCreator` simplify this?

alvaroaleman: Ah I see it also creates the CA. Maybe put the CA in a distinct secret? We could use `./pkg/resources/certificates.GetCACreator` for that

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -15,22 +27,199 @@ import (
...
+
+	if seed.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
+		log.Debugw("ignoring request for wrong worker name", "worker-name", seed.Labels[kubermaticv1.WorkerNameLabelKey])
+		return nil
+	}
+
+	// get pre-constructed seed client
+	client, exists := r.seedClients[seedName]
+	if !exists {
+		log.Debug("ignoring request for existing but uninitialized seed; the controller will be restarted once the kubeconfig is available")
```
alvaroaleman: Nit: Maybe write `reloaded` so ppl do not expect the pod to get restarted?

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -15,22 +27,199 @@ import (
...
+		return nil
+	}
+
+	// find the owning KubermaticConfiguration
+	configList := &operatorv1alpha1.KubermaticConfigurationList{}
+	listOpts := &ctrlruntimeclient.ListOptions{
+		Namespace:     r.namespace,
+		LabelSelector: r.workerNameSelector,
```
alvaroaleman: Nit: Is it necesary to also have the worker-name concept here? For the Cluster CRD it is necesary because they are cluster-scoped, but here the namespace should suffice or WDYT?

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -15,22 +27,199 @@ import (
...
+	}
+
+	if err := r.masterClient.List(r.ctx, configList, listOpts); err != nil {
+		return fmt.Errorf("failed to find KubermaticConfigurations: %v", err)
+	}
+
+	if len(configList.Items) != 1 {
+		log.Debug("ignoring request for namespace without KubermaticConfiguration")
```
alvaroaleman, line 86: Maybe split this up to emit an error log if `len > 1`?

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -15,22 +27,199 @@ import (
...
+		return nil
+	}
+
+	if seed.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
+		log.Debugw("ignoring request for wrong worker name", "worker-name", seed.Labels[kubermaticv1.WorkerNameLabelKey])
+		return nil
+	}
+
+	// get pre-constructed seed client
+	client, exists := r.seedClients[seedName]
```
alvaroaleman: nit: Would it be easier to understand if this was named `seedClient`?

api/pkg/controller/operator/seed/reconciler.go
alvaroaleman: I think the exponential backoff is fine

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
alvaroaleman: Sure, but why leave it as a comment? Its just copied from the chart anyways, right?

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,290 @@
...
+					},
+				},
+			}
+
+			return d, nil
+		}
+	}
+}
+
+func copyKubernetesAddonsContainer(cfg *operatorv1alpha1.KubermaticConfiguration, addonVolume string) corev1.Container {
```
alvaroaleman: This is a bit confusingly named as it doesn't copy anything but instead creates something, right?

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
```diff
@@ -0,0 +1,290 @@
...
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      addonVolume,
+				MountPath: "/opt/addons/",
+			},
+		},
+	}
+}
+
+func copyOpenshiftAddonsContainer(cfg *operatorv1alpha1.KubermaticConfiguration, addonVolume string) corev1.Container {
```
alvaroaleman: Same comment regarding the naming

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
xrstf: Yep, can remove it.

api/pkg/controller/operator/seed/reconciler.go
xrstf: SGTM

api/pkg/controller/operator/common/resources.go
xrstf: Refactored serving cert into two Secrets and made more use of existing helpers.

api/pkg/controller/operator/common/util.go
```diff
@@ -27,6 +30,18 @@ const (
 	ManagedByLabel = "app.kubernetes.io/managed-by"
 )
 
+var (
+	// ManagedByOperatorPredicate is a predicate that matches all resources created by
+	// the Kubermatic Operator, based on the ManagedBy label.
+	ManagedByOperatorPredicate = predicate.Factory(func(m metav1.Object, _ runtime.Object) bool {
```
alvaroaleman, line 37: I think we can just use the ownerRef to figure this out instead of duplicating that info

api/pkg/resources/certificates/triple/triple.go
```diff
@@ -48,6 +48,29 @@ type KeyPair struct {
 	Cert *x509.Certificate
 }
 
+func NewRSAKeyPair(certPEM, keyPEM []byte) (*KeyPair, error) {
```
alvaroaleman: Can we rename this to `ParseRSAKeyPair` as it doesn't create anything `New`?

As an aside, I would like to rename this package (in a different PR), the current name is a remainder of the fact that it originally was a kube package named `triple` which was removed upstream and then upstream removed even more cert-related code which I moved all into this so yeah...

api/pkg/controller/operator/common/util.go
xrstf: But if I already have a label so I can create a label selector, why not re-use the same thing for the predicate?

api/pkg/resources/certificates/triple/triple.go
xrstf: Renamed.

api/pkg/controller/operator/common/util.go
alvaroaleman: Because there shouldn't be more than one mechanism to figure out ownership

xrstf: I changed it to using to the owner refs and I understand why, though I still feel somewhat ... don't know. Now we check if the object is owned by a seed or kubermatic config, but saying that this means that the operator owns it.

api/pkg/controller/operator/master/resources/kubermatic/api.go
```diff
@@ -22,6 +22,12 @@ func apiPodLabels() map[string]string {
 	}
 }
 
+/*
+The openshiftseedsyncer controller is responsible for syncing resources from Openshift userclusters
```
alvaroaleman: I don't think that belongs here

api/pkg/controller/operator/seed/reconciler.go
```diff
@@ -28,9 +41,182 @@ type Reconciler struct {
...
+	// into the seed cluster.
+	seedCopy := &kubermaticv1.Seed{}
+	name := types.NamespacedName{
+		Name:      seedName,
+		Namespace: r.namespace,
+	}
+
+	if err := seedClient.Get(r.ctx, name, seedCopy); err != nil {
+		if kerrors.IsNotFound(err) {
+			return errors.New("seed cluster has not yet been provisioned and contains no Seed CR yet")
```
alvaroaleman, line 108: I may have missed it but we don't seem to watch Seed resources in the seeds so we won't get triggered when this appears

api/pkg/controller/operator/master/resources/kubermatic/api.go
xrstf: Fixed.

api/pkg/controller/operator/seed/reconciler.go
xrstf: Fixed.

alvaroaleman (approved):
/lgtm
/approve

PR #4635, opened 2019-11-06
Remove worker-name handling from the seedsGetter

alvaroaleman:

This PR makes the `seedsGetter` not respect the worker name anymore. The rationale behind that is:
* This was only half-implemented as the `seedGetter`  never respected the worker-name
* Using the `worker-name` breaks our dev workflow where the components are running locally but use the config of the seed, because it would require to have `seed` crd with all correct workername
* We already separate `seeds` via Namespace they are in, having a second mechanism is redudant

Fixes #

**Documentation**:

```release-note
none
```

/assign @xrstf


xrstf:
Love it! The inconsistent worker-name handling confused me a couple of times during the seed validation webhook PRs :-)

/lgtm
/approve

PR #4646, opened 2019-11-08
Add proposal for openshift ingress

alvaroaleman:

Fixes #

**Documentation**:

```release-note
none
```

/assign @thz

### DIFF
--- a/docs/proposals/openshift-ingress.md
+++ b/docs/proposals/openshift-ingress.md
@@ -25,11 +25,20 @@ directed at that DNS name will end up at an Ingress Controller running for the g
 
 The proposed implementation looks like this:
 
-* The human Kubermatic operator must delegate a DNS zone per Kubermatic seed and configure that zone in the seed
-* Kubermatic runs a DNS resolver that will create an A record per ready user cluster node with a name of
-	`*.<<clusterid>>.<<DNS_ZONE>>` that points to the nodes external address
+* The human Kubermatic operator must provide a DNS domain per seed under which kubermatic will set up a wildcard domain per cluster
+* [External DNS](https://github.com/kubernetes-sigs/external-dns/) must be set up in the seed and configured
+	with the customers provider of choice
+* Kubermatic runs a controller that manages a [`DNSEndpoint`](https://github.com/kubernetes-sigs/external-dns/blob/f763d2a4139746abd775c61642cb9e776b387ba6/docs/contributing/crd-source.md) custom resource and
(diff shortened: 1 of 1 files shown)


docs/proposals/openshift-ingress.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Alternatives considered
+
+### Using a service of type LoadBalancer
+
+Depending on the implementation, services of type LoadBalancer may not have an associated DNS record. So
+in order to support this, we would have to create the DNS entry and consequently support any possible DNS
+provider.
+
+Furthermore, services of type LoadBalancer are not available anywhere.
```
thz: "anywhere"? You mean "everywhere", right?

docs/proposals/openshift-ingress.md
```diff
@@ -0,0 +1,58 @@
...
+**Status**: Draft proposal
+
+## Goal
+
+The goal of this proposal is to describe a way to make ingress for Openshift user clusters work without any further configuration.
+
+## Non-Goals
+
+* Setting up certificates for an ingress controller, Openshift creates self-signed ones. Getting valid certificates can be a follow-up.
+* Solve Ingress for Kubernetes clusters, this can be a follow-up.
```
thz, line 14: If I understand the resolver idea correctly, it would be good to state a non-goal to clarify that the accessibility of the ingress by a dns name is not a goal but only to make openshift happy and proceed with what it does to get ingress going.

thz: The comment above is obsolete but:
I propose to use external-dns instead of implementing dns-server within the cluster itself.

alvaroaleman: Updated, PTAL

docs/proposals/openshift-ingress.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+## Implementation
+
+In order to make Ingress work for Openshift by default, we need a wildcard DNS set up in a way that traffic
+directed at that DNS name will end up at an Ingress Controller running for the given cluster.
+
+The proposed implementation looks like this:
+
+* The human Kubermatic operator must delegate a DNS zone per Kubermatic seed and configure that zone in the seed
```
thz: Isn't the dedicated zone at least optional (if not obsolete) with the external-dns idea?

alvaroaleman: I reworded it to "domain", as theoretically there can be other, non kubermatic-related entries in it as long as they do not clash. PTAL

PR #4656, opened 2019-11-11
Use an owner reference instead of an annotation

xrstf:
Originally we wanted to avoid strong ownership, but we cannot really remember why. Also since the operator has been turned into a namespaced thing and Seed CRs now exist, the old weak ownership system via annotations seems like a poor reimplementation of actual ownership references.

This PR replaces the old system with ownership refs.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/operator/master/controller.go
+++ b/api/pkg/controller/operator/master/controller.go
@@ -17,7 +17,6 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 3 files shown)


api/pkg/controller/operator/master/controller.go
```diff
@@ -84,24 +83,18 @@ func Add(
 			return nil
 		}
 
-		owner := a.Meta.GetAnnotations()[common.ConfigurationOwnerAnnotation]
-		if owner == "" {
-			return nil
+		for _, ref := range a.Meta.GetOwnerReferences() {
+			if ref.Kind == dummyConfig.Kind && ref.APIVersion == dummyConfig.APIVersion {
```
xrstf: Re-using `dummyConfig` here is a bit cheeky, but avoids having to hardcode Kind/Version. IMHO it's harmless enough to keep it.

api/pkg/controller/operator/master/controller.go
```diff
@@ -73,9 +72,9 @@ func Add(
 		}
 	})
 
-	obj := &operatorv1alpha1.KubermaticConfiguration{}
-	if err := c.Watch(&source.Kind{Type: obj}, kubermaticConfigHandler, namespacePredicate, workerNamePredicate); err != nil {
-		return fmt.Errorf("failed to create watcher for %T: %v", obj, err)
+	dummyConfig := &operatorv1alpha1.KubermaticConfiguration{}
```
alvaroaleman: This sounds like it was a mock

api/pkg/controller/operator/master/controller.go
alvaroaleman: Shouldn't we just do this based on the namespace? We watch all of them in our NS?

xrstf: Fixed.

api/pkg/controller/operator/master/controller.go
xrstf: Well it IMHO was just a dummy that we abused for its type information. But it's not relevant anymore because of the other change. The variable is now called `cfg` again.

PR #4662, opened 2019-11-11
Fix seed-validation cluster check

xrstf:
For testing the operator, I wanted to create a test seed in my own testing namespace on dev. The current webhook however rejected this Seed:

```yaml
kind: Seed
metadata:
  name: xrstf-europe
  namespace: xrstf
spec:
  country: DE
  location: Hamburg
  datacenters: {}
  kubeconfig:
    apiVersion: v1
    fieldPath: kubeconfig
    kind: Secret
    name: kubeconfig-europe
    namespace: xrstf
```

Error message was:

```
Error from server: error when creating "eu.yaml": admission webhook "seeds.kubermatic.io" denied the request: datacenter "do-fra1" is still in use, cannot delete it
```

The root cause was that the cluster-existence check was just checking *all* clusters against the subject Seed, and of course the subject does not neccessarily have datacenters for all clusters.

```release-note
Fix Seed Validation Webhook rejecting new Seeds in certain situations
```

### DIFF
--- a/api/pkg/validation/seed/seed.go
+++ b/api/pkg/validation/seed/seed.go
@@ -55,30 +55,37 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error 
 	return sv.validate(seed, client, seeds, isDelete)
 }
 
-func (sv *seedValidator) validate(seed *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, existingSeeds map[string]*kubermaticv1.Seed, isDelete bool) error {
-	// this can be nil on fresh seed clusters
-	existingSeed := existingSeeds[seed.Name]
+func (sv *seedValidator) validate(subject *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, existingSeeds map[string]*kubermaticv1.Seed, isDelete bool) error {
+	// this can be nil on new seed clusters
+	existingSeed := existingSeeds[subject.Name]
 
 	// remove the seed itself from the list, so uniqueness checks won't fail
-	delete(existingSeeds, seed.Name)
+	delete(existingSeeds, subject.Name)
 
-	ourDatacenters := sets.NewString()
-	for datacenter := range seed.Spec.Datacenters {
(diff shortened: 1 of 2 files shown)


api/pkg/validation/seed/seed.go
```diff
@@ -55,30 +55,37 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
 	return sv.validate(seed, client, seeds, isDelete)
 }
 
-func (sv *seedValidator) validate(seed *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, existingSeeds map[string]*kubermaticv1.Seed, isDelete bool) error {
-	// this can be nil on fresh seed clusters
-	existingSeed := existingSeeds[seed.Name]
+func (sv *seedValidator) validate(subject *kubermaticv1.Seed, seedClient ctrlruntimeclient.Client, existingSeeds map[string]*kubermaticv1.Seed, isDelete bool) error {
```
xrstf, line 58: Using `subject` as a parameter name makes it IMHO much easier to reason about what Seed this actually is. Especially since `ourDatacenters` confused me when reading [my own :trollface:] code. I was thinking "ours" = "what we have in the system right now".

alvaroaleman (approved):
/lgtm
/approve

api/pkg/validation/seed/seed.go
```diff
@@ -55,30 +55,37 @@ func (sv *seedValidator) Validate(seed *kubermaticv1.Seed, isDelete bool) error
...
-	for datacenter := range seed.Spec.Datacenters {
-		ourDatacenters.Insert(datacenter)
+	// collect datacenter names
+	subjectDatacenters := sets.NewString()
+	existingDatacenters := sets.NewString()
+
+	if !isDelete {
+		// this has no effect on the DC uniqueness check, but makes the
+		// cluster-remaining-in-DC check easier
+		subjectDatacenters = sets.StringKeySet(subject.Spec.Datacenters)
```
alvaroaleman, line 72: TIL'd

PR #4668, opened 2019-11-12
client-sync: remove not-posix parameter from the script

kdomanski:
Apparently the `--suffix` param is a GNU thing and doesn't exist in busybox mktemp.

```release-note
NONE
```


nikhita (approved):
/lgtm

PR #4670, opened 2019-11-12
CI scripts: use hardcoded Github hostkey

kdomanski:
Now we no longer skip hostkey checking when contacting Github.

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/api/hack/ci/ci-sync-apiclient.sh
+++ b/api/hack/ci/ci-sync-apiclient.sh
@@ -2,6 +2,8 @@
 
 set -eu
 
+. $(dirname $0)/../lib.sh
+
 URL="git@github.com:kubermatic/go-kubermatic.git"
 
 commit_and_push() {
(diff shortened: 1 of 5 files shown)


api/hack/lib.sh
```diff
@@ -62,10 +62,28 @@ write_junit() {
 EOF
 }
 
+ensure_github_host_pubkey() {
+  # check whether we already have a known_hosts entry for Github
+  if ssh-keygen -F github.com >/dev/null 2>&1; then
+    echo " [*] Github's SSH host key already present" >/dev/stderr
+  else
+    local github_rsa_hash
```
alvaroaleman: This declaration is never used

api/hack/ci/ci-sync-apiclient.sh
```diff
@@ -15,7 +17,7 @@ commit_and_push() {
 
         git config --local user.email "dev@loodse.com"
         git config --local user.name "Prow CI Robot"
-        git config --local core.sshCommand 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /ssh/id_rsa'
+        git config --local core.sshCommand 'ssh -o CheckHostIP=no -i /ssh/id_rsa'
```
alvaroaleman, line 20: Nit: Instead of defining this as ssh command in every git config, wouldn't it make sense to just set it in `.ssh/config` once and for all?

alvaroaleman (approved):
/lgtm
/approve

Lets see how much burning this creates

api/hack/ci/ci-sync-apiclient.sh
kdomanski: Like mentioned yesterday on slack, I meant for this script to mess with the files on the host as little as possible, so that I could test it locally.

PR #4671, opened 2019-11-12
OpenShift: deploy metrics-server

### LINKED ISSUE #4355: Openshift: Add metrics server
**User Story**

As a user of openshift, I'd like to have some metrics

**Acceptance criteria**

* The metrics server gets deployed for openshift
* It uses the same code as for Kubernetes unless modifications are needed
* It comes up successfully

kron4eg:

Fixes #4355

**Documentation**:

```release-note
openshift: add metrics-server
```

### DIFF
--- a/api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -111,8 +111,10 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 }
 
 func (r *reconciler) ensureAPIServices(ctx context.Context, data reconcileData) error {
-	creators := []reconciling.NamedAPIServiceCreatorGetter{}
 	caCert := triple.EncodeCertPEM(data.caCert.Cert)
+	creators := []reconciling.NamedAPIServiceCreatorGetter{
+		metricsserver.APIServiceCreator(caCert),
+	}
 
 	if r.openshift {
 		openshiftAPIServiceCreators, err := openshift.GetAPIServicesForOpenshiftVersion(r.version, caCert)
@@ -122,11 +124,10 @@ func (r *reconciler) ensureAPIServices(ctx context.Context, data reconcileData) 
 		creators = append(creators, openshiftAPIServiceCreators...)
 	}
 
-	creators = append(creators, metricsserver.APIServiceCreator(caCert))
-
 	if err := reconciling.ReconcileAPIServices(ctx, creators, metav1.NamespaceNone, r.Client); err != nil {
 		return fmt.Errorf("failed to reconcile APIServices: %v", err)
 	}
+
 	return nil
 }
 
--- a/api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -6,18 +6,18 @@ import (
 
 	openshiftresources "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/openshift/resources"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/clusterautoscaler"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/controller-manager"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/dnat-controller"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/kube-state-metrics"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/kubernetes-dashboard"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/machine-controller"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/metrics-server"
+	controllermanager "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/controller-manager"
+	dnatcontroller "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/dnat-controller"
+	kubestatemetrics "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/kube-state-metrics"
+	kubernetesdashboard "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/kubernetes-dashboard"
+	machinecontroller "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/machine-controller"
+	metricsserver "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/metrics-server"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/openshift"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/openvpn"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/prometheus"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/scheduler"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/system-basic-user"
-	"github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/user-auth"
+	systembasicuser "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/system-basic-user"
+	userauth "github.com/kubermatic/kubermatic/api/pkg/controller/user-cluster-controller-manager/resources/resources/user-auth"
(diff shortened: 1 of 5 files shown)


api/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -113,16 +113,17 @@ func (r *reconciler) reconcile(ctx context.Context) error {
...
 		openshiftAPIServiceCreators, err := openshift.GetAPIServicesForOpenshiftVersion(r.version, caCert)
 		if err != nil {
 			return fmt.Errorf("failed to get openshift apiservice creators: %v", err)
 		}
 		creators = append(creators, openshiftAPIServiceCreators...)
-	} else {
-		creators = append(creators, metricsserver.APIServiceCreator(caCert))
 	}
 
+	creators = append(creators, metricsserver.APIServiceCreator(caCert))
```
alvaroaleman: No need for the append, you can just define it in line 114 where the var is initialized

kron4eg: 👍

api/hack/run-controller.sh
```diff
@@ -8,16 +8,17 @@ set -x
 cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
 # Deploy a user-cluster/ipam-controller for which we actuallly
 # have a pushed image
-export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse origin/master)}"
+export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse HEAD)}"
```
alvaroaleman: This needs to stay the way it was, else it will be broken by default as in most cases there won't be an image for HEAD. If you actually need and want that you can override the value by setting the KUBERMATICCOMMIT env var.

kron4eg: Yeah, after thinking about it a bit more, I realized this is very thin ice. Consider this knee-jerk reaction on multiple WTF/second produced by the fact that pods used unexpected image tag :)

Removed.

kron4eg:
metrics-server complains:
```
E1113 20:31:55.085672       1 authentication.go:65] Unable to authenticate the request due to an error: [x509: subject with cn=apiserver-aggregator is not in the allowed list, x509: certificate signed by unknown authority]
```

kron4eg:
/hold

needs additional testing

kron4eg:
/hold cancel

it works

kron4eg:
/assign @alvaroaleman 

PTAL

api/hack/run-api.sh
```diff
@@ -1,31 +1,34 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -o errexit
 set -o nounset
 set -o pipefail
 set -x
 
-make -C $(dirname $0)/.. kubermatic-api
+make -C ./.. kubermatic-api
```
alvaroaleman: Why did you change this? This change means that the script will stop working of the api folder is not the current working directoy

api/hack/run-api.sh
```diff
@@ -1,31 +1,34 @@
...
 
 KUBERMATIC_WORKERNAME=${KUBERMATIC_WORKERNAME:-$(uname -n)}
+KUBERMATIC_SECRETS_DIR=${KUBERMATIC_SECRETS_DIR:-"../../secrets"}
```
alvaroaleman: Please use `$0` as base for relative paths

api/hack/run-api.sh
```diff
@@ -36,4 +39,5 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
   -log-debug=true \
   -log-format=Console \
   -logtostderr \
-  -v=4 $@
+  -v=4 \
+  "$@"
```
alvaroaleman: Doesn't this break the script when passing more than one extra arg?

api/hack/run-userclustercontroller.sh
```diff
@@ -1,64 +1,72 @@
...
+    jq -r '.data.kubeconfig' |
+    base64 -d |
+    yaml2json |
+    jq --arg url "${CLUSTER_URL}" '.clusters[0].cluster.server = $url' >"${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE}"
+echo "Using kubeconfig ${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE}"
 
-SEED_SERVICEACCOUNT_TOKEN="$(kubectl get secret -o json \
-  |jq '.items[]|select(.metadata.annotations["kubernetes.io/service-account.name"] == "kubermatic-usercluster-controller-manager")|.data.token' -r \
-  |base64 -d)"
+OPENVPN_SERVER_SERVICE_RAW="$(kubectl get service openvpn-server -o json)"
```
alvaroaleman: Not re-arranging this would make this pull easier to review :grimacing:

alvaroaleman: And generally it would be nice to create dedicated PRs for this kind of refactoring, as it doesn't really have anything to do with the main purpose of the PR

api/hack/run-userclustercontroller.sh
```diff
@@ -1,64 +1,72 @@
...
 ./_build/user-cluster-controller-manager \
-    -kubeconfig=${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE} \
+    -kubeconfig="${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE}" \
     -metrics-listen-address=127.0.0.1:8087 \
     -health-listen-address=127.0.0.1:8088 \
-    -namespace=${NAMESPACE} \
-    -openvpn-server-port=${OPENVPN_SERVER_NODEPORT} \
-    -cluster-url=${CLUSTER_URL} \
-    -version=${CLUSTER_VERSION} \
+    -pprof-listen-address="${KUBERMATIC_PPROF_ADDRESS}" \
```
alvaroaleman: Same here, if you add new args at the end instead of in the middle, the diff stays way smaller which makes reviewing easier

alvaroaleman:
@kron4eg the RBAC issues were all fixes by allowing the new client name?

api/hack/run-userclustercontroller.sh
kron4eg: I had to fix pprof port, so I had to edit this file and then all hell broke loose:
* https://github.com/mvdan/sh reformatted me the document
* shellcheck (shell linter) screamed in pains. How could I not fix linter complaining

¯\\\_(ツ)\_/¯

api/hack/run-api.sh
kron4eg: Fixed

api/hack/run-api.sh
kron4eg: Fixed

api/hack/run-api.sh
kron4eg: no, it's all good, and even safer: https://github.com/koalaman/shellcheck/wiki/SC2068

kron4eg:
> @kron4eg the RBAC issues were all fixes by allowing the new client name?

Yup, adding missing `apiserver-aggregator` to the clientName fixed the issue with Unauthorized reply from api, when I tried to access metrics.

api/hack/run-userclustercontroller.sh
kron4eg: The reason to stick pprof-listen-address= into the place where it is, is to group it with others -listen-address= parameters.

api/hack/run-userclustercontroller.sh
alvaroaleman: So learn to not use `git -a`? If you have an editorconfig that automatically changes stuff, that should be very much your problem to solve and not the one of your reviewer.

api/hack/run-controller.sh
```diff
@@ -1,38 +1,43 @@
-#!/bin/bash
```
alvaroaleman: This script doesn't work anymore and without having tried, I assume the same is true for all the other scripts. Please remove all changes to the scripts out of this PR and if you care, create a new one. Those do not have anything to do with the metrics server. I tested this with the versions in `master` of the scripts.

alvaroaleman (changes requested):
This still fails to scrape metrics with a `E1122 16:58:48.641179       1 authentication.go:65] Unable to authenticate the request due to an error: [x509: subject with cn=apiserver-aggregator is not in the allowed list, x509: certificate signed by unknown authority]` in the metrics server.

Changing the frontproxy client cert to use `kube-apiserver-proxy` as CN fixes that, indicating something with the aggregration layer is still not set up correctly:

```
diff --git a/api/pkg/resources/apiserver/frontproxy-client-certificate.go b/api/pkg/resources/apiserver/frontproxy-client-certificate.go
index a6aa44f57..dcaecc3ae 100644
--- a/api/pkg/resources/apiserver/frontproxy-client-certificate.go
+++ b/api/pkg/resources/apiserver/frontproxy-client-certificate.go
@@ -15,7 +15,7 @@ type frontProxyClientCertificateCreatorData interface {
 func FrontProxyClientCertificateCreator(data frontProxyClientCertificateCreatorData) reconciling.NamedSecretCreatorGetter {
        return certificates.GetClientCertificateCreator(
                resources.ApiserverFrontProxyClientCertificateSecretName,
-               "apiserver-aggregator",
+               "kube-apiserver-proxy",
                nil,
                resources.ApiserverProxyClientCertificateCertSecretKey,
                resources.ApiserverProxyClientCertificateKeySecretKey,
```

kron4eg:
> This still fails to scrape metrics with a `E1122 16:58:48.641179 1 authentication.go:65] Unable to authenticate the request due to an error: [x509: subject with cn=apiserver-aggregator is not in the allowed list, x509: certificate signed by unknown authority]` in the metrics server.

Deployments in user-cluster namespace needs to be restarted to pickup the change.

alvaroaleman:
> Deployments in user-cluster namespace needs to be restarted to pickup the change.

That should happen automatically, we pick up changes to all mounts by putting a label for each mount with a value that is the resourceVersion of the mounted thing. If this does not happen then that is a bug that needs fixing.

kron4eg:
latest push is just a push force without commits regarding shell scripts

alvaroaleman (approved):
/lgtm
/approve

PR #4674, opened 2019-11-13
Openshift: Manage admin kubeconfig from UCCM

alvaroaleman:

* Moves the management of the ServiceAccount and associated RBAC that created inside Openshift userclusters for the admin kubeconfig from the openshift controller into the resources userclustercontroller, as it has informers inside the usercluster
* Moves the management of the `cluster-admin` kubeconfig in the seed from the openshift controller into the user cluster controller manager in  a new `openshift-seed-syncer` controller that has informers in both the seed and the usercluster
* Fixes the `Revoke Admin Kubeconfig` endpoint to also work for openshift, so far, this has been a complete no-op when Openshift was used

Fixes #

**Documentation**:

```release-note
none
```


api/hack/run-userclustercontroller.sh
```diff
@@ -8,9 +8,23 @@ cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
 make user-cluster-controller-manager
 
 # Getting everything we need from the api
-# TODO: append -n cluster-$CLUSTER_NAME here or switch to the namespace before
-ADMIN_KUBECONFIG_RAW="$(kubectl get secret admin-kubeconfig -o json)"
-CLUSTER_RAW="$(kubectl get cluster $(echo $ADMIN_KUBECONFIG_RAW|jq -r '.metadata.namespace'|sed 's/cluster-//') -o json)"
+# This script assumes you are in your cluster namespace, which you can configure via `kubectl config set-context $(kubectl config current-context) --namespace=<<cluster-namespace>>`
+NAMESPACE="$(kubectl config view --minify|grep namespace |awk '{ print $2 }')"
```
xrstf: I have a similar situation with the run-operator.sh script, where I constantly have to edit the default `-namespace=kubermatic` to `-namespace=xrstf`. That made me think that it would be nice to allow accepting env variables there, so I can do `NAMESPACE=xrstf ./hack/run-operator.sh`.

Maybe the same could be useful here? I never set the namespace shell- or even session-wide.

api/pkg/controller/seed-controller-manager/openshift/openshift_controller.go
```diff
@@ -344,9 +339,13 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
...
 
-	reconcileRequest, err := r.createClusterAccessToken(ctx, osData)
-	if reconcileRequest != nil || err != nil {
-		return reconcileRequest, err
+	reachable, err := r.clusterIsReachable(ctx, osData.Cluster())
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if cluster is reachable: %v", err)
+	}
+	if !reachable {
+		log.Debug("Cluster is not reachable yet, retrying later")
```
xrstf, line 347: `clusterIsReachable` already logged that the cluster is not reachable.

api/pkg/controller/user-cluster-controller-manager/openshift-seed-syncer/doc.go
```diff
@@ -0,0 +1,6 @@
+package openshiftseedsyncer
+
+/*
+The openshiftseedsyncer controller is responsible for syncing resources from Openshift userclusters
+into the seed cluster namespace.
+*/
```
xrstf, line 4: Shouldn't this be a comment on the package declaration itself?

alvaroaleman:
@xrstf all comments addressed, PTAL

alvaroaleman:
/assign @zreigz 

The test failure is due to the `pull-kubermatic-api-e2e` presubmit using go 1.11

alvaroaleman:
/hold
openshift failure is legit

PR #4676, opened 2019-11-13
sync-apiclient: use -exec instead of -execdir which is non-POSIX

kdomanski:
```release-note
NONE
```


alvaroaleman (approved):
/lgtm
/approve

PR #4678, opened 2019-11-14
ci-update-docs: source missing bash lib

kdomanski:
I've switched this script to adding the hardcoded Github host pubkey using a function from `lib.sh`, but forgot to make sure the lib is actually getting sourced.

```release-note
NONE
```


nikhita (approved):
/lgtm

PR #4680, opened 2019-11-14
Move management of the oauth bootstrap password into the UCCM

alvaroaleman:

Moves the management of the oauth bootstrap password from the openshift controller into the usercluster controller manager, as the latter has informers in there.

Fixes #

**Documentation**:

```release-note
none
```


api/pkg/controller/user-cluster-controller-manager/resources/resources/openshift/secrets.go
```diff
@@ -27,3 +33,42 @@ func RegistryServingCert(caCert *triple.KeyPair) reconciling.NamedSecretCreatorG
...
+
+		rawPassword, err := generateNewOAuthSecret()
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate password: %v", err)
+		}
+		hashedPassword, err := bcrypt.GenerateFromPassword([]byte(rawPassword), 12)
+		if err != nil {
+			return nil, fmt.Errorf("failed to hash password: %v", err)
+		}
+		s.Data[OAuthBootstrapSecretName] = hashedPassword
```
xrstf, line 62: For future reference [to myself]: Openshift expects the secret to be a bcrypt hash, even if we don't have the plaintext password anymore. Hashing here does not increase any security whatsoever.

PR #4685, opened 2019-11-14
refactor chart-sync to deal with weeklies

kdomanski:
This commit removes some logic that is obsolete now that we only
synchronize charts on tags. In addition, it allows dealing with
weekly snapshot tags.

```release-note
NONE
```


kdomanski:
/assign @alvaroaleman 

I tested this by pushing tags `v2.12.666-please-ignore` and `weekly-2004-66`. At this time the jobs are still visible in prow history.
image

I've since removed the tags, as well as the jobs' effects in the `kubermatic-installer` repository.

alvaroaleman (approved):
/lgtm
/approve
/retest

PR #4686, opened 2019-11-15
rename SelectProjectCardView into SelectProjectTableView

kgroschoff:
Rename SelectProjectCardView into SelectProjectTableView, as it is more clear for frontend work. Wasn't used before, so renaming will not cause problems

/

**Documentation**:

```release-note
NONE
```


alvaroaleman (approved):
/lgtm
/approve

If anyone created `UserSettings` for themselves already, the setting will be lost.

PR #4688, opened 2019-11-15
Fix nginx chart

xrstf:
This syncs the nginx-ingress-controller chart with what upstream does.

* A bunch of missing permissions have been defined. Not sure if all of them are really required with our feature set, but better be safe than sorry.
* The publish-service stuff has been removed. We don't use it and if someone wants to use it, they can easily get it back via extraArgs.
* Same goes for the tcp/udp configmaps.
* The dnspolicy stuff could be a helpful addition.
* The old naming scheme for RBAC stuff has been harmonized with the other charts and updated to v1.

```release-note
Fix missing nginx-ingress-controller permissions.
ACTION REQUIRED: Removed default publishing of services in nginx-ingress-controller.
```

### DIFF
--- a/config/nginx-ingress-controller/Chart.yaml
+++ b/config/nginx-ingress-controller/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx-ingress-controller
-version: 1.0.3
+version: 1.0.4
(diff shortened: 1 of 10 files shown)


kron4eg (approved):
/lgtm

PR #4690, opened 2019-11-15
upgrade-test: include our lib, to be able to use ensure_github_host_pubkey()

kdomanski:
I switched to the usage of that helper, but yet again forgot to include the file that contains it.

/assign @alvaroaleman 

```release-note
NONE
```


alvaroaleman (approved):
/lgtm
/approve

PR #4695, opened 2019-11-18
add default namespace/cluster roles for addons

zreigz:
**What this PR does / why we need it**: add predefined namespace/cluster roles for addons. Those roles contain `component: userClusterRole` label to distinguish them from the other roles. Those are used to create a binding in the user cluster.

Fixes #

```release-note
add default namespace/cluster roles for addons
```


addons/rbac/cluster-admin.yaml
```diff
@@ -0,0 +1,258 @@
...
+kind: ClusterRole
+metadata:
+  labels:
+    component: userClusterRole
+  name: cluster-admin
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods/attach
```
alvaroaleman: Should we simply use a `*` for all of apiGroups, resources and verbs?

zreigz: done

addons/rbac/cluster-admin.yaml
```diff
@@ -0,0 +1,13 @@
...
+  labels:
+    component: userClusterRole
+  name: cluster-admin
+rules:
+  - apiGroups:
+      - "*"
+    resources:
+      - "*"
+    verbs:
+      - "*"
```
alvaroaleman: So it seems this is incomplete, it also needs a 
```
- nonResourceURLs:
  - '*'
  verbs:
  - '*'
```

I got this from looking at the `cluster-admin` clustterrole. WDYT about trying to label existing built-in roles instead of duplicating it?

addons/rbac/cluster-editor.yaml
```diff
@@ -0,0 +1,238 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
alvaroaleman: What about labeling the existing `editor` clusterrole instead?

addons/rbac/cluster-viewer.yaml
```diff
@@ -0,0 +1,110 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
alvaroaleman: What about using the existing `view` clusterrole instead and labeling it?

addons/rbac/space-editor.yaml
```diff
@@ -0,0 +1,239 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
alvaroaleman: Would you mind to rename to `namespace-editor.yaml` :grimacing: ?

addons/rbac/cluster-admin.yaml
zreigz: yes, but how to extend built-in roles with new labels using addon yaml? This solution is clear and easy to understand for the customer who wants to add own predefined roles

zreigz: I can create a new controller to label built-in cluster roles:
```
var mapFn = handler.ToRequestsFunc(func(o handler.MapObject) []reconcile.Request {
	return []reconcile.Request{
		{NamespacedName: types.NamespacedName{
			Name:      "edit",
			Namespace: "",
		}},
		{NamespacedName: types.NamespacedName{
			Name:      "view",
			Namespace: "",
		}},
		{NamespacedName: types.NamespacedName{
			Name:      "admin",
			Namespace: "",
		}},
	}
})
```
this is what you meant?

alvaroaleman: I was hoping that its possible to just omit the `rules` field and only set the metadata but if that is not possible a controller sounds like a good idea as well.

The main reason why I think that would be better is that we currently have to have a complete list of all apigroups and resources that exist in the apiserver and this may change over time and our rules may sulently rot. So just re-using what is built-in would avoid this problem.

zreigz:
@alvaroaleman PTAL

api/pkg/controller/user-cluster-controller-manager/cluster-role-labeler/cluster_role_labeler.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %v", err)
+	}
+
+	// Watch for changes to ClusterRoles
+	if err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: mapFn}); err != nil {
```
alvaroaleman: Instead of using a map function to enq all three clusterroles we care about whenever any clusterrole that may be completely unrelated changes, wouldn't it make more sense to use `handler.EnqeueRequestForObject` and use a `predicate` in the watch to filter out changes to all other clusterroles as we do not care about them?

addons/rbac/namespace-editor.yaml
```diff
@@ -0,0 +1,239 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
alvaroaleman, line 1: The sebastian suggested to write a simlistic controller that copies all of our roles into all namespaces, so they are not only available in the `kube-system` namespace. WDYT? (But can totally be a follow-up)

zreigz: Yes, I am aware of this, it will be the next step.

api/pkg/controller/user-cluster-controller-manager/cluster-role-labeler/cluster_role_labeler.go
zreigz: done, I hope I did this correct.

api/pkg/controller/user-cluster-controller-manager/cluster-role-labeler/cluster_role_labeler.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %v", err)
+	}
+
+	// Watch for changes to ClusterRoles
+	if err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestForObject{}, predicateutil.ByName("view")); err != nil {
```
alvaroaleman: I don't think this will do what you want it to do, you have to create something similiar to `predicateutil.ByName` that allows more than one name

api/pkg/controller/util/predicate/predicate.go
```diff
@@ -30,3 +30,10 @@ func ByNamespace(namespace string) predicate.Funcs {
 		return m.GetNamespace() == namespace
 	})
 }
+
+// ByName returns a predicate func that only includes objects in the given name
+func ByName(name string) predicate.Funcs {
```
alvaroaleman: Maybe just do it like:
```
func ByName(names ...string) predicate.Funcs{
namesSet := sets.NewString(names...)
return Factory(func(m metav1.Object, r runtime.Object) bool {
    return namesSet.Has(m.GetName()
})

zreigz: great, thanks a lot, I will try this

zreigz:
@alvaroaleman I've applied your comments. I've created a manager only with this controller to test it. It works perfectly fine. Thanks a lot.

alvaroaleman (approved):
/lgtm
/approve

Thanks!

PR #4702, opened 2019-11-18
Introduce accessible addon configurations

### LINKED ISSUE #1691: Store and use addon logos and descriptions
**What would you like to be added? / User Story**: We should be able to store logos and descriptions of all accessible addons that are configured or can be configured in our system. This could be done using config maps for example.

After the data will be available we should use in the dashboard to make the feature look like in the designs.

maciaszczykm:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Fixes https://github.com/kubermatic/dashboard-v2/issues/1691.

**Documentation**:

```release-note
Introduced addon configurations.
```


api/pkg/crd/kubermatic/v1/addon_config.go
```diff
@@ -0,0 +1,66 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// AddonConfigResourceName represents "Resource" defined in Kubernetes
+	AddonConfigResourceName = "addonconfigs"
```
alvaroaleman: All these consts are completely unused or did I miss it?

config/kubermatic/crd/crd-addon-configs.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: addonconfigs.kubermatic.k8s.io
+spec:
+  group: kubermatic.k8s.io
+  names:
+    kind: AddonConfig
+    listKind: AddonConfigList
+    plural: addonconfigs
+    singular: addonconfig
+  scope: Cluster
```
alvaroaleman, line 12: Addons are namespaced, so wouldnt it be better to also make the `addonconfigs` namespaced?

maciaszczykm: Addon configs are global, there will be only 1 config per accessible addon. Addons can differ for each cluster, depending on the user's choice.

api/pkg/crd/kubermatic/v1/addon_config.go
maciaszczykm: Yes, I will remove them for now.

maciaszczykm:
@alvaroaleman PTAL

alvaroaleman:
/hold
```
The CustomResourceDefinition "addonconfigs.kubermatic.k8s.io" is invalid: spec.scope: Invalid value: "Cluster": field is immutable
```

That's a valid failure.

maciaszczykm:
/hold cancel

The issue was caused by the old version of CRD deployed by the previous check, with different scope. I have removed the CRD from the CI cluster and it should pass right now.

PR #4703, opened 2019-11-18
Remove not needed port

bashofmann:
In a recent discussion with @kdomanski it came up that it is not necessary anymore to expose the kubelet port 10250 to the outside.
This PR removes this from the OpenStack, AWS and Azure security groups.

**Documentation**:

```release-note
For new clusters, the Kubelet port 12050 is not exposed publicly anymore
```


alvaroaleman:
thanks!
/lgtm
/approve

alvaroaleman:
I've edited the releasenote to clarify that this only affects new clusters

PR #4704, opened 2019-11-18
Seed Operator: Prometheus Customization

xrstf:
This PR adds the Prometheus  customization options from the Helm chart:

* allow disabling default scraping configs / alerting rules
* allow defining your own configs / rules

In the process of doing this, a couple of more changes had to be made:

* The fields inside the `KubermaticConfiguration` were turned into `runtime.RawExtension`, mainly because this makes editing the KubermaticConfiguration resource much more natural than having a gigantic multiline YAML string.
* Configs and rules are now treated as Go templates. Previously, it was impossible to configure a scraping rule that was restricted to the cluster namespace. Likewise it was difficult to figure out what certificates to use for connecting to the user cluster.
The template makes Cluster, ApiserverTLS, EtcdTLS, APIServerHost and ScrapingAnnotationPrefix available to the user.
* The scraping rule for endpoints has been removed, as there are no endpoints in cluster namespaces.
* Histogram metrics from the APIServer are now ignored by Prometheus. They contribute a good 10k time series (roughly one third of all) and are never used for anything. In preparation of moving away from federation to Thanos, we (Artiom and me) decided to skip these metrics.
The big pro of not using the "federate everything based on recording rules" is that we can re-use seed-level Grafana dashboards for user-cluster stuff, because metric names stay consistent.

```release-note
NONE
```


api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -112,16 +113,16 @@ type KubermaticAddonsConfiguration struct {
...
-type KubermaticSeedMonitoringConfiguration struct {
+// KubermaticUserClusterMonitoringConfiguration can be used to fine-tune to in-cluster Prometheus.
+type KubermaticUserClusterMonitoringConfiguration struct {
 	// DisableDefaultRules disables the recording and alerting rules.
 	DisableDefaultRules bool `json:"disableDefaultRules,omitempty"`
 	// DisableDefaultScrapingConfigs disables the default scraping targets.
 	DisableDefaultScrapingConfigs bool `json:"disableDefaultScrapingConfigs,omitempty"`
 	// CustomRules can be used to inject custom recording and alerting rules.
-	CustomRules string `json:"customRules,omitempty"`
+	CustomRules runtime.RawExtension `json:"customRules,omitempty"`
```
alvaroaleman: What is the reason for using the `runtime.RawExtension` over the string? Doing a basic smoketest by trying to `yaml.Marshal` the rules should also be possible when psasing them as string or not?

The main advantage (and drawback) of using `runtime.RawExtension` is that it can be used to defer unmarshalling, e.G. because the target types are not known to the part of the application that does the initial unmarshaling. But if I read the code correctly, they are never known in our example here

xrstf: Having the rules as a string makes it harder to maintain them, because your Yaml highlighting won't work anymore and you need to be careful with multi-line string indentation stuff.

At first I had it as `interface{}`, but that cannot have DeepCopy functions, so I needed some type for that. It's correct that runtime.RawExtension is kind of abused here, but it seemed simpler than defining a custom dummy type and writing DeepCopy functions manually.

kron4eg:
/lgtm
/hold

Hold to let `runtime.RawExtension` dispute settle.

api/pkg/crd/operator/v1alpha1/configuration.go
xrstf: To illustrate my point :-)

I like this better:
Sublime Text

than this:
Sublime Text

kron4eg: Visually, the former option looks like contents of `customScrapeingConfigs` is part of our API.

alvaroaleman: Something is wrong, I agree with Artiom again.

xrstf: It's strings now.

xrstf:
This is part of #4723.

PR #4705, opened 2019-11-18
Global Monitoring Proposal

xrstf:
This describes the neccessary steps to improve our multi-cluster monitoring.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/global-monitoring.md
+++ b/docs/proposals/global-monitoring.md
@@ -1,0 +1,195 @@
+# Global Monitoring
+
+**Author**: Christoph (@xrstf)
+**Status**: Draft proposal
+
+This is a proposal to create a centralized view for all monitoring data across all seed
+and user clusters. Karma (the centralized Alertmanager dashboard) is a great step in
+that direction, but being able to run queries and see dashboards in a single Grafana is
+still missing.
+
+## Motivation
+
+* Currently, user-cluster monitoring is severly lacking; while some data is federated,
+  we only have a single etcd dashboard that actually uses it. Prometheus is mainly used
+  for alerting right now.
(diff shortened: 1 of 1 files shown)


docs/proposals/global-monitoring.md
```diff
@@ -0,0 +1,195 @@
...
+* **Keep User-Cluster Prometheus stateless.**
+
+  We could give every Prometheus a 10Gi disk, but this would quickly balloon. And
+  balloon unneccesarily because the data is streamed somewhere else anyway in order to
+  eventually make it durable.
+
+* **Be able to scale to accomodate many user clusters.**
+
+  If for example the Seed-Prometheus would scrape all User-Cluster Prometheus instances,
+  it would quickly eat more memory than we're willing to give it. If it then also has to
```
alvaroaleman, line 105: How is "we are willing to give it" defined here?

xrstf: Defined as "on Loodse's test environments" and "more than we should demand from our customers to provision for Prometheus".

xrstf: Specifically in regards to resources, it was written under the assumption that managing more smaller pods is easier in k8s than managing fewer but larger pods. I always "feel bad" for requesting something like 16Gi because our usual instances are much smaller. Are "real-world" cluster usually consisting of instances with 32+ Gi RAM?

alvaroaleman: The question to ask is IMHO if our customers consider this an issue

xrstf: My sample size for feedback is very small and entirely "yes, it uses too much". And that is also a good year old. ^^

kron4eg (approved):
/lgtm

PR #4708, opened 2019-11-19
clean cluster role binding methods

zreigz:
**What this PR does / why we need it**: clean cluster role binding methods

Fixes #

```release-note
NONE
```


api/pkg/handler/test/helper.go
```diff
@@ -69,7 +69,7 @@ const (
 	UserID = "1233"
 	// UserID2 holds a test user ID
 	UserID2 = "1523"
-	// UserName holds a test user name
+	// UserEmail holds a test user name
```
kdomanski: ??

api/pkg/handler/v1/cluster/binding_test.go
```diff
@@ -613,206 +597,3 @@ func TestListClusterRoleBinding(t *testing.T) {
 		})
 	}
 }
-
```
kdomanski, line 616: Why are you deleting all these test?

api/pkg/handler/test/helper.go
zreigz: my IDE refactor to many items automatically

api/pkg/handler/v1/cluster/binding_test.go
kdomanski: Ok, I see now that the relevant endpoints are gone.

zreigz: we don't need those endpoints anymore. They were designed for a different view

zreigz: I forgot to delete them from binding.go I will commit new changes.

zreigz:
@kgroschoff @kdomanski PTAL

kdomanski:
Since the updated `BindUserToClusterRoleEndpoint` now uses `client.Patch`, is it now possible to use this endpoint to update an existing role binding? If yes, then `TestPatchClusterRoleBinding` should probably remain (although updated).

api/pkg/handler/v1/cluster/binding.go
```diff
@@ -920,12 +723,6 @@ func convertInternalRoleBindingsToExternal(roleBindings []rbacv1.RoleBinding) []
 
 func convertInternalClusterRoleBindingToExternal(clusterRoleBinding *rbacv1.ClusterRoleBinding) *apiv1.ClusterRoleBinding {
 	binding := &apiv1.ClusterRoleBinding{
-		ObjectMeta: apiv1.ObjectMeta{
```
kdomanski, line 923: Why is this no longer needed?

zreigz:
> Since the updated `BindUserToClusterRoleEndpoint` now uses `client.Patch`, is it now possible to use this endpoint to update an existing role binding? If yes, then `TestPatchClusterRoleBinding` should probably remain (although updated).

According to the new view we just need to bind/unbind user to some role. I will create a new method to unbind the user. In this case, we don't need `patch` endpoint and tests for it. It would be better to
add more tests for `TestBindUserToClusterRole` to cover scenarios:
- add already existing user
- add the user for existing binding to check is the subject list was expanded.
WDYT?

api/pkg/handler/v1/cluster/binding.go
zreigz: right now the ClusterRoleBinding is created automatically if doesn't exist during binding user to the role. It means Every Role contains exactly one binding which holds users with the role. To find this ClusterRoleBinding I need an only label and role name. The API model doesn't need to hold ClusterRoleBinding  ID and name in this case. We don't display binding name anywhere.

zreigz:
@kdomanski I've added new test scenarios

PR #4715, opened 2019-11-20
Initial rancher server integration

### LINKED ISSUE #4631: Rancher server integration
**User Story**
The idea is to use Rancer server as a management tool for user clusters. Initially, it should be deployed in the user cluster namespace on the seed cluster, and the user cluster should be imported in it automatically.

**Acceptance criteria**
- [x] Rancher server should be deployed in user cluster namespace
- [x] Rancher server UI should accessible externally
- [x] User cluster should be imported into rancher server
- [x] Basic Rancher functionality for imported cluster should work out of the box

moelsayed:
Add initial support for running rancher server UI per user cluster. 
Fixes #4631

Still needs:
~~- Persistent volumes for rancher deployment pod~~
~~- Rancher node agent can't auth with the master.~~

**Does this PR introduce a user-facing change?**: No

```release-note
NONE
```


api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
+package rancherinit
```
alvaroaleman: I think it would be better to rename the controller to just `rancher` without the trailing init, at least supposively it should be able to reconcile the thing (I know that is not 100% true currently)

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	"time"
+
+	"go.uber.org/zap"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+
+	"github.com/rancher/norman/clientbase"
+	normantypes "github.com/rancher/norman/types"
+	rancherv3 "github.com/rancher/types/client/management/v3"
```
alvaroaleman: If anyhow possible I would prefer not to import the types but instead to copy them or define a subset of them, because they have dependencies on Kubernetes which means that we must keep our kube dependency version compatible with the one of rancher which I would prefer to avoid

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+}
+
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	log := r.log.With("request", request)
+	log.Debug("Processing")
+	if request.Name != resources.RancherStatefulSetName {
+		return reconcile.Result{}, nil
```
alvaroaleman: This makes more sense as a predicate, similiar to `github.com/kubermatic/kubermatic/api/pkg/controller/util/predicate.ByNamespace`. Check `pkg/controller/operator/master/controller.go` for a sample on how to do that

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	statefulSet := &appsv1.StatefulSet{}
+	if err := r.Get(ctx, request.NamespacedName, statefulSet); err != nil {
+		if kubeapierrors.IsNotFound(err) {
+			log.Errorw("Couldn't find statefulSet", zap.Error(err))
```
alvaroaleman: This happening is completely normal and should at most be logged on debug level.

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, nil
+	}
+	// wait for statefulSet to be ready
+	if statefulSet.Status.ReadyReplicas == 0 {
+		return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+	if statefulSet.Annotations == nil {
+		statefulSet.Annotations = make(map[string]string)
+	}
+	if statefulSet.Annotations["kubermatic.io/rancher-server-ready"] != "true" { // if we have ready replicas, we should be ready
```
alvaroaleman: If possible, we should not look metadata to check on state, because if the state (here: StatefulSet has ReadyReplicaCount > 0) changes, the metadata won't get changed with it.

So instead of writing something in the metadata, just look up the state. The basic idea behind this is that whenever something about the state changes, your controller gets triggered and it then checks if and what it has to do.

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		statefulSet.Annotations["kubermatic.io/rancher-server-ready"] = "true"
+		if err := r.Update(ctx, statefulSet); err != nil {
+			if kubeapierrors.IsConflict(err) {
+				return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			log.Errorw("failed to update rancher statefulSet", zap.Error(err))
+			return nil, err
```
alvaroaleman: Please always wrap errors, because since go errors are return values they have no stacktrace associated, which means finding out where exactly an error was caused can be pretty hard when it didn't get wrapped.

Also, we commonly do the error log once after the `r.reconcile` call, as that way we can be sure that we log all errors and have no code duplicaiton

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	if port == 0 {
+		return "", fmt.Errorf("Can't find rancher server service nodeport")
+	}
+
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: service.OwnerReferences[0].Name}, cluster); err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("https://%s:%d", cluster.Address.IP, port), nil
```
alvaroaleman: If you use `cluster.Address.ExternalName` instead, you get something that is as a SAN on the cert

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+			port = svcPort.NodePort
+			break
+		}
+	}
+	if port == 0 {
+		return "", fmt.Errorf("Can't find rancher server service nodeport")
+	}
+
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: service.OwnerReferences[0].Name}, cluster); err != nil {
```
alvaroaleman: This will make the controller manager panic if there are no owner refs

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		if err := r.Update(ctx, statefulSet); err != nil {
+			if kubeapierrors.IsConflict(err) {
+				return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			log.Errorw("failed to update rancher statefulSet", zap.Error(err))
+			return nil, err
+		}
+	}
+	// initialize rancher statefulSet
+	if statefulSet.Annotations["kubermatic.io/rancher-server-initialized"] != "true" {
```
alvaroaleman: Pretty much everything this controller does seems to be based on things that are inside the usercluster, wouldn't it make more sense to place it in the usercluster controller manager?

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		if err := r.initRancherServer(ctx, log, statefulSet); err != nil {
+			log.Errorw("failed to initialize Rancher Server", zap.Error(err))
+			return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
alvaroaleman: This should return an error instead of a `RequeueAfter` so the exponential backoff is used

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+		}
+	}
+
+	// Setup the user cluster
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: statefulSet.OwnerReferences[0].Name}, cluster); err != nil {
+		log.Errorw("can't find cluster", zap.Error(err))
```
alvaroaleman: This is not an error, it just means the cluster got deleted and we don't have anything to do anymore

api/pkg/controller/rancherinit/controller.go
```diff
@@ -0,0 +1,462 @@
...
+	if statefulSet.Annotations["kubermatic.io/rancher-server-initialized"] != "true" {
+		if err := r.initRancherServer(ctx, log, statefulSet); err != nil {
+			log.Errorw("failed to initialize Rancher Server", zap.Error(err))
+			return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+	}
+
+	// Setup the user cluster
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: statefulSet.OwnerReferences[0].Name}, cluster); err != nil {
```
alvaroaleman: Accessing the `OwnerReferences` like that may lead to panics

api/pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -222,6 +227,7 @@ func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuth
 		metricsserver.DeploymentCreator(data),
 		usercluster.DeploymentCreator(data, false),
 		kubernetesdashboard.DeploymentCreator(data),
+		// rancherserver.DeploymentCreator(),
```
alvaroaleman: Please remove

api/pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -350,6 +357,7 @@ func (r *Reconciler) ensureConfigMaps(ctx context.Context, c *kubermaticv1.Clust
 func GetStatefulSetCreators(data *resources.TemplateData, enableDataCorruptionChecks bool) []reconciling.NamedStatefulSetCreatorGetter {
 	return []reconciling.NamedStatefulSetCreatorGetter{
 		etcd.StatefulSetCreator(data, enableDataCorruptionChecks),
+		rancherserver.StatefulSetCreator(data),
```
alvaroaleman: I think we should featuregate the whole thing

api/pkg/resources/rancherserver/clusterrolebinding.go
```diff
@@ -0,0 +1,30 @@
...
+		return resources.RancherServerClusterRoleBindingName, func(crb *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, error) {
+			crb.Labels = resources.BaseAppLabel(resources.RancherStatefulSetName, nil)
+
+			crb.RoleRef = rbacv1.RoleRef{
+				APIGroup: rbacv1.GroupName,
+				Name:     "cluster-admin", // nope! should use something with lower privileges
+				Kind:     "ClusterRole",
+			}
+			crb.Subjects = []rbacv1.Subject{{
+				Kind:      rbacv1.ServiceAccountKind,
```
alvaroaleman: This has to be `Kind: User`, the kubeconfigs we generate use x509-based auth and are hence considered a normal user

api/pkg/resources/rancherserver/statefulset.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+	}
+)
+
+// StatefulSetCreator returns the function to reconcile the etcd StatefulSet
+func StatefulSetCreator(data *resources.TemplateData) reconciling.NamedStatefulSetCreatorGetter {
+	return func() (string, reconciling.StatefulSetCreator) {
+		return resources.RancherStatefulSetName, func(set *appsv1.StatefulSet) (*appsv1.StatefulSet, error) {
+			set.Name = resources.RancherStatefulSetName
+			set.Spec.Template.Spec.ImagePullSecrets = []corev1.LocalObjectReference{{Name: resources.ImagePullSecretName}}
```
alvaroaleman, line 39: I don't think this is needed

api/pkg/resources/rancherserver/statefulset.go
```diff
@@ -0,0 +1,139 @@
...
+						{
+							ContainerPort: 80,
+							Protocol:      corev1.ProtocolTCP,
+						},
+					},
+					Resources: defaultResourceRequirements,
+				},
+			}
+
+			diskSize, _ := resource.ParseQuantity("2Gi")
```
alvaroaleman: Just put this into a package-level var and use `MustParse` so we do not have to parse this over and over again

api/pkg/resources/rancherserver/serviceaccount.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+)
+
+// ServiceAccountCreator creates the service account for rancherserver
+func ServiceAccountCreator() (string, reconciling.ServiceAccountCreator) {
```
alvaroaleman: This is completely unused and won't work anyways, as the rancher server is deployed in a the seed cluster but needs permissions for the user cluster, not for the seed.

You need a x509-based kubeconfig instead, check the various `resources.GetInternalKubeconfigCreator` calls in the kubernetes controller for inspiration

api/pkg/resources/rancherserver/statefulset.go
moelsayed: There is a test that ensure that all statefulsets should have this set: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/test/load_files_test.go#L633-L636

api/pkg/controller/rancherinit/controller.go
moelsayed: I am not sure. I am not very failure with different controllers' scope yet, but I think it should be in the seed controller manager, since it manages resources deployed on the seed plane. It's logically similar to addons, monitoring, and openshift controllers.

api/pkg/resources/rancherserver/service.go
```diff
@@ -0,0 +1,47 @@
...
+func ServiceCreator(exposeStrategy corev1.ServiceType) reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		return resources.RancherServerServiceName, func(s *corev1.Service) (*corev1.Service, error) {
+			s.Name = resources.RancherServerServiceName
+			s.Labels = resources.BaseAppLabel(resources.RancherStatefulSetName, nil)
+			if s.Annotations == nil {
+				s.Annotations = map[string]string{}
+			}
+			if exposeStrategy == corev1.ServiceTypeNodePort {
+				s.Annotations["nodeport-proxy.k8s.io/expose"] = "true"
```
alvaroaleman, line 22: I don't think this should be done, we should instead expose the service over an endpoint on our api, as it has valid certs

alvaroaleman: Which pretty much elimantes the purpose of the service as a whole, since port-forwarding can only target pods

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	predicates := predicateutil.ByName(resources.RancherStatefulSetName)
+
+	if err := c.Watch(&source.Kind{Type: &appsv1.StatefulSet{}}, &handler.EnqueueRequestForObject{}, predicates); err != nil {
+		return fmt.Errorf("failed to create watch for statefulsets: %v", err)
+	}
+	return nil
+}
+
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
alvaroaleman, line 91: When accessing the rancher UI it asks me:
```
What URL should be used for this installation? All the nodes in your clusters will need to be able to reach this.
```

Shouldn't we pre-configure that as well?

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	// wait for statefulSet to be ready
+	if statefulSet.Status.ReadyReplicas == 0 {
+		return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+	if statefulSet.Annotations == nil {
+		statefulSet.Annotations = make(map[string]string)
+	}
+	// initialize rancher statefulSet
+	if statefulSet.Annotations["kubermatic.io/rancher-server-initialized"] != "true" {
```
alvaroaleman: Since hypothetically the statefulset could get deleted and re-created and then not have the annotation anymore - Would about tracking this on the cluster object?

Also: Would it be possible to check if we already initialized it via its API instead of tracking it this way?

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,385 @@
...
+}
+
+func randString() string {
+	rand.Seed(time.Now().UnixNano())
+	charset := "abcdefghijklmnopqrstuvwxyz" +
+		"ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"0123456789" +
+		"!@#$%^&*()"
+
+	b := make([]byte, 10)
```
alvaroaleman: Could be a bit longer :P

api/pkg/resources/test/fixtures/deployment-aws-1.10.0-rancher-server.yaml
```diff
@@ -0,0 +1,69 @@
...
+    spec:
+      containers:
+      - args:
+        - --debug
+        - --http-listen-port=80
+        - --https-listen-port=443
+        - --add-local=false
+        - --k8s-mode=embedded
+        env:
+        - name: CATTLE_NAMESPACE
```
alvaroaleman, line 26: What does this do?

api/pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -357,10 +360,14 @@ func (r *Reconciler) ensureConfigMaps(ctx context.Context, c *kubermaticv1.Clust
 }
 
 // GetStatefulSetCreators returns all StatefulSetCreators that are currently in use
-func GetStatefulSetCreators(data *resources.TemplateData, enableDataCorruptionChecks bool) []reconciling.NamedStatefulSetCreatorGetter {
-	return []reconciling.NamedStatefulSetCreatorGetter{
+func GetStatefulSetCreators(data *resources.TemplateData, enableDataCorruptionChecks, ancherServerIntegration bool) []reconciling.NamedStatefulSetCreatorGetter {
```
alvaroaleman: I believe you lost an `r` here

api/pkg/resources/test/fixtures/deployment-aws-1.10.0-rancher-server.yaml
moelsayed: It's used internally by rancher, and it expects to be running it. I didn't set it initially, but it caused issues because the k8s client in rancher was reading the cluster name space from the pod service account configuration, while rancher client was using the default `cattle-system` namespace.

api/pkg/controller/rancher/controller.go
moelsayed: The initialization process is to simply reset the admin user password. So, we can login with the new user password, then rancher is initialized. 

I added a couple of conditions to the cluster object, though.

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,406 @@
...
+
+	statefulSet := &appsv1.StatefulSet{}
+	if err := r.Get(ctx, request.NamespacedName, statefulSet); err != nil {
+		if kubeapierrors.IsNotFound(err) {
+			log.Debugw("couldn't find statefulSet", zap.Error(err))
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get rancher statefulset: %v", err)
+	}
+	// all done.
```
alvaroaleman: Nit: Is this still needed?

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,406 @@
...
+			log.Errorw("failed to initialize Rancher Server", zap.Error(err))
+			return &reconcile.Result{}, err
+		}
+		kubermaticv1helper.SetClusterCondition(cluster, kubermaticv1.ClusterConditionRancherInitialized, corev1.ConditionTrue, "", "Rancher server initialized successfully")
+		if err := r.Update(ctx, cluster); err != nil {
+			return &reconcile.Result{}, err
+		}
+		log.Infow("rancher server statefulSet initialized successfully")
+
+		// }
```
alvaroaleman: Needed?

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,406 @@
...
+	// if statefulSet.Annotations["kubermatic.io/rancher-server-cluster-imported"] == trueStr {
+	// 	return reconcile.Result{}, nil
+	// }
+
+	result, err := r.reconcile(ctx, log, statefulSet)
+	if result == nil {
+		result = &reconcile.Result{}
+	}
+	if err != nil {
+		log.Errorf("failed to reconcile %s: %v", ControllerName, zap.Error(err))
```
alvaroaleman: I think we should add the clusters name here rather than the controller name

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,406 @@
...
+	if statefulSet.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: strings.ReplaceAll(statefulSet.Namespace, "cluster-", "")}, cluster); err != nil {
+		log.Debugw("can't find cluster", zap.Error(err))
+		return nil, nil
+	}
+	// rancher integration is disabled
+	if !cluster.Spec.RancherIntegration {
```
alvaroaleman: Can we re-use the existing `Cluster.Spec.Features` for this if its just a boolean?

api/pkg/controller/rancher/controller.go
```diff
@@ -0,0 +1,406 @@
...
+		if user.Username == RancherUsername {
+			err = client.SetUserPassword(&user, &rancherclient.SetPasswordInput{NewPassword: string(initSecret.Data["password"])})
+			if err != nil {
+				return fmt.Errorf("failed to set rancher user password: %v", err)
+			}
+			break
+		}
+	}
+	// at this point, the rancher statefulSet is initialized and ready to use
+	// statefulSet.Annotations["kubermatic.io/rancher-server-initialized"] = trueStr
```
alvaroaleman: Still needed?

PR #4718, opened 2019-11-20
Reconcile certificate for kubermatic

xrstf:
This adds Certificate reconciling to the Kubermatic Operator. The plan is to reconcile a certificate if we can and wait for someone else to install the cert-manager for us. This means that the operator has to wait until the cert-manager CRDs appear.

I chose to vendor the cert-manager dependencies because it was oh-so easy to do so. Funny enough, our own cert-manager does not yet have these CRDs, so the Operator currently will wait forever until we update cert-manager.

```release-note
NONE
```


xrstf:
/hold

Still need to update cert-manager itself.

api/pkg/controller/operator/certificate/controller.go
```diff
@@ -0,0 +1,151 @@
...
+) error {
+	// wait for the cert-manager CRDs to appear
+	err := wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
+		certCRD := &apiextensionsv1beta1.CustomResourceDefinition{}
+		key := types.NamespacedName{
+			Name: "certificates.cert-manager.io",
+		}
+
+		// must use API reader instead of client because the caches have not been started yet
+		if err := mgr.GetAPIReader().Get(ctx, key, certCRD); err != nil {
```
alvaroaleman: Apart from the crd existing it must also be initialized or sth before it can be used IIRC

xrstf: Apparently not. When changing the code to use the v1alpha1 version we're currently using on dev, as long as the v1alpha1 is registered in our scheme, I can successfully create a Certificate without any issues :-)

```
I1121 12:33:49.597629   23036 ensure.go:79] Created *v1alpha1.Certificate kubermatic in Namespace "xrstf"
2019-11-21T12:33:49.597+0100    debug   groucho.controller-runtime.controller   controller/controller.go:242    Successfully Reconciled {"controller": "kubermatic-master-certificate-operator", "request": "xrstf/test-kubermatic"}
```

xrstf:
I have now merged the new cert controller into the existing master controller, but simply start the watch later after a compatible CRD was found.

Also, the operator now emits fancy events on KubermaticConfiguration and Seed resources.

api/pkg/controller/operator/master/controller.go
```diff
@@ -123,6 +126,49 @@ func Add(
 		}
 	}
 
+	// watch CRDs to detect cert-manager Certificates
+	crdEventHandler := newEventHandler(func(a handler.MapObject) []reconcile.Request {
+		// we only care for the first time we detect a matching CRD; if the CRD is removed the
+		// existing watches break anyway and will crash the controller
```
xrstf: ... at least *I think* this is what should happen.

api/pkg/controller/operator/master/resources/kubermatic/common.go
```diff
@@ -22,20 +24,14 @@ const (
...
 }
 
-func NamespaceCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedNamespaceCreatorGetter {
-	return func() (string, reconciling.NamespaceCreator) {
-		return cfg.Namespace, func(ns *corev1.Namespace) (*corev1.Namespace, error) {
-			return ns, nil
-		}
-	}
-}
-
```
xrstf, line 38: This was just an accidental leftover from when KubermaticConfigurations could reference other namespaces.

api/pkg/controller/operator/master/resources/kubermatic/common.go
```diff
@@ -102,7 +98,8 @@ func IngressCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.N
 
 			i.Spec.TLS = []extensionsv1beta1.IngressTLS{
 				{
-					Hosts: []string{cfg.Spec.Domain},
+					Hosts:      []string{cfg.Spec.Domain},
+					SecretName: certificateSecretName,
```
xrstf, line 102: Let's finally start to use dedicated certs instead of relying on the default-fallback-cert in the nginx chart :-)

kron4eg:
IMHO logic for handling absent cert-manager CRDs is excessive (and we should let operator simply crash and be restarted by k8s), but I don't want to block this.

/lgtm

alvaroaleman:
> IMHO logic for handling absent cert-manager CRDs is excessive (and we should let operator simply crash and be restarted by k8s), but I don't want to block this.

Happens rarely but I very much agree with Artiom here. The operator is not responsible for its dependendencies and should just send a very clear "This is not going to work" signal rather than trying to be smart

alvaroaleman (approved):
/lgtm
/approve

PR #4719, opened 2019-11-20
Fix component health status shown during cluster creation

### LINKED ISSUE #3880: All components are marked as failed on cluster creation
Related to https://github.com/kubermatic/dashboard-v2/issues/717, all components (api server, etcd, machine controller,...) are marked as failed during cluster creation.

There already is some progress with this PR https://github.com/kubermatic/kubermatic/pull/3784. 
Nevertheless as said in the PR comments, there is space to improve.

floreks:
Makes control plane appear as pending instead of failed until the cluster is fully initialized.

Fixes #3880

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/seed-controller-manager/kubernetes/health.go
+++ b/api/pkg/controller/seed-controller-manager/kubernetes/health.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1/helper"
 	"github.com/kubermatic/kubermatic/api/pkg/provider/kubernetes"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cluster) (*kubermaticv1.ExtendedClusterHealth, error) {
@@ -34,7 +37,7 @@ func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cl
 		if err != nil {
 			return nil, fmt.Errorf("failed to get dep health %q: %v", name, err)
 		}
-		*healthMapping[name].healthStatus = status
+		*healthMapping[name].healthStatus = getStatus(status, cluster)
 	}
 
 	var err error
@@ -44,7 +47,7 @@ func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cl
 	if err != nil {
 		return nil, fmt.Errorf("failed to get etcd health: %v", err)
 	}
-	extendedHealth.Etcd = etcdHealthStatus
+	extendedHealth.Etcd = getStatus(etcdHealthStatus, cluster)
 
 	return extendedHealth, nil
 }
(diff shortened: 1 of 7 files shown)


zreigz (approved):
/lgtm
/approve

api/pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -4,10 +4,13 @@ import (
 	"context"
```
alvaroaleman, line 4: What about openshift?

floreks: Is there a separate endpoint for openshift? I thought only `ExtendedClusterHealth` is returned and it does not have openshift specific code.

alvaroaleman: This is not an endpoint, this is part of the `kubernetes` controller. The openshift controller has `syncHealth` for this, which needs a similiar change.

api/pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -60,5 +63,41 @@ func (r *Reconciler) syncHealth(ctx context.Context, cluster *kubermaticv1.Clust
...
+}
+
+func isClusterInitialized(cluster *kubermaticv1.Cluster) bool {
+	isInitialized := cluster.Status.HasConditionValue(kubermaticv1.ClusterConditionClusterInitialized, corev1.ConditionTrue)
+	// If was set to true at least once just return true
+	if isInitialized {
+		return true
+	}
+
+	_, success := helper.ClusterReconciliationSuccessful(cluster)
```
alvaroaleman: Thinking about this, we should also require the `kubermaticv1.ClusterConditionSeedResourcesUpToDate` condition which is excluded from `ClusterReconciliationSuccessful`, as the former checks if our controllers do the right thing and the latter check if the underlying seed realizes it. However from the APIs POV we don't care who does it, we only care if it was done

api/pkg/controller/seed-controller-manager/openshift/openshift_controller.go
```diff
@@ -14,6 +14,7 @@ import (
 	openshiftresources "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/openshift/resources"
 	controllerutil "github.com/kubermatic/kubermatic/api/pkg/controller/util"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1/helper"
```
alvaroaleman: Please alias this as `kubermaticv1helper`, because `helper` is a super generic name and to keep it consistent

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -570,3 +571,11 @@ func (cluster *Cluster) GetSecretName() string {
 	}
 	return ""
 }
+
+func (cluster *Cluster) IsOpenshift() bool {
+	return cluster.Annotations["kubermatic.io/openshift"] != ""
```
alvaroaleman, line 576: Finally someone adds this, we should be using this everywhere instead of duplicating it

api/pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -4,10 +4,13 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	"github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1/helper"
```
alvaroaleman: some comment wrt to aliasing the import

alvaroaleman (approved):
/lgtm
/approve

PR #4729, opened 2019-11-22
shfmt all our shell scripts

kron4eg:
like gofmt but for shell

https://github.com/mvdan/sh#shfmt

```release-note
NONE
```


kron4eg:
As a follow up we probably need to include `make shfmt-check` in pipeline.

alvaroaleman:
@kron4eg maybe bring this up in the dev meeting? Generally it seems there is not much interest thought, so in doubt we should refrain from using this, as it will force everyone to install stuff without a clear benefit

kron4eg:
Yeah, will do.

PR #4732, opened 2019-11-25
Create ClusterIssuers in cert-manager chart

### LINKED ISSUE #4725: Kubermatic Operator: Provider ClusterIssuer for Certificates
**User Story**
The old Helm chart created a Certificate and an Issuer in the kubermatic namespace, so that cert-manager knew how to provide the cert. We chose to instead rely on a ClusterIssuer by default, because:

* it enables admins to configure ACME challenge stuff in one central place
* we did not want to hardwire ACME challenge stuff / DNS config into the KubermaticConfiguration
* it reduces the surface between the operator and the cert-manager
* using an issuer would mean that every Helm chart now needs its own Issuer (and therefore its own config of the acme stuff) because we want to move away from the "one cert for everything! let's rely on nginx fallback default cert mechanism" approach to handling certs per chart/application

So in order for Kubermatic to work, a ClusterIssuer must exist. This means refactoring our cert-manager and certs chart a bit.

This is part of #4723.

**Acceptance criteria**
* [ ] A ClusterIssuer is provided when installing our cert-manager and certs chart. It doesn't matter which of the two charts provides it for now.

xrstf:
For the Kubermatic Operator and for more charts in general, we want to move away from a single cert for everything (which is then also configured as the global nginx fallback cert), towards one cert per app (i.e. kubermatic can get its own cert, so can the keycloak proxies and every other chart). To make this managable, we need s single place to configure the ACME solvers. Up until a few weeks ago there wasn't much configuration for those solvers, but now that we support DNS via Route53, it make sense to ensure we have a single place to configure that.

This PR therefore allows creating ClusterIssuers right in the cert-manager chart, relying on Helm to install the CRD before setting up the ClusterIssuers.

While I'm at it, this PR also includes the compat fixes for cert-manager.

Fixes #4725

```release-note
The cert-manager Helm chart now creates global ClusterIssuers for Let's Encrypt.
```

### DIFF
--- a/config/cert-manager/Chart.yaml
+++ b/config/cert-manager/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: cert-manager
-version: 1.1.4
+version: 1.1.5
 appVersion: v0.12.0-beta.1
 description: A Helm chart for cert-manager
 keywords:
--- a/config/cert-manager/templates/clusterissuer.yaml
+++ b/config/cert-manager/templates/clusterissuer.yaml
@@ -1,0 +1,42 @@
+{{ range $name, $issuer := .Values.certManager.clusterIssuers }}
+---
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
+  name: {{ $name }}
+spec:
+  acme:
+    server: {{ $issuer.server }}
+    email: {{ $issuer.email }}
+    privateKeySecretRef:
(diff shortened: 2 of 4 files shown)


kron4eg (approved):
/lgtm

/hold

I assume this will blow up our dev, so dev should be clean before applying this. Please unhold once ready.

alvaroaleman:
Just a quick test of the cherrypicker....
/cherrypick release/v2.12

PR #4731, opened 2019-11-25
Add remaining important customization options from the Helm chart

xrstf:
**What this PR does / why we need it**:
This updates the reconciling code for master and seed clusters to include all the important (i.e. not replica count or worker count) options.

**Which issue(s) this PR fixes**:
Fixes #4724 

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #4738, opened 2019-11-25
Operator Versioning

### LINKED ISSUE #4726: Kubermatic Operator: Allow configuring image versions via build flags
**User Story**
To use the operator on CI, we need to tell it which version of the kubermatic/api Docker image it's supposed to deploy. I imagine doing this via build flags at compile time, but we could also configure it via whatever mechanism. My thought is that while we could simply patch a KubermaticConfiguration to make the operator deploy a new version, hardcoding it into the binary has the benefit that when we tag a new release, the operator automatically is working for that release. No commit to bump a version needed.

This is part of #4723.

**Acceptance criteria**
* [ ] Compiling the operator spits out a version that deploys the same version of Kubermatic.

xrstf:
This PR implements the versioning for Kubermatic. For this it clarifies that the KubermaticConfiguration *never* contains any version information and that all versions are embedded into the operator binary.

To make the distinction between Seed Docker Repository and User-Cluster Docker Repository more clear, a new `userCluster` option has been added to the KubermaticConfiguration and some things have been moved over there.

Fixes #4726

```release-note
NONE
```


xrstf:
/hold

This will (merge-)conflict with #4731.

api/pkg/controller/operator/common/versions.go
```diff
@@ -0,0 +1,22 @@
...
+	Kubermatic string
+	UI         string
+}
+
+func NewDefaultVersions() Versions {
+	return Versions{
+		Kubermatic: resources.KUBERMATICCOMMIT,
+		UI:         UIVERSION,
+	}
+}
```
xrstf, line 22: At first I had a structure like `resources.TemplateData` in mind, but that would easily lead to people putting stuff in there instead of into the KubermaticConfiguration. That's why I chose to restrict this helper struct to only contain versions. And that's also why it's not an interface, because it's so tiny.

api/pkg/controller/operator/master/defaults.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"go.uber.org/zap"
```
kron4eg, line 8:
```suggestion
```

api/pkg/controller/operator/master/defaults.go
```diff
@@ -5,6 +5,7 @@ import (
...
 
+	"github.com/docker/distribution/reference"
 	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
```
kron4eg, line 9:
```suggestion

	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
```

api/pkg/controller/operator/master/defaults.go
xrstf: Should be fixed now.

kron4eg (commented):
/lgtm

PR #4740, opened 2019-11-25
Create example YAML for KubermaticConfiguration

### LINKED ISSUE #4739: Kubermatic Operator: Create example YAML for KubermaticConfiguration
**User Story**
Just like with Seed CRs we must provide a nice, documented example YAML for KubermaticConfigurations in our documentation.

This is part of #4723.

**Acceptance criteria**
* [ ] There is a (possibly not yet public) example YAML in our documentation.
* [ ] Changes are automatically synced, just like with Seed CRs.

xrstf:
This PR creates a new example KubermaticConfiguration YAML for our documentation. For this it uses the existing defaulting mechanism of the operator. But I noticed that our current approach where we *save* the defaulted config can lead to unexpected results. I wanted to change my domain and was then confused that the Dex URLs did not update (because they were only defaulted once many moons ago).

To prevent this headache, defaulted configurations are not saved anymore. Also, the defaulting logic has been significantly extended to actually cover what our Helm chart did before.

Fixes #4739

```release-note
NONE
```


kron4eg (commented):
/lgtm

PR #4741, opened 2019-11-26
Converge node labelers

alvaroaleman:

* Moves the logic of the `nodelabel` controller into the `node-labeler` controller, as what they do is very similiar
* Adds tests for the logic of the `nodelabeler`
* Removes the node labeling from the container linux operator
* Add a predicate to the container linux operator to ignore non-coreos nodes
* Remove the update predicate from the container linux operator, [the node lease API is in beta and enabled by default since 1.14](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/), so nodes shouldn't be spamming update events anymore
* Updates the distribution label key to not contain the word "kubermatic" to allow better whitelabeling
* Updates all existing distribution label references to the new key

Fixes #

**Documentation**:

```release-note
none
```


addons/csi/nodeplugin.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       serviceAccount: csi-cinder-node-sa
       hostNetwork: true
       nodeSelector:
-        kubermatic.io/distribution: ubuntu
+        x-kubernetes.io/distribution: ubuntu
```
xrstf, line 27: This has been sooo overdue, thanks for finalling fixing it.

PR #4746, opened 2019-11-26
Make inherited labels a status rather than a top-level attribute

alvaroaleman:

Moves the inheritedLabels attribute on the cluster from the top level to status, as Kube objects should only contain spec, status and metadata: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

Fixes #

**Documentation**:

```release-note
none
```
/assign @maciaszczykm

### DIFF
--- a/api/pkg/controller/project-label-synchronizer/project_label_synchronizer.go
+++ b/api/pkg/controller/project-label-synchronizer/project_label_synchronizer.go
@@ -152,7 +152,7 @@ func (r *reconciler) reconcile(log *zap.SugaredLogger, request reconcile.Request
 			}
 			oldCluster := cluster.DeepCopy()
 			cluster.Labels = newClusterLabels
-			cluster.InheritedLabels = getInheritedLabels(project.Labels)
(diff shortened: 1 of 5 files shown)


maciaszczykm:
The change looks good, but it will probably cause null pointers in the frontend. We have to fix it at the same time. Am I right, @floreks?

alvaroaleman:
@maciaszczykm I don't think so, the only api-visible change this introduces is that a nil map will not be serialized with a value of `null` anymore but instead just omitted. I would expect javascript to treat a json field with a value of `null` equaly to an absent field.

maciaszczykm:
@alvaroaleman Just checked the frontend, it should work as we have null checks for inherited labels field. We have to solve it quickly anyways as the frontend feature will not work until we will port it to use inherited labels from status instead.

/lgtm
/approve

floreks:
All good. No need to update frontend as external object sent to the frontend has not changed.

maciaszczykm:
Ah, I see. I looked at wrong structure 😓

PR #4753, opened 2019-11-27
prevent kubermatic namespaces from getting stuck on CI

xrstf:
On bad days (you know, Mondays) sometimes Tiller has a sneeze and fails to either properly install or purge the Kubermatic chart. In these situations the Seed Validation Webhook will block the namespace from being deleted, because the webhook service never comes up.

This PR makes sure that even if the webhook is stuck, the Seed CR is deleted so that at worst we end up with unused, dead webhooks in the CI cluster.

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-run-minimal-conformance-tests.sh
+++ b/api/hack/ci/ci-run-minimal-conformance-tests.sh
@@ -92,6 +92,10 @@ function cleanup {
   helm delete --purge kubermatic-$BUILD_ID  \
     --tiller-namespace=$NAMESPACE
 
+  # Delete our custom seed so that a stuck webhook will not cause the
+  # namespace to be stuck in Terminating state.
+  kubectl delete -n $NAMESPACE seeds $SEED_NAME
+
   # Delete the Helm installation
   kubectl delete clusterrolebinding -l prowjob=$BUILD_ID
   kubectl delete namespace $NAMESPACE --wait=false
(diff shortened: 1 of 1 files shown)


api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -475,7 +479,7 @@ git checkout ${GIT_HEAD_HASH}
 build_tag_if_not_exists
 
 echodate "Installing current version of Kubermatic"
-retry 3 helm upgrade --install --force --wait --timeout 300 \
+retry 3 helm upgrade --install --force --atomic --timeout 300 \
```
alvaroaleman, line 482: Why remove the wait? What does the atomic do? IIRC it prevents retrying from working, which is worse than the status quo

xrstf:
```
      --atomic                   if set, upgrade process rolls back changes made in case of failed upgrade, also sets --wait flag
```

Cannot say anything about the retrying, but wouldn't know of the top of my head why atomic should prevent retries.

alvaroaleman: hm okay, I can not find the issue anymore, lets try it, if it doesn't work we can still rollback
/lgtm
/approve

PR #4756, opened 2019-11-27
Enable external CCM for Opentstack on cluster creation when supported

### LINKED ISSUE #4730: Openstack: move to out-of-tree cloud provider
**User Story**
As a part of out effort to get out out of in-tree cloud providers, we will start with switching Openstack to https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager.md

**Acceptance criteria**
- [ ] For Openstack cluster, we no longer configure the built-in cloud provider in new clusters
- [ ] Instead, we deploy https://github.com/kubernetes/cloud-provider-openstack
- [ ] If possible, all existing clusters are migrated from using the built-in to out-of-tree without destroing volumes or LBs
  - [ ] If that is not possible, then in-tree  provider configuration is preserved for existing clusters and they are marked to prevent breakage through auto-migration

alvaroaleman:

Enables Opentstack external CCM by default on creation of 1.16+ clusters.

Fixes #4730 

**Documentation**:

```release-note
Openstack: New Kubernetes 1.16+ clusters use the external Cloud Controller Manager and CSI by default
```

### DIFF
--- a/api/pkg/crd/kubermatic/v1/cluster.go
+++ b/api/pkg/crd/kubermatic/v1/cluster.go
@@ -93,7 +93,7 @@ type ClusterSpec struct {
 	OIDC OIDCSettings `json:"oidc,omitempty"`
 
 	// Feature flags
-	Features map[string]bool `json:"features"`
+	Features map[ClusterFeature]bool `json:"features"`
 
 	// Openshift holds all openshift-specific settings
 	Openshift *Openshift `json:"openshift,omitempty"`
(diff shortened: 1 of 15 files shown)


api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -103,6 +105,13 @@ type ClusterSpec struct {
 	AuditLogging *AuditLoggingSettings `json:"auditLogging,omitempty"`
 }
 
+const (
+	// ClusterFeatureExternalCloudProvider describes the external cloud provider feature. It is
+	// only supported on a limited set of providers for a specific set of Kube versions. It must
+	// not be set if its not supported.
+	ClusterFeatureExternalCloudProvider = "externalCloudProvider"
+)
+
```
xrstf, line 114: For the operator I put all feature flags from all packages into https://github.com/kubermatic/kubermatic/blob/master/api/pkg/features/features.go -- maybe this one should go there, too?

xrstf: No idea why I forgot to move the existing constant for this feature over to the features.go as well :-D

alvaroaleman: Well I put it there because its part of the CRD which AFAIK is not the case for the other feature flags

PR #4757, opened 2019-11-27
add controller to clone roles for all namespaces

zreigz:
**What this PR does / why we need it**: a new controllers clones roles for all namespaces and new namespace

Fixes #

```release-note
a new controllers clones roles for all namespaces and new namespace
```

### DIFF
--- a/api/pkg/controller/user-cluster-controller-manager/role-cloner/role_cloner.go
+++ b/api/pkg/controller/user-cluster-controller-manager/role-cloner/role_cloner.go
@@ -89,29 +89,38 @@ func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 func (r *reconciler) reconcile(log *zap.SugaredLogger, role *rbacv1.Role) error {
 	oldRole := role.DeepCopy()
 
+	namespaces := []string{}
 	namespaceList := &corev1.NamespaceList{}
 	if err := r.client.List(r.ctx, namespaceList); err != nil {
 		return fmt.Errorf("failed to get namespaces: %v", err)
 	}
 
+	for _, n := range namespaceList.Items {
+		// This NS is the authoritative source of roles we configure
+		if n.Name == v1.NamespaceSystem {
+			continue
+		}
+		namespaces = append(namespaces, n.Name)
+	}
+
 	if oldRole.DeletionTimestamp != nil {
-		if kuberneteshelper.HasFinalizer(oldRole, kubermaticapiv1.UserClusterRoleCleanupFinalizer) {
-			for _, namespace := range namespaceList.Items {
-				if oldRole.Namespace != namespace.Name {
-					if err := r.client.Delete(r.ctx, &rbacv1.Role{
-						ObjectMeta: v1.ObjectMeta{
-							Name:      oldRole.Name,
-							Namespace: namespace.Name,
-						},
-					}); err != nil {
-						if kerrors.IsNotFound(err) {
-							continue
-						}
-						return fmt.Errorf("failed to delete role: %v", err)
-					}
+		if !kuberneteshelper.HasFinalizer(oldRole, kubermaticapiv1.UserClusterRoleCleanupFinalizer) {
+			return nil
+		}
+		for _, namespace := range namespaces {
+			if err := r.client.Delete(r.ctx, &rbacv1.Role{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      oldRole.Name,
+					Namespace: namespace,
+				},
+			}); err != nil {
+				if kerrors.IsNotFound(err) {
+					continue
 				}
+				return fmt.Errorf("failed to delete role: %v", err)
 			}
 		}
+
 		kuberneteshelper.RemoveFinalizer(oldRole, kubermaticapiv1.UserClusterRoleCleanupFinalizer)
 		if err := r.client.Update(r.ctx, oldRole); err != nil {
 			return fmt.Errorf("failed to update role: %v", err)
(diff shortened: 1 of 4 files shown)


api/pkg/controller/user-cluster-controller-manager/role-cloner/namespace/namespace_role_cloner.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %v", err)
+	}
+
+	// Watch for changes to Namespace
+	if err = c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForObject{}); err != nil {
```
alvaroaleman: It also needs to watch for changes to roles and either:

* Enqeue all namespaces if its one of the roles we created
* Enqeue the namespace the role is in

api/pkg/controller/user-cluster-controller-manager/role-cloner/namespace/namespace_role_cloner.go
```diff
@@ -0,0 +1,111 @@
...
+	return reconcile.Result{}, err
+}
+
+func (r *reconciler) reconcile(log *zap.SugaredLogger, namespace *corev1.Namespace) error {
+	roleList := &rbacv1.RoleList{}
+	if err := r.client.List(r.ctx, roleList, ctrlruntimeclient.MatchingLabels{cluster.UserClusterComponentKey: cluster.UserClusterRoleComponentValue}, ctrlruntimeclient.InNamespace("kube-system")); err != nil {
+		return fmt.Errorf("failed to get roles: %v", err)
+	}
+
+	for _, role := range roleList.Items {
```
alvaroaleman: This is probably a bit too simple, as it won't reflect any changes made to the role.  What about using [the normal approach we have to reconcile resources](https://github.com/kubermatic/kubermatic/blob/master/docs/resource-handling.md#example-implementation) and just copying over the name, the labels and the `Rules`?

api/pkg/controller/user-cluster-controller-manager/role-cloner/role/role_cloner.go
```diff
@@ -0,0 +1,117 @@
+package role
```
alvaroaleman: I don't really understand what differentiates this controller from the `namesapce_role_cloner` other than it watching roles instead of namespaces...?

alvaroaleman (commented):
Additionally, I think we need to also clone ClusterRoles or did you want to do that in a follow-up?

api/pkg/controller/user-cluster-controller-manager/role-cloner/role/role_cloner.go
zreigz: `role_cloner` watches role with a specific label and make a copy for all other namespaces. When you create a new namespace the roles with a specific label are not created in this new namespace. The `namespace_role_cloner` takes care to do this. I've tested it locally and works fine. The logic is spread for `Roles` and `Namespaces` which makes it clear and simple.

Why we need clone ClusterRoles? Clone where?

api/pkg/controller/user-cluster-controller-manager/role-cloner/namespace/namespace_role_cloner.go
zreigz: The `clone_role_controller` does this

api/pkg/controller/user-cluster-controller-manager/role-cloner/namespace/namespace_role_cloner.go
zreigz: The `clone_role_controller` does this

zreigz: oh no, this is missing. The same situation with deleting the Role.

zreigz:
The question is if I should go with this approach and just update `role_cloner` to support updating/deleting roles? Making copy works fine.

api/pkg/controller/user-cluster-controller-manager/role-cloner/namespace/namespace_role_cloner.go
zreigz: Extending the reconcile method for the update is simple. But deleting is not. How to distinguish between deleting namespace where all roles are deleted and deleting the single role. For the `clone_role_controller` is the same. Deleting namespace will delete all roles from other namespaces :(

zreigz:
@alvaroaleman I've changed approach as you suggested. PTAL

api/pkg/controller/user-cluster-controller-manager/role-cloner/role_cloner.go
```diff
@@ -0,0 +1,187 @@
...
+func (r *reconciler) reconcile(log *zap.SugaredLogger, role *rbacv1.Role) error {
+	oldRole := role.DeepCopy()
+
+	namespaceList := &corev1.NamespaceList{}
+	if err := r.client.List(r.ctx, namespaceList); err != nil {
+		return fmt.Errorf("failed to get namespaces: %v", err)
+	}
+
+	if oldRole.DeletionTimestamp != nil {
+		if kuberneteshelper.HasFinalizer(oldRole, kubermaticapiv1.UserClusterRoleCleanupFinalizer) {
```
alvaroaleman: Nit: Instead of nesting this deeper and deeper, another option is to negate the condition and use `continue` to avoid arrow-style code:

```
if oldRole.DeletionTimestamp != nil {
    if !kuberneteshelper.HasFinalizer(oldRole, kubermaticapiv1.UserClusterRoleCleanupFinalizer) {
        continue
    }
[...]
}
```

also instead of doing the `if oldRole.Namespace != namespace.Name` it would be easier to understand by removing it from the `NamespaceList` and adding a comment "This NS is the authoritative source of roles we configure" or sth the like

zreigz: done

alvaroaleman (approved):
/lgtm
/approve

Issue #3213, opened 2019-04-15
Move VPA handling into a dedicated controller + Trigger updates


mrIncompetent:
**User Story**
One of the drawbacks of the VPA is, that it actively kills Pods to apply the resource updates.
At the moment we restrict the VPA to only kill pods if there exist at least 2 replicas of it - Which has the drawback that we almost never update pods with 1 replicas.

A potential solution would be to have a dedicated controller which monitors the VPA objects +  the VPA targets (Deployments/StatefulSets). As soon as the controller detects a significant(10%) difference between the VPA recommendation (`vpa.status.recommendation.["container-name"].target`) and the resource request mentioned in the target (Deploymen/StatefulSet), it modifies the target to trigger a RollingUpdate.

On top we move all VPA reconciling into this controller - to avoid having to reconcile the VPA objects across multiple other controllers.

**Acceptance criteria**
- [ ] All VPA objects get reconciled by a new controller (`vpa_controller`)
- [ ] The VPA controller updates Deployments & StatefulSets in case it seees a significat difference between the VPA recommendation and the current configured resources

PR #4763, opened 2019-11-27
Implement the vpa-updater controller

xmudrii:

This PR implements the `vpaupdater` controller responsible for reconciling the VPA resources and setting the resource requirements on Deployments and StatefulSets based on the provided recommendations.

The controller takes the following steps for each VPA resource:
1. Get the appropriate Deployment or StatefulSet
1. Check is there an `updated-by-vpa` label
    * If the label exists, parse it to get the last updated time. The controller upscales resources only once in 30 seconds and downscales only once in 5 minutes. This ensures that the pods will not be constantly rotated and that the VPA recommendations are stable.
1. Parse the VPA resource and update requirements for given Deployment and StatefulSet
    * Right now, we **only** set the memory requirements. The reason for this is that the CPU requirements change more often causing resources to be rotated more often. This can be problematic for components such as API server and etcd, as rotating them can cause other components (mainly controllers) to go down.
1. Update the `updated-by-vpa` label with new values
1. Trigger the update

The `updated-by-vpa` label contains JSON encoded struct containing the resource requirements set by the controller, as well as, the time when the last update happened. This label is used by other controllers that are responsible for Deployments/StatefulSets, so the new resource requirements are not overridden with default values by responsible controllers. See #4810 for more details.

Addresses part of #3213
It is up to be decided what we will do with the CPU requirements.

* This feature is up to be tested on the OpenShift clusters.
* It is up to be tested how the controller behaves when ComponentOverrides are being used.
* It is up to be tested how the controller behaves where there is a higher load on the components
* What is going to happen when VPA is not deployed? I assume that the controller would be enabled, but it will do nothing as there are no VPA resources to be reconciled. This should be tested somehow.
* The VPA resources are not created for the UserClusterController and Kubernetes Dashboard.

**Documentation**:
We should eventually document this feature, but I'm not sure where that should be done.

```release-note
Implement the vpaupdater controller. The controller is responsible for reconciling the VPA resources and applying the recommendations on Deployments and StatefulSets.
```

### DIFF
--- a/api/cmd/seed-controller-manager/controllers.go
+++ b/api/cmd/seed-controller-manager/controllers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/rancher"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/seedresourcesuptodatecondition"
 	updatecontroller "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/update"
+	"github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-manager/vpaupdater"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/features"
 	"github.com/kubermatic/kubermatic/api/pkg/version"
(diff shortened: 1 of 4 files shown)


xmudrii:
/hold
until I don't test this properly

xmudrii:
The PR is ready for review. The E2E tests are now green as well. I'd prefer to keep this PR on hold until I don't perform a few more tests. I've updated the PR description to match the current state of the PR.

/assign @xrstf 
/unassign @alvaroaleman 
/uncc @alvaroaleman

xmudrii:
Closing this PR to switch to the fork, so we don't break Dashboard E2E tests.
/close

PR #4767, opened 2019-11-27
Kubermatic Operator: Reconcile SeedAdmissionWebhook

### LINKED ISSUE #4764: Kubermatic Operator: Seed Webhook Reconciling
**User Story**
The operator is not yet creating the Seed Validation webhook. It should though.

This is part of #4723.

**Acceptance criteria**
* [ ] The ValidatingWebhookConfiguration for the Seed Webhook is reconciled in master and all seed clusters.

xrstf:
This adds reconciling for the Seed validation webhook. The webhook must exist in master and seed clusters, but we support shared master/seed clusters, so we must ensure no naming conflicts arise from this. I had basically two options:

1. Create two Webhooks, just like with ClusterRoleBindings. But that is super akward and confusing.
2. Create a single Webhook, but try to be smart where it points. This is what has been implemented: If the master-controller-manager is found, it is the webhook's target, otherwise it's the seed-controller-manager.

To ensure that webhooks only work on their "own" namespace, a namespace label selector is used and that's the reason why we now reconcile the clearly already namespace.

Fixes #4764

```release-note
NONE
```


api/pkg/controller/operator/common/resources.go
```diff
@@ -101,3 +130,137 @@ func SeedWebhookServingCertSecretCreator(cfg *operatorv1alpha1.KubermaticConfigu
 
 	return servingcerthelper.ServingCertSecretCreator(caGetter, SeedWebhookServingCertSecretName, seedWebhookCommonName, altNames, nil)
 }
+
+func seedAdmissionWebhookName(cfg *operatorv1alpha1.KubermaticConfiguration) string {
+	return fmt.Sprintf("kubermatic-seeds-%s", cfg.Namespace)
```
xrstf, line 134: This is slightly akward and in a perfect world I would not be creating such a stuttery name ("kubermatic-seeds-kubermatic" for installations inside the "kubermatic" namespace). If this was my project, I would not append the "-kubermatic" suffix for the "default case", just so it reads neatly.

api/pkg/controller/operator/common/resources.go
```diff
@@ -101,3 +130,137 @@ func SeedWebhookServingCertSecretCreator(cfg *operatorv1alpha1.KubermaticConfigu
...
+// master has precedence.
+func determineWebhookServiceSelector(cfg *operatorv1alpha1.KubermaticConfiguration, client ctrlruntimeclient.Client) (map[string]string, error) {
+	deployment := appsv1.Deployment{}
+	key := types.NamespacedName{
+		Name:      MasterControllerManagerDeploymentName,
+		Namespace: cfg.Namespace,
+	}
+
+	err := client.Get(context.Background(), key, &deployment)
+	if err == nil {
```
alvaroaleman, line 230: Can we test on the specific errors, i.E ` if err != nil && !kerrors.IsNotFound(err) { return err }`?

api/pkg/controller/operator/common/resources.go
```diff
@@ -4,31 +4,60 @@ import (
...
+	seedWebhookCommonName                 = "seed-webhook"
+	seedWebhookServiceName                = "seed-webhook"
+	IngressName                           = "kubermatic"
+	MasterControllerManagerDeploymentName = "kubermatic-master-controller-manager"
+	SeedControllerManagerDeploymentName   = "kubermatic-seed-controller-manager"
+
+	OpenIDAuthFeatureFlag = "OpenIDAuthPlugin"
 )
 
+func LabelledNamespaceCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedNamespaceCreatorGetter {
```
alvaroaleman: Can we just name this `NamespaceCreator`? We don't e.G. name the APIServerCreator `APIServerWithOpenVPNSideCarAndSomeLabelsOnIt` creator either, thats an implementation detail that doesn't really have much to do with what it is.

When I read this initially I thought it was a func that takes a labelmap and then applies namespace to a Namespace

api/pkg/controller/operator/common/resources.go
```diff
@@ -101,3 +130,137 @@ func SeedWebhookServingCertSecretCreator(cfg *operatorv1alpha1.KubermaticConfigu
...
+		}
+	}
+}
+
+func SeedAdmissionServiceCreator(cfg *operatorv1alpha1.KubermaticConfiguration, client ctrlruntimeclient.Client) reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		return seedWebhookServiceName, func(s *corev1.Service) (*corev1.Service, error) {
+			s.Spec.Type = corev1.ServiceTypeClusterIP
+
+			if len(s.Spec.Ports) < 1 {
```
alvaroaleman: Can we change this to `!= 1` so we also catch cases where ppl add something?

api/pkg/controller/operator/common/resources.go
```diff
@@ -101,3 +130,137 @@ func SeedWebhookServingCertSecretCreator(cfg *operatorv1alpha1.KubermaticConfigu
...
+			NameLabel: MasterControllerManagerDeploymentName,
+		}, nil
+	}
+
+	key = types.NamespacedName{
+		Name:      SeedControllerManagerDeploymentName,
+		Namespace: cfg.Namespace,
+	}
+
+	err = client.Get(context.Background(), key, &deployment)
```
alvaroaleman, line 241: Similarly, please make a precise assertion on the error rather than just "is nil or is not nil"

api/pkg/controller/operator/common/resources.go
xrstf: > When I read this initially I thought it was a func that takes a labelmap and then applies namespace to a Namespace

Well technically all this creator ever does it labelling stuff, which is what I wanted to emphasize with the `Labelled..` prefix (so people don't think the creator is useless because it "creates its own namespace").

But eh, yeah, can change it.

xrstf:
All feedback addressed.

api/pkg/controller/operator/common/resources.go
```diff
@@ -4,31 +4,61 @@ import (
...
+	SeedAdmissionWebhookName              = "kubermatic.io-seeds"
+	SeedWebhookServingCASecretName        = "seed-webhook-ca"
+	SeedWebhookServingCertSecretName      = "seed-webhook-cert"
+	seedWebhookCommonName                 = "seed-webhook"
+	seedWebhookServiceName                = "seed-webhook"
+	IngressName                           = "kubermatic"
+	MasterControllerManagerDeploymentName = "kubermatic-master-controller-manager"
+	SeedControllerManagerDeploymentName   = "kubermatic-seed-controller-manager"
+
+	OpenIDAuthFeatureFlag = "OpenIDAuthPlugin"
```
alvaroaleman: Can we put this in some kind of api package, maybe by moving `cmd/seed-controller-manager/features.go`? Right now this is defined in a total of three places and it must be identical in all of those

xrstf: Yeah, I've now collected all feature flags and put them in the `pkg/features` package.

alvaroaleman (approved):
/lgtm
/approve

PR #4768, opened 2019-11-27
Correctly return token-based kubeconfigs via API

alvaroaleman:

* Makes the API correctly return the token-based kubeconfig
* Removes all GetKubeconfig and RevokeKubeconfig functions from the userClusterConnectionProvider and instead defines them directly in the ClusterProvider because the UserClusterConnectionProvider is used by various other components of Kubermatic, none of them need this functions, they are only used in the API

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/pkg/cluster/client/client.go
+++ b/api/pkg/cluster/client/client.go
@@ -4,14 +4,11 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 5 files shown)


zreigz:
beautiful PR, thanks
/lgtm
/approve

PR #4773, opened 2019-11-28
Addon controller: Only react to cluster update that change the reconciliation success condition

alvaroaleman:

Currently, we always enqeue all addons for a given cluster whenever we get an event for it. Since we started setting a condition per controller on the cluster, we get a lot of cluster events and hence do a lot of addon reconciliation, even thought nothing that is of relevance for the addons changed.

Fixes #

**Documentation**:

```release-note
none
```


xrstf:
/lgtm

Very nice solution :)

PR #4775, opened 2019-11-28
Make seed-proxy controller work in its own namespace

xrstf:
It's pretty bad for testing Kubermatic when the seed-proxy controller always operates in the same, fixed namespace. This PR changes:

* Master
  * Namespace for all resources is taken from the Seed CR instead of the hardcoded `kubermatic`.
* Seed
  * The ServiceAccount was moved from `kube-system` into the Seed CR's namespace.
  * The RBAC stuff in the `monitoring` namespace is now suffixed with the Seed CR's namespace, so the Role is now named `seed-proxy-kubermatic` for example.

Old resources are automatically removed after their new replacements have been provisioned.

```release-note
Fix seed-proxy controller not working in namespaces other than `kubermatic`.
```

### DIFF
--- a/api/pkg/controller/seed-proxy/reconciler.go
+++ b/api/pkg/controller/seed-proxy/reconciler.go
@@ -344,7 +344,7 @@ func (r *Reconciler) reconcileMasterGrafanaProvisioning(seeds map[string]*kuberm
 func (r *Reconciler) deleteResource(client ctrlruntimeclient.Client, name string, namespace string, obj runtime.Object) error {
 	key := types.NamespacedName{Name: name, Namespace: namespace}
 
-	if err := client.Get(context.Background(), key, obj); err != nil {
+	if err := client.Get(r.ctx, key, obj); err != nil {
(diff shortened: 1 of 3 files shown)


api/pkg/controller/seed-proxy/reconciler.go
```diff
@@ -337,3 +340,21 @@ func (r *Reconciler) reconcileMasterGrafanaProvisioning(seeds map[string]*kuberm
 
 	return nil
 }
+
+func (r *Reconciler) deleteResource(client ctrlruntimeclient.Client, name string, namespace string, obj runtime.Object) error {
+	key := types.NamespacedName{Name: name, Namespace: namespace}
+
+	if err := client.Get(context.Background(), key, obj); err != nil {
```
kron4eg: tiny notice: probably should be
```suggestion
	if err := client.Get(r.ctx, key, obj); err != nil {
```

kron4eg (approved):
/lgtm

PR #4780, opened 2019-11-29
Do not GC seed-proxy secrets in foreign namespaces

xrstf:
#4775 forgot to also restrict the GC to a given namespace. With two Kubermatic namespaces in the same cluster, the namespaces were constantly deleting each others' secrets.

```release-note
NONE
```


kron4eg (commented):
/lgtm
/approve

PR #4782, opened 2019-11-29
Wait until UCCM is up before deploying monitoring resources

alvaroaleman:

Currently, we check in the monitoring controller if the UCCM is down and only then skip reconciliation. However we started to initially put everything into the "provisioning" state, which is why it is neither down nor up, which means that we start deploying the monitoring resources before the usercluster is ready and RBAC exists, which makes them start off crashlooping.

Fixes #

**Documentation**:

```release-note
none
```


xmudrii (approved):
/lgtm

Issue #4723, opened 2019-11-22
Kubermatic Operator


xrstf:
**User Story**
This epic tracks the progress for getting the Kubermatic Operator implemented and replacing the Helm chart.

**Acceptance criteria**

PR #4789, opened 2019-12-02
Ignore seed validation requests for other namespaces

xrstf:
Seed validation can happen either in the master-controller-manager or the seed-controller-manager, depending on whether or not we're in a master or seed cluster (on shared master/seed clusters, the master-controller-manager has precedence).

* The master-controller has a seedKubeconfigGetter that is capable of working in any namespace, because it takes the namespace and name of the kubeconfig reference to find a Seed's kubeconfig. This is why the master-controller so far did not interfere much with my dummy Seeds I use for testing for operator (these dummy seeds have no Clusters, so maybe that is also helping me a bit).
* The seed-controller however has a special seedKubeconfigGetter that can only return its own Seed. This makes the webhook effectively only able to validate its own seed. For every other seed, even in the same namespace, it will complain about not being able to find the Seed's kubeconfig.

This problem goes away once the Kubermatic Operator replaces the Helm chart, because the Operator will always create Webhooks that are restricted to a given namespace. This is done by labelling the Kubermatic namespace and then using a label selector on the webhook.

To aid me in testing and make the migration easier, we have IMHO two options:

1. Label the Kubermatic namespace. Some controller could do that.
2. Ignore requests for other namespaces inside the seed webhook.

For simplicity sake and because it's only relevant during the migration, I chose option 2.

Relates to #4723 

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/main.go
+++ b/api/cmd/master-controller-manager/main.go
@@ -198,6 +198,7 @@ func main() {
 		seedValidationWebhookServer, err := runOpts.seedvalidationHook.Server(
 			ctx,
 			log,
+			ctrlCtx.namespace,
 			runOpts.workerName,
 			ctrlCtx.seedsGetter,
 			provider.SeedClientGetterFactory(ctrlCtx.seedKubeconfigGetter),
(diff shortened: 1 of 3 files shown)


kron4eg (approved):
/lgtm
/approve

PR #4790, opened 2019-12-02
Test idempotency of the Kubernetes resource deployment

alvaroaleman:

Adds an idempotency test for the resource deployment of the Kubernetes controller. I'll add somthing somiliar for the UCCM and openshift controllers as follow-ups.

Requires https://github.com/kubermatic/infra/pull/794
Fixes #

**Documentation**:

```release-note
none
```


api/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
```diff
@@ -0,0 +1,109 @@
...
+				Scope: apiextensionsv1beta1.ClusterScoped,
+			},
+		}},
+	}
+	if _, err := envtest.InstallCRDs(cfg, crdInstallOpts); err != nil {
+		t.Fatalf("failed install crds: %v", err)
+	}
+
+	go func() {
+		if err := mgr.Start(make(chan struct{})); err != nil {
```
kron4eg, line 63: A better way would be to have a cancel function in defer that will close this channel.

api/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
```diff
@@ -0,0 +1,109 @@
...
+		// Results in timeouts, maybe a controller-runtime bug?
+		// Paths: []string{"../../../../../config/kubermatic/crd"},
+		CRDs: []*apiextensionsv1beta1.CustomResourceDefinition{{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "clusters.kubermatic.k8s.io",
+			},
+			Spec: apiextensionsv1beta1.CustomResourceDefinitionSpec{
+				Group:   "kubermatic.k8s.io",
+				Version: "v1",
+				Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
```
kron4eg, line 48: I'm a bit worried about hardcoding GVR, would it make sense to have some kind of factory or grab it from the schema?

alvaroaleman: The test will fail if this is not correct

api/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
alvaroaleman: Yeah but since closing that channel will only close other channels, it doesn't really make a difference

kron4eg (commented):
/lgtm

PR #4795, opened 2019-12-02
Cleanup Resources when deleting KubermaticConfigurations/Seeds

### LINKED ISSUE #4755: Kubermatic Operator: Implement Cleanup
**User Story**
When deleting a KubermaticConfiguration, we should remove all resources created by the operator. For most things this is handled by our existing OwnerReferences, but for cluster-wide resources we need to handle the cleanup manually. Finalizers would be a good way.

This is part of #4723.

**Acceptance criteria**
* [ ] After deleting a KubermaticConfiguration, all resources _the operator created_ are gone (this does for example not include the Secret that is created by the cert-manager for our Certificate -- users who want to get rid of that need to remove the entire namespace).

xrstf:
This PR uses finalizers to ensure that cluster-wide resources are removed and Seeds are cleaned up when a KubermaticConfiguration or Seed is removed.

* `KubermaticConfigurations` inside master clusters have a new finalizer from the operator so that it can delete the webhook and ClusterRoleBinding.
* `Seeds` inside the seed cluster (!) have an identical finalizer so that the seed-operator-controller can cleanup the seed cluster.
* `Seeds` inside the master cluster now also get a finalizer from the `seed-sync` controller. This ensures that this controller can propagate the deletion into the seed cluster.

The workflow for deleting a Seed is simple:

1. Delete the Seed CR inside the master cluster.

(Deleting the Seed CR inside the seed cluster will make the operator clean up the seed cluster, but since the seed-sync controller will immediately re-create the Seed CR, it's pointless unless you want to force-reconcile a seed cluster.)

It works like this for separate master/seed clusters:

1. Master-Seed CR has a seed-sync-cleanup finalizer.
2. This triggers the seed-sync controller to delete the Seed-Seed CR.
3. This triggers the operator to cleanup the seed cluster. Afterwards it removes its finalizer.
4. If this happened quicker than 30s, the Delete operation from step 2 succeeds and the seed-sync controller removes the finalizer on the Master-Seed CR. If this is slower, the controller errors out and retries later to delete again.
5. The Master-Seed is removed by the apiserver.

For shared master/seed clusters this is a bit more complex:

1. Seed CR has a seed-sync-cleanup *and* a operator-seed-cleanup finalizers.
2. Upon deletion, both controllers race to cleanup their finalizers:
   1. The seed-sync controller will fetch the Seed CR and notice that Master=Seed by comparing the UIDs. This makes it not perform an indefinitely blocking Delete op, but instead it will just remove its finalizer.
   2. The operator will simply cleanup as usual and then remove its finalizer.
3. It's not important in which order the two steps above happen. Afterwards, both finalizers are gone.
5. The Master-Seed is removed by the apiserver.

There are some known limitations:

1. Deleting the kubeconfig-secret for a Seed will make the cleanup impossible. In this case the seed-sync controllers errors indefinitely. It's assumed that cluster admins will either fix the missing kubeconfig to allow cleanup or remove the finalizer because they nuked the seed cluster already. This could be tackled by making the kubeconfig Secret *own* the Seed CR, but that seems weird and is out of scope for this PR.
2. Deleting a KubermaticConfiguration or a Kubermatic namespace will obviously kill controllers and the operator and make Seed deletion stuck. But I think the procedure to first delete all Seeds and then the KubermaticConfiguration is fine.

Fixes #4755 

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/operator/master/reconciler.go
+++ b/api/pkg/controller/operator/master/reconciler.go
@@ -17,7 +17,6 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -84,7 +83,7 @@ func (r *Reconciler) reconcile(config *operatorv1alpha1.KubermaticConfiguration,
 	oldConfig := config.DeepCopy()
 	kubernetes.AddFinalizer(config, common.CleanupFinalizer)
 	if err := r.Patch(r.ctx, config, ctrlruntimeclient.MergeFrom(oldConfig)); err != nil {
-		return err
+		return fmt.Errorf("failed to add finalizer: %v", err)
 	}
 
 	if err := r.reconcileNamespaces(config, logger); err != nil {
@@ -135,7 +134,7 @@ func (r *Reconciler) reconcile(config *operatorv1alpha1.KubermaticConfiguration,
 }
 
 func (r *Reconciler) cleanupDeletedConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
-	if sets.NewString(config.Finalizers...).Has(common.CleanupFinalizer) {
+	if kubernetes.HasAnyFinalizer(config, common.CleanupFinalizer) {
 		logger.Debug("KubermaticConfiguration was deleted, cleaning up cluster-wide resources")
 
 		if err := common.CleanupClusterResource(r, &rbacv1.ClusterRoleBinding{}, kubermatic.ClusterRoleBindingName(config)); err != nil {
@@ -149,7 +148,11 @@ func (r *Reconciler) cleanupDeletedConfiguration(config *operatorv1alpha1.Kuberm
 		oldConfig := config.DeepCopy()
 		kubernetes.RemoveFinalizer(config, common.CleanupFinalizer)
 
-		return r.Patch(r.ctx, config, ctrlruntimeclient.MergeFrom(oldConfig))
+		if err := r.Patch(r.ctx, config, ctrlruntimeclient.MergeFrom(oldConfig)); err != nil {
+			return fmt.Errorf("failed to remove finalizer: %v", err)
+		}
+
+		return nil
 	}
 
 	return nil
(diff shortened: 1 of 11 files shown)


api/pkg/controller/seed-sync/reconciler.go
```diff
@@ -58,6 +61,20 @@ func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, logger *zap.SugaredLogge
...
+		return r.cleanupDeletedSeed(seed, client, logger)
+	}
+
+	// ensure we always have a cleanup finalizer on the original
+	// Seed CR inside the master cluster
+	if err := common.PatchSeed(r.Client, seed, func(s *kubermaticv1.Seed) error {
+		kubernetes.AddFinalizer(s, common.CleanupFinalizer)
+		return nil
+	}); err != nil {
+		return err
```
xrstf: I chose to sneakily re-use the common package from the operator. Not sure if we want to move the PatchSeed and PatchKubermaticConfiguration into the `controller/util` package.

api/pkg/controller/operator/master/reconciler.go
```diff
@@ -71,6 +75,19 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 func (r *Reconciler) reconcile(config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
 	logger.Debug("Reconciling Kubermatic configuration")
 
+	// config was deleted, let's clean up
+	if config.DeletionTimestamp != nil {
+		return r.cleanupDeletedConfiguration(config, logger)
+	}
+
+	// ensure we always have a cleanup finalizer
+	if err := common.PatchKubermaticConfiguration(r, config, func(cfg *operatorv1alpha1.KubermaticConfiguration) error {
```
kron4eg: functions like PatchKubermaticConfiguration and PatchSeed shouldn't exists for the sole reason to provide concrete type in the callback, especially given that callback you've provided don't take any advantage of the concrete type.

Please converge them to the runtime.Object or metav1.Object interfaces.

kron4eg (changes requested):
Aside from small nit, looks good

api/pkg/controller/operator/master/reconciler.go
```diff
@@ -118,6 +135,27 @@ func (r *Reconciler) reconcile(config *operatorv1alpha1.KubermaticConfiguration,
 	return nil
 }
 
+func (r *Reconciler) cleanupDeletedConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
+	if sets.NewString(config.Finalizers...).Has(common.CleanupFinalizer) {
```
kron4eg: btw, we have api/pkg/kubernetes.HasAnyFinalizer() which can be used interchangeably here

api/pkg/controller/operator/master/reconciler.go
xrstf: Neither runtime Objects nor metav1 Objects were usable because one didn't have a DeepCopy and the other wasn't compatible with the Finalizer helper functions. So I just removed the helpers alltogether.

kron4eg (approved):
/lgtm
/approve

xrstf:
/hold

This makes clusters get stuck on CI.

alvaroaleman (approved):
/lgtm
/approve

PR #4796, opened 2019-12-03
De-attach cluster ids from the UserSSHKey object when the cluster is deleted 

### LINKED ISSUE #4358: Cleaning up cluster ids in the UserSSHKey object once a key is detached from a cluster
**Bug**
When a UserSShKey is assigned to a given cluster, the cluster id will be attached to the UserSSHKey object. However, when the cluster is deleted or cleaned up, the cluster id remains the UserSSHKey specs which makes it hard to sync those cluster with the designated key.

edit from @kdomanski 2010-10-04-14:51
After talking to everyone involved, it appears that this cleanup belongs into a controller. It might end up being racy, since cleaning up the keys and deleting the cluster are not a single atomic operation.

**Acceptance criteria**
- [ ] Once the cluster is deleted, verify the cluster id has been cleared from the UserSSHKey object that was assigned to the cluster.

moadqassem:
When a UserSShKey is assigned to a given cluster, the cluster id will be attached to the UserSSHKey object. However, when the cluster is deleted or cleaned up, the cluster id remains the UserSSHKey specs.
Fixes #4358

No

```release-note
None
```


moadqassem:
**I am definitely open  for a better name for this controller `api/pkg/controller/usersshkeysclustersynchronizer`  if any has a better name!**

api/cmd/master-controller-manager/controllers.go
```diff
@@ -249,3 +203,58 @@ func userSSHKeysSynchronizerFactoryCreator(ctrlCtx *controllerContext) seedcontr
...
+			ctrlCtx.workerCount,
+		)
+	}
+	return func(mgr manager.Manager) (string, error) {
+		return usersshkeysclustersynchronizer.ControllerName, factory(mgr)
+	}
+}
+
+func createSeedManagers(ctrlCtx *controllerContext, mgr manager.Manager) (map[string]manager.Manager, error) {
+	log := ctrlCtx.log.Named("usersshkeys-cluster-synchronizer-factory")
```
alvaroaleman: Please change the name here, as this isn't necessarily part of the `usersshkeys-cluster-synchronizer-factory` anymore

api/cmd/master-controller-manager/controllers.go
```diff
@@ -24,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
```
alvaroaleman: Please leave the imports grouped as they were

api/pkg/controller/usersshkeysclustersynchronizer/controller.go
```diff
@@ -0,0 +1,165 @@
+package usersshkeysclustersynchronizer
```
alvaroaleman: I think the current name is somewhat ambigious. Also I think it would be better to make it part of the existing `usersshkeyssynchronizer` controller:

* Make it add a finalizer to the cluster
* When the clusters deletion timestamp is non nil: Iterate over all ssh keys and remove the cluster from them

You can read about finalizers here: https://book.kubebuilder.io/reference/using-finalizers.html

This way, this will work even when the controller wasn't up at the time a given cluster got deleted

moadqassem: addressed: https://github.com/kubermatic/kubermatic/pull/4796/commits/63f91829a9d8defd7bde4128cf6f2831a6ff3f77

api/pkg/controller/usersshkeyssynchronizer/controller.go
```diff
@@ -157,6 +160,35 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, request reconcile.Request
 		return fmt.Errorf("failed to reconcile ssh key secret: %v", err)
 	}
 
+	kubernetes.AddFinalizer(cluster, UserSSHKeysClusterIDsCleanupFinalizer)
+	if err := seedClient.Update(r.ctx, cluster); err != nil {
```
alvaroaleman: Can we use PATCH? There are by now a lot of controllers that modify the cluster, which is why update requests have a very high chance of failing

api/pkg/controller/usersshkeyssynchronizer/controller.go
```diff
@@ -157,6 +160,35 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, request reconcile.Request
...
+	}
+
+	if cluster.DeletionTimestamp != nil {
+		if err := r.reconcileUserSSHKeys(userSSHKeys.Items, cluster.Name); err != nil {
+			return fmt.Errorf("failed reconciling usersshkey: %v", err)
+		}
+
+		if kubernetes.HasFinalizer(cluster, UserSSHKeysClusterIDsCleanupFinalizer) {
+			kubernetes.RemoveFinalizer(cluster, UserSSHKeysClusterIDsCleanupFinalizer)
+			if err := seedClient.Update(r.ctx, cluster); err != nil {
```
alvaroaleman: Same here, please use PATCH

api/pkg/controller/usersshkeyssynchronizer/controller.go
```diff
@@ -157,6 +160,35 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, request reconcile.Request
 		return fmt.Errorf("failed to reconcile ssh key secret: %v", err)
 	}
 
+	kubernetes.AddFinalizer(cluster, UserSSHKeysClusterIDsCleanupFinalizer)
+	if err := seedClient.Update(r.ctx, cluster); err != nil {
+		return fmt.Errorf("failed adding %s finalizer: %v", UserSSHKeysClusterIDsCleanupFinalizer, err)
+	}
+
+	if cluster.DeletionTimestamp != nil {
+		if err := r.reconcileUserSSHKeys(userSSHKeys.Items, cluster.Name); err != nil {
```
alvaroaleman: Can we give this a better name, like `cleanupUserSSHKeys`?

alvaroaleman (approved):
/lgtm
/approve

PR #4803, opened 2019-12-03
Use RBAC-based authorization to access Kubernetes Dashboard

floreks:
Based on the user role in Kubermatic proper access will be granted in Kubernetes Dashboard.

Zrzut ekranu z 2019-12-03 12-31-02

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/kubernetes/cluster.go
+++ b/api/pkg/provider/kubernetes/cluster.go
@@ -328,6 +328,28 @@ func (p *ClusterProvider) GetClientForCustomerCluster(userInfo *provider.UserInf
 	return p.userClusterConnProvider.GetClient(c, p.withImpersonation(userInfo))
 }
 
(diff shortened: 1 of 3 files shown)


api/pkg/provider/kubernetes/cluster.go
```diff
@@ -328,6 +328,28 @@ func (p *ClusterProvider) GetClientForCustomerCluster(userInfo *provider.UserInf
 	return p.userClusterConnProvider.GetClient(c, p.withImpersonation(userInfo))
 }
 
+func (p *ClusterProvider) GetTokenForCustomerCluster(userInfo *provider.UserInfo, cluster *kubermaticv1.Cluster) (string, error) {
+	parts := strings.Split(userInfo.Group, "-")
```
floreks, line 332: Can we assume that the group format will be `name-yyy`?

zreigz: maybe simpler:
```		
if strings.HasPrefix(userInfo.Group, "viewers") {

} else {

}
```

PR #4806, opened 2019-12-04
Move Prow jobs to Kubermatic repo

xrstf:
This moves all Prow jobs over to this repository and bumps Helm to 2.16 because we need the k8s 1.16 support.

I have sorted the Prow jobs a bit and moved the repo name to the front, so that the job names do not collide with what we have in the infra repo right now.

```release-note
NONE
```


.prow.yaml
```diff
@@ -0,0 +1,1053 @@
...
+    clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
+    spec:
+      containers:
+      - image: quay.io/kubermatic/promtool:2.7.0-3
+        command:
+        - "./api/hack/verify-user-cluster-prometheus-configs.sh"
+      imagePullSecrets:
+      - name: quay
+
+  - name: pull-kubermatic-validate-prow-yaml
```
alvaroaleman: This one is not needed, we maintain it in the infra repo to keep the version in sync with our prow version

.prow.yaml
```diff
@@ -0,0 +1,1053 @@
...
+          privileged: true
+        resources:
+          requests:
+            memory: 2.5Gi
+            cpu: 500m
+          limits:
+            memory: 4Gi
+            cpu: 2
+
+  - name: pull-kubermatic-e2e-upgrade-2.11-2.12
```
alvaroaleman: Not needed on this branch

.prow.yaml
```diff
@@ -0,0 +1,1053 @@
...
+          privileged: true
+        resources:
+          requests:
+            memory: 2.5Gi
+            cpu: 500m
+          limits:
+            memory: 4Gi
+            cpu: 2
+
+  - name: pull-kubermatic-e2e-upgrade-weekly
```
alvaroaleman: We can not maintain it here as that means it will also run on master

PR #4810, opened 2019-12-04
Make controllers not override resource requests set by VPA

xmudrii:

Relevant to the upcoming `vpa-updater` controller (#4763).

This PR ensures that controllers will not override the resource requests set by `vpa-updater`. This is done by checking is there the annotation set by `vpa-updater`.

Relevant to #4763 #3213

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
+++ b/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
@@ -295,11 +295,7 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 					kubeControllerManagerContainerName: defaultKubeControllerResourceRequirements.DeepCopy(),
 					openvpnSidecar.Name:                openvpnSidecar.Resources.DeepCopy(),
 				}
-				overrides := map[string]*corev1.ResourceRequirements{}
-				if data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil {
-					overrides[kubeControllerManagerContainerName] = data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources.DeepCopy()
-				}
-				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, overrides, dep.Annotations)
+				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, resources.GetOverrides(data.Cluster().Spec.ComponentsOverride), dep.Annotations)
 				if err != nil {
 					return nil, fmt.Errorf("failed to set resource requirements: %v", err)
 				}
--- a/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
+++ b/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
@@ -251,11 +251,6 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 					return nil, err
 				}
 
-				resourceRequirements := defaultKubeControllerResourceRequirements
-				if data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil {
-					resourceRequirements = *data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources
-				}
-
 				env, err := controllermanager.GetEnvVars(data)
 				if err != nil {
 					return nil, fmt.Errorf("failed to get controller manager env vars: %v", err)
@@ -266,10 +261,11 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 					return nil, fmt.Errorf("failed to get openvpn sidecar: %v", err)
 				}
 				kubeControllerManagerContainerName := "kube-controller-manager"
+
 				dep.Spec.Template.Spec.Containers = []corev1.Container{
 					*openvpnSidecar,
 					{
-						Name:    "kube-controller-manager",
+						Name:    kubeControllerManagerContainerName,
 						Image:   image,
 						Env:     env,
 						Command: []string{"hyperkube", kubeControllerManagerContainerName},
@@ -279,7 +275,6 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 							kubeControllerManagerCACertPath,
 							"/etc/kubernetes/pki/front-proxy/ca/ca.crt",
 						),
-						Resources: resourceRequirements,
 						ReadinessProbe: &corev1.Probe{
 							Handler: corev1.Handler{
 								HTTPGet: &corev1.HTTPGetAction{
@@ -296,7 +291,18 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 						VolumeMounts: volumeMounts,
 					},
 				}
-
+				defResourceRequirements := map[string]*corev1.ResourceRequirements{
+					kubeControllerManagerContainerName: defaultKubeControllerResourceRequirements.DeepCopy(),
+					openvpnSidecar.Name:                openvpnSidecar.Resources.DeepCopy(),
+				}
+				overrides := map[string]*corev1.ResourceRequirements{}
+				if data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil {
+					overrides[kubeControllerManagerContainerName] = data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources.DeepCopy()
+				}
+				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, overrides, dep.Annotations)
+				if err != nil {
+					return nil, fmt.Errorf("failed to set resource requirements: %v", err)
+				}
 				podLabels, err := data.GetPodTemplateLabels(resources.ControllerManagerDeploymentName, dep.Spec.Template.Spec.Volumes, nil)
 				if err != nil {
 					return nil, err
--- a/api/pkg/controller/seed-controller-manager/openshift/resources/openshift_apiserver_deployment.go
+++ b/api/pkg/controller/seed-controller-manager/openshift/resources/openshift_apiserver_deployment.go
@@ -173,11 +173,6 @@ func OpenshiftAPIServerDeploymentCreator(ctx context.Context, data openshiftData
 				return nil, fmt.Errorf("failed to get dnat-controller sidecar: %v", err)
 			}
 
-			resourceRequirements := openshiftAPIServerDefaultResourceRequirements.DeepCopy()
-			if data.Cluster().Spec.ComponentsOverride.Apiserver.Resources != nil {
-				resourceRequirements = data.Cluster().Spec.ComponentsOverride.Apiserver.Resources
-			}
-
 			// TODO: Make it cope with our registry overwriting
 			image, err := hypershiftImage(data.Cluster().Spec.Version.String())
 			if err != nil {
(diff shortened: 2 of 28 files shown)


api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
```diff
@@ -266,10 +261,23 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
...
+					openvpnSidecar.Name:                openvpnSidecar.Resources,
+				}
+				if data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil {
+					resourceRequirements[kubeControllerManagerContainerName] = *data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources
+				} else if _, ok := dep.ObjectMeta.Annotations[kubermaticv1.UpdatedByVPALabelKey]; ok {
+					for _, container := range dep.Spec.Template.Spec.Containers {
+						resourceRequirements[container.Name] = container.Resources
+					}
+					openvpnSidecar.Resources = resourceRequirements[openvpnSidecar.Name]
+				}
```
xmudrii: This approach is not going to work well if we use VPA and ComponentOverrides at the same time:
* since `data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil` is true, we will override the resource requirements to the predefined value. This is going to make a conflict between controller and vpa-updater, which will mostly like crash the cluster very quickly
* in this case, we skip setting the correct resource requirements for `openvpnSidecar`, which is also going to make conflicts between controllers.

I'm not sure if this is a concern or not. I'd say that the user mustn't have both enabled at the same time, but I'm not sure is anyone requiring that.

api/pkg/controller/operator/master/resources/kubermatic/api.go
```diff
@@ -146,6 +160,14 @@ func APIDeploymentCreator(cfg *operatorv1alpha1.KubermaticConfiguration, workerN
 			}
 
 			d.Spec.Template.Spec.Volumes = volumes
+
+			resourceRequirements := apiResourceRequirements
```
alvaroaleman: Please always `DeepCopy()` the default to avoid bugs.

Also, I would probably factor the logic out, as its currently very repetetive, something along the lines of:

```
// SetResourcesForDeployment sets the resources for the Deployment unless it has the `UpdatedByVPA`
// label.
// The key of the `default` map is the name of the container.
SetResourcesForDeployment(d *appsv1.Deployment, default map[string]resources.ResourceRequirements)
```

api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
alvaroaleman: IMHO we should still allow `ComponentsOverride` and respect it as the first priority. Maybe we can add something to the VPA controller to not change deployments and make sure the label is absent for which the cluster has `ComponentsOverride`?

api/pkg/controller/operator/master/resources/kubermatic/api.go
xmudrii: Makes sense, I'll refactor that part. Do you have preferences regarding where to put that function or wherever I think it makes sense?

xmudrii: Fixed and simplified the logic.

api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
xmudrii: The priority part is fixed. We will see what we can add to the VPA controller to not change deployments when I get back to the other PR.

xmudrii:
@alvaroaleman Fixed, PTAL. I'll be testing this now.

api/pkg/resources/resources.go
```diff
@@ -927,3 +935,30 @@ func GetHTTPProxyEnvVarsFromSeed(seed *kubermaticv1.Seed, inClusterAPIServerURL
 
 	return envVars
 }
+
+func SetResourceRequirements(containers []corev1.Container, requirements, overrides map[string]*corev1.ResourceRequirements, annotations map[string]string) error {
```
alvaroaleman, line 919: This definitely needs a godoc explaining what it is supposed to do :)

api/pkg/resources/resources.go
```diff
@@ -927,3 +935,30 @@ func GetHTTPProxyEnvVarsFromSeed(seed *kubermaticv1.Seed, inClusterAPIServerURL
...
+	val, ok := annotations[kubermaticv1.UpdatedByVPALabelKey]
+	if ok {
+		var req []Requirements
+		err := json.Unmarshal([]byte(val), &req)
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal resource requirements provided by vpa: %v", err)
```
alvaroaleman: `...from annotation $annotationkey`

api/pkg/resources/resources.go
```diff
@@ -927,3 +935,30 @@ func GetHTTPProxyEnvVarsFromSeed(seed *kubermaticv1.Seed, inClusterAPIServerURL
...
+		}
+		for _, r := range req {
+			requirements[r.Name] = r.Requires
+		}
+	}
+	if overrides != nil {
```
alvaroaleman: Nit: The nilcheck is not needed here, `range` can cope with nil maps

api/pkg/resources/resources_test.go
```diff
@@ -79,3 +82,66 @@ func TestUserClusterDNSResolverIP(t *testing.T) {
...
+
+	tests := []struct {
+		name                 string
+		containers           []corev1.Container
+		annotations          map[string]string
+		defaultRequirements  map[string]*corev1.ResourceRequirements
+		expectedRequirements map[string]*corev1.ResourceRequirements
+	}{
+		{
+			name: "test with valid resource requirements json",
```
alvaroaleman: I think we should also have some testing for the correct precedence, i.E. overwrite > annotation > default

api/pkg/resources/usercluster/deployment.go
```diff
@@ -168,6 +167,12 @@ func DeploymentCreator(data userclusterControllerData, openshift bool) reconcili
 					},
 				},
 			}
+			err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, map[string]*corev1.ResourceRequirements{
```
alvaroaleman: In order to make all of these defaults less verbose, we could probably move the inline map creation into a `func defaultResourceRequirements() map[string]*corev1.ResourceRequirements { ... }`. WDYT?

api/pkg/resources/prometheus/statefulset.go
```diff
@@ -155,7 +150,17 @@ func StatefulSetCreator(data *resources.TemplateData) reconciling.NamedStatefulS
 					},
 				},
 			}
-
+			defResourceRequirements := map[string]*corev1.ResourceRequirements{
+				name: defaultResourceRequirements.DeepCopy(),
+			}
+			overrides := map[string]*corev1.ResourceRequirements{}
+			if data.Cluster().Spec.ComponentsOverride.Prometheus.Resources != nil {
```
alvaroaleman: Would it maybe be better to extend the `ComponentsOverride` type with a `func (co *ComponentsOverride) GetOverrides() map[string]*corev1.ResourceRequirements` and then just pass the result of that func into all `SetResourceRequirements` calls everywhere?

That has the advantage that if someone would extend the overrides in the future, it will just work and does not require extending code below the `resources` package.

xmudrii: I'm trying to understand how this function would look like. Would it return a map that contain requirements for all resources supporting overrides or something else?

api/pkg/resources/usercluster/deployment.go
xmudrii: I've fixed this at the places where it made sense. At other places, we had a few sidecars, so it was whatever.

api/pkg/resources/resources_test.go
xmudrii: I've added a few new test cases that should cover such situations.

api/pkg/resources/prometheus/statefulset.go
alvaroaleman: Yeah, a map containerName -> ResourceOverride

api/pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -268,6 +269,31 @@ type ComponentSettings struct {
...
+	}
+	if cs.Scheduler.Resources != nil {
+		r["scheduler"] = cs.Scheduler.Resources.DeepCopy()
+	}
+	if cs.Etcd.Resources != nil {
+		r["etcd"] = cs.Etcd.Resources.DeepCopy()
+	}
+	if cs.Prometheus.Resources != nil {
+		r["prometheus"] = cs.Prometheus.Resources.DeepCopy()
+	}
```
xmudrii: @alvaroaleman I don't like the idea that we have static names here. If for any reason we change the container name, this can break everything.

The problem is that those names are defined in two places - the `resources` package and the package where the resource is defined (e.g. `resources/apiserver`). If I import those packages, I get import cycles, which is as expected.

I thought about moving this function to another package, but that is also not possible because in that case, it couldn't be a method of `ComponentSettings`.

alvaroaleman: I agree on the static names. Maybe move it to another package and not make it a method of `ComponentSettings` but instead just take `ComponentSettings` as arg?

xmudrii: It's now moved to the `resources` package.

api/pkg/controller/seed-controller-manager/openshift/resources/openshift_apiserver_deployment.go
```diff
@@ -259,7 +253,19 @@ func OpenshiftAPIServerDeploymentCreator(ctx context.Context, data openshiftData
...
-
+			defResourceRequirements := map[string]*corev1.ResourceRequirements{
+				OpenshiftAPIServerDeploymentName: openshiftAPIServerDefaultResourceRequirements.DeepCopy(),
+				openvpnSidecar.Name:              openvpnSidecar.Resources.DeepCopy(),
+				dnatControllerSidecar.Name:       dnatControllerSidecar.Resources.DeepCopy(),
+			}
+			overrides := map[string]*corev1.ResourceRequirements{}
+			if data.Cluster().Spec.ComponentsOverride.Apiserver.Resources != nil {
+				overrides[OpenshiftAPIServerDeploymentName] = data.Cluster().Spec.ComponentsOverride.Apiserver.Resources.DeepCopy()
+			}
```
xmudrii, line 264: At some places, we still use the old overrides logic, because OpenShift components have a different name (e.g. the OpenShift API server is called `openshift-apiserver`).

api/pkg/resources/etcd/statefulset.go
```diff
@@ -87,14 +89,10 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
...
 			}
-			resourceRequirements := defaultResourceRequirements
-			if data.Cluster().Spec.ComponentsOverride.Etcd.Resources != nil {
-				resourceRequirements = *data.Cluster().Spec.ComponentsOverride.Etcd.Resources
-			}
 
 			set.Spec.Template.Spec.Containers = []corev1.Container{
 				{
-					Name:    name,
+					Name:    resources.EtcdStatefulSetName,
```
xmudrii, line 95: I renamed some containers so we have the consistent names between GetOverrides function and containers. The old and new names are supposed to match.

api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
```diff
@@ -296,7 +291,14 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 						VolumeMounts: volumeMounts,
 					},
 				}
-
+				defResourceRequirements := map[string]*corev1.ResourceRequirements{
+					kubeControllerManagerContainerName: defaultKubeControllerResourceRequirements.DeepCopy(),
+					openvpnSidecar.Name:                openvpnSidecar.Resources.DeepCopy(),
+				}
+				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, resources.GetOverrides(data.Cluster().Spec.ComponentsOverride), dep.Annotations)
```
xmudrii, line 298: The overrides are not going to work here since container names are different. To be fixed.

PR #4818, opened 2019-12-05
Handle unready Caches

xrstf:
Currently when the master-controller-manager starts up and acquired its leader lease, the usersshkeysyncer and projectlabelsyncer controllers will attempt to reconcile as soon as possible. This causes problems because the caches have not yet been started, so the pod is logging a bunch of serious sounding errors with stacktraces. This is especially misleading when debugging something, because you fetch the pod's logs and are greeted with many errors.

This PR extends the two controllers a bit to

1. Handle the worker-name labels as early as possible. This is not strictly required to fix the errors, but makes the controller play nicer when other Kubermatic installations are in the same cluster (as we now have due to the Operator and me testing it on dev).
2. Return the `ErrCacheNotStarted` so that we can handle it gracefully and simply retry.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/util/util.go
+++ b/api/pkg/controller/util/util.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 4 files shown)


api/pkg/controller/util/util.go
```diff
@@ -144,3 +151,14 @@ func ConcurrencyLimitReached(ctx context.Context, client ctrlruntimeclient.Clien
...
 	return clustersUpdatingInProgressCount >= limit, nil
 }
+
+// IsCacheNotStarted returns true if the given error is not nil and an instance of
+// cache.ErrCacheNotStarted.
+func IsCacheNotStarted(err error) bool {
+	if err == nil {
+		return false
+	}
+
```
xrstf, line 161: I'm not 100000% sure that we need this nil check. But it feels safe to have it.

kron4eg: true, it's absolutely excessive: https://play.golang.org/p/vf1uN8EjVHB

kron4eg (approved):
/lgtm
/approve

/hold

@xrstf there is microscopic nil, if you don't feel like removing it, then please simply unhold the PR

PR #4823, opened 2019-12-05
Share seed managers across controllers in seed-controller-lifecycle

xrstf:
This prepares a set of managers to be used by all controllers, with the purpose of improving performance and reducing memory usage. It also makes constructing controllers for the seed-lifecycle much easier. We also now give all controllers a shared context and cancel this context when we teardown the controllers again.

As a bonus this now also both fixes the seedKubeconfigSuccess metric and makes it work across all controllers.

```release-note
NONE
```

### DIFF
--- a/api/cmd/master-controller-manager/controllers.go
+++ b/api/cmd/master-controller-manager/controllers.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"go.uber.org/zap"
-
 	projectlabelsynchronizer "github.com/kubermatic/kubermatic/api/pkg/controller/project-label-synchronizer"
 	"github.com/kubermatic/kubermatic/api/pkg/controller/rbac"
 	seedcontrollerlifecycle "github.com/kubermatic/kubermatic/api/pkg/controller/seed-controller-lifecycle"
@@ -21,7 +19,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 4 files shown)


api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -181,5 +233,46 @@ func (r *Reconciler) reconcile() error {
...
+	cache.Cache
+}
+
+func (m *unstartableCache) Start(_ <-chan struct{}) error {
+	return nil
+}
+
+func (m *unstartableCache) WaitForCacheSync(_ <-chan struct{}) bool {
+	return true
+}
```
xrstf, line 279: This is copied verbatim from the master-controller-manager, which does a very similar logic to wait for caches to sync, but in its case so that non-manager-managed components can work (most notably the seed admission webhook, which is needed for the migrations before the regular manager is started).

Not sure if/how we can prevent code duplication for this.

api/pkg/controller/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -64,12 +80,26 @@ func Add(
...
+
+	go func() {
+		if err := cache.Start(ctx.Done()); err != nil {
+			log.Fatalw("failed to start cache", zap.Error(err))
+		}
+	}()
+
+	if !cache.WaitForCacheSync(ctx.Done()) {
+		log.Fatal("failed to wait for caches to synchronize")
+	}
```
xrstf, line 96: Without this, the master-controller-manager still works (presumably because it has its own super mega cache that is started and maintained outside of the manager), but the kubermatic-operator panics every time the seeds change. This even happens to a naked controller-manager binary that has no actual controllers whatsoever (besides the seed-lifecycle controller).

The panic is "close of closed channel" and so far I could track it down to the `close` statement in `api/vendor/k8s.io/client-go/tools/cache/shared_informer.go`'s `processorListener::pop()` function.

api/cmd/master-controller-manager/controllers.go
```diff
@@ -71,71 +69,32 @@ func rbacControllerFactoryCreator(
 	workerCount int,
 	selectorOps func(*metav1.ListOptions),
 ) seedcontrollerlifecycle.ControllerFactory {
-
+	// register metrics
```
alvaroaleman: super useless comment

alvaroaleman:
/hold
As this has mergeconflicts, retesting will not help

xrstf:
Merge conflicts fixed.

/hold cancel

PR #4825, opened 2019-12-06
Seed components: Check if cluster is up by using HealthStatusUp

alvaroaleman:

We changed the health status to be `Provisioning` for as long as the cluster hasn't been ready once and is not fully ready. This breaks our current assumption of "If its not down its up" which leads to a lot of error on initial cluster creation. This PR fixes that by checking for "up" instead.

Fixes #

**Documentation**:

```release-note
none
```


xrstf:
very nice

/lgtm

PR #4839, opened 2019-12-09
Add parent cluster readable name to default worker names

### LINKED ISSUE #1838: Make node names contain human-readable cluster name
### Todo
Change node names to contain the human-readable cluster name by default e. g. cluster-name-xxx-xxx  

Was generally thumbs upped when proposed in slack, comes from here  kubermatic/customer-projects#33

### Question
Does this adjust when you rename the cluster? Or do we carry around names from the whole lifecycle of the cluster on old nodes?

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard-v2/issues/1838

**Documentation**:

```release-note
Added parent cluster readable name to default worker names.
```

### DIFF
--- a/api/pkg/resources/machine/machinedeployment.go
+++ b/api/pkg/resources/machine/machinedeployment.go
@@ -33,7 +33,7 @@ func Deployment(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc *kubermati
 	} else {
 		// GenerateName can be set only if Name is empty to avoid confusing error:
 		// https://github.com/kubernetes/kubernetes/issues/32220
-		md.GenerateName = "worker-"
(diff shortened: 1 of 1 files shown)


kdomanski:
Uhm, it looks like this function along with its unit test and perhaps some of the fixtures have been dead code ever since we switched e2e tests to MachineDeployments in 54ffc5df26dfee946534314ff0dc78fb2a9b048a

I imagine the change belongs to `func Deployment` instead. And we could get rid of the dead code while we're at it.

maciaszczykm:
@kdomanski True, my mistake. Tried to do it too quickly. However, the code is used in the tests. I will change the correct method.

maciaszczykm:
@kdomanski PTAL

PR #4849, opened 2019-12-10
Remove ci-run-minimal-conformance-tests.sh symlink

xrstf:
This has confused me constantly and there is no real reason to have this backwards compatibility hack anymore.

```release-note
NONE
```


xrstf:
/hold

The presubmits for older branches are still defined in the infra repo. We cannot change them without also updating the older branches. So in effect the old branches keep us from fixing this legacy hack.

PR #4853, opened 2019-12-10
Fix missing deployment selector

xrstf:
Since appsv1 it's not allowed to skip defining the selectors.

```release-note
NONE
```


kron4eg (approved):
/lgtm

Issue #4777, opened 2019-11-28
Support MachineDeployment specific kubelet config


chrkl:
**User Story**

The kubelet configuration is currently applied as a ConfigMap from the addons image. Customer wants to change the API request limits per cluster or even MachineDeployment. E.g., different limits for small and large nodes.

**Acceptance criteria**

Kubelet configuration should be adaptable per MachineDeployment

toschneck:
As first step we make QPS parameter configurable, see https://github.com/kubermatic/kubermatic/pull/4854

kdomanski:
That PR is for configuring it globally, has nothing to do with MD-specific.
/unassign

PR #4854, opened 2019-12-10
addons/kubelet-configmap: allow configuring QPS

kdomanski:
Addresses immediate requirements of kubermatic/customer-projects#256, thus reducing the pressure for #4777

```release-note
The QPS settings of Kubeletes can now be configured per-cluster using addon Variables
```


alvaroaleman:
I _think_ we don't use this configmap unless kubeadm was used to join the nodes

kdomanski:
All providers now optionally support dynamic kubelet config.

PR #4855, opened 2019-12-11
Improve log flag handling

xrstf:
All controllers had their own way of handling the logging flags. Some used a string-based format, some used the dedicated `log.Format` type. Some validated the flags, some did not. Some put the log options in their logOptions struct, some kept them independent.

This PR cleans the mess up and makes sure that we always validate the logging flags. I had panics in my controller because I accidentally ran it with `-log-format=json` (lowercase was not allowed).

```release-note
NONE
```

### DIFF
--- a/api/pkg/test/clusterexposer/cmd/main.go
+++ b/api/pkg/test/clusterexposer/cmd/main.go
@@ -86,7 +86,7 @@ func isRequired(flagName string) bool {
 }
 
 func run(cmd *cobra.Command, _ []string) error {
-	rawLog := log.New(debug, log.Format(string(log.FormatJSON)))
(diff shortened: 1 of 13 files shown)


api/pkg/test/clusterexposer/cmd/main.go
```diff
@@ -86,7 +86,7 @@ func isRequired(flagName string) bool {
 }
 
 func run(cmd *cobra.Command, _ []string) error {
-	rawLog := log.New(debug, log.Format(string(log.FormatJSON)))
+	rawLog := log.New(debug, log.FormatJSON)
```
xrstf, line 89: This was ma favorite line to fix in this PR :-D

PR #4856, opened 2019-12-11
Use etcd v3.4 for Kube 1.17+

alvaroaleman:

* Upgrades etcd to 3.4 for Kube 1.17+ clusters
* Changes the backup cronjob to also work with the new etcd version

I have manually checked that:
* Upgrading etcd works
* The backup job works for both etcd 3.4 and 3.3
* The defragger job works for both etcd 3.4 and 3.3

Fixes #

**Documentation**:

```release-note
Etcd was upgraded to 3.4 for 1.17+ clusters
```

/CC @shahbazn

### DIFF
--- a/api/pkg/controller/seed-controller-manager/backup/backup_controller.go
+++ b/api/pkg/controller/seed-controller-manager/backup/backup_controller.go
@@ -44,7 +44,7 @@ const (
 	// will write the backup to
 	SharedVolumeName = "etcd-backup"
 	// DefaultBackupContainerImage holds the default Image used for creating the etcd backups
-	DefaultBackupContainerImage = "gcr.io/etcd-development/etcd:"
+	DefaultBackupContainerImage = "gcr.io/etcd-development/etcd"
 	// DefaultBackupInterval defines the default interval used to create backups
 	DefaultBackupInterval = "20m"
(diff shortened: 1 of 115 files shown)


api/pkg/controller/seed-controller-manager/backup/backup_controller.go
```diff
@@ -44,7 +44,7 @@ const (
 	// will write the backup to
 	SharedVolumeName = "etcd-backup"
 	// DefaultBackupContainerImage holds the default Image used for creating the etcd backups
-	DefaultBackupContainerImage = "gcr.io/etcd-development/etcd:" + etcd.ImageTag
+	DefaultBackupContainerImage = "gcr.io/etcd-development/etcd:"
```
xrstf: This should break the image-loader. I also don't think having the ":" at the end here is a good idea.

xrstf: What happens if someone runs the seed-controller with a custom backup container image that contains a version already? Is this a BC break and should be documented?

alvaroaleman: Changed it to check for a `:` before adding a tag. Also removed this from the imageloader, as its the same image used by the etcd statefulset (par definition) and that is already being pulled in.

PR #4857, opened 2019-12-11
cert-manager 0.12.0

xrstf:
There were no changes between beta1 and final for the deployment manifests and the upstream chart. So I guess this is a safe little happy bump.

```release-note
Update cert-manager to 0.12.0
```


kron4eg (approved):
/lgtm

PR #4885, opened 2019-12-16
Fix GCP orphaned routes issue

### LINKED ISSUE #3926: Orphaned GCP routes
**User Story**
Based on the amount of unused routes we had on GCP it looks like Kubermatic clusters will orphan routes on GCP when they get deleted.

**Acceptance criteria**
- [ ] Deleting a Kubermatic cluster (GCP) will not orphan routes.

moelsayed:
`NetworkName` is used by the gce cloud provider to populate the provider's `NetworkURL`.
This value can be provided in the config as a name or a url. Internally,
the gce cloud provider checks it and if it's a name, it will infer the URL from it.
However, if the name has a '/', the provider assumes it's a URL and uses it as is.
This breaks routes cleanup since the routes are matched against the URL,
which would be incorrect in this case.
On the provider side, the `"global/networks/default"` format is the valid
one since it's used internally for firewall rules and and network interfaces,
so it has to be kept this way.
tl;dr: use "default" or a full network URL, not "global/networks/default"

Fixes #3926

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/cloud/gcp/provider.go
+++ b/api/pkg/provider/cloud/gcp/provider.go
@@ -66,11 +66,13 @@ func (g *gcp) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update prov
 	}
 
 	// add the routes cleanup finalizer
-	cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
-		kuberneteshelper.AddFinalizer(cluster, routesCleanupFinalizer)
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to add %s finalizer: %v", routesCleanupFinalizer, err)
+	if !kuberneteshelper.HasFinalizer(cluster, routesCleanupFinalizer) {
+		cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
+			kuberneteshelper.AddFinalizer(cluster, routesCleanupFinalizer)
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to add %s finalizer: %v", routesCleanupFinalizer, err)
+		}
 	}
 	return cluster, nil
 }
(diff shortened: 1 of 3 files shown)


api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	defaultNetwork               = "global/networks/default"
+	DefaultNetwork               = "global/networks/default"
```
alvaroaleman, line 28: We also need to add a finalizer during initialization and have it block deletion until all routes for the given cluster are gone. Is there a sane way to figure out what cluster routes belong to?

moelsayed: I think it's possible to filter on node names. I will look into it.

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -262,3 +295,88 @@ func isHTTPError(err error, status int) bool {
 	gerr, ok := err.(*googleapi.Error)
 	return ok && gerr.Code == status
 }
+
+// cleanUnusedRoutes finds and remove unused gcp routes
+func (g *gcp) cleanUnusedRoutes(cluster *kubermaticv1.Cluster) error {
+	serviceAccount, err := GetCredentialsForCluster(cluster.Spec.Cloud, g.secretKeySelector)
+	if err != nil {
+		return err
```
alvaroaleman: Please ~warp~ wrap the errors, because otherwise the message may be hard to interpret (i.E. both here and below, you could get an "Access Denied" message)

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -262,3 +295,88 @@ func isHTTPError(err error, status int) bool {
...
+func (g *gcp) networkURL(project, network string) string {
+	url, err := url.Parse(network)
+	if err == nil && url.Host != "" {
+		return network
+	}
+	return computeAPIEndpoint + strings.Join([]string{"projects", project, "global", "networks", path.Base(network)}, "/")
+}
+
+// isClusterRoute checks if the route CIDR is part of the Cluster CIDR
+func isClusterRoute(cluster *kubermaticv1.Cluster, route *compute.Route) (bool, error) {
```
alvaroaleman, line 352: Could you write two  simple unittest demonstrating the functionality of this (Ip is in CIDR, IP is not in CIDR)?

api/pkg/provider/cloud/gcp/provider.go
```diff
@@ -52,6 +65,13 @@ func (g *gcp) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update prov
 		return nil, err
 	}
 
+	// add the routes cleanup finalizer
+	cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
```
alvaroaleman: I think we should check if the finalizer already exists first, the `AddFinalizer` doesn't add it again but we will attempt a no-up update request

alvaroaleman (approved):
/lgtm
/approve

PR #4886, opened 2019-12-16
Improve Dex handling in e2e tests

xrstf:
So this turned out to be an adventure. In order to use the Kubermatic Operator in our e2e tests, I wanted to understand how they actually work. I came across the `oidc-proxy-client` and dug deeper.

It turns out the name is slightly confusing. It's an "OIDC proxy"-client, not an OIDC "proxy-client". Well at least that confused me at first. The reason why it existed is that in the 3-legged OAuth2 workflow, the provider is redirecting the user's browser to the application via the predefined `redirectURI`. And so somebody needs to listen somewhere for that very request, which contains the temporary code which can then be exchanged for a real access token.

The insanity here was that instead of at least adjusting the oidc-proxy-client to match our needs, we instead added regex-based HTML parsing to extract the token at the end. The original purpose of the oidc-proxy-client was a demonstration after all, prettyprinting the token and claim at the end. After replacing that futility by making the oidc-proxy-client output JSON, I noticed further things:

* The only reason we had a Github connector defined in the `oauth_values.yaml` was because this makes Dex output different HTML on the `/dex/auth` endpoint than if only one login method (user+pass) was enabled (in which case Dex would immediately present the login form). The reason we want Github was that the cloud-based e2e tests (that run as postsubmits to check that cloud is still functioning) has Github (and others) enabled. So it was cheaper to mimic that instead of having 2 parser codes to handle both Dex templates.
* The `kubermaticIssuer` is never used.

So I begun to remove the oidc-proxy-client entirely. Instead of it, we can talk to Dex directly. "But what about the redirect at the end?", you may ask. Fear not! With the magic of inspecting HTTP headers, we can simply observe the response to *our* `POST /dex/auth/local` request and just take the `Location` header, `url.Parse()` it and take the code from there. No server needed.

Similarily, for the login form we can simply observe the Location header when Dex redirects from the initial `/dex/auth?client_id=...&claims=...` request to the login page at `/dex/auth/local?req=<request token>`. No HTML parsing needed.

This got rid of the oidc-proxy-client. It was replaced by the new `api/pkg/test/e2e/api/utils/dex` package, which is called `dex` and not `oauth` or `oidcclient` because it is very specific to Dex' UI, which could change at any time. The package offers a sane `Client` and some nice helpers. However, we can do better.

Earlier we had sprinkled the credentials and issuer stuff all across YAML files, shell scripts and Go constants. In the new code, we simply re-use the existing `oauth_values.yaml` and take everything we need to construct a Dex client.

---

This PR originally only intended to make our OAuth chart flexible enough to not require hacks during e2e. These changes are still included, because they are required.

* A custom ingress class is nice to have and also a good way to make it possible to disable the Ingress alltogether, by using the familiar `non-existent` ingress class.
* The scheme can be changed to HTTP, with a scary warning that people should please never do this in production.
* The naming confusion between `staticPasswords` and `staticPasswordLogins` has been fixed while retaining BC with older versions of the chart.

---

I've left some comments throughout the diff to explain some changes.

Also, I don't want to put anyone down who previously worked on this, but I spent way too much time staring at this code not to be emotional now ^^

```release-note
NONE
```

All this work and not even a release note... that's not how you get promoted, kids.


api/cmd/conformance-tests/main.go
```diff
@@ -74,6 +75,7 @@ type Opts struct {
 	onlyTestCreation             bool
 	pspEnabled                   bool
 	createOIDCToken              bool
+	dexHelmValues                string
```
xrstf: Throughout *new* code I tried to call the thing "dex" and not "oauth". It was IMHO a mistake to ever name the Helm chart "oauth". All other charts are named after their product, never their purpose. We don't have a "logshipper" chart, or a "log viewer", a "metrics-collector" or "dashboard" chart.

At some point I will simply rename the chart.

api/hack/ci/ci-run-cloud-e2e.sh
```diff
@@ -2,26 +2,24 @@
...
 export KUBERMATIC_SCHEME="https"
 export KUBERMATIC_HOST="cloud.kubermatic.io"
 
-
-(
-cd ${SDIR}/../../pkg/test/e2e/api/utils/oidc-proxy-client
-make build
-)
-
-export KUBERMATIC_OIDC_USER=${KUBERMATIC_DEX_DEV_E2E_USERNAME:-"roxy@loodse.com"}
```
xrstf, line 22: No point in defaulting a login name and password that don't even work on cloud.kubermatic.io (NB: on cloud the test user us roxy@`kubermatic.io`). Better to exit early.

api/hack/ci/ci-run-minimal-conformance-tests.sh
```diff
@@ -425,13 +425,13 @@ if ! ls ./api/_build/conformance-tests &>/dev/null; then
 fi
 
 echodate "Starting conformance tests"
-export KUBERMATIC_APISERVER_ADDRESS="kubermatic-api.prow-kubermatic-${BUILD_ID}.svc.cluster.local.:80"
+export KUBERMATIC_SCHEME="http"
+export KUBERMATIC_HOST="kubermatic-api.prow-kubermatic-${BUILD_ID}.svc.cluster.local.:80"
```
xrstf: Already used these envs for other stuff, so I renamed the APISERVER_ variant.

api/hack/ci/ci-setup-kubermatic-in-kind.sh
```diff
@@ -451,22 +461,6 @@ else
...
-
-echodate "Waiting for dex to be ready"
-retry 5 curl --fail http://127.0.0.1:5556/healthz
-echodate "Dex got ready"
-
-echodate "Waiting for api to be ready"
-retry 5 curl --fail http://127.0.0.1:8080/api/v1/healthz
-echodate "API got ready"
-
 function cleanup_kubermatic_clusters_in_kind {
```
xrstf, line 471: Setting the trap up earlier prevents stalling kubectl processes and builds.

api/hack/ci/ci_run_api_e2e.sh
```diff
@@ -47,23 +40,10 @@ spec:
...
   admin: true
 EOF
+echodate "Creating roxy2 user..."
 retry 2 kubectl apply -f user.yaml
 
-# Create and run OIDC proxy client
-# TODO: since OIDC_CLIENT_ID and OIDC_CLIENT_SECRET are defined in the docker image
-#       they could be exposed as envs by that image
-export KUBERMATIC_OIDC_CLIENT_ID="kubermatic"
-export KUBERMATIC_OIDC_CLIENT_SECRET="ZXhhbXBsZS1hcHAtc2VjcmV0"
```
xrstf, line 56: Fun fact: That is actually the secret for cloud.kubermatic.io.

api/hack/ci/testdata/oauth_values.yaml
```diff
@@ -1,81 +1,27 @@
...
-#    userID: "08a8684b-db88-4b73-90a9-3cd1661f5466"
+    scheme: http
+    host: dex.oauth:5556
+    path: "/dex"
+    class: non-existent
   clients:
   - id: kubermatic
     name: Kubermatic
-    secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+    secret: rZpo4bZFgy2Gsjxb5KCmf68K
```
xrstf: Not important, just a new random secret.

api/pkg/test/e2e/api/cluster_rbac_test.go
```diff
@@ -1,6 +1,6 @@
 // +build create
 
-package e2e
+package api
```
xrstf, line 3: I like it when the package name at least somewhat resembles the directory name.

api/pkg/test/e2e/api/utils/dex/client.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	params := loginURL.Query()
+	params.Set("client_id", c.clientID)
+	params.Set("redirect_uri", c.redirectURI)
+	params.Set("response_type", "code")
+	params.Set("scope", "openid profile email")
+	params.Set("state", "we-dont-care-for-replay-attacks-here")
+	// make sure we are redirected and not greeted with a "choose your login method page"
+	params.Set("connector_id", "local")
```
xrstf, line 109: And boom, the "we need to define github just so we have the same markup as on cloud"-issue is gone.

api/cmd/conformance-tests/main.go
alvaroaleman: I very much agree with this, this is super confusing today.

alvaroaleman: Can we rename this var and the corresponding flag to `dexHelmValuesFile`? I initially thought these were the literal values

api/cmd/conformance-tests/main.go
```diff
@@ -238,15 +241,22 @@ func main() {
 		opts.versions = append(opts.versions, semver.NewSemverOrDie(s))
 	}
 
-	kubermaticAPIServerAddress := os.Getenv("KUBERMATIC_APISERVER_ADDRESS")
+	kubermaticAPIServerAddress := os.Getenv("KUBERMATIC_HOST")
```
alvaroaleman, line 246: Nothing for this PR, but at some point we should find some lib that allows to easely define flags and corresponding env vars, so ppl can figure out what vars exist and if they are mandatory

api/hack/ci/ci-run-cloud-e2e.sh
```diff
@@ -2,26 +2,24 @@
...
 
-SDIR=$(dirname $0)
-
-
-export KUBERMATIC_OIDC_CLIENT_ID="kubermatic"
-export KUBERMATIC_OIDC_CLIENT_SECRET="ZXhhbXBsZS1hcHAtc2VjcmV0"
-export KUBERMATIC_OIDC_ISSUER="https://cloud.kubermatic.io/dex"
-export KUBERMATIC_OIDC_REDIRECT_URI="http://localhost:8000"
-export KUBERMATIC_OIDC_ISSUER_URL_PREFIX="dex"
+cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic
```
alvaroaleman, line 5: Can we copy this job as an optional presubmit so we can verify that it still works? :grimacing:

api/hack/ci/ci-run-minimal-conformance-tests.sh
alvaroaleman: This script is basically unused, the only remaining tests on the master branch that use it are the upgrade tests for the weeklys, so plase don't touch it and lets remove it after I upgraded those tests :)

api/cmd/conformance-tests/main.go
```diff
@@ -238,15 +241,22 @@ func main() {
...
 	if kubermaticAPIServerAddress == "" {
-		log.Fatalf("Kubermatic apiserver address must be set via KUBERMATIC_APISERVER_ADDRESS env var")
+		log.Fatal("Kubermatic apiserver address must be set via KUBERMATIC_HOST env var")
 	}
-	opts.kubermaticClient = apiclient.New(httptransport.New(kubermaticAPIServerAddress, "", []string{"http"}), nil)
+	kubermaticAPIServerScheme := os.Getenv("KUBERMATIC_SCHEME")
```
alvaroaleman, line 250: I think this will always be http for the conformance tester, so just hardcode it and do not set it in the scripts until we actually need to make it configurable?

api/pkg/test/e2e/api/utils/dex/helm.go
```diff
@@ -0,0 +1,69 @@
...
+	"os"
+
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
+)
+
+type helmValues struct {
+	Dex struct {
+		Ingress struct {
+			Scheme string `yaml:"scheme"`
```
alvaroaleman, line 14: Nit: Json tags will also work and have the advantage, that they still work when the input is json

api/pkg/test/e2e/api/utils/dex/retry.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type RetryingClient struct {
+	Client Client
+
+	interval time.Duration
+	timeout  time.Duration
+}
+
+func NewRetryingClient(client *Client, interval time.Duration, timeout time.Duration) *RetryingClient {
```
alvaroaleman: What is the reason for having another constructor for this and exposing both a normal and a retrying client? Is there any case where we do _not_ want retrying?

Also IMHO the interval and timeout should be optional and setup via functional opts

api/pkg/test/e2e/api/oidc_test.go
```diff
@@ -0,0 +1,41 @@
+package api
```
alvaroaleman: Can we move the content of this into `oidc.go`? The current naming seems to suggest that it tests `oidc.go` which is not the case.

config/oauth/templates/configmap.yaml
```diff
@@ -17,27 +17,30 @@ data:
...
+    # this is a requirement for static passwords, so we just enable it by default
+    enablePasswordDB: true
 {{ if .Values.dex.expiry }}
     expiry:
 {{ toYaml .Values.dex.expiry | indent 6 }}
 {{- end }}
 {{ if .Values.dex.grpc }}
-    enablePasswordDB: true
     grpc:
-{{ toYaml .Values.dex.grpc.api | indent 7 }}
```
alvaroaleman, line 27: How did this work before?

config/oauth/values.yaml
```diff
@@ -4,8 +4,13 @@ dex:
     tag: "v2.21.0"
   replicas: 2
   ingress:
-    host: ""
+    host: "" # this option is required
```
alvaroaleman: Can we stick to one of "comment after" or "comment above" instead of mixing?

config/oauth/values.yaml
```diff
@@ -30,7 +35,7 @@ dex:
 #    - http://localhost:8000/projects
 #    - https://dev.kubermatic.io/projects
 #
-#  staticPasswordLogins:
+#  staticPasswords:
```
alvaroaleman, line 39: This was just wrong before?

alvaroaleman (commented):
Some nits, but overall this is a great improvement!

Since this unfortunately has a pretty big blastradius, could you open a PR in the dashboard repo that changes the basebranch for the cloned kubermatic repo to the one of this PR, so we can verify that this PR doesn't break the dashboard tests?

api/hack/ci/ci-run-minimal-conformance-tests.sh
xrstf: So I should revert changes to this file in this PR?

api/hack/ci/ci-run-cloud-e2e.sh
xrstf: I let it run locally and could successfully authenticate against cloud.kubermatic.io, the tests only failed because I had no AWS credentials env variables setup in my shell.

api/cmd/conformance-tests/main.go
xrstf: We use scheme and host in other places and I thought it would be nice to use them consistently, even though we only to HTTP here. So I'd like to keep it this way please :-)

api/pkg/test/e2e/api/utils/dex/retry.go
xrstf: Separation of concern was the idea here. I first implemented the non-retrying client and then wrote the RetryingClient as a decorator. I can see that we probably never need a non-retrying, so I could make the `NewClient` private, but I would like to keep the two implementations as they are.

api/pkg/test/e2e/api/oidc_test.go
xrstf: Funny enough, this renaming was one of the last changes. I did it because otherwise the `masterToken` variable became part of the regular app (*gasp* occupying a few bytes of memory!). So I moved it into a _test.go file, even though, as you noted, the naming is slightly confusing (shame on Go that there is no sane way to have a "test helper").

config/oauth/templates/configmap.yaml
xrstf: We never ever had configured the GRPC stuff anywhere, so we never noticed that it was broken.

config/oauth/values.yaml
xrstf: It was just a bad naming choice in our Helm values. Did not dig further when it was introduced, was just focussing on fixing it ^^

api/pkg/test/e2e/api/utils/dex/helm.go
xrstf: JSON tags work in yaml.v2? Even if it did, I find it a bit confusing to load a YAML file and parse it as JSON.

api/hack/ci/ci-run-minimal-conformance-tests.sh
alvaroaleman: yeah

api/cmd/conformance-tests/main.go
alvaroaleman: Adding a mandatory config option for something we do not plan to have more than one value is IMHO almost the definition of premature optimization

api/pkg/test/e2e/api/utils/dex/helm.go
alvaroaleman: in sigs.k8s.io/yaml, yes. This is how its done pretty much everywhere, e.G. in the types used for the versions in `pkg/version/manager.go`

api/pkg/test/e2e/api/utils/dex/retry.go
alvaroaleman: Yeah that sounds fine, just do not make the non-retrying variant public, because then ppl will end up using it

xrstf:
The UI team seems to be happy and their tests are shown to continue to work with this, so I guess we can do a final round of review and then merge :-)

api/cmd/conformance-tests/main.go
```diff
@@ -184,6 +186,7 @@ func main() {
 	flag.BoolVar(&opts.onlyTestCreation, "only-test-creation", false, "Only test if nodes become ready. Does not perform any extended checks like conformance tests")
 	_ = flag.Bool("debug", false, "No-Op flag kept for compatibility reasons")
 	flag.BoolVar(&opts.createOIDCToken, "create-oidc-token", false, "Whether to create a OIDC token. If false, environment vars for projectID and OIDC token must be set.")
+	flag.StringVar(&opts.dexHelmValuesFile, "dex-helm-values-file", "", "Helm values.yaml of the OAuth (Dex) chart to read and configure a matching client for. Only needed if -create-oidc-token is enabled.")
```
alvaroaleman, line 191: We can not add new args, because they break upgrade tests, as we use the same flags for both old and new version of the conformance-tester, hence we have to use env vars

alvaroaleman:
/lgtm
/hold

for upgrade test to finish.

xrstf:
@alvaroaleman all regular and the upgrade test now pass.

To make the upgrade tests work, I bit the bullet and restored the oidc-proxy-client *temporarily* and left a big fat "DEPRECATED" notice on it. This is required because the upgrade tests will use the old branch's conformance tester, which does not yet know how to get the token from Dex directly.

The alternative would have been to make another PR to make the conformance-tester more flexible, then backport this into 2.12 and then re-apply this PR again. Not fun.

api/hack/ci/ci-run-conformance-tester.sh
```diff
@@ -67,10 +70,14 @@ if [ -n "${UPGRADE_TEST_BASE_HASH:-}" ]; then
   kubermatic_delete_cluster="false"
 fi
 
-echodate "Building OIDC proxy"
+# this is only used for upgrade tests, remove after 2.13 is released
```
alvaroaleman: Please use `TODO` for such cases so we can easely grep. Also, can you comment on Kamils PR for his release checklist and mention "Remove TODOS in conformance test code and scripting"? else this will be forgoton

api/hack/ci/ci-run-conformance-tester.sh
```diff
@@ -67,10 +70,14 @@ if [ -n "${UPGRADE_TEST_BASE_HASH:-}" ]; then
...
+echodate "Building legacy OIDC proxy"
 make -C api/pkg/test/e2e/api/utils/oidc-proxy-client build
 echodate "Finished building OIDC proxy"
 
+# this is also just for pre 2.13-compat
```
alvaroaleman: use `TODO`, do not expect ppl to read your code serially

api/pkg/test/e2e/api/utils/dex/helm.go
```diff
@@ -0,0 +1,64 @@
...
+	} `yaml:"dex"`
+}
+
+// NewClientFromHelmValues is a helper for e2e tests, reading the hack/ci/testdata/oauth_values.yaml
+// to provide a matching OIDC client. We use this instead of spreading the client ID etc.
+// in tons of shell scripts and env vars.
+func NewClientFromHelmValues(valuesFile string, clientID string, log *zap.SugaredLogger) (*Client, error) {
+	values := helmValues{}
+
+	f, err := os.Open(valuesFile)
```
alvaroaleman, line 32: Wouldn't `io/ioutil.ReadFile` be simpler?

api/pkg/test/e2e/api/utils/dex/retry.go
alvaroaleman: This didn't get addressed, please unexport the non-retrying client

api/pkg/test/e2e/api/utils/oidc_client.go
```diff
@@ -1,3 +1,5 @@
+// Package utils is deprecated and only used for upgrade tests from
```
alvaroaleman: Please use a `TODO` for the removal after the release

api/pkg/test/e2e/api/utils/dex/helm.go
xrstf: Well *technically* using ReadFile + Unmarshal means that I have to read the whole YAML file into memory, instead of parsing it as a stream. Also code-wise ReadFile+Unmarshal are not really that much shorter, if even.

api/hack/ci/ci-run-conformance-tester.sh
xrstf: Left a comment, but was originally intending to open a dedicated issue for cleaning this up after the 2.13 release.

alvaroaleman: Dedicated issue may also be a good idea

alvaroaleman:
/lgtm
/approve
/hold
/test pre-kubermatic-e2e-upgrade
Because I am paranoid regarding the upgrade test. Feel free to remove the hold once it passed.

PR #4887, opened 2019-12-16
Openshift console: Fix login

### LINKED ISSUE #4880: Access to OpenShift Console not working
**User Story**

Using the `OpenConsole` button results in the following error message:
`did not get an OAuth code back from Openshift OAuth server<a href="/api/v1/projects/hd8tsh8xjf/dc/europe-west3-c/clusters/wthlsngrd6/openshift/console/proxy/auth/callback?code=&amp;state=">Found</a>.`

**Acceptance criteria**

OpenShift Console should be shown

alvaroaleman:

When moving the management of the bootstrap secret into the usercluster controller, i started syncing the password hash rather than the password. This obviously does not work, hence this PR changes the whole thing to:

* Store the password encrypted via the UID of the CA secret that is in the seed in the usercluster secret
* Decrypt it and sync that into the seed

Fixes #4880 

~A further improvement could probably be move the reconciling of the secret inside the usercluster controller from the usercluster resource controller into the openshift-seed-syncer and then just check if the value in the seed when hashed matches whats in the usercluster. This is a bit tricky thought, as it ties the creation of two secrets together.~

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/user-cluster-controller-manager/openshift-seed-syncer/controller.go
+++ b/api/pkg/controller/user-cluster-controller-manager/openshift-seed-syncer/controller.go
@@ -204,7 +204,7 @@ func oauthBootstrapSecretCreatorGetter(userClusterClient ctrlruntimeclient.Clien
 			if s.Data == nil {
 				s.Data = map[string][]byte{}
 			}
-			s.Data[name] = userClusterOAuthSecret.Data[name]
+			s.Data[name] = userClusterOAuthSecret.Data[userclusteropenshiftresources.OAuthBootstrapUnencryptedKeyName]
 
 			return s, nil
 		}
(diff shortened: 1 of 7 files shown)


xrstf:
We could rot13 it for added protection and a :white_check_mark: on the next audit.

/lgtm

alvaroaleman:
/hold

Lets encrypt the raw password rather than storing it in plaintext

alvaroaleman:
/hold cancel
@xrstf Updated the PR to not store the password in plaintext in the usercluster, PTAL

xrstf:
I'm not sure the UIDs are a good source of crypto randomness, but it should be good enough(tm).

/lgtm

PR #4912, opened 2019-12-20
Fix setting overrides for kube-controller-manager in OpenShift controller

xmudrii:

Fix setting overrides for kube-controller-manager in OpenShift controller. This was overlooked in #4810.

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
+++ b/api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
@@ -295,7 +295,11 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 					kubeControllerManagerContainerName: defaultKubeControllerResourceRequirements.DeepCopy(),
 					openvpnSidecar.Name:                openvpnSidecar.Resources.DeepCopy(),
 				}
-				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, resources.GetOverrides(data.Cluster().Spec.ComponentsOverride), dep.Annotations)
+				overrides := map[string]*corev1.ResourceRequirements{}
+				if data.Cluster().Spec.ComponentsOverride.ControllerManager.Resources != nil {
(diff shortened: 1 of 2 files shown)


api/pkg/controller/seed-controller-manager/openshift/resources/kube_controller_manager.go
```diff
@@ -295,7 +295,11 @@ func KubeControllerManagerDeploymentCreatorFactory(data kubeControllerManagerDat
 					kubeControllerManagerContainerName: defaultKubeControllerResourceRequirements.DeepCopy(),
 					openvpnSidecar.Name:                openvpnSidecar.Resources.DeepCopy(),
 				}
-				err = resources.SetResourceRequirements(dep.Spec.Template.Spec.Containers, defResourceRequirements, resources.GetOverrides(data.Cluster().Spec.ComponentsOverride), dep.Annotations)
+				overrides := map[string]*corev1.ResourceRequirements{}
```
alvaroaleman: I don't get it, the way it was before looks correct to me, it should be using `resources.GetOverrides` or not?

xmudrii: @alvaroaleman Names of containers are different, so overrides would never get applied. In this case, the container is called `kube-controller-manager`, but the `GetOverrides` function uses the `resources.ControllerManagerDeploymentName` const which equals to just `controller-manager`.

Another possibility is to rename the container. But I'm not sure about side effects, such as, when you upgrade Kubermatic, it would rotate all controller manager pods managed by this controller.

alvaroaleman: Yeah it will rotate all of them but IMHO thats the better option here

alvaroaleman (approved):
/lgtm
/approve

alvaroaleman:
/hold

Tests are legitimately failing here

xmudrii:
@alvaroaleman Test should be fixed now. I guess this is how it should be done, but if it isn't okay, let me know.

alvaroaleman (approved):
/lgtm

PR #4913, opened 2019-12-20
Addons: Allow specyfing required api resources

### LINKED ISSUE #4881: Addon reconciling errors during OpenShift cluster creation
**User Story**

Errors during cluster creation:

Eventually the cluster is created successfully and working.

**Acceptance criteria**

Cluster should be created without reconciling errors

alvaroaleman:

* Allows addons to specify that they need some resource to exist and have the addon controller check if thats the case, before applying them
* Add new `-{openshift,kubernetes}-addons-file` flags that are mutually exclusive with the existing `-{openshift,kubernetes}-addons-list` flag and allow to specify more options for the default addons than just the name
* Adds the required resources to the openshift `network` and `registry` addon
* Moves the setting of the `spec.IsDefault` boolean from the `addon` to the `addoninstaller` controller, because the `addon` controller shouldn't need to know which addons are default

The initial approach to this problem was to allow addons to specify other addons as being required and to store a `SuccessfullyDeployed` info in the `.Status` of each addon, allowing the addon controller to wait with one addon until all its required addons are deployed successfully.

I decided against this approach, because it will force us to implement something similiar for cleanup: If an addon a) is required by another addon b), deleting a) has to delete b) first. That would force us to implement dependency management for addons, which is not something we want to support or export as public API.

Fixes #4881 

**Documentation**:

```release-note
none
```


api/cmd/seed-controller-manager/options.go
```diff
@@ -98,8 +106,10 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
 	flag.StringVar(&c.openshiftAddonsPath, "openshift-addons-path", "/opt/addons/openshift", "Path to addon manifests. Should contain sub-folders for each addon")
-	flag.StringVar(&c.kubernetesAddonsList, "kubernetes-addons-list", "canal,csi,dns,kube-proxy,openvpn,rbac,kubelet-configmap,default-storage-class,node-exporter,nodelocal-dns-cache,logrotate", "Comma separated list of Addons to install into every user-cluster")
-	flag.StringVar(&c.openshiftAddonsList, "openshift-addons-list", "openvpn,rbac,crd,network,default-storage-class,registry,logrotate", "Comma separated list of addons to install into every openshift user cluster")
+	flag.StringVar(&defaultKubernetesAddonsList, "kubernetes-addons-list", defaultKubernetesAddons, "Comma separated list of Addons to install into every user-cluster. Mutually exclusive with `--kubernetes-addons-file`")
+	flag.StringVar(&defaultKubernetesAddonsFile, "kubernetes-addons-file", "", "File that contains a list of default kubernetes addons. Mutually exclusive with `--openshift-addons-list`")
```
xrstf: Must be `mutually exclusive with *--kubernetes-addons-list* `.

api/hack/run-controller.sh
```diff
@@ -20,9 +20,9 @@ KUBERMATIC_WORKERNAME=${KUBERMATIC_WORKERNAME:-$(uname -n)}
   -kubeconfig=../../secrets/seed-clusters/dev.kubermatic.io/kubeconfig \
   -versions=../config/kubermatic/static/master/versions.yaml \
   -updates=../config/kubermatic/static/master/updates.yaml \
-  -master-resources=../config/kubermatic/static/master \
   -kubernetes-addons-path=../addons \
   -openshift-addons-path=../openshift_addons \
+  -openshift-addons-file=../config/kubermatic/static/master/openshift-addons.yaml \
```
xrstf, line 25: The documentation for the new CLI flags says that -file and -path are mutually exclusive. Why are both given here?

config/kubermatic/templates/kubermatic-controller-manager-dep.yaml
```diff
@@ -69,8 +68,18 @@ spec:
...
+        {{- end }}
+        {{- if .Values.kubermatic.controller.addons.kubernetes.defaultAddonsFile }}
+        - -kubernetes-addons-file=/opt/master-files/{{ .Values.kubermatic.controller.addons.kubernetes.defaultAddonsFile }}
+        {{- end }}
+        {{- if .Values.kubermatic.controller.addons.openshift.defaultAddons }}
         - -openshift-addons-list={{ join "," .Values.kubermatic.controller.addons.openshift.defaultAddons }}
+        {{- end }}
+        {{- if .Values.kubermatic.controller.addons.openshift.defaultAddonsFile }}
+        - -openshift-addons-file=/opt/master-files/{{ .Values.kubermatic.controller.addons.openshift.defaultAddonsFile }}
+        {{- end }}
```
xrstf, line 82: These changes should also be reflected in `api/pkg/controller/operator`. Likewise the new default openshift addon file must be a default value in the operator (like it already defaults the versions.yaml).

api/hack/run-controller.sh
alvaroaleman: No, it says the `-list` and `-path` flags are mutually exclusive.

`-path` is the location of the addon manifests.

PR #4917, opened 2019-12-23
Fix patch requests send from go-kubermatic client

mfranczy:
Currently, the generated go-kubermatic client is not able to send an appropriate PATCH request to the API server, it's because of the `Patch` field type which is `[]byte` ([example](https://github.com/kubermatic/kubermatic/pull/4917/files#diff-2ace21458585d6d9d5e1ce87136e0fdcL991)) instead it should be `json.RawMessage`. 

`[]byte` representation in JSON is base64 so the API server returns bad request code when receiving such a payload

Additionally, the go-swagger pkg must be upgraded to handle json.RawMessage type.

```release-note
NONE
```


api/pkg/handler/v1/cluster/kubeconfig.go
```diff
@@ -348,12 +348,12 @@ type CreateOIDCKubeconfigReq struct {
...
-	encodedState     string
-	decodedState     OIDCState
-	phase            int
-	cookieNonceValue string
+	// skip for swagger spec
+	code             string    `json:"-"`
+	encodedState     string    `json:"-"`
+	decodedState     OIDCState `json:"-"`
+	phase            int       `json:"-"`
+	cookieNonceValue string    `json:"-"`
```
mfranczy: the new swagger version adds to spec unexported params it seems I will have to fix it there

mfranczy: I created a new issue with PR, https://github.com/go-swagger/go-swagger/issues/2164, let's wait for their response

mfranczy:
`pre-kubermatic-verify` job fails because of https://github.com/go-swagger/go-swagger/issues/2066, shortly I will create a PR with fix there.

zreigz (changes requested):
You need to apply changes (new parameter names) for this file: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/test/e2e/api/runner.go
Then tests should pass

zreigz:
I don't see any version changes in `Gopkg.toml` only in autogenerated `Gopkg.lock` file. The command `dep ensure` can override the changes in my opinion

mfranczy:
@zreigz thanks for review

> You need to apply changes (new parameter names) for this file: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/test/e2e/api/runner.go
Then tests should pass

yeah I am fixing this now, I see there are more problems with naming.

mfranczy:
> I don't see any version changes in Gopkg.toml only in autogenerated Gopkg.lock file. The command dep ensure can override the changes in my opinion

I see we use go-swagger master branch that's why Gokpg.toml didn't change, however, I will lock to v0.22.0 version if go-swagger folks will agree to backport this PR https://github.com/go-swagger/go-swagger/pull/2209

api/pkg/api/v1/types.go
```diff
@@ -610,7 +610,7 @@ const (
 // Cluster has a custom MarshalJSON method defined
 // and thus the output may vary
 //
-// swagger:model Cluster
+// swagger:model ClusterResource
```
mfranczy: This had to be renamed to `ClusterResource` we have a name collision with `cmdv1.Config` (Kubeconfig) it contains `Cluster` field (it comes from vendor pkg so I can do nothing about that).

However from the client usage perspective nothing changes, we can still use `model.Cluster`.

zreigz:
nice work, you need to run `update-swagger` and `gen-api-client` again

api/pkg/api/v1/types.go
mfranczy: damn it, now swagger generator is inconsistent with results (similar as spec in the past).... The above comment is not valid. Sometimes generator produces output for `ClusterResource` and sometimes for `Cluster` from `cmdv1.Config`.

mfranczy:
@zreigz thanks, `gen-api-client` won't work because of https://github.com/kubermatic/kubermatic/pull/4917#discussion_r375900556, swagger has so many bugs, it doesn't know how to handle embedded models. I will try to find a way how to fix that in our code.

zreigz:
> @zreigz thanks, `gen-api-client` won't work because of [#4917 (comment)](https://github.com/kubermatic/kubermatic/pull/4917#discussion_r375900556), swagger has so many bugs, it doesn't know how to handle embedded models. I will try to find a way how to fix that in our code.

Ach, OK. Thanks. Tests are passing now :+1:

mfranczy:
@zreigz I had to change `Kubeconfig` to be swagger `response` not `model` (https://github.com/kubermatic/kubermatic/pull/4917/commits/d61ddecf2a22a0679731d9392b8aa3e4a2f6c50a#diff-7f8b07edb0cdb9a092e601081f5ddc8bR501) I think it's even better because it's used only as a response type for `api/v1/kubeconfig` (https://github.com/kubermatic/kubermatic/blob/master/api/pkg/handler/routes_v1_optional.go#L35) and sadly it cannot be embedded struct.

Let's try to re-run tests a few times just to be sure that go-swagger does not surprise us, locally after a few iterations of `swagger generate` I had consistent results.

zreigz:
everything works fine, thanks for this PR :+1:

api/cmd/kubermatic-api/swagger.json
```diff
@@ -231,11 +231,7 @@
             "name": "Patch",
             "in": "body",
             "schema": {
-              "type": "array",
```
kron4eg, line 395: Isn't such change in API is a breaking change?

mfranczy: It's not. Before `Patch` fields were of `[]byte` type, they should be `json.RawMessage` (https://github.com/kubermatic/kubermatic/pull/4917/commits/c36b8aa64cb624e2b0489dd862588433d3340c39#diff-a57f88db269544b698f3e8b0e675c4d2L73) which is basically the `[]byte` type with Marshal and Unmarshal functions. So the type is still the same for `Patch` fields.

However, when it comes to type change in swagger.json, `json.RawMessage` for go-swagger is an `object`, for `[]byte` is `[]uint8`, so after that change, finally, the generated kubermatic client will be able to send an appropriate payload which is not base64 encoded (`[]byte` in json is base64 encoded). 

That change from `[]uint8` to `object` in swagger is very demaned and it's not a breaking change as it didn't work before.

mfranczy:
I had to rebase the branch against master, so please take a look again after all tests are green

mfranczy:
in the meantime, some other PR has been merged into master, I had to rebase and regenerate the client again

PR #4919, opened 2019-12-23
Operator: Fix addon defaulting

alvaroaleman:

In #4913 the operator was hardcoded to use addon manifests configured via the master files and ignore the `Spec.UserCluster.Addons.Openshift.Default` setting.

This PR makes it correctly set either of the two and only default if both are unset. Also, the setting was moved from the master files map into the `KubermaticAddonConfiguration` and the controller was updated to also use either of the two for Kubernetes.
Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/api/pkg/controller/operator/common/defaults.go
+++ b/api/pkg/controller/operator/common/defaults.go
@@ -129,7 +129,7 @@ func DefaultConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logg
 
 	if len(copy.Spec.UserCluster.Addons.Openshift.Default) == 0 && copy.Spec.UserCluster.Addons.Openshift.DefaultManifests == "" {
 		copy.Spec.UserCluster.Addons.Openshift.DefaultManifests = strings.TrimSpace(defaultOpenshiftAddons)
-		logger.Debugw("Defaulting field", "field", "usercluster.Addons.Openshift.DefaultManifests")
+		logger.Debugw("Defaulting field", "field", "userCluster.Addons.Openshift.DefaultManifests")
 	}
 
 	if len(copy.Spec.API.AccessibleAddons) == 0 {
--- a/api/pkg/controller/operator/common/defaults.go
+++ b/api/pkg/controller/operator/common/defaults.go
@@ -38,16 +38,6 @@ var (
 		"logrotate",
 	}
 
-	openshiftDefaultAddons = []string{
-		"crd",
-		"openvpn",
-		"rbac",
-		"network",
-		"default-storage-class",
-		"registry",
-		"logrotate",
-	}
-
 	defaultAccessibleAddons = []string{
 		"node-exporter",
 	}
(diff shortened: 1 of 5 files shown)


api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
```diff
@@ -149,30 +148,40 @@ func SeedControllerManagerDeploymentCreator(workerName string, versions common.V
...
 			}
 
-			if len(cfg.Spec.MasterFiles) > 0 {
-				args = append(
-					args,
-					"-versions=/opt/master-files/versions.yaml",
-					"-updates=/opt/master-files/updates.yaml",
-					"-openshift-adons-file=/opt/master-files/openshift-addons.yaml",
-				)
+			args = append(
```
alvaroaleman, line 151: I've removed the condition, because not having these is a defunct config that will send the controller manager into a crashloop. Also IMHO we should not have the `MasterFiles map[string]string` but instead specify what exactly we want (versions, updates) and have settings for those.

Thats a topic for another PR thought.

api/pkg/controller/operator/common/resources.go
```diff
@@ -81,9 +88,15 @@ func DexCASecretCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconcili
 }
 
 func MasterFilesSecretCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedSecretCreatorGetter {
+	data := map[string]string{}
```
alvaroaleman, line 91: I am copying the map here because otherwise I would have to copy the value of `Addons{Kubernetes,Openshift}.DefaultManifests` into the `MasterFiles` map, making them appear twice in the generated sample and in the CRs in the cluster, even thought the latter will always be identical to the former.

api/pkg/controller/operator/common/defaults.go
```diff
@@ -132,14 +122,14 @@ func DefaultConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logg
...
 		copy.Spec.UserCluster.Addons.Kubernetes.Default = kubernetesDefaultAddons
 		logger.Debugw("Defaulting field", "field", "userCluster.addons.kubernetes.default", "value", copy.Spec.UserCluster.Addons.Kubernetes.Default)
 	}
 
-	if len(copy.Spec.UserCluster.Addons.Openshift.Default) == 0 {
-		copy.Spec.UserCluster.Addons.Openshift.Default = openshiftDefaultAddons
-		logger.Debugw("Defaulting field", "field", "userCluster.addons.openshift.default", "value", copy.Spec.UserCluster.Addons.Openshift.Default)
+	if len(copy.Spec.UserCluster.Addons.Openshift.Default) == 0 && copy.Spec.UserCluster.Addons.Openshift.DefaultManifests == "" {
+		copy.Spec.UserCluster.Addons.Openshift.DefaultManifests = strings.TrimSpace(defaultOpenshiftAddons)
+		logger.Debugw("Defaulting field", "field", "usercluster.Addons.Openshift.DefaultManifests")
```
xrstf: `userCluster` please :)

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
xrstf: `though`

Also true, due to the defaulting this condition can never be false, except for synthetic test cases. So it made no real sense anymore.

And yeah, I didn't like the MasterFiles concept either, but I assumed it was an intended mechanism for customers to mount arbitrary files into our containers for whatever reason, so I kept it. Would definitely support removing it and having the two master files be actual go structs (pls no yaml string :-D).

xrstf (changes requested):
Just one nitpick.

alvaroaleman:
@xrstf updated, PTAL

xrstf (approved):

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
alvaroaleman: To my knowledge it isn't used like that. And even if it was, I would prefer something like `additionalMasterFiles` and have something properly typed for the ones that are mandatory.

PR #4932, opened 2020-01-06
Replica/Ingress-Class Support for Operator

xrstf:
This PR has been extracted from a larger branch with the goal of using the Kubermatic Operator in our e2e tests. For this the operator needs to support a few more minor options. And while I was at it, this PR also includes a fix for the webhook reconciling, which ended up in a endless loop because we didn't correctly default some fields.

Relates to #4723 

```release-note
NONE
```


api/pkg/controller/operator/master/reconciler.go
```diff
@@ -300,6 +305,11 @@ func (r *Reconciler) reconcileIngresses(config *operatorv1alpha1.KubermaticConfi
 }
 
 func (r *Reconciler) reconcileCertificates(config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
+	if config.Spec.Ingress.ClassName == common.NonExistentIngressClass && config.Spec.Ingress.CertificateIssuer.Name == "" {
```
kron4eg: What will happen if ingress class is non-existing, but CertificateIssuer.Name has some value?

Such situation probably is a miss-configuration. But it might be not, if the original intention of `NonExistentIngressClass` was to provide ability to configure custom Ingress outside of the operator. In that case `NonExistentIngressClass` name (of the constant and its value) is misleading and should be renamed to something like `CustomIngressClass` or `DoNotManageIngressClass` or something similar.

kron4eg (commented):
mostly LGTM, but have a question to clarify.

api/pkg/controller/operator/master/reconciler.go
xrstf: The original intent behind `NonExistentIngressClass` is solely to disable Ingress creation, primarily for the e2e tests. The current e2e test stuff is hacky and simply does `rm config/kubermatic/templates/ingress.yaml` to prevent the Ingress from appearing, but that is brittle and I flat out don't really like it ^^ Because of this I wanted to put the ability to disable the Ingress right into the Operator and since we already use the concept of `non-existent` in other places, it seemed like a good fit.

> What will happen if ingress class is non-existing, but CertificateIssuer.Name has some value?

No Ingress should be created, but the Certificate should be reconciled. Without an Ingress it probably will never be successfully retrieved, but that should not hurt the operator or Kubermatic itself.

kron4eg: >primarily for the e2e tests

OK then, sounds good, except the name :) `DevE2EIngressClass` / `E2ETestsIngressClass` maybe?

xrstf: The "non-existent" value is the one we use everywhere for "disable Ingress creation", so I don't think we should conflate it with e2e tests. It was just a nice coincidence to kill two birds with one stone.

kron4eg: re: "everywhere", I only found it (outside of helm charts) in 3 places:

```
hack/ci/ci-setup-kubermatic-in-kind.sh
320:  --set=kubermatic.ingressClass=non-existent \

hack/ci/testdata/oauth_values.yaml
14:    class: non-existent

hack/ci/ci-run-offline-test.sh
227:  --set=kubermatic.ingressClass=non-existent \
```

It looks like a good opportunity to fix this leaked-from-the-helm "hack" and have some boolean flag in API to control Ingress creation.

xrstf: True, I might have had some more uses of `non-existent` in mind. However if we remove it, we lose the ability to override the ingress class, something I cannot decide. If we can remove this option, sure, let's add a dedicated `disableIngress` flag. If not, I would propose keeping the old behaviour.

kron4eg: on no, I didn't meant to remove `ingressClass`, I meant to add `ingressEnabled bool` in the operator.

api/pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -179,6 +182,27 @@ type KubermaticAddonConfiguration struct {
...
+	Domain string `json:"domain"`
+
+	// ClassName is the Ingress resource's class name, used for selecting the appropriate
+	// ingress controller.
+	ClassName string `json:"className,omitempty"`
+
+	// Disable will prevent an Ingress from being created at all. This is mostly useful
+	// during testing. If the Ingress is disabled, the CertificateIssuer setting can also
+	// be left empty, as no Certificate resource will be created.
+	Disable bool `json:"disable,omitempty"`
```
xrstf, line 197: `disable` or `disabled`?

kron4eg: disable 👍

kron4eg (approved):
/lgtm
/approve

PR #4936, opened 2020-01-08
Add configurable ExposeStrategy on Seed cluster level

### LINKED ISSUE #4721: Configure ExposeStrategy per SeedCluster
**User Story**
As a customer running Kubermatic on prem in vSphere, I am using MetalLB with `exposeStrategy: LoadBalancer`
https://docs.kubermatic.io/concepts/expose-strategy/expose_strategy/

This works fine for my on-prem workload, because it allows my firewalling team to have fixed port allocations, for a fixed subnet of IP addresses.

But now I want to add a new Seed Cluster on a cloud provider to my Kubermatic Installation and I would like to use the `NodePortProxy` expose strategy for this new seed cluster, so I don't have lot's of expensive LoadBalancers.

**Acceptance criteria**
- [ ] `exposeStrategy` can be configured on seed cluster level (or even finer)

moelsayed:
Adds configurable ExposeStrategy on Seed cluster level
Fixes #4721

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
+++ b/api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
@@ -83,8 +83,14 @@ func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, client ctrlruntimeclient
 	seedCreators := []reconciling.NamedSeedCreatorGetter{
 		seedCreator(seed),
 	}
-	if !isValidExposeStrategy(seed.Spec.ExposeStrategy) {
-		return fmt.Errorf("failed to validate seed: invalid Seed Expose Strategy %s", seed.Spec.ExposeStrategy)
+	supportedStrategies := map[corev1.ServiceType]struct{}{
+		corev1.ServiceTypeNodePort:     {},
+		corev1.ServiceTypeLoadBalancer: {},
+	}
+	if seed.Spec.ExposeStrategy != "" {
+		if _, ok := supportedStrategies[seed.Spec.ExposeStrategy]; !ok {
+			return fmt.Errorf("failed to validate seed: invalid Seed Expose Strategy %s", seed.Spec.ExposeStrategy)
+		}
 	}
 	if err := reconciling.ReconcileSeeds(r.ctx, seedCreators, seed.Namespace, client); err != nil {
 		return fmt.Errorf("failed to reconcile seed: %v", err)
@@ -140,8 +146,4 @@ func (r *Reconciler) cleanupDeletedSeed(seedInMaster *kubermaticv1.Seed, seedCli
 	}
 
 	return nil, nil
-}
-
-func isValidExposeStrategy(es corev1.ServiceType) bool {
-	return es == "" || (es == corev1.ServiceTypeNodePort || es == corev1.ServiceTypeLoadBalancer)
 }
--- a/api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
+++ b/api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
@@ -83,7 +83,9 @@ func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, client ctrlruntimeclient
 	seedCreators := []reconciling.NamedSeedCreatorGetter{
 		seedCreator(seed),
 	}
-
+	if !isValidExposeStrategy(seed.Spec.ExposeStrategy) {
+		return fmt.Errorf("failed to validate seed: invalid Seed Expose Strategy %s", seed.Spec.ExposeStrategy)
+	}
 	if err := reconciling.ReconcileSeeds(r.ctx, seedCreators, seed.Namespace, client); err != nil {
 		return fmt.Errorf("failed to reconcile seed: %v", err)
 	}
@@ -138,4 +140,8 @@ func (r *Reconciler) cleanupDeletedSeed(seedInMaster *kubermaticv1.Seed, seedCli
 	}
 
 	return nil, nil
+}
+
+func isValidExposeStrategy(es corev1.ServiceType) bool {
+	return es == "" || (es == corev1.ServiceTypeNodePort || es == corev1.ServiceTypeLoadBalancer)
 }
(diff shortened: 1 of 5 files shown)


api/pkg/controller/master-controller-manager/seed-sync/reconciler_test.go
```diff
@@ -123,3 +130,9 @@ func TestReconcilingSeed(t *testing.T) {
 		})
 	}
 }
+
+func TestInvalidExposeStrategy(t *testing.T) {
```
moadqassem: Can you please add a test case inside the test table instead of creating a separate test for it? I mean eventually, this is part of the reconciliation cycle.

api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
```diff
@@ -139,3 +141,7 @@ func (r *Reconciler) cleanupDeletedSeed(seedInMaster *kubermaticv1.Seed, seedCli
 
 	return nil, nil
 }
+
+func isValidExposeStrategy(es corev1.ServiceType) bool {
```
moadqassem: If I see this right, this function is only used here and in the test. Can you please just add it as a condition in the method body instead of extracting it explicitly? To test it please see the comment above https://github.com/kubermatic/kubermatic/pull/4936/files#r364148683

api/pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -65,6 +65,8 @@ type SeedSpec struct {
 	// Optional: ProxySettings can be used to configure HTTP proxy settings on the
 	// worker nodes in user clusters. However, proxy settings on nodes take precedence.
 	ProxySettings *ProxySettings `json:"proxy_settings,omitempty"`
+	// Optional: ExposeStrategy explicitly sets the expose strategy for this seed cluster, if not set, the default provided by the master is used.
+	ExposeStrategy corev1.ServiceType `json:"exposeStrategy,omitempty"`
```
moadqassem: I think the json tag is wrong! should be ` expose_strategy` right?

api/pkg/controller/master-controller-manager/seed-sync/reconciler_test.go
```diff
@@ -21,6 +22,11 @@ import (
 	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
+const (
```
moadqassem: Those two constants should be added in the test case which tests the exposeStrategy :)

api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
moelsayed: I usually wrap multi-condition if statements in a function because it's easier to read the main code path. Being easier to test in this case is just a happy coincidence :)
What do you think? wrap or unwrap ?

api/pkg/controller/master-controller-manager/seed-sync/reconciler_test.go
moelsayed: I initially tried to do this, but ensuring the fail case in the test table structure is kind of messy. So, I put the "happy path" case in the test table and put the fail path in a separate test.

I will try to include it in the table cleanly.

api/pkg/controller/master-controller-manager/seed-sync/reconciler.go
moadqassem: Sure, but the conditions which are used in this place are not too complex and convoluted to have them in separate function especially in the form(one-liner) that you chose to represent the conditions. Plus as I said earlier those conditions are very specific to this case and aren't used in other places ;)

api/pkg/controller/master-controller-manager/seed-sync/reconciler_test.go
```diff
@@ -80,6 +120,25 @@ func TestReconcilingSeed(t *testing.T) {
...
+					Name:      "my-seed",
+					Namespace: "kubermatic",
+				},
+				Spec: kubermaticv1.SeedSpec{
+					Country:        "Germany",
+					ExposeStrategy: corev1.ServiceTypeClusterIP,
+				},
+			},
+			existingSeeds: existingSeeds,
+			validate: func(input, result *kubermaticv1.Seed) error {
```
moadqassem: Please remove the `validate` field:  and inside the test check if the `validate` function is `nil`: 
```
if test.validate != nil {
  if err := test.validate(test.input, result); err != nil {
    t.Fatalf("reconciling did not lead to a valid end state: %v", err)
  }
}

File: api/pkg/controller/master-controller-manager/seed-sync/reconciler.go [hunk 6]
```diff
@@ -83,7 +83,15 @@ func (r *Reconciler) reconcile(seed *kubermaticv1.Seed, client ctrlruntimeclient
 	seedCreators := []reconciling.NamedSeedCreatorGetter{
 		seedCreator(seed),
 	}
-
+	supportedStraties := map[corev1.ServiceType]bool{
```
moadqassem: typo supportedStraties/supportedStrategies

moadqassem: Nit: Also since the values of the map don't matter you can use empty structs `struct{}` instead of `bool`

api/pkg/controller/master-controller-manager/seed-sync/reconciler_test.go
```diff
@@ -22,8 +23,26 @@ import (
...
+				},
+			},
+			Spec: kubermaticv1.SeedSpec{
+				Country: "Val Verde",
+			},
+		},
+	}
 	tests := []struct {
 		name          string
+		shouldFail    bool
```
moadqassem, line 45: Why this is needed?

moelsayed: because this test case will always err out from the reconciler. This is to flag it as an always fail test.

PR #4948, opened 2020-01-10
changelog for v2.12.5

kdomanski:
```release-note
NONE
```

/assign @xrstf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+### [v2.12.5]()
+
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+### [v2.12.5]()
+
+
+- A bug that occasionally resulted in a `Error: no matches for kind &#34;MachineDeployment&#34; in version &#34;cluster.k8s.io/v1alpha1&#34; visible in the UI was fixed. [#4870](https://github.com/kubermatic/kubermatic/issues/4870) ([alvaroaleman](https://github.com/alvaroaleman))
```
xrstf: The backtick quoting didn't really work here. The closing ` is missing.

Also why the HTML encoding?

xrstf (changes requested):
Nits

CHANGELOG.md
kdomanski: I fixed the quotation. The HTML encodings seems to be what we get from the API when `gchl` scrapes the release-notes from the github API. Usually it works just fine, except for inside backtick quotations.

PR #4959, opened 2020-01-13
ci-setup-kubermatic-in-kind.sh: checkout corresponding charts when deploying older kubermatic

kdomanski:
```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-setup-kubermatic-in-kind.sh
+++ b/api/hack/ci/ci-setup-kubermatic-in-kind.sh
@@ -331,6 +331,11 @@ if [[ "${KUBERMATIC_USE_OPERATOR}" = "false" ]]; then
   TEST_NAME="Deploy Kubermatic"
   echodate "Deploying Kubermatic using Helm..."
 
+  OLD_HEAD="$(git rev-parse HEAD)"
+  if [[ -n ${CHARTS_VERSION:-} ]]; then
+    git checkout "$CHARTS_VERSION"
+  fi
+
   # --force is needed in case the first attempt at installing didn't succeed
   # see https://github.com/helm/helm/pull/3597
   retry 3 helm upgrade --install --force --wait --timeout 300 \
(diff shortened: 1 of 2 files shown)


api/hack/ci/ci-setup-kubermatic-in-kind.sh
```diff
@@ -331,37 +331,43 @@ if [[ "${KUBERMATIC_USE_OPERATOR}" = "false" ]]; then
...
-    --namespace=kubermatic \
-    ${ADDITIONAL_HELM_ARGS} \
-    ${OPENSHIFT_HELM_ARGS:-} \
-    --values ${VALUES_FILE} \
-    kubermatic \
-    ./config/kubermatic/
+  (
+    if [[ -n ${CHARTS_VERSION:-} ]]; then
+      git checkout "$CHARTS_VERSION"
+    fi
```
xrstf: I like the approach, but I don't like that after the "if use-operator or else" shebang we are now left with different git checkouts. If Helm is used, after the if-else-end thingy, we're on CHARTS_VERSION, otherwise we're still on HEAD. Can we make it consistent?

kdomanski: We're not left with different checkouts, because putting it in `( )` cause the thing to be run in a subshell and therefore return the previous workdir.

xrstf: How does a subshell revert changes to the filesystem?

```
/github.com/kubermatic/infra/apps/node-reaper $ git rev-parse HEAD && \
   (git co master) && \
   git rev-parse HEAD
64bdd01f643e8a3d8159a96945955bf7aebbf99c
Switched to branch 'master'
Your branch is ahead of 'origin/master' by 138 commits.
  (use "git push" to publish your local commits)
2d5f758bfd1ca69257304ad6cbb30009369e2367
```

kdomanski: ooops, brainfart

kdomanski: Added a checkout that returns repo to previous state.

Issue #4728, opened 2019-11-22
Kubermatic Operator: Use on dev/cloud


xrstf:
**User Story**
This is the final step of the migration from the Helm chart to the new operator. This step involves setting up the operator on the two master clusters, adjusting the postsubmit deploy jobs to just patch the KubermaticConfiguration to the newer version, and removing the Kubermatic chart from master and seed clusters.

This is part of #4723.

**Acceptance criteria**
* [ ] The Kubermatic chart is not installed on any master/seed cluster for dev/cloud.
* [ ] A single Kubermatic Operator runs in each master cluster.

PR #4962, opened 2020-01-13
Allow using the operator instead of Helm for deploying Kubermatic

xrstf:
Until now we had the operator installed in parallel to the Helm chart, with the Helm chart being the main attraction. This PR changes the behaviour to use the operator **instead of** Helm when `USE_KUBERMATIC_OPERATOR=true` is set. The flag is not yet switched because all hell would break loose when the installation is not properly, manually prepared.

Relates to #4728

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-deploy-dev.sh
+++ b/api/hack/ci/ci-deploy-dev.sh
@@ -19,15 +19,16 @@ retry 5 vault write \
 export VAULT_TOKEN="$(cat /tmp/vault-token-response.json| jq .auth.client_token -r)"
 export KUBECONFIG=/tmp/kubeconfig
 export VALUES_FILE=/tmp/values.yaml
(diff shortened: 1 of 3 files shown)


api/hack/ci/ci-deploy-dev.sh
```diff
@@ -19,15 +19,17 @@ retry 5 vault write \
 export VAULT_TOKEN="$(cat /tmp/vault-token-response.json| jq .auth.client_token -r)"
 export KUBECONFIG=/tmp/kubeconfig
 export VALUES_FILE=/tmp/values.yaml
+export DOCKER_CONFIG=/tmp/dockercfg
 export KUBERMATIC_CONFIG=/tmp/kubermatic.yaml
+# export USE_KUBERMATIC_OPERATOR=true
```
kdomanski: please remove the commented-out code

PR #4963, opened 2020-01-14
Add utility to convert Helm values.yaml to KubermaticConfiguration

xrstf:
We shuffled quite a few settings around when designing the KubermaticConfiguration CRD. To make it easier for customers to migrate to the operator, we should also offer a tool to convert their existing configuration. This PR adds such a utility.

The code has been put into a dedicated binary so that we can ship it individually and in order to not make the existing operator binary more complicated by introducing a CLI framework.

The conversion makes great effort to not output default values, in order to create a minimal KubermaticConfiguration. This helps as it does not hardcode (possibly soon outdated) default values and prevents misconfiguration. It also makes it easier to either use the customer's values.yaml (with only their adjustments) or the merged Helm values (e.g. by taking the values from Tiller's ConfigMap for the Kubermatic release), as default values are consistently filtered out.

Part of #4723 

```release-note
NONE
```

### DIFF
--- a/api/cmd/kubermatic-operator-util/main.go
+++ b/api/cmd/kubermatic-operator-util/main.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/urfave/cli"
+	"go.uber.org/zap"
+	"sigs.k8s.io/yaml"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/operator/conversion"
+	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
+	"github.com/kubermatic/kubermatic/api/pkg/log"
+)
+
+var logger *zap.SugaredLogger
+
+func main() {
+	app := cli.NewApp()
+	app.Name = "Kubermatic Operator Utility"
+	app.Version = "v1.0.0"
+
+	defaultLogFormat := log.FormatConsole
+
+	app.Flags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "log-debug",
+			Usage: "Enables more verbose logging",
+		},
+		cli.GenericFlag{
+			Name:  "log-format",
+			Value: &defaultLogFormat,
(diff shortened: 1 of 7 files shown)


api/cmd/kubermatic-operator-util/main.go
```diff
@@ -0,0 +1,88 @@
...
+		return cli.NewExitError("no values.yaml file given", 2)
+	}
+
+	config, err := conversion.HelmValuesFileToKubermaticConfiguration(valuesFile)
+	if err != nil {
+		return cli.NewExitError(err, 1)
+	}
+
+	config.APIVersion = operatorv1alpha1.SchemeGroupVersion.String()
+	config.Kind = "KubermaticConfiguration"
```
xrstf, line 91: Very open for ways to improve this.

xrstf:
/assign @moelsayed 

Congratulations, you have been volunteered for this PR ;-)

api/pkg/controller/operator/conversion/helm.go
```diff
@@ -0,0 +1,454 @@
...
+
+	storeContainer := strings.TrimSpace(values.StoreContainer)
+	if storeContainer == strings.TrimSpace(common.DefaultBackupStoreContainer) {
+		storeContainer = ""
+	}
+
+	cleanupContainer := strings.TrimSpace(values.CleanupContainer)
+	if cleanupContainer == strings.TrimSpace(common.DefaultBackupCleanupContainer) {
+		cleanupContainer = ""
+	}
```
moelsayed, line 245: Why are we clearing those if they have the defaults ?

xrstf: Because a customer might use their effective values.yaml (the one Helm creates when merging the Chart's values.yaml with the one the customer provides) and in this case, we would read the default backup/store containers. And then put these default values into the KubermaticConfiguration, where they now will stay forever. And when we update the default to fix something, customers will not get the fix because the "overrode" the containers.

That's why the converter never puts default values in the result.

xrstf:
/hold

Need to remove the binary I accidentially committed.

PR #4964, opened 2020-01-15
Migrate datacenters/presets when converting Helm values to Operator config

xrstf:
This PR extends the `kubermatic-operator-util` to not only output a KubermaticConfiguration, but also Seeds when the Helm values.yaml still included the old datacenters.yaml. The conversion code existed already, but now we offer to skip the migration in the master-controller-manager and perform it while the customer migrates from Helm to operator.

This PR also converts the old presets.yaml into a Preset CRD when possible.

Part of #4723 

```release-note
NONE
```


api/pkg/crd/migrations/util/kubeconfig.go
```diff
@@ -0,0 +1,64 @@
...
+	encoded, err := clientcmd.Write(*kubeconfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to serialize kubeconfig: %v", err)
+	}
+
+	secret := &corev1.Secret{}
+	secret.Name = name
+	secret.Namespace = namespace
+	secret.Data = map[string][]byte{
+		KubeconfigFieldPath: encoded,
```
kron4eg, line 60: isn't `encoded` should be base64 encoded?

kron4eg: it will became so automatically, by encoding/json :D

PR #4966, opened 2020-01-15
Add OpenVPN exporter

### LINKED ISSUE #1025: Instrument the OpenVPN in all cluster namespaces
https://github.com/kumina/openvpn_exporter

moelsayed:
Add OpenVPN exporter to all user clusters
Fixes #1025

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/resources/openvpn/deployment.go
+++ b/api/pkg/resources/openvpn/deployment.go
@@ -40,7 +40,7 @@ var (
 				corev1.ResourceCPU:    resource.MustParse("10m"),
 			},
 		},
-		"exporter": openvpnResourceRequirements.DeepCopy(),
+		"openvpn-exporter": openvpnResourceRequirements.DeepCopy(),
 	}
 )
 
@@ -259,8 +259,8 @@ done`,
 					},
 				},
 				{
-					Name:    "exporter",
-					Image:   data.ImageRegistry(resources.RegistryQuay) + "/kubermatic/openvpn-exporter:v0.2.2",
+					Name:    "openvpn-exporter",
+					Image:   data.ImageRegistry(resources.RegistryDocker) + "/kumina/openvpn-exporter:v0.2.2",
 					Command: []string{"/bin/openvpn_exporter"},
 					Args: []string{
 						"-openvpn.status_paths",
--- a/api/pkg/resources/openvpn/deployment.go
+++ b/api/pkg/resources/openvpn/deployment.go
@@ -3,6 +3,7 @@ package openvpn
 import (
 	"fmt"
 	"net"
+	"path"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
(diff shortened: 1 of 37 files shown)


api/pkg/resources/openvpn/deployment.go
```diff
@@ -244,6 +258,28 @@ done`,
 						ProcMount:  &procMountType,
 					},
 				},
+				{
+					Name: "exporter",
+					// FIXME: use proper image
```
xrstf: Where does this image come from and what did you patch to make it work?

xrstf (commented):
One question, otherwise very nice.

api/pkg/resources/openvpn/deployment.go
moelsayed: Source repo is https://github.com/kumina/openvpn_exporter (mentioned in the original issue). I just compiled, docker-built it and pushed it. No modifications.

xrstf: https://github.com/kumina/openvpn_exporter#docker mentions that `kumina/openvpn-exporter` exists already, so that's why I'm wondering.

moadqassem: @moelsayed  I would suggest that we use the original docker image from the author which is `kumina/openvpn-exporter` and then get rid of the 
```// FIXME: use proper image```

Comment (moelsayed) [reply api/pkg/resources/openvpn/deployment.go]: @moadqassem The fixme comment is removed in a later commit along with using an image hosted in kubermatic repo.

I opted for using our repo because this image is in the cluster deployment critical path. If, for any reason, this image is not accessible, the openvpn deployment would fail to start the pod and and the cluster will never be ready.

Comment (moadqassem) [reply api/pkg/resources/openvpn/deployment.go]: @moelsayed Sorry I haven't seen that you have actually changed the image to use `kubermatic` sorry my bad. One small question though other wise lgtm, have you patched your changes and saw that the metrics are being instrumented?

File: api/pkg/resources/openvpn/deployment.go [hunk 2]
```diff
@@ -244,6 +258,27 @@ done`,
 						ProcMount:  &procMountType,
 					},
 				},
+				{
+					Name:    "exporter",
```
moadqassem: I think this should be `openvpn-exporter` instead of `exporter`

api/pkg/resources/openvpn/deployment.go
xrstf: > I opted for using our repo because this image is in the cluster deployment critical path. If, for any reason, this image is not accessible, the openvpn deployment would fail to start the pod and and the cluster will never be ready.

What reasons would prevent pulling the original image, but not prevent pulling our fork?

moelsayed: Maintainer account gets compromised, they remove the image by mistake..etc..

xrstf: Can we just use the upstream image in this PR and then, if you really want to argue for mirrored Docker images, you can open another PR where you change *all* our Helm charts and start to mirror *all* of the Docker images? I don't think it's something we actually want and would rather discuss this separately from the "do we want an openvpn exporter?" question :-)

moelsayed: Makes sense, updated :+1:

PR #4969, opened 2020-01-16
Implement the vpa-updater controller

xmudrii:
Created a new PR in place of #4763 to prevent breaking Dashboard E2E tests.

This PR implements the `vpaupdater` controller responsible for reconciling the VPA resources and setting the resource requirements on Deployments and StatefulSets based on the provided recommendations.

The controller takes the following steps for each VPA resource:
1. Get the appropriate Deployment or StatefulSet
1. Check is there an `updated-by-vpa` label
    * If the label exists, parse it to get the last updated time. The controller upscales resources only once in 30 seconds and downscales only once in 5 minutes. This ensures that the pods will not be constantly rotated and that the VPA recommendations are stable.
1. Parse the VPA resource and update requirements for given Deployment and StatefulSet
    * Right now, we **only** set the memory requirements. The reason for this is that the CPU requirements change more often causing resources to be rotated more often. This can be problematic for components such as API server and etcd, as rotating them can cause other components (mainly controllers) to go down.
1. Update the `updated-by-vpa` label with new values
1. Trigger the update

The `updated-by-vpa` label contains JSON encoded struct containing the resource requirements set by the controller, as well as, the time when the last update happened. This label is used by other controllers that are responsible for Deployments/StatefulSets, so the new resource requirements are not overridden with default values by responsible controllers. See #4810 for more details.

Addresses part of #3213
It is up to be decided what we will do with the CPU requirements.

* This feature is up to be tested on the OpenShift clusters.
* It is up to be tested how the controller behaves when ComponentOverrides are being used.
* It is up to be tested how the controller behaves where there is a higher load on the components
* What is going to happen when VPA is not deployed? I assume that the controller would be enabled, but it will do nothing as there are no VPA resources to be reconciled. This should be tested somehow.
* The VPA resources are not created for the UserClusterController and Kubernetes Dashboard.

**Documentation**:
We should eventually document this feature, but I'm not sure where that should be done.

```release-note
Implement the vpaupdater controller. The controller is responsible for reconciling the VPA resources and applying the recommendations on Deployments and StatefulSets.
```

### DIFF
--- a/api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
+++ b/api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
@@ -8,6 +8,7 @@ import (
 
 	"go.uber.org/zap"
 
+	"github.com/kubermatic/kubermatic/api/pkg/controller/util"
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	kubermaticv1helper "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1/helper"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
@@ -26,8 +27,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-
-	"github.com/kubermatic/kubermatic/api/pkg/controller/util"
 )
 
 const (
--- a/api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
+++ b/api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
@@ -1,0 +1,311 @@
+package vpaupdater
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"go.uber.org/zap"
+
+	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
+	kubermaticv1helper "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1/helper"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	autoscalingv1beta2 "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1beta2"
+	"k8s.io/client-go/tools/record"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/util"
+)
+
+const (
+	ControllerName = "kubermatic_vpaupdater_controller"
+	UpdatedByVPA   = "updated-by-vpa"
+)
+
+// Reconciler stores necessary components that are required to manage in-cluster Add-On's
+type Reconciler struct {
+	log        *zap.SugaredLogger
+	workerName string
+	ctrlruntimeclient.Client
+	recorder record.EventRecorder
+}
+
+// Add creates a new Addon controller that is responsible for
+// managing in-cluster addons
+func Add(
+	mgr manager.Manager,
+	log *zap.SugaredLogger,
+	numWorkers int,
+	workerName string,
+) error {
+	log = log.Named(ControllerName)
+	client := mgr.GetClient()
+
+	reconciler := &Reconciler{
+		log:        log,
+		Client:     client,
+		workerName: workerName,
+		recorder:   mgr.GetEventRecorderFor(ControllerName),
+	}
+
+	ctrlOptions := controller.Options{
+		Reconciler:              reconciler,
+		MaxConcurrentReconciles: numWorkers,
+	}
+	c, err := controller.New(ControllerName, mgr, ctrlOptions)
+	if err != nil {
+		return err
+	}
+
+	objTypes := []runtime.Object{
+		&appsv1.Deployment{},
+		&appsv1.StatefulSet{},
+		&autoscalingv1beta2.VerticalPodAutoscaler{},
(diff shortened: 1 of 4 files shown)


api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,311 @@
...
+	"k8s.io/client-go/tools/record"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/util"
```
kron4eg: this import seems lost its ordering

kron4eg (commented):
looks good, microscopic nil in comments

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
xmudrii: Fixed.

irozzo-1A:
@xmudrii apparently this PR implements the [updater] (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/vertical-pod-autoscaler.md#updater) component described int the VPA design proposal, is it correct? Did you take a look at https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler/pkg/updater ?

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+	updated := false
+	for _, cr := range containerRecommendations {
+		for i, c := range podSpec.Containers {
+			if c.Name == cr.ContainerName {
+				requirements := corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceMemory: *cr.Target.Memory(),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceMemory: *cr.UpperBound.Memory(),
```
irozzo-1A, line 258: Not sure it is correct to do so, according to the [recomendation mode spec] (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/vertical-pod-autoscaler.md#recommendation-model) in this early stage VPA just focuses on requests and not limits.

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	return nil
+}
+
+func updatePodSpec(podSpec *corev1.PodSpec, containerRecommendations []autoscalingv1beta2.RecommendedContainerResources, lastUpdated *time.Time) ([]resources.Requirements, bool) {
+	podsRequirements := []resources.Requirements{}
+	updated := false
+	for _, cr := range containerRecommendations {
+		for i, c := range podSpec.Containers {
```
irozzo-1A, line 251: Minor, if you create a map of container name -> container you avoid some iterations. In case the number of recommendations is big it could be worthy.

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+
+				scaleUp := false
+				scaleDown := false
+
+				var valOld float64
+				if podSpec.Containers[i].Resources.Requests.Memory() != nil {
+					valOld = float64(podSpec.Containers[i].Resources.Requests.Memory().MilliValue())
+				}
+				valNew := float64(requirements.Requests.Memory().MilliValue())
+				if valNew >= valOld*1.15 {
```
irozzo-1A, line 276: Magic number `1.15`, I think it's better to define a configurable threshold e.g. 10%

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+					scaleUp = true
+				} else if valNew <= valOld*0.7 {
```
irozzo-1A, line 278: Same here.
BTW why it is not symmetric?

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+					scaleDown = true
+				}
+
+				valOld = 0
+				if podSpec.Containers[i].Resources.Limits.Memory() != nil {
+					valOld = float64(podSpec.Containers[i].Resources.Limits.Memory().MilliValue())
+				}
+				valNew = float64(requirements.Limits.Memory().MilliValue())
```
irozzo-1A, line 286: Same comment as before are limits in the scope of VPA.

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+					scaleDown = true
+				}
+
+				if lastUpdated == nil {
+					podSpec.Containers[i].Resources = requirements
+					updated = true
+					continue
+				}
+
+				if scaleUp && time.Since(*lastUpdated).Seconds() >= 30 {
```
irozzo-1A, line 299: This threshold also should be configurable

api/pkg/controller/seed-controller-manager/vpaupdater/vpaupdater.go
```diff
@@ -0,0 +1,310 @@
...
+					podSpec.Containers[i].Resources = requirements
+					updated = true
+				} else if scaleDown && time.Since(*lastUpdated).Minutes() >= 5 {
```
irozzo-1A, line 302: Same here

irozzo-1A (commented):
I could think it would be nice and more relevant to consider the logic used by the updater MVP to take decision about updating or not deployments: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/pkg/updater/priority/update_priority_calculator.go#L118
The problem is that this logic acts at POD level, but we could apply it to all or a sample of pods controlled by deployments and stateful set we control.

Note that this way of working of the updater has been considered in the project upstream and the solution is briefly discussed [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/autoscaling/vertical-pod-autoscaler.md#actuation-using-the-deployment-update-mechanism). I think the points regarding possible conflicts with user updates and fallbacks are relevant. What do you think?

kdomanski:
Can we get a rebase and continue the work?

xmudrii:
@kdomanski Let's sync about this tomorrow

kdomanski:
@xmudrii please rebase

xmudrii:
This PR is quite old and a lot of things changed since then. Additionally, this approach doesn't work well because:

- it turns out that it's much more complicated to predict the resource usage and thresholds
- restarting many components at the same time would make resource predictions wrong, which would cause even more restarting, which eventually causes the cluster to fail

Instead, we should reinvestigate this and see can we use the upstream updater instead of implementing our own.

/close

PR #4970, opened 2020-01-16
Do not tag :latest when doing canary deployments in presubmits

xrstf:
Some PRs cause a canary deployment and we accidentally always tag a "latest" Docker image, probably because of the assumption that the `ci-push-images.sh` is only ever used in postsubmits. This means that the "latest" tag is more or less random and this causes all sorts of random test failures in the dashboard repo (and of course can cause major issues if a customer ever wanted to live dangerously and run "latest").

This PR simply disable the "latest" tag if the script is called in a canary presubmit.

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-push-images.sh
+++ b/api/hack/ci/ci-push-images.sh
@@ -7,8 +7,13 @@ cd "$(git rev-parse --show-toplevel)"
 
 GIT_HEAD_HASH="$(git rev-parse HEAD)"
 GIT_HEAD_TAG="$(git tag -l --points-at HEAD)"
-# FIXME: use `latest` only on the master branch
-TAGS="$GIT_HEAD_HASH $GIT_HEAD_TAG latest"
+TAGS="$GIT_HEAD_HASH $GIT_HEAD_TAG"
+
+# we only want to create the "latest" tag if we're building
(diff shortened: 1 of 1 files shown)


api/hack/ci/ci-push-images.sh
```diff
@@ -7,8 +7,13 @@ cd "$(git rev-parse --show-toplevel)"
 
 GIT_HEAD_HASH="$(git rev-parse HEAD)"
 GIT_HEAD_TAG="$(git tag -l --points-at HEAD)"
-# FIXME: use `latest` only on the master branch
-TAGS="$GIT_HEAD_HASH $GIT_HEAD_TAG latest"
+TAGS="$GIT_HEAD_HASH $GIT_HEAD_TAG"
+
+# during canary deployments (in presubmits) we do not want to tag
+# a "latest" Docker image, but in postsubmits we do
+if [ -z "${CANARY_DEPLOYMENT:-}" ]; then
```
floreks: Could we rather check if current branch is master and only then add latest? This seems to cover only this single faulty case.

maciaszczykm:
@xrstf Can we move it to the past releases as it is breaking our e2e?

PR #4972, opened 2020-01-17
Configurable update window

bashofmann:
This adds a configurable update window with a start time, (e.g. "02:00", or :"Thu 02:00") and length (e.g. "1h") to a cluster.

It can be used to limit various update operations to a given time frame only.

The first usage is to configure the CoreOS operator node reboot window, see https://github.com/coreos/container-linux-update-operator/blob/master/doc/reboot-windows.md.

In chose to make it a generic setting since we are planning to use this in the future for other things, like automatic cluster upgrades or in the addon controller.

```release-note
Add a configurable update window to a cluster
Use configurable update window for coreos-operator node reboots
```


bashofmann:
We will also provide a PR for the dashboard once this is merged, in case we want to change something like field names or other details during the review.

kdomanski:
Hi @bashofmann . This PR got lost in the cracks, because no reviewer was assigned to it. Would you care to rebase it against the current version?

kdomanski:
Closed, since prow was stuck on the test jobs. I'll try to reopen/recreate.

PR #4991, opened 2020-01-29
filter presets by datacenter

### LINKED ISSUE #4937: presets can be used in DCs/regions for which they don't work
**User Story**
Sales ran into an issue, where they attempted to use an AWS preset in a region, but the preset specified VPC/subnet that were only present in another reason. Please design a clean way to make this impossible and update this ticket.

zreigz:
**What this PR does / why we need it**: filter out presets by datacenter

Fixes #4937

```release-note
filter out presets by datacenter
```


zreigz:
PTAL  @moadqassem

PR #4994, opened 2020-01-29
Add support for custom Dex CA in kubermatic-api

xrstf:
When Dex uses a custom CA, we already support telling the kubernetes-apiserver in user clusters about that. This PR makes it so that the kubermatic-api itself can also use the custom CA.

```release-note
Support custom CA for OpenID provider in Kubermatic API.
```


zreigz:
the failing e2e test is flaky, I will fix this today

PR #5011, opened 2020-01-30
Fix FluentbitManyRetries error being too sensitive

xrstf:
Retries are a normal part of fluent-bit's scheduler operations and do not necessarily indicate a problem worth alerting about. Even on nodes that are pretty much idle fluent-bit seems to retry stuff internally without reason. The debug logging shows clearly that all Elasticsearch requests succeed and nothing on any level except debug is logged at all.

```release-note
Fix FluentbitManyRetries Prometheus alert being too sensitive to harmless backpressure.
```


kron4eg (approved):
/lgtm 
/approve

PR #5035, opened 2020-02-04
Disable master alerts for non-master clusters

### LINKED ISSUE #3873: Promtheus charts: Disable master alerts if `values.kubermatic.ismaster` is defined and has a value of false
**User Story**

As a kubermatic operator I do not want to have master alerts in my seeds, as they will permanently go off.

**Acceptance criteria**

* If the key `.Values.Kubermatic.IsMaster` is defined and has a value of `false`, our prometheus chart doesn't deploy any alerts for Kubermatic master components anymore

moelsayed:
Disable master alerts for non-master clusters.
Fixes #3873

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/hack/deploy.sh
+++ b/api/hack/deploy.sh
@@ -16,9 +16,9 @@ fi
 HELM_EXTRA_ARGS=${@:3}
 VALUES_FILE=$(realpath ${2})
 if [[ "${1}" = "master" ]]; then
-  MASTER_FLAG="--set=kubermatic.isMaster=true"
+  MASTER_FLAG="--set=kubermatic.isMaster=true --set=prometheus.isMaster=true"
 else
-  MASTER_FLAG="--set=kubermatic.isMaster=false"
+  MASTER_FLAG="--set=kubermatic.isMaster=false --set=prometheus.isMaster=false"
 fi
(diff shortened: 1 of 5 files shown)


xrstf:
I'm not really sure that trading the "change the list of ruleFiles" step when setting up a non-master cluster for "change this isMaster flag" is worth it, tbh. While it is definitly easier to just flip a single switch (especially because you can do that easily via `--set ....`), it hardcodes another layer of abstraction into the Prometheus chart. Prometheus so far did not really care about master vs. seeds, it only cared for a list of rules files to load (that we have one for masters and one for seeds is just coincidence).

moelsayed:
@xrstf TBH, I am not very happy with the solution either. The prometheus values file is well commented and it's possible to achieve this by commenting a single line. But, I wasn't sure if there is a specific requirement to do this with a switch.

If there is no explicit requirement for this, I am perfectly fine with just adding a comment about this in the `ruleFiles` section and avoid the extra complexity.

xrstf:
I would do exactly that, just add an even more expressive comment than I left originally and wait for someone to complain that managing the rule files is too hard. :)

PR #5040, opened 2020-02-04
Add SLES as an Operating System option

### LINKED ISSUE #3878: Add SLES as an OS option

kgroschoff:
Add SLES as an Operating System option

Fixes #3878

/

```release-note
Add SLES as an Operating System option
```


api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -60,7 +60,7 @@ func main() {
 
 func availableUsersPaths() ([]string, error) {
 	var paths []string
-	for _, user := range []string{"root", "core", "ubuntu", "centos"} {
+	for _, user := range []string{"root", "core", "ubuntu", "centos", "sles"} {
```
kgroschoff: Not sure: is `sles` a valid `userPath`?

api/pkg/machine/util.go
```diff
@@ -22,6 +22,7 @@ var userNameMap = map[string]string{
 	"AWS:Ubuntu":                  "ubuntu",
 	"AWS:ContainerLinux":          "core",
 	"AWS:CentOS":                  "centos",
+	"AWS:SLES":                    "",
```
kgroschoff: I have no idea what the right value is. 🤔

zreigz: cc @moadqassem

kgroschoff: done :)

api/cmd/user-ssh-keys-agent/main.go
moadqassem: The problem here is, in the `aws` cloud provider, there will be one home dir for `sles`, which is `ec2-user`. Now, this is problematic since this functionality doesn't judge the cloud provider, but it checks for the os distro. This means if we add `ec2-user` here it will work for `aws` provider but it will not work for other providers since the `ec2-user` is exclusive to `aws`.

kgroschoff: I see. :/ Any idea how we should handle it?

moadqassem:
@kgroschoff 
cc @zreigz 

At the moment this PR looks good except a small detail that would block the functionality of updating the usersshkeys inside the provisioned vm. The reason for that is, in aws there is a user which is called `ec2-user` that exists by default for the `sles`image, and adding this user home won't be convenient because this is a cloud provider detail not an os provider, thus adding those user home dirs isn't really idiomatic. and even if we do it this way, the problem will arise again once we use other os. e.g: RHEL. 

In the past we had only ubuntu, centos and core so we didn't need a lot of customization to solve this issue, however now we need a different approach to solve this issue and we should make dynamic instead of static. 

One approach can be to list all the dirs which are in the `/home` dir and which dir exists in there we simply copy those keys in.

zreigz:
In my opinion, we can merge this. We can think to block updating user ssh keys in the next PR or in the UI. We have to rethink this.

zreigz (approved):

api/cmd/user-ssh-keys-agent/main.go
moadqassem: Well for now, please adjust the ‘sles’ path to ‘ec2-user’ because there is no home dir called sles ;)

moadqassem:
Alright then, after changing this https://github.com/kubermatic/kubermatic/pull/5040#discussion_r375088471 from ‘sles’ to ‘ec2-user’, things will work as expected and I can add a dynamic path configuring in later step

api/cmd/user-ssh-keys-agent/main.go
kgroschoff: sure, done :)

zreigz:
@moadqassem you have to approve it or someone from lifecycle team

PR #5041, opened 2020-02-04
prow.yaml: configure canary deployment to only ever run against master

kdomanski:
```release-note
NONE
```


xrstf:
/lgtm

As a reminder: We do this because in https://github.com/kubermatic/infra/pull/810 we accidentally switched the job from "master-only" to "everything but master/weekly", which is just an oversight in the mass-edit / done because the job was already moved to the .prow.yaml and having a duplicate in the infra repo would be no good.

PR #5044, opened 2020-02-05
add admission controller plugins

zreigz:
**What this PR does / why we need it**: add admission controller plugins. It will allow adding some extra plugins to the user cluster API. The plugins will be defined in CRD what makes this solution configurable. Right now we have only two checkboxes for it. We can replace it with something like this:
react-multiselect-dropdown

Fixes #

```release-note
make possible to extend admission controller plugins for user cluster API using cluster structure
```

### DIFF
--- a/api/pkg/resources/apiserver/deployment.go
+++ b/api/pkg/resources/apiserver/deployment.go
@@ -228,7 +228,7 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		nodePortRange = defaultNodePortRange
 	}
 
-	admissionPlugins := []string{
+	admissionPlugins := sets.NewString(
 		"NamespaceLifecycle",
 		"LimitRanger",
 		"ServiceAccount",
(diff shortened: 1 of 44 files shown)


api/pkg/resources/apiserver/deployment.go
```diff
@@ -244,6 +245,14 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
...
 		admissionPlugins = append(admissionPlugins, "PodNodeSelector")
 	}
+	if data.Cluster().Spec.AdmissionPlugins != nil {
+		admissionPluginSet := sets.NewString(admissionPlugins...)
+		for _, plugin := range data.Cluster().Spec.AdmissionPlugins {
+			if !admissionPluginSet.Has(plugin) {
+				admissionPlugins = append(admissionPlugins, plugin)
+			}
+		}
+	}
```
xrstf: How about

```go
admissionPlugins := sets.NewString(
   "NamespaceLifecycle",
   "LimitRanger",
   "ServiceAccount",
   "DefaultStorageClass",
   "DefaultTolerationSeconds",
   "MutatingAdmissionWebhook",
   "ValidatingAdmissionWebhook",
   "Priority",
   "ResourceQuota",
)

admissionPlugins.Insert(data.Cluster().Spec.AdmissionPlugins)

if data.Cluster().Spec.UsePodNodeSelectorAdmissionPlugin {
   admissionPlugins.Insert("PodNodeSelector")
}

// yadda yadda yadda

"--enable-admission-plugins", strings.Join(admissionPlugins.List(), ",")
```

xrstf (commented):
How 'bout this?

kron4eg:
Just FYI, in different kubernetes versions the "enabled by default list" of admission plugins is different, please keep this in mind.

xrstf:
That is actually something we should maybe really check. Good thinking!

kron4eg:
We already faced that problem in kubeone, and hacked though it. Not good, not terrible. But it works for now.

zreigz:
> Just FYI, in different kubernetes versions the "enabled by default list" of admission plugins is different, please keep this in mind.

We going to refactor the way of setting the plugins. Then I will add validation for that

zreigz:
I have to update fixtures. The plugins are sorted right now

kron4eg (approved):
/lgtm

PR #5051, opened 2020-02-07
add an endpoint to list admission plugins

zreigz:
**What this PR does / why we need it**: add an endpoint to list admission plugins

Fixes #

```release-note
add an endpoint to list admission plugins: GET /api/v1/admission/plugins/{version}
```


api/pkg/handler/routes_v1.go
```diff
@@ -559,6 +560,12 @@ func (r Routing) RegisterV1(mux *mux.Router, metrics common.ServerMetrics) {
 	mux.Methods(http.MethodGet).
 		Path("/labels/system").
 		Handler(r.listSystemLabels())
+
+	//
```
floreks, line 564: Leftover?

zreigz: it's everywhere like this

floreks: That's a bit weird but ok.

PR #5052, opened 2020-02-07
Add Openshift support in image-loader

moelsayed:
**What this PR does / why we need it**:
Adds OpenShift support in the image-loader
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3623

**Special notes for your reviewer**:
Depends on #5055
**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5055, opened 2020-02-10
Add overwriteRegistry support for Openshift

### LINKED ISSUE #5059: Add overwriteRegistry support for Openshift resource deployment
**User Story**
Currently, the Openshift resource deployment code doesn't respect the overwriteRegistry passed to the seed controller. It should be supported.
**Acceptance criteria**
- [ ] When `overwriteRegistry` is used, all Openshift related images should be pulled from that registry.

moelsayed:
This PR enables Openshift resources/versions generator to use an external registry. 
Fixes #5059

Since the Openshift images use a single base image with different digests, each pointing to a layer representing the actual image used. It's not possible to "re-tag" image digests, so, as a work around, the added helper function will use the ~~digest hex as a tag~~ component name and the Openshift version as a tag. PR #5052 will use the same convention to push the images to a private registry. 
**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/seed-controller-manager/openshift/resources/utils.go
+++ b/api/pkg/controller/seed-controller-manager/openshift/resources/utils.go
@@ -8,7 +8,7 @@ import (
 
 // OpenshiftImageWithRegistry will return docker image name for Openshift images. The function is
 // digest-aware and can be used with the overwriteRegistry option and with image-loader.
-func OpenshiftImageWithRegistry(image, registry string) (string, error) {
+func OpenshiftImageWithRegistry(image, componentName, version, registry string) (string, error) {
 	if registry == "" {
 		return image, nil
 	}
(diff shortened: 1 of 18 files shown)


irozzo-1A:
Personally I'm not fond of this solution as we discussed offline. Using the image digest as a tag is not very clean and prevents us from using the official ocp mirroring procedure https://docs.openshift.com/container-platform/4.2/installing/install_config/installing-restricted-networks-preparations.html#installation-mirror-repository_installing-restricted-networks-preparations . Even if it costs more I think it's worthy to look at the OCP 4 installer and take inspiration to solve this particular issue.

moelsayed:
@irozzo-1A I opened issue #5059 for this, I think we should move the discussion there.

xrstf:
/approve
/assign @irozzo-1A 

I approve, but cannot really judge the implementation. Seems like @irozzo-1A knows much more about this than me :-)

api/pkg/controller/seed-controller-manager/openshift/resources/utils.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	if taggedImageRef, ok := imageRef.(reference.NamedTagged); ok {
+		return fmt.Sprintf("%s/%s:%s", registry, reference.Path(imageRef), taggedImageRef.Tag()), nil
+	} else if _, ok := imageRef.(reference.Digested); ok {
+		// if the image is passed with digest, we use the component name and version to
+		// tag the image
+		if componentName == "" || version == "" {
+			return "", fmt.Errorf("failed to set Openshift image tag. Component name and Openshift version must be set")
+		}
+		return fmt.Sprintf("%s/%s:%s-%s", registry, reference.Path(imageRef), version, componentName), nil
```
irozzo-1A, line 30: When we mirror the openshift images to a V2 registry the digest remain the same. Wouldn't be better to keep retrieving the images by digest in such case? It would require some logic to detect if the registry support V2 though. I don't consider this a blocker, we can improve later.

irozzo-1A: As we discussed offline we can think about this and add it as an improvement for later.

irozzo-1A (approved):

PR #5057, opened 2020-02-10
Update Thanos to 0.10.1

xrstf:
Thanos 0.10 promises lower resource requirements and other nicities. The UI component now also beings to emitting metrics, so the bad scrape annotations from before now actually work.

```release-note
Update Thanos to 0.10.1
```

### DIFF
--- a/config/monitoring/prometheus/values.yaml
+++ b/config/monitoring/prometheus/values.yaml
@@ -121,7 +121,7 @@ prometheus:
     enabled: false
     image:
       repository: quay.io/thanos/thanos
-      tag: v0.8.1
+      tag: v0.10.1
 
     store:
       replicas: 2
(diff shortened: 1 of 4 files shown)


config/monitoring/prometheus/values.yaml
```diff
@@ -245,7 +245,13 @@ prometheus:
...
           memory: 1Gi
     thanosStore:
-      resources: {}
+      resources:
+        requests:
+          cpu: 250m
+          memory: 1536Mi
+        limits:
+          cpu: 1
+          memory: 3Gi
```
xrstf, line 254: We did not have resource constraints for the store component because we wanted to learn more about it over some time. That time has passed and these constraints should work fine.

PR #5066, opened 2020-02-10
Add VPA Support for Kubermatic Operator

xrstf:
This has been a missing feature so far. This mostly just copies what the Kubermatic Helm chart does, but cleans up the naming a tiny bit here and there.

Relates to #4723 

```release-note
NONE
```


api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/operator/common"
+	operatorv1alpha1 "github.com/kubermatic/kubermatic/api/pkg/crd/operator/v1alpha1"
+	"github.com/kubermatic/kubermatic/api/pkg/resources"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/certificates/triple"
+	"github.com/kubermatic/kubermatic/api/pkg/resources/reconciling"
+	"k8s.io/apimachinery/pkg/util/intstr"
```
moadqassem: Please regroup the imports

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+	AdmissionControllerServiceCertSecretName = "vpa-tls-certs"
+	AdmissionControllerServingCertName       = "serverCert.pem"
+	AdmissionControllerServingCertKeyName    = "serverKey.pem"
+	AdmissionControllerServingCertCAName     = "caCert.pem"
+
+	// WebhookServiceName is defined by the admission-controller's self-registration.
+	WebhookServiceName = "vpa-webhook"
+)
+
+func AdmissionControllerServiceAccountCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceAccountCreatorGetter {
```
moadqassem: The `cfg` parameter is not used

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+					Resources: cfg.Spec.VerticalPodAutoscaler.AdmissionController.Resources,
+				},
+			}
+
+			return d, nil
+		}
+	}
+}
+
+func AdmissionControllerServiceCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceCreatorGetter {
```
moadqassem: This function is not used and the parameter `cfg` isn't used either.

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+
+			s.Spec.Selector = appPodLabels(AdmissionControllerName)
+
+			return s, nil
+		}
+	}
+}
+
+func AdmissionControllerServingCertCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedSecretCreatorGetter {
```
moadqassem: Same goes here, the parameter `cfg` isn't used.

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+							SecretName: AdmissionControllerServiceCertSecretName,
+						},
+					},
+				},
+			}
+
+			d.Spec.Template.Spec.ServiceAccountName = AdmissionControllerName
+			d.Spec.Template.Spec.Containers = []corev1.Container{
+				{
+					Name:    "admission-controller",
```
moadqassem, line 73: Why not using the constant which represents the controller name, `AdmissionControllerName`?  Is it because it's called after the binary?

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+					Image:   cfg.Spec.VerticalPodAutoscaler.AdmissionController.DockerRepository + ":" + versions.VPA,
+					Command: []string{"/admission-controller"},
+					Args: []string{
+						"--address=:8944",
```
moadqassem: The port `8944` has been spotted more than two times already, what about making it a constant?

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+
+			s.Spec.Selector = appPodLabels(AdmissionControllerName)
+
+			return s, nil
+		}
+	}
+}
+
+func AdmissionControllerServingCertCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedSecretCreatorGetter {
+	commonName := WebhookServiceName
```
moadqassem: Why declaring a new var here, while you can use the constant directly?

api/pkg/controller/operator/common/vpa/admissionctrl.go
```diff
@@ -0,0 +1,214 @@
...
+			se.Data[AdmissionControllerServingCertKeyName] = triple.EncodePrivateKeyPEM(key)
+			se.Data[AdmissionControllerServingCertName] = triple.EncodeCertPEM(cert)
+			se.Data[AdmissionControllerServingCertCAName] = triple.EncodeCertPEM(ca.Cert)
+
+			return se, nil
+		}
+	}
+}
+
+func hasValidCertificate(secret *corev1.Secret, commonName string, altNames certutil.AltNames) bool {
```
moadqassem, line 183: I would suggest that this method returns also an error so we can know why certificates validation has failed. Returning only false here gives us no hints about what went wrong!

api/pkg/controller/operator/common/vpa/rbac.go
```diff
@@ -0,0 +1,216 @@
...
+			return cr, nil
+		}
+	}
+}
+
+func ClusterRoleBindingCreators(cfg *operatorv1alpha1.KubermaticConfiguration) []reconciling.NamedClusterRoleBindingCreatorGetter {
+	recommender := rbacv1.Subject{
+		Kind:      "ServiceAccount",
+		Name:      RecommenderName,
+		Namespace: "kube-system",
```
moadqassem: Nit: using maybe using `k8s.io/apimachinery/pkg/apis/meta/v1.NamespaceSystem` for all `kube-system` namespace here and down?

api/pkg/controller/operator/common/vpa/rbac.go
```diff
@@ -0,0 +1,216 @@
...
+			{
+				APIGroups: []string{"autoscaling.k8s.io"},
+				Resources: []string{"verticalpodautoscalers"},
+				Verbs:     []string{"get", "list", "watch"},
+			},
+		}),
+	}
+}
+
+func clusterRoleCreator(cfg *operatorv1alpha1.KubermaticConfiguration, name string, rules []rbacv1.PolicyRule) reconciling.NamedClusterRoleCreatorGetter {
```
moadqassem: Again `cfg` parameter is not used.

api/pkg/controller/operator/common/vpa/rbac.go
```diff
@@ -0,0 +1,216 @@
...
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "ClusterRole",
+			Name:     AdmissionControllerRoleName,
+		}, []rbacv1.Subject{
+			admissionController,
+		}),
+	}
+}
+
+func clusterRoleBindingCreator(cfg *operatorv1alpha1.KubermaticConfiguration, name string, roleRef rbacv1.RoleRef, subjects []rbacv1.Subject) reconciling.NamedClusterRoleBindingCreatorGetter {
```
moadqassem: `cfg` parameter is not used.

api/pkg/controller/operator/common/vpa/recommender.go
```diff
@@ -0,0 +1,82 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	RecommenderName = "vpa-recommender"
+)
+
+func RecommenderServiceAccountCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceAccountCreatorGetter {
```
moadqassem: Unused `cfg` parameter.

api/pkg/controller/operator/common/vpa/admissionctrl.go
xrstf: We do not care *why* a cert is not valid, we only want to reconcile here. What would you do with an error here? Log it? That would be the only place in all our controllers where we log a warning if the resource differs from the desired one. ;)

api/pkg/controller/operator/common/vpa/admissionctrl.go
xrstf: Yep, I don't like ultra complex container names just because someone was reusing a constant.

api/pkg/controller/operator/common/vpa/admissionctrl.go
moadqassem: Ok, of there was no valid certificate we simply create a new one. I think that makes sense :)

PR #5077, opened 2020-02-12
update Masterminds/semver to 3.0.3

xrstf:
This PR just updates one of our dependencies. Nothing fancy to see here.

```release-note
NONE
```


xrstf:
This just opens a can of worms we should not open until the box of pandora, i.e. go modules, are available for all our dependencies.

PR #5080, opened 2020-02-12
Generate Kubermatic Chart static files

xrstf:
Keeping our helper manifests up to date in two places is tedious and easy to fuck up. This PR makes it so that the Operator is the single source of truth and the manifests inside the Kubermatic Helm chart are merely derived from it.

```release-note
NONE
```


config/kubermatic/static/master/accessible-addons.yaml
```diff
@@ -0,0 +1,4 @@
+# This file has been generated using hack/update-kubermatic-chart.sh, do not edit.
+
+addons:
+- node-exporter
```
xrstf, line 4: This has to be a map at the top level because Helm's fromYaml expects it.

api/codegen/kubermatic_operator/main.go
```diff
@@ -0,0 +1,57 @@
...
+// based on the canonical source of truth, the Kubermatic Operator package.
+
+package main
+
+import (
+	"io/ioutil"
+	"log"
+	"strings"
+
+	"github.com/kubermatic/kubermatic/api/pkg/controller/operator/common"
```
moadqassem: Maybe a better import regrouping?

api/codegen/kubermatic_operator/main.go
```diff
@@ -0,0 +1,57 @@
...
+func main() {
+	writeYAML(common.DefaultBackupStoreContainer, "config/kubermatic/static/store-container.yaml")
+	writeYAML(common.DefaultBackupCleanupContainer, "config/kubermatic/static/cleanup-container.yaml")
+	writeYAML(common.DefaultVersionsYAML, "config/kubermatic/static/master/versions.yaml")
+	writeYAML(common.DefaultUpdatesYAML, "config/kubermatic/static/master/updates.yaml")
+	writeYAML(common.DefaultKubernetesAddons, "config/kubermatic/static/master/kubernetes-addons.yaml")
+	writeYAML(common.DefaultOpenshiftAddons, "config/kubermatic/static/master/openshift-addons.yaml")
+	writeJSON(common.DefaultUIConfig, "config/kubermatic/static/master/ui-config.json")
+
+	markup, err := yaml.Marshal(map[string]interface{}{
```
moadqassem, line 25: Why an interface? `[]string` wouldn't be enough?

api/codegen/kubermatic_operator/main.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Fatalf("Failed to encode accessible addons as YAML: %v", err)
+	}
+
+	writeYAML(string(markup), "config/kubermatic/static/master/accessible-addons.yaml")
+}
+
+func writeYAML(container string, filename string) {
+	log.Printf("Generating %s...", filename)
+
+	container = strings.Join([]string{
```
moadqassem, line 38: What about using a buffer instead of using string joins?  it is more efficient and easier to read to be honest :)

api/codegen/kubermatic_operator/main.go
```diff
@@ -0,0 +1,57 @@
+// This application generates some manifests in the Kubermatic Helm chart
```
moadqassem, line 1: Can we maybe have a test for this bin functionality?

api/codegen/kubermatic_operator/main.go
```diff
@@ -0,0 +1,57 @@
...
+// based on the canonical source of truth, the Kubermatic Operator package.
+
+package main
+
+import (
+	"io/ioutil"
+	"log"
```
moadqassem, line 8: Nit: What about using zap here since we already using in all kubermatic?

api/codegen/kubermatic_operator/main.go
moadqassem, line 1: Nit: In go package names shouldn't have underscores.

moadqassem (changes requested):

api/codegen/kubermatic_operator/main.go
xrstf: Because I need to output a map, see my other Github comment on the generated file.

api/codegen/kubermatic_operator/main.go
xrstf: What should that test, except for what the new verify-kubermatic-chart job already does?

api/codegen/kubermatic_operator/main.go
xrstf: Because it's just a codegen thingy and setting up zap requires more code than it's worth it, IMHO. This thing will never produce logs that are meant to be machine-readable anyway.

api/codegen/kubermatic_operator/main.go
xrstf: I was keeping the convention started with `openshift_versions` ;-)

moadqassem: How Is it a convention? I mean having underscores in package name is not idiomatic in golang please check this https://blog.golang.org/package-names.

PR #5088, opened 2020-02-13
replace employee email addresses in Dockerfiles

xrstf:
Employees come and go, and just like with our Helm charts we should use a generic maintainer label for our Docker images.

```release-note
NONE
```


moadqassem:
Really cool idea

PR #5091, opened 2020-02-14
Operator chart

xrstf:
Even though the operator takes care of setting up Kubermatic, we currently have a bunch of other dependencies for which we still use Helm. So the install procedure for customers would currently be

* install something with Helm
* install something with Helm
* install something with Helm
* do something completely different for the Kubermatic Operator

Which feels akward. This PR turns the static manifests for the operator into a very, very basic Helm chart.

Relates to #4723 

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve

PR #5093, opened 2020-02-14
WIP - Helm 3

xrstf:
This prepares all Docker images to contain Helm 3 in addition to Helm 2. It also configures the Prow jobs to use these new images and gives the deploy.sh the ability to use Helm 3.

This should prepare us for next steps in the Helm3 migration.

Requires kubermatic/infra#909.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -68,7 +68,7 @@ presubmits:
     clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
     spec:
       containers:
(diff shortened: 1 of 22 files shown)


xrstf:
This is nice, but there have been too many changes to resurrect this branch. Other PRs have to be made to make all Helm charts compatible with Helm 3. Some other older PRs have already introduced a Helm 3 binary to our images.

In short: We can do better than this, and when it's time, we will do so.

PR #5097, opened 2020-02-16
Rollback Karma to v0.52

youssefazrak:
Rollback previous upgrade for Karma v0.55 to Karma v0.52.
At this point it seems the issue isn't coming from the upgrade as it appeared couple of hours later. Moreover a downgrade didn't solve the issue of having a crashloopbackoff pod.
I will proceed with further testing then re-write a PR for the upgrade to latest Karma version

`time="2020-02-16T20:56:47Z" level=info msg="Reading configuration file /etc/karma/karma.yaml"
{"level":"fatal","msg":"No valid Alertmanager URIs defined","time":"2020-02-16T20:56:47Z"}`

```release-note
NONE
```

### DIFF
--- a/config/monitoring/karma/Chart.yaml
+++ b/config/monitoring/karma/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: karma
-version: 0.0.6
-appVersion: v0.55
+version: 0.0.5
+appVersion: v0.52
 description: Dashboard for Prometheus Alertmanager
 keywords:
 - kubermatic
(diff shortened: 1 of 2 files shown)


xrstf:
Karma is not broken because of its update, but because due the kubermatic operator, for which I cleaned up the `kubermatic` namespace. This brought to light that the seed-proxy controller is defunct and its Deployments are what makes Karma find the Alertmanagers (see its init container).

xrstf:
https://github.com/kubermatic/kubermatic/pull/5101 should fix the underlaying issue.

youssefazrak:
> Karma is not broken because of its update, but because due the kubermatic operator, for which I cleaned up the `kubermatic` namespace. This brought to light that the seed-proxy controller is defunct and its Deployments are what makes Karma find the Alertmanagers (see its init container).

Superb, thanks Christoph

PR #5101, opened 2020-02-17
Fix seed-proxy controller not being triggered

xrstf:
The controller only watched the resources it creates, but did not watch Seeds themselves, so it was never triggered in the first place. The only reason this ever worked is that I guess I left some Deployments are developing the controller originally and these re-triggered the controller since then. When moving to the kubermatic operator, I cleaned up everything in the `kubermatic` namespace, so the error surfaced.

```release-note
Fix seed-proxy controller not being triggered.
```


kron4eg:
nice catch!
/lgtm
/approve

PR #5103, opened 2020-02-17
Make nodeport-proxy compatible with the operator

xrstf:
When using the kubermatic operator, users are no longer using the old `kubermatic` Helm chart. This means that the nodeport-proxy would not be able to find the Image Pull Secret.

This PR makes it so that the operator Helm chart is properly recognized.

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve

PR #5109, opened 2020-02-19
[release/v2.13] add Kubernetes v1.15.10, v1.16.7, v1.17.3 (#5102)

kdomanski:
```release-note
Added Kubernetes v1.15.10, v1.16.7, v1.17.3
```

### DIFF
--- a/api/pkg/controller/operator/common/defaults.go
+++ b/api/pkg/controller/operator/common/defaults.go
@@ -406,6 +406,8 @@ versions:
   default: false
 - version: "v1.15.9"
   default: true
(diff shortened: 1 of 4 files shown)


api/pkg/controller/operator/common/defaults.go
```diff
@@ -406,6 +406,8 @@ versions:
   default: false
 - version: "v1.15.9"
   default: true
```
xrstf: Now we would have two default versions and IIRC the first one is chosen.

xrstf (commented):
One minor thing.

PR #5110, opened 2020-02-19
Use explicit versions/updates structs instead of YAML

xrstf:
This replaces the (what I consider legacy) YAML documents inside the KubermaticConfiguration CRD with dedicated Go structs. The reason for doing so is that it just feels weird to document

> to change the versions Kubermatic supports, change this YAML file. But not directly, inside this CRD. No, really, please don't change the master files Secret yourself.

Now that we generate all YAML files inside the Kubermatic chart anyway, the door is open to get fancier with the versions and updates. This PR therefore changes the structure inside the CRD to

```yaml
spec:
  versions:
    kubernetes:
      default: 1.5.2
      versions:
        - 1.5.1
        - 1.5.2
        - 1.6.3
        - 1.7.4
        - 1.7.5
        - 1.7.6
      updates:
        - from: 1.5.*
          to: 1.6.*
          automatic: false
        - from: 1.6.*
          to: 1.7.*
          automatic: false
    openshift:
      default: 4.1.18
      versions:
        - 4.1.9
        - 4.1.18
      updates: []
```

The structure wasn't changed just for the sake of changing things, but instead the goal was to make misconfigurations harder. We had multiple cases where we ourselves accidentally configured multiple default versions, for example.

Even though the structure has changed, the operator continues to create a compatible versions.yaml/updates.yaml, so for the containers nothing has changed. But now the YAML format is just an implementation detail and not our public API anymore.

With the last two master files gone, this PR also removes the `masterFiles` field entirely from the KubermaticConfiguration. From what I can see, it's a remnant from a time when lots and lots of k8s manifests were shipped this way, and slowly over time fewer files remained there.

Also, because we already name the secret `extra-files`, this PR aligns the naming inside the code as well. "master files" never meant "files only on master clusters", it meant "important files".

```release-note
NONE
```


docs/zz_generated.kubermaticConfiguration.yaml
```diff
@@ -1,504 +1,593 @@
...
-    # NodePortRange is the port range for customer clusters - this must match the NodePort
-    # range of the seed cluster.
-    nodePortRange: 30000-32767
-    # OverwriteRegistry specifies a custom Docker registry which will be used for all images
-    # used inside user clusters (user cluster control plane + addons). This also applies to
-    # the KubermaticDockerRepository and DNATControllerDockerRepository fields.
-    overwriteRegistry: ""
+    # API configures the frontend REST API used by the dashboard.
+    api:
+        # AccessibleAddons is a list of addons that should be enabled in the API.
```
xrstf: The indentation changed because my old `replace "    " with "  "` hack was messing with arrays, so I had to remove it. A PR that improves it is in review at https://github.com/kubernetes/test-infra/pull/16359 -- this also fixes that each and every update is fully documented.

api/pkg/controller/operator/seed/resources/kubermatic/seed-controller-manager.go
```diff
@@ -149,35 +149,34 @@ func SeedControllerManagerDeploymentCreator(workerName string, versions common.V
 
 			args = append(
 				args,
-				"-versions=/opt/master-files/versions.yaml",
-				"-updates=/opt/master-files/updates.yaml",
+				"-versions=/opt/extra-files/versions.yaml",
+				"-updates=/opt/extra-files/updates.yaml",
```
xrstf, line 153: And this here, the usage of "master-files" for the *seed* controller manager, was what triggered the `s/master-/extra-/g` change.

kron4eg:
OMG this is awesome! 👍 

/lgtm
/approve

PR #5115, opened 2020-02-20
Create/update credentials on update and migrations

### LINKED ISSUE #4552: Remove raw credentials from cluster after edit
**User Story** It seems that we may have a bug related to cluster edition. After the cluster is created the credentials are stored in the `credentialsReference` and removed from other properties (like `token` for DigitalOcean). When user edits the cluster in the UI it seems that it is not happening. Please see: https://github.com/kubermatic/dashboard-v2/pull/1718#issuecomment-546233867.

zreigz:
 - Removes the projectID label during secret creation since that label isn't used anywhere.
 - Adds logic to update credentials in PatchEndpoint.
 - Adds code to create/update credentials for migration. This also changes the migration-related code to use controller-runtime.
 - Adds unit test for credentials for migration.

Fixes #4552

```release-note
NONE
```

### DIFF
--- a/api/pkg/crd/migrations/seed/migrations.go
+++ b/api/pkg/crd/migrations/seed/migrations.go
@@ -15,7 +15,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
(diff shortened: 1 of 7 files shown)


api/pkg/crd/migrations/seed/migrations.go
```diff
@@ -152,17 +154,29 @@ func setExposeStrategyIfEmpty(cluster *kubermaticv1.Cluster, ctx *cleanupContext
...
 
 func cleanupDashboardAddon(cluster *kubermaticv1.Cluster, ctx *cleanupContext) error {
-	if err := ctx.kubermaticClient.KubermaticV1().Addons(cluster.Status.NamespaceName).Delete("dashboard", &metav1.DeleteOptions{}); err != nil {
+	dashboardAddon := &kubermaticv1.Addon{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: cluster.Status.NamespaceName,
+			Name:      "dashboard",
+		},
+	}
+	if err := ctx.client.Delete(context.TODO(), dashboardAddon); err != nil {
```
maciaszczykm: You are using `context.TODO()` in a few places. Can this be modified to use existing context where possible, i.e. you have some context in this method as a param, would it work?

zreigz: good point, I've changed for the existing context

PR #5117, opened 2020-02-20
[release/v2.12] add Kubernetes v1.15.10, v1.16.7

kdomanski:
```release-note
NONE
```

### DIFF
--- a/config/kubermatic/Chart.yaml
+++ b/config/kubermatic/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubermatic
-version: 1.0.24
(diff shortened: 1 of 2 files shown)


config/kubermatic/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubermatic
-version: 1.0.24
+version: 1.0.25
```
xrstf, line 3: This is not the place to discuss, but bumping chart versions in old release branches makes me wonder if we should use build suffixes for this (i.e. `1.0.24.a` or `1.0.24.1`).

PR #5119, opened 2020-02-20
Use head commit hash for KUBERMATICCOMMIT in run-controller.sh

irozzo-1A:
The KUBERMATICCOMMIT env variable in the `run-controller.sh` script is used to determine the kubermatic container images to be used, thus it should be related to the current HEAD and not `origin/master` otherwise the level of code and the image used risk to be misaligned when running the script for other branches/commits. 

Fixes #

**Documentation**:

```release-note
```

### DIFF
--- a/api/hack/run-controller.sh
+++ b/api/hack/run-controller.sh
@@ -3,12 +3,24 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x
 
 cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
-# Deploy a user-cluster/ipam-controller for which we actuallly
-# have a pushed image
-export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse origin/master)}"
+DEFAULT_IMAGE_TAG="$(git rev-parse 'origin/master')"
+if [[ -z "${KUBERMATICCOMMIT:-}" && \
+    "$(git rev-parse HEAD)" != "${DEFAULT_IMAGE_TAG}" ]]; then
+    cat << EOF >&2
+WARNING: The KUBERMATICCOMMIT is defaulting to origin/master head commit
+and your current HEAD is not on it.
+
+If you are on an old commit and you want to use the images corresponding to
+such version run the following command instead:
+
+$ KUBERMATICCOMMIT="$(git rev-parse HEAD)" $0
+EOF
+fi
+# KUBERMATICCOMMIT is used to define the tag of the k8c images used by the
+# seed controller.
+export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-${DEFAULT_IMAGE_TAG}}"
 make seed-controller-manager
 
 KUBERMATIC_WORKERNAME=${KUBERMATIC_WORKERNAME:-$(uname -n)}


api/hack/run-controller.sh
```diff
@@ -8,7 +8,7 @@ set -x
 cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
 # Deploy a user-cluster/ipam-controller for which we actuallly
 # have a pushed image
-export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse origin/master)}"
+export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse HEAD)}"
```
kron4eg: This wouldn't work in most cases. As images with tag from your local HEAD most likely wouldn't exist in quay

irozzo-1A: well, this would not work when your HEAD is ahead of origin/master because the images have not been push yet.
What we have now does not work either because if you go to an old commit (e.g. investigate a ticket on an old version) you expect to use the images that are coherent with that version, but it's not the case.
What would be better is to detect that there are no images tagged with current commit ID and either print a useful message of fallback to the HEAD of `origin/master`. WDYT?

kron4eg: the way `KUBERMATICCOMMIT` is defined is that at first it will try to find existing `KUBERMATICCOMMIT`, e.g.

```bash
KUBERMATICCOMMIT=123 ./api/hack/run-controller.sh
```

and only if it's not set it will resort to git rev-parse. So if you need any non `origin/master` hash, simply export it in shell before launching run-controller.

irozzo-1A: I saw that `KUBERMATICCOMMIT` can be overridden. My point is that we can either have a default value that is coherent with your current HEAD, i.e. my initial proposal or always point to `origin/master` i.e. current default. Both solutions have some limits, I thought my proposal was more intuitive, but apparently you do not agree with me :-).

kron4eg: I made exactly the same mistake as you here, and it blows up local dev ;) Been there, done that :D Current way is safe and flexible enough.

irozzo-1A: If we did the same mistake is a good indicator that something can be improved :-) I propose at least to output a warning (even if it's not my preferred option):
https://github.com/kubermatic/kubermatic/pull/5119/files#diff-20d4ab45cfb646bdb16543600d05af32R12

api/hack/run-controller.sh
```diff
@@ -3,12 +3,24 @@
 set -o errexit
 set -o nounset
 set -o pipefail
-set -x
```
kron4eg, line 6: why?

api/hack/run-controller.sh
```diff
@@ -3,12 +3,24 @@
...
 
 cd $(go env GOPATH)/src/github.com/kubermatic/kubermatic/api
-# Deploy a user-cluster/ipam-controller for which we actuallly
-# have a pushed image
-export KUBERMATICCOMMIT="${KUBERMATICCOMMIT:-$(git rev-parse origin/master)}"
+DEFAULT_IMAGE_TAG="$(git rev-parse 'origin/master')"
+if [[ -z "${KUBERMATICCOMMIT:-}" && \
+    "$(git rev-parse HEAD)" != "${DEFAULT_IMAGE_TAG}" ]]; then
+    cat << EOF >&2
+WARNING: The KUBERMATICCOMMIT is defaulting to origin/master head commit
```
kron4eg, line 12: Instead of writing in every run-* script things like this, I'd resort to documenting this in here https://github.com/kubermatic/kubermatic/tree/master/api#running-locally as those scripts are not meant for general public, we can keep their profile low.

api/hack/run-controller.sh
irozzo-1A: Because it's for debug, in a normal usage it just pollutes the output with not interesting info

api/hack/run-controller.sh
irozzo-1A: I think that one don't exclude the other

api/hack/run-controller.sh
xrstf: I don't see any major improvements in this PR compared to the current situation. When working on a release branch you cannot be certain that all images have already been pushed by some Prow job, so in 99.999% of cases it is much easier to just take the latest master branch version and be done with it. In the rare cases where you need some special tag, you can override the variable before running the script.

Using `HEAD` would be the worst of all approaches, as I for example often have PRs locally with unpushed commits, so Docker images with the HEAD hash will never exist on quay. Relying on the master branch for these auxiliary images is the safest solution here.

irozzo-1A: @xrstf My ambition was not to bring a major improvement here, whatever it means :-). 

> Using HEAD would be the worst of all approaches, as I for example often have PRs locally with unpushed commits, so Docker images with the HEAD hash will never exist on quay.

All default values have their limits, I already mentioned the example where you checkout an old version to reproduce an issue. You and @kron4eg think that the current default is better in most cases, fair enough. 

That said as you can see my last proposal was simply raise a warning when you run this script and you're not aligned with `origin/master`. It's just a small thing that can turn useful in the future for people that like me, and apparently @kron4eg, run this script without realising that the images used can be incoherent with the level of code.

xrstf: > That said as you can see my last proposal was simply raise a warning when you run this script and you're not aligned with origin/master

So this warning would basically be printed always for everyone all the time, no? When you are on the master branch, the current solution would already work. If you are not on the master branch, you get a warning. That means in most cases a warning will be printed.

irozzo-1A: It will be printed whenever you risk to use an image that is not aligned with your level of code. If you are not on the master branch you can get it or not depending if you are aligned with master branch or not. That means almost always, indeed :-)
Anyway it's just a log, it does not harm. I think we spent too many words on this already. If you don't like it for whatever reason just close it, it's not that important ;-)

PR #5122, opened 2020-02-20
Improve global settings WebSocket

maciaszczykm:
- Adds keep-alive mechanism for settings watcher. It is required as I have noticed that from time to time it is failing. We need to restart it each time it fails. WebSocket connection will be restarted by the client, so this should be fine.
- Exposes only external objects in the WebSocket.
- Implements `CheckOrigin` method. Used a default method extended by an allowance for requests coming from same host but different port.
- Use combined token extractor to extract auth token from header, cookie or query param. 

**Documentation**:

```release-note
NONE
```


maciaszczykm:
/hold cancel

I've added one more thing, that will be useful for clients and I forgot to do it earlier - WebSockets should use only send settings object.

PR #5126, opened 2020-02-20
Update Velero to v1.2.0

youssefazrak:
Update Velero to v1.2.0.
Minor upgrade. New features and bug fixes.

New mandatory flag `--plugins` to execute the `velero install` command that generates the CRDs.

**Documentation**:

Release note: https://github.com/vmware-tanzu/velero/releases/tag/v1.2.0

```release-note
Update Velero to v1.2.0
```

### DIFF
--- a/config/backup/velero/values.yaml
+++ b/config/backup/velero/values.yaml
@@ -3,8 +3,8 @@ velero:
   # if you are using restic, make sure to use an official image
   # that also contains the restic binary
   image:
-    repository: gcr.io/heptio-images/velero
-    tag: v1.1.0
+    repository: velero/velero
+    tag: v1.2.0
     pullPolicy: IfNotPresent
 
   # CLI flags to pass to velero server; note that the two flags
--- a/config/backup/velero/values.yaml
+++ b/config/backup/velero/values.yaml
@@ -3,7 +3,7 @@ velero:
   # if you are using restic, make sure to use an official image
   # that also contains the restic binary
   image:
-    repository: velero/velero
+    repository: docker.io/velero/velero
     tag: v1.2.0
     pullPolicy: IfNotPresent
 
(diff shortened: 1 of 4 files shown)


youssefazrak:
@xrstf Christoph, I updated the "update.sh" file with the `--plugins` flag as it is now a mandatory flag with Velero 1.2.0. When I launch the script with bash the formating is wrong (same in version 1.1.0) so that might be due to a different output with a newer version of Velero. It seems the update.sh was written for Velero 1.0.0.
Also, do you know if this is a manually triggered file or is it triggered by ci when a cluster is deployed? I can't seem to find any mention of "update.sh".
Cheers!

youssefazrak:
@xrstf I changed the update.sh script to work with the latest version of yq (v3.1.1). Now I have a proper output.
Regarding the newly generated CRD.yaml file, I guess we can remove the new annotation `controller-gen.kubebuilder.io/version: (unknown)`?
Also I suppose we need to apply those manually? 
Cheers

config/backup/velero/values.yaml
```diff
@@ -3,8 +3,8 @@ velero:
   # if you are using restic, make sure to use an official image
   # that also contains the restic binary
   image:
-    repository: gcr.io/heptio-images/velero
-    tag: v1.1.0
+    repository: velero/velero
```
xrstf: Please always use a fully-qualified Docker image name, i.e. `docker.io/velero/velero`.

xrstf:
> Regarding the newly generated CRD.yaml file, I guess we can remove the new annotation controller-gen.kubebuilder.io/version: (unknown)?

Nah, why remove them? They offer no real value, but they also don't hurt.

> Also I suppose we need to apply those manually?

Helm applies the CRDs for us, like with everything in `templates/`.

xrstf (changes requested):
minor

config/backup/velero/values.yaml
youssefazrak: Updated

PR #5127, opened 2020-02-20
Fix release branch builds

xrstf:
Currently, when building Docker images, the ci-push-images.sh script would search for the latest hash in the same branch on the dashboard repo. The exception were tagged Kubermatic releases, which require a matching tag in the dashboard.

However, the dashboard does not build Docker images for untagged revisions in release branches. It only has images for each master branch revision and all tagged releases. This leads to cases where we take the latest revision from a dashboard release branch and then cannot find a Docker image for it.

This PR fixes that, hopefully.

```release-note
NONE
```

### DIFF
--- a/api/hack/lib.sh
+++ b/api/hack/lib.sh
@@ -85,7 +85,7 @@ ensure_github_host_pubkey() {
 }
 
 get_latest_dashboard_hash() {
-  FOR_BRANCH="$1"
+  local FOR_BRANCH="$1"
 
   ensure_github_host_pubkey
   git config --global core.sshCommand 'ssh -o CheckHostIP=no -i /ssh/id_rsa'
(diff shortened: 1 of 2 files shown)


api/hack/lib.sh
```diff
@@ -99,6 +99,30 @@ get_latest_dashboard_hash() {
...
+  local DEPTH="${2:-25}"
+
+  ensure_github_host_pubkey
+  git config --global core.sshCommand 'ssh -o CheckHostIP=no -i /ssh/id_rsa'
+  local DASHBOARD_URL="git@github.com:kubermatic/dashboard-v2.git"
+
+  local TMPDIR
+  TMPDIR=$(mktemp -d dashboard.XXXXX)
+
+  # git ls-remote cannot list tags in a meaningful way, so we have to clone the repo
```
kdomanski, line 115: What does "list tags in a meaningful way" mean?

kdomanski:
Any particular reason why you didn't just restore parts of the code removed in 1b9ac115b77944eabf0244720636915e72378ea5 ?

api/hack/lib.sh
xrstf: "meaningful" as in "you can tell which branch a tag belongs to", vs. just a flat list of all tags with no chance to determine the latest tag in a given branch.

xrstf:
> Any particular reason why you didn't just restore parts of the code removed in 1b9ac11 ?

Because I did not search the history on this. I also think the old implementation removed in 1b9ac11 is inferior, as it seems to rely on string prefixes to match tags to branches.

xrstf:
/retest

Fixed the CI system, the job should now actually work.

PR #5144, opened 2020-02-25
Update openvpn container to v2.4.8

youssefazrak:
Openvpn update to v2.4.8. Security fixes, improvements etc.

This will update:
_ Openvpn for the user-clusters (control plane side and on worker nodes)
_ vpnsidecar
_ openshift addons

**Documentation**:

Release note: 
- https://github.com/OpenVPN/openvpn/releases/tag/v2.4.8
- https://github.com/OpenVPN/openvpn/releases/tag/v2.4.7

```release-note
Update openvpn to 2.4.8
```

### DIFF
--- a/containers/openvpn/Dockerfile
+++ b/containers/openvpn/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.9
+FROM alpine:3.11
 LABEL maintainer="support@loodse.com"
 
 RUN apk add -U \
--- a/containers/openvpn/Dockerfile
+++ b/containers/openvpn/Dockerfile
@@ -8,7 +8,7 @@ RUN apk add -U \
 	iptables \
 	jq \
 	openssl \
-	openvpn \
+	openvpn=2.4.8-r1 \
 	--no-cache
 
 # add kubectl
--- a/containers/openvpn/Dockerfile
+++ b/containers/openvpn/Dockerfile
@@ -1,4 +1,5 @@
 FROM alpine:3.11
+ARG OPENVPN_VERSION
 LABEL maintainer="support@loodse.com"
 
 RUN apk add -U \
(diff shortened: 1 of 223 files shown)


containers/openvpn/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.9
```
irozzo-1A, line 1: Why not setting the package versions explicitly, especially openvpn?

containers/openvpn/release.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-ver=v0.5
```
irozzo-1A, line 3: would't be worthy to use the same versioning of openvpn?

containers/openvpn/Dockerfile
youssefazrak: Alpine 3.11 comes by default with Openvpn 2.4.8.

containers/openvpn/release.sh
youssefazrak: That seems logical. @xrstf what do you think?

containers/openvpn/Dockerfile
irozzo-1A: Be aware that `3.9` is a floating tag and that in general when you use a dependency manager (e.g. `apk`) you have no guarantee that the version you get today is the same you'll get tomorrow, unless you specify explicitly. I think that being explicit about the version makes the creation of the image more determinist and reproducible, and this is important in particular if we decide to follow the same versioning of openvpn.

youssefazrak: Fair point. Package version is now specified for Openvpn.

containers/openvpn/release.sh
youssefazrak: Updated to 2.4.8

PR #5147, opened 2020-02-25
Fix vsphere/alibaba credential migration

xrstf:
The code for managing vsphere/alibaba credential secrets was not yet using the "create or update" strategy, but was always trying to create a new secret and quickly failing. This PR refactors the credential migration code a bit and fixes the issue.

```release-note
NONE
```

### DIFF
--- a/api/pkg/provider/kubernetes/credentials.go
+++ b/api/pkg/provider/kubernetes/credentials.go
@@ -1,13 +1,16 @@
 package kubernetes
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 
 	kubermaticv1 "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
 	"github.com/kubermatic/kubermatic/api/pkg/resources"
 	providerconfig "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
 
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


api/pkg/provider/kubernetes/credentials.go
```diff
@@ -48,624 +51,300 @@ func CreateOrUpdateCredentialSecretForCluster(ctx context.Context, seedClient ct
...
-		}
-
-		var updateNeeded bool
-		if existingSecret.Data != nil {
-			if tenantID, ok := existingSecret.Data[resources.AzureTenantID]; ok {
-				if string(tenantID) != spec.TenantID {
-					existingSecret.Data[resources.AzureTenantID] = []byte(spec.TenantID)
-					updateNeeded = true
-				}
-			}
```
xrstf, line 174: I think this was generally broken, as this code would never set a missing field in an existing Secret. It would only ever update an existing field.

api/pkg/provider/kubernetes/credentials.go
```diff
@@ -48,624 +51,300 @@ func CreateOrUpdateCredentialSecretForCluster(ctx context.Context, seedClient ct
...
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: secretData,
+	}
 
+	namespacedName := types.NamespacedName{Namespace: resources.KubermaticNamespace, Name: name}
+	existingSecret := &corev1.Secret{}
+	if err := seedClient.Get(ctx, namespacedName, existingSecret); err != nil && !kerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to probe for secret %q: %v", name, err)
+	}
```
xrstf, line 61: The logic was changed from "search for a Secret if the ref exists, otherwise create it", to a more reconciling-like approach "try to find it, reconcile it, return the reference for it."

zreigz:
I've added some changes because of the controller failure. Now works stable

PR #5151, opened 2020-02-26
Update controller-runtime to 0.4.0

xrstf:
This PR updates controller-runtime to 0.4.0. A few notes:

* ctrl-runtime 0.5.0 requires Go 1.13+, so that's out.
* ctrl-runtime 0.4.0 does not work with k8s 1.15 dependencies anymore, so it was also required to bump all of them to kubernetes-1.16.
* Due to what some consider a bug, the ctrl-runtime now always sets GVK on resources, so we have to specify the TypeMeta everywhere in our tests now. Likewise, the ctrl-runtime has begun to also mimic the resource version and bumps it on every update (https://github.com/kubernetes-sigs/controller-runtime/pull/620). So that's why we need to track that one as well in our tests.
* The pre-kubermatic-lint job did not complete anymore, because the linter was just consistently OOM-killed. Bumping the golang-ci to the most recent version not only fixed that, but dropped the memory usage to around 7-8GB. Unfortunately a few new linter errors needed to be fixed.
* `//lint:ignore <code>` comments do not work in the most recent golangci-lint version: https://github.com/golangci/golangci-lint/issues/741, that's why we have the "nolint" comment as well.
* ctrl-runtime 0.4 also begun throwing errors when bad owner refs were set (https://github.com/kubernetes-sigs/controller-runtime/pull/675). This unearthed quite a few bugs in the kubermatic operator. In a future PR some of the cleanups need to be implemented manually (for the seed admission webhook, clusterroles etc.)

```release-note
NONE
```


api/cmd/kubermatic-api/swagger.json
```diff
@@ -11059,16 +11059,6 @@
...
-      "properties": {
-        "Raw": {
-          "description": "Raw is the underlying serialization of this object.\n\nTODO: Determine how to detect ContentType and ContentEncoding of 'Raw' data.",
-          "type": "array",
-          "items": {
-            "type": "integer",
-            "format": "uint8"
-          }
-        }
-      },
```
irozzo-1A, line 11071: Why this has been removed? It's just for me to understand.

api/pkg/handler/v1/label/label.go
```diff
@@ -42,9 +42,7 @@ func GetSystemLabels() apiv1.ResourceLabelMap {
 // FilterLabels removes system labels from the provided labels map
 func FilterLabels(resource apiv1.ResourceType, labels map[string]string) map[string]string {
 	for _, label := range systemLabels[resource] {
-		if _, exists := labels[label]; exists {
-			delete(labels, label)
-		}
+		delete(labels, label)
```
irozzo-1A, line 45: good catch, the test was superfluous

api/pkg/handler/v1/label/label_test.go
```diff
@@ -40,7 +40,6 @@ func TestListSystemLabels(t *testing.T) {
 			req := httptest.NewRequest("GET", fmt.Sprintf("/api/v1/labels/system"), strings.NewReader(""))
 			res := httptest.NewRecorder()
 			kubermaticObj := []runtime.Object{}
-			kubermaticObj = append(kubermaticObj)
```
irozzo-1A, line 43: this was superuseful too :-)

irozzo-1A (commented):
LGTM, I just asked for some clarifications before approving

api/cmd/kubermatic-api/swagger.json
xrstf: go-swagger did whatever go-swagger does. I didn't even look at what it generated, as I did not touch the go-swagger version.

irozzo-1A: ok, did not know this was generated. With swagger you can either use a top-down approach where you generate code from spec or a bottom-down approach where you do the reverse. So this change was driven by a change you did in the code.

PR #5157, opened 2020-02-28
update endpoints to get/delete projects

### LINKED ISSUE #5154: Update endpoints to get/delete any project by admin
**User Story**
Update endpoints to get/delete projects. Allow admin user to delete/get any project

zreigz:
**What this PR does / why we need it**: the admin user can get/delete any project

Fixes #5154

```release-note
allow admin user to delete/get any project
```

### DIFF
--- a/api/pkg/handler/v1/project/project.go
+++ b/api/pkg/handler/v1/project/project.go
@@ -138,7 +138,7 @@ func isStatus(err error, status int32) bool {
 }
 
 // DeleteEndpoint defines an HTTP endpoint for deleting a project
-func DeleteEndpoint(projectProvider provider.ProjectProvider, userInfoGetter provider.UserInfoGetter) endpoint.Endpoint {
+func DeleteEndpoint(projectProvider provider.ProjectProvider, privilegedProjectProvider provider.PrivilegedProjectProvider, userInfoGetter provider.UserInfoGetter) endpoint.Endpoint {
 	return func(ctx context.Context, request interface{}) (interface{}, error) {
 		req, ok := request.(deleteRq)
 		if !ok {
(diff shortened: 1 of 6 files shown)


api/pkg/handler/v1/project/project.go
```diff
@@ -148,6 +148,19 @@ func DeleteEndpoint(projectProvider provider.ProjectProvider, userInfoGetter pro
 			return nil, errors.NewBadRequest("the id of the project cannot be empty")
 		}
 
+		// check if admin user
+		adminUserInfo, err := userInfoGetter(ctx, "")
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+		// allow to delete any project for the admin user
+		if adminUserInfo.IsAdmin {
```
maciaszczykm, line 157: I wonder if it wouldn't be easier to read if this method would end here. I would look like this:

```go
if adminUserInfo.IsAdmin {
  err := privilegedProjectProvider.DeleteUnsecured(req.ProjectID)
} else {
  (a new private method to delete as regular user)
}
```

Thanks to that we could get rid of some comments like `delete project by regular user` as the method names would explain what is happenning. And we would avoid mixing the code for admins and regular users. What do you think?

api/pkg/handler/v1/project/project.go
```diff
@@ -204,15 +217,31 @@ func GetEndpoint(projectProvider provider.ProjectProvider, memberProvider provid
...
-		if err != nil {
-			return nil, common.KubernetesErrorToHTTPError(err)
+
+		var kubermaticProject *kubermaticapiv1.Project
+
+		if adminUserInfo.IsAdmin {
+			// get any project for admin
+			kubermaticProject, err = privilegedProjectProvider.GetUnsecured(req.ProjectID, &provider.ProjectGetOptions{IncludeUninitialized: true})
+			if err != nil {
+				return nil, common.KubernetesErrorToHTTPError(err)
```
maciaszczykm: Same as above, it's easier to read code if we don't have too many nested returns.

PR #5161, opened 2020-02-28
remove pre-kubermatic-build job

xrstf:
This job does not do anything useful. We compile the code in basically all other jobs anyway. If the code is good, it will succeed. Otherwise *all* job will fail. We gain nothing from having this dedicated compile job, so good riddance.

```release-note
NONE
```


kron4eg (approved):
/lgtm
/approve

PR #5163, opened 2020-02-28
Dedicated Certificates for Kubermatic/IAP

xrstf:
So far we had a big "catch all" `certs` chart that just created one big cert and we hoped that it always contains all relevant domains. This was weird to manage (adding a new IAP meant re-deploying the certs chart) and weird to understand ("Kubermatic uses HTTPS, but the Ingress says nothing about what cert? Where does it come from?" -- until you find out about the certs chart).

This PR removes the catch-all-cert and instead makes the kubermatic/iap charts create dedicated certificates for each ingress/domain. The `certs` chart is therefore not needed anymore.

The cert-manager in Kubermatic 2.13 is already configured to create a ClusterIssuer for Let's Encrypt and we moved the DNS stuff already over to the cert-manager as well. It should be safe to update.

**Documentation:**
https://github.com/loodse/docs/pull/248

```release-note
TLS certificates for Kubermatic/IAP are now not managed by a shared `certs` chart anymore, but handled individually for each Ingress.
```

### DIFF
--- a/api/hack/ci/ci-setup-kubermatic-in-kind.sh
+++ b/api/hack/ci/ci-setup-kubermatic-in-kind.sh
@@ -327,6 +327,12 @@ function check_all_deployments_ready() {
   return 0
 }
 
+# We don't need a valid certificate (and can't even get one), but still need
+# to have the CRDs installed so we can at least create a Certificate resource.
+TEST_NAME="Deploy cert-manager CRDs"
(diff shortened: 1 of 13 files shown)


kron4eg:
/lgtm
/hold

@youssefazrak PTAL

youssefazrak:
All good.
/hold cancel

kron4eg (approved):
/lgtm

PR #5164, opened 2020-02-28
begin to publish the new logging stack

xrstf:
We already document on the master branch that Loki is replacing the ELK stack, so it's time to also publish our charts to the kubermatic-installer repo. The ELK stack charts are still published because maaaaaybe there will be a bugfix in them.

```release-note
Grafana Loki replaces the ELK logging stack.
```


kron4eg (approved):
/lgtm
/approve

PR #5165, opened 2020-02-28
bump envoyproxy/go-control-plane to 0.9.4

xrstf:
This updates the go-control-plane library to the latest version. This involves a lot of tiny syntax changes (slices of pointers instead of structs, for example).

The PR also

* changes the logging library to zap
* removes the log proxy, as envoy now makes use of debug/warn levels on its own (and Zap is compatible to their logging interface)
* fixes the `Using deprecated option 'envoy.api.v2.listener.Filter.config' from file listener_components.proto.` deprecation
* fixes the seemingly random `SocketAddressValidationError.Address: ["value length must be at least " '\x01' " bytes"])` error which was caused by a bad loop adress bug in our `getReadyServicePods()`

```release-note
NONE
```


api/cmd/nodeport-proxy/envoy-manager/controller.go
```diff
@@ -122,40 +127,42 @@ func (r *reconciler) getInitialResources() (listeners []envoycache.Resource, clu
 		},
 		HttpFilters: []*envoyhttpconnectionmanagerv2.HttpFilter{
 			{
-				Name:       envoyutil.HealthCheck,
-				ConfigType: &envoyhttpconnectionmanagerv2.HttpFilter_Config{Config: healthCheckMsg},
+				Name: envoywellknown.HealthCheck,
+				ConfigType: &envoyhttpconnectionmanagerv2.HttpFilter_TypedConfig{
+					TypedConfig: healthCheckMarshalled,
+				},
```
xrstf, line 133: This is one of the example fixes for the deprecation warning about filters.

api/cmd/nodeport-proxy/envoy-manager/controller.go
```diff
@@ -360,9 +374,9 @@ func (r *reconciler) getReadyServicePods(service *corev1.Service) ([]*corev1.Pod
 	}
 
 	// Filter for ready pods
-	for _, pod := range servicePods.Items {
+	for idx, pod := range servicePods.Items {
 		if PodIsReady(&pod) {
-			readyPods = append(readyPods, &pod)
+			readyPods = append(readyPods, &servicePods.Items[idx])
```
xrstf, line 379: This fixes the warning about bad socket addresses because we filtered the pods badly.

api/cmd/nodeport-proxy/lb-updater/main.go
```diff
@@ -168,28 +185,29 @@ func (u *LBUpdater) syncLB(s string) error {
...
+	buf.WriteString("======================\n")
+	buf.WriteString("Updated LB Ports:\n")
+	for _, p := range lb.Spec.Ports {
+		buf.WriteString(fmt.Sprintf("Name: %s\n", p.Name))
+		buf.WriteString(fmt.Sprintf("Port: %d\n", p.Port))
+		buf.WriteString(fmt.Sprintf("NodePort: %d\n", p.NodePort))
+		buf.WriteString("\n")
 	}
+	buf.WriteString("======================\n")
+	u.log.Debug(buf.String())
```
xrstf, line 210: Log these diffs only on the debug level, because they are not that important in production.

api/cmd/nodeport-proxy/lb-updater/main_test.go
```diff
@@ -500,17 +501,14 @@ func TestReconciliation(t *testing.T) {
 		},
 	}
 
-	var breakNow bool
 	for _, tc := range testCases {
-		if breakNow {
-			break
-		}
```
xrstf, line 507: No idea why this was needed if the test fails with `t.Fatalf`.

api/cmd/nodeport-proxy/envoy-manager/controller.go
```diff
@@ -209,33 +217,35 @@ func (r *reconciler) sync() error {
 
 		// If we have no pods, dont bother creating a cluster.
 		if len(pods) == 0 {
-			r.log.Debugf("skipping service %s/%s as it has no running pods", service.Namespace, service.Name)
+			serviceLog.Debugf("Skipping service: it has no running pods")
```
moadqassem: So `Debugf` is not needed since you are not formating any explicit values, thus `Debug` can be used. However, maybe you can keep it `Debugf` but with adding the namespace of the service in the message.

api/cmd/nodeport-proxy/lb-updater/main.go
```diff
@@ -122,21 +137,24 @@ func (u *LBUpdater) syncLB(s string) error {
...
+
 		if service.Annotations[exposeAnnotationKey] != "true" {
-			klog.V(4).Infof("skipping service %s/%s as the annotation %s is not set to 'true'", service.Namespace, service.Name, exposeAnnotationKey)
+			serviceLog.Debugw("Skipping service as the annotation is not set to 'true'", "annotation", exposeAnnotationKey)
 			continue
 		}
 
 		if service.Spec.ClusterIP == "" {
-			klog.V(4).Infof("skipping service %s/%s as it has no clusterIP set", service.Namespace, service.Name)
+			serviceLog.Debugw("Skipping service as it has no clusterIP set")
```
moadqassem: Maybe `serviceLog.Debug()` instead of `serviceLog.Debugw()`

moadqassem (commented):
Looks good except those two minor comments

api/cmd/nodeport-proxy/envoy-manager/controller.go
xrstf: Fixed. The service namespace is already one of the extra fields on the `serviceLog` object.

api/cmd/nodeport-proxy/lb-updater/main.go
xrstf: Fixed.

PR #5181, opened 2020-03-04
Revert "Dedicated Certificates for Kubermatic/IAP"

kron4eg:
Reverts kubermatic/kubermatic#5163


xrstf:
I think https://github.com/kubermatic/kubermatic/pull/5184 solves the remaining problems, no need to revert and basically restore again with some minimal additions.

PR #5184, opened 2020-03-04
Configure dedicated Certificate for Dex Ingress

xrstf:
In #5163 we forgot that Dex also has its own Ingress and was still relying on the default cert. Even when the Kubermatic Ingress already contains a cert for a given domain, nginx in this case used the non-existing cert from the Dex Ingress for the domain.

As the two Ingresses are in different namespaces, they cannot share the same cert/secret. To keep them somewhat independent, I chose to give Dex its own certificate, so the chart simply works without weird nginx hacks and another helper chart.

This leaves us with two certificates for the same domain (as Kubermatic and Dex are in 99% of cases on the same domain), so nginx will simply choose whatever it likes best.

```release-note
NONE
```

### DIFF
--- a/config/oauth/Chart.yaml
+++ b/config/oauth/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: oauth
-version: 1.0.11
+version: 1.1.0
 appVersion: v2.22.0
 description: Dex
 keywords:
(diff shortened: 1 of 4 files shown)


kron4eg (approved):
/approve

PR #5188, opened 2020-03-05
Add endpoints to fetch alibaba zones 

kgroschoff:
Add endpoints to fetch alibaba zones 

```release-note
Added endpoints to fetch alibaba zones 
```


api/pkg/api/v1/types.go
```diff
@@ -1158,7 +1168,7 @@ type AlibabaNodeSpec struct {
 	InstanceType            string            `json:"instanceType"`
 	DiskSize                string            `json:"diskSize"`
 	DiskType                string            `json:"diskType"`
-	VSwitchID               string            `json:"vswitchID"`
+	VSwitchID               string            `json:"vSwitchID"`
```
kgroschoff, line 1171: FYI: this one was a typo, which would have lead to problems when editing the node deployment via the ui 😉

PR #5192, opened 2020-03-06
Fix validation for alibaba

kgroschoff:
Fix validation for alibaba. 
Previously it wasn't possible to just change the name via the UI, as the backend complained, that no credentials are provided. This should not happen.

```release-note
NONE
```


kgroschoff:
@moadqassem could you have a look please? Hopefully the last alibaba PR to review for a while 😅

PR #5193, opened 2020-03-06
Add configurable default addon enforce option

### LINKED ISSUE #4979: enforcement of default addons in user cluster
**User Story**

**Acceptance criteria**

moelsayed:
Add configurable default addon enforce option.
Fixes #4979

**Documentation**:

```release-note
Add a configuration flag for seed-controller-manager to enforce default addons on userclusters. Enabled by default.
```


kron4eg (changes requested):
@moelsayed please see https://github.com/kubermatic/kubermatic/issues/4979#issuecomment-601651108

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -196,7 +201,14 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
 	if result == nil {
-		result = &reconcile.Result{}
+		// we check for this after the ClusterReconcileWrapper() call because otherwise the cluster would never reconcile since we always requeue
+		if r.addonEnforceInterval == 0 { // addon enforce is disabled.
+			result = &reconcile.Result{}
+		} else {
+			// All is well, requeue in addonEnforceInterval minutes. We do this to enforce default addons and prevent cluster admins from disabling them.
+			result = &reconcile.Result{RequeueAfter: time.Duration(r.addonEnforceInterval) * time.Minute}
+		}
```
kron4eg:
```suggestion
		// we check for this after the ClusterReconcileWrapper() call because otherwise the cluster would never reconcile since we always requeue
		result = &reconcile.Result{}
		if r.addonEnforceInterval != 0 { // addon enforce is enabled
			// All is well, requeue in addonEnforceInterval minutes. We do this to enforce default addons and prevent cluster admins from disabling them.
			result.RequeueAfter = time.Duration(r.addonEnforceInterval) * time.Minute
		}
```

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -234,6 +246,12 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, addo
 		}
 		return nil, nil
 	}
+	// This is true when the addon: 1) is fully deployed, 2) doesn't have a `addonEnsureLabelKey` set to true.
+	// we do this to allow users to "edit/delete" resources deployed by unlabeled addons,
+	// while we enfornce the labeled ones
```
kron4eg, line 250:
```suggestion
	// while we enforce the labeled ones
```

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -543,3 +573,48 @@ func (r *Reconciler) ensureRequiredResourceTypesExist(ctx context.Context, log *
...
+}
+
+func hasEnsureResoucesLabel(addon *kubermaticv1.Addon) bool {
+	if len(addon.Labels) != 0 {
+		if ensure := addon.Labels[addonEnsureLabelKey]; ensure == "true" {
+			return true
+		}
+	}
+	return false
+}
```
kron4eg: It's completely legal to read from uninitialized map (in case when it's empty or nil).
```suggestion
func hasEnsureResoucesLabel(addon *kubermaticv1.Addon) bool {
	return addon.Labels[addonEnsureLabelKey] == "true"
}
```

kron4eg (commented):
Looks good! Albeit left few minor comments.

kron4eg (approved):
/hold

@moadqassem PTAL too

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -543,3 +572,43 @@ func (r *Reconciler) ensureRequiredResourceTypesExist(ctx context.Context, log *
...
+
+func addonResourcesCreated(addon *kubermaticv1.Addon) bool {
+	_, cond := getAddonCondition(addon, kubermaticv1.AddonResourcesCreated)
+	if cond != nil && cond.Status == corev1.ConditionTrue {
+		return true
+	}
+	return false
+}
+
+func hasEnsureResoucesLabel(addon *kubermaticv1.Addon) bool {
```
moadqassem: typo hasEnsureResoucesLabel should be hasEnsureResourcesLabel

moadqassem: and does it really need to do this check, in a function instead of having it inline? I mean 1) it is an unexported function 2) it is only used one time.

moelsayed: Fixed the typo. I prefer to wrap it in a function because I do think it's more readable this way..

multi-io:
Shouldn't the addon installer controller also honour this setting for existing clusters? Ie. update the default addons in those clusters and update the label on them.

PR #5211, opened 2020-03-11
Manage NodePortProxy using Kubermatic Operator

xrstf:
Kubermatic needs the nodeport-proxy (NPP) in each seed to function properly, and the NPP is useless without Kubermatic. The operator is already reconciling each Seed, so it would be nice if it also quickly installed the NPP. If there's a 1:1 relation between the two, why not merge them?

This PR makes it so that the operator also installs the NPP. It will then live inside the Kubermatic namespace, so it's explicitly not compatible with the old Helm chart. This means that it will not 1:1 replace it or try to migrate. The intention is that new customers use the new NPP and existing customers should slowly migrate while the NPP chart is deprecated. Running both NPP installations in parallel is possible (you end up with 2 identical LoadBalancer services), so migrations can happen slowly over time. If customers really do not want the new NPP, they can always disable it per seed in the Seed CRD.

```release-note
NONE
```


api/pkg/controller/operator/common/defaults.go
```diff
@@ -481,6 +525,44 @@ func DefaultConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logg
 	return copy, nil
 }
 
+func DefaultSeed(seed *kubermaticv1.Seed, logger *zap.SugaredLogger) (*kubermaticv1.Seed, error) {
+	logger = logger.With("seed", seed.Name)
+	logger.Debug("Applying defaults to Seed")
+
+	copy := seed.DeepCopy()
```
moadqassem, line 532: Nit: Isn't this a builtin function? maybe a better naming?

api/pkg/controller/operator/common/defaults.go
```diff
@@ -120,6 +123,47 @@ var (
...
+	}
+
+	DefaultNodeportProxyUpdaterResources = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("50m"),
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("150m"),
+			corev1.ResourceMemory: resource.MustParse("32Mi"),
```
moadqassem, line 155: Nit: maybe `48Mi` to be different than what it requests?

api/pkg/controller/operator/seed/resources/nodeportproxy/rbac.go
```diff
@@ -0,0 +1,109 @@
...
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+const (
+	RoleName        = "nodeport-proxy"
+	RoleBindingName = "nodeport-proxy"
+)
+
+func RoleCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedRoleCreatorGetter {
```
moadqassem: `cfg` is an unused parameter ;)

api/pkg/controller/operator/seed/resources/nodeportproxy/rbac.go
```diff
@@ -0,0 +1,109 @@
...
+					Name:      ServiceAccountName,
+				},
+			}
+
+			return crb, nil
+		}
+	}
+}
+
+func ServiceAccountCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedServiceAccountCreatorGetter {
```
moadqassem, line 103: same goes here `cfg` is an unused parameter ;)

api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+const (
+	ServiceAccountName    = "nodeport-proxy"
+	EnvoyDeploymentName   = "nodeport-proxy-envoy"
+	UpdaterDeploymentName = "nodeport-proxy-updater"
+	EnvoyPort             = 8002
+)
+
+func EnvoyDeploymentCreator(cfg *operatorv1alpha1.KubermaticConfiguration, seed *kubermaticv1.Seed, versions common.Versions) reconciling.NamedDeploymentCreatorGetter {
```
moadqassem: `cfg` parameter is not used.

api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
```diff
@@ -0,0 +1,231 @@
...
+				MatchLabels: map[string]string{
+					common.NameLabel: EnvoyDeploymentName,
+				},
+			}
+			return pdb, nil
+		}
+	}
+}
+
+func UpdaterDeploymentCreator(cfg *operatorv1alpha1.KubermaticConfiguration, seed *kubermaticv1.Seed, versions common.Versions) reconciling.NamedDeploymentCreatorGetter {
```
moadqassem: `cfg` parameter is not used.

api/pkg/controller/operator/common/defaults.go
xrstf: I don't understand that reasoning at all :-D Use a different value just to be different? What's bad about a fixed-size (non-burstable) pod?

moadqassem: Sorry 😂! my expressing ability has failed me( happening quite a lot lately 🤔). I thought that you have actually forgotten to adjust this as the other resource( the `DefaultNodeportProxyEnvoyManagerResources`).

api/pkg/controller/operator/common/defaults.go
xrstf: What do you mean? Yes, the DeepCopy() is auto-generated.

moadqassem: `copy` is a builtin function

xrstf:
PR feedback integrated.

PR #5231, opened 2020-03-19
Use Loki stack instead of Elastic Stack on offline-env

xrstf:
We want to scale down the offline environment and for this purpose we decided that running Loki instead of the Elastic stack allows us to use smaller instances. This PR makes it so that the offline deployment actually updates the monitoring/logging stacks.

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-deploy-offline.sh
+++ b/api/hack/ci/ci-deploy-offline.sh
@@ -56,8 +56,8 @@ helm template oauth | ../api/hack/retag-images.sh
 helm template iap | ../api/hack/retag-images.sh
 helm template minio | ../api/hack/retag-images.sh
 helm template s3-exporter | ../api/hack/retag-images.sh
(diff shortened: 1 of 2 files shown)


api/hack/ci/ci-deploy-offline.sh
```diff
@@ -56,8 +56,8 @@ helm template oauth | ../api/hack/retag-images.sh
 helm template iap | ../api/hack/retag-images.sh
 helm template minio | ../api/hack/retag-images.sh
 helm template s3-exporter | ../api/hack/retag-images.sh
-helm template nodeport-proxy --set=nodePortProxy.image.tag=${GIT_HEAD_HASH}	\
-	| ../api/hack/retag-images.sh
+helm template nodeport-proxy --set=nodePortProxy.image.tag=${GIT_HEAD_HASH} \
+  ../api/hack/retag-images.sh
```
xrstf, line 60: This just fixed weird inline tabs.

PR #5243, opened 2020-03-23
Fix node-exporter addon being impossible to scrape

xrstf:
Up until now it was pretty much impossible to configure Prometheus to scrape the node-exporter addon in user clusters. This PR fixes that. For those interested, the failed attempts are described below the release notes.

## Foreword

The most important piece of knowledge here is that the node-exporter **runs in the host network.** This puts it (and its pod IP) outside the Pod IP range of the user cluster and therefore outside of the VPN.

Old solution: In order to not make the node-exporter available on all interfaces of a node (and hence make it public, unless someone put firewall rules into place), we make the exporter run on `127.0.0.1:9100` and then have a *kube-rbac-proxy* sidecar which binds on the pod IP (this can be an internal or external IP, dependending on the cloud provider), checks credentials and then forwards to the upstream node-exporter on `127.0.0.1`.

## Solution: Escape host network via UNIX sockets

The kube-rbac-proxy is gone entirely. Instead the node-exporter pod now has a socat sidecar, which listens on a socket on the host and forwards traffic to the node-exporter. A second DaemonSet runs another socat proxy, but this time taking Prometheus scrape requests and forwarding them to the local socket. This removes the need to find the neighboring pod, as the proxy simply uses whatever is present on the host's disk.

## Addon Configuration

This PR also extends the addon configuration so that the scrape annotation prefix is available for templating in addon manifests. This makes it so that cluster operators can configure the prefix once and have the addons be reconciled properly.

In order to not leak our Product name, we have to revert the default value for the flag to an empty string.

The node-exporter addon now expects a `port` variable and defaults to a **non-default** port for the exporter. This makes it so that customers can install their own node-exporter without conflicting with the addon.

## Security Considerations

This PR makes a socket available on every node, under which anyone can reach the node-exporter. I don't think this makes it less secure: If you can read the socket file, you could also just talk to port 9100 locally.

## Backwards Compatibility

The default port for node-exporters changed from 9100 to 49100. Given that nobody so far could feasably have scraped it, I don't think we need to classify this as "breaking change". The new port should hopefully be safer and cluster operators can override it by setting the variable for each user-cluster manually.

The changed default value for the scrape annotation however is breaking, but I think it's okay to document that customers need to set `kubermatic.monitoringScrapeAnnotationPrefix=whatever.they.like` when updating.

```release-note
The node-exporter addon gets automatically scraped when installed into user clusters.
Action required: default scrape annotation prefix default value was removed.
```

#### Attempt 1: Scrape Pod via kube-apiserver proxy.

The obvious approach and what we already do for all other pods in the user-cluster (if they have the special scraping prefix annotation). This makes Prometheus send the scraping request to the kube-apiserver, which proxies it to the pod.

I tried

* `/api/v1/namespaces/kube-system/pods/node-exporter-xyz:9100/proxy` (sic)
* `/api/v1/namespaces/kube-system/pods/node-exporter-xyz:9100/proxy/metrics`
* `/api/v1/namespaces/kube-system/pods/https:node-exporter-xyz:9100/proxy/metrics`

All ended up in a timeout. Even if this worked, the credentials are not forwarded by the kube-apiserver, so the bearer token would be received by kube-rbac-proxy anyway.

#### Attempt 2: Scrape Nodes via kube-apiserver proxy.

Same as before, but doing Prometheus service discovery on nodes instead of pods. I tried

* `/api/v1/nodes/<ip>:9100/proxy/metrics`
* `/api/v1/nodes/https:<ip>:9100/proxy/metrics`
* `/api/v1/nodes/<hostname>:9100/proxy/metrics`
* `/api/v1/nodes/https:<hostname>:9100/proxy/metrics`

All failed.

#### Attempt 3: Scrape the node directly

This fails because

* we might have firewall rules in place that prevent access to the nodes
* we never know if a node IP is internal or external
* internal node IPs are not reachable via VPN

#### Attempt 4: Remove kube-rbac-proxy, bind to private IP

This would make the node-exporter bind to the node's internal IP instead of localhost. Not all cloud providers provide internal IPs, so this is out. It would also not solve the issue of reaching the node in the first place (see attempt 3).

#### Attempt 5: Run a kubectl-proxy sidecar container

This sidecar cannot run in the node-exporter pod, because it would not solve anything attempt 4 would not have solved (removing authentication). So the "sidecar" would be a "sidecar pod", i.e. a second DaemonSet. The problem then is that the pod needs to find its neighboring node-exporter pod. This is simple with kubectl and jq, but we need to do this in a public Docker image, because due to whitelabelling we cannot create a quay.io/kubermatic/... helper image with the required tools. Good luck finding good kubectl+jq images out there.

### DIFF
--- a/addons/node-exporter/daemonset-proxy.yaml
+++ b/addons/node-exporter/daemonset-proxy.yaml
@@ -1,0 +1,70 @@
+{{- $port := .Variables.port | default 49100 -}}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: node-exporter-proxy
+  namespace: kube-system
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+    app.kubernetes.io/name: node-exporter-proxy
+    app.kubernetes.io/version: v0.18.0
+spec:
+  updateStrategy:
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: node-exporter-proxy
+  template:
+    metadata:
+{{- with .ScrapeAnnotationPrefix }}
+      annotations:
+        {{ . }}/path: /metrics
+        {{ . }}/port: "{{ $port }}"
+{{- end }}
+      labels:
+        app.kubernetes.io/name: node-exporter-proxy
+    spec:
+      containers:
+      - name: socat
+        image: '{{ Registry "docker.io" }}/alpine/socat:1.7.3.4-r0'
+        imagePullPolicy: IfNotPresent
+        command: [/bin/sh]
+        args:
+        - -c
+        - |
+          set -e
+
+          socket=/shared/node-exporter-socket
+          if [ ! -S "$socket" ]; then
+            echo "Socket $socket has not yet been opened, exiting."
+            exit 1
+          fi
+
+          socat -d -d TCP4-LISTEN:{{ $port }},fork,reuseaddr UNIX:$socket
+        resources:
+          requests:
+            cpu: 10m
+            memory: 24Mi
+          limits:
+            cpu: 100m
+            memory: 48Mi
+        ports:
+        - containerPort: {{ $port }}
+          hostPort: {{ $port }}
+          name: http
+        volumeMounts:
+        - name: shared
+          mountPath: /shared
+          mountPropagation: HostToContainer
+      tolerations:
+      - effect: NoExecute
+        operator: Exists
+      - effect: NoSchedule
+        operator: Exists
+      volumes:
+      - name: shared
+        hostPath:
+          path: /tmp
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
--- a/addons/node-exporter/daemonset.yaml
+++ b/addons/node-exporter/daemonset.yaml
@@ -1,3 +1,4 @@
+{{- $port := .Variables.port | default 49100 -}}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
@@ -15,7 +16,6 @@ spec:
       app.kubernetes.io/name: node-exporter
   template:
     metadata:
-      name: node-exporter
       labels:
         app.kubernetes.io/name: node-exporter
     spec:
@@ -29,7 +29,7 @@ spec:
         - '--path.procfs=/host/proc'
         - '--path.sysfs=/host/sys'
         - '--path.rootfs=/host/root'
-        - '--web.listen-address=127.0.0.1:9100'
+        - '--web.listen-address=127.0.0.1:{{ $port }}'
         resources:
           requests:
             cpu: 10m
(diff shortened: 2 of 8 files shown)


addons/node-exporter/daemonset-proxy.yaml
```diff
@@ -0,0 +1,70 @@
...
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: node-exporter-proxy
+  template:
+    metadata:
+{{- with .ScrapeAnnotationPrefix }}
+      annotations:
+        {{ . }}/path: /metrics
+        {{ . }}/port: "{{ $port }}"
+{{- end }}
```
xrstf, line 23: We could/should do this thing in all our addons, maybe? In a future PR.

irozzo-1A:
@xrstf Really valuable to have all this context in PR ;-).

If I'm not mistaken you did not explore the possibility to extend the mechanism in place to enable communication from apiserver to the kubelet through a VPN tunnel (maybe @youssefazrak  can advise here).

Regarding the attempts you made the following points are not clear to me:

1. What caused the timeout you mentioned for attempts 1 and 2.
2. Why someone would deploy node-exporter if it is already there (e.g. different version, configuration, ...) ?at first glance it looks to me just a waste of resources.

xrstf:
> What caused the timeout you mentioned for attempts 1 and 2.

If only I knew. Nobody logged anything, and I reconfigured my kube-apiserver to run with `--v=12` and still saw nothing except "go-restblabla is handling this proxy request" and that's it.

It also depends on the cloud provider, as there seem to be more restrictive firewalling rules on AWS than on DigitalOcean.

> Why someone would deploy node-exporter if it is already there (e.g. different version, configuration, ...) ?at first glance it looks to me just a waste of resources.

Because the addon might be configured as a default addon, because the cluster operator wants all user clusters to have a node-exporter for monitoring purposes.

irozzo-1A:
> Because the addon might be configured as a default addon, because the cluster operator wants all user clusters to have a node-exporter for monitoring purposes.

If the operator impose to have node-exporter deployed on all user clusters it's a constraint of the system. The user cluster administrator should be simply aware that node-exporter is already deployed and he can scrape it if he has a personal prometheus deployment. IMHO this should be the preferred way.

youssefazrak:
> The most important piece of knowledge here is that the node-exporter runs in the host network. This puts it (and its pod IP) outside the Pod IP range of the user cluster and therefore outside of the VPN.

IIRC node-exporter is bound to both localhost and private IP of the node on port 9100. Which in this case makes it reachable from the VPN.
Currently, the VPN simply serves for the API-server and metrics-server to connect to the kubelet on the node. This is done using the node-access-network which will get dnat'ed when arriving on the node and reach the private IP of the node (at least on AWS).

Using the VPN I can see the node-exporter replying:
/ # curl -k --http1.1 https://10.254.55.198:9100/metrics
Unauthorized
/ # curl -k --http1.1 https://10.254.55.198:9100/proxy/metrics
Unauthorized

Those commands are executed from the apiserver pod. Obviously I don't know about this token stuff but if it can be passed via CURL we might get what we want.

xrstf:
Whoops, fat-fingered while cleaning up my branches.

xrstf:
Alternative implementation is in #5264.

PR #5262, opened 2020-03-27
Keep most recent backup for every user cluster on deletion

xrstf:
For easier cluster recovery after a user cluster was deleted, it would be nice to still have backups. Currently we simply wipe everything, which is something cluster operators should opt-in to.

This PR makes it so that the most recent backup for a cluster is kept. And it leaves a nice example for customers who want to actually delete everything. Removing the cleanup container by default would have made it much more difficult for operators to correctly configure the cleanup.

```release-note
Action required: The most recent backup for user clusters is kept when the cluster is deleted. Adjust the cleanup-container to get the old behaviour (delete all backups) back.
```

### DIFF
--- a/api/pkg/controller/operator/common/defaults.go
+++ b/api/pkg/controller/operator/common/defaults.go
@@ -644,8 +644,12 @@ command:
 - -c
 - |
   set -euo pipefail
-  s3-storeuploader store --endpoint minio.minio.svc.cluster.local:9000 --bucket kubermatic-etcd-backups --create-bucket --prefix $CLUSTER --file /backup/snapshot.db
-  s3-storeuploader delete-old-revisions --endpoint minio.minio.svc.cluster.local:9000 --bucket kubermatic-etcd-backups --prefix $CLUSTER --file /backup/snapshot.db --max-revisions 20
+
+  endpoint=minio.minio.svc.cluster.local:9000
+  bucket=kubermatic-etcd-backups
+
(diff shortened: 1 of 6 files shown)


api/pkg/controller/operator/common/defaults.go
```diff
@@ -670,7 +674,15 @@ command:
 - -c
 - |
   set -euo pipefail
-  s3-storeuploader delete-all --endpoint minio.minio.svc.cluster.local:9000 --bucket kubermatic-etcd-backups --prefix $CLUSTER
+
+  endpoint=minio.minio.svc.cluster.local:9000
+  bucket=kubermatic-etcd-backups
+
+  # be default, we keep the most recent backup for every user cluster
```
moelsayed: typo: s/be/by/

api/pkg/controller/operator/common/defaults.go
```diff
@@ -670,7 +674,15 @@ command:
...
+  s3-storeuploader delete-old-revisions --max-revisions 1 --endpoint "$endpoint" --bucket "$bucket" --prefix $CLUSTER
+
+  # alternatively, delete all backups for this cluster
+  #s3-storeuploader delete-all --endpoint "$endpoint" --bucket "$bucket" --prefix $CLUSTER
```
moelsayed, line 685: Do we still need this commented line?

xrstf: Yes, it's explicitly left there as an example for a customer who wants to modify it. Otherwise you would have to trial&error our S3 tools to figure out that there even is a `delete-all` command.

PR #5263, opened 2020-03-29
addon controller: requeue reconciliations if customer apiserver is down

multi-io:

In the current master, when the addon controller tries to reconcile an addon and sees that the customer apiserver isn't ready, the reconciliation request will be dropped, essentially as if the reconciliation had been successful. If no further reconciliation requests are ever received for that addon, e.g. because no further changes to the addon or cluster resource occur, the addon won't ever be reconciled until the controller manager is restarted. Correct me if I'm missing something. We saw this problem happen in our staging environment numerous times after the 2.13 merge, especially in integration tests where we create a cluster and install a whole bunch of addons very shortly after. The cluster’s AddonControllerReconciledSuccessfully condition ends up being set to True, but none of the addons are in fact reconciled.

This PR modifies the addon controller so that if the customer apiserver is down, addon reconciliation requests are re-queued after 10 seconds rather than dropped. I can no longer reproduce the problem with this patch in place.

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/seed-controller-manager/addon/addon_controller.go
+++ b/api/pkg/controller/seed-controller-manager/addon/addon_controller.go
@@ -203,8 +203,8 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
 	if cluster.Status.ExtendedHealth.Apiserver != kubermaticv1.HealthStatusUp {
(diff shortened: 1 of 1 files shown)


xrstf:
/ok-to-test
/approve

I like it, but I'm wondering why the status change for the apiserver does not trigger a reconcile in itself. The controller is watching cluster resources already, so it should be triggered the moment the API server becomes ready.

multi-io:
> /ok-to-test
> /approve
> 
> I like it, but I'm wondering why the status change for the apiserver does not trigger a reconcile in itself. The controller is watching cluster resources already, so it should be triggered the moment the API server becomes ready.

I think that's because of https://github.com/kubermatic/kubermatic/pull/4773

xrstf:
Ah of course, haven't read the controller that far yet.

/lgtm

PR #5270, opened 2020-03-30
Add rhel in kubermatic

### LINKED ISSUE #5206: Adding RHEL as an os option in Kubermatic
**User Story**
Since Red Hat Enterprise Linux(rhel) is now added in all cloud provider which supports rhel, the next step would be to integrate it in Kubermatic.

**Acceptance criteria**
- [ ] RHEL is available as an os option for AWS 
- [ ] RHEL is available as an os option for Azure
- [ ] RHEL is available as an os option for GCE
- [ ] RHEL is available as an os option for Kubevirt
- [ ] RHEL is available as an os option for Openstack
- [ ] RHEL is available as an os option for vSphere

moadqassem:
Adding rhel as an OS option in Kubermatic, in addition to fixing the dynamic user ssh keys syncing. Previously, we hard coded the users which would exist in the home dir based on the os, where in this PR we simply check which are the users that exist in the home dir and copy the user ssh keys to those `.ssh` folder.

Fixes kubermatic/kubermatic#5206

RHEL is available as an os option for AWS
RHEL is available as an os option for Azure
RHEL is available as an os option for GCE
RHEL is available as an os option for Kubevirt
RHEL is available as an os option for Openstack
RHEL is available as an os option for vSphere

**Documentation**:
The `MachineDeployment` specs are: 
RHEL_SUBSCRIPTION_MANAGER_USER string 
RHEL_SUBSCRIPTION_MANAGER_PASSWORD string
REDHAT_SUBSCRIPTIONS_OFFLINE_TOKEN string

Yes 

```release-note
None
```


api/cmd/conformance-tests/main.go
```diff
@@ -494,6 +497,11 @@ func getScenarios(opts Opts, log *zap.SugaredLogger) []testScenario {
 				filteredScenarios = append(filteredScenarios, scenario)
 			}
 		}
+		if osspec.Rhel != nil {
+			if !opts.excludeSelector.Distributions[providerconfig.OperatingSystemRHEL] {
+				filteredScenarios = append(filteredScenarios, scenario)
+			}
+		}
```
xrstf, line 504: In the future, this code can really use some `&&` for all these if statements.

api/cmd/user-ssh-keys-agent/main.go
```diff
@@ -138,3 +144,21 @@ func createFileIfNotExist(path string, uid, gid int) error {
 
 	return nil
 }
+
+func availableHomeUsers() ([]string, error) {
+	files, err := ioutil.ReadDir("/home")
```
xrstf, line 149: I just want to mention it once, but not all users have a home directory and not all home directories are in `/home`.

api/pkg/machine/convert.go
```diff
@@ -69,6 +71,18 @@ func GetAPIV1OperatingSystemSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv
 		operatingSystemSpec.SLES = &apiv1.SLESSpec{
 			DistUpgradeOnBoot: config.DistUpgradeOnBoot,
 		}
+
+	case providerconfig.OperatingSystemRHEL:
+		config := &rhel.Config{}
+		if err := json.Unmarshal(decodedProviderSpec.OperatingSystemSpec.Raw, &config); err != nil {
+			return nil, fmt.Errorf("failed to parse sles config: %v", err)
```
xrstf: Actual change request: `RHEL` instead of `sles`.

api/cmd/conformance-tests/main.go
moadqassem: Just following what is already there ;)

api/cmd/user-ssh-keys-agent/main.go
moadqassem: Well, even in the previous version of this agent, we were relying on what's already in the `/home` dir. This solution was only to dynamically do what's already done.

zreigz:
looks really good, thanks for your work :+1:

PR #5274, opened 2020-03-31
migrate the RBAC controller to controller-runtime, part 1

kdomanski:
```release-note
NONE
```

"Part 1", because it is intertwined with some code from the resource controller, which also needs migration. After the resource controller has been migrated as well, the cleanup of legacy cluster providers will be possible.


api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -186,15 +178,22 @@ func (c *projectController) ensureClusterRBACRoleBindingForResources(projectName
 			}
 
 			if projectResource.destination == destinationSeed {
-				for _, seedClusterProvider := range c.seedClusterProviders {
-					seedClusterRESTClient := seedClusterProvider.kubeClient
-					err := ensureClusterRBACRoleBindingForResource(seedClusterRESTClient, groupName, projectResource.gvr.Resource, seedClusterProvider.kubeInformerProvider.KubeInformerFactoryFor(metav1.NamespaceAll).Rbac().V1().ClusterRoleBindings().Lister())
+				for _, seedClusterRESTClient := range c.seedClientMap {
+					err := ensureClusterRBACRoleBindingForResource(
```
zreigz, line 182: you can also add `return` statement if err != nil

api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -153,15 +146,14 @@ func (c *projectController) ensureClusterRBACRoleForResources() error {
...
-					err := ensureClusterRBACRoleForResource(seedClusterRESTClient, groupPrefix, projectResource.gvr.Resource, projectResource.kind, seedClusterProvider.kubeInformerProvider.KubeInformerFactoryFor(metav1.NamespaceAll).Rbac().V1().ClusterRoles().Lister())
+				for _, seedClusterRESTClient := range c.seedClientMap {
+					err := ensureClusterRBACRoleForResource(c.ctx, seedClusterRESTClient, groupPrefix, projectResource.gvr.Resource, projectResource.kind)
 					if err != nil {
 						return err
 					}
 				}
 			} else {
-				err := ensureClusterRBACRoleForResource(c.masterClusterProvider.kubeClient, groupPrefix, projectResource.gvr.Resource, projectResource.kind, c.masterClusterProvider.kubeInformerProvider.KubeInformerFactoryFor(metav1.NamespaceAll).Rbac().V1().ClusterRoles().Lister())
+				err := ensureClusterRBACRoleForResource(c.ctx, c.client, groupPrefix, projectResource.gvr.Resource, projectResource.kind)
```
zreigz, line 156: you can also add return statement if err != nil
```
if err := ensureClusterRBACRoleForResource(c.ctx, c.client, groupPrefix, projectResource.gvr.Resource, projectResource.kind); err != nil {
return err
}
```

api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -285,25 +277,26 @@ func (c *projectController) ensureRBACRoleForResources() error {
 		for _, groupPrefix := range AllGroupsPrefixes {
 
 			if projectResource.destination == destinationSeed {
-				for _, seedClusterProvider := range c.seedClusterProviders {
-					seedClusterRESTClient := seedClusterProvider.kubeClient
-					err := ensureRBACRoleForResource(seedClusterRESTClient,
+				for _, seedClusterRESTClient := range c.seedClientMap {
+					err := ensureRBACRoleForResource(
```
zreigz, line 281: you can also add return statement if err != nil

api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -359,23 +347,24 @@ func (c *projectController) ensureRBACRoleBindingForResources(projectName string
 			}
 
 			if projectResource.destination == destinationSeed {
-				for _, seedClusterProvider := range c.seedClusterProviders {
-					seedClusterRESTClient := seedClusterProvider.kubeClient
-					err := ensureRBACRoleBindingForResource(seedClusterRESTClient,
+				for _, seedClusterRESTClient := range c.seedClientMap {
+					err := ensureRBACRoleBindingForResource(
```
zreigz, line 351: you can also add return statement if err != nil

api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -359,23 +347,24 @@ func (c *projectController) ensureRBACRoleBindingForResources(projectName string
...
-						projectResource.namespace,
-						seedClusterProvider.kubeInformerProvider.KubeInformerFactoryFor(projectResource.namespace).Rbac().V1().RoleBindings().Lister().RoleBindings(projectResource.namespace))
+						projectResource.namespace)
 					if err != nil {
 						return err
 					}
 				}
 			} else {
-				err := ensureRBACRoleBindingForResource(c.masterClusterProvider.kubeClient,
+				err := ensureRBACRoleBindingForResource(
```
zreigz, line 362: you can also add return statement if err != nil

api/pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -467,15 +447,14 @@ func (c *projectController) ensureProjectCleanup(project *kubermaticv1.Project)
...
-					err := cleanUpClusterRBACRoleBindingFor(seedClusterRESTClient, groupName, projectResource.gvr.Resource)
+				for _, seedClient := range c.seedClientMap {
+					err := cleanUpClusterRBACRoleBindingFor(c.ctx, seedClient, groupName, projectResource.gvr.Resource)
 					if err != nil {
 						return err
 					}
 				}
 			} else {
-				err := cleanUpClusterRBACRoleBindingFor(c.masterClusterProvider.kubeClient, groupName, projectResource.gvr.Resource)
+				err := cleanUpClusterRBACRoleBindingFor(c.ctx, c.client, groupName, projectResource.gvr.Resource)
```
zreigz, line 457: you can also add return statement if err != nil

zreigz (changes requested):
some small comments. Looks good and seems to work fine. I've tested it locally and I didn't find any problem so far

api/pkg/controller/master-controller-manager/rbac/sync_project.go
kdomanski: Then it looks like this:
image
And the logic flow is difficult to see. Same with the remaining cases.

One could reformat the `ensureClusterRBACRoleBindingForResource` to have all arguments in one line, but in most other cases that's impractical, because the line would be too long.

PR #5275, opened 2020-04-01
Turn addon.TemplateData into a stable, documented interface

xrstf:
We have always offered templating in addon manifests (or at least for quite some time), yet we have never officially documented the available variables. This made the "templating interface" for addon authors needlessly hard to figure out.

This PR fixes that by generating a Go snippet for the struct that we actually inject into the Go templates. This Go snippet is generated by recursively finding all structs referenced in the `addon.TemplateData` struct.

As our CRDs are not our API and hence never considered stable, this PR changes `TemplateData` to have its very own data structure with a `NewTemplateData` mapping helper. This is similar to how we handle CRDs in our REST API.

https://github.com/loodse/docs/pull/246 contains the upgrade notes for this change.

```release-note
Action required: Addon manifest templating is now a stable API, but different to the old implicit data. Custom addons might need to be adjusted.
```

### DIFF
--- a/api/cmd/addon-godoc-generator/main.go
+++ b/api/cmd/addon-godoc-generator/main.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"go/ast"
+	"go/format"
+	"go/token"
+	"log"
+	"reflect"
+	"strings"
+
+	"github.com/kubermatic/kubermatic/api/pkg/addon"
+	"golang.org/x/tools/go/packages"
+)
+
+var packageCache = map[string]*packages.Package{}
+
+func main() {
+	t := addon.TemplateData{}
+	snippets := generateDocumentation([]string{}, reflect.ValueOf(t))
+
+	fmt.Println(strings.Join(snippets, "\n\n"))
+}
+
+func generateDocumentation(docs []string, v reflect.Value) []string {
+	t := v.Type()
+
+	if v.Kind() == reflect.Struct {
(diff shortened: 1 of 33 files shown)


api/pkg/addon/template.go
```diff
@@ -32,16 +40,120 @@ func txtFuncMap(overwriteRegistry string) template.FuncMap {
...
+	SeedName       string
+	DatacenterName string
+	Cluster        ClusterData
+	Credentials    Credentials
+	Variables      map[string]interface{}
+}
+
+func NewTemplateData(
+	cluster *kubermaticv1.Cluster,
+	addon *kubermaticv1.Addon,
```
moadqassem: After you removed `Addon` from the `TemplateData` struct(https://github.com/kubermatic/kubermatic/pull/5275/files#diff-b640da135f39d8bda15a0be1eb3dc8beL36) this parameter is not used anymore so we can get rid of it.

api/cmd/addon-godoc-generator/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"errors"
+	"fmt"
+	"go/ast"
+	"go/format"
+	"go/token"
+	"log"
+	"reflect"
+	"strings"
+
+	"github.com/kubermatic/kubermatic/api/pkg/addon"
```
moadqassem: Please regroup the imports:
```
	"github.com/kubermatic/kubermatic/api/pkg/addon"
	
        "golang.org/x/tools/go/packages"
```

PR #5304, opened 2020-04-14
Add proxy support to the Kubermatic Operator

### LINKED ISSUE #4783: Kubermatic Operator: Add Proxy Settings
**User Story**
The Helm chart allows to configure HTTP proxies. This should be possible via the Operator as well.

This is part of #4723.

**Acceptance criteria**
* [ ] All proxy-related settings from the chart are possible via KubermaticConfiguration.

xrstf:
This PR gives the operator the options to configure Kubermatic to use a proxy for HTTP/HTTPS communication. The operator itself naturally does not use the proxy settings (though it would be fun if the operator reconciled itself :D ), but since it does not talk to the outside world, it's okay. Otherwise the proper place to put the proxy settings would be the Kubermatic Installer.

The idea for the default NoProxy value and appending the user-supplied values is taken from KubeOne. These guys usually know what they're doing.

Disclaimer: I have no way to test this.

Fixes #4783

**Documentation**:
This is a new feature for the unreleased operator, the generated docs already show how to configure the proxy.

```release-note
NONE
```


docs/zz_generated.kubermaticConfiguration.yaml
```diff
@@ -105,6 +105,23 @@ spec:
...
+    # HTTPS is the full URL to the proxy to use for encrypted HTTPS
+    # connections, e.g. "http://secureinternalproxy.example.com:8080".
+    https: ""
+    # NoProxy is a comma-separated list of hostnames / network masks
+    # for which no proxy shall be used. If you make use of proxies,
+    # this list should contain all local and cluster-internal domains
+    # and networks, e.g. "10.0.0.0/8,172.16.0.0/12,192.168.0.0/16,mydomain".
+    # The operator will always prepend the following elements to this
+    # list if proxying is configured (i.e. HTTP/HTTPS are not empty):
+    # "127.0.0.1/8", "localhost", ".local", "kubernetes", ".default", ".svc"
```
toschneck: some code in the machinecontroller query `webhook. ....cluster.local.`.  This not match somehow to `.local` so we would need add `.local.` as well

api/pkg/controller/operator/common/defaults.go
```diff
@@ -39,6 +39,11 @@ const (
 	DefaultVPAAdmissionControllerDockerRepository = "gcr.io/google_containers/vpa-admission-controller"
 	DefaultNodeportProxyDockerRepository          = "quay.io/kubermatic/nodeport-proxy"
 	DefaultEnvoyDockerRepository                  = "docker.io/envoyproxy/envoy-alpine"
+
+	// DefaultNoProxy is a set of domains/networks that should never be
+	// routed through a proxy. All user-supplied values are appended to
+	// this constant.
+	DefaultNoProxy = "127.0.0.1/8,localhost,.local,kubernetes,.default,.svc"
```
toschneck: some code in the machinecontroller query `webhook. ....cluster.local.`.  This not match somehow to `.local` so we would need add `.local.` as well

toschneck (changes requested):
Improve of some defaults needed

api/pkg/controller/operator/common/defaults.go
xrstf: Fixed.

toschneck (approved):

PR #5312, opened 2020-04-16
Adding azure multiple az support

moadqassem:
Adding the azure multiple az support to kubermatic

Fixes #kubermatic/kubermatic#4473

There should be another ticket to add this change to the UI

No 
```release-note
None
```


maciaszczykm:
>Does this PR introduce a user-facing change?:
No

I think it does. Both API and UI seem to be user-facing in my opinion.

api/pkg/handler/v1/provider/azure.go
```diff
@@ -248,6 +248,75 @@ func azureSize(ctx context.Context, subscriptionID, clientID, clientSecret, tena
 	return sizeList, nil
 }
 
+func AzureAvailabilityZonesEndpoint(presetsProvider provider.PresetProvider, userInfoGetter provider.UserInfoGetter) endpoint.Endpoint {
```
maciaszczykm, line 251: Is this the same as `AvailabilitySet` from the `AzureCloudSpec`? There is no `AvailabilityZone` in the Azure yet, so if this is not the same then you should add this. If this is the same then please align the name to be the same.

moadqassem: `AvailabilityZones` are kinda different than the  `AvailabilitySet`. `AvailabilityZones` or as it is mentioned on the azure vm `Zones` is  on a `MachineDeployment` scope, which means it should be added to probably here https://github.com/kubermatic/kubermatic/blob/master/api/pkg/api/v1/types.go#L1062. However, the machine-controller vendor should be updated in order to get the `Zones`field in the `MachineDeployment` object https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/azure/types/types.go#L38

moadqassem:
> > Does this PR introduce a user-facing change?:
> > No
> 
> I think it does. Both API and UI seem to be user-facing in my opinion.

I see, I wasn't sure if the users, use our api explicitly instead of using the UI. About the API yeah it has changed, the UI not yet. However, I will change this to `Yes`

api/pkg/handler/v1/provider/azure.go
maciaszczykm: Yeah, true it's on the node level, not cluster.

You've used the same link twice but I think I got the point. We should update vendor and also make the change in the node deployment creation to be able to set `AvailabilityZones` for Azure. As there is no point of merging this change without the rest I think we should wait. Could you take care of that? Do you know the process here?

moadqassem: I have updated the link. Regarding updating the vendor, sure, I can take care of that ;)

PR #5315, opened 2020-04-17
Add properties required for RHEL

maciaszczykm:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Related to https://github.com/kubermatic/dashboard-v2/pull/2190.

@moadqassem can you confirm if these 2 will be enough to create clusters and nodes with RHEL?

**Documentation**:

```release-note
Added custom image property to GCP clusters.
Added image ID property to Azure clusters.
```


api/pkg/provider/kubernetes/preset.go
```diff
@@ -313,6 +314,7 @@ func (m *PresetsProvider) setAzureCredentials(userInfo *provider.UserInfo, prese
 	cloud.Azure.SecurityGroup = credentials.SecurityGroup
 	cloud.Azure.SubnetName = credentials.SubnetName
 	cloud.Azure.VNetName = credentials.VNetName
+	// TODO ImageID?
```
maciaszczykm: @zreigz @moadqassem do you think that we should include it also in the presets?

api/pkg/provider/kubernetes/preset.go
```diff
@@ -217,6 +217,7 @@ func (m *PresetsProvider) setGCPCredentials(userInfo *provider.UserInfo, presetN
 	cloud.GCP.ServiceAccount = credentials.ServiceAccount
 	cloud.GCP.Network = credentials.Network
 	cloud.GCP.Subnetwork = credentials.Subnetwork
+	// TODO CustomImage?
```
maciaszczykm: @zreigz @moadqassem do you think that we should include it also in the presets?

zreigz: depends if this parameter is optional or needed to create a cluster. If someone will pick predefined preset then this parameter will be always used together with others. I don't know if it's true or not. cc @moadqassem

moadqassem: But this custom image is not a cluster level, it is a node deployment level. Thus I am not quite sure about having a preset for it. In addition, this field can be different for two different node deployment(in theory of course since we have only one rhel custom image for gcp atm).

maciaszczykm:
PTAL @moadqassem

PR #5318, opened 2020-04-20
Configurable update window

kdomanski:
Copy of #4972 which doesn't work properly anymore since Bastian's fork is deleted.

```release-note
Added configurable time window for coreos-operator node reboots
```


zreigz:
please regenerate client `/hack/gen-api-client.sh`

kdomanski:
@zreigz done

PR #5322, opened 2020-04-20
openstack cloud controller manager: use unique cluster name

multi-io:

We noticed that we’re not setting the clusterName parameter for the openstack cloud controller manager, which means it’ll default to “kubernetes”. That will lead to non-unique openstack load balancer names at https://github.com/kubernetes/cloud-provider-openstack/blob/8c02748fab095675514861628ea57317bb36c263/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L521

This will result in a name clash if a user has more than one cluster in the same openstack tenant and then creates a loadbalancer service with the same name in the same namespace in all those clusters. Those clusters would then all try to use the same Openstack loadbalancer. This would definitely be the case for quite a few of our customers, hence this PR.

This definitely means though that for for users that use openstack and have already deployed 2.13 and created loadbalancer services, new openstack loadbalancers with new external IPs would be created for those services because of the name change. (it won’t be a problem for us because we haven’t rolled out 2.13 to production yet)

```release-note
Set clusterName parameter in Openstack cloud controller to the Kubermatic cluster id to ensure unique names of created cloud resources like load balancers.
```

### DIFF
--- a/api/pkg/resources/cloudcontroller/openstack.go
+++ b/api/pkg/resources/cloudcontroller/openstack.go
@@ -149,6 +149,7 @@ func getOSFlags(data *resources.TemplateData) []string {
 		"--v=1",
 		"--cloud-config=/etc/kubernetes/cloud/config",
 		"--cloud-provider=openstack",
+		"--cluster-name=" + data.Cluster().Name,
 	}
 	return flags
 }


xrstf:
@multi-io Does this cause migration issues (new loadbalancers with new IPs) when updating Kubermatic?

multi-io:
> @multi-io Does this cause migration issues (new loadbalancers with new IPs) when updating Kubermatic?

Yeah, it does. If a user already created an LB service with 2.13 (which is the first version to roll out cloud-controller-manager, right?), then if this patch is rolled out, the name of the openstack LB created for the service will change, so K8s will no longer find the existing (previously created) openstack LB and thus create a new one with a new floating IP. So first of all the old previously created openstack LB will no longer be managed and will have to be deleted manually, and second of all the new LB will have a different floating IP, which the user would have to account for, e.g. by modifying DNS records etc. So this should be merged with care, if at all. OTOH, without this patch, the created LB name is kube_service_kubernetes_`<lb service namespace>_<lb service name>`, which is not unique if a user has two clusters in the same openstack tenant and then creates an lb service with the same name in the same namespace in both clusters.

kdomanski:
Is there no way to update the LB names so that they are not recreated?

Issue #3372, opened 2019-04-29
Move DNS handling into user cluster controller


mrIncompetent:
**User Story**
All DNS specific resources should not be deployed using addons.
Addons should be optional - customer specific - resources.

DNS is a core component of a Kubernetes cluster.
With https://github.com/kubermatic/kubermatic/pull/3369 we now split logic across an addon & the controllers - Giving more potential for errors.

To avoid having logic split between addons & controllers + to ease [CIDR changes](https://docs.kubermatic.io/operation/control-plane/increase-service-pod-cidr/)
all DNS specific resources should be deployed by the UCCM (Potentially a own controller)

**Acceptance criteria**
- [ ] The DNS & nodelocal-dns-cache addons got removed
- [ ] CoreDNS & nodelocal-dns-cache are now being deployed by a own controller

PR #5341, opened 2020-04-27
Fix possible service IP collision for kube-dns

### LINKED ISSUE #5232: Ensure kube-dns svc IP is not used by any other service
**User Story**
When a cluster is deployed, on the user side we have couple of services that are created. One of them is kube-dns svc. This one is using a hardcoded IP.
Right now we can hit race conditions where another service could get allocated the "hardcoded" IP of kube-dns. In this scenario, this leads to whole user-cluster failure as DNS resolution can't work properly.
The probability of this happening is still very slim (4094 usable addresses in the svc network range).
 
**Acceptance criteria**
kube-dns IP is not attributed to any other services in the user cluster.

moelsayed:

Fixes #5232

It seems the first 3 services created in `kube-system` are `kubeapi`, `dashboard-metrics-scraper` and `kube-dns`. So, `dashboard-metrics-scraper` is only one likely to beat `kube-dns` to that particular IP.
This best way I figured to fix this is to also set the IP for `dashboard-metrics-scraper` but make sure it's not conflicting with the `kube-dns` IP. Theoretically, there is a possibility that this might conflict with an existing service IP, but it's very unlikely since this is the second service created on the user cluster.
**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/controller/master-controller-manager/rbac/mapper.go
+++ b/api/pkg/controller/master-controller-manager/rbac/mapper.go
@@ -20,12 +20,16 @@ const (
 	// ViewerGroupNamePrefix represents viewers group prefix
 	ViewerGroupNamePrefix = "viewers"
 
+	// WatcherGroupNamePrefix represents watchers group prefix, this doesn't need to be added to AllGroupsPrefixes
+	WatcherGroupNamePrefix = "watchers"
+
 	// RBACResourcesNamePrefix represents kubermatic group prefix
 	RBACResourcesNamePrefix = "kubermatic"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


moelsayed:
FWIW, this seems like an overkill considering the probability of the issue. An alternative, solution is to create a place-holder `kube-dns` service to allocate the IP until the kube-dns addon is applied later during the deployment.

moelsayed:
/hold

I am planning to work on #3372 next, it should clean this nicely and make this PR unnecessary.

moelsayed:
Closing in favor of #5370

PR #5325, opened 2020-04-21
Add Support for Kubernetes 1.18

xrstf:
**What this PR does / why we need it**:
This does what it says on the tin, i.e. add support for Kubernetes 1.18. For reviewing I strongly recommend to go commit-by-commit, so you can skip aa6dc60f34dcfc ("rebuild test fixtures for 1.15-1.18").

**Does this PR introduce a user-facing change?**:
```release-note
Add support for Kubernetes 1.18, drop support for Kubernetes < 1.15.
```

PR #5344, opened 2020-04-28
[release/v2.14] Add Support for Kubernetes 1.18 (#5325)

xrstf:
**What this PR does / why we need it**:
This backports the Kubernetes 1.18 support (#5325) into the 2.14 release branch.

**Does this PR introduce a user-facing change?**:
```release-note
Add support for Kubernetes 1.18, drop support for Kubernetes < 1.15.
```

PR #5345, opened 2020-04-28
restore f8598963e (make openshift tests optional)

xrstf:
The change from #5335 got somehow lost in #5325, even though I thought I carefully rebased and was aware of the change. In contrast to the original PR, I chose to still at least run the tests by default when the openshift addons changed.

```release-note
NONE
```


xrstf:
/cherry-pick release/v2.14

Because #5344 will apply the broken prow.yaml change to the 2.14 release branch.

PR #5346, opened 2020-04-28
push basic metrics during e2e tests

xrstf:
This is a first attempt to gain more insight into our tests and their runtimes. The conformance-tester is extended by a few metrics that are pushed to a shared pushgateway running in the CI cluster.

```release-note
NONE
```

### DIFF
--- a/api/pkg/test/e2e/api/oidc.go
+++ b/api/pkg/test/e2e/api/oidc.go
@@ -52,8 +52,14 @@ func retrieveMasterToken() (string, error) {
 	return masterToken, nil
 }
 
+var adminMasterToken = ""
+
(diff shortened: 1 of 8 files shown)


api/pkg/test/e2e/api/oidc.go
```diff
@@ -52,8 +52,14 @@ func retrieveMasterToken() (string, error) {
 	return masterToken, nil
 }
 
+var adminMasterToken = ""
```
xrstf, line 55: We do this runtime caching for the regular access token already, so the test logs are not constantly spammed with very spammy login logs. However the cache was missing for the admin token.

PR #5368, opened 2020-05-05
add Flatcar Linux to the API

kdomanski:
```release-note
Added Flatcar Linux
```


api/cmd/conformance-tests/scenarios_aws.go
```diff
@@ -36,6 +36,16 @@ func getAWSScenarios(versions []*semver.Semver) []testScenario {
 				},
 			},
 		})
+		// CoreOS
```
moadqassem: Should be Flatcar instead of `CoreOS`

api/cmd/kubermatic-api/swagger.json
```diff
@@ -9361,6 +9361,18 @@
       },
       "x-go-package": "github.com/kubermatic/kubermatic/api/pkg/crd/kubermatic/v1"
     },
+    "FlatcarSpec": {
+      "description": "FlatcarSpec contains Flatcar Linux specific settings",
+      "type": "object",
+      "properties": {
+        "disableAutoUpdate": {
+          "description": "disable container linux auto-update feature",
```
moadqassem: instead of `container linux` should be `flatcar` ;)

api/pkg/api/v1/types.go
```diff
@@ -985,6 +985,13 @@ type ContainerLinuxSpec struct {
 	DisableAutoUpdate bool `json:"disableAutoUpdate"`
 }
 
+// FlatcarSpec contains Flatcar Linux specific settings
+// swagger:model FlatcarSpec
+type FlatcarSpec struct {
+	// disable container linux auto-update feature
```
moadqassem: Same goes here `flatcar` instead of `container linux`

api/cmd/conformance-tests/runner.go
```diff
@@ -1154,6 +1154,9 @@ func (r *testRunner) getGinkgoRuns(
 		case osSpec.ContainerLinux != nil:
 			args = append(args, "--node-os-distro=coreos")
 			env = append(env, "KUBE_SSH_USER=core")
+		case osSpec.Flatcar != nil:
```
moadqassem, line 1157: Please add the flatcar os to the `getOSNameFromSpec` method:
https://github.com/kubermatic/kubermatic/blob/master/api/cmd/conformance-tests/os.go#L7

.prow.yaml
```diff
@@ -229,6 +229,8 @@ presubmits:
         env:
         - name: VERSIONS_TO_TEST
           value: "v1.15.11"
+        - name: EXCLUDE_DISTRIBUTIONS
```
moadqassem, line 232: Actually this is not needed as we have this env var `EXCLUDE_DISTRIBUTIONS `: https://github.com/kubermatic/kubermatic/blob/master/api/hack/ci/ci-run-conformance-tester.sh#L10

moadqassem: Also would be cool to add those in our offline tests for consistency :) 
https://github.com/kubermatic/kubermatic/blob/master/api/hack/ci/ci-run-offline-test.sh#L256

api/pkg/api/v1/types.go
```diff
@@ -985,6 +985,13 @@ type ContainerLinuxSpec struct {
 	DisableAutoUpdate bool `json:"disableAutoUpdate"`
 }
 
+// FlatcarSpec contains Flatcar Linux specific settings
+// swagger:model FlatcarSpec
+type FlatcarSpec struct {
```
moadqassem, line 990: This also should be added to the api for the `node_labeler`: 
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/user-cluster-controller-manager/node-labeler/api/api.go#L20

moadqassem (changes requested):

moadqassem:
@kdomanski Can you please add the flatcar ssh username for our supported cloud provider: 
https://github.com/kubermatic/kubermatic/blob/master/api/pkg/machine/util.go#L10

.prow.yaml
kdomanski: Implicit is better than explicit and I don't really understand the logic of skipping the setting in a few particular cases without comment because 2 scripts down there is a default value. If anything, we should remove the default values from the script to avoid surprises.

api/cmd/conformance-tests/scenarios_aws.go
kdomanski: Fixed

api/cmd/kubermatic-api/swagger.json
kdomanski: fixed

api/pkg/api/v1/types.go
kdomanski: fixed

api/cmd/conformance-tests/runner.go
kdomanski: done

api/pkg/api/v1/types.go
kdomanski: done

api/pkg/controller/user-cluster-controller-manager/node-labeler/api/api.go
```diff
@@ -18,6 +18,9 @@ const (
 
 	// RHELLabelValue is the value of the label for RHEL
 	RHELLabelValue = "rhel"
+
+	// FlatcarLabelValue is the value of the label for RHEL
```
moadqassem: Should be `FlatcarLabelValue is the value of the label for Flatcar` not `RHEL` ;)

kdomanski: Fixed

kdomanski:
/hold
@zreigz could someone from your team take a look as well?

PR #5370, opened 2020-05-06
Switch CoreDNS from addon to a user-cluster-controller-manager resource

moelsayed:

Fixes-partially #3372

**Documentation**:

```release-note
Kubedns is no longer a manifest addon. 
```


api/pkg/test/e2e/api/runner.go
```diff
@@ -678,7 +678,7 @@ func cleanUpProject(id string, attempts int) func(t *testing.T) {
 			if err != nil {
 				break
 			}
-			time.Sleep(3 * time.Second)
+			time.Sleep(6 * time.Second)
```
xrstf: It might have been better to increase the number of attempts rather than the interval.

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -521,6 +520,10 @@ func (r *Reconciler) ensureResourcesCreatedConditionIsSet(ctx context.Context, a
 func (r *Reconciler) cleanupManifests(ctx context.Context, log *zap.SugaredLogger, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) error {
 	kubeconfigFilename, manifestFilename, done, err := r.setupManifestInteraction(log, addon, cluster)
 	if err != nil {
+		if strings.Contains(err.Error(), "no such file or directory") { // if the manifest is already deleted, that's ok
```
xrstf, line 524: Please return a dedicated error instead of asserting the string form on an error.

api/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
```diff
@@ -0,0 +1,201 @@
...
+			dep.Namespace = metav1.NamespaceSystem
+			dep.Labels = resources.BaseAppLabels(resources.CoreDNSDeploymentName, nil)
+
+			dep.Spec.Replicas = resources.Int32(2)
+			// The Selector is immutable, so we don't change it if it's set. This happens in upgrade cases
+			// where coredns is switched from a manifest based addon to a user-cluster-controller-manager resource
+			if dep.Spec.Selector == nil {
+				dep.Spec.Selector = &metav1.LabelSelector{
+					MatchLabels: resources.BaseAppLabels(resources.CoreDNSDeploymentName,
+						map[string]string{"k8s-app": "kube-dns", "kubermatic-addon": "dns"}),
```
xrstf: Please use the modern recommended labels https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ and do not use the word "kubermatic" anywhere a user could see it. Nothing inside a user cluster must hint at Kubermatic.

api/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/service.go
```diff
@@ -0,0 +1,37 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ServiceCreator creates the service for the CoreDNS
+func ServiceCreator(dnsClusterIP string) reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		labels := map[string]string{
+			"kubernetes.io/cluster-service": "true",
+			"kubernetes.io/name":            "KubeDNS",
```
xrstf: `app.kubernetes.io/name` is more consistent :)

api/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/serviceaccount.go
```diff
@@ -0,0 +1,21 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ServiceAccountCreator creates the service account for CoreDNS
+func ServiceAccountCreator() reconciling.NamedServiceAccountCreatorGetter {
+	return func() (string, reconciling.ServiceAccountCreator) {
+		return resources.CoreDNSServiceAccountName, func(sa *corev1.ServiceAccount) (*corev1.ServiceAccount, error) {
+			sa.Name = resources.CoreDNSServiceAccountName
+			sa.Namespace = metav1.NamespaceSystem
```
xrstf: FYI, both lines are redundant. Name and namespace are always set by the reconciling library.

api/pkg/resources/resources.go
```diff
@@ -474,6 +474,16 @@ const (
...
+const (
+	CoreDNSClusterRoleName         = "system:coredns"
+	CoreDNSClusterRoleBindingName  = "system:coredns"
+	CoreDNSServiceAccountName      = "coredns"
+	CoreDNSServiceName             = "kube-dns"
+	CoreDNSConfigMapName           = "coredns"
+	CoreDNSDeploymentName          = "coredns"
+	CoreDNSPodDisruptionBudgetName = "coredns"
+)
+
```
xrstf, line 486: Why not have these in the new `coredns` package?

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
moadqassem: maybe he can use os.IsNotExist instead if a dedicated error ;)

xrstf (changes requested):
Some minor things.

api/hack/run-userclustercontroller.sh
```diff
@@ -16,7 +16,11 @@ CLUSTER_NAME="$(echo $NAMESPACE|sed 's/cluster-//')"
 CLUSTER_RAW="$(kubectl get cluster $CLUSTER_NAME -o json)"
 CLUSTER_URL="$(echo $CLUSTER_RAW|jq -r '.address.url')"
 # You must set the email address of the cluster owner, otherwise the controller will fail to start
-OWNER_EMAIL=""
+if [ -z "${OWNER_EMAIL:-}" ]
```
moadqassem, line 19: There is already a check for this in the main.go file:
https://github.com/kubermatic/kubermatic/blob/master/api/cmd/user-cluster-controller-manager/main.go#L100-L102

api/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
moelsayed: We actually ensure that particular "kubermatic" label on all addon resources: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/controller/seed-controller-manager/addon/addon_controller.go#L354, do we want to stop doing that everywhere?

api/pkg/resources/resources.go
moelsayed: I would prefer that as well, but all the resources-related const's are in this file. Moved them here for consistency.

api/hack/run-userclustercontroller.sh
moelsayed: That's true, however. It's not currently possible to pass it to the script through an environment variable, you actually need to edit the script to set the value. So, I added the check to inform the user with the environment variable name.

api/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
xrstf: That is genuinely surprising me, cause I can remember the talks about whitelabelling. But since you didn't make it worse, 👍  ;-)

PR #5371, opened 2020-05-06
prepare sync-charts script for syncing master branch

xrstf:
We often have the case where we want to install the master branch, just because it has some new features or fixes that are important in very special environments. This PR makes it so that the sync-charts script can handle syncing the master branch as well (another PR will then change the Prow postsubmit to make it run for every branch).

```release-note
NONE
```

### DIFF
--- a/api/hack/ci/ci-sync-charts.sh
+++ b/api/hack/ci/ci-sync-charts.sh
@@ -7,14 +7,24 @@ source $(dirname $0)/../lib.sh
 cd $(dirname $0)/../../..
 git fetch --tags
 
-# we only synchronize charts for tags
-if ! git tag -l | grep -q "^$PULL_BASE_REF\$"; then
-  echo "Base ref $PULL_BASE_REF is not a tag! Exiting..."
+# we only synchronize charts for the master branch and tagged versions
+if [ "$PULL_BASE_REF" != "master" ] && ! (git tag -l | grep -q "^$PULL_BASE_REF\$"); then
+  echo "Base ref $PULL_BASE_REF is neither master nor a tag! Exiting..."
   exit 1
 fi
 
+# The dashboard tag should match the Kubermatic tag for tagged releases
+KUBERMATIC_VERSION="$PULL_BASE_REF"
+DASHBOARD_VERSION="$KUBERMATIC_VERSION"
+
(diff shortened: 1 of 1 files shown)


api/hack/ci/ci-sync-charts.sh
```diff
@@ -7,14 +7,24 @@ source $(dirname $0)/../lib.sh
...
+DASHBOARD_VERSION="$KUBERMATIC_VERSION"
+
 # find out which kubermatic-installer branch to synchronize to
-if grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+.*' <<< "$PULL_BASE_REF"; then
+if [ "$PULL_BASE_REF" == "master" ]; then
+  INSTALLER_BRANCH="master"
+  # do not use the actual head hash to avoid spamming the installer
+  # repo for every new commit on master branch
+  KUBERMATIC_VERSION="latest"
+  DASHBOARD_VERSION="master"
```
xrstf, line 26: `latest` is the latest tagged release for the dashboard repo, whereas in Kubermatic `latest` is just the latest non-test build I think.

api/hack/ci/ci-sync-charts.sh
```diff
@@ -7,14 +7,24 @@ source $(dirname $0)/../lib.sh
 cd $(dirname $0)/../../..
 git fetch --tags
 
-# we only synchronize charts for tags
-if ! git tag -l | grep -q "^$PULL_BASE_REF\$"; then
-  echo "Base ref $PULL_BASE_REF is not a tag! Exiting..."
+# we only synchronize charts for the master branch and tagged versions
+if [ "$PULL_BASE_REF" != "master" ] && ! git tag -l | grep -q "^$PULL_BASE_REF\$"; then
```
kdomanski: I think a few brackets would add readability

kdomanski:
/approve

Looks roughly fine, but I'd like more eyes on this.

api/hack/ci/ci-sync-charts.sh
xrstf: I had the second part in `[ ... ]`, but then bash didn't want to parse it anymore... :D I'm open to suggestions.

PR #5375, opened 2020-05-07
Go modules

kron4eg:
This is needed in order to take advantage of the go new errors (introduced since 1.13).

Go version need to be updated to 1.14

```release-note
* migrated to gomods
```


xrstf (changes requested):
Tests must pass before I can review this.

kron4eg:
> Tests must pass before I can review this.

I'm not changing infra repo to pass this PR 🤷‍♂️. When someone will want to do gomods migration, this PR can be used as a reference.

PR #5391, opened 2020-05-11
[EE] Move MultiSeed behaviour into dedicated packages

xrstf:
This PR introduces the distinction between a *Community Edition* (ce) and an *Enterprise Edition* (ee). For this it makes use of Go build tags (`-tags ee`) to load alternative implementations during build time. By default the Kubermatic source code builds a Community Edition (typing `make` will get you a CE).

This PR also makes the multi seed implementation in Kubermatic an Enterprise feature. While most of the codebase still works based on `map[string]Seed`, the Community edition will always only work with a single seed (so `len(seedMap) == 1` everywhere). So the SeedsGetter will always return a 1-element map and the SeedGetter will always only work with a Seed named `kubermatic`. I had to make a decision as to what happens when a user is creating a second Seed resource: Prefer the first? The oldest? The largest? I settled on "In the community edition, the Seed must always be named `kubermatic`."

Restricting seeds in the datacenters.yaml however was pointless. Since the datacenters.yaml is already just a legacy feature, I decided to remove it entirely from the Community Edition.

There is still lots of leftovers (like the `-datacenters` flag, which is 100% useless in the CE) that we need to adjust later. Similarily, we have a bunch of controllers that don't use the seedsGetter/seedGetter, but rather use informers to directly watch Seed resources. These controllers also need to be looked at and maybe adjusted. We could also extend the Seed Validation Webhook to reject any Seed not named "kubermatic".

With regards to the code, this PR is not pretty. To have directory-based licensing, I needed to extract a few things into their own packages and then add even more packages to fix circular dependencies.

The environment variable `KUBERMATIC_EDITION` is used during CI to toggle what version we build. I think we will maybe have distinct jobs for CE and EE, which would just differ in this variable, but I'm not sure yet. For now the build artifacts are EE and therefore compatible with the rest of our infrastructure.

```release-note
NONE
```

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -18,6 +18,7 @@ HAS_DEP:= $(shell command -v dep 2> /dev/null)
 HAS_GIT:= $(shell command -v git 2> /dev/null)
 BUILD_DEST?=_build
 GOTOOLFLAGS?=$(GOBUILDFLAGS) -ldflags '-w $(LDFLAGS)'
+KUBERMATIC_EDITION?=ce
 
 default: all
 
@@ -27,7 +28,7 @@ all: check vendor build test
 build: $(CMD)
 
 $(CMD): download-gocache
-	go build $(GOTOOLFLAGS) -o $(BUILD_DEST)/$@ ./cmd/$@
+	go build -tags "$(KUBERMATIC_EDITION)" $(GOTOOLFLAGS) -o $(BUILD_DEST)/$@ ./cmd/$@
 
 install:
 	go install $(GOTOOLFLAGS) ./cmd/...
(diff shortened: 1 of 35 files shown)


api/Makefile
```diff
@@ -27,7 +28,7 @@ all: check vendor build test
 build: $(CMD)
 
 $(CMD): download-gocache
-	go build $(GOTOOLFLAGS) -o $(BUILD_DEST)/$@ ./cmd/$@
+	go build -tags "$(KUBERMATIC_EDITION)" $(GOTOOLFLAGS) -o $(BUILD_DEST)/$@ ./cmd/$@
```
kdomanski, line 31: Will this still work locally or is it another thing that only works on Prow now?

xrstf: `KUBERMATIC_EDITION` defaults to `ce` in line 21.

xrstf:
I've rebased this, but kept the 1.12 compatibility for easier backporting to 2.14.

xrstf:
Rebased by reverse-cherry-picking, squasing and force-rebasing afterwards. #5410 really got merged at an unfortunate point in time.

PR #5397, opened 2020-05-12
admin list dc per provider endpoint

lsviben:
Adds a datacenter list per provider endpoint for admins which is a step towards managing dynamic datacenters

```release-note
NONE
```


api/pkg/handler/routes_v1.go
```diff
@@ -1240,6 +1244,30 @@ func (r Routing) datacenterHandler() http.Handler {
 	)
 }
 
+// swagger:route GET /api/v1/providers/{provider_name}/dc listDCForProvider
```
zreigz: change for:
```
// swagger:route GET /api/v1/providers/{provider_name}/dc datacenter listDCForProvider
```

zreigz (changes requested):

PR #5399, opened 2020-05-12
Add datastore and datastore cluster configs on cluster level

### LINKED ISSUE #5136: [vSphere] Support for datastore cluster in Kubermatic
**User Story**
In PR https://github.com/kubermatic/machine-controller/pull/671 we added support for datastore clusters in machine controller. Now `MachineDeployments` for vSphere can specify a `datastoreCluster` instead of a `datastore` in the `cloudProviderSpec` as showed in [this](https://github.com/kubermatic/machine-controller/blob/master/examples/vsphere-datastore-cluster-machinedeployment.yaml) example.

The goal of this US is to wire this functionality to Kubermatic. Meaning that the `datastoreCluster` should be supported in the vSphere [datacenter spec](https://github.com/kubermatic/kubermatic/blob/a38fc5dd4f9eca74b2fade7fc17cb06cd2227eef/api/pkg/crd/kubermatic/v1/datacenter.go#L183).
This will allow users to create `NodeDeployment` from the API that include `datastoreCluster`, `datastore` and `datastoreCluster` should be mutually exclusive.

Note that the `datacenter spec` is also used to generate the `cloud config` used by the user cluster control plane, and that such config does not allow to specify a datastore cluster as a default value for volumes provisioning, but only a datastore (see https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#supported-parameters for more details). 
We have two options here:
1. Leave the default datastore empty
2. Pick one of the datastores included in the datastore cluster to be used as default datastore
3. Make the distinction between the datastore to be used as default for dynamic volumes provisioning and the datastore/datastore cluster to be used for VMs placement in vSphere datacenter spec (if we go for this solution we should also consider to do the same for `MachineDeployment` vSphere cloudProviderSpec).

**Acceptance criteria**
Being able to create vSphere `NodeDeployments` (including initial one at cluster creation) specifying a `datastoreCluster` instead of a `datastore`.

moadqassem:
We added support for datastore clusters in machine controller. Now MachineDeployments for vSphere can specify a datastoreCluster instead of a datastore in the cloudProviderSpec. The goal of this PR is to wire this functionality to Kubermatic. Meaning that the datastoreCluster should be supported in the vSphere cluster spec. 

Fixes kubermatic/kubermatic#5136

This PR should also be reflected by adding two new fields in Kubermatic UI where the user should specify whether to use a datastore cluster or simply a specific datastore.

```release-note
Supporting datastore cluster and default datastore in Kubermatic  
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1012,6 +1012,47 @@ presubmits:
           limits:
             memory: 4Gi
 
+  - name: pre-kubermatic-e2e-vsphere-coreos-1.18-datastore-cluster
+    decorate: true
+    optional: true
+    run_if_changed: "pkg/provider/cloud/vsphere"
+    clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
+    labels:
+      preset-vsphere: "true"
+      preset-vault: "true"
+      preset-docker-pull: "true"
+      preset-docker-push: "true"
+      preset-repo-ssh: "true"
+      preset-kubeconfig-ci: "true"
+      preset-kind-volume-mounts: "true"
+    spec:
+      containers:
+        - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.17
+          env:
+            - name: KUBERMATIC_EDITION
+              value: ee
+            - name: VERSIONS_TO_TEST
+              value: "v1.18.6"
+            - name: PROVIDER
+              value: "vsphere"
+            - name: SCENARIO_OPTIONS
+              value: "datastore-cluster"
+            - name: SERVICE_ACCOUNT_KEY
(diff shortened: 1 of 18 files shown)


irozzo-1A:
@moadqassem I think that specifying the datastore/datastore cluster at cluster level instead of datacenter level makes sense, oh the other hand I would not [rename datastore](https://github.com/kubermatic/kubermatic/pull/5399/files#diff-8abe8482a316ffcf190408bdfc231e19R217) to `default_datastore` to avoid backward compatibility/migration issues. WTYT?

irozzo-1A:
/hold
Still working on it

PR #5400, opened 2020-05-13
datacenter management - added get per provider endpoint

lsviben:
Adds a get endpoint for datacenters per provider, which is a step towards dynamic dc management

```release-note
NONE
```


api/pkg/handler/v1/dc/datacenter_test.go
```diff
@@ -171,3 +171,83 @@ func TestDatacenterListForProviderEndpoint(t *testing.T) {
...
+func TestDatacenterGetForProviderEndpoint(t *testing.T) {
+	t.Parallel()
+	testcases := []struct {
+		name                   string
+		provider               string
+		dc                     string
+		expectedResponse       string
+		httpStatus             int
+		existingAPIUser        *apiv1.User
+		existingKubermaticObjs []runtime.Object
```
zreigz: it's not used

zreigz (changes requested):

PR #5418, opened 2020-05-18
[EE] Multi-Seed Cleanup & CLI Flags

xrstf:
This PR builds upon the previous and improves the CE/EE split in the following way:

* The big picture here is: **The CE/EE split has been moved from deeply nested packages (`pkg/provider`) directly into each of the `cmd/...` binaries we have.** This means the master-controller-manager already decides what implementations to compile, not the `provider` package anymore.
So the commands each got a `wrappers_ce.go` and `wrappers_ee.go` that toggle which SeedGetter/SeedsGetter/SeedKubeconfigGetter to load.
* The point above means that we can restore the `pkg/provider` package back to being a normal Go package without any CE/EE wrappers in it. :-) So now `pkg/provider` is simply "the common Community implementation" and nothing else. If you do `pkg/provider.SeedsGetterFactory()`, you will always get the Community Edition implementation, no matter what build tags you have.
* The point above means that we now have two totally distinct provider packages: `pkg/provider` and `pkg/ee/provider`. And since the implementations are separate and are separately instantiated, we can also refactor the Seed(s)Getter to only use the parameters it needs.
So `pkg/provider.SeedsGetterFactory` was previously `SeedsGetterFactory(ctx, client, dcFile, namespace, dynamicDatacenters) (SeedsGetter, error)`.
But now we can remove `dcFile` and `dynamicDatacenters` from the flags, because the Community Edition does not need them. Only the EE implementation still has these parameters.
* This all allows to define the EE CLI flags (`-datacenters` and `-dynamic-datacenters`) in `pkg/ee/cmd/*` packages, so that the Community Edition does not have these flags anymore.

The pattern is the same across all affected binaries:

* kubermatic-api
* master-controller-manager
* seed-controller-manager

And finally, to make it easier to identify CE/EE builds, I introduced a simple log helper that prints a log line telling the user about the application.

```release-note
NONE
```


api/cmd/conformance-tests/main.go
```diff
@@ -409,7 +412,7 @@ func main() {
 		log.Warn("Environment variable `NAMESPACE` was unset, defaulting to `kubermatic`")
 		namespaceName = "kubermatic"
 	}
-	seedGetter, err := provider.SeedGetterFactory(context.Background(), seedClusterClient, seedName, "", namespaceName, true)
+	seedGetter, err := provider.SeedGetterFactory(context.Background(), seedClusterClient, seedName, namespaceName)
```
xrstf, line 415: Here we can see that now the "community implementation" in `pkg.provider` is now really only caring for CE feature, so no more datacentersFile or dynamicDatacenters flag necessary here.

api/cmd/kubermatic-api/main.go
```diff
@@ -142,11 +145,11 @@ func createInitProviders(options serverRunOptions) (providers, error) {
 	if err != nil {
 		return providers{}, fmt.Errorf("failed to construct manager: %v", err)
 	}
-	seedsGetter, err := provider.SeedsGetterFactory(context.Background(), mgr.GetClient(), options.dcFile, options.namespace, options.dynamicDatacenters)
+	seedsGetter, err := seedsGetterFactory(context.Background(), mgr.GetClient(), options)
```
xrstf, line 150: This is there the magic CE/EE split happens now. The calls to `provider.****` have been wrapped.

api/cmd/kubermatic-api/wrappers_ee.go
```diff
@@ -0,0 +1,25 @@
...
+	eeapi.AddFlags(fs)
+}
+
+func seedsGetterFactory(ctx context.Context, client ctrlruntimeclient.Client, opt serverRunOptions) (provider.SeedsGetter, error) {
+	return eeapi.SeedsGetterFactory(ctx, client, opt.namespace)
+}
+
+func seedKubeconfigGetterFactory(ctx context.Context, client ctrlruntimeclient.Client, opt serverRunOptions) (provider.SeedKubeconfigGetter, error) {
+	return eeapi.SeedKubeconfigGetterFactory(ctx, client, opt.kubeconfig)
+}
```
xrstf, line 25: This is basically "wrapping a wrapper". Instead of calling `pkg.ee.provider.SeedKubeconfigGetterFactory` directly here, we *have* to have another wrapper in `pkg.ee.cmd.kubermatic-api` because that's where the 2 required CLI flags are defined. Without these (dcFile and dynamicDatacenters) we cannot build an EE SeedsGetter.

So here in the main package we only forward to the "ee main" package in `pkg.ee.cmd.kubermatic-api`.

api/cmd/master-controller-manager/main.go
```diff
@@ -189,32 +179,19 @@ func main() {
...
-			log,
-			ctrlCtx.namespace,
-			runOpts.workerName,
-			ctrlCtx.seedsGetter,
-			provider.SeedClientGetterFactory(ctrlCtx.seedKubeconfigGetter),
-			migrationOptions.SeedMigrationEnabled())
-		if err != nil {
-			log.Fatalw("failed to create validatingAdmissionWebhook server for seeds", zap.Error(err))
+		if err := setupSeedValidationWebhook(ctx, mgr, log, runOpts, ctrlCtx); err != nil {
+			log.Fatalw("failed to start seed validation webhook", zap.Error(err))
```
xrstf, line 193: The seed webhook stuff had to be wrapped because it needs to know if we're in "migration mode" (i.e. datacenters.yaml is given, but also -dynamic-datacenters), because in this mode the validation webhook works a teensie bit differently.

And since the "migration mode" depends on EE features, the webhook stuff needed its own wrapper (`setupSeedValidationWebhook`).

PR #5419, opened 2020-05-18
[EE] Use kubermatic/api-ce repository for Community Edition

xrstf:
When building a CE version of Kubermatic, the Docker images will be put into a different Docker repository. This will aid in permission management (compared to using "EE tags" only, in the same repository). To make migration for existing users easier, the existing repo is now declared to be the EE repo, and `api-ce` / `dashboard-ce` will be for the Community Edition.

This PR extends the Kubermatic Operator so it is built for either the CE or the EE variant.

```release-note
NONE
```

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -1,5 +1,6 @@
 export CGO_ENABLED?=0
-REPO=quay.io/kubermatic/api
+KUBERMATIC_EDITION?=ce
+REPO=quay.io/kubermatic/api$(shell [ "$(KUBERMATIC_EDITION)" != "ee" ] && echo "-$(KUBERMATIC_EDITION)" )
 CMD=$(filter-out OWNERS nodeport-proxy kubeletdnat-controller, $(notdir $(wildcard ./cmd/*)))
 GOBUILDFLAGS?=-v
 GITTAG=$(shell git describe --tags --always)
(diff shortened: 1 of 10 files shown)


kdomanski:
I think this alone will break a couple of things. Most notably the image in the charts is not updated.

xrstf:
> I think this alone will break a couple of things. Most notably the image in the charts is not updated.

Care to explain? The Kubermatic Helm chart is deprecated, so only existing EE customers should ever need it. New customers should use the operator, which can handle CE and EE just fine.

kdomanski:
The official documentation still refers to installation by wizard and/or charts and not the operator.
Our own automated deployments use the charts as well.

xrstf:
> The official documentation still refers to installation by wizard and/or charts and not the operator.

Wrong: https://docs.kubermatic.com/kubermatic/master/installation/install_kubermatic/

> Our own automated deployments use the charts as well.

So? So they will continue to use quay.io/kubermatic/api as if nothing changed. Just like our existing customers.

kdomanski:
Ah sorry, somehow I got redirected to v2.13 documentation and never noticed.

xrstf:
> Our own automated deployments use the charts as well.

Also, yes, this is really a TODO we should get to soon.

PR #5431, opened 2020-05-22
Do not use deprecated Go tags syntax anymore

xrstf:
Our retry function does not handle strings with spaces very well:

```
retry 2 go test -tags "foo bar" ./...
```

Results in `go test -tags foo bar ./...` being executed. Instead of fiddling with bash's quoting, this PR simply removes compatibility with older Go versions by using comma-separated tags.

```release-note
NONE
```


xmudrii (approved):

xrstf:
/retest

The test is failing because the container-linux-update-agent cannot find Container Linux nodes. All nodes are Flatcars. The DaemonSet is created by a controller that should only react if a Container Linux node exists:

```
	predicates := predicateutil.Factory(func(m metav1.Object, _ runtime.Object) bool {
		return m.GetLabels()[nodelabelerapi.DistributionLabelKey] == nodelabelerapi.ContainerLinuxLabelValue
	})
	return c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}, predicates)
```

Given that this test is super flaky, I guess at some point there is a Container Linux node and the controller then is not capable of cleaning up the DaemonSet when it's not neede anymore.

PR #5432, opened 2020-05-22
addon installer controller: update existing addons if needed

multi-io:

This extends the addon installer controller so it not only installs default addons if they don't exist, but also updates them in case the in-cluster resource isn't the same as the one from the DefaultKubernetesAddons list. This became necessary for us after https://github.com/kubermatic/kubermatic/pull/5193 and was mentioned briefly in https://github.com/kubermatic/kubermatic/pull/5193#issuecomment-630142192.

**Documentation**:

```release-note
Existing default addons will be updated if their definition is changed.
```

### DIFF
--- a/api/pkg/controller/seed-controller-manager/addoninstaller/addoninstaller_controller.go
+++ b/api/pkg/controller/seed-controller-manager/addoninstaller/addoninstaller_controller.go
@@ -3,6 +3,7 @@ package addoninstaller
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


kdomanski:
pkg/controller/seed-controller-manager/addoninstaller/addoninstaller_controller_test.go:77: File is not `gofmt`-ed with `-s` (gofmt)

multi-io:
Rebased, formatting fixed.

PR #5437, opened 2020-05-25
ensure the node-labeler controller uses consistent labels

xrstf:
Depending on the map iteration order, the node labeler controller labelled "flatcar container linux" systems sometimes as "container linux" because of its substring matching. To make the logic consistent, we now only consider the longest label match.

This should make e2e tests much less flaky.

```release-note
Fix nodes sometimes not having the correct distribution label applied.
```


kdomanski:
Nice.
/lgtm
/approve

kdomanski:
cc @xrstf

PR #5439, opened 2020-05-25
Go Modules Proposal

xrstf:
This describes how Kubermatic switches from dep to Go modules.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/go-modules.md
+++ b/docs/proposals/go-modules.md
@@ -87,6 +87,6 @@ keep semver-compatible version tags.
 ### Vendoring
 
 Currently we use vendoring to have the repository all contained. This speeds up compile times and makes
-builds more resilient against some cloud provider's network infrastructure and peering to GitHub. As
-having the `vendor` directory or not does not touch any of the proposals above, I propose that we keep
-the directory for now.
+builds more resilient against some cloud provider's network infrastructure and peering to GitHub. However
+moving forward the build time speed up should be achieved with proper modcache handling instead, making
+the `vendor` directory obsolete. I propose that we remove it from the repository.
--- a/docs/proposals/go-modules.md
+++ b/docs/proposals/go-modules.md
@@ -1,0 +1,92 @@
+# Using Go Modules
+
+**Author**: Christoph Mewes (@xrstf)
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+
+Kubermatic uses [dep](https://github.com/golang/dep) to manage its dependencies. Since its introduction
+to the codebase it has become apparent that
+
+* dep is brittle and confusing to use sometimes,
+* dep is not developed anymore and of course,
+* [Go modules](https://github.com/golang/go/wiki/Modules) are the way forward.
+
+The goal of this proposal is to remove dep from Kubermatic and instead manage dependencies using
+Go modules.
+
+## Implementation
+
+### Module Naming
+
+Go modules must have a name and so does Kubermatic. The [Modules specs](https://github.com/golang/go/wiki/Modules#semantic-import-versioning)
+say
+
+> If the module is version v2 or higher, the major version of the module must be included as a `/vN` at
+> the end of the module paths used in `go.mod` files and in the package import path.
+
+Given that Kubermatic is currently in version 2.14, this requires us to include a version number in
+the module path. This gives us `github.com/kubermatic/kubermatic/v2` as the module name (note the missing
+`/api`).
+
+### api/
+
+The `api` directory is a leftover from when this repository was a monorepo containing Kubermatic, the
+bare metal provider and an addon manager (see PR [#385](https://github.com/kubermatic/kubermatic/pull/385)).
+Now that this is no longer the case, I propose that we get rid of it. This also nicely solves the question
+of whether or not, and where, to put the `api` directory in the module path: just don't.
+
+So we would do a `mv api/* .` in the repository.
+
+### Custom Repository
+
+As Kubernetes does, I propose we use our `kubermatic.com` domain to abstract away GitHub and have custom
+module names: `kubermatic.com/kubermatic/v2` sounds good. We will also use this for other products like
+KubeOne and KubeCarrier.
+
+As [documented](https://golang.org/cmd/go/#hdr-Remote_import_paths), we need to host a HTML snippet on
+`kubermatic.com` to direct `go get` to the GitHub repository. As the HTML snippet is pretty static (does
+not change with versioning, only for new majors), I propose we just keep it in our website's repository.
+
+This is how the HTML snippet for `k8s.io/klog/v2` looks like, as an example:
+
+```html
+<html>
+  <head>
+    <meta name="go-import"
+          content="k8s.io/klog
+                   git https://github.com/kubernetes/klog">
+    <meta name="go-source"
+          content="k8s.io/klog
+                   https://github.com/kubernetes/klog
+                   https://github.com/kubernetes/klog/tree/master{/dir}
+                   https://github.com/kubernetes/klog/blob/master{/dir}/{file}#L{line}">
+  </head>
+</html>
+```
+
+### New Major?
+
+The [Modules specs](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher) say
+
+> Note that if you are adopting modules for the first time for a pre-existing repository or set of packages
+> that have already been tagged v2.0.0 or higher before adopting modules, then the recommended best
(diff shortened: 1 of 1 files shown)


kron4eg:
@xrstf get out of my head, you mind-reader!

Big 👍✅✔️➕

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+
+Given that Kubermatic is currently in version 2.14, this requires us to include a version number in
+the module path. This gives us `github.com/kubermatic/kubermatic/v2` as the module name (note the missing
+`/api`).
+
+### api/
+
+The `api` directory is a leftover from when this repository was a monorepo containing Kubermatic, the
+bare metal provider and an addon manager (see PR [#385](https://github.com/kubermatic/kubermatic/pull/385)).
+Now that this is no longer the case, I propose that we get rid of it. This also nicely solves the question
```
irozzo-1A, line 37: 👍

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+coincides with a "real" version 3).
+
+### Release Strategy
+
+Our branching model is not affected by the change. We can still have `release/vX.Y` branches, as long as we
+keep semver-compatible version tags.
+
+### Vendoring
+
+Currently we use vendoring to have the repository all contained. This speeds up compile times and makes
```
irozzo-1A, line 89: IMO we should eventually get rid of `vendor`. Having a warmed mod cache is equivalent to having  vendor folder. In addition I'm not sure that cloning a very large repo including vendoring is more convenient than cloning a small one and download dependencies relying on [GOPROXY](https://proxy.golang.org/). I don't exclude that the opposite is true.

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+Go modules.
+
+## Implementation
+
+### Module Naming
+
+Go modules must have a name and so does Kubermatic. The [Modules specs](https://github.com/golang/go/wiki/Modules#semantic-import-versioning)
+say
+
+> If the module is version v2 or higher, the major version of the module must be included as a `/vN` at
```
kdomanski, line 26: I think this mostly meant to deal with libraries changing their API, and even then it's mostly a remainder of the "build only from master" orthodoxy of the early Go ecosystem. @themue WDYT?

docs/proposals/go-modules.md
kdomanski: :+1:

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+The `api` directory is a leftover from when this repository was a monorepo containing Kubermatic, the
+bare metal provider and an addon manager (see PR [#385](https://github.com/kubermatic/kubermatic/pull/385)).
+Now that this is no longer the case, I propose that we get rid of it. This also nicely solves the question
+of whether or not, and where, to put the `api` directory in the module path: just don't.
+
+So we would do a `mv api/* .` in the repository.
+
+### Custom Repository
+
+As Kubernetes does, I propose we use our `kubermatic.com` domain to abstract away GitHub and have custom
```
kdomanski, line 44: I think a vanity import path is more trouble than it's worth. cc @scheeles

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+
+Our branching model is not affected by the change. We can still have `release/vX.Y` branches, as long as we
+keep semver-compatible version tags.
+
+### Vendoring
+
+Currently we use vendoring to have the repository all contained. This speeds up compile times and makes
+builds more resilient against some cloud provider's network infrastructure and peering to GitHub. As
+having the `vendor` directory or not does not touch any of the proposals above, I propose that we keep
+the directory for now.
```
kdomanski: :+1:

docs/proposals/go-modules.md
themue: Working on a larger modules based project right now w/o vendoring. Local mod cache gets filled transparent and only needs time on first run. No downsides due to modules. Also `go.mod` makes used modules and versions very transparent as well as `go mod graph`.

docs/proposals/go-modules.md
themue: Yes, the reason is to avoid API breaks when using master or at least while doing a non-top-level version change with an imported module. So far it leads to three results:

- Long time usage of a v0 by the creators to signal that an API is still not stable.
- So far only few API breaks in v1 modules.
- API changes not use a v2, v3, etc but change the whole import path. See Go protocol buffers API (https://blog.golang.org/protobuf-apiv2).

But that's simply something all projects have to find their way as Go modules are the way to go.

docs/proposals/go-modules.md
themue: 👍

docs/proposals/go-modules.md
themue: Which troubles in detail do you expect? Kubernetes itself is using `k8s.io` pretty fine.

docs/proposals/go-modules.md
themue: Trouble with network infrastructure is pretty much handled by using the GOPROXYs. See also https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/. As backup we also could setup an own proxy. That would help at CI.

themue (commented):
Would like to see `/vendor` dropped and appreciate a remote import path like described at
https://golang.org/cmd/go/#hdr-Remote_import_paths

docs/proposals/go-modules.md
irozzo-1A: IMO it's mostly aesthetic thing unless we use it significantly shorten the import paths. If we go that way I would favour something like `k8c.io` with respect to `kubermatic.com`.

The only small drawback I see is that it could generate ambiguity with import paths.

themue: Having a shorter name, but IMHO hard to assign to our brand. So would kubermatic.

docs/proposals/go-modules.md
```diff
@@ -0,0 +1,92 @@
...
+
+As Kubernetes does, I propose we use our `kubermatic.com` domain to abstract away GitHub and have custom
+module names: `kubermatic.com/kubermatic/v2` sounds good. We will also use this for other products like
+KubeOne and KubeCarrier.
+
+As [documented](https://golang.org/cmd/go/#hdr-Remote_import_paths), we need to host a HTML snippet on
+`kubermatic.com` to direct `go get` to the GitHub repository. As the HTML snippet is pretty static (does
+not change with versioning, only for new majors), I propose we just keep it in our website's repository.
+
+This is how the HTML snippet for `k8s.io/klog/v2` looks like, as an example:
```
themue, line 52: Done it in my case via the nginx configuration, makes it pretty simply. But surely depends on individual environment.

docs/proposals/go-modules.md
xrstf: I don't think we should question the very clear wording in the Go wiki. They have

> As a result of Semantic Import Versioning, code opting in to Go modules **must comply with these rules**:

even in bold font.

docs/proposals/go-modules.md
xrstf: This is all true and we will eventually kill the vendor directory once somebody has time to update our CI's gocache logic. It should be rather trivial, but still something for later (sooner than later, but still not now ^^).

docs/proposals/go-modules.md
xrstf: We discussed this further among the team leads and have settled on using what the proposal ... proposes. A shorter import path, better branding and (though minimally) the independence of the repo location outweight the inconvenience of having to manage HTML snippets on kubermatic.com.

docs/proposals/go-modules.md
xrstf: Site is probably hosted by Netlify, built with Hugo. HTML snippets will probably much easier to do in our case :-)

docs/proposals/go-modules.md
xrstf: Thinking about it, you're right. The proposal should define the end goal, not only the first (or how many arbitrary) step. I will rephrase to reflect that.

docs/proposals/go-modules.md
xrstf: Eventually we will get rid of `vendor/` and the proposal reflects that target now. Doesn't mean it has to be one gigantic cuthulu PR.

PR #5453, opened 2020-05-27
[release/v2.14] update machine-controller to v1.14.1

kdomanski:
```release-note
Updated machine-controller to `v1.14.1`
```

### DIFF
--- a/api/vendor/github.com/kubermatic/machine-controller/pkg/userdata/ubuntu/provider.go
+++ b/api/vendor/github.com/kubermatic/machine-controller/pkg/userdata/ubuntu/provider.go
@@ -114,6 +114,11 @@ hostname: {{ .MachineSpec.Name }}
 {{- /* Never set the hostname on AWS nodes. Kubernetes(kube-proxy) requires the hostname to be the private dns name */}}
 {{ end }}
 
+{{- if .OSConfig.DistUpgradeOnBoot }}
+package_upgrade: true
+package_reboot_if_required: true
+{{- end }}
+
 ssh_pwauth: no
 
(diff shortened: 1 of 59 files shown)


api/vendor/github.com/kubermatic/machine-controller/pkg/userdata/ubuntu/provider.go
```diff
@@ -241,36 +246,32 @@ write_files:
...
 
-    export CR_PKG='docker-ce=5:18.09.9~3-0~ubuntu-bionic'
+{{- /* We need to explicitly specify docker-ce and docker-ce-cli to the same version.
+	See: https://github.com/docker/cli/issues/2533 */}}
 
+    DOCKER_VERSION='5:18.09.9~3-0~ubuntu-bionic'
     DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install -y \
       curl \
       ca-certificates \
-      ceph-common \
```
irozzo-1A, line 249: Not related to this PR, but why did we remove those dependencies? They should be there according to me.

irozzo-1A (commented):
/lgtm

api/vendor/github.com/kubermatic/machine-controller/pkg/userdata/ubuntu/provider.go
kdomanski: https://github.com/kubermatic/machine-controller/pull/751
Well in the end they can be added by a DaemonSet. But you're right, this was unmandated.

PR #5470, opened 2020-06-02
[EE] switch Docker repository logic for Kubermatic

xrstf:
We decided to take the opportunity to rename our default Docker repository to `kubermatic` (from `api`).

```release-note
NONE
```

### DIFF
--- a/config/kubermatic/values.yaml
+++ b/config/kubermatic/values.yaml
@@ -84,7 +84,7 @@ kubermatic:
     nodeportRange: "30000-32767"
     replicas: 2
     image:
-      repository: "quay.io/kubermatic/api"
+      repository: "quay.io/kubermatic/kubermatic-ee"
       tag: "__KUBERMATIC_TAG__"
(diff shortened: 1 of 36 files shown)


config/kubermatic/values.yaml
```diff
@@ -84,7 +84,7 @@ kubermatic:
     nodeportRange: "30000-32767"
     replicas: 2
     image:
-      repository: "quay.io/kubermatic/api"
+      repository: "quay.io/kubermatic/kubermatic-ee"
```
xrstf, line 87: The EE version is referenced here because the Helm Chart is legacy and only for existing (= EE) customers. New installations should not use this chart, but instead the Operator, which itself defaults to the CE version.

xrstf:
/hold

Need to create the Docker repositories first.

xrstf:
/hold cancel

Quay Repositories exist now.

zreigz:
@xrstf thanks :+1:

PR #5471, opened 2020-06-02
remove unused oidc-proxy-client and other deprecated stuff

xrstf:
Since we updated the way our e2e tests communicate with Dex in #4886, the oidc-proxy-client was only left for upgrade tests. Since we removed those in favor of the Kubermatic Installer, it's now finally time to delete the unused code.

```release-note
NONE
```


api/hack/ci/testdata/oauth_values.yaml
```diff
@@ -15,24 +15,12 @@ dex:
   clients:
   - id: kubermatic
     name: Kubermatic
-    # change this to a new random secret after 2.13 is released
-    secret: ZXhhbXBsZS1hcHAtc2VjcmV0
+    secret: BDZleMTgqON5kiJybIBZM4Si
```
xrstf, line 18: @zreigz Not sure if this needs to be updated in the dashboard as well. We deploy our own Dex, whereas the dashboard uses dev.kubermatic.io, so maybe it's independent?

zreigz: @floreks WDYT?

floreks: Dashboard uses dev only for local e2e test, when we are writing them and want to test locally. CI also uses Kubermatic in kind. I don't remember right now what impact it had on the final installation.

PR #5473, opened 2020-06-03
Update Dashboard dependencies to use dashboard-ee by default

floreks:
As `dashboard` image is no longer considered EE, I have updated the code to use `dashboard-ee` instead.

```release-note
NONE
```


xrstf:
I've adopted your fix to the `values.yaml` (dashboard => dashboard-ee) into my PR #5470, so I guess this one can be closed?

floreks:
Yep

Issue #4942, opened 2020-01-09
Tunneling Expose Strategy Alpha Stage


kdomanski:
**User Story**
We're running into a limit of 25 AWS clusters caused by the fact, that there can only be 2 listeners on an AWS LB, and we're using 2 listeners per cluster.

https://loodse.slack.com/archives/CC3BRUSJH/p1573035498034900

**Note:** Issues relating to this epic are managed [via ZenHub](https://app.zenhub.com/workspace/o/kubermatic/kubermatic/issues/4942).

irozzo-1A:
@kdomanski If my understanding is good this issue should be described this way:

When running a seed cluster in AWS with [global load-balancer](https://docs.loodse.com/kubermatic/v2.13/concepts/expose-strategy/expose_strategy/#global-loadbalancer) expose strategy we use 2 listeners for each user cluster.

- one for the apiserver
- the other for the openvpn server

The reason is that currently we cannot use SNI for the apiserver due to https://github.com/kubernetes/enhancements/pull/1216, so the loadbalancing is handled the following way.

A service of type loadbalancer is created for an envoyproxy deployment. For each service/user-cluster the [lb-updater](https://github.com/kubermatic/kubermatic/tree/master/api/cmd/nodeport-proxy/lb-updater) is adding a port pointing to an envoy cluster that is configured dynamically by a [sidecar container](https://github.com/kubermatic/kubermatic/tree/master/api/cmd/nodeport-proxy/envoy-manager) running on each envoyproxy pod. This on AWS triggers the creation of a new listener.

irozzo-1A:
@youssefazrak have you something to amend/add with respect to my comment?

youssefazrak:
@irozzo-1A Nothing to add, that's well summarized.

Now the question is how we could overcome this:
- Raise the quota for the ELB (Easiest way **IF** accepted). 
- Use multiple ELBs. Not cost-effective.
- Find a way to LB the packets without using SNI. This would mean TLS termination at an LB or Ingress level. Also, apiserver traffic is not the same as OpenVPN traffic (by default udp 1194, can use TLS 443 for example) and other specific things related to the type of traffic and headers that we are able to play with needs to be taken into account.

irozzo-1A:
> * Raise the quota for the ELB (Easiest way **IF** accepted).

This is my preferred one, simple and effective ;-) https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-limits.html

[...]

PR #5483, opened 2020-06-04
Add expose strategy proposal based on HTTP tunneling and SNI

irozzo-1A:
Co-authored-by: Youssef Azrak <yazrak.tech@gmail.com>
Co-authored-by: Iacopo Rozzo <iacopo@loodse.com>

This proposal addresses issue #4942 

Fixes #

**Documentation**:

```release-note
NONE
```


thetechnick:
Very nice proposal, looks very cool!

Have you considered/investigated, if we can also support another HTTP Proxy between the Worker Nodes and the Seed Cluster?
I have seen very rigid and air-gaped environments, where the Worker nodes MUST go through a whitelisting HTTP Proxy.

As everything is now speaking HTTP, I don't see a big issue with that, but it would be dope, if we could validate and document such a setup. I am sure the colleagues in PS would highly appreciate that!

@toschneck / @scheeles Might have more information on customer requirements on this.

scheeles:
LGTM from me but agree to Nico, we should consider HTTP Proxy

irozzo-1A:
> Have you considered/investigated, if we can also support another HTTP Proxy between the Worker Nodes and the Seed Cluster?

@thetechnick you raised a very good point, me and @youssefazrak considered it even if we did not write a dedicated section yet. 

I think that it should be feasible to have an HTTP proxy in between the tunneler and the HTTP proxy running in the seed cluster. IMO to enable this, the tunneler should pass from the middle HTTP proxy which should simply forward the CONNECT requests addressed to the seed cluster HTTP proxy, but I want to be cautious here because we did not setup an environment to demonstrate this and maybe I'm missing something and/or there are better options.

What is sure is that it makes the picture a bit more complicated, but I understand it's a mandate in some environments.

@toschneck feedbacks from PS team are more than welcome regarding this topic.

youssefazrak:
I just updated the proposal with the addition of mTLS:

- Diagram has been updated for the tunneling part. Now pods and the Openvpn client uses the same HTTP2 Tunneler.
- Security sections: should be "resolved" as I managed to get mTLS setup and working between the two clusters.

irozzo-1A:
@thetechnick @scheeles We added the section regarding HTTP proxy support, we also opened an issue on envoy to see if we can natively support double tunneling:

https://github.com/envoyproxy/envoy/issues/11569

docs/proposals/http-tunnel-expose-strategy.md
```diff
@@ -0,0 +1,137 @@
...
+
+To overcome this limitation, we have two strategies that rely on load balancers. 
+
+The first one: “Global LoadBalancer” is using one entry point and is routing requests based on ports uniquely assigned in the nodeport range for each user cluster service (apiserver and openvpn). One downside of this approach is having a limited number of listeners on the cloud load balancer. In AWS, there is a limit of 50 listeners per ELB (ALB or NLB) which means 50 services. As we are using 2 services per user cluster this leads to a maximum of 25 user clusters per load balancer. Another downside is that each service consumes two node ports, which could lead to exhaustion of the node port ranges for large clusters (e.g. default nodeport range is 30000-32767 meaning ~700 user clusters). 
+
+This leads us to the second load balancing strategy: “One Load Balancer Per User Cluster”. The downside of this solution is cost related. The cost of the load balancers increases linearly with the number of user clusters.
```
irozzo-1A, line 25: I think that in many cases the cost of the LBs is dominated by the cost of the rest of the infrastructure especially VMs (e.g. if the number of nodes per cluster is in the order of ~100). But we do have use cases in which a lot of small user cluster are created (e.g. ~5 nodes per user cluster).

Another problem, more relevant in general maybe, is related to LB quota issues (e.g. 50 by default in AWS): https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html

docs/proposals/http-tunnel-expose-strategy.md
```diff
@@ -0,0 +1,137 @@
...
+The HTTP Proxy on seed cluster must only accept CONNECT requests addressed to
+the OpenVPN server and Kube Apiserver, all other requests should be dropped.
+
+Moreover and to prevent disclosure of information sent in cleartext between the 
+CONNECT requests, we have implemented mTLS between the two Envoy pods 
+(HTTP Tunneler on the user cluster and HTTP Proxy on the seed cluster). 
+
+## **Alternatives considered**
+
+*   Using SNI everywhere: in this case, we would use the OpenVPN3 client instead of OpenVPN2 client. This one supports SNI. The problem we encounter is that OpenVPN uses the OpenVPN protocol which is based on TLS. In OpenVPN protocol the ClientHello is not the first packet exchanged after the TCP handshake, thus standard SNI routers cannot be used. This behaviour has been tested with Nginx ingress controller and Envoy.
```
irozzo-1A, line 121: We should evaluate https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/ for k8s >= 1.18 as a drop-in replacement for OpenVPN.

I this case we would be left with the problem of in-cluster communication with the apiserver where SNI does not apply.

irozzo-1A:
Closing this in favour of #5560 as after the repository was made public my fork got disconnected and I had to create a new one.

PR #5487, opened 2020-06-05
Seeds api adjust dc

### LINKED ISSUE #5486: dynamic datacenters - adjust seed API datacenter objects
**User Story**
The Seed and dynamic datacenter API use different datacenter structs. We should make it use one so its simpler.

lsviben:
Adjusts the seeds API to use the same API Datacenter struct as the dynamic datacenter API and adds the missing fake provider to the dynamic datacenter API struct.

Fixes #5486 

```release-note
action required - Seed API changes. Seed now contains a map of Datacenters instead of SeedDatacenters to be aligned with the Datacenter API. 
```


api/pkg/api/v1/types.go
```diff
@@ -45,9 +45,16 @@ type ObjectMeta struct {
...
+	// Name of the seed this datacenter belongs to.
+	Seed string `json:"seed,omitempty"`
+	// Optional: Country of the seed as ISO-3166 two-letter code, e.g. DE or UK.
+	// For informational purposes in the Kubermatic dashboard only.
+	Country string `json:"country,omitempty"`
+	// Optional: Detailed location of the cluster, like "Hamburg" or "Datacenter 7".
+	// For informational purposes in the Kubermatic dashboard only.
+	Location string `json:"location,omitempty"`
+	// Name of the datacenter provider. Extracted based on which provider is defined in the spec.
+	// For informational purposes in the Kubermatic dashboard only.
```
maciaszczykm: I'd avoid the part that says that is only for Kubermatic Dashboard's informational purposes. Some of our clients may use it too, as well as we might in the future add another client. Can we make it `It is used for informational purposes` or `It is used for informational purposes only`?

lsviben: Makes sense, Ill change it.

api/pkg/handler/v1/dc/datacenter.go
```diff
@@ -545,7 +545,7 @@ func kubermaticToAPI(dc *kubermaticv1.Datacenter, dcName, seedName string) (*api
 	}, nil
 }
 
-func apiSpec(dc *kubermaticv1.Datacenter, seedName string) (*apiv1.DatacenterSpec, error) {
+func KubermaticToAPISpec(dc *kubermaticv1.Datacenter, seedName string) (*apiv1.DatacenterSpec, error) {
```
maciaszczykm: Is it used outside the package?

maciaszczykm: By the way, we could name use `internal` and `external` for names here instead of `Kubermatic` and `API`.

cc @zreigz

api/pkg/api/v1/types.go
maciaszczykm: I just noticed it was like this before. I would change it anyways :)

maciaszczykm:
Looks good. I've added minor remarks.

api/pkg/handler/v1/dc/datacenter.go
zreigz: yes, good point. We do like this for others, for example, `convertInternalUserToExternal`

PR #5496, opened 2020-06-08
Update cert-manager to 0.15.1

xrstf:
This updates cert-manager in the legacy variant (to support k8s <1.15 for seed/master clusters, see https://cert-manager.io/docs/release-notes/release-notes-0.14/#improving-our-deployment-and-release-process). This PR also changes how CRDs are handled:

Previously, CRDs were installed by Helm when the chart was originally installed. Helm then never ever updated the CRDs. This is bad because cert-manager actually specs out their CRDs and we need them to be up-to-date.

To improve this, and to make the chart compatible with Helm 3, (which does not support the old `helm.sh/crd-install` hook and would need adjustments either way), this PR makes it so that CRDs are **always** installed manually by the user **before** installing/updating the Helm chart.

I tested:

* clean installing the master branch with Helm 2, then upgrading to this version using Helm 2, and
* clean installing the chart with Helm 3,

and both worked fine with no CRDs being lost or certs deleted. No further manual migration required, except you can delete the now unused `webhook-tls` Secret. But keeping it doesn't do anything either.

**Documentation:**
https://github.com/kubermatic/docs/pull/337

```release-note
ACTION REQUIRED: Update cert-manager to 0.15.1, change CRD handling
```


xrstf:
Closed due to repo change.

PR #5498, opened 2020-06-08
Normalize Prometheus scraping annotations

xrstf:
Historically we have been using `kubermatic/scrape` to annotations to mark pods for scraping. This was mostly done, IIRC, to prevent Prometheus from sccraping the entire seed cluster.

Over time, the monitoring stack has evolved and contains more metrics and dashboards for other components. There are now a number of reasons to use a more standard `prometheus.io/` prefix:

* It makes Kubermatic be one component in the cluster.
* Other monitoring solutions can easily find Kubermatic scrape targets.
* We already use `prometheus.io/` as the prefix inside the user-cluster Prometheus.
* We use `fluentbit.io/parser` for the parser name, not `kubermatic/fluentbit-parser`.

This PR therefore changes the prefix everywhere *but* the nginx-ingress and nodeport-proxy, to prevent re-deployments and LoadBalancer changes. The PR also

* normalizes `metric_path` to `metrics_path`,
* normalizes `scrape_port` to `port`,
* adds a new `kubermatic.io/chart` annotation to pods managed by Helm charts, in case we ever want to specifically target a certain chart by its name (or all pods just having this annotation),
* adjusts the scraping rules in Prometheus to handle both old and new annotations, so the rollout can happen over time.

I tested this on dev (Prometheus only, i.e. all other charts still use their original annotations) and the changes scraping rules match a broader, i.e. more complete set of pods (52 instead of 47). Node-exporter were also still matched and since rules/annotation for user-cluster components have not been touched, I didn't test it.

Currently, without this PR, the scraping of Kubermatic components is broken when using the Operator, because the Operator acciden^Wbravely already uses the new annotations. I will create a dedicated PR for the 2.14 release branch to just add the old annotations, which is much safer for an existing release.

**Documentation**:

TODO

```release-note
Prometheus scraping annotations have been normalized to use `prometheus.io/` as their prefix. Old annotations (`kubermatic/`) are still supported, but deprecated.
```

### DIFF
--- a/api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
+++ b/api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
@@ -44,10 +44,10 @@ func EnvoyDeploymentCreator(seed *kubermaticv1.Seed, versions common.Versions) r
 
 			d.Spec.Template.Labels = d.Spec.Selector.MatchLabels
 			d.Spec.Template.Annotations = map[string]string{
-				"kubermatic/scrape":       "true",
-				"kubermatic/scrape_port":  strconv.Itoa(EnvoyPort),
-				"kubermatic/metrics_path": "/stats/prometheus",
-				"fluentbit.io/parser":     "json_iso",
+				"prometheus.io/scrape":       "true",
+				"prometheus.io/scrape_port":  strconv.Itoa(EnvoyPort),
+				"prometheus.io/metrics_path": "/stats/prometheus",
+				"fluentbit.io/parser":        "json_iso",
 			}
 
 			d.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyAlways
--- a/api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
+++ b/api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
@@ -45,7 +45,7 @@ func EnvoyDeploymentCreator(seed *kubermaticv1.Seed, versions common.Versions) r
 			d.Spec.Template.Labels = d.Spec.Selector.MatchLabels
 			d.Spec.Template.Annotations = map[string]string{
 				"prometheus.io/scrape":       "true",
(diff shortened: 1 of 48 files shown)


config/monitoring/prometheus/config/scraping/pods.yaml
```diff
@@ -2,26 +2,54 @@ job_name: 'pods'
 kubernetes_sd_configs:
 - role: pod
 relabel_configs:
-# drop everything within test clusters
-- source_labels: [__meta_kubernetes_namespace]
-  regex: prow-kubermatic-.*
-  action: drop
```
xrstf, line 8: This is a leftover from before we used Kind for e2e tests.

api/pkg/controller/operator/seed/resources/nodeportproxy/deployments.go
```diff
@@ -44,10 +44,10 @@ func EnvoyDeploymentCreator(seed *kubermaticv1.Seed, versions common.Versions) r
...
 			d.Spec.Template.Labels = d.Spec.Selector.MatchLabels
 			d.Spec.Template.Annotations = map[string]string{
-				"kubermatic/scrape":       "true",
-				"kubermatic/scrape_port":  strconv.Itoa(EnvoyPort),
-				"kubermatic/metrics_path": "/stats/prometheus",
-				"fluentbit.io/parser":     "json_iso",
+				"prometheus.io/scrape":       "true",
+				"prometheus.io/port":         strconv.Itoa(EnvoyPort),
+				"prometheus.io/metrics_path": "/stats/prometheus",
+				"fluentbit.io/parser":        "json_iso",
```
xrstf, line 50: Operator-managed nodeport-proxy is not yet stable and actually used/usable, so it's fine to still change the annotations.

xrstf:
/hold

This PR uses Chart versions that are based on other PRs. To prevent confusing version increments, the other PRs should be merged first.

xrstf:
The other PRs that this is waiting for have more things to discuss, so I reduced the version bumps in this PR in order to get it merged.

/hold cancel

PR #5512, opened 2020-06-10
do not scrape kube-state-metrics twice

xrstf:
Since adjusting the scraping annotations, we scraped kubestate-metrics twice. This is a mistake and can make other computed metrics fail because of cardinality issues.

This PR makes it so that we continue to use the explicit scraping rule for KSM and ignore it in the regular pod scraping job.

```release-note
NONE
```


youssefazrak (approved):
/lgtm

PR #5531, opened 2020-06-18
add token blacklist for the user

zreigz:
**What this PR does / why we need it**: token blacklist secret holds the authorization bearer tokens which were invalidated during logout operation. It enforces a new login to get API access

Fixes #

```release-note
NONE
```


kdomanski:
Holding so that it won't get merged until @maciaszczykm takes a look as well.

PR #5540, opened 2020-06-19
add logout endpoint

zreigz:
**What this PR does / why we need it**: logout method invalidates current user token. Adds token to the blocked list. Every API call checks if the token belongs to the blacklist and rejects the connection in this case.

Fixes #

```release-note
new logout endpoint: POST /api/v1/me/logout
```


xrstf:
Well this should not have happened again.

/test pre-kubermatic-e2e-aws-coreos-1.18

PR #5550, opened 2020-06-22
etcd launcher image base overridable

### LINKED ISSUE #5549: etcd launcher image location should be dynamically configurable
**User Story**

Currently the deployed etcd launcher statefulset's image is harcoded to `quay.io/kubermatic/etcd-launcher-v3[34]:${KUBERMATICCOMMIT}`. This should be user-settable via the controller-manager command line, just like it is for the kubermatic api and dnatcontroller images.

**Acceptance criteria**

* [ ] image location (the part up to baseTag, i.e. "quay.io/kubermatic/etcd-launcher-" by default) can be configured via command-line parameters
* [ ] `--overwrite-registry=...` command-line parameter, if given, should be honoured

multi-io:

This adds a command line option `--etcd-launcher-image-base` to the seed-controller-manager, which allows the user to set the base image name of rolled out etcd pods, similarly to `--kubermatic-image` and `--dnatcontroller-image` for the kubermatic api and dnat-controller image, respectively.

In the current master the etcd image name is handled differently from the api and dnat-controller image names, and it isn't fully configurable -- e.g. the path component `/kubermatic/` is [hardcoded](https://github.com/kubermatic/kubermatic/blob/7dd0fbaec5028c195a90e39dc8b6166e7795c1ef/api/pkg/resources/etcd/statefulset.go#L315), making it impossible to pull the etcd launcher image from external (non-kubermatic) docker repositories unless they have `/kubermatic/` in their name.

The PR modifies the etcd launcher image base setting so it works the same as the kubermatic api and dnat-controller image setting. The new `--etcd-launcher-image-base` sets the "base name" prefix of the launcher image, which is the first part of the image name from the beginning (i.e. including the repository) up to but not including `"v3[34]:<KUBERMATICCOMMIT>"`. The default value is "quay.io/kubermatic/etcd-launcher-", which is what the current master uses. The registry name can be overridden separately using `--overwrite-registry=`.

Fixes #5549

```release-note
New seed controller manager option --etcd-launcher-image-base
```

### DIFF
--- a/api/cmd/image-loader/main.go
+++ b/api/cmd/image-loader/main.go
@@ -386,7 +386,7 @@ func getTemplateData(version *kubermaticversion.Version) (*resources.TemplateDat
 		// Since this is the image-loader we hardcode the default image for pulling.
 		resources.DefaultKubermaticImage,
 		resources.DefaultDNATControllerImage,
-		resources.DefaultEtcdLauncherImageBase,
+		"",
 		false,
 	), nil
 }
--- a/api/cmd/image-loader/main.go
+++ b/api/cmd/image-loader/main.go
@@ -386,6 +386,7 @@ func getTemplateData(version *kubermaticversion.Version) (*resources.TemplateDat
 		// Since this is the image-loader we hardcode the default image for pulling.
 		resources.DefaultKubermaticImage,
 		resources.DefaultDNATControllerImage,
+		resources.DefaultEtcdLauncherImageBase,
 		false,
 	), nil
 }
--- a/api/cmd/seed-controller-manager/controllers.go
+++ b/api/cmd/seed-controller-manager/controllers.go
@@ -130,7 +130,7 @@ func createOpenshiftController(ctrlCtx *controllerContext) error {
 		ctrlCtx.runOptions.externalURL,
 		ctrlCtx.runOptions.kubermaticImage,
 		ctrlCtx.runOptions.dnatControllerImage,
-		ctrlCtx.runOptions.etcdLauncherImageBase,
+		ctrlCtx.runOptions.etcdLauncherImage,
 		openshiftcontroller.Features{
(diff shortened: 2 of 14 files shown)


multi-io:
:O I'll fix

moelsayed:
@multi-io I understand the motivation behind this, but I don't think it should be only for the image base as it would be confusing and not consistent with the other similar options. I think it would make more sense have the option override the full image name instead.

multi-io:
> @multi-io I understand the motivation behind this, but I don't think it should be only for the image base as it would be confusing and not consistent with the other similar options. I think it would make more sense have the option override the full image name instead.

OK yeah, makes sense. I'll change that.

multi-io:
> > @multi-io I understand the motivation behind this, but I don't think it should be only for the image base as it would be confusing and not consistent with the other similar options. I think it would make more sense have the option override the full image name instead.
> 
> OK yeah, makes sense. I'll change that.

I've implemented this change, but haven't pushed it yet because I reconsidered -- obviously, if the user can only override the full image name, they'll lose the ability to roll out different images (etcd 3.3 vs. 3.4) depending on the K8s version of the user cluster. So we'd trade functionality against being "consistent with the other similar options". Not sure if this is a good tradeoff. Ideas? Maybe it makes sense to say, let's only make the full image name overridable and if a user wants to retain the above ability with their own repository, they'll have to make a fork and change the image name prefix in the code. I'll push the change no problem if full image override is what you guys want to merge.

multi-io:
> I'll push the change no problem if full image override is what you guys want to merge.

OK I've just pushed it now to shorten the discussion. :)

moelsayed:
@multi-io I see your point. I initially thought about this in terms of testing/development, didn't consider production use. 
@kron4eg had suggested that we bundle the binary in the Kubermatic main image and use an initcontainer to copy it in the upstream etcd container avoiding the extra image all together. We might go with that.

multi-io:
Closing this PR -- we're abandoning this approach and going with https://github.com/kubermatic/kubermatic/pull/5710 instead.

PR #5551, opened 2020-06-22
added seed name list endpoint

lsviben:
adds a seed name list endpoint. It will allow us to cleanup the datacenters from seeds.

```release-note
Added a seed name list endpoint.
```


api/pkg/api/v1/types.go
```diff
@@ -1517,6 +1517,12 @@ type SeedSpec struct {
 	ExposeStrategy corev1.ServiceType `json:"expose_strategy,omitempty"`
 }
 
+// SeedNamesList represents a object for seed names.
+// swagger:model SeedNamesList
+type SeedNamesList struct {
+	Names []string `json:"names,omitempty"`
```
maciaszczykm: I wondering if pure array wouldn't be better here. WDYT?

PR #5571, opened 2020-06-25
Enable etcd cluster scaling for userclusters

### LINKED ISSUE #5545: Support configurable user cluster etcd cluster size
**User Story**
Currently, the user cluster etcd is hardcoded to 3 nodes only. The user should be able to scale up or down to improve reliability or reduce cost/load.

**Acceptance criteria**
- [ ] user cluster etcd cluster size should by dynamically configurable.
- [ ] UI should allow user to resize etcd cluster. Also should make sure the user uses an odd number.

moelsayed:
- Enables configurable etcd cluster size in the user cluster spec.
- Managed scale up/down for etcd stateful set.
- Automatic etcd member join and reconcile.
- Adds http based liveness probe for etcd pods
Fixes #5545

**Documentation**:

```release-note
Adds configurable etcd cluster size to user cluster spec. 
```


pkg/resources/etcd/statefulset.go
```diff
@@ -314,3 +338,23 @@ func getLauncherImage(data etcdStatefulSetCreatorData) (string, error) {
...
+	isEtcdHealthy := data.Cluster().Status.ExtendedHealth.Etcd == kubermaticv1.HealthStatusUp
+	// at required size. do nothing
+	if etcdClusterSize == replicas {
+		return replicas
+	}
+	if isEtcdHealthy { // no scaling until we are healthy
+		if etcdClusterSize > replicas {
+			return replicas + 1
+		}
+		return replicas - 1
```
multi-io, line 362: Tested with a pre-existing, running cluster with 3 replicas. The statefulset ended up being reduced to 2 replicas, I assume here (.Spec.EtcdClusterSize not set => etcdClusterSize:=0; replicas:=3; isEtcdHealthy:=true => return 2). The etcd launcher then refuses to launch a 2-node cluster because that's smaller than the defaultClusterSize (3), leading to a permanently non-running cluster. Is the idea that the admin must always set .Spec.EtcdClusterSize?

moelsayed: No, it's defaulted [here](https://github.com/kubermatic/kubermatic/pull/5571/commits/11068e39b1e90f875b5b62d4b6b96b1fc7d7190e#diff-eea4352ca8918a218965fcc028c01dceR34-R36), which runs as part of the API server. I am not 100% sure if it will default existing clusters as well. But it is a good point, we should be more defensive here.

multi-io: Confirmed that an existing cluster is migrated to the new launcher properly if .spec.etcdClusterSize is set to 3 first. Maybe we should just treat an unspecified (==0) value for the field as if the field was set to 3.

multi-io: > No, it's defaulted [here](https://github.com/kubermatic/kubermatic/pull/5571/commits/11068e39b1e90f875b5b62d4b6b96b1fc7d7190e#diff-eea4352ca8918a218965fcc028c01dceR34-R36), which runs as part of the API server. I am not 100% sure if it will default existing clusters as well. But it is a good point, we should be more defensive here.

Yeah the defaulting I saw too, but I'm pretty sure that's only [called when new clusters are created](https://github.com/kubermatic/kubermatic/blob/a5e42e57fe7d64bd8a1052b0ec901e4458c72162/pkg/resources/cluster/cluster.go#L59).

moelsayed: @multi-io I just pushed a fix for it. Can you please confirm it's working?

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	err := e.getConfigFromEnv()
 	if err != nil {
 		log.Fatalf("failed to get launcher configuration: %v", err)
 	}
 
+	initialMembers := initialMemberList(e.config.clusterSize, e.config.namespace)
 	// not required, will leave it for now.
-	os.Setenv(initialStateEnvName, "new")
-	os.Setenv(initialClusterEnvName, initialMemberList(config.clusterSize, config.namespace))
+	os.Setenv(initialStateEnvName, e.config.initialState)
```
moadqassem: This line and the line underneath it has no error check, I know the original code had no error check either but would make since to add some checks here.

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	os.Setenv(initialClusterEnvName, strings.Join(initialMembers, ","))
```
moadqassem: One other thing regarding this line and the one above, are you using it for debugging? because normally env vars are populated via the job instead of setting explicitly in the code.

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	}
+	return endpoints
 }
 
-func getConfigFromEnv() (*envConfig, error) {
+func (e *etcdCluster) endpoint() string {
+	return fmt.Sprintf("%s.etcd.%s.svc.cluster.local:2380", e.config.podName, e.config.namespace)
+}
+
+func (e *etcdCluster) getConfigFromEnv() error {
```
moadqassem: The way we actually pass configs to the bin/cmd, is via having the args/options pattern. In other words, we pass those configs as flags to the cmd instead of env vars. So it follows as:

* From out side(e.g: job runner) env vars are being created and passed to the container for example.
* Inside the container/job the configs are passed as flags. 

For more information take a look at the [user-cluster-controller-manager](https://github.com/kubermatic/kubermatic/blob/master/cmd/user-cluster-controller-manager/main.go) and how those args are read via the hack [script](https://github.com/kubermatic/kubermatic/blob/master/hack/run-user-cluster-controller-manager.sh)

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
-		"/usr/local/bin/etcd",
-		etcdCmd(config),
-		os.Environ())
+	// setup and start etcd command
+	cmd := exec.Command("/usr/local/bin/etcd", etcdCmd(e.config)...)
+	cmd.Env = os.Environ()
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	if err = cmd.Start(); err != nil {
+		log.Fatalf("failed to start etcd: %v", err)
```
moadqassem, line 78: Can you please use our logging lib `"github.com/kubermatic/kubermatic/pkg/log"` ? here is an [example](https://github.com/kubermatic/kubermatic/blob/master/cmd/user-cluster-controller-manager/main.go#L115-L116) how it is initialised and used

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
 	log.Print("initializing etcd..")
 	log.Printf("initial-state: %s", os.Getenv(initialStateEnvName))
 	log.Printf("initial-cluster: %s", os.Getenv(initialClusterEnvName))
 
-	err = syscall.Exec(
-		"/usr/local/bin/etcd",
-		etcdCmd(config),
-		os.Environ())
+	// setup and start etcd command
+	cmd := exec.Command("/usr/local/bin/etcd", etcdCmd(e.config)...)
```
moadqassem: Shouldn't we make suer that `etcd` cmd exists before doing any calls or preparation? if os.Status("<etcd-path>") => not exist return an error as early as possible?

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	cmd.Env = os.Environ()
```
moadqassem, line 73: why this is needed?

cmd/etcd-launcher/main.go
```diff
@@ -130,6 +214,139 @@ func etcdCmd(config *envConfig) []string {
...
+
+func (e *etcdCluster) isEndpointHealthy(endpoint string) (bool, error) {
+	client, err := e.getClientWithEndpoints([]string{endpoint})
+	if err != nil {
+		return false, err
+	}
+	// just get a key from etcd, this is how `etcdctl endpoint health` works!
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	_, err = client.Get(ctx, "healthy")
+	cancel()
```
moadqassem: Nit:maybe check the error then cancel? `defer cancel()`

cmd/etcd-launcher/main.go
```diff
@@ -130,6 +214,139 @@ func etcdCmd(config *envConfig) []string {
...
+}
+
+func (e *etcdCluster) isHealthy() (bool, error) {
+	if err := e.getClient(); err != nil {
+		return false, err
+	}
+	// just get a key from etcd, this is how `etcdctl endpoint health` works!
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	_, err := e.client.Get(ctx, "healthy")
+	cancel()
```
moadqassem: Nit:maybe check the error then cancel? `defer cancel()`

cmd/etcd-launcher/main.go
```diff
@@ -130,6 +214,139 @@ func etcdCmd(config *envConfig) []string {
 			"--experimental-corrupt-check-time=10m",
 		}...)
 	}
-
 	return cmd
 }
+
+func (e *etcdCluster) getClient() error {
+	if e.client != nil {
```
moadqassem, line 233: I don't understand this one, so when the client is not nil you are returning nil, shouldn't you return the client here?

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	cmd.Env = os.Environ()
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	if err = cmd.Start(); err != nil {
+		log.Fatalf("failed to start etcd: %v", err)
+	}
+
+	// if existing, we need to reconcile
+	var healthy bool
+	for i := 0; i < 5; i++ {
```
moadqassem: So here you are trying to make sure that the storage is healthy, so you are trying to get valid results in 5 times. In general that's valid of course but what about adding some semantics something similar to this
```
	var (
		tries    int
		maxTries int
                healthy bool
	)

	for tries < maxTries {
		if healthy, err = e.isHealthy(); !healthy || err != nil {
			log.Error("...")
			time.Sleep(500 * time.Milisecond)
			tries++
			continue
		}
		log.Info("...")
		break
	}

      if !healthy || err != nil {
       log.Fatal("failed for ever")
     } 
```

pkg/resources/etcd/statefulset.go
multi-io: Confirmed. Existing etcd cluster was migrated without having to manually set .spec.etcdClusterSize or anything else in the cluster resource.

cmd/etcd-launcher/main.go
moadqassem: Or you can try out this lovely package:
https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Poll

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+	isMemeber, err := e.isClusterMember(e.config.podName)
+	if err != nil {
+		log.Fatalf("failed to check cluster membership: %v", err)
 	}
 
+	if isMemeber {
+		for { // reconcile dead members
+			members, err := e.listMembers()
+			if err != nil {
+				time.Sleep(10 * time.Second)
```
moadqassem, line 95: Shouldn't we log the error in the for loop? this line and the lines underneath it?

multi-io:
I think the cluster size field should be .spec.etcd.clusterSize instead of .spec.etcdClusterSize, i.e. have a spec.etcd map in there that can accommodate future etcd-related settings.

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
+			leader, err := e.isLeader()
+			if err != nil || !leader {
+				time.Sleep(10 * time.Second)
+				continue
+			}
+			if err := e.removeDeadMembers(); err != nil {
+				continue
+			}
+		}
+	} else { // new etcd member, need to join hte cluster
```
moadqassem: typo: s/hte/the

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
 	if config.namespace, ok = os.LookupEnv("NAMESPACE"); !ok || config.namespace == "" {
-		return nil, errors.New("NAMESPACE is not set")
+		return errors.New("NAMESPACE is not set")
 	}
 
 	config.clusterSize = defaultClusterSize
 	if s := os.Getenv("ECTD_CLUSTER_SIZE"); s != "" {
 		if config.clusterSize, err = strconv.Atoi(s); err != nil {
-			return nil, fmt.Errorf("failed to read ECTD_CLUSTER_SIZE: %v", err)
+			return fmt.Errorf("failed to read ECTD_CLUSTER_SIZE: %v", err)
```
moadqassem: typo: s/ECTD/ETCD

cmd/etcd-launcher/main.go
```diff
@@ -26,66 +33,137 @@ type envConfig struct {
...
 		}
-		if config.clusterSize > defaultClusterSize {
-			return nil, fmt.Errorf("ECTD_CLUSTER_SIZE is smaller then %d", defaultClusterSize)
+		if config.clusterSize < defaultClusterSize {
+			return fmt.Errorf("ECTD_CLUSTER_SIZE is smaller then %d", defaultClusterSize)
```
moadqassem: typo: s/ECTD/ETCD

typo: s/then/than

cmd/etcd-launcher/main.go
```diff
@@ -130,6 +214,139 @@ func etcdCmd(config *envConfig) []string {
...
+	return true, nil
+
+}
+
+func (e *etcdCluster) isLeader() (bool, error) {
+	var err error
+	if err = e.getLocalClient(); err != nil {
+		return false, err
+	}
+	var resp *clientv3.StatusResponse
```
moadqassem: This can be derived instead and declared inside the for loop since you are not using outside of it: 
```
resp, err = e.localClient.Status(context.Background(), e.endpoint())
```

cmd/etcd-launcher/main.go
```diff
@@ -130,6 +214,139 @@ func etcdCmd(config *envConfig) []string {
...
+	for _, member := range members {
+		if member.Name == e.config.podName {
+			continue
+		}
+		healthy, err := e.isEndpointHealthy(member.PeerURLs[0])
+		if err != nil {
+			continue
+		}
+		if !healthy {
+			if _, err := e.client.MemberRemove(context.Background(), member.ID); err != nil {
```
moadqassem: you don't need the check here and you are not wrap the error, thus remove the if statement please and return the method. 
```
return e.client.MemberRemove(context.Background(), member.ID)
```

cmd/etcd-launcher/main.go
moelsayed: `client` is set in the `etcdCluster` struct. If it's nil, I create it and assign it to e.client instead of passing it around.

cmd/etcd-launcher/main.go
moelsayed: I always pass forward the environment for commands as they might be used. One common use case is the `*_PROXY` variables.

pkg/resources/etcd/statefulset.go
moelsayed: @multi-io I think this should work https://github.com/kubermatic/kubermatic/pull/5571/files#diff-60471266513513d81d30f4855c2c01e2R316

cmd/etcd-launcher/main.go
```diff
@@ -3,28 +3,30 @@ package main
...
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"path"
-	"strconv"
 	"strings"
 	"time"
 
+	kubermaticlog "github.com/kubermatic/kubermatic/pkg/log"
```
moadqassem: Please regroup the imports respectively :

* go sdk imports(`"context"`)
* third party imports(`"go.uber.org/zap"`)
* kubermatic/machine-controller imports 
* k8s imports

cmd/etcd-launcher/main.go
```diff
@@ -88,30 +89,40 @@ func main() {
 		for { // reconcile dead members
 			members, err := e.listMembers()
 			if err != nil {
+				log.Warnf("failed to list memebers: %v ", err)
```
moadqassem: Please use logging with additional context with error wrapping: 
`log.Warnw("failed to list memebers ", zap.Error(err))`

cmd/etcd-launcher/main.go
```diff
@@ -37,45 +39,44 @@ type envConfig struct {
...
 
-	log.Print("initializing etcd..")
-	log.Printf("initial-state: %s", os.Getenv(initialStateEnvName))
-	log.Printf("initial-cluster: %s", os.Getenv(initialClusterEnvName))
+	log.Info("initializing etcd..")
+	log.Infof("initial-state: %s", e.config.initialState)
+	log.Infof("initial-cluster: %s", strings.Join(initialMembers, ","))
 
+	if _, err := os.Stat(etcdCommandPath); os.IsNotExist(err) {
+		log.Fatalf("can't find etcd command [%s]: %v", etcdCommandPath, err)
```
moadqassem: Please follow the pattern that we use regaring contextual logs:  
`log.Fatalw("can't find command","command-path", etcdCommandPath, zap.Error(err))`

cmd/etcd-launcher/main.go
```diff
@@ -88,30 +89,40 @@ func main() {
...
 			}
 			// we only need to reconcile if we have more members than we should
 			if len(members) <= e.config.clusterSize {
+				log.Info("cluster members reconciled..")
 				break
 			}
 			// to avoide race conditions, we will run only on the cluster leader
 			leader, err := e.isLeader()
 			if err != nil || !leader {
+				if err != nil {
```
moadqassem: You don't need to check the error two times, instead make the log message more general: 
`log.Warnw("failed to remove member, error occurred or didn't get the current leader", zap.Error(err))`

In this case, if there was an error it will be printed out as an error

cmd/etcd-launcher/main.go
```diff
@@ -88,30 +89,40 @@ func main() {
...
+					log.Warnf("failed leader check: %v", err)
+				}
 				time.Sleep(10 * time.Second)
 				continue
 			}
-			if err := e.removeDeadMembers(); err != nil {
+			if err := e.removeDeadMembers(log); err != nil {
+				log.Warnf("failed to remove member: %v", err)
```
moadqassem: Please adjust.
`log.Warnw("failed to remove member", zap.Error(err))`

cmd/etcd-launcher/main.go
```diff
@@ -338,14 +349,13 @@ func (e *etcdCluster) removeDeadMembers() error {
...
-		}
-		if !healthy {
-			if _, err := e.client.MemberRemove(context.Background(), member.ID); err != nil {
-				return err
-			}
+
+		if err = wait.Poll(1*time.Second, 30*time.Second, func() (bool, error) {
+			return e.isEndpointHealthy(member.PeerURLs[0])
+		}); err != nil {
+			log.Infof("member [%s] is not responding, removing from cluster", member.Name)
```
moadqassem: Please use:
`log.Infow("member is not responding, removing from cluster", "member-name", member.Name)`

cmd/etcd-launcher/main.go
```diff
@@ -37,45 +39,44 @@ type envConfig struct {
...
 	e := &etcdCluster{}
-	err := e.getConfigFromEnv()
+	err := e.parseConfigFlags()
 	if err != nil {
 		log.Fatalf("failed to get launcher configuration: %v", err)
 	}
 
+	logOpts := kubermaticlog.NewDefaultOptions()
+	rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
+	log := rawLog.Sugar()
```
moadqassem, line 59: This variable collides with the `"log"` imported package

pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -82,6 +82,21 @@ func (r *Reconciler) syncHealth(ctx context.Context, cluster *kubermaticv1.Clust
 	if err != nil {
 		return err
 	}
+	// set ClusterConditionEtcdClusterInitialized, this should be don't only once
```
moadqassem: I don't quite understand the comment, can you please make it clearer :-)

pkg/resources/etcd/statefulset.go
```diff
@@ -162,9 +171,13 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
 							Value: "/etc/etcd/pki/client/apiserver-etcd-client.key",
 						},
 						{
-							Name:  "ETCDCTL_ENDPOINTS",
-							Value: "https://127.0.0.1:2379",
+							Name:  "INITIAL_STATE",
+							Value: initialState,
 						},
+						// {
```
moadqassem: If this is not needed can you please remove it?

pkg/resources/etcd/statefulset.go
```diff
@@ -314,3 +327,24 @@ func getLauncherImage(data etcdStatefulSetCreatorData) (string, error) {
...
 	return data.ImageRegistry(resources.RegistryQuay) + "/kubermatic/etcd-launcher-" + baseTag + ":" + resources.KUBERMATICCOMMIT, nil
 }
+
+func computeReplicas(data etcdStatefulSetCreatorData, set *appsv1.StatefulSet) int {
+	etcdClusterSize := data.Cluster().Spec.EtcdClusterSize
+	if set.Spec.Replicas == nil { // new replicaset
+		return etcdClusterSize
+	}
+	replicas := int(*set.Spec.Replicas)
+	isEtcdHealthy := data.Cluster().Status.ExtendedHealth.Etcd == kubermaticv1.HealthStatusUp
```
moadqassem: Can you please move this line to 341, as this line is evaluated and not used if the `etcdClusterSize` == `replicas `

cmd/etcd-launcher/main.go
moelsayed: The imported package is aliased, so it should be ok?

PR #5577, opened 2020-06-29
fix charts crds location to helm3 conform location

toschneck:
Like mentioned in the docs (https://helm.sh/docs/topics/charts_hooks/#helm), CRDs should be stored at `crds/` folder, to enable correct hook management.
If not the following error occurs in e.g. on an `helm upgrade
--install`:
```
helm upgrade --install --create-namespace --wait --values ./kubermatic-setup-files/values.yaml --namespace cert-manager cert-manager kubermatic-repo/charts/cert-manager/ 
Release "cert-manager" does not exist. Installing it now.
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ClusterIssuer" in version "cert-manager.io/v1alpha2"
```

Fixes #

**Documentation**:

```release-note
NONE
```


kron4eg:
Retest won’t help


```
2020-06-29T22:12:34+00:00] Deploying Kubermatic CRDs
119
error: the path “charts/kubermatic/crd/“ does not exist
120
Retry 1/5 exited 1, retrying in 1 seconds...
```

xrstf:
This would reverse https://github.com/kubermatic/kubermatic/pull/5552.

/hold

xrstf:
To be more clear, we do not want Helm to manage CRDs anymore. Helm is terrible at managing CRDs. CRDs should be managed by our own installer / KubeCarrier / other orchestration tools.

This is also already a documented change in https://docs.kubermatic.com/kubermatic/master/upgrading/2.14_to_2.15/

toschneck:
Hmm in helm 3 the CRD management works quite ok, the problem if we don't use helm, after someone removes the helm release he needs to be aware that he also needs to remove CRDs (what most people will forget). For my perspective I would prefer we can manage the single components atomic. 
We can't expect that end users have there deploy script as long as are not the official installation way. 
it's just a renaming of `/crd/` to `/crds/`, why this would overwrite your change?
image

toschneck:
Please give me a hint, it looks I'm not into all the details of the changes you planed

xrstf:
> Hmm in helm 3 the CRD management works quite ok

If you consider "creates CRD, but never updates or deletes them" "quite ok", then yes. I consider it broken by design ;-)

> the problem if we don't use helm, after someone removes the helm release he needs to be aware that he also needs to remove CRDs (what most people will forget).

Helm will also not delete the CRDs when a release is removed. See https://helm.sh/docs/topics/charts/#limitations-on-crds for more information.

Before the change I mentioned, our CRD handling was entirely broken.

* The Kubermatic CRDs, which are just dummies with no specs, have alread always been applied manually and were never part of the Helm chart. So our upgrade procedure was always to manually `kubectl apply`, even though it does only matter for *new* CRDs here. So funny enough, for the Kubermatic chart it would have been *fine* (not great) to use Helm.
* For Velero/cert-manager we used Helm, so we never ever got CRD updates. Which is terrible because these projects actually spec out their CRDs and we actually *need* the CRDs to be updated. This is something Helm does not do. So our CRD handling for these charts so far was broken by design and we had to rely on manual upgrade notes with every Kubermatic release.

=> We did the opposite of what would have made sense.

To finally gain more control over CRD lifetimes, we decided to clean up and manage CRDs manually.

toschneck:
Alright understand ... maybe we would nee a CRD operator :laughing: 

Maybe we should think about two create for the dependencies dedicated [KUDO](https://kudo.dev/) operators and manage the dependencies with a simple operator. e.g. the Redis kudo operator looks pretty straight forward: https://github.com/kudobuilder/operators/tree/master/repository/redis

PR #5591, opened 2020-07-02
bump machine-controller to v1.15.0

kdomanski:
```release-note
updated machine-controller to `v1.15.0`
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -875,7 +875,7 @@
   version = "v0.1.0"
 
 [[projects]]
-  digest = "1:887211ee7dd190f9ae613304687f079e1397626a2a47eff320a26e260ee5e8d7"
+  digest = "1:ad2ebd3aa5ca619bf2a77cf6f8221f3c940fd9ea6a8531dbb5bef2941943a86a"
   name = "github.com/kubermatic/machine-controller"
   packages = [
     "pkg/apis/cluster/common",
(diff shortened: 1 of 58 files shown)


pkg/resources/machinecontroller/deployment.go
```diff
@@ -53,7 +53,7 @@ var (
 const (
 	Name = "machine-controller"
 
-	tag = "v1.14.3"
+	tag = "v1.15.0"
```
xmudrii, line 56: We should consider using 1.15.1 instead.

Gopkg.lock
```diff
@@ -904,7 +904,7 @@
     "pkg/userdata/ubuntu",
   ]
   pruneopts = "NUT"
-  revision = "1dd6b128854077daac70fb1a1d1975f365501604"
+  revision = "0f96e356b1f0ea9f7866547c1b5f9820e905a9ae"
   version = "v1.14.3"
```
xmudrii, line 908: Should we bump the MC dependency to v1.15.0/v1.15.1 in [`Gopkg.toml`](https://github.com/kubermatic/kubermatic/blob/e2ff54983224a4a9e6ba96ac05525fe28d04aa63/Gopkg.toml#L139)?

xmudrii:
/lgtm
/hold
in case you want to fix the comments, otherwise, remove hold

pkg/resources/machinecontroller/deployment.go
xmudrii: Although, it seems like the change in 1.15.1 doesn't really affect Kubermatic, so I'll leave it up to you to choose the version.

PR #5609, opened 2020-07-06
Fix label for nodeport-proxy when deployed with the operator

youssefazrak:
When installing Kubermatic with the operator, the nodeport-proxy is deployed with Label Selectors. 
Currently, both the deployment and the svc have different labels, leading to the non-creation of endpoints. In short, communications between the userclusters and the control planes are not possible.

**Documentation PR**
https://github.com/kubermatic/docs/pull/359

```release-note
NONE
```


irozzo-1A:
@youssefazrak could you please apply it to branch 2.14?

PR #5611, opened 2020-07-07
Fix seed validation

### LINKED ISSUE #5601: Failure to create Seed resource on CE
**User Story**
The creation of a Seed resource in the CE version is currently not possible, throwing the following error:
`Error from server: error when creating "seed.yaml": admission webhook "seeds.kubermatic.io" denied the request: failed to get seeds: Seed.kubermatic.k8s.io "kubermatic" not found`

Seems to be related to a "chicken and egg" problem and the fact that we are only allowed to have one seed with the CE edition. The controller checks the existence of the "Kubermatic" seed before creating one and thus fails. With the EE version, we have the possibility of having a list of multiple seeds thus, even if "Kubermatic" seed doesn't exist, it still creates it.

Pointers:
https://github.com/kubermatic/kubermatic/blob/master/pkg/ee/provider/datacenter.go#L172
https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/datacenter.go#L60
https://github.com/kubermatic/kubermatic/blob/master/pkg/validation/seed/seed.go#L56

**Acceptance criteria**

Having the possibility to create one seed resource when using the CE version.

irozzo-1A:
At the moment it is not possible to create Seed resources in CE mode, as the validation is broken.

Fixes #5601 

**Documentation**:

```release-note
Fix Seed validation for Community Edition.
```

### DIFF
--- a/cmd/master-controller-manager/wrappers_ce.go
+++ b/cmd/master-controller-manager/wrappers_ce.go
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/kubermatic/kubermatic/pkg/provider"
+	seedvalidation "github.com/kubermatic/kubermatic/pkg/validation/seed"
 
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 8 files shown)


youssefazrak:
@zreigz PTAL

PR #5613, opened 2020-07-07
restrict project creation

zreigz:
**What this PR does / why we need it**: Customer would like to be able to restrict users, including Owner, Viewer and Editor roles, from creating projects beneath groups. This would mean that only Admin users can create projects for any group. The reason for this is so that the admins can ensure that naming conventions are adhered to and project permissions are correct for each project created.

Fixes #

```release-note
Restrict project creation for nonadmin users
```

### DIFF
--- a/pkg/handler/v1/project/project.go
+++ b/pkg/handler/v1/project/project.go
@@ -54,6 +54,10 @@ func CreateEndpoint(projectProvider provider.ProjectProvider, privilegedProjectP
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
+		if err := checkProjectRestriction(user, settings); err != nil {
+			return nil, err
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/handler/v1/project/project.go
```diff
@@ -76,6 +80,16 @@ func CreateEndpoint(projectProvider provider.ProjectProvider, privilegedProjectP
 	}
 }
 
+func checkProjectRestriction(user *kubermaticapiv1.User, settings *kubermaticapiv1.KubermaticSetting) error {
+	if user.Spec.IsAdmin {
+		return nil
+	}
+	if settings.Spec.RestrictProjectCreation {
+		return errors.New(http.StatusForbidden, "project creation restricted")
```
maciaszczykm: `project creation is restricted`

PR #5293, opened 2020-04-08
[WIP] Remove apiserver internal service

irozzo-1A:
**What this PR does / why we need it**:

This PR addresses the following points:

- The apiserver pods bind to the same port as the nodeport service port that exposes them externally. Not sure this is necessary.
- The internal service pointing to port 8080 of the apiserver seems to be useless as it binds to the localhost interface, thus not accessible.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
@kdomanski This PR is based on my current understanding of the problem, please let me know if I missed something.

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
```

PR #5621, opened 2020-07-08
Remove apiserver internal service

irozzo-1A:
Apiserver internal service is not useful as it exposes a port that is not open. The service port does not need to match the nodeport.

Replaces #5293 

Fixes #

**Documentation**:

```release-note
Remove unused apiserver internal service.
```


pkg/controller/seed-controller-manager/openshift/openshift_controller_test.go
```diff
@@ -142,7 +142,7 @@ func TestResources(t *testing.T) {
 				},
 				&corev1.Service{
 					ObjectMeta: metav1.ObjectMeta{
-						Name:      "apiserver-external",
+						Name:      "apiserver",
```
youssefazrak: Service name inconsistency compared to the K8s one.

irozzo-1A: That's just the name used in the test so it's harmless, but it's a good starting point for a deeper reflection. It comes from a previous attempt I made where I renamed `apiserver-external` to `apiserver`. This probably makes more sense, but it requires some clean-up logic to avoid port clashes. After better thinking it could be worthy to have it.

youssefazrak:
Finished testing on 2.14.3. Working fine, nodeport-proxy is forwarding correctly the traffic of userclusters to the control planes.
/lgtm

PR #5626, opened 2020-07-09
Fix incorrect error handling in collectAddon function

phiphi282:
There is a return statement missing, when handling a broken or missing ownerReference in the collectAddon function.
Without this, the controller will print an error message and panic.

**Documentation**:

```release-note
NONE
```


xrstf:
/ok-to-test
/test all
/approve
/lgtm

Thank you very much :-)

xrstf:
Oh, backporting this is not actually required, because in 2.14 we determine the cluster name in a different way.

PR #5640, opened 2020-07-12
seed-controller-manager: disable-leader-election option

multi-io:

This adds a --disable-leader-election option to the seed-controller-manager, which can be handy during debugging sessions where you don't want your worker to be cancelled by leader election losses.

```release-note
seed-controller-manager: -disable-leader-election option
```

### DIFF
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -256,26 +256,32 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		leaderCtx, stopLeaderElection := context.WithCancel(rootCtx)
 		defer stopLeaderElection()
 
+		runner := func(ctx context.Context) error {
+			log.Info("Executing migrations...")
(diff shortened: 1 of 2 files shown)


multi-io:
(force push to fix linter error)

xrstf:
@irozzo-1A you're working on the master ctrlmgr; is this conflicting with the direction you're heading? Otherwise I would tend to approve this.

irozzo-1A:
> @irozzo-1A you're working on the master ctrlmgr; is this conflicting with the direction you're heading? Otherwise I would tend to approve this.

I'm doing some refactoring and I incorporated this requirement into #5683. I'm not done yet, but if my PR gets merged we don't need this any longer.

I would hold until tomorrow. I'm not far from from finishing. In case it will take me longer I will take care of this ;-)

irozzo-1A:
@multi-io closing as I introduced this feature here: https://github.com/kubermatic/kubermatic/pull/5683

PR #5668, opened 2020-07-16
Etcd backups and restore controllers

multi-io:

EtcdBackupConfig CRD and controller.

EtcdRestore CRD and controller.

Disabled by default; enabled via new CLI flag `--enable-etcd-backups-restores`. See "Enabling it" below for details.

**Backups**

This provides this CRD for the seed clusters:

```
kind: EtcdBackupConfig
metadata:
  name: mybackup
  namespace: cluster-zvc78xnnz7
  ownerReferences:
  - apiVersion: kubermatic.k8s.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Cluster
    name: zvc78xnnz7
    uid: ....
spec:
  cluster:
    kind: Cluster
    name: zvc78xnnz7
    uid: ....
  name: mybackup
  schedule: '*/3 * * * *'
  keep: 4
EOF
```

You create this in the cluster-xxxx namespace, and the controller will start a new backup every time the schedule demands it (i.e. every 3 minutes in the example), and delete backups once more than .spec.keep (4 in the example) have been created.

Backup creations and deletions are done by spawning k8s jobs: When a backup needs to be created, the controller gives it a name like $backupconfigname-$timestamp, where $backupconfigname is the name of the CRD instance ("mybackup" in the example), and timestamp is the timestamp, and it then creates a job with an init container that creates a snapshot and a main container that will run the customizable store container. It will receive the cluster name and backup name in two environment variables. The store container must upload the snapshot somewhere and return exit code 0 if successful. When a backup needs to be deleted, the controller starts another job with a customizable delete container, which again receives the cluster and backup names in env variables and must delete the backup at the place where it was uploaded by the create job earlier, and exit 0 if successful. The controller maintains an array `.status.lastBackups` where it tracks the creation and deletion status of all backups that haven't been successfully deleted yet.

See below for more details on the store and delete containers.

You can leave out the .spec.schedule and .keep fields, in which case it'll create just one backup, immediately (named just $backupconfigname, i.e. no timestamp). This can be used e.g. when you need to backup some or all clusters right now.

**Restores**

Only supported on S3 for now, as we discussed.

New CRD:

```
kind: EtcdRestore
metadata:
  name: mybackup-restore
  namespace: cluster-zvc78xnnz7
  ownerReferences:
  - apiVersion: kubermatic.k8s.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Cluster
    name: zvc78xnnz7
    uid: ...
spec:
  name: mybackup-restore
  cluster:
    kind: Cluster
    name: zvc78xnnz7
    uid: ...
  backupName: mybackup-2020-09-02T15:24:00
```

You create that in the cluster-xxx namespace, and the restore controller will:

- check if the backup is accessible:

  - there is an optional field `spec.backupDownloadCredentialsSecret` in the crd, which you can set to a secret in the cluster-xxx namespace container S3 credentials, bucket name and endpoint

  - if that field is NOT set, the controller will dynamically create such a secret and point the field to it. The values of the secret will be set from the (backwards-compatible) kube-system/s3-credentials secret for the credentials and from the kube-system/s3-settings configmap for the endpoint and bucket name

- pause the cluster

- delete the etcd sts & volumes

- set the `EtcdClusterInitialized` cluster condition to false (this is the one that until now is only ever set to true once etcd becomes healthy, and never back to false again)

- unpause the cluster and wait until the cluster controller has recreated the etcd sts (see next section for what the launcher does here) and `EtcdClusterInitialized` becomes true again (which also means the etcd cluster was rebuilt with whatever componentOverride.etcd settings the user set in the cluster resource).

Meanwhile the launcher pods will notice that there is an active (.status.phase==stsRebuilding) EtcdRestore resource in the seed, and download the backup named in .spec.backupName from S3 and initialize the etcd datadir from it (using the `etcdrestore.spec.backupDownloadCredentialsSecret` field to obtain S3 credentials, endpoint and bucket name

**Backup store and delete containers**

The restore controller is currently fixed to use S3 only. (we agreed that this is ok for now)

The backup controller spawns k8s jobs to perform the backups and backup deletions. The init container of the backup jobs creates the snapshot and puts it in a well-known place in the filesystem (/backups/snapshot.db) that's shared with the main container. The main container can be customized by the Kubermatic admin (similar to the old backup controller, to which it should be compatible -- see below). It receives the name to upload the backup as in an env variable ($BACKUP_TO_CREATE). It also receives the cluster name in $CLUSTER (like with the old backup controller). Sample container that uploads to s3:

```
command:
- /bin/sh
- -c
- |
  set -e
  s3cmd --access_key=$ACCESS_KEY_ID --secret_key=$SECRET_ACCESS_KEY put /backup/snapshot.db s3://some-backups-bucket/$CLUSTER-$BACKUP_TO_CREATE
image: d3fk/s3cmd:latest
name: store-container
env:
- name: ACCESS_KEY_ID
  valueFrom:
    secretKeyRef:
      name: s3-credentials
      key: ACCESS_KEY_ID
- name: SECRET_ACCESS_KEY
  valueFrom:
    secretKeyRef:
      name: s3-credentials
      key: SECRET_ACCESS_KEY
volumeMounts:
- name: etcd-backup
  mountPath: /backup
```

Similarly, backup deletions are also performed in a job, also with a customizable container (passed to the seed-controller-manager via new optional commandline arg `--backup-delete-container`). The name of the backup to delete will be passed in env variable $BACKUP_TO_DELETE. Sample container corresponding to the above store container:

```
command:
- /bin/sh
- -c
- |
  s3cmd --access_key=$ACCESS_KEY_ID --secret_key=$SECRET_ACCESS_KEY del s3://some-backups-bucket/$CLUSTER-$BACKUP_TO_DELETE
  case $? in
  12)
    # backup no longer exists, which is fine
    exit 0
    ;;
  0)
    exit 0
    ;;
  *)
    exit $?
    ;;
  esac
image: d3fk/s3cmd:latest
name: store-container
env:
- name: ACCESS_KEY_ID
  valueFrom:
    secretKeyRef:
      name: s3-credentials
      key: ACCESS_KEY_ID
- name: SECRET_ACCESS_KEY
  valueFrom:
    secretKeyRef:
      name: s3-credentials
      key: SECRET_ACCESS_KEY
```

The delete container is optional. If none is specified, no delete jobs will be spawned. In that case, the store container must handle old backup deletion by itself. This is mainly for compatibility with existing customized store containers -- these should work just as they do now -- so they won't use the $BACKUP_TO_CREATE variable and thus will upload the backup under a different name that they choose themselves, rather than the one chosen by the controller. The new controller also supports the cleanup container that the old one supports: If a cleanup container is configured (it's optional now), it is used when deleting/finalizing etcdbackupconfig resources (i.e. delete the backups). All in all this should make the new backup controller a fully compatible drop-in replacement for the old one. For new setups using a cleanup container isn't recommended, as cleanups there will be handled using the delete container -- i.e. when deleting/finalizing an etcdbackupconfig resource, the controller will spawn delete jobs for all the backups that still remain.

The controller keeps track of the performed backups, their jobs and statuses in the .status field of the EtcdBackupConfig resource. Example:

```
...
kind: EtcdBackupConfig
metadata:
  creationTimestamp: "2020-09-16T00:13:38Z"
  name: jobtest1
  namespace: cluster-zvc78xnnz7
  ownerReferences:
  - apiVersion: kubermatic.k8s.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Cluster
    name: zvc78xnnz7
    uid: 3a2a5b58-b958-48e1-af1b-030c8bff16c1
spec:
  cluster:
    kind: Cluster
    name: zvc78xnnz7
    uid: 3a2a5b58-b958-48e1-af1b-030c8bff16c1
  keep: 4
  name: jobtest1
  schedule: '*/3 * * * *'
status:
  lastBackups:
  - backupFinishedTime: "2020-09-16T01:48:22Z"
    backupName: zvc78xnnz7-jobtest1-2020-09-16T03:48:00
    backupPhase: Completed
    deleteFinishedTime: "2020-09-16T02:01:43Z"
    deleteJobName: zvc78xnnz7-backup-jobtest1-delete-s8sb8mdsp5
    deletePhase: Completed
    jobName: zvc78xnnz7-backup-jobtest1-create-rjjq9d4k6d
    scheduledTime: "2020-09-16T01:48:00Z"
  - backupFinishedTime: "2020-09-16T01:51:25Z"
    backupName: zvc78xnnz7-jobtest1-2020-09-16T03:51:00
    backupPhase: Completed
    deleteFinishedTime: "2020-09-16T02:04:08Z"
    deleteJobName: zvc78xnnz7-backup-jobtest1-delete-wcwvc8nmdt
    deletePhase: Completed
    jobName: zvc78xnnz7-backup-jobtest1-create-96tblznngk
    scheduledTime: "2020-09-16T01:51:00Z"
  - backupFinishedTime: "2020-09-16T01:54:21Z"
    backupName: zvc78xnnz7-jobtest1-2020-09-16T03:54:00
    backupPhase: Completed
    deleteFinishedTime: "2020-09-16T02:08:24Z"
    deleteJobName: zvc78xnnz7-backup-jobtest1-delete-cgmqhlw5fw
    deletePhase: Completed
    jobName: zvc78xnnz7-backup-jobtest1-create-v2gz2ntn46
    scheduledTime: "2020-09-16T01:54:00Z"
...
```

**Default EtcdBackupConfig resources**

A default EtcdBackupConfig resource is created for each cluster, with the schedule set to "@every X", with X being 20m by default, overridable via the normal `--backup-interval` cli parameter that's used for the old backups too.

**Enabling it**

There's a new boolean CLI flag `--enable-etcd-backups-restores`, which defaults to false. Only if this is set, the new backup and restore controller will be enabled. At the same time the old backup controller will be disabled and its cronjobs will be deleted (not the old backups though).

Fixes #

- EtcdBackup and EtcdRestore resources: namespaced, created in cluster namespace, redundancy metadata.name/spec.name and metadata.ownerReferences[0]/spec.cluster. Similar structure to `addon` resources. Is this OK? Do we want to get rid of at least one of those cluster references? Do we want a MutatingWebhookConfiguration that ensures the resources are "well formed"?

- have a look at `ClusterReconcileWrapper()` -- I modified that a little bit to make the "set a cluster condition" part optional. This was needed to make the function usable in the backup controller, which doesn't set a cluster condition (because etcdbackup resources are supposed to be created continuously, there's not really one time when a "backups reconciled" condition could be set on the cluster)

- backup reconciliation is split linearly into 5 functions, essentially for better readability. Each of them patches the backupConfig.status at the end. We could probably do away with some of those patches (ie. modify the config only in memory in several functions and then perform one patch at the end). Would have to be careful though to not create any jobs that might then be "lost" because they were not recorded in backupConfig.status after a patch failure

- I'm using Update() instead of Patch() to update etcdbackupconfig resources because Patch() or the associated ctrlruntimeclient.MergeFrom(oldBackupConfig) sometimes mysteriously messed up the etcdbackupconfig.status.lastBackups array. Update() sometimes fails with the well-known error "object has been modified; please apply your changes to the latest version and try again", which is benign but can spam the controller logs and events, which is why I've added special handling in the reconcile function so it only debug-logs this error and doesn't generate an event

**Documentation**:

```release-note
New etcd backup and restore controllers
```


multi-io:
force-push small change -- removed apiserver heathy check. We want to still be able to create backups when the apiserver is down.

charts/kubermatic/templates/kubermatic-controller-manager-dep.yaml
```diff
@@ -95,6 +95,11 @@ spec:
         {{- else }}
         - -openshift-addons-list={{ join "," .Values.kubermatic.controller.addons.openshift.defaultAddons }}
         {{- end }}
+        - -backup-snapshot-dir=/backup-snapshots
+        - -backup-s3-endpoint={{ .Values.kubermatic.controller.etcdBackups.s3Endpoint }}
+        - -backup-s3-bucket={{ .Values.kubermatic.controller.etcdBackups.s3Bucket }}
+        - -backup-s3-access-key={{ .Values.kubermatic.controller.etcdBackups.s3AccessKeyId }}
+        - -backup-s3-secret-access-key={{ .Values.kubermatic.controller.etcdBackups.s3SecretAccessKey }}
```
moelsayed: We probably shouldn't pass configuration as args or envvars in this case, as it would require a redeploy to change them. Maybe a configmap and secret for the credentials. This way it will be possible for external containers to access them as well.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,348 @@
...
+
+	cluster := &kubermaticv1.Cluster{}
+	if err := r.Get(ctx, types.NamespacedName{Name: backup.Spec.Cluster.Name}, cluster); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	log = r.log.With("cluster", cluster.Name, "backup", backup.Name)
+
+	// Add a wrapping here so we can emit an event on error
+	result, err := kubermaticv1helper.ClusterReconcileWrapper(
```
moelsayed, line 186: We can probably use this to set a condition with the last backup status for the cluster.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,348 @@
...
+	if result == nil {
+		result = &reconcile.Result{}
+	}
+	return *result, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, backup *kubermaticv1.EtcdBackup, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+	if backup.DeletionTimestamp != nil {
+		log.Debug("Cleaning up all backups")
+		return nil, wrapErrorMessage("error cleaning up all backups: %v", r.deleteAllBackups(ctx, log, backup))
```
moelsayed: I am not sure if we should delete _all_ backups when the config is removed. Maybe have a flag in the CRD spec for this?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	durationToNextBackup := schedule.Next(lastBackupTime.Time).Sub(r.clock.Now())
+	if durationToNextBackup < 0 {
+		durationToNextBackup = 0
+	}
+	return &reconcile.Result{Requeue: true, RequeueAfter: durationToNextBackup}, nil
+}
+
+func (r *Reconciler) createBackup(ctx context.Context, log *zap.SugaredLogger, fileName string, backup *kubermaticv1.EtcdBackup, cluster *kubermaticv1.Cluster) error {
```
moelsayed: As we discussed offline, we shouldn't do this inside the seed controller, instead we will keep using the external container pattern.

pkg/crd/kubermatic/v1/etcd_backup.go
```diff
@@ -0,0 +1,89 @@
...
+type EtcdBackupList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []EtcdBackup `json:"items"`
+}
+
+type EtcdBackupStatus struct {
+	LastBackupTime *metav1.Time `json:"lastBackupTime,omitempty"`
+	CurrentBackups []string     `json:"lastBackups,omitempty"`
```
moelsayed: I think it would be nice if we could have name, timestamp, status of the backup. Would make easy to list in the UI.

moelsayed (commented):
Thank you so much. I put a few inline comments. I would like to also suggest:

- We should automatically create the Backup object with sensible defaults. I like the idea of using the old backup controller to this. For existing clusters, we could check if it has a backup cronjob (old etcd backup controller), if so, we pick up the cronjob config, use it to create the backup object and delete the cronjob. Thus, migrating the cluster to the new backup controller. For new clusters, we simply create the backup object instead of the cronjob.
- Use jobs instead of running inside the seed controller. We should track the job execution/completion status in the backup list in the backup object status.
- We should tls auth for etcd access when needed. 
- As you suggested, we should rename the CRD to reflect it's current function.

multi-io:
> * Use jobs instead of running inside the seed controller. We should track the job execution/completion status in the backup list in the backup object status.

I agree that scalability will be better if we have separate jobs, and that the customizable storage container is nice. Just to confirm -- the backup controller would still evaluate the (optional) schedule field of the CRD, but then whenever a backup needs to be created it would create a job resource rather than creating the backup in-process in the seed controller manager. Right?

pkg/crd/kubermatic/v1/etcd_backup.go
multi-io: Yeah. I didn't even think of the whole UI/API access angle. 😂

moelsayed:
> Just to confirm -- the backup controller would still evaluate the (optional) schedule field of the CRD, but then whenever a backup needs to be created it would create a job resource rather than creating the backup in-process in the seed controller manager. Right?

Yes, exactly.

multi-io:
I thought a little more about changing the backup controller so it runs the backups in k8s jobs, while also supporting a customizable storage container for uploading the snapshots like the current upstream backup controller does. This can be done, and I think it'd be quite straightforward to implement, but I'm not sure we can keep the storage container backwards compatible with the ones the current upstream backup controller uses:

The "protocol" between the current upstream backup controller and the storage container is very simple: The storage container knows that the snapshot to upload is located at a well-known place in the local filesystem (/backups/snapshot.db). It can then upload it wherever it wants, and it handles upload object names (including timestamps) and old backup deletion entirely at its own discretion. The rest of the system doesn't know these things.

With the new backup controller we have tighter requirements now:

- we want to track uploaded object names (i.e. "backup names") in the CRD

- the controller needs some way of initiating old backup deletion

- we want to support instant one-time backups

- for regular scheduled backups, the number of backups to keep is configurable in the CRD (.spec.keep) and thus the controller must be able to control this

Design proposal:

- each k8s job basically uploads the snapshot to a controller-defined object name and also deletes some other already uploaded objects whose names the controller defines

- just like current upstream, the k8s job would contain an init container that takes the snapshot and puts it into a well-known location (/backup/snapshot.db) that's shared with the main container.

- the main container is the storage container that uploads the backup to some safe backend (object or other) storage. May be s3 or something else. The container is user-definable, but must adhere to a small "protocol" defined as follows:

  - just like with the current upstream, the snapshot to upload exists at /backup/snapshot.db when the container is started

  - controller passes (via env variables injected into the container):

    - full object name to upload snapshot as (full object name includes timestamp and everything, if applicable)

    - list of full object names to delete

I guess in this case, since the controller defines all the backup names and scheduling, the store container wouldn't really need to run the s3-storeuploader binary/image anymore; instead, a simple s3cmd image running a shell script would suffice.

moelsayed:
@multi-io This looks great. 

I just want to make sure that we don't break existing custom setups, so just a small clarification:
If I understand correctly, the second point in the protocol you defined means that the new controller will simply pass this information to the main container, and the main container will do the work. This way, even if the user custom container doesn't respect this configuration, the process will not fail, correct?

If that is the case, we ca go ahead with this proposal :+1:

multi-io:
> @multi-io This looks great.
> 
> I just want to make sure that we don't break existing custom setups, so just a small clarification:
> If I understand correctly, the second point in the protocol you defined means that the new controller will simply pass this information to the main container, and the main container will do the work. This way, even if the user custom container doesn't respect this configuration, the process will not fail, correct?
> 
> If that is the case, we ca go ahead with this proposal 👍

I guess existing storage containers would kinda work, but wouldn't adhere to the settings in the CRD, i.e. the backup name or keep count. They would basically continue doing what they do now, i.e. they'd upload the backup as `$clustername-storeuploader-$timestamp-snapshot.db` and perform their own backup deletion, just like they do now (or do whatever the customer decided to do in their own customized storage container defition). There may also be issues if you have more than one backup config for the same cluster, which the current upstream backup controller doesn't do and thus the storage containers aren't prepared for. I.e. maybe those different customer backup containers belonging to different backup configs would then delete each other's backups.

Thanks for the heads-up, I'll implement this then, with an eye on at reasonable compat w/ existing backup containers.

moelsayed:
@multi-io The limitations caused by users using their own custom images (manage deletion on their own, upload their backups..etc) are fine, as long as they continue doing what their doing with no problem.

The multiple config issue is interesting. I don't think it would be an issue for users using default containers, since it's wasn't support initially. For users with custom containers, we can clarify this in the release notes.

cmd/etcd-launcher/main.go
```diff
@@ -226,6 +243,22 @@ func (e *etcdCluster) parseConfigFlags() error {
...
+	}
+
+	if config.backupS3AccessKeyID == "" {
+		return errors.New("s3 access key ID is not set")
+	}
+
+	if config.backupS3SecretAccessKey == "" {
+		return errors.New("s3 secret access key is not set")
+	}
+
```
moelsayed: I don't think we should always fail here. We should fail on these if we are in restore mode only since some users might not use this restore path.

cmd/etcd-launcher/main.go
```diff
@@ -409,3 +442,64 @@ func (e *etcdCluster) removeDeadMembers(log *zap.SugaredLogger) error {
...
+	objectName := fmt.Sprintf("%s-%s", k8cCluster.GetName(), activeRestore.Spec.BackupName)
+	downloadedSnapshotFile := fmt.Sprintf("/tmp/%s", objectName)
+
+	if err := s3Client.FGetObject(e.config.backupS3BucketName, objectName, downloadedSnapshotFile, minio.GetObjectOptions{}); err != nil {
+		return fmt.Errorf("failed to download backup (%s/%s): %v", e.config.backupS3BucketName, objectName, err)
+	}
+
+	if err := os.RemoveAll(e.config.dataDir); err != nil {
```
moelsayed: I think it would be safer to rename it and only delete it once the restore is successful.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: restore.Spec.Cluster.Name}, cluster); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	log = r.log.With("cluster", cluster.Name, "restore", restore.Name)
+
+	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName {
+		return reconcile.Result{}, nil
+	}
+
```
moelsayed, line 137: Does it make sense to use a cluster condition here and update it with the latest restore status?

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	if err := r.updateRestore(ctx, restore, func(restore *kubermaticv1.EtcdRestore) {
+		restore.Status.Phase = kubermaticv1.EtcdRestorePhaseStarted
+		kuberneteshelper.AddFinalizer(restore, RebuildStatefulsetFinalizer)
+	}); err != nil {
+		return nil, fmt.Errorf("failed to add finalizer: %v", err)
+	}
+
+	// delete etcd sts
+	sts := &v1.StatefulSet{}
+	err := r.Get(ctx, types.NamespacedName{Namespace: cluster.Status.NamespaceName, Name: resources.EtcdStatefulSetName}, sts)
```
moelsayed: I think we agreed this would be done only when we have a new storage class set.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		deletePropagationForeground := metav1.DeletePropagationForeground
+		delOpts := &ctrlruntimeclient.DeleteOptions{
+			PropagationPolicy: &deletePropagationForeground,
+		}
+		if err := r.Delete(ctx, &pvc, delOpts); err != nil {
+			return nil, fmt.Errorf("failed to delete pvc %v: %v", pvc.GetName(), err)
+		}
+	}
+
+	if len(pvcs.Items) > 0 {
```
moelsayed, line 248: I think we should List() again before this?

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		kuberneteshelper.RemoveFinalizer(restore, RebuildStatefulsetFinalizer)
+	}); err != nil {
+		return nil, fmt.Errorf("failed to mark restore completed: %v", err)
+	}
+
+	return nil, nil
+}
+
+func (r *Reconciler) updateCluster(ctx context.Context, cluster *kubermaticv1.Cluster, modify func(*kubermaticv1.Cluster)) error {
+	oldBackup := cluster.DeepCopy()
```
moelsayed: oldCluster*

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -159,6 +160,7 @@ const (
 	ClusterConditionAddonControllerReconcilingSuccess          ClusterConditionType = "AddonControllerReconciledSuccessfully"
 	ClusterConditionAddonInstallerControllerReconcilingSuccess ClusterConditionType = "AddonInstallerControllerReconciledSuccessfully"
 	ClusterConditionBackupControllerReconcilingSuccess         ClusterConditionType = "BackupControllerReconciledSuccessfully"
+	ClusterConditionEtcdRestoreControllerReconcilingSuccess    ClusterConditionType = "EtcdRestoreControllerReconciledSuccessfully"
```
moelsayed: Not currently used, we should use it.

pkg/crd/kubermatic/v1/etcd_restore.go
```diff
@@ -0,0 +1,80 @@
...
+	EtcdRestoreResourceName = "etcdrestores"
+
+	// EtcdRestoreKindName represents "Kind" defined in Kubernetes
+	EtcdRestoreKindName = "EtcdRestore"
+
+	// EtcdRestorePhase value indicating that the restore has started
+	EtcdRestorePhaseStarted = "Started"
+
+	// EtcdRestorePhase value indicating that the old Etcd statefulset has been deleted and is now rebuilding
+	EtcdRestorePhaseStsRebuilding = "StsRebuilding"
```
moelsayed, line 35: `EtcdRestorePhaseRebuildingSTS` is more readable?

pkg/resources/etcd/statefulset.go
```diff
@@ -240,6 +247,11 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
 					},
 				},
 			}
+
+			set.Spec.Template.Spec.Containers[0].VolumeMounts = append(set.Spec.Template.Spec.Containers[0].VolumeMounts, resources.GetHostCACertVolumeMounts()...)
```
moelsayed, line 258: is this dup of https://github.com/kubermatic/kubermatic/pull/5668/commits/41a43e33a3f8ffddbc4d2ab8fd0d8d43541fba3c#diff-5e4f08420bfa19f5d6cd42269a047618R340

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	return *result, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, restore *kubermaticv1.EtcdRestore, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+	log.Infof("performing etcd restore from backup %v", restore.Spec.BackupName)
+
+	if restore.Status.Phase == kubermaticv1.EtcdRestorePhaseCompleted {
+		return nil, nil
+	}
+
```
moelsayed, line 156: I think we should try to make sure we have everything we need before starting the destructive actions. We should at least check if we have the settings and the secrets available.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	if err == nil {
+		if err := r.Delete(ctx, sts); err != nil && !kerrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to delete etcd statefulset: %v", err)
+		}
+	} else if !kerrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get etcd statefulset: %v", err)
+	}
+
+	// delete PVCs
+	pvcSelector, err := labels.Parse(fmt.Sprintf("%s=%s", resources.AppLabelKey, resources.EtcdStatefulSetName))
```
moelsayed, line 228: Also this was only in case of new storage class.

moelsayed (commented):
Thank you, looks good :+1: , just a few comments.

cmd/etcd-launcher/main.go
multi-io: Changed it -- the etcdrestore CRD now contains a field `BackupDownloadCredentialsSecret` that names a secret in the cluster-xxx namespace that must contain S3 credentials and endpoint + bucket name. The restore controller, when starting the restore, checks if the field is set, and if it is NOT, it creates a secret in cluster-xxx and copies the data from secret kube-system/s3-credentials and (new) configmap kube-system/s3-settings into it. See PR description for more details.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
multi-io: Since we just deleted at least one PVC (as evidenced by `len(pvcs.Items) > 0`) and the deletion happens asynchronously in the background, I wouldn't expect the PVC to be gone immediately afterwards (and wouldn't it still be in the cache anyway?). That's why I'm just arranging for the reconcile to be called again 30 seconds later, and this repeats until no PVCs are found anymore.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
multi-io: 👍  oldRestore too (next function). Fixed, will push soon.

pkg/resources/etcd/statefulset.go
multi-io: Hm no, one is the volumes, the other is the volume mounts.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
multi-io: The controller now checks that the backup exists in S3 before proceeding here...

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// check that the backup to restore from exists and is accessible
+	s3Client, bucketName, err := resources.GetEtcdRestoreS3Client(ctx, restore, true, r.Client, cluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to obtain S3 client: %w", err)
+	}
+
+	objectName := fmt.Sprintf("%s-%s", cluster.GetName(), restore.Spec.BackupName)
+	if _, err := s3Client.StatObject(bucketName, objectName, minio.StatObjectOptions{}); err != nil {
```
multi-io, line 199: ...this is the check.

charts/kubermatic/templates/kubermatic-controller-manager-dep.yaml
multi-io: Removed the options -- they're no longer used or supported. See the PR description for how it gets this data now.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: Done (see PR description).

pkg/crd/kubermatic/v1/etcd_backup.go
multi-io: Done (see PR description).

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: Not with `ClusterReconcileWrapper`, I don't think. It sets the condition when the reconcile method returned nil,nil -- which the etcd backup controller's reconcile method never will for backup configs with a schedule until the backup config is deleted (because with a schedule set, I'm always returning a "requeueAfter: [time to next scheduled backup]" result, not nil). That's why I thought that aspect of `ClusterReconcileWrapper()` doesn't make too much sense here -- it seems to be basically meant for controllers that reconcile something once per cluster and are then done with it.

If we really wanted to set a condition explicitly every time a backup succeeds, I'd have to add explicit support for it.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
multi-io: As we talked about in Slack briefly, I've postponed this for now because it wasn't so easy to implement reliably. When deleting and immediately recreating all etcd pods, it happened sometimes that data would "spill" from still-running old pods to already-running new pods, so that after the restore the etcd sometimes contained not just data that was in the backup, but also some keys that had been added after the backup was taken. I tried to circumvent this by scaling the sts down to 0 and then up again, but currently the cluster resource controller can't scale the etcd sts up from 0 (look at https://github.com/kubermatic/kubermatic/blob/c764c003af56e1c11090e20d314d5806e7b83645/pkg/resources/etcd/statefulset.go#L353-L384 for the case *set.Spec.Replicas==0 to see this -- actually not sure if this is a bug or not).

I've pushed the "delete all pods" and "scale sts down/up" implementations here
https://github.com/multi-io/kubermatic/blob/etcd-backups-controller-del-sts-pods/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go and here
https://github.com/multi-io/kubermatic/blob/etcd-backups-controller-scaledown-sts/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go in case we want to revisit this.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
multi-io: see above.

pkg/crd/kubermatic/v1/etcd_backup_config.go
```diff
@@ -0,0 +1,150 @@
...
+
+const (
+	// EtcdBackupConfigResourceName represents "Resource" defined in Kubernetes
+	EtcdBackupConfigResourceName = "etcdbackupconfig"
+
+	// EtcdBackupConfigKindName represents "Kind" defined in Kubernetes
+	EtcdBackupConfigKindName = "EtcdBackupConfig"
+
+	DefaultKeptBackupsCount = 20
+	MaxKeptBackupsCount     = 20
```
moelsayed: Can we increase the max? it should be more than the default to allow more/less setting.

pkg/resources/etcd/statefulset.go
```diff
@@ -253,6 +253,11 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
...
+
+			set.Spec.Template.Spec.Tolerations = data.Cluster().Spec.ComponentsOverride.Etcd.Tolerations
```
moelsayed, line 260: Why do we need to override the tolerations?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package etcdbackup
+
+import (
```
moelsayed, line 19: Could you please reorganize the imports:
- stdlib
- 3rd party
- k8c
- k8s

cmd/seed-controller-manager/controllers.go
```diff
@@ -180,6 +184,40 @@ func createKubernetesController(ctrlCtx *controllerContext) error {
...
+		ctrlCtx.runOptions.workerCount,
+		ctrlCtx.runOptions.workerName,
+		storeContainer,
+		deleteContainer,
+		cleanupContainer,
+		ctrlCtx.runOptions.backupContainerImage,
+	)
+}
+
 func createBackupController(ctrlCtx *controllerContext) error {
```
moelsayed, line 244: Should we disable this if the new controller is enabled?

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -190,6 +193,8 @@ var AllClusterConditionTypes = []ClusterConditionType{
 	ClusterConditionClusterControllerReconcilingSuccess,
 	ClusterConditionAddonControllerReconcilingSuccess,
 	ClusterConditionBackupControllerReconcilingSuccess,
+	ClusterConditionEtcdRestoreControllerReconcilingSuccess,
+	ClusterConditionBackupScheduleControllerReconcilingSuccess,
```
moelsayed: These are causing the cluster to get stuck during provisioning. Check the failed tests.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -319,17 +324,21 @@ type APIServerSettings struct {
 }
 
 type DeploymentSettings struct {
-	Replicas  *int32                       `json:"replicas,omitempty"`
-	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+	Replicas    *int32                       `json:"replicas,omitempty"`
+	Resources   *corev1.ResourceRequirements `json:"resources,omitempty"`
+	Tolerations []corev1.Toleration          `json:"tolerations,omitempty"`
```
moelsayed, line 359: Why are we adding this?

pkg/resources/resources.go
```diff
@@ -1097,3 +1117,92 @@ func GetHostCACertVolumeMounts() []corev1.VolumeMount {
 		},
 	}
 }
+
+// GetEtcdRestoreS3Client returns an S3 client for downloading the backup for a given EtcdRestore.
+// If the EtcdRestore doesn't reference a secret containing the credentials and endpoint and bucket name data,
+// one can optionally be created from a well-known secret and configmap in kube-system.
+func GetEtcdRestoreS3Client(ctx context.Context, restore *kubermaticv1.EtcdRestore, createSecretIfMissing bool, client ctrlruntimeclient.Client, cluster *kubermaticv1.Cluster) (*minio.Client, string, error) {
```
moelsayed, line 1142: Does it make sense to move this to the restore controller?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+// "schedule a backup" means set the scheduled time, backup name and job names of the corresponding element of backupConfig.Status.CurrentBackups
+func (r *Reconciler) ensureNextBackupIsScheduled(ctx context.Context, backupConfig *kubermaticv1.EtcdBackupConfig, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+	if backupConfig.DeletionTimestamp != nil || cluster.DeletionTimestamp != nil {
+		// backupConfig is deleting. Don't schedule any new backups.
+		return nil, nil
+	}
+
+	var backupToSchedule *kubermaticv1.BackupStatus
+
+	if len(backupConfig.Status.CurrentBackups) > 2*backupConfig.GetKeptBackupsCount() {
```
moelsayed, line 290: Does this mean that if delete jobs fail for a while, we stop trying to create new backups ?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+			kubermaticv1.EtcdBackupConfigConditionSchedulingActive,
+			corev1.ConditionFalse,
+			"TooManyActiveBackups",
+			"too many active backups already; not scheduling new ones") {
+
+			// condition changed, need to persist and generate an event
+			if err := r.Update(ctx, backupConfig); err != nil {
+				return nil, errors.Wrap(err, "failed to update backup config")
+			}
+			r.recorder.Event(backupConfig, corev1.EventTypeWarning, "TooManyActiveBackups", "too many active backups already; not scheduling new ones")
```
moelsayed: I think `active backups` indicates an ongoing backup process, not an existing snapshot. Maybe we can reword this?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+		kubermaticv1.EtcdBackupConfigConditionSchedulingActive,
+		corev1.ConditionTrue,
+		"",
+		"") {
+
+		// condition changed, need to persist and generate an event
+		if err := r.Update(ctx, backupConfig); err != nil {
+			return nil, errors.Wrap(err, "failed to update backup config")
+		}
+		r.recorder.Event(backupConfig, corev1.EventTypeNormal, "NotTooManyActiveBackups", "backup count low enough; scheduling new backups")
```
moelsayed: Maybe reword to "normal backup count" or something similar.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+
+func (r *Reconciler) limitNameLength(name string) string {
+	if len(name) <= 63 {
+		return name
+	}
+	randomness := r.randStringGenerator()
+	return name[0:63-len(randomness)] + randomness
+}
+
+// set a condition on a backupConfig, return true iff the condition's status was changed
```
moelsayed, line 394: nit: s/iff/if

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+			kept++
+			if kept > keepCount && backup.DeletePhase == "" {
+				backupsToDelete = append(backupsToDelete, backup)
+			}
+		}
+	}
+
+	modified := false
+	for _, backup := range backupsToDelete {
+		if err := r.createDeleteJob(ctx, backupConfig, cluster, backup); err != nil {
```
moelsayed: nit: Can you rename this to `createBackupDeleteJob` or something similar? This name indicates it's a generic function, and it doesn't reflect it's actually updating the backup object. Maybe the new name is slightly clearer?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+
+			if age < retentionTime {
+				// don't delete the job yet, but reconcile when the time has come to delete it
+				returnReconcile = minReconcile(returnReconcile, &reconcile.Result{RequeueAfter: retentionTime - age})
+
+			} else {
+				// delete job
+				job := &batchv1.Job{}
+
+				err := r.Get(ctx, types.NamespacedName{Namespace: metav1.NamespaceSystem, Name: backup.JobName}, job)
```
moelsayed, line 621: Why do we need to get the job first, instead of just delete right away?

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+					}
+					deleteJobDeleted = true
+				case !kerrors.IsNotFound(err):
+					return nil, errors.Wrapf(err, "backup %s: failed to get job %s", backup.BackupName, backup.DeleteJobName)
+				}
+			}
+		}
+
+		if backupJobDeleted && deleteJobDeleted {
+			continue // delete backup from backupConfig.Status.CurrentBackups
```
moelsayed: this comment is a bit unclear. You are not deleting the backup from the list, you are just skipping it.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+
+// Delete backup and delete jobs that have been finished for a while.
+// For backups where both the backup and delete jobs have been deleted, delete the backup status entry too.
+func (r *Reconciler) deleteFinishedBackupJobs(ctx context.Context, backupConfig *kubermaticv1.EtcdBackupConfig, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+	var returnReconcile *reconcile.Result
+
+	var newBackups []kubermaticv1.BackupStatus
+	for _, backup := range backupConfig.Status.CurrentBackups {
+		// if the backupConfig is being deleted and this backup hasn't even started yet,
+		// we can just delete the backup from backupConfig.Status.CurrentBackups directly
```
moelsayed, line 591: also a bit unclear, please check below.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
```diff
@@ -0,0 +1,983 @@
...
+			"failed to reconcile etcd backup config %q: %v", backupConfig.Name, err)
+	}
+	if result == nil {
+		result = &reconcile.Result{}
+	}
+	return *result, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, backupConfig *kubermaticv1.EtcdBackupConfig, cluster *kubermaticv1.Cluster) (*reconcile.Result, error) {
+	if err := r.ensureEtcdClientSecret(ctx, cluster); err != nil {
```
moelsayed, line 225: Why do we need this ?

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,312 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package etcdrestore
+
+import (
```
moelsayed, line 19: Same as above.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -0,0 +1,312 @@
...
+	if err := r.updateCluster(ctx, cluster, func(cluster *kubermaticv1.Cluster) {
+		cluster.Spec.Pause = true
+	}); err != nil {
+		return nil, fmt.Errorf("failed to pause cluster: %v", err)
+	}
+
+	if err := r.updateRestore(ctx, restore, func(restore *kubermaticv1.EtcdRestore) {
+		restore.Status.Phase = kubermaticv1.EtcdRestorePhaseStarted
+	}); err != nil {
+		return nil, fmt.Errorf("failed to add finalizer: %v", err)
```
moelsayed: incorrect error message.

moelsayed (changes requested):
Looks great, thank you. Just a few small comments.

moelsayed:
@multi-io I think these should also be included in this PR:
```
if the flag is set, kubermatic must:
-  disable old backup controller
-     enable new one
-     create one etcdbackupconfig per cluster that uses the same schedule that the cronjob had
-     in the helm chart, if the deleteContainer is not set, install a default one from the static/ directory just like with the store and cleanup containers for the old backup controller
```

pkg/resources/etcd/statefulset.go
multi-io: Ah, we need that in our fork because we run the etcd pods on special local storage nodes now which are tainted so nothing else is scheduled on them. So the etcd pods need to tolerate those taints. So yeah it feels a bit arbitrary to have tolerations overridable in the etcd settings only, but it's what we needed. I you want I can remove it from the PR.

multi-io:
> Looks great, thank you. Just a few small comments.

Thanks. I'm on vacation for 2 weeks, might take a bit longer to answer. But I'm working on it.

pkg/resources/resources.go
multi-io: It's needed in the launcher too, that's why I put it here.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: Yes.

Basically delete jobs are the only things that know how to delete a backup, so if they fail, we can't be sure if there are remnants left behind in the object store. So I'm retrying delete jobs until they succeed, and only then remove the backup entry from `.Status.CurrentBackups`. So if they keep failing (e.g. because there's something wrong with the seed cluster), entries pile up in .Status.CurrentBackups, so at some point I stop scheduling new backups and set the `SchedulingActive` condition to false on the backupconfig to indicate this. Everything will recover automatically once the delete jobs start succeeding. I basically wanted to avoid a situation where an administrator would have to clean up backups manually in the object store after a problem with the seed cluster prevented jobs from running for a while.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: https://mathworld.wolfram.com/Iff.html (used in K8s source code ~40 times, so some people use it, but I've noticed many others don't know it  -- I'll reword it if you insist :) )

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: I copied this from the old backup controller (where the function is named `ensureCronJobSecret`). It creates the etcd client certificate and private key for the cluster, to be used by the backup jobs. At least one controller needs to do it I figure.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: I'm skipping it in newBackups, but since I'm writing that to backupConfig.Status.CurrentBackups in the end, I'll end up deleting it from there. I'll try to make that more explicit in the comment.

cmd/seed-controller-manager/controllers.go
multi-io: OK as discussed below I'll do that.

pkg/crd/kubermatic/v1/cluster.go
multi-io: :-O OK. I'm not setting/using those conditions at all. So we should either delete them or use them -- not sure how. Set it to true once the first backup in the cluster has succeeded? Backups aren't really something that's reconciled once per cluster and then never updated again, so it doesn't really make sense to have that condition? For restores it may make a bit more sense, but not sure there either.

pkg/crd/kubermatic/v1/cluster.go
multi-io: See above. It's kinda arbitrary, but wee need it in our fork to override resources and tolerations for the etcd pods.

pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go
multi-io: I didn't find a delete API that doesn't need to be given a full object, at least not in controller-runtime. What would be the best way to delete the job given just the name and namespace?

pkg/crd/kubermatic/v1/cluster.go
multi-io: Looked at it again -- the new backup controller uses the old backup controller's condition (ClusterConditionBackupControllerReconcilingSuccess) [here](https://github.com/multi-io/kubermatic/blob/etcd-backups-controller/pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go#L184). So it sets it to true the first time a backup was reconciled successfully for the cluster. I.e. it won't be set to true until a backupconfig is created for the cluster -- which it will be once I've added the code that adds a default one for every cluster, but it still feels wrong to rely on that, right? So it comes back to what I wrote in the PR description:

```
what ClusterReconcileWrapper() is doing here probably makes no sense for BackupController:

SetClusterCondition(cluster, conditionType, reconcilingStatus, "", "")

(because etcdbackup resources are supposed to be created continuously, there's not really
one time when the controller's reconciliation is "finished" for this cluster. Maybe make this
activity optional in ClusterReconcileWrapper()?
```

So I've now removed the unused conditions and also extended `ClusterReconcileWrapper()` a little bit so the "set a cluster condition" part is optional, and I've adapted the new backup controller so it no longer sets a condition. If some other solution is preferred, I can change it back easily.

multi-io:
Rebased and addressed comments. Changes added after your review comments start at commit 48cece643b06392bf56c70e33b97df71cc553bb9

multi-io:
rebased. New stuff starts at f829ad5a7b68cc1513ac96972f2548e5c82a9e7e

moelsayed:
@multi-io Thank you for the update. Looks pretty good! Could you please rebase, update the description to reflect the latest changes and remove the WIP?

charts/kubermatic/values.yaml
```diff
@@ -268,12 +268,19 @@ kubermatic:
...
   deleteContainer: null
   cleanupContainer: null
+  # Settings for the the controllers for EtcdBackupConfig and EtcdRestore CRDs,
+  # which allow dynamically creating multiple arbitrary backup schedules for each
+  # cluster, as well as restoring a cluster's etcd from any of the created backups
+  etcdBackupsRestores:
+    enabled: false
+    s3:
+      endpoint: minio.minio.svc.cluster.local:9000
+      bucket: kubermatic-etcd-backups
```
moelsayed, line 283: I don't think this is supported anymore. We moved to using the kubermatic-installer now, which uses the [`KubermaticConfiguration`](https://github.com/kubermatic/kubermatic/blob/master/pkg/crd/operator/v1alpha1/configuration.go) CRD and runs the operator with it. You probably should add the needed configuration into the CRD and update the operator chart instead to reflect the changes.

moelsayed (changes requested):

moelsayed:
@multi-io do we have a specific reason to running the backup and restore jobs in kube-system?

charts/kubermatic/static/store-container-new.yaml
```diff
@@ -0,0 +1,34 @@
+# This file has been generated using hack/update-kubermatic-chart.sh, do not edit.
+
+name: store-container
+image: d3fk/s3cmd@sha256:2061883abbf0ebcf0ea3d5d218558c9c229f212e9c08af4acdaa3758980eb67a
+command:
+- /bin/sh
+- -c
+- |
+  set -e
+  s3cmd --access_key=$ACCESS_KEY_ID --secret_key=$SECRET_ACCESS_KEY --host=$ENDPOINT --host-bucket='%(bucket).'$ENDPOINT put /backup/snapshot.db s3://$BUCKET_NAME/$CLUSTER-$BACKUP_TO_CREATE
```
moelsayed: Our current defaults to the same behavior as s3cmd's `--no-check-certificate` option. So we should should default to it here, too. It would be awesome if we could make that configurable as well!

multi-io:
> @multi-io do we have a specific reason to running the backup and restore jobs in kube-system?

As we talked about briefly in slack, it's because the s3 upload and delete container usually need to access the s3 credentials in the kube-system/s3-credentials (which is also why the old backup controller's cronjobs exist in kube-system). I guess even if the admin uses their own customized upload/delete containers they might want them to read a central credentials secret, which is easier if they all run in the same namespace. So it's kind of a "user friendliness" (for the admin, not the end user) thing too. As you said, we can potentially change this later though.

pkg/crd/kubermatic/v1/cluster.go
multi-io: I removed the two constants again -- i.e. not setting any cluster conditions for now. Rationale for the backup controller is discussed under "Special notes for your reviewer" above. For the restore controller, I thought it doesn't make too much sense to have a "restore complete" cluster condition because many clusters won't ever be restored from a backup. To be discussed though.

charts/kubermatic/static/store-container-new.yaml
multi-io: Done. (not configurable for now)

multi-io:
Had to rebase again because of 1-line conflict in the chart version. The last 3 commits are new.

multi-io:
Updated the PR description.

charts/kubermatic/values.yaml
moelsayed: @multi-io Is this resolved? I think this is the only remaining point in the PR.

multi-io: Sorry I overlooked that. Will work on it asap. I may need some help because we're not using the operator currently.

multi-io:
Forgot --signoff, hence the repush of the last 2 commits. Those two are also the only new ones -- all the others I just rebased.

cmd/etcd-launcher/main.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/minio/minio-go"
```
zreigz: Please  fix imports also in other places. They should be in order:
- golang
- other 
- kubermatic
- k8s

zreigz (changes requested):

moadqassem (commented):
/approve
/lgtm

PR #5681, opened 2020-07-20
Fix for panic when user/settings watch connection closes

lsviben:
fix for the panic that happens on user/settings watch connection close

```release-note
NONE
```

### DIFF
--- a/pkg/handler/websocket/common.go
+++ b/pkg/handler/websocket/common.go
@@ -23,7 +23,7 @@ import (
 )
 
 // writes close control message to the websocket connection. Same as the default websocket close handler.
-func writeCloseMessage(ws *websocket.Conn, code int) {
+func writeCloseMessage(ws *websocket.Conn, code int) error {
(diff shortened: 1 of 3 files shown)


pkg/handler/websocket/common.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+// writes close control message to the websocket connection. Same as the default websocket close handler.
+func writeCloseMessage(ws *websocket.Conn, code int) {
+	message := websocket.FormatCloseMessage(code, "")
+	_ = ws.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
```
floreks: I guess this returns potential err, wouldn't it be better to return it instead of returning `nil`?

lsviben: you are right, i did it automatically like websocket does it, returning err is better

PR #5685, opened 2020-07-20
Add current WIP for the installer

xrstf:
To make it easier to track progress and keep the installer in-sync with the remaining Kubermatic codebase, this PR moves the code from the `new-installer` branch of the old installer repo into this repo.

This is a heavy WIP and currently can install Helm charts and CRDs, but nothing more. Some things are still hardcoded, some things are still missing from the old-old installer (e.g. DNS hints).

```release-note
NONE
```


Gopkg.toml
```diff
@@ -223,3 +223,7 @@ required = [
 [[constraint]]
   name = "go.etcd.io/etcd"
   version = "3.4.9"
+
+[[constraint]]
+  name = "github.com/sirupsen/logrus"
```
xrstf: I chose logrus because it was easier to have pretty, well-indented log output. The installer is something that should, in my opinion, be pleasing to the eyes and nice looking.

cmd/installer/cmd_deploy.go
```diff
@@ -0,0 +1,236 @@
...
+
+	content, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	raw := &unstructured.Unstructured{}
+	if err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(content), 1024).Decode(raw); err != nil {
+		return nil, nil, fmt.Errorf("failed to decode %s: %v", filename, err)
+	}
```
xrstf: Need the raw config in order to not accidentally created lots of empty objects in the cluster. Loading the config and unmarshalling it directly into a KubermaticConfiguration would interfere with the defaulting the operator does later.

cmd/installer/cmd_deploy.go
```diff
@@ -0,0 +1,236 @@
...
+
+	greetings := []string{
+		"Have a nice day!",
+		"Time for a break, maybe? ☺",
+		"✌",
+		"Thank you for using Kubermatic ❤",
+	}
+
+	return greetings[rand.Intn(len(greetings))]
+}
```
xrstf: `kind` does something similar and I like it.

cmd/installer/helm/cli.go
```diff
@@ -0,0 +1,164 @@
+package helm
```
xrstf: Importing Helm as a Go package is much more difficult (though maybe more controlled) than just shelling out to a `helm` binary, hence this tiny wrapper.

cmd/installer/stack/kubermatic/stack.go
```diff
@@ -0,0 +1,258 @@
...
+	CertManagerReleaseName = CertManagerChartName
+	CertManagerNamespace   = CertManagerChartName
+
+	DexChartName   = "oauth"
+	DexReleaseName = DexChartName
+	DexNamespace   = DexChartName
+
+	KubermaticOperatorChartName   = "kubermatic-operator"
+	KubermaticOperatorReleaseName = KubermaticOperatorChartName
+	KubermaticOperatorNamespace   = "kubermatic"
```
xrstf: TODO: Not have this hardcoded.

docs/proposals/kubermatic-installer.md
```diff
@@ -0,0 +1,128 @@
...
+speed for ease of use.
+
+During `./installer deploy`, the chosen "stack" is installed. By default (and currently
+because it's the only one) this is the "kubermatic" stack. The stack then does whatever
+it needs to run. For Kubermatic this means
+
+* setting up a "kubermatic-fast" StorageClass; this is a leftover from very early installer
+  versions and requires us to know the cloud provider. This requires yet another setting
+  to be made from the user, so maybe we should instead just inform the user that they need
+  to create a StorageClass?
```
youssefazrak, line 107: From the code there is already an implementation for the storageClass. I would say we keep it this way and determine which cloud provider the user is using. Obviously a user might install it somewhere else than one of the 3 major cloud providers (Azure, AWS, Gcloud) and in this case, we need to inform the inform the user that he needs to provision it manually.

pkg/install/stack/kubermatic/stack.go
```diff
@@ -0,0 +1,274 @@
...
+	if err == nil {
+		logger.Info("✅ StorageClass exists, nothing to do.")
+		return nil
+	}
+
+	if !kerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to check for StorageClass %s: %v", StorageClassName, err)
+	}
+
+	sc := storageClassForProvider(StorageClassName, "gke")
```
youssefazrak, line 111: Right now, gke is hardcoded which would need to be changed for a production version.
Tested with eks in my dev env by overwriting the value according to the storageClass defined, works fine.

pkg/install/stack/kubermatic/stack.go
```diff
@@ -0,0 +1,274 @@
...
+		return fmt.Errorf("failed to load Helm chart: %v", err)
+	}
+
+	if err := util.EnsureNamespace(ctx, sublogger, kubeClient, NginxIngressControllerNamespace); err != nil {
+		return fmt.Errorf("failed to create namespace: %v", err)
+	}
+
+	release, err := util.CheckHelmRelease(ctx, sublogger, helmClient, NginxIngressControllerNamespace, NginxIngressControllerReleaseName)
+	if err != nil {
+		return fmt.Errorf("failed to check to Helm release: %v", err)
```
youssefazrak, line 140: Typo.

youssefazrak, line 140: CheckHelmRelease is giving back a non-nil error value causing the installation to abort. This is the case as I'm doing a  new installation so the Helm chart doesn't exist and it leads to failure.

Edit: checking further. Seems like even with an existing chart, it still fails with the same error ` Operation failed: failed to deploy nginx-ingress-controller: failed to check to Helm release: failed to check for an existing release: failed to list releases: .`.

xrstf: You probably maybe don't have the Helm binary in the same symlinked location that I do.

youssefazrak: @xrstf can't we embed Helm in the installer in this case? For references: [link to the Helm interesting code ](https://github.com/helm/helm/tree/4bb8483a28e53b39ddb6b758f0974e9956729cb6/pkg/action)
I modified the path to my Helm3 binary and it works now but for production purposes, I don't think this might be the best way to go.

xrstf: The plan is to ship the installer as a Docker image with a Helm binary and the Helm charts included. When the installer was started, embedding Helm was a PITA and I have since not looked again, but it's the reason why we have a very simple Helm interface in the code, so we can easily change the implementation *later*.

xrstf: And the name for the current Helm 3 binary in the installer is mostly because I have roughly 8 different Helm binaries on my machine. ^^

xrstf: I also mentioned this in the proposal:

> Helm is not easy to embed, so I chose to just shell out to calling a Helm binary. As the installer is delivered in a Docker image (so it can bundle the Helm charts), this seemed a simple choice. The interaction with Helm goes through a super simple interface, so it would be possible to swap out the implementation later.

;-)

youssefazrak: Noted

docs/proposals/kubermatic-installer.md
```diff
@@ -0,0 +1,128 @@
...
+  release in the cluster is identical to the Helm chart that is shipped with the installer,
+  nothing happens unless the user runs the installer with `--force`.
+* `kubectl apply` the given `KubermaticConfiguration`.
+
+If anything goes wrong during this, the installation is aborted. The user then can fix
+the problem and re-run the installer.
+
+It would be possible to also manage Seeds while we're at it, but the win between having to
+explain to the user that the installer can also install the seeds, vs. just making them
+`kubectl apply` them is questionable.
```
youssefazrak, line 120: Fair point. The advantage I would see though is to have an all in one type of installation and not having a "mix". Would an optional flag make sense here?

youssefazrak:
@xrstf Install is complete from my side 👍 

I had an issue with the CRDs of cert-manager since the crd-install hook has been removed in favor of the crds/ directory in Helm 3 (https://v3.helm.sh/docs/topics/charts_hooks/). I had to manually create the crds/ folder and move the crd.yaml inside for the install to complete.
This is not related directly to the installer but just a heads up to modify the script creating the tarball releases taking this into account.

Else lgtm

docs/proposals/kubermatic-installer.md
xrstf: I think this is a bridge we can burn when we come to it. This does not affect the rest of the installer in any way. :-)

xrstf:
@youssefazrak can you please /lgtm this formally?

PR #5689, opened 2020-07-21
add external cluster CRD and controller to cleanup finalizers

### LINKED ISSUE #5688: Create imported cluster CRD
**User Story**

Create a new structure for the imported cluster with kubeconfig reference. Add a new controller to cleanup the secret with kubeconfig.

This is part of https://github.com/kubermatic/kubermatic/issues/5687.

zreigz:
**What this PR does / why we need it**: add new CRD to reflect the external cluster. Create the controller to cleanup kubeconfig secret during cluster deletion.

Fixes #5688

```release-note
New CRD for external k8s cluster
```


pkg/controller/master-controller-manager/imported-cluster/cluster_controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			cluster := &kubermaticv1.ImportedCluster{}
+			err = kubermaticFakeClient.Get(context.TODO(), client.ObjectKey{Name: test.clusterName}, cluster)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if kuberneteshelper.HasFinalizer(cluster, kubermaticapiv1.ImportedClusterKubeconfigCleanupFinalizer) {
+				t.Fatalf("the finilizer should be deleted")
```
xrstf: fin*a*lizer, also Fatal() instead of Fatalf()

pkg/controller/master-controller-manager/imported-cluster/cluster_controller_test.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+
+			secretKubeconfig := &corev1.Secret{}
+			err = kubermaticFakeClient.Get(context.TODO(), client.ObjectKey{Name: cluster.GetKubeconfigSecretName()}, secretKubeconfig)
+			if err == nil {
+				t.Fatal("expected error")
+			}
+			if !kerrors.IsNotFound(err) {
+				t.Fatal("expected not found error")
```
xrstf: `Fatalf("expected not-found error, but got %v", err)`

pkg/controller/master-controller-manager/imported-cluster/cluster_controller_test.go
zreigz: done

pkg/controller/master-controller-manager/imported-cluster/cluster_controller_test.go
zreigz: done

zreigz:
@xrstf ready for review PTAL

pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if icl.DeletionTimestamp != nil {
+		if kuberneteshelper.HasOnlyFinalizer(icl, kubermaticapiv1.ExternalClusterKubeconfigCleanupFinalizer) {
+			if err := r.cleanUpKubeconfigSecret(icl); err != nil {
+				log.Errorf("Could not delete kubeconfig secret, %v", err)
+				return reconcile.Result{}, err
+			}
+		}
+	}
```
xrstf, line 98: If this is the only logic, this should be achievable by just using an owner reference, no?

xrstf (commented):
Questions

pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
zreigz: This is a similar situation like with Cluster and Credentials Secret. In create method we first create a secret and then cluster, so we don't have a reference because the cluster doesn't exist yet. We can't store sensitive data like credentials or kubeconfig in cluster or externalCluster object event for a second. First, we store the sensitive data into secret then we create cluster. This is how it works right now.

xrstf: I appreciate that, but the order of creation does not mean that in the future, you couldn't just create the secret, create the cluster, then update the secret to add the owner reference.

zreigz: 1 endpoint creates a secret with a special label (atomic operation)
2 external_cluster_controller watches secrets with this label
3 reconcile method creates an external cluster with owner reference 

sounds reasonable?

xrstf: Sounds good to me.

xrstf (approved):
I changed my mind, kind of. All approaches to atomatically adding the two resources (ExternalCluster plus it's kubeconfig Secret) kind of suck, so it's fine if we do it via a controller instead of an owner ref.

PR #5695, opened 2020-07-22
add a proposal for cluster import

zreigz:
**What this PR does / why we need it**: add a proposal for cluster import

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/import-cluster.md
+++ b/docs/proposals/import-cluster.md
@@ -1,0 +1,56 @@
+# Importing a Cluster
+
+**Author**: Lukasz Zajaczkowski (@zreigz)
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+The end user would like to import existing Kubernetes cluster to Kubermatic platform. Therefore, Kubermatic does not provision
+Kubernetes, but only sets up the connection to communicate with the cluster and install controllers to enable some Kubermatic features.
+Kubermatic features, including displaying cluster details, metrics and role-based access control, will be available for imported clusters.
+The cluster details will be displayed in the Dashboard in Kubermatic way. The Kubermatic doesn't manipulate or change control plane.
+The configuration of an imported cluster still has to be edited outside of Kubermatic.
+
+## Prerequisites
+
+To import cluster user has to deliver an admin kubeconfig. The API server must be accessible for the Kubermatic.
+
+## Implementation
+
+During the cluster import the Kubermatic creates `ImportedCluster` CRD with the reference to the `Secret` with kubeconfig.
+
+```
+// ImportedCluster is the object representing a imported kubernetes cluster.
+type ImportedCluster struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ImportedClusterSpec `json:"spec"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ImportedClusterList specifies a list of imported kubernetes clusters
+type ImportedClusterList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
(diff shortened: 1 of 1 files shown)


docs/proposals/import-cluster.md
```diff
@@ -0,0 +1,56 @@
+# Adding an external cluster to Kubermatic
+
+**Author**: Lukasz Zajaczkowski (@zreigz)
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+The end user would like to add existing Kubernetes cluster to Kubermatic platform. Therefore, Kubermatic does not provision
+Kubernetes, but only sets up the connection to communicate with the cluster and install controllers to enable some Kubermatic features.
```
xrstf, line 9: What controllers are we installing into external clusters?

docs/proposals/import-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+Kubermatic features, including displaying cluster details, metrics and role-based access control, will be available for external clusters.
```
xrstf, line 10: Would we install the metrics-server ourselves, if it's not there?

docs/proposals/import-cluster.md
```diff
@@ -0,0 +1,56 @@
...
+
+	// HumanReadableName is the cluster name provided by the user
+	HumanReadableName string `json:"humanReadableName"`
+
+	KubeconfigReference *providerconfig.GlobalSecretKeySelector `json:"kubeconfigReference,omitempty"`
+}
+
+```
+
+The Kubermatic implements a provider with all necessary methods to access and control the cluster. It uses delivered kubeconfig
```
xrstf, line 52: What does "controlling" mean exactly? How do we even know that we have enough permissions to do anything in the external cluster?

zreigz: in my opinion, the user should deliver admin kubeconfig. This is how Rancher import external cluster. It should be in the prerequisites.

docs/proposals/import-cluster.md
zreigz: It would be good to have this for the metrics. It's correlated with your second question about kubeconfig privileges.

docs/proposals/import-cluster.md
zreigz: rbac, addon

docs/proposals/import-cluster.md
xrstf: I am asking because suppose a user is evil and does not provide an admin kubeconfig. Now our controllers might run into reconcile loops and if enough bad external clusters have been added, we will run into the maxParallelReconcile limit and effectively be deadlocked.

For clusters that we did not create, we must be really, really careful.

docs/proposals/import-cluster.md
xrstf: But these will be running in our seed clusters, right? I first understood this as "installing into the cluster", i.e. letting our controllers run on user cluster nodes.

docs/proposals/import-cluster.md
xrstf: Yes it would be good for metrics. But what if the metrics-server is not there? Will we install it or will we simply display a warning "No metrics available, please do X, Y Z to fix."?

docs/proposals/import-cluster.md
zreigz: I think it will be necessary to install some controller manger into the cluster, for example, user cluster RBAC controller. I didn't think about others. The first idea was to create POC and get feedback from the customers and then extend this for other functionality

docs/proposals/import-cluster.md
zreigz: I have no clear answer for that. I think we should discuss some subjects during tech lead meeting

xrstf: We decided on not installing anything in the cluster ourselves, but rather document the requirements that must be met in order to external clusters to be shown our dashboard with all details and stuff.

xrstf (approved):
@scheeles LGTM'ed this today outside of GitHub.

/approve

PR #5705, opened 2020-07-23
add KubermaticAddonTakesTooLongToReconcile alert

furkhat:

It adds an alert that is fired if there is some problem with add-on reconciliation for over 30m.

**Documentation**:
https://github.com/kubermatic/docs/pull/383

```release-note
add KubermaticAddonTakesTooLongToReconcile alert
```

### DIFF
--- a/charts/monitoring/prometheus/rules/kubermatic-seed-kubermatic.yaml
+++ b/charts/monitoring/prometheus/rules/kubermatic-seed-kubermatic.yaml
@@ -43,6 +43,17 @@ groups:
     for: 0m
     labels:
       severity: warning
+  - alert: KubermaticAddonTakesTooLongToReconcile
+    annotations:
+      message: Addon {{ $labels.addon }} in cluster {{ $labels.cluster }} has no related
+        resources created for more than 30min.
+      runbook_url: https://docs.kubermatic.com/kubermatic/master/monitoring/runbook/#alert-kubermaticaddonreconciliationtakestoolong
+    expr: kubermatic_addon_reconcile_failed * on(cluster) group_left() kubermatic_cluster_created
+      - kubermatic_addon_reconcile_failed * on(cluster) group_left() kubermatic_cluster_deleted
+      > 0
+    for: 30m
+    labels:
+      severity: warning
   - alert: KubermaticControllerManagerDown
     annotations:
       message: KubermaticControllerManager has disappeared from Prometheus target
--- a/charts/monitoring/prometheus/rules/src/kubermatic-seed/kubermatic.yaml
+++ b/charts/monitoring/prometheus/rules/src/kubermatic-seed/kubermatic.yaml
@@ -57,6 +57,21 @@ groups:
           Manually deleted resources inside of the user cluster is a common reason for failing deletions.
         - If all resources of the addon inside the user cluster have been cleaned up, remove the blocking finalizer (e.g. `cleanup-manifests`) from the addon resource.
 
+  - alert: KubermaticAddonTakesTooLongToReconcile
+    annotations:
+      message: Addon {{ $labels.addon }} in cluster {{ $labels.cluster }} has no related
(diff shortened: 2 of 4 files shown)


charts/monitoring/prometheus/rules/src/kubermatic-seed/kubermatic.yaml
```diff
@@ -57,6 +57,19 @@ groups:
           Manually deleted resources inside of the user cluster is a common reason for failing deletions.
         - If all resources of the addon inside the user cluster have been cleaned up, remove the blocking finalizer (e.g. `cleanup-manifests`) from the addon resource.
 
+  - alert: KubermaticAddonTakesTooLongToReconcile
+    annotations:
+      message: Addon {{ $labels.addon }} in cluster {{ $labels.cluster }} has no related resources
+        created for more than 30min.
+      runbook_url: https://docs.loodse.com/kubermatic/master/monitoring/runbook/#alert-kubermaticaddonreconciliationtakestoolong
```
xrstf: Can you please use `docs.kubermatic.com`?

pkg/collectors/addon.go
```diff
@@ -88,12 +88,21 @@ func (cc *AddonCollector) collectAddon(ch chan<- prometheus.Metric, addon *kuber
...
 	clusterName := addon.OwnerReferences[0].Name
 
+	resourcesCreated := "0"
+
+	for _, cond := range addon.Status.Conditions {
+		if cond.Type == kubermaticv1.AddonResourcesCreated {
+			resourcesCreated = "1"
+		}
+	}
+
```
xrstf: Not sure if adding this as a label to the creation timestamp metric is the optimal approach. What about having a dedicated metric for this?

xrstf (changes requested):
Some minor things

pkg/collectors/addon.go
furkhat: Could you elaborate more on this?

charts/monitoring/prometheus/rules/kubermatic-seed-kubermatic.yaml
```diff
@@ -28,6 +28,15 @@ groups:
     for: 0m
     labels:
       severity: warning
+  - alert: KubermaticAddonTakesTooLongToReconcile
+    annotations:
+      message: Addon {{ $labels.addon }} in cluster {{ $labels.cluster }} has no related resources
+        created for more than 30min.
+      runbook_url: https://docs.loodse.com/kubermatic/master/monitoring/runbook/#alert-kubermaticaddonreconciliationtakestoolong
```
ekeih: I guess this also should be docs.kubermatic.com

pkg/collectors/addon.go
xrstf: Instead of extending the `kubermatic_addon_created` metric with a (somewhat) unrelated label, why not create a new like `kubermatic_addon_has_resources`?

charts/monitoring/prometheus/rules/kubermatic-seed-kubermatic.yaml
xrstf: Yeah but this one is generated from the `src/` file :)

xrstf (changes requested):
Please use a compatible yq version so that indentation in the generated file matches (see the verify-prometheus-rules job). And something does not compile anymore ;-)

furkhat:
> see the verify-prometheus-rules job

@xrstf I cannot see the CI jobs for the PR. I get 403.


I fixed what I found, hope it's good now.

xrstf:
We're working on making the build logs available. :)

xrstf:
@furkhat You should now be able to view the logs at least from the three failed jobs. And all logs for all future PRs.

xrstf (approved):
/approve

PR #5708, opened 2020-07-23
allow admin to get any datacenter from the seed

### LINKED ISSUE #5607: DC not found when trying to use OTC
**User Story**
image

I'm trying to create a cluster on OTC. It fails both using a preset and using manual credentials, claiming that the datacenter does not exist. However the datacenter does definitely exist, since I was able to select it on the previous screen.

zreigz:
**What this PR does / why we need it**: allow admin to get any datacenter from the seed

Fixes #5607

```release-note
NONE
```

### DIFF
--- a/pkg/provider/conversions.go
+++ b/pkg/provider/conversions.go
@@ -50,7 +50,7 @@ iterateOverSeeds:
 		requiredEmailDomain := datacenter.Spec.RequiredEmailDomain
 		requiredEmailDomains := datacenter.Spec.RequiredEmailDomains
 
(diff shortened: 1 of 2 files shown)


pkg/provider/conversions.go
```diff
@@ -50,7 +50,7 @@ iterateOverSeeds:
 		requiredEmailDomain := datacenter.Spec.RequiredEmailDomain
```
floreks, line 50: They seem not to be used anymore

zreigz: they are used in else statement

PR #5709, opened 2020-07-23
Cleanup API e2e Tests / The Unflakaning

xrstf:
This PR does a general cleanup of our API's e2e tests, triggered by the fact that they are very flaky (basically more than half of them fail). The following changes have been made:

* a custom HTTP client takes care of silently ignoring temporary network errors and unwanted HTTP status codes (e.g. 409 Conflict)
* hardcoded waitloops have been replaced with PollImmediate()
* some tests now don't create nodes anymore, in order to speed up the tests
* a lot of wait loops have been removed, as they would only hide issues
* the default k8s version has been bumped from 1.15.x to 1.8.8
* typos have been fixed
* logging has been extended a lot, so it's now much easier to follow along what tests are doing
* the tests themselves have not changed, i.e. we test the same things as before

I left a few comments explaining the highlights.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1165,7 +1165,10 @@ presubmits:
       - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.21
         imagePullPolicy: Always
         command:
-        - "./hack/ci/ci_run_api_e2e.sh"
+        - "./hack/ci/ci-run-api-e2e.sh"
+        env:
+        - name: VERSION_TO_TEST
+          value: v1.18.8
         securityContext:
           privileged: true
         resources:
--- a/hack/ci/ci-setup-kubermatic-in-kind.sh
+++ b/hack/ci/ci-setup-kubermatic-in-kind.sh
@@ -63,13 +63,11 @@ dockerd > /tmp/docker.log 2>&1 &
 echodate "Started Docker successfully"
 
 function docker_logs {
-  originalRC=$?
-  if [[ $originalRC -ne 0 ]]; then
+  if [[ $? -ne 0 ]]; then
     echodate "Printing Docker logs"
     cat /tmp/docker.log
     echodate "Done printing Docker logs"
   fi
-  return $originalRC
 }
 appendTrap docker_logs EXIT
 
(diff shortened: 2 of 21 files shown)


hack/ci/ci-setup-kubermatic-in-kind.sh
```diff
@@ -63,13 +63,11 @@ dockerd > /tmp/docker.log 2>&1 &
...
 
 function docker_logs {
-  originalRC=$?
-  if [[ $originalRC -ne 0 ]]; then
+  if [[ $? -ne 0 ]]; then
     echodate "Printing Docker logs"
     cat /tmp/docker.log
     echodate "Done printing Docker logs"
   fi
-  return $originalRC
```
xrstf, line 72: Returning non-zero codes from traps actually prevents other traps from working. Also it's not necessary to return the original RC, the script's RC will stay the same even if all traps return 0.

This is why I removed the "get the original RC, then return it" logic in all our traps.

pkg/test/e2e/api/cluster_rbac_test.go
```diff
@@ -46,9 +44,9 @@ func TestCreateClusterRoleBinding(t *testing.T) {
 			name:                     "create cluster/role binding",
 			dc:                       "kubermatic",
 			location:                 "do-fra1",
-			version:                  "v1.18.8",
+			version:                  getKubernetesVersion(),
 			credential:               "e2e-digitalocean",
-			replicas:                 1,
+			replicas:                 0,
```
xrstf, line 49: No need for nodes in this test. Saves a good 4-5 minutes of runtime.

.prow.yaml
```diff
@@ -1165,7 +1165,7 @@ presubmits:
       - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.21
         imagePullPolicy: Always
         command:
-        - "./hack/ci/ci_run_api_e2e.sh"
+        - "./hack/ci/ci-run-api-e2e.sh"
```
xrstf, line 1168: Thankfully this is only a presubmit, so we don't need to make the filename change backwards compatible (as we would have to do with a postsubmit job).

PR #5710, opened 2020-07-23
make etcd launcher image name overridable

multi-io:

This adds an option `-etcd-launcher-image` to the seed-controller-manager which lets the user set the image name for the etcd launcher initcontainer image. This is the same thing that `-kubermatic-image` does for the kubermatic api image, i.e. it allows you to pull the image from somewhere other than Kubermatic's own quay repo, which is needed if people run their own fork with KUBERMATICCOMMIT values that don't exist upstream.

```release-note
etcd launcher image name overridable
```

### DIFF
--- a/pkg/resources/etcd/statefulset.go
+++ b/pkg/resources/etcd/statefulset.go
@@ -61,6 +61,7 @@ type etcdStatefulSetCreatorData interface {
 	GetPodTemplateLabels(string, []corev1.Volume, map[string]string) (map[string]string, error)
 	ImageRegistry(string) string
 	EtcdDiskSize() resource.Quantity
+	EtcdLauncherImage() string
 	GetClusterRef() metav1.OwnerReference
 	SupportsFailureDomainZoneAntiAffinity() bool
 }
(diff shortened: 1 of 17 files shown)


pkg/resources/etcd/statefulset.go
```diff
@@ -101,9 +102,8 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
 
 			set.Spec.Template.Spec.InitContainers = []corev1.Container{
 				{
-					Name: "etcd-launcher-init",
-
-					Image:           data.ImageRegistry(resources.RegistryQuay) + "/kubermatic/etcd-launcher:" + resources.KUBERMATICCOMMIT,
+					Name:            "etcd-launcher-init",
+					Image:           data.EtcdLauncherImage() + ":" + resources.KUBERMATICCOMMIT,
```
moelsayed, line 106: How would this work? Even if I override the image via the new option, it will still append `KUBERMATICCOMMIT` to it, right ?

moelsayed: Oh you will override both image and the kubermaticcommit, right ?

multi-io: Exactly. We'll have our own KUBERMATICCOMMIT values but they're determined the same way as upstream. I.e. we'll build and push the image for our fork, and KUBERMATICCOMMIT containing the sha1 of the commit will be baked into the binary and also used for the image tag name when pushing to our own docker repo.

multi-io: Api image override is handled the same way: https://github.com/kubermatic/kubermatic/blob/dcd7e764b61a754522fac99dc395ccd133f2cd3c/pkg/resources/usercluster/deployment.go#L162

multi-io:
I just remembered that I forgot to adapt the Helm chart. That should probably be part of this PR too, right? I'll do that on Monday.

kdomanski:
Does this resolve #5549 ?

PR #5731, opened 2020-08-05
Use Athens instead of homegrown go cache

xrstf:
Previously we uploaded the compiled Go packages to Minio in order to speed up download and compilation time. For now we replce the cache with a local proxy, so the download times are sped up a bit. The compilation time however will increase until we maybe re-create a caching mechanism for it.

```release-note
NONE
```


kron4eg (approved):

xrstf:
/override pre-kubermatic-canary-deployment-ci-kubermatic-io

The dashboard is currently broken, so the canary deployment fails with a ImagePullBackoff.

PR #5737, opened 2020-08-06
Fix prometheus alertmanager queries for KubeClientCertificateExpiration

cedi:
In the alert-manager config there is a problem where

`apiserver_client_certificate_expiration_seconds_count{job="apiserver"} > 0`
and
`histogram_quantile(0.01, sum by (job, le) (rate(apiserver_client_certificate_expiration_seconds_bucket{job="apiserver"} [5m]))) < 604800`
are joined together by `and`.

However the first portion contains a label `instance` while the second
portion of the query does a `sum(` doesn't include the `instance` label
and therefore prometheus cannot join together those two queries.

This patch adds the instance label to the histogram query in order to
make this alert work.
This can cause an production outage, since you wont be alerted
on a expiring certificate in time.

The current query is this:
Screenshot 2020-07-23 at 11 19 42

While the query submitted in this PR is this:
Screenshot 2020-07-23 at 11 16 18

**Which issue(s) this PR fixes** *
No issue opened on the Kubermatic repository

None

**Documentation**:
None

```release-note
Fixing the KubeClientCertificateExpiration Prometheus Alert which did not alert in for expiring certificates
```

### DIFF
--- a/charts/monitoring/prometheus/Chart.yaml
+++ b/charts/monitoring/prometheus/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: prometheus
-version: 2.2.25
+version: 2.2.26
 appVersion: v2.19.2
 description: Prometheus Monitoring for Kubernetes
 keywords:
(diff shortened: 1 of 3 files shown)


xrstf (approved):
Thanks, very nice.

PR #5739, opened 2020-08-07
fix codegen

multi-io:

I'm not sure if I misunderstand something, but "update-codegen" (and thus also verify-codegen) fails in the current master for me, pretty obviously, with `sed: can't read v2/pkg/crd/client/informers/externalversions/generic.go: No such file or directory`. This fixes that. I'm not sure why prow on the master branch apparently doesn't run the the `pre-kubermatic-verify` job, which would trigger this.

```release-note
NONE
```

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -47,7 +47,7 @@ echodate "Generating operator:v1alpha1"
   --go-header-file /tmp/headerfile
 
 # Temporary fixes due to: https://github.com/kubernetes/kubernetes/issues/71655
-GENERIC_FILE="v2/pkg/crd/client/informers/externalversions/generic.go"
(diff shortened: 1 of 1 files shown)


multi-io:
Hm OK, thanks for running this. So the reverted change was actually needed to make the prow pipeline work. I'll have to grok this further then.

xrstf:
We have also experienced similar inconsistent issues with some coworkers, but settled on keeping it this way. @irozzo-1A is preparing a PR to run the codegen in Docker, hopefully fixing this for everyone.

irozzo-1A:
indeed https://github.com/kubermatic/kubermatic/pull/5741

multi-io:
Closing the PR.

PR #5727, opened 2020-08-04
custom default etcd replicas number for new clusters

furkhat:
Signed-off-by: furkhat <vailodf@gmail.com>

**What this PR does / why we need it**:

Adds `-etcd-replicas-number` option to the `seed-controller-manager` that will be used as value for`componentsOverride.etcd.clusterSize` for all new clusters.

**Special notes for your reviewer**:

Also fixed some hardcodes (go templates) that assume that there are only 3 etcd replicas.

**Documentation**:

What documentation would you like me to add?

**Does this PR introduce a user-facing change?**:

```release-note
Add `-etcd-replicas` option to `seed-controller-manager` to customize number of etcd replicas for new clusters using `componentsOverride`
```

PR #5756, opened 2020-08-12
componentsOverrides configmap

furkhat:
Signed-off-by: furkhat <vailodf@gmail.com>

Moves all the componentOverrides to ConfigMap so we can configure values applied for new clusters.
Advantage of it is that we can extend componentOverrides and don't need to add cli flags for each option. 

Fixes #

Based on the #5727

**Documentation**:

https://github.com/kubermatic/docs/pull/438

```release-note
componentOverrides applied for new clusters now configurable in ConfigMap. Added optional cli flag to specify ConfigMap. Removed previously used cli flags that customize some of componentOverride values.
```


furkhat:
Hey @xrstf, could you take a look?

cmd/etcd-launcher/main.go
```diff
@@ -44,14 +44,14 @@ import (
 )
 
 const (
-	defaultClusterSize       = 3
+	defaultEtcdReplicas      = 3
```
xrstf: :rocket:

charts/kubermatic/templates/default-cluster-overrides-configmap.yaml
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: default-cluster-overrides
+data:
+  yaml: |
```
xrstf: THis is an unfortunate naming, because it makes it hard to tell `data.yaml` from a filename and a path expression. Can we make the key just `config.yaml` (i.e. the path being `spec."config.yaml"`)?

cmd/seed-controller-manager/options.go
```diff
@@ -143,12 +140,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
-	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
-	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-reconciling-disabled-by-default", false, "Whether to disable reconciling for the apiserver endpoints by default")
-	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
-	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
 	flag.IntVar(&c.concurrentClusterUpdate, "max-parallel-reconcile", 10, "The default number of resources updates per cluster")
 	flag.IntVar(&c.addonEnforceInterval, "addon-enforce-interval", 5, "Check and ensure default usercluster addons are deployed every interval in minutes. Set to 0 to disable.")
+	flag.StringVar(&c.clusterComponentsSettings, "default-cluster-overrides", "default-cluster-overrides", "Default components settings applied for every new cluster")
```
xrstf: The help text does not make it clear that it's the name of a ConfigMap.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -291,20 +291,20 @@ type AuditLoggingSettings struct {
...
 
 type APIServerSettings struct {
-	DeploymentSettings `json:",inline"`
+	ResourceSettings `json:",inline"`
 
 	EndpointReconcilingDisabled *bool `json:"endpointReconcilingDisabled,omitempty"`
 }
 
-type DeploymentSettings struct {
+type ResourceSettings struct {
```
xrstf: The replica count is logically not really expected to be nested below `resources`, so I would tend to disagree with this renaming. Also, `DeploymentSettings` allows for easier extensibility.

xrstf (changes requested):
Love it! :slightly_smiling_face: Just a few notes.

/ok-to-test
/test all

xrstf:
Also, this is a backwards incompatible change to the Helm values, no? So it needs a sister PR to our documentation with a migration step.

furkhat:
Added documentation https://github.com/kubermatic/docs/pull/438

charts/kubermatic/Chart.yaml
```diff
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubermatic
-version: 1.1.13
+version: 2.0.0
```
irozzo-1A: Is there any reason why you increased major version `2.0.0`? TBH I'm not sure to understand how semantic versioning applies to helm charts :-)

charts/kubermatic/templates/default-cluster-overrides-configmap.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
irozzo-1A: I would make the whole manifest optional, we can keep our hardcoded values in case.

cmd/seed-controller-manager/controllers.go
```diff
@@ -84,26 +83,36 @@ func createSeedConditionUpToDateController(ctrlCtx *controllerContext) error {
...
 
+func readDefaultComponentsSettings(ctrlCtx *controllerContext) (kubermaticv1.ComponentSettings, error) {
+	ret := kubermaticv1.ComponentSettings{}
+	cm := &corev1.ConfigMap{}
+	r := ctrlCtx.mgr.GetAPIReader()
+	ns := ctrlCtx.runOptions.namespace
+	n := ctrlCtx.runOptions.clusterComponentsSettings
+	if err := r.Get(ctrlCtx.ctx, types.NamespacedName{Namespace: ns, Name: n}, cm); err != nil {
+		return ret, fmt.Errorf("get components settings configmap: %w", err)
+	}
```
irozzo-1A: I'm not very fond of this way of consuming a ConfigMap, I would rather mount it as a volume to make explicit the dependency in the manifests (also avoiding potential transient ordering issues) and simply read the file.

cmd/seed-controller-manager/options.go
```diff
@@ -143,12 +140,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
-	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
-	flag.BoolVar(&c.apiServerEndpointReconcilingDisabled, "apiserver-reconciling-disabled-by-default", false, "Whether to disable reconciling for the apiserver endpoints by default")
-	flag.IntVar(&c.controllerManagerDefaultReplicas, "controller-manager-default-replicas", 1, "The default number of replicas for usercluster controller managers")
-	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
 	flag.IntVar(&c.concurrentClusterUpdate, "max-parallel-reconcile", 10, "The default number of resources updates per cluster")
 	flag.IntVar(&c.addonEnforceInterval, "addon-enforce-interval", 5, "Check and ensure default usercluster addons are deployed every interval in minutes. Set to 0 to disable.")
+	flag.StringVar(&c.clusterComponentsSettings, "default-cluster-overrides", "default-cluster-overrides", "ConfigMap name with default components settings applied for every new cluster")
```
irozzo-1A: Consistently with my previous comment I would suggest here to provide a filename instead of a ConfigMap name, and make this flag optional (it should be legal to run the seed controller manager without overriding the defaults).

hack/ci/ci-setup-kubermatic-in-kind.sh
```diff
@@ -189,7 +189,18 @@ function print_cluster_exposer_logs {
 
   return $originalRC
 }
+
```
irozzo-1A: I imagine this is something you added to debut the E2E failures and should be removed before integrating the PR.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -304,18 +304,18 @@ type APIServerSettings struct {
 	EndpointReconcilingDisabled *bool `json:"endpointReconcilingDisabled,omitempty"`
 }
 
-type DeploymentSettings struct {
+type EtcdSettings struct {
```
irozzo-1A: This is breaking backward compatibility without IMO bringing much value. I would rather avoid it.

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -162,13 +162,10 @@ type KubermaticUserClusterConfiguration struct {
...
 	Monitoring KubermaticUserClusterMonitoringConfiguration `json:"monitoring,omitempty"`
-	// DisableAPIServerEndpointReconciling can be used to toggle the `--endpoint-reconciler-type` flag for
-	// the Kubernetes API server.
-	DisableAPIServerEndpointReconciling bool `json:"disableApiserverEndpointReconciling,omitempty"`
 	// EtcdVolumeSize configures the volume size to use for each etcd pod inside user clusters.
 	EtcdVolumeSize string `json:"etcdVolumeSize,omitempty"`
-	// APIServerReplicas configures the replica count for the API-Server deployment inside user clusters.
-	APIServerReplicas *int32 `json:"apiserverReplicas,omitempty"`
+	// DefaultClusterOverridesConfigMap is the name of the configmap with default components override settings.
+	DefaultClusterOverridesConfigMap string `json:"defaultClusterOverridesConfigMap"`
```
irozzo-1A: Same as per previous comments: we could avoid breaking backward compatibility and use file instead of ConfigMap name as a contract.

irozzo-1A (commented):
Thx for the contribution, I think it is useful. I left a few comments, but they are all around the following ideas:

- I would prefer mounting the ConfigMap as a volume and having a filename as input instead of the CM name itself.
- I would avoid breaking backward compatibility wherever there is not a big benefit in doing it.

irozzo-1A:
At the moment the CM is read once and if it gets updated the changes are ignored until the seed-controller-manager is restarted. For me it's ok to start with, as far as it is well documented, but probably we should improve it later. @xrstf WDYT?

cmd/seed-controller-manager/controllers.go
irozzo-1A: After thinking a bit I think that turning `ComponentSettings` into a CRD with a controller taking care of it could be even better option. I think we could do something similar to: https://github.com/openshift/cluster-version-operator/blob/2114f132b6a1e9149d864019162fb795da8d4e3d/pkg/cvo/cvo.go#L545

We would gain:
- Better validation
- Easy way to take changes to the default values dynamically

@xrstf WDYT?

furkhat: I like the idea of passing configmap name and mounting it. I like flexibility of it and that it makes it explicit that config is not dynamic.
I don't like crd for component overrides because I think we will ever have 1 crd instance at a time and having multiple component override crd instances is no use.

charts/kubermatic/Chart.yaml
furkhat: I moved some options from values.yaml so they no longer work which is backward incompatible. hence I decided to increase major version. Though I was hesitant to do it, do you think increasing minor is enough?

hack/ci/ci-setup-kubermatic-in-kind.sh
furkhat: Sure.

furkhat:
> At the moment the CM is read once and if it gets updated the changes are ignored until the seed-controller-manager is restarted. For me it's ok to start with, as far as it is well documented, but probably we should improve it later. @xrstf WDYT?

If component overrides are supplied using volume mount + passing filename it should make it explicit that it is not dynamic. 
Thanks for taking time to review my code @irozzo-1A. I will make corresponding changes to the PR now.

cmd/seed-controller-manager/controllers.go
irozzo-1A: > I like the idea of passing configmap name and mounting it. I like flexibility of it and that it makes it explicit that config is not dynamic.

We had a discussion today during our daily and I think that having config changes taken into account dynamically and benefitting from validation, defaulting etc. would not be bad. On that respect the CRD approach would make sense. We discussed two possible approaches that could be considered:

- Having the component settings in Seed resource: https://github.com/kubermatic/kubermatic/pull/5821/files
- Having a dedicated CRD: https://github.com/kubermatic/kubermatic/pull/5822/files

IMO One important thing to take into account is that whatever approach we take, we have to make sure not to impact existing clusters if defaults are changed. I think the ideal would be to have the defaulting of Clusters `componentsOverrides` done at admission phase and not in a controller, because it should be edge-driven and not level-driven. We should basically do a snapshot of the default component settings at the moment of the Cluster creation (see https://github.com/irozzo-1A/kubermatic-1/blob/2f04e66bc4c354ca7b3f10bc82099347344472e6/pkg/controller/seed-controller-manager/clustercomponentdefaulter/clustercomponentdefaulter.go#L132-L162). 

> I don't like crd for component overrides because I think we will ever have 1 crd instance at a time and having multiple component override crd instances is no use.

I see your point and I think it's valid, but on the other hand we could think about extending this mechanism in the future and having a sort of component setting presets or classes, that can be referenced by user clusters instead of configuring each of them individually. Do you think this would be useful for you?

@xrstf @moelsayed @furkhat Thoughts?

furkhat: For a long run it seems custom CRD would be better. I will start to work on it if you don't have additional comments here @xrstf @moelsayed.

xrstf: I think extending our Seed CRD instead could save us lots of work w.r.t to validation. We already have a webhook and logic to ensure currently in-use settings are not accientally removed (like datacenters).

irozzo-1A: @xrstf I'm not sure it would save lot of work. For sure it would be cheaper, but this should not be our single and ultimate design goal.

> We already have a webhook and logic to ensure currently in-use settings are not accientally removed

I'm not sure this is relevant. What I proposed, I don't know if I have consensus here, is that when clusters get created, we take a snapshot of the referenced component settings. This means that if this get removed there should be no impact at all on the existing `Cluster`s. This would be similar to the relationship between `StorageClass` and `PersistentVolume`.

@furkhat As this is not trivial as it seamed at first glance I would propose to have a proposal first so that we can agree on where we want to go.

xrstf: But then updating the components override will not affected existing clusters, isn't that a major downside? You as the cluster administrator want to ensure a consistent set of overrides and then later you change your mind and now you have to manually edit all existing clusters. That IMHO ruins the usefulness of this feature entirely.

irozzo-1A: I cannot say wether it is a downside or not if we don't first agree on what we want to achieve. What I had in mind is actually classical `defaulting`, what you proposed is more like having two sources of config that get's merged to define the final spec. I think that some specification is definitely needed here.

furkhat: > But then updating the components override will not affected existing clusters, isn't that a major downside?

@xrstf Hm, I haven't thought about default component overrides to be part of continuous reconcile. Specifically for our needs we wanted them to apply for new clusters only during creation and allow independent change of per cluster component override by editing cluster resource `.spec.componentOverrides`.  I will think about it.

> @furkhat As this is not trivial as it seamed at first glance I would propose to have a proposal first so that we can agree on where we want to go.

@irozzo-1A good idea. I will prepare a non formal proposal and put it here.

furkhat:
## What do we want?

- Specify default components settings for new clusters.
- Override some of components settings values for a specific cluster.
- Dynamically change components settings values of all the clusters that were not overridden/customized.

## Use cases

#### Specify default components settings for new clusters.
- Create/configure `ComponentsSettings`.
- Adjust controller manager deployment by setting correct values for `default-components-settings-ns` and `default-components-settings-name`.

#### Override some of components settings values for a specific cluster.

- Create/Configure `ComponentsSettings` resource in cluster's namespace with customized values.

Deprecated:
- Edit cluster's `componentsOverrides` field

#### Dynamically change components settings values of all the clusters that were not overridden/customized.
- Edit `ComponentsSettings` that is passed to controller manager via `default-components-settings-ns`/`default-components-settings-name`

# How?

- Introduce new custom resource definition `ComponentsSettings`.
- Add deprecation warning for using `Cluster`'s `.Spec.ComponentOverrides` field.
- Provide default `ComponentsSettings` to controller manager via required cli options `default-components-settings-ns` and `default-components-settings-name`.
- During cluster reconcile determine cluster's components settings to use by taking default, overriding it's values with`ComponentsSettings` in the cluster's namespace if there is any and override the result with deprecated cluster's `.Spec.ComponentsOverrides` values.

Hope that makes sense. WDYT? @irozzo-1A @xrstf

irozzo-1A:
Hi @furkhat, if I recall well we decided to move the component settings defaulting from the controller to a mutating/validating webhook for Cluster resources, as this would simplify the logic and allow to specify empty values without having side effects on existing clusters if the defaults get eventually changed. #5866 should make this quite easy.

Another topic we discussed is whether we want to use a file or configmap to provide defaults or a CRD based approach. I tend to prefer the second approach as:
- we can get easily admission control
- changes are taken into account dynamically
- no need for providing defaults at bootstrap
- seems to be a more extensible approach. e.g. we can imagine having something like `ComponentSettingsPreset` and apply the defaults to clusters based on label selection or explicit binding. Of course, this would involve some additional complexity e.g. what if multiple presets match?

Don't know if you are still available to work on this? In such case what I propose to progress on this topic is to adjust your proposal according to the elements we discussed and move it [here](https://github.com/kubermatic/kubermatic/tree/master/docs/proposals) to reach an agreement. Otherwise, if you are not available let me know and we will work out this internally. Thanks ;-)

Best regards,
Iacopo

docs/proposals/cluster-default-components-settings.md
```diff
@@ -0,0 +1,31 @@
+# Cluster default component settings webhook
+
+**Author**: Furkhat @furkhat
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+
+*short description of the topic e.g.*
+Add mutating admission webhook that reads configured configmap with default components settings and sets `.spec.componentsOverride` for new clusters.
```
irozzo-1A: As I mentioned [here](https://github.com/kubermatic/kubermatic/pull/5756#issuecomment-730229579) I would prefer introducing a new CRD, e.g. `ComponentSettingsPreset` that includes the component settings and a label selector to match the clusters (similarly to https://github.com/kubernetes/community/blob/master/contributors/design-proposals/service-catalog/pod-preset.md). Not sure if I'm overthinking here, but this approach seems more flexible and future-proof than having a static configuration at `seed-controller-manager` level.

I imagine that KKP administrators could provide several cluster flavours, that could be selected at cluster creation time by the platform users (UI could allow choosing between the available presets in the cluster creation wizard). @zreigz @shaase-ctrl  Any thoughts on this?

irozzo-1A: I would add some links for reference:
https://github.com/kubermatic/kubermatic/blob/2a489f4e4de8fb38a5ef2248a5873af9da8843d8/pkg/crd/kubermatic/v1/cluster.go#L307
And mention that this would replace the current solution based on a controller:
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/clustercomponentdefaulter/clustercomponentdefaulter.go#L110-L141

furkhat: No I think we should implement the simplest approach for our current needs and don't over engineer. I can imagine `ComponentSettingsPreset` could be useful but we don't need it at the moment. And I think that could be added latter on.

With this proposal not much have changed from users perspective because this approach doesn't really change the way things work. Instead of cli flags and reconciling defaults for all the clusters all the time the proposal is to set defaults at creation instead.

furkhat: I added the link and mentioned the controller in goals section. Thanks!

irozzo-1A: @furkhat I agree with the idea of defaulting instead of reconciling, and that we could start with a simple approach and building on top if needed, but would be better if we can do it in an incremental way and trying to avoid another breaking change later.

One aspect I don't like very much with this approach is that it brings a start time dependency on `seed-controller-manager`. If I understood it correctly, the config-map containing the settings should be there when `seed-controller-manager` starts. Practically speaking this means that there is an impact on the operator and on the helm charts too, that should be modified to deploy the cm. You could also decide to read the cm dynamically each time there is a cluster admission, but at this point, I would prefer a CRD based approach because it's more structured, and we can use the admission mechanism for validation. To keep it simple we could also decide that to start with, only one CRD instance is admitted (similarly to what is done for global Kubermatic config).

Other folks asked for a similar feature, I will try to put them into the loop to see if this approach you proposed fits their needs too.

irozzo-1A:
@furkhat The defaulting of componentSettings is now moved to a mutating webhook, I think that now it should be easier to proceed with this. Do you still can/want to work on this or we should take ownership?

furkhat:
@irozzo-1A I can work on this.

irozzo-1A:
> @irozzo-1A I can work on this.

Thx @furkhat !

PR #5766, opened 2020-08-20
bring back the gocache

xrstf:
Athens is nice for reducing network times, but the majority of time is of course spent compiling stuff. This PR brings back the caching mechanism I replaced with Athens, thinking caching would work differently when using Go modules compared to dep. That was wrong however, `GOCACHE` is still relevant.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ all: check build test
 .PHONY: $(CMD)
 build: $(CMD)
 
-$(CMD):
+$(CMD): download-gocache
 	GOOS=$(GOOS) go build -tags "$(KUBERMATIC_EDITION)" $(GOTOOLFLAGS) -o $(BUILD_DEST)/$@ ./cmd/$@
 
 install:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -43,7 +43,7 @@ all: check build test
 .PHONY: $(CMD)
 build: $(CMD)
 
-$(CMD):
+$(CMD): download-gocache
```
kron4eg, line 46: So locally on my laptop `make build` it attempt to download cache, right? I suggest to not do this as a dependency for $(CMD).

xrstf: This PR is not about refactoring, it's about bringing back the old mechanism.

kron4eg: Is there anything that prevents the so-called "refactoring" in this case?

moelsayed:
One question: at any point do we clean the minio cache storage?
otherwise:
/lgtm

xrstf:
> One question: at any point do we clean the minio cache storage?

Manually

PR #5777, opened 2020-08-25
Migrate from Keycloak-Gatekeeper to OAuth2-Proxy

xrstf:
Keycloak-Gatekeeper has been moved to https://github.com/louketo/louketo-proxy, which was then shortly thereafter also [declared dead](https://github.com/louketo/louketo-proxy/issues/683) and users are advised to migrate to https://github.com/oauth2-proxy/oauth2-proxy -- this PR accomplishes that.

Most of the configuration options are compatible, but with different names, so a manual migration step is absolutely required.

**Documentation**:
https://github.com/kubermatic/docs/pull/432

```release-note
Replaced deprecated Keycloak-Gatekeeper with OAuth2-Proxy
```

### DIFF
--- a/charts/iap/templates/deployments.yaml
+++ b/charts/iap/templates/deployments.yaml
@@ -90,7 +90,7 @@ spec:
       nodeSelector:
 {{ toYaml $.Values.iap.nodeSelector | indent 8 }}
       affinity:
-{{ (tpl (toYaml $.Values.iap.affinity) (merge $ .)) | fromYaml | toYaml | indent 8 }}
+{{ (tpl (toYaml $.Values.iap.affinity) (merge (deepCopy $) .)) | fromYaml | toYaml | indent 8 }}
       tolerations:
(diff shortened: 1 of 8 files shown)


charts/iap/templates/deployments.yaml
```diff
@@ -83,7 +90,7 @@ spec:
       nodeSelector:
 {{ toYaml $.Values.iap.nodeSelector | indent 8 }}
       affinity:
-{{ (tpl (toYaml $.Values.iap.affinity) (merge $ .)) | fromYaml | toYaml | indent 8 }}
+{{ (tpl (toYaml $.Values.iap.affinity) (merge (deepCopy $) .)) | fromYaml | toYaml | indent 8 }}
```
xrstf, line 93: This fixes a problem whre settings from one Deployment would affect all other deployments, i.e. if Grafana sets `pass_user_headers`, suddenly Alertmanager and Prometheus have the same value.

PR #5778, opened 2020-08-25
Fallback to in-tree cloud provider for OTC

irozzo-1A:
When using OpenStack provider with Open Telekom Cloud we have to fallback to the in-tree cloud provider, otherwise services of type LoadBalancer do not work properly due to:

https://github.com/kubernetes/cloud-provider-openstack/issues/960

Fixes #

This is a temporary solution and should be removed when/if OTC will have a dedicated K8C provider.

**Documentation**:

```release-note
Fallback to in-tree cloud provider when OpenStack provider is used with Open Telekom Cloud
```


xrstf (approved):
/approve

PR #5785, opened 2020-08-26
Fix retag-images script for 'exotic' Docker registries

xrstf:
The new nginx repository is `eu.gcr.io/k8s-artifacts-prod/ingress-nginx/controller:v0.34.1`, which is our first Docker image that is not just `REGISTRY/ORG/NAME:TAG`. The [Docker Registry API v2](https://docs.docker.com/registry/spec/api/#overview) explicitly does **not** require paths to be `ORG/NAME`:

> Classically, repository names have always been two path components where each path component is less than 30 characters. The V2 registry API does not enforce this. The rules for a repository name are as follows:

This PR fixes our retag script to make fewer assumptions and just treat everything between domain and tag as the "name".

```release-note
NONE
```

### DIFF
--- a/hack/retag-images.sh
+++ b/hack/retag-images.sh
@@ -19,27 +19,32 @@ set -euo pipefail
 TARGET_REGISTRY=${TARGET_REGISTRY:-127.0.0.1:5000}
 
 function retag {
-  local IMAGE="$1"
+  local image="$1"
 
-  ORG="$(echo ${IMAGE} | cut -d / -f2)"
-  NAME="$(echo ${IMAGE} | cut -d / -f3 | cut -d : -f1)"
-  TAG="$(echo ${IMAGE} | cut -d / -f3 | cut -d : -f2)"
-  TARGET_IMAGE="${TARGET_REGISTRY}/${ORG}/${NAME}:${TAG}"
+  # trim registry
+  local local_image="$(echo ${image} | cut -d/ -f1 --complement)"
 
-  echo -n "Retagging ${IMAGE} => ${TARGET_IMAGE}"
+  # split into name and tag
+  local name="$(echo ${local_image} | cut -d: -f1)"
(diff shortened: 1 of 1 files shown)


kron4eg:
```
$ echo $image
eu.gcr.io/k8s-artifacts-prod/ingress-nginx/controller:v0.34.1

$ IFS=':' read -r -a image_with_tag_array <<< "$image"

$ echo ${image_with_tag_array[0]}
eu.gcr.io/k8s-artifacts-prod/ingress-nginx/controller

$ echo ${image_with_tag_array[1]}
v0.34.1

$ echo ${#image_with_tag_array[@]} # <----- this one should be used as a guard, if length >2, then error out
2
```

hack/retag-images.sh
```diff
@@ -19,27 +19,32 @@ set -euo pipefail
...
 function retag {
-  local IMAGE="$1"
+  local image="$1"
 
-  ORG="$(echo ${IMAGE} | cut -d / -f2)"
-  NAME="$(echo ${IMAGE} | cut -d / -f3 | cut -d : -f1)"
-  TAG="$(echo ${IMAGE} | cut -d / -f3 | cut -d : -f2)"
-  TARGET_IMAGE="${TARGET_REGISTRY}/${ORG}/${NAME}:${TAG}"
+  # trim registry
+  local local_image="$(echo ${image} | cut -d/ -f1 --complement)"
```
kron4eg, line 25: There is no such flag `--complement` on macos, if that's important

xrstf: Is MacOS for this script relevant? Also I kind of like the old variant (using cut instead of IFS) more readable than array magic. ^^

kron4eg: I guess it's not important if  this script is intended to only run in CI

xrstf: Sooo... LGTY? ^^

kron4eg (approved):

PR #5794, opened 2020-08-26
Add Kubernetes 1.19, Remove 1.15

xrstf:
This adds support for Kubernetes 1.19 to Kubermatic, while also removing Kubernetes 1.15 from the list of supported versions (the upgrade paths are, as always, left so users can migrate their 1.15.x clusters to 1.16).

The 1.19 tests were switched to using Ubuntu instead of Flatcar because Kubernetes 1.19 (at the least with our Peoseidon Kubelet image) needs Docker 19 to make pod statistics work. See https://github.com/kubernetes/kubernetes/issues/94281 for more information. This is also the reason why this PR bumps the machine-controller to 1.17.1. The missing metrics are the only broken thing in 1.19+Flatcar, everything else including the conformance tests work flawlessly.

This PR also includes a fix for the conformance-tester, which for over a year now forgot to ever put SSH keys onto the nodes, which made debugging oh so annoying. Now the SSH stuff works as intended again and, using our e2e SSH keypair you can login to all test nodes again :-)

```release-note
Add support for Kubernetes 1.19, make 1.18 the new default version, remove 1.15
Update machine-controller to 1.17.1
```


pkg/resources/scheduler/deployment.go
```diff
@@ -66,8 +66,6 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
 				"--client-ca-file", "/etc/kubernetes/pki/ca/ca.crt",
 				// We're going to use the https endpoints for scraping the metrics starting from 1.13. Thus we can deactivate the http endpoint
 				"--port", "0",
-				// Force the authentication lookup to succeed, otherwise if it fails all requests will be treated as anonymous and thus fail
```
xrstf, line 69: `false` has been the default for a long time, so there's no need to set this. I removed it because it failed anyway because boolean flags do not like the `-flagname true` format anymore.

xmudrii (approved):
/lgtm
/approve
/hold
if anyone else wants to take a look, otherwise remove the hold

PR #5768, opened 2020-08-24
OPA integration proposal

lsviben:
**What this PR does / why we need it**:
Adds a proposal for OPA integration with Kubermatic.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5760 

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

Issue #5713, opened 2020-07-28
Cluster Stuck in Deletion since weekly-2020-30


toschneck:
**User Story**
A change between the version `weekly-2020-29` and the current version. Cluster stuck in deletion.

Environment:
Dev (current master) + run-2 (weekly-2020-30)

It seams that the `controller-manager` and `scheduler` is crash looping, what could be the root cause:
```
NAME                                          READY   STATUS             RESTARTS   AGE
apiserver-68b8f9cc85-dqjbn                    3/3     Running            0          5h49m
apiserver-68b8f9cc85-fhn6h                    3/3     Running            0          11h
controller-manager-68bb775bfb-57kjf           1/2     CrashLoopBackOff   7          19m
dns-resolver-8494786c54-f52mm                 2/2     Running            0          19m
dns-resolver-8494786c54-h5nq6                 2/2     Running            0          5h49m
etcd-0                                        1/1     Running            0          20h
etcd-1                                        1/1     Running            0          20h
etcd-2                                        1/1     Running            0          11h
kube-state-metrics-cf5dbd78-fkqsw             1/1     Running            0          19m
kubernetes-dashboard-74ffdfd584-cnf29         1/1     Running            0          19m
kubernetes-dashboard-74ffdfd584-k2b6m         1/1     Running            0          9h
machine-controller-b6c7c9958-sntz4            1/1     Running            0          19m
machine-controller-webhook-7c7c5cf8b6-6hz89   1/1     Running            0          19m
metrics-server-8b9b459b5-kvftk                3/3     Running            0          5h49m
metrics-server-8b9b459b5-nd54c                3/3     Running            0          20h
openvpn-server-69bb74887b-w8pq4               3/3     Running            0          19m
prometheus-0                                  1/1     Running            0          19m
scheduler-747dc9f5d-cr4m8                     1/2     CrashLoopBackOff   7          19m
usercluster-controller-67b9fbd5fd-68w9q       1/1     Running            0          5h49m
```
Cluster Object:
```
address:
  adminToken: v6d6j6.s8658z5kjglp9dl8
  externalName: xdgxb7vnk5.europe-west3-c.dev.kubermatic.io
  internalURL: apiserver-external.cluster-xdgxb7vnk5.svc.cluster.local.
  ip: 35.198.93.90
  port: 31279
  url: https://xdgxb7vnk5.europe-west3-c.dev.kubermatic.io:31279
apiVersion: kubermatic.k8s.io/v1
[...]

PR #5814, opened 2020-09-03
Deploy gatekeeper

### LINKED ISSUE #5784: Deploy gatekeeper for user cluster by default
**User Story**
Deploy gatekeeper by default as part of the OPA integration with Kubermatic.

lsviben:
This PR adds gatekeeper deployment as part of the Kubermatic OPA integration

Fixes #5784 

**Documentation:**
https://github.com/kubermatic/kubermatic/pull/5768/files

Tested on dev cluster. Testing procedure is like this:
1. Run local `seed-controller` and `api` with the same `worker-name` arg set.
2. Run local `dashboard`.
3. Create a new cluster.
4. When the cluster is ready (you can already check if the seed resources are deployed correctly), edit is `user-cluster-controller` deployment and set `replicas` to 0.
5. Run your local `user-cluster-controller` with the arguments set for the the proper user cluster - hint , you can check what args the original `user-cluster-controller` has.
6. Check to see if the gatekeeper resources are deployed on user cluster.
7. You can use some of the constraint templates from the gatekeeper demo to check if its working correctly - https://github.com/open-policy-agent/gatekeeper/tree/master/demo/basic. Keep in mind to create ConstraintTemplates and Constraints and Configs in the user cluster.

An addtional step would be to remove the addon gatekeeper installation option. Let me know if you want it in this PR or it can be a separate task.

```release-note
Gatekeeper is now deployed automatically for the user clusters as part of Kubermatic OPA integration
```


zreigz (commented):
I think you should remove the gatekeeper from /addons directory. Rest looks good

cmd/user-cluster-controller-manager/main.go
```diff
@@ -110,6 +111,7 @@ func main() {
 	flag.StringVar(&runOp.ownerEmail, "owner-email", "", "An email address of the user who created the cluster. Used as default subject for the admin cluster role binding")
 	flag.StringVar(&runOp.updateWindowStart, "update-window-start", "", "The start time of the update window, e.g. 02:00")
 	flag.StringVar(&runOp.updateWindowLength, "update-window-length", "", "The length of the update window, e.g. 1h")
+	flag.BoolVar(&runOp.opaIntegration, "opa-integration", false, "Weather the managed cluster should integrate with OPA")
```
moadqassem: I think you mean `whether` not `weather`. Can you please change the message to: 
`Enable OPA integration in user cluster`.

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/crds.go
```diff
@@ -0,0 +1,75 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gatekeeper
+
+import (
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
moadqassem: Please regroup the imports: 
* go packages 
* 3rd Party packages
* k8c packages 
* k8s packages

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/crds.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// ConfigCRDCreator returns the gatekeeper config CRD definition
+func ConfigCRDCreator() reconciling.NamedCustomResourceDefinitionCreatorGetter {
+	return func() (string, reconciling.CustomResourceDefinitionCreator) {
+		return resources.GatekeeperConfigCRDName, func(crd *apiextensionsv1beta1.CustomResourceDefinition) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
+			crd.Annotations = map[string]string{"controller-gen.kubebuilder.io/version": "v0.2.4"}
+			crd.Labels = map[string]string{"gatekeeper.sh/system": "yes"}
+			crd.Spec.Group = configAPIGroup
+			crd.Spec.Version = configAPIVersion
```
moadqassem: I am not sure that you actually need to add this field since you already add it to the versions field. Plus this field is actually deprecated thus we can simply remove it.

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/crds.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ConstraintTemplateCRDCreator returns the gatekeeper constraintTemplate CRD definition
+func ConstraintTemplateCRDCreator() reconciling.NamedCustomResourceDefinitionCreatorGetter {
+	return func() (string, reconciling.CustomResourceDefinitionCreator) {
+		return resources.GatekeeperConstraintTemplateCRDName, func(crd *apiextensionsv1beta1.CustomResourceDefinition) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
+			crd.Annotations = map[string]string{"controller-gen.kubebuilder.io/version": "v0.2.4"}
+			crd.Labels = map[string]string{"gatekeeper.sh/system": "yes", "controller-tools.k8s.io": "1.0"}
+			crd.Spec.Group = constraintTemplateAPIGroup
+			crd.Spec.Version = constraintTemplateAPIVersion
```
moadqassem: Same thing goes here. Please see the comment above.

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/webhook.go
```diff
@@ -0,0 +1,128 @@
...
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/certificates/triple"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// ValidatingwebhookConfigurationCreator returns the ValidatingwebhookConfiguration for gatekeeper
```
moadqassem: The comment here should be `ValidatingWebhookConfigurationCreator` not `ValidatingwebhookConfigurationCreator`

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/webhook.go
```diff
@@ -0,0 +1,128 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gatekeeper
+
+import (
```
moadqassem, line 19: Please regroup the imports: 
* go packages 
* 3rd Party packages
* k8c packages 
* k8s packages

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/webhook.go
```diff
@@ -0,0 +1,128 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/certificates/triple"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// ValidatingwebhookConfigurationCreator returns the ValidatingwebhookConfiguration for gatekeeper
+func ValidatingWebhookConfigurationCreator(caCert *x509.Certificate, namespace string) reconciling.NamedValidatingWebhookConfigurationCreatorGetter {
+	return func() (string, reconciling.ValidatingWebhookConfigurationCreator) {
+		return resources.GatekeeperValidatingWebhookConfigurationName, func(validatingWebhookConfigurationWebhookConfiguration *admissionregistrationv1beta1.ValidatingWebhookConfiguration) (*admissionregistrationv1beta1.ValidatingWebhookConfiguration, error) {
+			failurePolicyIgnore := admissionregistrationv1beta1.Ignore
```
moadqassem, line 36: Why this should be ignored instead of fail?

pkg/resources/gatekeeper/service.go
```diff
@@ -0,0 +1,47 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// ServiceCreator returns the function to reconcile the gatekeeper service
+func ServiceCreator() reconciling.NamedServiceCreatorGetter {
+	return func() (string, reconciling.ServiceCreator) {
+		return resources.GatekeeperWebhookServiceName, func(se *corev1.Service) (*corev1.Service, error) {
```
moadqassem, line 30: Don't you probably need to add a service type here? Would suggest: 
```go
se.Spec.Type = corev1.ServiceTypeClusterIP
```

pkg/resources/resources.go
```diff
@@ -122,6 +126,8 @@ const (
 	OpenVPNServerServiceName = "openvpn-server"
 	//MachineControllerWebhookServiceName is the name of the machine-controller webhook service
 	MachineControllerWebhookServiceName = "machine-controller-webhook"
+	//GatekeeperServiceName is the name of the gatekeeper webhook service
```
moadqassem: Please change the comment to `GatekeeperWebhookServiceName`

pkg/resources/resources.go
```diff
@@ -352,11 +360,19 @@ const (
 	MachineDeploymentCRDName = "machinedeployments.cluster.k8s.io"
 	// ClusterCRDName defines the CRD name for cluster objects
 	ClusterCRDName = "clusters.cluster.k8s.io"
+	// GatekeeperConfigCRDName defines the CRD name for gatekeeper config objects
+	GatekeeperConfigCRDName = "configs.config.gatekeeper.sh"
+	// GatekeeperConfigCRDName defines the CRD name for gatekeeper constraint template objects
```
moadqassem: This should be `GatekeeperConstraintTemplateCRDName` instead of `GatekeeperConfigCRDName`

moadqassem:
@lsviben Please regroup all the imports in this PR as they are not correct, the right order is: 
* go packages
* 3rd Party packages
* k8c packages
* k8s packages

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/webhook.go
lsviben: I changed it because of this issue: https://github.com/kubermatic/kubermatic/issues/5713#issuecomment-666573791 . 
When set to fail, when the validating webhook is malfuncioning it breaks our clusters.

cmd/user-cluster-controller-manager/main.go
lsviben: much better, thanks

pkg/resources/gatekeeper/service.go
lsviben: it defaults to it, but better to make it explicit.

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/webhook.go
moadqassem: I see, what I would suggest is the following. If the webhook is not working as expected, we should offer a mechanism to simply deactivate the webhook and remove it from the user cluster. Because using this policy at this point is kinda confusing. If I enable gatekeeper, then I am expecting some actions and if those actions are not satisfying my scenario(e.g: break my cluster for example) I should be able to simply deactivate it. Another approach can be, to make the policy flexible.

lsviben: Right, added a ticket for this - https://github.com/kubermatic/kubermatic/issues/6008. 

There should be a way to do it differently and I try to find a way before 2.16 rolls out.

moadqassem:
Thanks for your work, other than finding a mechanism to remove the OPA integration when deactivated, it looks good. 
/lgtm
/approve

PR #5795, opened 2020-08-27
Refactor clients for etcd-launcher

moelsayed:
**What this PR does / why we need it**:
Restructure etcd client code and remove some duplicated code.
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5825, opened 2020-09-07
Add dynamic etcd cluster size for tls certificate

phiphi282:
Currently when specifying a custom number of etcd replicas, the certificate won't be updated accordingly.

TLS certificate now uses dynamic etcd cluster size instead of hardcoded
number to make sure that all etcds are reachable with the certificate

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/resources/etcd/statefulset.go
+++ b/pkg/resources/etcd/statefulset.go
@@ -348,11 +348,7 @@ func ImageTag(c *kubermaticv1.Cluster) string {
 }
 
 func computeReplicas(data etcdStatefulSetCreatorData, set *appsv1.StatefulSet) int {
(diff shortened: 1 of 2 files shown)


moelsayed:
@phiphi282 Thank you for the PR. 
Unfortunately, setting this dynamically will break etcd scale up/down. When scaling down, the seed controller will immediately update the certificate to the new etcd cluster size. In the mean time, the replica set is scaled down 1 pod at a time, which create a case where a pod (etcd node) is not included in the certificate. 

I am currently working on another PR that has a workaround for this issue: https://github.com/kubermatic/kubermatic/pull/5795/files#diff-7e1fd4089925d42838c19c86d921db2cR55-R56

phiphi282:
Okay, then I will close this pr :)

PR #5829, opened 2020-09-08
allow custom envvar definitions for dex to be passed via the oauth chart

### LINKED ISSUE #5826: Support for HTTP proxy environment variables in Oauth chart
**User Story**
When the master cluster has internet access through an egress HTTP proxy environment variables (i.e. `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`) to Dex through the OAuth chart:

https://github.com/kubermatic/kubermatic/blob/99236a3d679400eedc4db7d220a4366860cc47b5/charts/oauth/templates/deployment.yaml#L38-L52

**Acceptance criteria**
OAuth chart is extended to support HTTP proxy environment variables.

erdii:
**What this PR does / why we need it**: allow custom envvar definitions for dex to be passed via the oauth chart 

Fixes #5826 

**Documentation**:
 not yet- only if this approach is deemed good by @irozzo-1A

```release-note
Allow custom envvar definitions for dex to be passed via the oauth chart. values key: `dex.env`
```

### DIFF
--- a/charts/oauth/templates/deployment.yaml
+++ b/charts/oauth/templates/deployment.yaml
@@ -39,6 +39,9 @@ spec:
       - image: {{ .Values.dex.image.repository }}:{{ .Values.dex.image.tag }}
         name: dex
         command: ["/usr/local/bin/dex", "serve", "/etc/dex/cfg/config.yaml"]
+        {{- if .Values.dex.env }}
+        env: {{ .Values.dex.env | toYaml | trim | nindent 8}}
+        {{- end }}
         ports:
         - name: https
           containerPort: 5556
--- a/charts/oauth/templates/deployment.yaml
+++ b/charts/oauth/templates/deployment.yaml
@@ -40,7 +40,7 @@ spec:
         name: dex
         command: ["/usr/local/bin/dex", "serve", "/etc/dex/cfg/config.yaml"]
         {{- if .Values.dex.env }}
-        env: {{ .Values.dex.env | toYaml | trim | nindent 8}}
+        env: {{ toYaml .Values.dex.env | trim | nindent 8 }}
         {{- end }}
         ports:
         - name: https
--- a/charts/oauth/values.yaml
+++ b/charts/oauth/values.yaml
@@ -20,6 +20,13 @@ dex:
   # this options allows setting custom envvars in the dex container
   # this list is directly handed to the container spec
   env: []
+  #  - name: HTTP_PROXY
+  #    value: "http://USER:PASSWORD@IPADDR:PORT"
+  #  - name: HTTPS_PROXY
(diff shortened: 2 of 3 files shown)


charts/oauth/values.yaml
```diff
@@ -17,6 +17,9 @@ dex:
     repository: "quay.io/dexidp/dex"
     tag: "v2.24.0"
   replicas: 2
+  # this options allows setting custom envvars in the dex container
+  # this list is directly handed to the container spec
+  env: []
```
irozzo-1A, line 22: I would just add a commented example below:

```
  env: []
  #  - name: HTTP_PROXY
  #     value: "http://USER:PASSWORD@IPADDR:PORT"
  #  - name: HTTPS_PROXY
  #     valueFrom:
  #       secretKeyRef:
  #         key: HTTPS_PROXY
  #         name: http-proxy-secret
```

charts/oauth/templates/deployment.yaml
```diff
@@ -39,6 +39,9 @@ spec:
       - image: {{ .Values.dex.image.repository }}:{{ .Values.dex.image.tag }}
         name: dex
         command: ["/usr/local/bin/dex", "serve", "/etc/dex/cfg/config.yaml"]
+        {{- if .Values.dex.env }}
+        env: {{ .Values.dex.env | toYaml | trim | nindent 8}}
```
irozzo-1A: I think you can avoid one pipe:
```
{{ toYaml .Values.dex.env | trim | nindent 8 }}
```

irozzo-1A:
Overall LGTM, just left 2 small comments

charts/oauth/templates/deployment.yaml
erdii: Personally, I prefer to read pipelines from left to right, which is kind of circumvented by having the first function on the left side of the input value.
Nevertheless: that thing is small enough and easy to understand in both variants. I'll add your recommendation and fix that missing space before the closing braces as well :smiley: 

Edit: replace `brackets` with `braces`

charts/oauth/values.yaml
erdii: Good catch! Done!

charts/oauth/templates/deployment.yaml
erdii: Also done!

irozzo-1A:
LGTM, just change the chart version to make the CI happy ;-)
https://github.com/kubermatic/kubermatic/blob/a7bd61404eb87b605997b4e28ae891439b5d74e0/charts/oauth/Chart.yaml#L17

erdii:
haha ofc I forgot that one! :see_no_evil:

PR #5831, opened 2020-09-08
define optional presubmit to test the github release code

xrstf:
This makes it easier to see if the GitHub release logic actually works (without actually talking to GitHub).

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -289,7 +289,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-      - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.18
+      - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.21
         command:
         - ./hack/ci/ci-test-github-release.sh
         resources:
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -281,7 +281,7 @@ presubmits:
       imagePullSecrets:
       - name: quay
 
-  - name: pre-kubermatic-github-release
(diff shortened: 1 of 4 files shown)


.prow.yaml
```diff
@@ -281,6 +281,22 @@ presubmits:
       imagePullSecrets:
       - name: quay
 
+  - name: pre-kubermatic-github-release
```
xmudrii: Can we name this job to make it clear it's dry-run? Maybe something along `pre-kubermatic-mock-github-release`.

xrstf: We can. I was thinking the `pre-` is enough, but apparently not. ^^

xrstf: It's `pre-kubermatic-simulate-github-release` now.

hack/ci/ci-test-github-release.sh
```diff
@@ -0,0 +1,49 @@
...
+
+set -euo pipefail
+
+cd $(dirname $0)/../..
+source hack/lib.sh
+
+export DRY_RUN=true
+export GITHUB_TOKEN="none"
+export GIT_TAG="v2.99.99"
+export RELEASE_PLATFORMS="linux-amd64 windows-amd64"
```
xmudrii, line 28: Why we don't test macOS releases here?

xrstf: Because the packaging process is identical to Linux (tar), so I didn't think it was worth it to spend the CPU cycles to compile it.

xmudrii (approved):
/lgtm

PR #5832, opened 2020-09-08
Bump k8s Go dependencies to v0.19.0

xrstf:
It's time to update to a stable release. The rc.3 was from when we did the big Go Modules refactoring andwanted to use the latest because it made the dependency tree easier.

```release-note
NONE
```


hack/lib.sh
```diff
@@ -266,14 +266,14 @@ check_all_deployments_ready() {
   local deployments
   deployments=$(kubectl -n $namespace get deployments -o json)
 
-  if [ $(jq '.items | length' <<< $deployments) -eq 0 ]; then
```
kron4eg: IIRC, this is the way shellcheck is insisting it needs to be. Do we want to drop it?

xrstf: Fixed

kron4eg (approved):

PR #5834, opened 2020-09-09
Use Go 1.15 for CI jobs

xrstf:
Updating just because we can. This also then needs updates to the postsubmits in https://github.com/kubermatic/infra/pull/1127.

```release-note
Kubermatic is built using Go 1.15.
```

### DIFF
--- a/pkg/test/e2e/api/cluster_test.go
+++ b/pkg/test/e2e/api/cluster_test.go
@@ -41,7 +41,7 @@ func getKubernetesVersion() string {
 	if len(version) > 0 {
 		return version
 	}
-	return "v1.14.2"
+	return "v1.18.8"
 }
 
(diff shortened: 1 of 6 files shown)


pkg/test/e2e/api/cluster_test.go
```diff
@@ -41,7 +41,7 @@ func getKubernetesVersion() string {
 	if len(version) > 0 {
 		return version
 	}
-	return "v1.14.2"
+	return "v1.18.8"
```
xrstf, line 44: I accidentally found this ancient and now even unsupported default when grepping through the repo. For all intents and purposes we're already defaulting to 1.18.8 for the e2e tests.

kron4eg (approved):

PR #5844, opened 2020-09-10
Make ImagePullSecret optional

### LINKED ISSUE #5694: KubermaticOperator CE requires an image pull secret
**User Story**
KubermaticOperator requires an image pull secret to be specified in the `KubermaticConfiguration` manifest even when CommunityEdition is used (it is allowed to be empty though), where all platform images are public.

This is the error thrown by the operator when no `imagePullSecret` is specified in `KubermaticConfiguration` spec:
```
{"level":"error","time":"2020-09-15T19:00:07.503Z","logger":"controller","caller":"controller/controller.go:209","msg":"Reconciler error","controller":"kubermatic-master-operator","name":"kubermatic","namespace":"kubermatic","error":"failed to reconcile Secrets: failed to ensure Secret kubermatic/dockercfg: failed to create *v1.Secret 'kubermatic/dockercfg': Secret \"dockercfg\" is invalid: data[.dockerconfigjson]: Invalid value: \"<secret contents redacted>\": unexpected end of JSON input"}
```

**Acceptance criteria**
KubermaticOperator do not fail if no image pull secret is provided (CE only).

xrstf:
This PR makes configuring the imagepullsecret always optional.

* For CE-based setups, there is no special secret required because all Docker images are public.
* For EE-based setups, a Secret is required *unless* the admin chose to mirror the images in a local Registry mirror, so we cannot be 100% sure that for EE a Secret is required.
* To make misconfiguration a tiny bit harder, the installer now checks the Docker repository used and if it's the EE repo, it will require a Docker Pull Secret (it can be empty, the installer will not validate it _that_ deeply).

Fixes #5694

```release-note
NONE
```

### DIFF
--- a/charts/kubermatic-operator/templates/dockercfg.yaml
+++ b/charts/kubermatic-operator/templates/dockercfg.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{- if .Values.kubermaticOperator.imagePullSecret }}
+{{ with .Values.kubermaticOperator.imagePullSecret }}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -21,5 +21,5 @@ metadata:
     app.kubernetes.io/name: kubermatic-operator
 type: kubernetes.io/dockerconfigjson
 data:
-  .dockerconfigjson: {{ .Values.kubermaticOperator.imagePullSecret | b64enc | quote }}
-{{- end }}
+  .dockerconfigjson: {{ . | b64enc | quote }}
+{{ end }}
--- a/pkg/controller/operator/common/defaults.go
+++ b/pkg/controller/operator/common/defaults.go
@@ -322,6 +322,14 @@ func DefaultConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logg
 
 	copy := config.DeepCopy()
 
+	if copy.Spec.ImagePullSecret == "" {
+		// This ensures that the created dockercfg Secrets later will contain valid JSON
+		// and that we do not have to make a distinction between unset, empty and full
+		// Docker Pull Secret; instead we *always* create a dockercfg Secret and reference
+		// it everywhere.
+		copy.Spec.ImagePullSecret = "{}"
+	}
+
 	if copy.Spec.ExposeStrategy == "" {
 		copy.Spec.ExposeStrategy = DefaultExposeStrategy
 		logger.Debugw("Defaulting field", "field", "exposeStrategy", "value", copy.Spec.ExposeStrategy)
--- a/pkg/install/stack/kubermatic/validation.go
+++ b/pkg/install/stack/kubermatic/validation.go
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"go.uber.org/zap"
@@ -29,6 +30,7 @@ import (
 	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
 	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
 	"k8c.io/kubermatic/v2/pkg/features"
+	"k8c.io/kubermatic/v2/pkg/resources"
 	"k8c.io/kubermatic/v2/pkg/serviceaccount"
 	"k8c.io/kubermatic/v2/pkg/util/yamled"
 )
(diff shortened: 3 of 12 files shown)


pkg/controller/operator/common/defaults.go
```diff
@@ -322,6 +322,14 @@ func DefaultConfiguration(config *operatorv1alpha1.KubermaticConfiguration, logg
...
 	copy := config.DeepCopy()
 
+	if copy.Spec.ImagePullSecret == "" {
+		// This ensures that the created dockercfg Secrets later will contain valid JSON
+		// and that we do not have to make a distinction between unset, empty and full
+		// Docker Pull Secret; instead we *always* create a dockercfg Secret and reference
+		// it everywhere.
+		copy.Spec.ImagePullSecret = "{}"
+	}
+
```
youssefazrak, line 332: Do we need this when we have [this](https://github.com/kubermatic/kubermatic/pull/5844/files#diff-bdbc7758829c9b98e5794ed1f89bffd5R23)?

xrstf: Yes, because those are different secrets. The one in the Kubermatic Operator chart is for pulling the _operator_ only, and then you have another ImagePullSecret in the KubermaticConfiguration, which is deployed by the operator to pull Kubermatic itself.

It's ugly, but an unfortunate side effect of our Helm chart structure at the moment.

charts/kubermatic-operator/templates/dockercfg.yaml
```diff
@@ -21,5 +20,4 @@ metadata:
     app.kubernetes.io/name: kubermatic-operator
 type: kubernetes.io/dockerconfigjson
 data:
-  .dockerconfigjson: {{ .Values.kubermaticOperator.imagePullSecret | b64enc | quote }}
-{{- end }}
+  .dockerconfigjson: {{ .Values.kubermaticOperator.imagePullSecret | default "{}" | b64enc | quote }}
```
irozzo-1A: @xrstf  Wouldn't be cleaner to have the operator tolerating no `imagePullSecret` rather than creating an empty one as default?

xrstf: @irozzo-1A döne

irozzo-1A:
@xrstf From what I observed this does not seem to fix #5694 I proposed a different approach here: #5874 PTAL.

xrstf:
@irozzo-1A from what i tested, it does work.

xrstf:
@irozzo-1A From what I've seen in your PR, I like the idea of using an object modifier for setting the pull secret. I wish we could use this "functional" approach more often.

What I'm no yet seeing is how this helps the user not misconfigure Kubermatic. I guess your PR still WIP and you'd handle that as well? The whole idea behind the installer is to make it as easy and smooth as possible to setup KKP. This should include warnings/errors when a user forgets/misconfigures the docker pull secret, for example.

pkg/install/stack/kubermatic/validation.go
```diff
@@ -101,11 +103,19 @@ func validateHelmValues(config *operatorv1alpha1.KubermaticConfiguration, helmVa
...
 
 	path = yamled.Path{"kubermaticOperator", "imagePullSecret"}
-	if value, _ := helmValues.GetString(path); value == "" {
+	if value, _ := helmValues.GetString(path); value == "" && config.Spec.ImagePullSecret != "" {
 		logger.Warnf("Helm values: %s is empty, setting to spec.imagePullSecret from KubermaticConfiguration", path.String())
 		helmValues.Set(path, config.Spec.ImagePullSecret)
 	}
 
+	repoPath := yamled.Path{"kubermaticOperator", "image", "repository"}
+	repo, _ := helmValues.GetString(repoPath)
```
moelsayed, line 112: Is it safe to ignore errors?

pkg/resources/resources.go
```diff
@@ -65,6 +65,16 @@ const (
...
 
 const (
+	// DefaultKubermaticCEImage defines the default Docker repository containing the Kubermatic CE API image.
+	DefaultKubermaticCEImage = "quay.io/kubermatic/kubermatic"
+	// DefaultKubermaticCEImage defines the default Docker repository containing the Kubermatic EE API image.
+	DefaultKubermaticEEImage = "quay.io/kubermatic/kubermatic-ee"
+
+	// DefaultDashboardCEImage defines the default Docker repository containing the dashboard CE API image.
+	DefaultDashboardCEImage = "quay.io/kubermatic/dashboard"
+	// DefaultDashboardEEImage defines the default Docker repository containing the dashboard EE API image.
```
moelsayed: nit: not the API image.

moelsayed (commented):
Couple of nit comments.

pkg/install/stack/kubermatic/validation.go
xrstf: This is not ignoring an error, the second return value is a boolean telling you if the value was found in the document (like a map access). This is not relevant here, as I'm selecting a string and the empty value will just be an empty string. The 2nd return value is useful if you're getting booleans out of the document, for example.

pkg/resources/resources.go
xrstf: Will fix soon

xrstf: Done

moelsayed:
/hold
Not sure if we are making modifications to accommodate @irozzo-1A's PR ?

pkg/controller/operator/common/defaults.go
irozzo-1A: Please revert this as we agreed, empty `imagePullSecret` will be covered in #5874

PR #5846, opened 2020-09-11
Take into account DOCKER_REPO environment variable for image release

irozzo-1A:
Fix the `hack/release-docker-images.sh` script to take into account the `DOCKER_REPO` environment variable for all images.
The problem is that some `Makefile` assign a value to `DOCKER_REPO` variable using `=` operator thus it is not overridden by the environment variable set by the script.

References:
https://www.gnu.org/software/make/manual/html_node/Environment.html#Environment
https://www.gnu.org/software/make/manual/html_node/Setting.html

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,30 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-export CGO_ENABLED?=0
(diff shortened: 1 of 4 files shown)


xrstf (approved):
/approve

PR #5851, opened 2020-09-11
do not deploy nodeport-proxy chart on dev anymore

xrstf:
It's time we finally make the switch to the operator-managed nodeport-proxy. This PR makes it so that dev/dev-asia do not have the old chart deployed to them anymore, so we can begin to cleanup these clusters.

```release-note
NONE
```


xmudrii (approved):
/lgtm

PR #5856, opened 2020-09-13
Validate serviceAccountKey in the installer

xrstf:
The Kubermatic API crashloops whenever the key is shorter than 32 bytes.

```release-note
NONE
```


youssefazrak (approved):
/lgtm

PR #5864, opened 2020-09-14
Make sure that the entries in nginx cm data are quoted

irozzo-1A:
Make sure that values in the nginx chart config-map template gets quoted to avoid issues with numeric and boolean values like:

```
Error: release nginx-ingress-controller failed: ConfigMap in version "v1" cannot be handled as a ConfigMap: v1.ConfigMap.Data: ReadString: expects " or n, but found f, error found in #10 byte of ...|edirect":false},"kin|..., bigger context ...|ata":{"load-balance":"least_conn","ssl-redirect":false},"kind":"ConfigMap","metadata":{"labels":{"ap|...
``` 

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/nginx-ingress-controller/Chart.yaml
+++ b/charts/nginx-ingress-controller/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: nginx-ingress-controller
-version: 1.1.6
+version: 1.1.7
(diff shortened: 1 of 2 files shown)


xrstf (approved):
/approve

I like it!

PR #5866, opened 2020-09-15
Refactor seed admission webhook server to make it generic

irozzo-1A:
Refactors the seed validation webhook to use `controller-runtime`. This should make easier to add validating or mutating webhooks for other resources.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/validation/seed/validator.go
```diff
@@ -33,71 +33,74 @@ import (
...
-	seeds, err := sv.seedsGetter()
+	seeds := kubermaticv1.SeedList{}
+	err := v.client.List(ctx, &seeds)
 	if err != nil {
 		return fmt.Errorf("failed to get seeds: %v", err)
 	}
+	seedsMap := map[string]*kubermaticv1.Seed{}
+	for _, s := range seeds.Items {
+		seedsMap[s.Name] = &s
+	}
```
xrstf: Are we sure that this works? Isn't `s` re-used in each iteration and so every element in `seedsMap` would en up with the same address?

pkg/validation/flags.go
```diff
@@ -0,0 +1,88 @@
...
+			return fmt.Errorf("failed to parse seed admission webhook listen address: %v", err)
+		}
+		s.Host = host
+		s.Port, _ = strconv.Atoi(port)
+	}
+	// controller-runtime server do not support cert file and key file being in
+	// different directories
+	// this is not fully backward compatible
+	if opts.CertFile != "" && opts.KeyFile != "" && filepath.Dir(opts.CertFile) != filepath.Dir(opts.KeyFile) {
+		return fmt.Errorf("certificate file %q and key file %q provided respectively with flags seed-admissionwebhook-cert-file and seed-admissionwebhook-cert-file should be located in the same directory", opts.CertFile, opts.certDir)
```
xrstf: should be or must be? The error sounds like a mere suggestion ^^

pkg/validation/flags.go
```diff
@@ -0,0 +1,88 @@
...
+	s.Port = opts.listenPort
+
+	// handle deprecated flags
+	if opts.ListenAddress != "" {
+		host, port, err := net.SplitHostPort(opts.ListenAddress)
+		if err != nil {
+			return fmt.Errorf("failed to parse seed admission webhook listen address: %v", err)
+		}
+		s.Host = host
+		s.Port, _ = strconv.Atoi(port)
```
xrstf, line 68: What happens if the user used `localhost` with a port? Will an empty port lead to errors?

pkg/validation/flags.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (opts *WebhookOpts) Configure(s *webhook.Server) error {
+	s.CertDir = opts.certDir
+	s.CertName = opts.certName
+	s.KeyName = opts.keyName
+	s.Host = opts.listenHost
+	s.Port = opts.listenPort
+
+	// handle deprecated flags
```
xrstf, line 61: Do we need to? The address was IIRC never configurable and is IMHO just an internal thing fully managed by the KKP Operator / Helm chart.

pkg/validation/flags.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (opts *WebhookOpts) AddFlags(fs *flag.FlagSet, includeDeprecatedFlags bool) {
+	if includeDeprecatedFlags {
+		fs.StringVar(&opts.ListenAddress, "seed-admissionwebhook-listen-address", ":8100", "The listen address for the seed amission webhook (Deprecated)")
+		fs.StringVar(&opts.CertFile, "seed-admissionwebhook-cert-file", "", "The location of the certificate file (Deprecated)")
+		fs.StringVar(&opts.KeyFile, "seed-admissionwebhook-key-file", "", "The location of the certificate key file (Deprecated)")
+	}
+	fs.StringVar(&opts.listenHost, "admissionwebhook-listen-host", "", "The listen host for the seed amission webhook")
+	fs.IntVar(&opts.listenPort, "admissionwebhook-listen-port", 8100, "The listen port for the seed amission webhook")
```
xrstf, line 48: FWIW, I kinda like specifying the address to listen on in one single CLI flag, instead of having to split it into 2 fields. Did you do this to save the call the SplitHostPort() ?

xrstf (commented):
Some questions

pkg/validation/seed/validator.go
irozzo-1A: true, this is definitely a bug I'm always assigning the address of `s`. Good catch.

pkg/validation/flags.go
irozzo-1A: Not really, just a matter of taste I think

pkg/validation/flags.go
irozzo-1A: don't know if someone is running the seed controller manager without our helm charts or operator, I think it's rather unlikely, but it does not cost much to keep the support with a deprecation notice

pkg/validation/flags.go
irozzo-1A: yes, it will lead to errors

pkg/validation/flags.go
irozzo-1A: yep, must be is more appropriate

pkg/validation/flags.go
xrstf: Should we then default the port to 80?

irozzo-1A: I don't think, the current behaviour should be consistent with legacy where `host:port` was expected: https://golang.org/pkg/net/http/#Server

pkg/validation/seed/validator_test.go
```diff
@@ -415,13 +433,28 @@ func TestValidate(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			sv := &Validator{
+			var obj []runtime.Object
+			for _, c := range tc.existingClusters {
+				obj = append(obj, c)
+			}
+			for _, s := range tc.existingSeeds {
+				obj = append(obj, s)
+			}
```
xrstf, line 442: This can be shortened to

```go
obj = append(obj, tc.existingClusters...)
obj = append(obj, tc.existingSeeds...)
```

irozzo-1A: true, why not

xrstf (changes requested):
One minor thing

pkg/validation/seed/validator_test.go
irozzo-1A: ah, right. I did it like that in order to have typed slices in the tests:
```
		existingSeeds    []*kubermaticv1.Seed
		existingClusters []*kubermaticv1.Cluster
```
As it's just syntactic sugar and does not add many lines I would rather keep it that way

xrstf (approved):
/approve

irozzo-1A:
@zreigz @xrstf Could you please approve this again? It was not merged due to a conflict.

PR #5873, opened 2020-09-15
Fix peerURL upgrade issues

moelsayed:
Fix a split brain situation that happens when we update the peer url of one of the etcd cluster members. When we have 3 pods, the first pod is updated successfully. During processing the second pod, there was a window where the 3rd pod is removed from the cluster. The result is 1st pod is in the cluster, 3rd pod is out of the cluster, and 2nd pod is trying to join a cluster with broken quorum(1st pod cluster), which would fail. 

This PR fixes the the problem by only matching/removing etcd nodes based on peerurl hostname, not the full peer url. 

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -375,7 +375,7 @@ func (e *etcdCluster) isClusterMember(name string) (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		// if the member is not started yet, it's name would be empty, in that case, we match for peerURL host.
+		// if the member is not started yet, its name would be empty, in that case, we match for peerURL host.
 		if member.Name == name || url.Host == fmt.Sprintf("%s.etcd.%s.svc.cluster.local:2380", e.config.podName, e.config.namespace) {
 			return true, nil
 		}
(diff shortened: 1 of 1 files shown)


cmd/etcd-launcher/main.go
```diff
@@ -366,7 +375,8 @@ func (e *etcdCluster) isClusterMember(name string) (bool, error) {
 		if err != nil {
 			return false, err
 		}
-		if member.Name == name && url.Host == fmt.Sprintf("%s.etcd.%s.svc.cluster.local:2380", e.config.podName, e.config.namespace) {
+		// if the member is not started yet, it's name would be empty, in that case, we match for peerURL host.
```
xrstf: `its name`

cmd/etcd-launcher/main.go
```diff
@@ -406,7 +422,8 @@ func (e *etcdCluster) isHealthyWithEndpoints(endpoints []string) (bool, error) {
 	_, err = client.Get(ctx, "healthy")
 	defer cancel()
 	if err != nil && err != rpctypes.ErrPermissionDenied {
-		return false, err
+		// wait.Poll() doesn't retry if we err out
```
xrstf: This comment assumes the caller of the function. A better comment would be agnostic to the caller and describe the intention, e.g. `// silently swallow/drop transient errors`. At least I think :-)

xrstf (commented):
Super nitpicking in its purest form.

xrstf (approved):
/approve
/cherrypick release/v2.15

PR #5883, opened 2020-09-17
Add operator.kubermatic.io/skip-reconciling annotation for Seeds

### LINKED ISSUE #5882: Legacy Installations: Allow step-by-step Seed Migrations
**User Story**
To facilitate easier migrations from the old `kubermatic` Helm chart to having seed clusters be automatically provisioned by the Kubermatic Operator in the master cluster, we must enable admins t o prevent the Operator from provisioning seeds _immediately_.

To achieve this, a new flag shall be added to `Seed` resources, similar to a `Cluster`'s `pause` flag. But we should _not _ call it "pause", as pausing a usercluster is a totally different thing from "pausing a seed" and we do not want to make the admin believe that they can make the entire seed cluster go into sleep mode, when in reality all that happens is that the Kubermatic Operator will not touch the seed cluster. The existing userclusters on that seed will continue to work and be reconciled by the existing seed-controller-manager.

This is part of #5845.

**Acceptance criteria**
- [ ] A new flag on Seed resources exist to prevent the operator from reconciling them.
- [ ] The flag is not just called "pause", to prevent confusion.
- [ ] The documentation must be clear that this is _only_ meant for the migration and that seed clusters should never be paused in a production setup.

xrstf:
This adds a new flag to prevent the operator from provisioning seed clusters.

Fixes #5882

```release-note
Add operator.kubermatic.io/skip-reconciling annotation to Seeds to allow step-by-step seed cluster upgrades.
```

### DIFF
--- a/docs/zz_generated.seed.yaml
+++ b/docs/zz_generated.seed.yaml
@@ -274,6 +274,11 @@ spec:
         requests:
           cpu: 50m
           memory: 32Mi
+  # PauseProvisioning should only be used when migrating seed clusters, as it
+  # will make the Kubermatic Operator not reconcile the required components
+  # inside the seed cluster (like the seed-controller-manager, webhooks etc.).
+  # This setting does not "disable" a seed or the userclusters on it.
+  pauseProvisioning: false
   # Optional: ProxySettings can be used to configure HTTP proxy settings on the
   # worker nodes in user clusters. However, proxy settings on nodes take precedence.
   proxy_settings:
--- a/docs/zz_generated.seed.yaml
+++ b/docs/zz_generated.seed.yaml
@@ -274,11 +274,6 @@ spec:
         requests:
           cpu: 50m
           memory: 32Mi
-  # PauseProvisioning should only be used when migrating seed clusters, as it
-  # will make the Kubermatic Operator not reconcile the required components
-  # inside the seed cluster (like the seed-controller-manager, webhooks etc.).
-  # This setting does not "disable" a seed or the userclusters on it.
-  pauseProvisioning: false
   # Optional: ProxySettings can be used to configure HTTP proxy settings on the
   # worker nodes in user clusters. However, proxy settings on nodes take precedence.
   proxy_settings:
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -95,7 +95,7 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, seedName string) error {
 
 	// to allow a step-by-step migration of seed clusters, it's possible to
 	// disable the operator's reconciling logic for seeds
-	if seed.Spec.PauseProvisioning {
+	if seed.Annotations[common.SkipReconcilingAnnotation] != "" {
 		log.Info("seed is marked as paused, skipping reconciliation")
 		return nil
 	}
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -95,7 +95,7 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, seedName string) error {
 
 	// to allow a step-by-step migration of seed clusters, it's possible to
 	// disable the operator's reconciling logic for seeds
-	if seed.Annotations[common.SkipReconcilingAnnotation] != "" {
+	if _, ok := seed.Annotations[common.SkipReconcilingAnnotation]; ok {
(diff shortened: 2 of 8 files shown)


docs/zz_generated.seed.yaml
```diff
@@ -274,6 +274,11 @@ spec:
         requests:
           cpu: 50m
           memory: 32Mi
+  # PauseProvisioning should only be used when migrating seed clusters, as it
+  # will make the Kubermatic Operator not reconcile the required components
+  # inside the seed cluster (like the seed-controller-manager, webhooks etc.).
+  # This setting does not "disable" a seed or the userclusters on it.
+  pauseProvisioning: false
```
irozzo-1A: I was thinking that maybe having a label instead (e.g. `operator.k8c.io/skip-reconciling`) as proposed by @moelsayed would make more sense. The only use-case we have at the moment is this migration, I would avoid polluting the API with a field with a such reduced scope. @xrstf WDYT?

xrstf: Good idea. I'll change that.

xrstf: @irozzo-1A done

pkg/controller/operator/seed/reconciler.go
```diff
@@ -93,6 +93,13 @@ func (r *Reconciler) reconcile(log *zap.SugaredLogger, seedName string) error {
 		return nil
 	}
 
+	// to allow a step-by-step migration of seed clusters, it's possible to
+	// disable the operator's reconciling logic for seeds
+	if seed.Annotations[common.SkipReconcilingAnnotation] != "" {
```
irozzo-1A: I would rather do a check on the presence of the annotation:
```
if _, ok := seed.Annotations[common.SkipReconcilingAnnotation]; ok {
```

pkg/ee/conversion/helm.go
```diff
@@ -307,6 +308,10 @@ func ConvertDatacenters(datacenterMeta map[string]provider.DatacenterMeta, globa
 		seed.Kind = "Seed"
 		seed.Namespace = targetNamespace
 
+		if pauseProvisioning {
+			seed.Annotations[common.SkipReconcilingAnnotation] = "true"
```
irozzo-1A: Consistently with my previous comment I would add an empty string, as I think that the discriminant should be the existence of the annotation more than its value

irozzo-1A:
LGTM, apart small comments.

pkg/controller/operator/seed/reconciler.go
xrstf: Well _technically_ in this case both variants are identical.

pkg/ee/conversion/helm.go
xrstf: @irozzo-1A done

pkg/controller/operator/seed/reconciler.go
irozzo-1A: right

irozzo-1A: nope sorry, I was confused. They are not identical. the zero value for strings is "", we cannot distinguish between non present entry and empty string entry without the `ok` value

xrstf: Well good thing I changed the condition.

pkg/ee/conversion/helm.go
```diff
@@ -307,6 +308,10 @@ func ConvertDatacenters(datacenterMeta map[string]provider.DatacenterMeta, globa
 		seed.Kind = "Seed"
 		seed.Namespace = targetNamespace
 
+		if pauseProvisioning {
+			seed.Annotations[common.SkipReconcilingAnnotation] = ""
```
irozzo-1A, line 316: Add nil check to avoid panics e.g.
```
                       if seed.Annotations == nil {
                               seed.Annotations = make(map[string]string)
                       }
```

xrstf: Good catch, fixed.

pkg/ee/cmd/kubermatic-installer/shared.go
```diff
@@ -35,6 +39,10 @@ var (
 		Usage: "Namespace to use for created resources",
 		Value: "kubermatic",
 	}
+	pauseSeedsFlag = cli.BoolFlag{
```
irozzo-1A: I think we are using this command only for this migration, would not be easier to have this defaulted to `true`?

xrstf: Good point

PR #5908, opened 2020-09-21
metrics endpoint for V2

zreigz:
**What this PR does / why we need it**: v2 endpoint for cluster metrics

Fixes #

```release-note
v2 endpoint for cluster metrics: GET /api/v2/projects/{project_id}/clusters/{cluster_id}/metrics
```


pkg/handler/v2/cluster/cluster_test.go
```diff
@@ -1144,6 +1146,193 @@ func TestGetClusterHealth(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	memoryQuantity, err := resource.ParseQuantity("687202304")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testcases := []struct {
+		Name                   string
+		Body                   string
```
lsviben: NIT: Body is not needed, you can just use `nil` for the reader or hardcode an empty string here:
```
req := httptest.NewRequest("GET", fmt.Sprintf("/api/v2/projects/%s/clusters/%s/metrics", test.ProjectName, tc.ClusterToGet), strings.NewReader(tc.Body))

```

lsviben (commented):
one small comment

PR #5915, opened 2020-09-21
Fix operator ingress compatibility

xrstf:
In 1.18, the new `PathType` field was added to Ingresses. It is defaulted to `ImplementationSpecific` by the kube apiserver.

Our code so far never touched the field, but because we set all HTTP paths at once, we inadvertantly overwrite the defaulted value, leading to an endless reconcile loop.

To prevent this, the IngressCreator() now probes for any defaulted PathType and then applies whatever it found. This means that on 1.18+ clusters, we will reconcile a new Ingress twice (once with `nil` and then again with `ImplementationSpecific`).

```release-note
Fix Kubermatic Operator getting stuck in Kubernetes 1.18 clusters when reconciling Ingresses.
```

### DIFF
--- a/pkg/controller/operator/master/resources/kubermatic/common.go
+++ b/pkg/controller/operator/master/resources/kubermatic/common.go
@@ -112,6 +112,23 @@ func IngressCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.N
 				ServicePort: intstr.FromInt(80),
 			}
 
+			// PathType has been added in Kubernetes 1.18 and defaults to
+			// "ImplementationSpecific". To prevent reconcile loops in previous
+			// Kubernetes versions, this code is carefully written to not
+			// overwrite a PathType that has been defaulted by the apiserver.
+			var pathType *extensionsv1beta1.PathType
+
+			// As we control the entire rule set anyway, it's enough to find
+			// the first pathType -- they will all be identical eventually.
+			if rules := i.Spec.Rules; len(rules) > 0 {
+				if http := rules[0].IngressRuleValue.HTTP; http != nil {
+					for _, path := range http.Paths {
+						pathType = path.PathType
+						break
+					}
+				}
+			}
+
 			i.Spec.Rules = []extensionsv1beta1.IngressRule{
 				{
 					Host: cfg.Spec.Ingress.Domain,
@@ -119,15 +136,17 @@ func IngressCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.N
 						HTTP: &extensionsv1beta1.HTTPIngressRuleValue{
 							Paths: []extensionsv1beta1.HTTPIngressPath{
 								{
-									Path: "/api",
+									Path:     "/api",
+									PathType: pathType,
 									Backend: extensionsv1beta1.IngressBackend{
 										ServiceName: apiServiceName,
 										ServicePort: intstr.FromInt(80),
 									},
 								},
 								{
-									Path:    "/",
-									Backend: *i.Spec.Backend,
+									Path:     "/",
+									PathType: pathType,
+									Backend:  *i.Spec.Backend,
 								},
 							},
 						},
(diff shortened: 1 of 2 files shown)


pkg/controller/operator/master/resources/kubermatic/common.go
```diff
@@ -112,22 +112,41 @@ func IngressCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.N
...
 
+			// PathType has been added in Kubernetes 1.18 and defaults to
+			// "ImplementationSpecific". To prevent reconcile loops in previous
+			// Kubernetes versions, this code is carefully written to not
+			// overwrite a PathType that has been defaulted by the apiserver.
+			var pathType *extensionsv1beta1.PathType
+
+			// As we control the entire rule set anyway, it's enough to find
+			// the first pathType -- they will all be identical eventually.
+			if rules := i.Spec.Rules; len(rules) > 0 {
```
irozzo-1A, line 123: This should make it, but I would suggest assigning the current rules to a new variable and getting the relevant pathType with something like:

```
func getPathType(host string, path string, rules []extensionsv1beta1.IngressRule) *extensionsv1beta1.PathType {
	for _, rule := range rules {
		if http := rule.IngressRuleValue.HTTP; rule.Host == host && http != nil {
			for _, p := range http.Paths {
				if p.Path == path {
					return p.PathType
				}
			}
		}
	}
	return nil
}
```

I find it clearer, a bit more robust and future-proof.

irozzo-1A (commented):
I would add a small unit test, it should be fairly easy.

pkg/controller/operator/master/resources/kubermatic/common.go
xrstf: I don't really understand this suggestion. What about your code is clearer than mine? Because it's in a dedicated function? Because you (IMHO needlessly) compare the host?

irozzo-1A: Because it tries to match the `PathType` associated to the same `Rule`, `Path` and not just taking the first `Path` of the first `Rule`. For instance if a rule is added on top of the manifest, it will be removed by the reconcile logic, but its PathType will be preserved and kept for all paths. This scenario is quite remote. Second scenario, is if in future we want to have different TypePath values we can keep this logic, as it is slightly more generic while your is tailored to this specific use case.

What I had in mind is something like:
```
			currRules := i.Spec.Rules
			i.Spec.Rules = []extensionsv1beta1.IngressRule{
				{
					Host: cfg.Spec.Ingress.Domain,
					IngressRuleValue: extensionsv1beta1.IngressRuleValue{
						HTTP: &extensionsv1beta1.HTTPIngressRuleValue{
							Paths: []extensionsv1beta1.HTTPIngressPath{
								{
									Path:     "/api",
									PathType: getPathType("", "/api", currRules),
									Backend: extensionsv1beta1.IngressBackend{
										ServiceName: apiServiceName,
										ServicePort: intstr.FromInt(80),
									},
								},
								{
									Path:     "/",
									PathType: getPathType("", "/api", currRules),
									Backend:  *i.Spec.Backend,
								},
							},
						},
					},
				},

```

I agree that the gain is small, but the added complexity is small too. I'm not going to be pushy on that, take it as a suggestion.

xrstf: Let's burn that bridge when we come to it. I can see the point of keeping the "wrong" PathType in some exotic situations, but this might also be seen as a feature :-D

irozzo-1A: Well, having two paths with different types in the same ingress does not seem to be that "exotic", anyway it's your call.

xrstf: Having two path types in *this particular* Ingress that is managed by the operator and therefore not the place where anyone can feasably add their own paths or path types, is indeed something I would call exotic ;-)

I had long discussions with previous project members about how much we want to "allow" others to modify the resources we reconcile (e.g. should the operator remove unexpected volume mounts on a deployment?) and the consensus was that for "our" resources, we do not want interference and will not "be kind" to other modifications. This means that the KKP operator does not make any statement over what modifications will survive and users should generally never touch "our" resources. This is the assumption under which all reconcile functions are written in KKP. Just for a little background. :)

xmudrii (approved):
Looks fine to me.
/lgtm

PR #5927, opened 2020-09-22
Add  Kuberentes 1.16.15, 1.17.12, 1.19.2

xrstf:
This just bumps the supported Kubernetes versions. 1.18.9 is left out because there is no image on quay.io for it [yet?].

```release-note
Add Kuberentes 1.16.15, 1.17.12, 1.19.2
```

### DIFF
--- a/charts/kubermatic/static/master/versions.yaml
+++ b/charts/kubermatic/static/master/versions.yaml
@@ -15,11 +15,9 @@ versions:
   version: 1.17.12
 - type: kubernetes
   version: 1.18.6
-- type: kubernetes
-  version: 1.18.8
 - default: true
   type: kubernetes
-  version: 1.18.9
+  version: 1.18.8
 - type: kubernetes
   version: 1.19.0
 - type: kubernetes
--- a/charts/kubermatic/static/master/versions.yaml
+++ b/charts/kubermatic/static/master/versions.yaml
@@ -6,16 +6,24 @@ versions:
 - type: kubernetes
   version: 1.16.14
 - type: kubernetes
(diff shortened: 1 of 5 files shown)


charts/kubermatic/static/master/versions.yaml
```diff
@@ -5,17 +5,25 @@ versions:
...
+- type: kubernetes
+  version: 1.18.8
 - default: true
   type: kubernetes
-  version: 1.18.8
+  version: 1.18.9
 - type: kubernetes
   version: 1.19.0
+- type: kubernetes
+  version: 1.19.2
```
xmudrii, line 24: As far as I see, we try to support as many patch releases as possible. To keep this consistent, should we also add support for 1.19.1? I don't see any value in doing that, so this is only for the sake of consistency.

xrstf: I'm not sure if/why anyone would want to use 1.19.1 if 1.19.2 is supported? But i don't know ^^

xmudrii: That's what I was thinking as well and that's why I just said for the sake of consistency. We can leave it as it is right now.

xrstf: One argument against it: I cannot easily have one PR that tests both patch releases ;-)

xmudrii (approved):
/lgtm
/retest

xrstf:
/hold

quay.io/poseidon/kubelet:v1.18.9 does not exist.

xmudrii (approved):
/lgtm

PR #5934, opened 2020-09-24
Makefile improvements

xmudrii:

* Remove lint from make all/default target

Lint takes a lot of resources and time to finish, so we should not run
it by default. We have a dedicated lint target that can be used to
properly run it.

* Extend check-dependencies make target with go mod tidy

* Run go mod tidy

* Remove check make target

It appears that the check target is not used anywhere, so let's remove
it.

```release-note
NONE
```

/assign @xrstf @kron4eg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,7 @@ verify:
 	./hack/verify-api-client.sh
 
 check-dependencies:
+	go mod tidy
 	go mod verify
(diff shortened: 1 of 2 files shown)


xrstf (approved):
/approve

kron4eg (approved):

PR #5948, opened 2020-09-25
etcd-launcher image tag is customizable

### LINKED ISSUE #5909: etcd image tag should be configurable
**User Story**
As an ops engineer 
I want to have option to avoid changes to etcd-launcher image when deploying newer version of the kubermatic images 
So that etcd statefulsets of clusters are not recreated

**Acceptance criteria**

furkhat:

Shipping new version of kubermatic will update etcd statefulsets etcd-launcher tag and they will be recreated even though new version doesn't have any updates related to etcd-launcher. For better service availability reasons we want to be able to set etcd-launcher image tag.

Fixes #5909

**Documentation**:

```release-note
seed-controller-manager option `etcd-launcher-image` receives tagged image
```

### DIFF
--- a/cmd/seed-controller-manager/options.go
+++ b/cmd/seed-controller-manager/options.go
@@ -140,7 +140,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.StringVar(&c.oidcIssuerClientID, "oidc-issuer-client-id", "", "Issuer client ID")
 	flag.StringVar(&c.oidcIssuerClientSecret, "oidc-issuer-client-secret", "", "OpenID client secret")
 	flag.StringVar(&c.kubermaticImage, "kubermatic-image", resources.DefaultKubermaticImage, "The location from which to pull the Kubermatic image")
-	flag.StringVar(&c.etcdLauncherImage, "etcd-launcher-image", resources.DefaultEtcdLauncherImage, "The location from which to pull the etcd launcher image")
+	flag.StringVar(&c.etcdLauncherImage, "etcd-launcher-image", resources.DefaultEtcdLauncherImage, fmt.Sprintf("The location from which to pull the etcd launcher image, if image tag is not specified tag %s is used", resources.KUBERMATICCOMMIT))
 	flag.StringVar(&c.dnatControllerImage, "dnatcontroller-image", resources.DefaultDNATControllerImage, "The location of the dnatcontroller-image")
 	flag.StringVar(&c.namespace, "namespace", "kubermatic", "The namespace kubermatic runs in, uses to determine where to look for datacenter custom resources")
 	flag.IntVar(&c.apiServerDefaultReplicas, "apiserver-default-replicas", 2, "The default number of replicas for usercluster api servers")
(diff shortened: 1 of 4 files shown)


cmd/seed-controller-manager/options.go
```diff
@@ -140,7 +140,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.StringVar(&c.oidcIssuerClientID, "oidc-issuer-client-id", "", "Issuer client ID")
 	flag.StringVar(&c.oidcIssuerClientSecret, "oidc-issuer-client-secret", "", "OpenID client secret")
 	flag.StringVar(&c.kubermaticImage, "kubermatic-image", resources.DefaultKubermaticImage, "The location from which to pull the Kubermatic image")
-	flag.StringVar(&c.etcdLauncherImage, "etcd-launcher-image", resources.DefaultEtcdLauncherImage, "The location from which to pull the etcd launcher image")
+	flag.StringVar(&c.etcdLauncherImage, "etcd-launcher-image", resources.DefaultEtcdLauncherImage, "The location from which to pull the etcd launcher image, if image tag is not corresponding dynamically derived kubermatic commit is used")
```
xrstf: Nit pick, but can we make the phrasing a bit easier to parse, like

> to pull the etcd-launcher; if image tag is not corresponding [to what?], *then* a dynamically derived Kubermsatic commit is used [which one?]

The seed-controller should have access to the version/git commit it was built for, maybe it could be nice to just fmt.Sprintf() the actual docker image tag that would be used. like

```
fmt.Sprintf("Do that, if not then the image tag %s is used", versions.KUBERMATICDOCKERTAG) // pseudo code
```

xrstf (commented):
Minor nitpick

xrstf:
> For better service availability reasons we want to be able to set etcd-launcher image tag.

Do you plan on running indefinitly with a version skew here or do you plan on then migrating userclustes manually step by step? If so, I think we can try to do better and automatically migrate (etcd-)clusters sequentially. Allowing custom tags might open the door for accidental foot shooting when the setting is forgotten or when inadvertently there are actual fixes in newerd etcd-launcher versions.

furkhat:
> Do you plan on running indefinitly with a version skew here or do you plan on then migrating userclustes manually step by step? If so, I think we can try to do better and automatically migrate (etcd-)clusters sequentially. Allowing custom tags might open the door for accidental foot shooting when the setting is forgotten or when inadvertently there are actual fixes in newerd etcd-launcher versions.

We plan to migrate the customer clusters by manually selecting one after one. This is also precautionary measures for future, we don't want the etcd pods recreated if there was no change to etcd-launcher.

xrstf:
This seems to be an issue every KKP admin is facing in their clusters. I don't think we should recommend every admin to manually adjust tags on every KKP release.

furkhat:
Overall images being coupled to single tag which is git version of last commit of kubermatic repo I find inconvenient. Why things can't be deployed independently? This is true for not only for `etcd-launcher`. WDYT?

moelsayed:
As discussed offline. This will be closed for now.

PR #5962, opened 2020-09-28
Provide way to skip certificate CR creation

### LINKED ISSUE #5554: Add support to provide own certificates for ingresses
**User Story**

Kubermatic 2.14 moved the creation of certificate resources from a dedicated Helm chart to the Kubermatic and IAP charts. This makes the affected charts relying on the cert-manager CRDs. Some on-premise installations use an internal CA since their ingresses are not exposed to the public internet.

With Kubermatic 2.13 or older, a simple workaround for such environments was skipping the installation for cert-manager and the certificates chart. Afterwards an own certificate could be deployed to the `default/kubermatic-tls-certificates` secret, which was configured as default certificate for the ingress controller.

Since the ingress controller Helm chart has a `.Values.nginx.extraArgs` parameter, setting a default certificate is easily possible. However, the dependency of the cert-manager CRDs make it impossible to run Kubermatic without cert-manager.

**Acceptance criteria**

The minimal solution would be to allow deploying Kubermatic without creating certificate resources. In best case also a custom certificate could be installed via Helm charts. However, I see the difficulty that some customers will use a wildcard certificate (which has to be deployed as a default certificate for the ingress controller), other will use an individual certificate per ingress resource.

irozzo-1A:
This PR provides a way of skipping certificate creation. This is useful for users that provide their own certificate, enabling them to skip cert-manager creation.

Fixes #5554 

This is especially useful for migration from v2.13 to v2.14 where the Certificate CR was moved to kubermatic chart.

**Documentation**:

```release-note
Provide a way of skipping Certificate cert-manager resources.
```

### DIFF
--- a/charts/values.example.ce.yaml
+++ b/charts/values.example.ce.yaml
@@ -48,6 +48,9 @@ dex:
     userID: 08a8684b-db88-4b73-90a9-3cd1661f5466
 
   # the cert-manager Issuer (or ClusterIssuer) responsible for managing the certificates
+  # If you want to deploy your own certificate without relying on cert-manager
+  # uncomment the next line and remove subsequent certIssuer configuration.
+  # certIssuer: null
   certIssuer:
     # For generating a certificate signed by a trusted root authority replace
     # with "letsencrypt-prod".
--- a/charts/kubermatic/Chart.yaml
+++ b/charts/kubermatic/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubermatic
-version: 1.1.15
+version: 1.1.16
 appVersion: '__KUBERMATIC_TAG__'
 description: Kubermatic chart for master and/or seed clusters.
 keywords:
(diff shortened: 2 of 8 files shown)


irozzo-1A:
Note that when using the operator based installation it wont be still possible to skip the certificate creation after this PR is integrated. This will be addressed with a dedicated PR.

charts/values.example.ce.yaml
```diff
@@ -48,6 +48,9 @@ dex:
     userID: 08a8684b-db88-4b73-90a9-3cd1661f5466
 
   # the cert-manager Issuer (or ClusterIssuer) responsible for managing the certificates
+  # If you want to deploy your own certificate without relying on cert-manager
+  # uncomment the next line and remove subsequent certIssuer configuration.
+  # certIssuer: null
```
xrstf, line 53: Can you put this comment also in the `charts/oauth/values.yaml`?

xrstf (commented):
One suggestion

charts/values.example.ce.yaml
irozzo-1A: ok

xrstf (approved):
/approve

PR #5536, opened 2020-06-18
Add etcd-launcher

moelsayed:
**What this PR does / why we need it**:
- Removes the current bash entrypoint for user cluster etcd statefulset. 
- Removes old code required for etcd-operator migrations. 
- Adds etcd-launcher binary and uses it as an entrypoint for the user cluster etcd statefulset
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5542

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5973, opened 2020-09-30
Etcd-launcher feature flag

moelsayed:
This PR introduces a feature flag for the etcd-launcher feature initially added in #5536. It also handles the situations where existing clusters already have the etcd-launcher enabled, to avoid breaking etcd on weekly releases upgrades.
Fixes #

**Documentation**:

```release-note
Adds feature flag `etcdLauncher` around etcd-launcher.
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -172,7 +172,7 @@ func main() {
 		}
 		defer client.Close()
 
-		if _, err := client.MemberAdd(context.Background(), []string{fmt.Sprintf("https://%s.etcd.%s.svc.cluster.local:2380", e.config.podName, e.config.namespace)}); err != nil {
+		if _, err := client.MemberAdd(context.Background(), []string{fmt.Sprintf("http://%s.etcd.%s.svc.cluster.local:2380", e.config.podName, e.config.namespace)}); err != nil {
 			log.Fatalf("failed to join cluster: %v", err)
 		}
 		log.Info("joined etcd cluster succcessfully.")
@@ -193,13 +193,13 @@ func (e *etcdCluster) updatePeerURL() error {
 		if err != nil {
 			return err
 		}
-		if member.Name == e.config.podName && peerURL.Scheme == "http" {
+		if member.Name == e.config.podName && peerURL.Scheme == "https" {
 			client, err := e.getClusterClient()
 			if err != nil {
 				return err
 			}
 			defer client.Close()
-			peerURL.Scheme = "https"
+			peerURL.Scheme = "http"
 			_, err = client.MemberUpdate(context.Background(), member.ID, []string{peerURL.String()})
 			if err != nil {
 				return err
@@ -213,7 +213,7 @@ func (e *etcdCluster) updatePeerURL() error {
 func initialMemberList(n int, namespace string) []string {
 	members := []string{}
 	for i := 0; i < n; i++ {
-		members = append(members, fmt.Sprintf("etcd-%d=https://etcd-%d.etcd.%s.svc.cluster.local:2380", i, i, namespace))
+		members = append(members, fmt.Sprintf("etcd-%d=http://etcd-%d.etcd.%s.svc.cluster.local:2380", i, i, namespace))
 	}
 	return members
 }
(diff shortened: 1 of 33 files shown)


pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -138,6 +138,8 @@ const (
 	// ClusterFeatureRancherIntegration enables the rancher server integration feature.
 	// It will deploy a Rancher Server Managegment plane on the seed cluster and import the user cluster into it.
 	ClusterFeatureRancherIntegration = "rancherIntegration"
+
+	ClusterFeatureEtcdLauncher = "etcdLauncher"
```
xrstf, line 144: Please add a comment to explain what this feature flag does exactly.

pkg/controller/seed-controller-manager/pvwatcher/pvwatcher_controller.go
```diff
@@ -111,6 +111,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName || cluster.Spec.Pause {
 		return reconcile.Result{}, nil
 	}
+	if flag := cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]; !flag {
```
xrstf: Nit: No need to put this into a dedicated temporary variable. This is true for many parts of this PR.

pkg/resources/etcd/statefulset.go
```diff
@@ -17,9 +17,12 @@ limitations under the License.
 package etcd
 
 import (
+	"bytes"
 	"fmt"
 	"strconv"
+	"text/template"
 
+	"github.com/Masterminds/sprig"
```
xrstf, line 25: Please separate 3rd party non-k8s imports with an empty line from the kubermatic packages.

xrstf (changes requested):
Reviewed

pkg/resources/etcd/statefulset.go
```diff
@@ -95,28 +100,42 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
...
-				{
-					Name:            "etcd-launcher-init",
-					Image:           data.EtcdLauncherImage() + ":" + resources.KUBERMATICCOMMIT,
-					ImagePullPolicy: corev1.PullIfNotPresent,
-					Command:         []string{"/bin/cp", "/etcd-launcher", "/opt/bin/"},
-					VolumeMounts: []corev1.VolumeMount{
-						{
-							Name:      "launcher",
-							MountPath: "/opt/bin/",
+			// luancher feature gate was added later in the process. So there is the possiblity to have some clusters with
```
irozzo-1A: s/luancher/launcher/

irozzo-1A: I'm not totally convinced we need this logic. The point is that if we want to fallback, disabling the `etcd-launcher` feature is not enough, and we also need to manually remove the init container from the manifest.

It's true that we have already RC versions with this feature activated by default, but they are not intended for production. I think we are still in time to avoid this.

xrstf: If we don't absolutely need this to fix an existing etcd cluster, I agree with @irozzo-1A and we should skip adding this logic. We would probably never get rid of it, even long after any possible migration is done.

irozzo-1A:
I would add some unit test coverage, from what I see the `load_files_test` only covers the case in which the feature is disabled.

cmd/etcd-launcher/main.go
```diff
@@ -290,8 +290,8 @@ func etcdCmd(config *config) []string {
 		fmt.Sprintf("--advertise-client-urls=https://%s.etcd.%s.svc.cluster.local:2379,https://%s:2379", config.podName, config.namespace, config.podIP),
 		fmt.Sprintf("--listen-client-urls=https://%s:2379,https://127.0.0.1:2379", config.podIP),
 		fmt.Sprintf("--listen-metrics-urls=http://%s:2378,http://127.0.0.1:2378", config.podIP),
-		fmt.Sprintf("--listen-peer-urls=https://%s:2380", config.podIP),
-		fmt.Sprintf("--initial-advertise-peer-urls=https://%s.etcd.%s.svc.cluster.local:2380", config.podName, config.namespace),
+		fmt.Sprintf("--listen-peer-urls=http://%s:2380", config.podIP),
```
irozzo-1A, line 293: you removed support for secure peer communication at the moment, right?

pkg/resources/etcd/testdata/etcd-command-with-launcher.golden.sh
```diff
@@ -0,0 +1 @@
+/opt/bin/etcd-launcher -namespace $(NAMESPACE) -etcd-cluster-size $(ETCD_CLUSTER_SIZE) -pod-name $(POD_NAME) -pod-ip $(POD_IP) -api-version $(ETCDCTL_API) -token $(TOKEN)
```
irozzo-1A, line 1: Is this used somewhere?

cmd/etcd-launcher/main.go
moelsayed: Yes, I just update the removed the other options :+1:

pkg/resources/etcd/testdata/etcd-command-with-launcher.golden.sh
moelsayed: Yes, it's used by the test to compare the output of the command function.

irozzo-1A: Ok, I missed it https://github.com/kubermatic/kubermatic/blob/2e30c15e5123cfa3d381ccaaa5f9641f561f3111/pkg/resources/etcd/statefulset_test.go#L76

PR #5987, opened 2020-10-02
[release/v2.13] Add configurable default addon enforce option

### LINKED ISSUE #4979: enforcement of default addons in user cluster

moelsayed:
Cherry-pick configurable default addon enforce option, along with related code
Fixes #4979

This PR extracts the basic logic from #5193, but since v2.13.x doesn't have a generated list of addons, the PR only adds the enforce labels only when addons are added using a list. When an addons file is used, the user should add the labels manually to the yaml Addon file.

**Documentation**:

```release-note
Add a configuration flag for seed-controller-manager to enforce default addons on userclusters. Disabled by default.
```


.prow.yaml
```diff
@@ -292,7 +292,7 @@ presubmits:
   # e2e tests for Kubernetes 1.15
   #########################################################
 
-  - name: pre-kubermatic-e2e-aws-coreos-1.15
```
irozzo-1A, line 295: We switched to `flatcar` in 2.14 and 2.15, maybe we should do the same for consistency.

api/pkg/resources/address/address_test.go
```diff
@@ -22,7 +22,7 @@ const (
 	fakeClusterNamespaceName = "cluster-ns"
 	// This must be the A record for *.europe-west3c.dev.kubermatic.io and
 	// *.alias-europe-west3-c.dev.kubermatic.io
-	externalIP = "35.198.93.90"
```
irozzo-1A, line 25: I'm cherry picking the proper fix for this here: https://github.com/kubermatic/kubermatic/pull/6020

.prow.yaml
irozzo-1A: well, probably in v2.13 was not supported yet

irozzo-1A: Yep, I confirm.

api/pkg/resources/address/address_test.go
moelsayed: Sounds good. I am going to keep it though until that PR is merged, otherwise I won't pass tests.

irozzo-1A: sure

moelsayed:
Since this is considered a change in behavior, not technically a bug fix. I am updating the charts to have it disabled by default so avoid the change in behavior for customers who don't need this.

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -242,7 +259,9 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, addo
 	if err := r.ensureFinalizerIsSet(ctx, addon); err != nil {
 		return nil, fmt.Errorf("failed to ensure that the cleanup finalizer existis on the addon: %v", err)
 	}
-
+	if err := r.ensureResourcesCreatedConditionIsSet(ctx, addon); err != nil {
+		return nil, fmt.Errorf("failed to set add ResourcesCreated Condition: %v", err)
```
xrstf: `set add` should be either `set` or `add` :)

api/pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -540,3 +559,53 @@ func (r *Reconciler) ensureRequiredResourceTypesExist(ctx context.Context, log *
...
+	}
+	return -1, nil
+}
+
+func addonResourcesCreated(addon *kubermaticv1.Addon) bool {
+	_, cond := getAddonCondition(addon, kubermaticv1.AddonResourcesCreated)
+	if cond != nil && cond.Status == corev1.ConditionTrue {
+		return true
+	}
+	return false
```
xrstf: This can be written as a single `return` statement.

xrstf (approved):
/approve
/hold

Two minor things that could be improved, but since this is just a backport, I'm fine with the current state also. :-) Please unhold when @irozzo-1A if @irozzo-1A is also good with this.

PR #5991, opened 2020-10-05
fix error message for PVs on cluster deletion

qeqar:
We had the same LB error message in shouldDeleteLBs an shouldDeletePVs

```release-notes
NONE
```


xrstf:
/ok-to-test
/test all
/approve

Thanks and welcome to Kubermatic :-)

xrstf (approved):
/approve

PR #5995, opened 2020-10-05
Constraint template controller

### LINKED ISSUE #5904: Constraint template controller
**User Story**
Implement constraint template controller which reconciles the CTs available on the seed cluster to the user cluster

lsviben:
Adds a controller for constraint templates which syncs seed kubermatic constraint templates to the user cluster. 

Fixes #5904 

```release-note
NONE
```


pkg/api/v1/types.go
```diff
@@ -1585,6 +1585,8 @@ const (
 	UserClusterRoleCleanupFinalizer = "kubermatic.io/user-cluster-role"
 	// ExternalClusterKubeconfigCleanupFinalizer indicates that secrets for kubeconfig still need cleanup
 	ExternalClusterKubeconfigCleanupFinalizer = "kubermatic.io/cleanup-kubeconfig-secret"
+	// ConstraintTemplateCleanupFinalizer indicates that synced Constraint Templates on user cluster need cleanup
+	ConstraintTemplateCleanupFinalizer = "kubermatic.io/cleanup-constraint-templates"
```
moadqassem: Please add `gatekeepr` to the finalizer name:
```go
GatekeeperConstraintTemplateCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-constraint-templates"
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// This controller syncs the kubermatic constraint templates to gatekeeper constraint templates on the user cluster.
+	controllerName = "constraint_template_controller"
```
moadqassem: Please add `gatekeeper` in the controller name:
```go
controllerName = "gatekeeper-constraint_template_controller"
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	// Watch for changes to ConstraintTemplates
+	if err = c.Watch(ctSource, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("failed to establish watch for the ConstraintTemplates %v", err)
+	}
+
+	return nil
+}
+
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.With("ConstraintTemplate", request.Name)
```
moadqassem: Why not logging the whole request? I mean what value we would have if we have the name only?
```go
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	log.Debug("Reconciling")
+
+	constraintTemplate := &v1.ConstraintTemplate{}
+	if err := r.seedClient.Get(r.ctx, request.NamespacedName, constraintTemplate); err != nil {
+		if kerrors.IsNotFound(err) {
+			log.Debug("constraint template not found, returning")
```
moadqassem: Would be nice to have the name of the constraint while reconciling: 
```go
log.Debugw("constraint template in namespace not found, returning", "constraint-name", constraintTemplate.Name, "namespace", constraintTemplate.Namespace)
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get constraint template: %v", err)
```
moadqassem: Please add the name of the constraint in the error message: 
```go
return reconcile.Result{}, fmt.Errorf("failed to get constraint template %s: %v", constraintTemplate.Name, err)
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+	if constraintTemplate.DeletionTimestamp != nil {
+		if !kuberneteshelper.HasFinalizer(constraintTemplate, kubermaticapiv1.ConstraintTemplateCleanupFinalizer) {
+			return reconcile.Result{}, nil
+		}
+		if err := r.userClient.Delete(r.ctx, &v1beta1.ConstraintTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: constraintTemplate.Name,
+			},
+		}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to delete constraint template: %v", err)
```
moadqassem: Please add the name of the constraint in the error message:
```go
return reconcile.Result{}, fmt.Errorf("failed to delete constraint template %s: %v", constraintTemplate.Name, err)
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+		return reconcile.Result{}, fmt.Errorf("failed to get constraint template: %v", err)
+	}
+
+	if constraintTemplate.DeletionTimestamp != nil {
+		if !kuberneteshelper.HasFinalizer(constraintTemplate, kubermaticapiv1.ConstraintTemplateCleanupFinalizer) {
+			return reconcile.Result{}, nil
+		}
+		if err := r.userClient.Delete(r.ctx, &v1beta1.ConstraintTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: constraintTemplate.Name,
```
moadqassem: Please add the namespace of the yet to be deleted resource.

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+			},
+		}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to delete constraint template: %v", err)
+		}
+
+		kuberneteshelper.RemoveFinalizer(constraintTemplate, kubermaticapiv1.ConstraintTemplateCleanupFinalizer)
+		if err := r.seedClient.Update(r.ctx, constraintTemplate); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to remove constraint template finalizer: %v", err)
```
moadqassem: Please add the name of the constraint in the error message:
```go
return reconcile.Result{}, fmt.Errorf("failed to remove constraint template finalizer %s: %v", constraintTemplate.Name, err)
```

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}{
+		{
+			name:               "sync ct to user cluster",
+			constraintTemplate: genConstraintTemplate("requiredlabels"),
+			requestName:        "requiredlabels",
+			expectedCT:         genConstraintTemplate("requiredlabels"),
+		},
+	}
+
+	for idx := range testCases {
```
moadqassem: You can use this loop instead, and then you don't need to worry about the index in the next line:
```go
```
so you can remove line 53

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+		tc := testCases[idx]
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
moadqassem: Why do you need this ?

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Watch for changes to ConstraintTemplates
+	if err = c.Watch(ctSource, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("failed to establish watch for the ConstraintTemplates %v", err)
+	}
+
+	return nil
+}
+
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
moadqassem: I would suggest to move the logic of the reconcile completely in the private method `reconcile`. Starting from line 99 to line 129.  Take a look at the gist https://gist.github.com/moadqassem/4a4cb9d065a4b52942a63b89c47e0126

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+			sch, err := v1beta1.SchemeBuilder.Build()
+			if err != nil {
+				t.Fatalf("building gatekeeper scheme failed: %v", err)
+			}
+			seedClient := fakectrlruntimeclient.NewFakeClientWithScheme(scheme.Scheme, tc.constraintTemplate)
```
moadqassem: Nit: can be defined in the testCase struct

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+			userClient := fakectrlruntimeclient.NewFakeClientWithScheme(sch)
```
moadqassem: Nit: can be defined in the testCase struct.

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller_test.go
```diff
@@ -0,0 +1,123 @@
...
+func TestReconcile(t *testing.T) {
+	testCases := []struct {
+		name               string
+		constraintTemplate *v1.ConstraintTemplate
+		requestName        string
+		expectedCT         *v1.ConstraintTemplate
+	}{
+		{
+			name:               "sync ct to user cluster",
+			constraintTemplate: genConstraintTemplate("requiredlabels"),
```
moadqassem: Please have the value `requiredlabels` as a constant.

moadqassem (changes requested):

pkg/controller/user-cluster-controller-manager/constraint-template-controller/controller.go
lsviben: its cluster scoped so no need here

lsviben:
Hi @moadqassem , @zreigz ,

I added new commits which change the approach from earlier. Now the constraint template controller will run as part of the master-controller, instead of the user-cluster controller, as it made more sense considering Constraint Templates being master resources plus it avoids some problems that I encountered with the previous solution.

moadqassem:
@lsviben Can you please brief me what kinda issues you had? I am not generally against having the gatekeeper on the master level(even though I would prefer to have on seed level which is kinda easier to manage, I mean what if I would like to activate this thing in Seed 1 and deactivate on Seed 2 for example?).

zreigz:
@moadqassem 
First, the Kubermatic admin installs ConstraintTemplates on the master. They are global and manage by the admin in the admin panel or manually. Second, we have to make a copy of those constraint templates for every newly created user cluster.

This controller should handle the situation when the admin creates/deletes a new template. When we have multiple seed controllers we have a problem to synchronize deleted constraint template resources on every seed. Finally, we have to delete one global  ConstraintTemplates on the master and this is the main reason why we decided to move the logic to this controller.

The goal for this controller is to manage resource which is installed on master so it's very natural to have this logic on the master.
We can achieve all requirements for this controller in an easy way.

zreigz:
@lsviben 
your controller just watches ConstarintTemplates. A newly created cluster won't have already existing templates. Am I right?
So also we need a seed controller to watch clusters and copy existing templates if are not in the deleted state.

lsviben:
@moadqassem 
Right, as @zreigz mentioned, the problem I had when It was on the user-cluster controller was that we would need to put a cleanup finalizer on the CT for each cluster with OPA integration. And it could get messy when deleting CTs plus there are also issues if a cluster is deleted or malfunctioning, because its finalizer needs to be removed from the CT.

Doing it on the seed cluster controller we would have a smaller but similar issue. 

For applying something on Seed 1 and not on Seed 2, there is a planned feature for the future in which we would add some filters to the CT to apply only on some clusters, but its a good idea to also have the option to filter by seed.

Also if CT was a seed resource we would need to have users repeat them for each seed, or have some default CTs in the master and some controller which syncs them to the seeds. 

All in all, for us it looked like the solution which is most natural and solves all these problems is to have the controller on master level.

pkg/controller/master-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,310 @@
...
+// Reconcile reconciles the kubermatic constraint template on the master cluster to all seed cluster user clusters
+// which have opa integration enabled
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.With("ConstraintTemplate", request)
+	log.Debug("Reconciling")
+
+	constraintTemplate := &kubermaticv1.ConstraintTemplate{}
+	if err := r.masterClient.Get(r.ctx, request.NamespacedName, constraintTemplate); err != nil {
+		if kerrors.IsNotFound(err) {
+			log.Debugw("constraint template not found, returning", "constraint-name", constraintTemplate.Name)
```
xrstf: If it's not found, then constraintTemplate will be empty. Also, the name has already been assigned to the log in the `r.log.With()` line, no need to add it here again.

pkg/controller/master-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,310 @@
...
+		return fmt.Errorf("failed to create watch for constraintTemplates: %v", err)
+	}
+
+	return nil
+}
+
+// Reconcile reconciles the kubermatic constraint template on the master cluster to all seed cluster user clusters
+// which have opa integration enabled
+func (r *reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.With("ConstraintTemplate", request)
```
xrstf: Please call this field `request` to be consistent with other controllers.

xrstf:
> First, the Kubermatic admin installs ConstraintTemplates on the master. 

Why? Apparently they influence User Clusters, so why not create them on Seeds?

> They are global and manage by the admin in the admin panel or manually.

So the admin panel (i.e. the Kubermatic API) can then just create them in each Seed. Or have a controller that just syncs them over (we already have https://github.com/kubermatic/kubermatic/tree/master/pkg/controller/master-controller-manager/seed-sync).

> The goal for this controller is to manage resource which is installed on master so it's very natural to have this logic on the master.

By that logic we can just get rid of Seeds entirely and manage everything from the master, no? We have Seeds to spread out the reconciling logic and keep latencies short. Keeping a map of connections to each and every user cluster in every Seed is just not the best approach.

I appreciate that this "is easier", but that's not a good reason to violate the entire KKP architecture.

zreigz:
> > First, the Kubermatic admin installs ConstraintTemplates on the master.
> 
> Why? Apparently they influence User Clusters, so why not create them on Seeds?
> 


This comes from the proposal: https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/opa-integration.md#implementation

zreigz:
> By that logic we can just get rid of Seeds entirely and manage everything from the master, no? We have Seeds to spread out the reconciling logic and keep latencies short. Keeping a map of connections to each and every user cluster in every Seed is just not the best approach.
> 
> I appreciate that this "is easier", but that's not a good reason to violate the entire KKP architecture.

We just follow what is implemented right now, not future plans. Can you explain how this code violates the entire KKP architecture? The same solution is done for: projects, users, RBACs, extern cluster, datacenters, global settings, user settings.

xrstf:
> This comes from the proposal: https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/opa-integration.md#implementation

I'm sorry, but I must have missed that during review, but

> High level overview on how it would work is that a _seed(master) cluster_ has a list of some default ConstraintTemplates deployed

(emphasis mine) is not really clear as to where things are living. 

> Can you explain how this code violates the entire KKP architecture? 

The master cluster manages the seeds, the seeds then manage the user clusters. We have Seeds to keep the latency between control planes and the worker nodes small, and to spread the control plane load across multiple seed clusters. As detailed in https://docs.kubermatic.com/kubermatic/v2.15/concepts/architecture/

> projects, users, RBACs, extern cluster, datacenters, global settings, user settings.

Not 100% sure about the RBAC stuff, but all the rest has nothing to do with reconciling user clusters. External clusters do not have a control plane attached, so they don't impose any load (yet, still unsure about the future), projects and users do not trigger reconcilings and neither do global settings or user settings.

But _all_ other things that we manage inside the user cluster are managed _on the Seed_ (in either the seed-ctrlmgr or the usercluster-ctrlmgr).

moadqassem:
@lsviben @zreigz So here are my thoughts about the current implementation. I understand that having every action comes from one place(master controller in this case) will make your life easier because one change in the resource will be propagated to all clusters(which is btw a big disadvantage too due to the lack of having custom setup). I honestly feel uneasy about the code over here: 
https://github.com/kubermatic/kubermatic/pull/5995/files#diff-4c6ff73433caffc8e3d859053bc31edecf079cf8a6508bcd17eacf533d6d1cbdR218-R245

This couples all seeds and all the user clusters in one place and make the update(or action as you call it) comes only from one place which is the master where the real update is actually bounded and should be executed on seed and user cluster as the PoC suggests: 

> Admins can add more and these are shared across all user clusters which have gatekeeper enabled, but admins have the possibility to mark some CTs for example just for AWS datacenter, so they are shared just to AWS DC clusters

Adding filters and conditions in order to cope with a global resource being shared along the whole architecture(which btw can bring up a lot of corner cases which in turns would make the logic fragile) can be simply avoided by deploying the resources where they belong. If the CTs are deployed everywhere unconditionally I would agree, but when you say this DC in that seed can have it and that seed can't then this is something for the seed controller to decide not the master.

lsviben:
Hey,

so just to sum up your proposal, correct me if I am wrong:

what you guys suggest is to have the CTs created and managed on the Master cluster.
Then a master-controller like "seed-sync" syncs them over to the seeds.
From there on the CT-controller on the seed-controller level does the syncing to the user clusters. 

PRO:
- we get rid of the big loop over all seeds and clusters (one cluster not working stops reconciles in other seeds and clusters)
- more localized for latency issues as @xrstf mentions

CON:
- we need a new controller for the CT syncing
- finalizers and cleanup are a bit more complicated as master CT would need to wait for all seed CTs to cleanup all user clusters and then could delete

Other way is the current one in which we control all through the master cluster controller. 

PRO: 
- cleanup is easier
- controlling all from one place lowers the complexity a bit

CON:
- the big loop in which one malfuncioning cluster can mess up the whole reconcile in all seeds
- possible scalability issues (holding all user cluster clients)

So for the filtering part, this is still not clear what exactly we want to do and where. IMO in both approaches corner cases. I agree that in the seed-controller case it would be easier to control in which seeds the CTs would be used as we could add that logic to the seed CT syncer part.

@zreigz @xrstf could you guys agree on the approach as the tech-leaders, then I will make changes if needed. 

BTW one question on the side. How does the master-to-seed-sync part work when the master and seed are the same cluster? Is it just turned off or how is that solved?

zreigz:
cc @shaase-ctrl @scheeles 

I don't have expertise in this area, so we should follow they rules. For the tech lead meeting, I can rise the general topic about the whole project management process because we shouldn't implement controllers as a UI team. It's a lifecycle team area. Right now we lost two weeks of work instead of finishing endpoints and frontend.

Thanks, @moadqassem, and @xrstf  for your concerns and review

lsviben:
Thanks @zreigz, @moadqassem and @xrstf .

Ill change it to the seed version then. I dont think all is lost, I can reuse most of the code so it will go faster then from scratch. Ill add the master-to-seed controller CT controller separately

lsviben:
@moadqassem @xrstf i moved the controller the seed level. Check it out and let me know how it looks. 

The master -> seed CT sync controller will be done in a separate PR.

pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+func enqueueAllConstraintTemplates(client ctrlruntimeclient.Client, log *zap.SugaredLogger) *handler.EnqueueRequestsFromMapFunc {
+	return &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+		var requests []reconcile.Request
+
+		cluster, ok := a.Object.(*kubermaticv1.Cluster)
+		if !ok {
+			err := fmt.Errorf("Object was not a cluster but a %T", a.Object)
```
moadqassem: Error shouldn't be capitalised.

pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,265 @@
...
+				userClusterClient, err = r.userClusterClientProvider.GetClient(&userCluster)
+				if err != nil {
+					return fmt.Errorf("error getting client for cluster %s: %w", userCluster.Spec.HumanReadableName, err)
+				}
+				r.userClusterClients[userCluster.Name] = userClusterClient
+			}
+
+			err = action(userClusterClient, constraintTemplate)
+			if err != nil {
+				return fmt.Errorf("Failed syncing constraint template for cluster %s: %w", userCluster.Spec.HumanReadableName, err)
```
moadqassem: Error shouldn't be capitalised.

pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
```diff
@@ -0,0 +1,265 @@
...
+	if err := r.seedClient.List(r.ctx, clusterList, &ctrlruntimeclient.ListOptions{LabelSelector: r.workerNameLabelSelector}); err != nil {
+		return fmt.Errorf("failed listing clusters: %w", err)
+	}
+
+	for _, userCluster := range clusterList.Items {
+		if userCluster.Spec.Pause {
+			log.Debugw("Cluster paused, skipping", "cluster", userCluster.Spec.HumanReadableName)
+			continue
+		}
+
```
moadqassem, line 201: Can you please also add the worker-name check here: 
```go
if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != r.workerName { 
 // Don't Reconcile 
}
```

moadqassem:
@lsviben Thanks for moving the logic back to the seed level, nice work. The PR LGTM, @xrstf do you have another opinion on this?

xrstf (approved):
/approve

PR #5996, opened 2020-10-05
Allow to run conformance-tests locally again

xrstf:
This PR refactors the conformance-tests in a few significant ways, with the goal of allowing devs to run it locally.

* The usage of environment variables has been reduced greatly. They were mostly introduced to keep the CLI flagset backwards-compatible, but since we removed the old version of our upgrade tests, this is no longer required and can be cleaned up. Most of the env vars have been turned into CLI flags, except for those that are also used by the API E2E tests. These test run with `go test` and it's IMHO not worth the effeort to duplicate the CLI flags, so I kept e.g. `KUBERMATIC_OIDC_LOGIN`.

  * `SEED_NAME` became `-kubermatic-seed-cluster`

  * `NAMESPACE` became `-kubermatic-namespace`

  * `KUBERMATIC_HOST` and `KUBERMATIC_SCHEME` became `-kubermatic-endpoint=http://host:port`

  * `KUBERMATIC_PROJECT_ID` became `-kubermatic-project-id`

  * `KUBERMATIC_SERVICEACCOUNT_TOKEN` became `-kubermatic-oidc-token`

  * `KUBERMATIC_DEX_VALUES_FILE` has been removed (`-dex-helm-values-file` existed already)

  * `KUBERMATIC_PSP_ENABLED` became `-enable-psp`

  Env vars are of course still used within the `hack/ci/` scripts to transport metadata, especially out of the `setup-kubermatic-in-kind` scripts to their callers, as well as in Prow jobs to configure stuff. But they are then explicitly translated to CLI flags in `hack/ci/run-conformance-tests.sh`.

* The default versions to be tested in the conformance-tests is now automatically determined based on the Operator config. Likewise, the API E2E tests automatically determine the most recent k8s version to tests against. Of course, our Prow jobs still set the versions to test explicitly.
* The list of distributions to test can now also be specified via `-distributions`, simplifying the prow jobs greatly (`-distributions=ubuntu` is much easier to understand than `-exclude-distributions=coreos,flatcar,centos,ubuntu`). Both flags cannot be used at the same time.
* If no project ID is given, a project is always created automatically. Previously, it would only be created if the tester also performs the OIDC login itself.
* The API calls against the Kubermatic API now use the same auto-retrying HTTP client that the API E2E tests use. This allowed to remove some `wait.Poll()` calls throughout the tests. Also, cluster creation became easier since no cluster listing calls are needed after the `.Create()` call.
* To allow running the tests against a pre-existing Kubermatic setup, which is not the same version as the conformance-tests, the `ClusterReconciliationSuccessful` function can now be instructed to ignore the Kubermatic version when comparing the Cluster Conditions.
* The number of parallel tests has been reduced greatly, as 3 (nodes) times 10 = 30 processes à 150-200MB resident memory usage is not nice locally. It did not affect the ginkgo runtimes from my local experiments.

The tests themselves do not have a default for the Kubermatic API endpoint, but the hack script defaults it to our dev environment. This was made because the majority or callers will probably be Kubermatic devs for now. Of course, this can be overriden using `-kubermatic-endpoint=https://kubermatic.example.com`.

Many of these changes have been motivated by the rewritten `hack/run-conformance-tests.sh`. The script now uses Docker
to run the locally produced `_build/conformance-tests` binary. Running in Docker provides a number of advantages, but of course means that developers need to be careful when referring to files outside the repository path or using environment variables. As most things are now CLI flags, this is less of a concern thankfully :)

The Docker container uses the same user/group IDs as the local user, so generated reports are then now owned by root. The home directory is also mounted to allow adding the user's SSH key to the nodes (this was always there, but can now be disabled via `-node-ssh-pub-key=`).

The new script allows to add additional CLI flags, so it's easy enough to override defaults, e.g.

```bash
./hack/run-conformance-tests.sh -providers=gcp
```

will override the default `-providers=aws`. What I have been doing to test things is log into our dev system, copy the token and run the tests like so:

```bash
export OIDC_TOKEN=...
export KUBERMATIC_EDITION=ee

./hack/run-conformance-tests.sh \
  -kubermatic-oidc-token=$OIDC_TOKEN \
  -kubermatic-seed-cluster=europe-west3-c \
  -versions=v1.19.0 \
  -distributions=ubuntu
```

Note that in this case, the env variable is resolved by my shell and therefore does not need to exist in the (hidden) container.

A downside of the Docker approach is that it is harder to run the tests against a local Kubermatic setup (i.e. using the other hack scripts to bring up the Kubermatic). Making it easier to run against local environments (e.g. by using --net=host or allowing to skip the Docker container entirely) is left for future endavours.

One big remaining problem is that tests cannot be properly cancelled. The signal is caught, but since we don't yet transport the same context throughout our tests, it does nothing. Hitting Ctrl-C a second time then just kills the tests, leaving cloud resources left behind. But this PR is already way to large, so I will tackle this in a future PR.

```release-note
NONE
```


cmd/conformance-tests/main.go
```diff
@@ -164,58 +169,66 @@ const (
 var defaultTimeout = 10 * time.Minute
 
 var (
-	providers  string
-	pubKeyPath string
-	sversions  string
+	providers             string
+	pubKeyPath            string
+	sversions             string
+	sdistributions        string
```
moadqassem, line 169: why the 's' here as a prefix(also in the line below)? and if it has a meaning can you please add it in the usage section in the args messages. 
btw what about camel case style here?

cmd/conformance-tests/main.go
```diff
@@ -662,28 +656,92 @@ func createProject(client *apiclient.KubermaticAPI, bearerToken runtime.ClientAu
...
+	list := []string{}
+	for _, v := range minorMap {
+		list = append(list, "v"+v.String())
+	}
+	sort.Strings(list)
+
+	return list
+}
+
+func getEffectiveDistributions(distributions string, excludeDistributions string) (map[providerconfig.OperatingSystem]struct{}, error) {
```
moadqassem, line 695: What does the name of the method hint?

cmd/conformance-tests/main.go
```diff
@@ -662,28 +656,92 @@ func createProject(client *apiclient.KubermaticAPI, bearerToken runtime.ClientAu
...
+	if distributions != "" && excludeDistributions != "" {
+		return nil, errors.New("-distributions and -exclude-distributions must not be given at the same time")
+	}
+
+	chosen := map[providerconfig.OperatingSystem]struct{}{}
+
+	if distributions != "" {
+		for _, value := range strings.Split(distributions, ",") {
+			exists := false
+			for _, os := range all {
```
moadqassem: Please change the name of this var as it collides with the package `os`:
https://github.com/kubermatic/kubermatic/pull/5996/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&hide-deleted-files=true#diff-eedbfd4f9f8fd17c8ce5a7afbe479ddbR31

cmd/conformance-tests/main.go
```diff
@@ -662,28 +656,92 @@ func createProject(client *apiclient.KubermaticAPI, bearerToken runtime.ClientAu
...
 
-			return true, nil
-		}); err != nil {
-			return fmt.Errorf("failed creating SSH key: %v", err)
+			chosen[providerconfig.OperatingSystem(value)] = struct{}{}
+		}
+	} else {
+		excluded := strings.Split(excludeDistributions, ",")
+
+		for _, os := range all {
```
moadqassem: Please change the name of this var as it collides with the package `os`:
https://github.com/kubermatic/kubermatic/pull/5996/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md&file-filters%5B%5D=.sh&file-filters%5B%5D=.yaml&hide-deleted-files=true#diff-eedbfd4f9f8fd17c8ce5a7afbe479ddbR31

cmd/conformance-tests/runner.go
```diff
@@ -439,14 +446,16 @@ func (r *testRunner) executeTests(
...
+		defer logEventsForAllMachines(context.Background(), log, userClusterClient)
+		defer logUserClusterPodEventsAndLogs(
+			log,
+			r.clusterClientProvider,
+			cluster.DeepCopy(),
+		)
+	}
 
-	var overallTimeout = 10 * time.Minute
+	var overallTimeout = 20 * time.Minute
```
moadqassem: I think this also can be configurable as a flag.

cmd/conformance-tests/main.go
xrstf: That it combines two sets of restrictions to an effective set of distributions. What do you mean?

cmd/conformance-tests/main.go
xrstf: The s prefix is probably because these are the stringified versions of the CLI flags, that we have so that the Options struct has properly parsed, usable configuration fields.

> and if it has a meaning can you please add it in the usage section in the args messages.

What do you mean by that? Is

```go
flag.StringVar(&sdistributions, "distributions", "", "a comma-separated list of distributions to test (cannot be used in conjunction with -exclude-distributions)")
```

not enough?

moadqassem: > The s prefix is probably because these are the stringified versions of the CLI flags

No it is fine.

xrstf:
@moadqassem Feedback addressed.

PR #5997, opened 2020-10-05
Add seed-controller feature gate for etcd-launcher

### LINKED ISSUE #5992: Provide the ability to enable etcd-laucher by default in seed contorller
**User Story**
Currently, etcd-launcher is behind a cluster-level feature flag. We need to provide the ability to enable it by default for all cluster and selectively disable it per cluster.
**Acceptance criteria**
- [x] Add a seed controller flag to enable etcd-launcher for all clusters
- [x] etcd-launcher can be disabled per cluster.

moelsayed:
Adds a feature gate for etcd-launcher in seed-controller. If the flag is enabled all user clusters will have the cluster level feature flag automatically enabled, except user clusters where the cluster level feature flags is explicitly disabled.
Fixes #5992

**Documentation**:

```release-note
Adds a new feature gate to the seed-controller to enable etcd-launcher for all user clusters.
```

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/pending.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/pending.go
@@ -43,6 +43,11 @@ func (r *Reconciler) reconcileCluster(ctx context.Context, cluster *kubermaticv1
 		return nil, err
 	}
 
+	// Apply etcdLauncher flag
+	if err := r.ensureEtcdLauncherFeatureFlag(ctx, cluster); err != nil {
+		return nil, err
+	}
+
 	// Deploy & Update master components for Kubernetes
 	if err := r.ensureResourcesAreDeployed(ctx, cluster); err != nil {
 		return nil, err
@@ -110,6 +115,22 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 	return r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
 		for _, modify := range modifiers {
 			modify(c)
+		}
+	})
+}
+
+// ensureEtcdLauncherFeatureFlag will apply seed controller etcdLauncher setting on the cluster level
+func (r *Reconciler) ensureEtcdLauncherFeatureFlag(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	return r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
+		if r.features.EtcdLauncher { // enabled at the controller level
+			// we only modify the cluster feature flag if it's not explicitly set, regardless if the value
+			_, set := c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]
+			if !set {
+				if c.Spec.Features == nil {
+					c.Spec.Features = make(map[string]bool)
+				}
+				c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] = true
+			}
 		}
 	})
 }
--- a/pkg/controller/seed-controller-manager/kubernetes/pending.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/pending.go
@@ -123,9 +123,8 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 func (r *Reconciler) ensureEtcdLauncherFeatureFlag(ctx context.Context, cluster *kubermaticv1.Cluster) error {
 	return r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
 		if r.features.EtcdLauncher { // enabled at the controller level
-			// we only modify the cluster feature flag if it's not explicitly set, regardless if the value
-			_, set := c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]
-			if !set {
+			// we only modify the cluster feature flag if it's not explicitly set, regardless of the value
+			if _, set := c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]; !set {
 				if c.Spec.Features == nil {
 					c.Spec.Features = make(map[string]bool)
 				}
--- a/pkg/controller/seed-controller-manager/kubernetes/pending_test.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/pending_test.go
@@ -1,0 +1,139 @@
+// +build integration
+
+/*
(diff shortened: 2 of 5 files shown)


pkg/controller/seed-controller-manager/kubernetes/pending.go
```diff
@@ -113,3 +118,19 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
...
 	})
 }
+
+// ensureEtcdLauncherFeatureFlag will apply seed controller etcdLauncher setting on the cluster level
+func (r *Reconciler) ensureEtcdLauncherFeatureFlag(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	return r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
+		if r.features.EtcdLauncher { // enabled at the controller level
+			// we only modify the cluster feature flag if it's not explicitly set, regardless if the value
+			_, set := c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]
+			if !set {
```
irozzo-1A: personally, I find it more elegant if you get the `set` value and do the check it in one line:
```
if _, set := c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher]; set {
...
```

pkg/controller/seed-controller-manager/kubernetes/pending.go
```diff
@@ -113,3 +118,18 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 		}
 	})
 }
+
+// ensureEtcdLauncherFeatureFlag will apply seed controller etcdLauncher setting on the cluster level
+func (r *Reconciler) ensureEtcdLauncherFeatureFlag(ctx context.Context, cluster *kubermaticv1.Cluster) error {
```
irozzo-1A, line 123: I think this logic should be unit tested, it should not be too complicated

pkg/controller/seed-controller-manager/kubernetes/pending.go
```diff
@@ -43,6 +43,11 @@ func (r *Reconciler) reconcileCluster(ctx context.Context, cluster *kubermaticv1
 		return nil, err
 	}
 
+	// Apply etcdLauncher flag
+	if err := r.ensureEtcdLauncherFeatureFlag(ctx, cluster); err != nil {
```
irozzo-1A, line 47: I think this is defaulting, and should be done at admission phase. At the moment we don't have an admission webhook for `Cluster` resources, but I think this is the way to go long term. This is not a blocker as what you did is consistent with the approach used today.

pkg/controller/seed-controller-manager/kubernetes/pending.go
```diff
@@ -113,3 +118,18 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 		}
 	})
 }
+
+// ensureEtcdLauncherFeatureFlag will apply seed controller etcdLauncher setting on the cluster level
+func (r *Reconciler) ensureEtcdLauncherFeatureFlag(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	return r.updateCluster(ctx, cluster, func(c *kubermaticv1.Cluster) {
+		if r.features.EtcdLauncher { // enabled at the controller level
+			// we only modify the cluster feature flag if it's not explicitly set, regardless if the value
```
irozzo-1A: should it be `regardless of value` instead of `regardless if the value`?

pkg/features/features.go
```diff
@@ -41,6 +41,10 @@ const (
 	// --experimental-initial-corrupt-check=true +
 	// --experimental-corrupt-check-time=10m
 	EtcdDataCorruptionChecks = "EtcdDataCorruptionChecks"
+
+	// EtcdLauncher if enabled will apply the cluster level etcd-launcher feature flag on all clusters,
+	// unless it's explicitly disabled on the cluster level
```
irozzo-1A: `s/on the cluster level/at cluster level/` ?

irozzo-1A:
@moelsayed I finished my review, overall looks good. I just left a few comments, and I would appreciate some unit-test coverage.

pkg/controller/seed-controller-manager/kubernetes/pending.go
moelsayed: I agree.

pkg/controller/seed-controller-manager/kubernetes/pending_test.go
```diff
@@ -0,0 +1,139 @@
...
+			clusterFeatures: map[string]bool{
+				kubermaticv1.ClusterFeatureEtcdLauncher: true,
+			},
+			seedEtcdLauncher:     false,
+			expectedEtcdLauncher: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
irozzo-1A, line 61: I think the test can be simplified a bit. I don't think for the scope of this test you need starting a the controller-runtime manager or getting the client.

I tested this locally and seems to be enough:
```
		t.Run(test.name, func(t *testing.T) {
			cluster := &kubermaticv1.Cluster{
					Name: "test-cluster",
				},
				Spec: kubermaticv1.ClusterSpec{
					Features: test.clusterFeatures,
				},
			}
			r := &Reconciler{
				features: Features{
					EtcdLauncher: test.seedEtcdLauncher,
				},
			}
			if err := r.ensureEtcdLauncherFeatureFlag(context.Background(), cluster); err != nil {
				t.Fatal(err)
			}
			if cluster.Spec.Features != nil && cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] != test.expectedEtcdLauncher {
				t.Errorf("expected clsuter flag to be %v , got %v instead", test.expectedEtcdLauncher, cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher])
			}
		})
```

irozzo-1A: Well, I simplified too much, the client is needed for the Patch call, but as we discussed offline you can make use of `"sigs.k8s.io/controller-runtime/pkg/client/fake"`

pkg/controller/seed-controller-manager/kubernetes/pending_test.go
```diff
@@ -0,0 +1,139 @@
+// +build integration
```
irozzo-1A: Also, this should be removed as we discussed as this is a unit test ;-)

PR #5986, opened 2020-10-02
[release/v2.14] Migrate from google_containers to k8s.gcr.io Docker registry

xmudrii:
**What this PR does / why we need it**:

Migrate from `gcr.io/google_containers` registry to `k8s.gcr.io` registry. The google_containers registry is read-only for several weeks, and therefore the migration is advised.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
xref #5773 #5967
The issue will be closed once we fix it for the release/v2.13 branch as well.

**Does this PR introduce a user-facing change?**:
```release-note
Migrate from google_containers to k8s.gcr.io Docker registry
```

PR #5758, opened 2020-08-13
Hyperkube image replaced with k8s.gcr.io respective images

kron4eg:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
xRef #5717

**Special notes for your reviewer**:
OpenShift usage of hypekube will be in separate PR

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Hyperkube image replaced with k8s.gcr.io respective images
```

PR #6012, opened 2020-10-09
[WIP] Set maxSurge to zero in apiserver deployment with more than one replica

irozzo-1A:
The idea behind this PR is to limit the consumption of additional resources (memory in particular) in seed clusters during the rollout of new k8c versions triggering the rollout of control plane components. e.g. #5986 #5758 

Configuring the apiserver deployment rolling strategy with maxSurge=0 and maxUnavailable=1 (default is the opposite) should guarantee that the replicaset associated with the old version is scaled down before the one associated with the new version is scaled up.

Fixes #

**Documentation**:

```release-note
```

### DIFF
--- a/pkg/resources/apiserver/deployment.go
+++ b/pkg/resources/apiserver/deployment.go
@@ -82,11 +82,19 @@ func DeploymentCreator(data *resources.TemplateData, enableOIDCAuthentication bo
 				dep.Spec.Replicas = data.Cluster().Spec.ComponentsOverride.Apiserver.Replicas
 			}
 
+			if *dep.Spec.Replicas > 1 {
+				// Avoid creating additional Apiserver PODs when rolling up a new
+				// version.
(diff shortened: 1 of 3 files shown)


irozzo-1A:
During my tests, I realized that this measure is not very effective in limiting additional resources consumption during the rollout of new apiserver pods.

The reason is that setting `maxSurge` to 0 and `maxUnavailable` to 1 ensures that during the rollout from version `n` to `n+1` the actual replicas of the replicaset associated to version `n` are decreased by 1 before the replicas of version `n+1` are increased by one. When a pod is in `Terminating` state it is not accounted any longer by the replicaset `n` enabling the scaleup of the replicaset `n+1`, but still consumes resources. If the time elapsed from the pod being terminating to the actual termination is greater than the time it takes for the new pod to start there is still an additional resource consumption. And from my observation this isn't unlikely to happen, and it's arriving often in my test environment.

PR #6017, opened 2020-10-12
Don't deploy cluster-autoscaler on unsupported clusters

xmudrii:

Don't deploy cluster-autoscaler on unsupported clusters.

```release-note
NONE
```

### DIFF
--- a/addons/cluster-autoscaler/cluster-autoscaler.yaml
+++ b/addons/cluster-autoscaler/cluster-autoscaler.yaml
@@ -20,6 +20,7 @@
 {{ $version = "v1.19.0 "}}
 {{ end }}
 
+{{ if not eq $version "UNSUPPORTED" }}
 ---
 apiVersion: v1
 kind: ServiceAccount
@@ -193,3 +194,4 @@ spec:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+{{ end }}


xrstf:
Are we sure that this does not fail? Even though we do a `len(manifests) == 0` check before kubectl-applying it, the license comments in the manifest files should make this check never pass anymore. Maybe we should actually parse the YAML into a list of some raw, unstructured objects and see if that list is empty.

xmudrii:
@xrstf I'll test this tomorrow and get back to you. I think we already have some YAML parsing logic in place, which might be useful here, but I have to check.

xmudrii:
@xrstf I went through the code and here's what I've found:

- the kubectl seems to be invoked by the [`ensureIsInstalled` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L485-L514)
- at the beginning, that function invokes the [`setupManifestInteraction` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L486-L490)
- the [`setupManifestInteraction` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L450-L477) has several tasks: parse the manifests, ensure there are appropriate labels on each object in the manifest, combine them into a single manifest, write the combined manifest along with the kubeconfig
- the `setupManifestInteraction` function invokes the [`getAddonManifests` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L296-L352), which prepares and parses the addon manifest (by running it through the templating logic)
- the `getAddonManifests` function invokes the [`addonutils.ParseFromFolder` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L346-L349) to parse the manifests and returns parsed manifests back to the `setupManifestInteraction` function
- the [`addonutils.ParseFromFolder` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/addon/template.go#L183-L251) reads all files, decodes them, and puts them in the manifests array
- mostly important, the `addonutils.ParseFromFolder` function [checks is the manifest empty after decoding it](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/addon/template.go#L241-L245), for example, if there are only comments in the file
- when `addonutils.ParseFromFolder` and `getAddonManifests` functions return the parsed addons to the `getAddonManifests` function, it ensures that there are the appropriate labels applied and then writes the manifest to a temporary location
- [the temporary path is returned to the `ensureIsInstalled` function](https://github.com/kubermatic/kubermatic/blob/ad6ea896e4fe812e709a1748ee793049ab46ae04/pkg/controller/seed-controller-manager/addon/addon_controller.go#L486-L500)
- the `ensureIsInstalled` function reads the file, ensures that it's not empty, and if it's empty, it just returns from the function and logs that the manifest was not applied

Based on that, if we just send a manifest that only has the copyright header, the `addonutils.ParseFromFolder` function should return empty manifest, as it decodes YAML, and then parses it. At the end, the parsed manifest is written to a temporary location, and the path is given to `kubectl` to apply the manifest.

xrstf (approved):
/approve

PR #6018, opened 2020-10-12
[WIP][release/v2.14] Tolerate legacy registry for k8s control plane components

irozzo-1A:
The goal of this PR is to avoid rolling out all control plane components in seed clusters when k8c patch version is upgraded due to #5986.

Fixes #

**Documentation**:

```release-note
NONE
```


irozzo-1A:
Unfortunately the apiserver POD (that is the main memory consumer) is always rolled-out when new `seed-controller-manager` is deployed, because of the `dnat-controller` image uses the TAG corresponds to the release commit.

irozzo-1A:
I will probably backport some of the logic to master, but the gain for v2.14 is very small.

PR #6027, opened 2020-10-13
Split kind/Kubermatic logic during CI

xrstf:
This PR is a first step towards bringing back the upgrade tests (KKP upgrades that is, not usercluster upgrades). These tests were removed a year back or so because the way we built them was severely limiting our ability to improve the product and lead to tons of BC hacks all over the place.

Now that we have a fancy installer and operator, it's time to bring these tests back and make sure that the installer can do what it promises. This will then replace the canary-ci-deployment tests, which currently test that our Helm charts are upgradable without problems. The problem with the canary job are many however: there can be only 1 job at any time, it depends on Kubermatic infrastructure and it does not test anyhting that is actually of importance.

The general test outline I aim for is this:

1. Setup a kind cluster.
2. Find the last stable KKP release, `git checkout` to that revision and then install KKP into the kind cluster. This of course requires that the release version also already has the installer (because otherwise we couldn't test the installer, as all our other code uses Helm 2, but the installer uses Helm 3). Instead of git-checkout and compiling, we could soon also just download the release archive from github. :-D
3. Wait for the old KKP installation to be complete.
4. Checkout HEAD of the Pull Request again, recompile installer.
5. Run the new installer and let it upgrade the KKP setup.
6. Wait for all Deployments to have finished rolling out.

This PR is a preparation for step 2, to ensure that we have a simple script that can just setup Kubermatic without also setting up Kind/clusterexposer. That's why this PR splits the `setup-kubermatic-in-kind` script into two, one just for kind and one just for KKP. To make it simpler, the large inline YAML documents for Seed/KubermaticConfiguration have been extracted as well.

```release-note
NONE
```

### DIFF
--- a/hack/ci/run-api-e2e.sh
+++ b/hack/ci/run-api-e2e.sh
@@ -23,6 +23,16 @@ set -euo pipefail
 cd $(dirname $0)/../..
 source hack/lib.sh
 
+TEST_NAME="Pre-warm Go build cache"
+echodate "Attempting to pre-warm Go build cache"
+
+beforeGocache=$(nowms)
+make download-gocache
+pushElapsed gocache_download_duration_milliseconds $beforeGocache
+
+export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
+
+source hack/ci/setup-kind-cluster.sh
 source hack/ci/setup-kubermatic-in-kind.sh
 
 echodate "Creating UI Azure preset..."
@@ -123,7 +133,6 @@ appendTrap print_kubermatic_logs EXIT
 
 echodate "Running API E2E tests..."
 
-export KUBERMATIC_DEX_VALUES_FILE=$(realpath hack/ci/testdata/oauth_values.yaml)
 go test -tags="create,$KUBERMATIC_EDITION" -timeout 20m ./pkg/test/e2e/api -v
 go test -tags="e2e,$KUBERMATIC_EDITION" -timeout 20m ./pkg/test/e2e/api -v
 go test -tags="logout,$KUBERMATIC_EDITION" -timeout 20m ./pkg/test/e2e/api -v
--- a/hack/ci/run-e2e-tests.sh
+++ b/hack/ci/run-e2e-tests.sh
@@ -26,6 +26,17 @@ source hack/lib.sh
 export GIT_HEAD_HASH="$(git rev-parse HEAD)"
 export KUBERMATIC_VERSION="${GIT_HEAD_HASH}"
 
+TEST_NAME="Pre-warm Go build cache"
+echodate "Attempting to pre-warm Go build cache"
+
+beforeGocache=$(nowms)
+make download-gocache
+pushElapsed gocache_download_duration_milliseconds $beforeGocache
+
+echodate "Creating kind cluster"
+export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
+source hack/ci/setup-kind-cluster.sh
+
 echodate "Setting up Kubermatic in kind on revision ${KUBERMATIC_VERSION}"
 
 beforeKubermaticSetup=$(nowms)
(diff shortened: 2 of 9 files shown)


hack/ci/run-api-e2e.sh
```diff
@@ -123,7 +133,6 @@ appendTrap print_kubermatic_logs EXIT
 
 echodate "Running API E2E tests..."
 
-export KUBERMATIC_DEX_VALUES_FILE=$(realpath hack/ci/testdata/oauth_values.yaml)
```
xrstf, line 126: This was just redundant.

hack/ci/run-conformance-tester.sh
```diff
@@ -27,11 +27,6 @@ export VERSIONS=${VERSIONS_TO_TEST:-"v1.18.8"}
 export EXCLUDE_DISTRIBUTIONS="${EXCLUDE_DISTRIBUTIONS:-ubuntu,centos,sles,rhel}"
 export ONLY_TEST_CREATION=${ONLY_TEST_CREATION:-false}
 
-export WORKER_NAME="${BUILD_ID}"
-if [ "${KUBERMATIC_NO_WORKER_NAME:-}" = "true" ]; then
-  WORKER_NAME=""
-fi
-
```
xrstf: Since we test in kind, there are never worker-names involved. This is a leftover from when we tested in a shared master/seed cluster.

hack/ci/setup-kubermatic-in-kind.sh
```diff
@@ -58,126 +53,19 @@ export KUBERMATIC_OIDC_PASSWORD="password"
...
-
-iptables -t nat -A PREROUTING -i eth0 -p tcp -m multiport --dports=30000:33000 -j DNAT --to-destination 172.18.0.2
-# By default all traffic gets dropped unless specified (tested with docker
-# server 18.09.1)
-iptables -t filter -I DOCKER-USER -d 172.18.0.2/32 ! -i $KIND_NETWORK_IF -o $KIND_NETWORK_IF -p tcp -m multiport --dports=30000:33000 -j ACCEPT
-
-# Docker sets up a MASQUERADE rule for postrouting, so nothing to do for us
-echodate "Successfully set up iptables rules for nodeports"
+# find Dex anymore. As this script can be run multiple times in the same CI job,
+# we must make sure to only add the alias once.
```
xrstf, line 58: Currently this is not true, the script is not run multiple times. But it soon will be.

hack/ci/run-e2e-tests.sh
```diff
@@ -36,7 +47,5 @@ else
 fi
 pushElapsed kind_kubermatic_setup_duration_milliseconds $beforeKubermaticSetup
 
-echodate "Done setting up Kubermatic in kind"
-
 echodate "Running conformance tests"
-KUBERMATIC_NO_WORKER_NAME=true ./hack/ci/run-conformance-tests.sh
```
moelsayed, line 42: Is there somewhere else where this was set to false?

hack/ci/setup-kind-cluster.sh
```diff
@@ -0,0 +1,133 @@
...
+
+echodate "Setting up kind cluster..."
+
+if [ -z "${JOB_NAME:-}" ]; then
+  echodate "This script should only be running in a CI environment."
+  exit 1
+fi
+
+if [ -z "${PROW_JOB_ID:-}" ]; then
+  echodate "Build id env variable has to be set."
```
moelsayed: nit: mention the env variable name in the message.

hack/ci/setup-kind-cluster.sh
```diff
@@ -0,0 +1,133 @@
...
+iptables -t mangle -A POSTROUTING -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --clamp-mss-to-pmtu
+
+# Make debugging a bit better
+echodate "Confuguring bash"
+cat <<EOF >>~/.bashrc
+# Gets set to the CI clusters kubeconfig from a preset
+unset KUBECONFIG
+
+cn() {
+  kubectl config set-context \$(kubectl config current-context) --namespace=\$1
```
moelsayed: nit: use `--current` to select the current context.

moelsayed (commented):
Couple of nit comments, otherwise lgtm. Thank you.

hack/ci/run-e2e-tests.sh
xrstf: No, it was only used in the one if statement that I removed in this PR as well.

PR #6043, opened 2020-10-18
fix tag-check in github release script

xrstf:
Previously, we only checked the first N tags, so the logic often seemingly randomly failed. This PR improves that by just checking the tag directly. Nicely enough, the Github API does _not_ perform a prefix check, so checking for `v2.15.0` will not match an existing `v2.15.0-rc.1`.

```release-note
NONE
```


xmudrii (approved):
/lgtm

PR #6044, opened 2020-10-18
[release/v2.15] fix tag-check in github release script

xrstf:
This is a manual backport of #6043. It is somewhat improved though, as it removes an unneeded empty-string check and makes the logic more obvious.

```release-note
NONE
```


xmudrii (approved):
/lgtm

PR #6051, opened 2020-10-19
Fix bug where cluster in deletion kills controller

phiphi282:
Sometimes a cluster in deletion can kill the controller possibly
indefinitely, when it tries to add a finalizer to them.
This happens for example when a cluster is in deletion and the
SecretsCleanupFinalizer gets removed when the controller-manager
restarts for some reason.
Then it will always try to add this finalizer again and fail.

Make sure that finalizers only get added when objects are not stuck in
deletion.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/api/pkg/crd/migrations/seed/migrations.go
+++ b/api/pkg/crd/migrations/seed/migrations.go
@@ -225,6 +225,10 @@ func migrateClusterUserLabel(cluster *kubermaticv1.Cluster, cleanupContext *clea
 }
 
 func createSecretsForCredentials(cluster *kubermaticv1.Cluster, cleanupContext *cleanupContext) error {
+	if cluster.GetDeletionTimestamp() != nil {
+		return nil
+	}
+
 	if err := kubernetesprovider.CreateOrUpdateCredentialSecretForCluster(cleanupContext.ctx, cleanupContext.client, cluster); err != nil {
 		return err
 	}


api/pkg/crd/migrations/seed/migrations.go
```diff
@@ -228,7 +228,9 @@ func createSecretsForCredentials(cluster *kubermaticv1.Cluster, cleanupContext *
 	if err := kubernetesprovider.CreateOrUpdateCredentialSecretForCluster(cleanupContext.ctx, cleanupContext.client, cluster); err != nil {
 		return err
 	}
-	kuberneteshelper.AddFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer)
+	if cluster.GetDeletionTimestamp() == nil {
+		kuberneteshelper.AddFinalizer(cluster, apiv1.CredentialsSecretsCleanupFinalizer)
+	}
```
xrstf: If I read this correctly, this will not add the finalizer when the cluster is already in deletion. That make sense.

But in line 228, right before this, we create the credentials Secrets regardless of the deletion status. Won't these secrets then be orphans once the cluster is gone?

xrstf (commented):
Some questions, otherwise nice catch :-)

api/pkg/crd/migrations/seed/migrations.go
phiphi282: Ah yes, changed it to just return at the beginning of the function if there is a deletionTimestamp.
Make more sense it is even better readable this way :D

xrstf (approved):
/approve
/hold

Holding until the release-blocking PRs are merged.

PR #6059, opened 2020-10-20
Fix cluster-autoscaler addon template error and addons templating tests

xmudrii:

* Fix cluster-autoscaler addon template error
* Fix addons templating tests

```release-note
NONE
```

/assign @xrstf


xrstf (approved):
/approve

PR #6060, opened 2020-10-20
Add explicit context handling in conformance-tests

xrstf:
Currently our conformance-tests do not use a context consistently, making it hard to Ctrl-C it and have it properly clean up after itself. This PR improves the situation by making the context handling explicit.

```release-note
NONE
```

### DIFF
--- a/cmd/conformance-tests/scenarios_vsphere.go
+++ b/cmd/conformance-tests/scenarios_vsphere.go
@@ -119,7 +119,7 @@ func (s *vSphereScenario) Cluster(secrets secrets) *apimodels.CreateClusterSpec 
 	return spec
 }
 
-func (s *vSphereScenario) NodeDeployments(ctx context.Context, num int, _ secrets) ([]apimodels.NodeDeployment, error) {
+func (s *vSphereScenario) NodeDeployments(_ context.Context, num int, _ secrets) ([]apimodels.NodeDeployment, error) {
 	osName := getOSNameFromSpec(s.nodeOsSpec)
 	replicas := int32(num)
 	return []apimodels.NodeDeployment{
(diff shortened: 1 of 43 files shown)


cmd/conformance-tests/scenarios_vsphere.go
```diff
@@ -118,7 +119,7 @@ func (s *vSphereScenario) Cluster(secrets secrets) *apimodels.CreateClusterSpec
 	return spec
 }
 
-func (s *vSphereScenario) NodeDeployments(num int, _ secrets) ([]apimodels.NodeDeployment, error) {
+func (s *vSphereScenario) NodeDeployments(ctx context.Context, num int, _ secrets) ([]apimodels.NodeDeployment, error) {
```
maciaszczykm: Is it needed here?

xrstf: Yes, because some cloud providers need a context, so the interface for scenarios is

```
type testScenario interface {
	Name() string
	Cluster(secrets secrets) *apimodels.CreateClusterSpec
	NodeDeployments(ctx context.Context, num int, secrets secrets) ([]apimodels.NodeDeployment, error)
	OS() apimodels.OperatingSystemSpec
}
```

AWS needs this.

xrstf: Or do you insist on calling the parameter `_` here to signify that it is not used for this scenario? ^^

maciaszczykm: I'd rather name it `_` if it is not used :)

xrstf: Updated.

PR #6063, opened 2020-10-20
Support KubermaticConfiguration in image-loader

xrstf:
The image-loader did not yet help users that do not manage the versions with the old YAML file, but instead use the operator and its KubermaticConfiguration. This PR rectifies that by adding support for it. I chose not to split the image loader into CE/EE binaries just because it's so little code and simply not worth it for the time we still support the old Helm chart.

This can be tested by building the image loader (make sure to set a KUBERMATICCOMMIT to a version that has Docker images available on quay), creating an empty KubermaticConfiguration file and the running:

```
export KUBERMATICCOMMIT=4818b3906e63588c0b75a5a07f8d1a9bc95d8756
make image-loader

docker run -d -p 5000:5000 --restart=always --name registry registry:2

_build/image-loader \
  -registry localhost:5000 \
  -configuration-file test-kubermaticconfig.yaml \
  -log-debug
```

```release-note
Add support for KubermaticConfiguration in image-loader utility
```

### DIFF
--- a/cmd/image-loader/main.go
+++ b/cmd/image-loader/main.go
@@ -85,7 +85,7 @@ func main() {
 
 	o := opts{}
 	flag.StringVar(&o.configurationFile, "configuration-file", "", "Path to the KubermaticConfiguration YAML file")
-	flag.StringVar(&o.versionsFile, "versions-file", "", "The versions.yaml file path (deprecated, EE-only)")
+	flag.StringVar(&o.versionsFile, "versions-file", "", "The versions.yaml file path (deprecated, EE-only, used only if no -configuration-file is given)")
 	flag.StringVar(&o.versionFilter, "version-filter", "", "Version constraint which can be used to filter for specific versions")
 	flag.StringVar(&o.registry, "registry", "", "Address of the registry to push to, for example localhost:5000")
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
@@ -99,6 +99,10 @@ func main() {
 			fmt.Println(err)
 		}
 	}()
+
+	if (o.configurationFile == "") != (o.versionsFile == "") {
+		log.Fatal("Either -configuration-file or -versions-file must be specified.")
+	}
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
--- a/cmd/image-loader/main.go
+++ b/cmd/image-loader/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -25,13 +26,16 @@ import (
 
 	"github.com/Masterminds/semver"
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
 
 	addonutil "k8c.io/kubermatic/v2/pkg/addon"
 	apiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
 	kubernetescontroller "k8c.io/kubermatic/v2/pkg/controller/seed-controller-manager/kubernetes"
 	"k8c.io/kubermatic/v2/pkg/controller/seed-controller-manager/monitoring"
 	containerlinux "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/container-linux"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
 	"k8c.io/kubermatic/v2/pkg/docker"
 	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
 	"k8c.io/kubermatic/v2/pkg/resources"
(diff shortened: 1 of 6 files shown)


cmd/image-loader/main.go
```diff
@@ -65,28 +68,32 @@ var (
...
-	klog.InitFlags(nil)
+	var err error
 
 	logOpts := kubermaticlog.NewDefaultOptions()
+	logOpts.Format = kubermaticlog.FormatConsole
 	logOpts.AddFlags(flag.CommandLine)
 
 	o := opts{}
-	flag.StringVar(&o.versionsFile, "versions", "charts/kubermatic/static/master/versions.yaml", "The versions.yaml file path")
+	flag.StringVar(&o.configurationFile, "configuration-file", "", "Path to the KubermaticConfiguration YAML file")
```
irozzo-1A, line 87: I would mention that the two flags are mutually (i.e. `configuration-file` and `versions-file`) exclusive and I would add some input validation. If both are specified or none of them I would fail immediately.

cmd/image-loader/main.go
```diff
@@ -65,28 +68,32 @@ var (
...
+	flag.StringVar(&o.versionsFile, "versions-file", "", "The versions.yaml file path (deprecated, EE-only)")
 	flag.StringVar(&o.versionFilter, "version-filter", "", "Version constraint which can be used to filter for specific versions")
-	flag.StringVar(&o.registry, "registry", "registry.corp.local", "Address of the registry to push to")
+	flag.StringVar(&o.registry, "registry", "", "Address of the registry to push to, for example localhost:5000")
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
 	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to the folder containing the addons")
```
irozzo-1A, line 92: This is not related to your changes, but would not make more sense to get the addons from the addons image related to the current version?

xrstf: This is precisely what the next PR will do. It's already implemented, but based on this branch.

cmd/image-loader/main.go
xrstf: Done.

cmd/image-loader/main.go
```diff
@@ -65,34 +68,42 @@ var (
...
-	log := kubermaticlog.New(logOpts.Debug, logOpts.Format)
+	rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
+	log := rawLog.Sugar()
 	defer func() {
 		if err := log.Sync(); err != nil {
 			fmt.Println(err)
 		}
 	}()
 
+	if (o.configurationFile == "") != (o.versionsFile == "") {
```
irozzo-1A: This is an XOR, it's not at all what we need, it fails if one of the two is provided. Anyway, after better thinking, we should probably allow when both are not specified as we have default versions. We should just fail when both are specified:
```
if o.configurationFile != "" && o.versionsFile != "" {
```

xrstf: I disagree. I specifically required one or the other so that users who forget to specify it don't accidentally get the wrong versions. Users of the old Helm chart _must_ give us the versions.yaml, they must not rely on the baked-in versions in the image-loader.

xrstf: That's not to say that this new check is maybe buggy and does represent the logic I wanted to have (and had, before adding this additional check).

irozzo-1A: Do you prefer something like `-configuration-file /dev/null`? I think that the tool would be easier to use if when versions are not overridden both arguments can be left empty. Anyway, do as you prefer. Just fix the logic ;-)

xrstf: Yes, I prefer to have this in our CI to having a bad CLI for everyone. I agree that it's super akward, but also super clever and I was even surprised that io.ReadFile(/dev/null) works and doesn't block infinitely.

The alternative would have been to create a dummy, empty config file and specify that in our CI scripts. That wasn't worth it when /dev/null works so nicely. ^^

Once the old Helm chart is gone and we don't need to support the versions.yaml anymore, _then_ we can make the config optional and use the inbuilt one. For now having an explicit config made the logic inside more obvious.

cmd/image-loader/main.go
```diff
@@ -65,34 +68,42 @@ var (
...
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
 	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to the folder containing the addons")
 	flag.Parse()
 
-	log := kubermaticlog.New(logOpts.Debug, logOpts.Format)
+	rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
+	log := rawLog.Sugar()
 	defer func() {
 		if err := log.Sync(); err != nil {
 			fmt.Println(err)
```
irozzo-1A, line 99: Locally (macos) I'm getting this output when the process terminates:
```
sync /dev/stderr: inappropriate ioctl for device
```
I think we can safely avoid logging `Sync` errors and do as mentioned in the quickstart: https://github.com/uber-go/zap#quick-start

```
rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
defer rawLog.Sync()
log := rawLog.Sugar()
```

xrstf: This should then affect all binaries we have, no? This is not the PR to touch most of our binaries ;-)

cmd/image-loader/main.go
irozzo-1A: > Yes, I prefer to have this in our CI to having a bad CLI for everyone

I think your definition of "bad CLI" is largely debatable, but as I said, do as you like.

cmd/image-loader/main.go
irozzo-1A: If we do the same probably affects the other binaries too. But this seems to be OS specific and only CLIs are supposed to run on macos.

irozzo-1A: Anyway, if you prefer to handle this separately it's not a problem as this issues was not generated by your changes.

PR #6064, opened 2020-10-21
Improve initial node deployment creation

### LINKED ISSUE #2928: Improve initial node deployment creation

maciaszczykm:
**What this PR does / why we need it**: Currently the whole cluster and initial node deployment creation process happens in the API. First we create the cluster, then start the background process for initial node deployment creation and return the created cluster to the user. The main issue with this approach is that if the API will crash or if it will restart during the background process run that handles initial node deployment creation, then the data will be lost and the cluster will have no initial deployment added to it. This pull request changes the approach and adds persistence layer to the process.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: Fixes https://github.com/kubermatic/kubermatic/issues/2928.

- The initial plan is to use cluster annotation (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) as persistence layer. CRDs are the alternative but annotations seem to be just enough to store the machine deployment definition until the cluster will be ready and we will be able to create machine deployment object.

**Documentation**:

```release-note
Improved initial node deployment creation process. Right now cluster annotation is used to save the node deployment object before it is created to improve stability.
```


maciaszczykm:
@multi-io Hello, you have mentioned that you were starting to do similar thing. Can we sync on this as this is the thing that I would like to push forward for us too?

pkg/controller/seed-controller-manager/update/update_controller.go
```diff
@@ -116,16 +117,36 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+				r.recorder.Eventf(cluster, corev1.EventTypeNormal, "InitialMachineDeployment", "Created initial machine deployment")
+				if err := r.removeInitialMachineDeploymentRequestAnnotation(cluster); err != nil {
+					return nil, fmt.Errorf("cannot remove initial machine deployment request from %s cluster: %v", cluster.Name, err)
+				}
+			}
+		} else {
+			if err := r.removeInitialMachineDeploymentRequestAnnotation(cluster); err != nil {
+				return nil, fmt.Errorf("cannot remove initial machine deployment request from %s cluster: %v", cluster.Name, err)
+			}
+		}
```
maciaszczykm: @xrstf @moadqassem can you take a look on the above code and tell me if this is be okay for you? I will implement methods soon but I would like to confirm that it is okay for you if the approach will be like this.

pkg/controller/seed-controller-manager/update/update_controller.go
```diff
@@ -116,16 +117,36 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-	if cluster.IsOpenshift() {
-		clusterType = v1.OpenShiftClusterType
+	clusterType := getClusterType(cluster)
+
+	// Check if cluster has non-empty initial machine deployment request serialized
+	// into its annotations and create it if its there.
+	if request, ok := cluster.Annotations[v1.InitialMachineDeploymentRequestAnnotation]; ok && request != "" {
+		var nodeDeployment *v1.NodeDeployment
+		if err := json.Unmarshal([]byte(request), &nodeDeployment); err != nil {
+			return nil, fmt.Errorf("cannot unmarshal initial machine deployment request for %s cluster: %v", cluster.Name, err)
```
maciaszczykm: Should this be changed in order to avoid throwing an error? For example sending some event and then possibly removing the annotation?

pkg/controller/seed-controller-manager/update/update_controller.go
```diff
@@ -116,16 +117,36 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 		return nil, nil
 	}
 
-	clusterType := v1.KubernetesClusterType
-	if cluster.IsOpenshift() {
-		clusterType = v1.OpenShiftClusterType
+	clusterType := getClusterType(cluster)
+
+	// Check if cluster has non-empty initial machine deployment request serialized
+	// into its annotations and create it if its there.
```
xrstf: _it's_

pkg/controller/seed-controller-manager/update/update_controller.go
xrstf: What made you choose this controller to add your logic to? Because it deals with nodes during the cluster upgrades? What would you say to writing a dedicated controller just for this purpose (turning the initial-node annotation into a MachineDeployment)?

maciaszczykm: I thought that this was our agreement during discussion on Slack as you agreed on "re-using an existing controllers" and @moadqassem pointed this one. It's good that the code is not finished yet and there won't be much to revert, so no worries here. However, at this point I'd rather leave it to lifecycle team as only the controller part is missing. I think it will be quicker for you to do and we will avoid any unnecessary misunderstandings.

moadqassem: @xrstf What is the problem of adding the initial node deployment here too. This controller is not just responsible for taking care of control plane components, it is also responsible for managing the machine deployment list within the user cluster. Also there is no real semantics here that this controller is only for upgrades, as it watches seed resources(cluster) and in the very same time it has the reconciled user cluster clients. Anyway, would it really make sense to create a controller only to add an initial deployment for a cluster? I mean if this place is not appropriate as what @xrstf suggested I am quite sure that we can find another controller that we can fit this logic in.

xrstf: Yeah it's fine.

pkg/controller/seed-controller-manager/update/update_controller.go
xrstf: For now I think this is sufficient.

pkg/controller/seed-controller-manager/update/update_controller.go
xrstf: So I think we're good here, @moadqassem WDYT?

moadqassem: SGTM

xrstf:
/lgtm cancel

Sorry, had a brainfart. Just wanted to signal that this can be developed further in its current form.

maciaszczykm:
/hold cancel

Transformed it to draft. I will continue with the implementation. Thanks for approval @xrstf.

xrstf:
Since Marcin is OOO, I am taking over.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
```diff
@@ -0,0 +1,243 @@
...
+
+	var nodeDeployment *v1.NodeDeployment
+	if err := json.Unmarshal([]byte(request), &nodeDeployment); err != nil {
+		// TODO: What now? Remove the invalid annotation? This will end up in a reconcile loop.
+		return nil, fmt.Errorf("cannot unmarshal initial MachineDeployment request: %v", err)
+	}
+
+	nodeDeployment, err := machineresource.Validate(nodeDeployment, cluster.Spec.Version.Semver())
+	if err != nil {
+		// TODO: What now? Remove the invalid annotation? This will end up in a reconcile loop.
```
maciaszczykm: I think that removing the annotation in both these cases makes sense since there is really no obvious way to fix these issues here. I would just leave some information in the event that creation has failed and remove the annotation.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, fmt.Errorf("initial node deployment is not valid: %v", err)
+	}
+
+	userClusterClient, err := r.userClusterConnectionProvider.GetClient(cluster)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get user cluster client: %v", err)
+	}
+
+	if err := r.createInitialMachineDeployment(nodeDeployment, cluster, userClusterClient); err != nil {
```
maciaszczykm, line 154: It seems that you have removed the check for machine deployment existence here (previously in the `isInitialMachineDeploymentCreated` method). It was added to cover case in which initial machine deployment is created but then the annotation cannot be removed due some reason. If it happens now, we will have duplicated machine deployments or error (depends if the name was set in the JSON). It is not common thing to happen but perhaps we should cover it too, what do you think? However,`InitialMachineDeploymentLabel` label is additional side effect in my solution.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	err = client.Create(r.ctx, machineDeployment)
+	if err != nil {
+		// in case we created the MD before but then failed to cleanup the Cluster resource's
+		// annotations, we can silently ignore AlreadyExists errors here and then re-try removing
+		// the annotation afterwards
+		if kerrors.IsAlreadyExists(err) {
+			return nil
+		}
```
maciaszczykm, line 213: Will this error happen if the name of the machine deployment is not set? I think in that case we might end up with multiple machine deployments. To avoid it I have added `InitialMachineDeploymentLabel` label and `isInitialMachineDeploymentCreated` check before. It can be done in a different way if you have any idea.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
maciaszczykm: All the errors go to event recorder already, so removing the annotation in that case is good way to go for me.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
```diff
@@ -0,0 +1,243 @@
...
+		KubermaticCluster: cluster,
+		Client:            r,
+	}
+
+	machineDeployment, err := machineresource.Deployment(cluster, nodeDeployment, datacenter, sshKeys, data)
+	if err != nil {
+		return fmt.Errorf("failed to assemble MachineDeployment: %v", err)
+	}
+
+	err = client.Create(r.ctx, machineDeployment)
```
maciaszczykm, line 206: What do you think about adding some event on success too?

maciaszczykm:
Looks very good. It seems a little bit better to have it in a new controller as it is easier to read code in that way. I have added some remarks/questions.

pkg/api/v1/types.go
```diff
@@ -1644,3 +1644,7 @@ func ToInternalClusterType(externalClusterType string) kubermaticv1.ClusterType
 	}
 	return kubermaticv1.ClusterTypeAll
 }
+
+const (
+	InitialMachineDeploymentRequestAnnotation = "InitialMachineDeploymentRequest"
```
moadqassem: Can you please change this to: 
```go
InitialMachineDeploymentRequestAnnotation = "kubermatic.io/initial-machinedeployment-request"
```

pkg/handler/v2/cluster/cluster.go
```diff
@@ -25,8 +25,6 @@ import (
 	"strconv"
 
 	"github.com/go-kit/kit/endpoint"
-	"github.com/prometheus/client_golang/prometheus"
-
```
moadqassem, line 29: Please add a new line here, to fix the regrouping

pkg/handler/common/cluster.go
```diff
@@ -59,12 +54,6 @@ import (
 // NodeDeploymentEvent represents type of events related to Node Deployment
 type NodeDeploymentEvent string
 
-const (
```
moadqassem, line 62: After removing those I guess this type `NodeDeploymentEvent` above is not used anymore.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
xrstf: Hm I was never really sure how an initial deployment looked like. As the user can specify a name for it in the dashboard in the cluster creation wizard, I was under the impression that the initial MD will always have a name set.

Will have to investigate.

xrstf: It turns out, for MachineDeployments the `pkg/resources/machine.Deployment()` function will set a `GenerateName` on the MD if there is no user-defined name set. This will then make Kubernetes auto-generate a name by adding a unique suffix to the GenerateName and using that as the MD's name. This means every time we would apply the same annotation, we would get a new MD.

To fix this, I simply made it so that the CreateEndpoint() already sets a fixed name (using the same logic as Kubernetes itself), so that the annotation always contains a name and we can then use AlreadyExists() to perform the uniqueness check.

pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go
xrstf: Added an event.

xrstf:
Added some unit tests as well. This is now ready for review.

pkg/api/v1/types.go
xrstf: Totally!

PR #6065, opened 2020-10-21
Support addons Docker image in image-loader

xrstf:
For users who do not have an addons folder locally (either because they do not develop their own addons or because they simply don't have it at hands), this PR extends our image-loader to

* determine the addon Docker image from the KubermaticConfiguration
* pull said image
* extract `/addons` onto the host system

This allows users to have the addons locally or let the image-loader take care of getting them. Very convenient!

As we do not support openshift in the image-loader, this only takes the Kubernetes addons image into account.

Docker offers the `docker cp` command, but that needs a Docker container, so I chose to keep it simple and just use `docker run`. As the addons image should always be inherited from our image, this means that we always have Alpine available.

```release-note
NONE
```

### DIFF
--- a/cmd/image-loader/main.go
+++ b/cmd/image-loader/main.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
 
 	"github.com/Masterminds/semver"
@@ -89,7 +90,7 @@ func main() {
 	flag.StringVar(&o.versionFilter, "version-filter", "", "Version constraint which can be used to filter for specific versions")
 	flag.StringVar(&o.registry, "registry", "", "Address of the registry to push to, for example localhost:5000")
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
-	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to the folder containing the addons")
+	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to a directory containing the KKP addons, if not given, falls back to the Docker image configured in the KubermaticConfiguration")
 	flag.Parse()
 
 	rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
@@ -131,6 +132,31 @@ func main() {
 		log.Fatalw("Error loading versions", zap.Error(err))
 	}
 
+	// if no local addons path is given, use the configured addons
+	// Docker image and extract the addons from there
+	if o.addonsPath == "" {
+		v := common.NewDefaultVersions()
+		addonsImage := kubermaticConfig.Spec.UserCluster.Addons.Kubernetes.DockerRepository + ":" + v.Kubermatic
+
+		tempDir, err := ioutil.TempDir("", "imageloader*")
+		if err != nil {
+			log.Fatalw("Failed to create temporary directory", zap.Error(err))
+		}
+		defer os.RemoveAll(tempDir)
+
+		log.Infow("Extracting addon manifests from Docker image", "image", addonsImage, "temp-directory", tempDir)
+
+		if err := docker.DownloadImages(ctx, log, false, []string{addonsImage}); err != nil {
+			log.Fatalw("Failed to download addons image", zap.Error(err))
+		}
+
+		if err := docker.ExtractDirectory(ctx, log, addonsImage, tempDir, "/addons"); err != nil {
+			log.Fatalw("Failed to extract addons", zap.Error(err))
+		}
+
+		o.addonsPath = tempDir
+	}
+
 	// Using a set here for deduplication
 	imageSet := sets.NewString(staticImages...)
 	for _, version := range versions {
@@ -144,7 +170,7 @@ func main() {
 			continue
 		}
 		versionLog.Info("Collecting images...")
-		images, err := getImagesForVersion(log, version, o.addonsPath)
+		images, err := getImagesForVersion(ctx, log, version, o.addonsPath)
 		if err != nil {
 			versionLog.Fatal("failed to get images", zap.Error(err))
 		}
@@ -172,7 +198,7 @@ func processImages(ctx context.Context, log *zap.SugaredLogger, dryRun bool, ima
 	return nil
 }
 
-func getImagesForVersion(log *zap.SugaredLogger, version *kubermaticversion.Version, addonsPath string) (images []string, err error) {
+func getImagesForVersion(ctx context.Context, log *zap.SugaredLogger, version *kubermaticversion.Version, addonsPath string) (images []string, err error) {
 	templateData, err := getTemplateData(version)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


cmd/image-loader/main.go
```diff
@@ -131,6 +132,31 @@ func main() {
 		log.Fatalw("Error loading versions", zap.Error(err))
 	}
 
+	// if no local addons path is given, use the configured addons
+	// Docker image and extract the addons from there
+	if o.addonsPath == "" {
+		v := common.NewDefaultVersions()
+		addonsImage := kubermaticConfig.Spec.UserCluster.Addons.Kubernetes.DockerRepository + ":" + v.Kubermatic
```
irozzo-1A: This panics when no `configuration-file` is given.

cmd/image-loader/main.go
```diff
@@ -172,7 +198,7 @@ func processImages(ctx context.Context, log *zap.SugaredLogger, dryRun bool, ima
 	return nil
 }
 
-func getImagesForVersion(log *zap.SugaredLogger, version *kubermaticversion.Version, addonsPath string) (images []string, err error) {
+func getImagesForVersion(ctx context.Context, log *zap.SugaredLogger, version *kubermaticversion.Version, addonsPath string) (images []string, err error) {
```
irozzo-1A: `ctx` is not used in `getImagesForVersion`, weird this was not reported as lint violation

pkg/docker/docker.go
```diff
@@ -151,3 +152,35 @@ func PushImage(ctx context.Context, log *zap.SugaredLogger, dryRun bool, image s
 
 	return nil
 }
+
+// ExtractDirectory copies the content from a directory out of the
+// container onto the host system
+func ExtractDirectory(ctx context.Context, log *zap.SugaredLogger, image string, dst string, src string) error {
```
irozzo-1A: why `ExtraceDirectory`? I think `Copy` is more clear as a name.

cmd/image-loader/main.go
```diff
@@ -89,7 +90,7 @@ func main() {
 	flag.StringVar(&o.versionFilter, "version-filter", "", "Version constraint which can be used to filter for specific versions")
 	flag.StringVar(&o.registry, "registry", "", "Address of the registry to push to, for example localhost:5000")
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
-	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to the folder containing the addons")
+	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to a directory containing the KKP addons, if not given, falls back to the Docker image configured in the KubermaticConfiguration")
```
irozzo-1A: Should not be better to have an `addons-image` flag that defaults to the addon image for this version instead of taking it from the `KubernetesConfiguration`? This way would work even when `version-file` is given.

pkg/docker/docker.go
xrstf: I started with a name that made sense to me, then found out about `docker cp` but decided that "Copy" was a bit vague. Apparently not.

cmd/image-loader/main.go
xrstf: Good catch!

xrstf:
Adressed all feedback points.

cmd/image-loader/main.go
```diff
@@ -131,6 +134,30 @@ func main() {
 		log.Fatalw("Error loading versions", zap.Error(err))
 	}
 
+	// if no local addons path is given, use the configured addons
+	// Docker image and extract the addons from there
+	if o.addonsPath == "" {
+		addonsImage := o.addonsImage
+		if addonsImage == "" {
+			if kubermaticConfig == nil {
```
irozzo-1A, line 146: Why you bother checking the kubermatic config? now that we have an explicit argument you could just use the default image if not specified otherwise.

xrstf: Because I want the users in the future to simply give us their KubermaticConfiguration and we figure out everything from there. The addons-path, addons-image etc. flags are only for customers who need to override it.

cmd/image-loader/main.go
```diff
@@ -131,6 +134,30 @@ func main() {
 		log.Fatalw("Error loading versions", zap.Error(err))
 	}
 
+	// if no local addons path is given, use the configured addons
+	// Docker image and extract the addons from there
+	if o.addonsPath == "" {
```
irozzo-1A, line 143: This means that `addonsPath` takes precedence. I would rather do the following that avoids unnecessary nesting and makes clear that they are mutually exclusives:

```
if o.addonsPath != "" && o.addonsImage != "" {
   fatal
}
if o.addonsPath != "" {
... // from fs
} else {
... // from image
}
```

cmd/image-loader/main.go
irozzo-1A: I'm not sure it's a very good idea. Users that are providing their custom images in the KubermaticConfiguration don't want probably to mirror them as they pushing them already in the right place. I think makes more sense to consider KubermaticConfiguration for versions only. And make this explicit in the flag description. This is how I see it but do as you want.

cmd/image-loader/main.go
xrstf: Leaving both addonsPath and addonsImage empty is a feature, not a bug.

irozzo-1A: I have no pb having both empty, they should not be both specified.

pkg/docker/docker.go
```diff
@@ -151,3 +152,35 @@ func PushImage(ctx context.Context, log *zap.SugaredLogger, dryRun bool, image s
...
+
+	log.Infow("Extracting image...", "src", src, "dst", dst)
+
+	dst, err = filepath.Abs(dst)
+	if err != nil {
+		return fmt.Errorf("failed to determine absolute path: %v", err)
+	}
+
+	mountPoint := "/kubermaticextractor"
+	args := []string{
```
irozzo-1A, line 171: sorry if I missed this in my previous pass, but here you should better use `docker cp` command instead. No need of running a container: 

https://docs.docker.com/engine/reference/commandline/cp/

cmd/image-loader/main.go
```diff
@@ -89,7 +91,8 @@ func main() {
 	flag.StringVar(&o.versionFilter, "version-filter", "", "Version constraint which can be used to filter for specific versions")
 	flag.StringVar(&o.registry, "registry", "", "Address of the registry to push to, for example localhost:5000")
 	flag.BoolVar(&o.dryRun, "dry-run", false, "Only print the names of found images")
-	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to the folder containing the addons")
+	flag.StringVar(&o.addonsPath, "addons-path", "", "Path to a directory containing the KKP addons, if not given, falls back to -addons-image, then the Docker image configured in the KubermaticConfiguration")
+	flag.StringVar(&o.addonsImage, "addons-image", "", "Docker image containing KKP addons, if not given, falls back to the Docker image configured in the KubermaticConfiguration")
```
irozzo-1A, line 95: I don't consider this a blocker as it is not prod ready, but if we want to support OpenShift addons as well we should allow specifying multiple images.

pkg/docker/docker.go
xrstf: As stated in the PR description:

> Docker offers the docker cp command, but that needs a Docker container, so I chose to keep it simple and just use docker run. As the addons image should always be inherited from our image, this means that we always have Alpine available.

Also stated in the documentation:

> Copy files/folders between a **container** and the local filesystem

xrstf: So it's either `docker create + docker cp + docker delete`, or a single `docker run --rm` with much less state handling and need to cleanup if any of the steps failed.

irozzo-1A: Yeah, you need to create the container, but you don't need to run it. I missed that sentence in the PR description. If we don't want to support pure data containers at the moment we can go for your solution as it is simpler.

PR #6066, opened 2020-10-21
Add Helm Support to image-loader

xrstf:
Previously, we had a shell script to run `helm template [chart] | ./retag-images.sh`. This script would then sed/grep for anything that looks like `image: ....` and return it.

This PR moves this logic into the image-loader. As we already have a neat little Helm client for the installer, it's easy to just extend it to allow rendering Helm charts.

For addons we already have a logic that would parse the output (YAML), read it into []runtime.RawExtension and then look for PodSpecs in a type-safe manner. This is now reused for handling the Helm output as well, making it much, MUCH safer.

This now reduces the work for users who want to mirror KKP to just running the image-loader (and having Helm 3 installed on their system).

The fixes for the kubernetes-dashboard are just here because the image-loader parses _all_ charts and this chart was simply broken. It will soon be moved out of this repository anyway and is currently not used anywhere, so this can be disregarded in this PR.

```release-note
NONE
```

### DIFF
--- a/pkg/install/helm/cli.go
+++ b/pkg/install/helm/cli.go
@@ -61,16 +61,7 @@ func (c *cli) InstallChart(namespace string, releaseName string, chartDirectory 
 		"--timeout", c.timeout.String(),
 	}
 
-	set := make([]string, 0)
-
-	for name, value := range values {
-		set = append(set, fmt.Sprintf("%s=%s", name, value))
-	}
-
-	if len(set) > 0 {
-		command = append(command, "--set", strings.Join(set, ","))
-	}
-
+	command = append(command, valuesToFlags(values)...)
 	command = append(command, flags...)
 	command = append(command, releaseName, chartDirectory)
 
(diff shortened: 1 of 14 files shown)


maciaszczykm:
/approve

For the k8s/dashboard part.

pkg/install/helm/cli.go
```diff
@@ -171,3 +175,17 @@ func (c *cli) run(namespace string, args ...string) ([]byte, error) {
 
 	return stdoutStderr, err
 }
+
+func valuesToFlags(values map[string]string) []string {
+	set := make([]string, 0)
```
irozzo-1A: Small thing, as we know already the size of the map you can use instead:
```
make([]string, 0, len(values))
```
to avoid allocations.

irozzo-1A:
I think overall is good, I did not see any evident issue. I would have preferred to have some unit test coverage (also to simplify my review), but I leave this up to you.

PR #6110, opened 2020-11-02
Proposal for KubeLB

mate4st:
Implementation of multi cluster load balancing

**Documentation**:

https://github.com/kubermatic-labs/kubelb

```release-note
NONE
```


docs/proposals/kubelb.md
```diff
@@ -0,0 +1,197 @@
...
+  <img src="images/kubelb-architecture.png" width="100%" />
+</div>
+
+
+It is required to have a service type "LoadBalancer" implementation on the load balancer cluster. This can be a cloud solution, or some single cluster implementations. 
+
+#### Implementation L4
+
+The agent watches for Services of type LoadBalancer/NodePort in the tenant cluster. In the tenant cluster itself for each service with type LoadBalancer a NodePort is allocated by default. 
+The agent informs the Manager which creates a Service and Endpoint in the LB cluster and adds the node IP addresses of the tenant cluster to the Endpoint IP addresses there. The agent watches for node changes like "remove", "add" and failures and will update the IP list in the Endpoint accordingly.
```
rastislavs: The usage of node IP addresses here makes an assumption on having them reachable from the LB cluster. This works if the LB cluster and the tenant clusters are part of the same admin domain (infrastructure), or if the tenant cluster nodes have public IPs.

Potentially, we could introduce some tunneling mechanism between the LB and the tenant clusters, which would also allow for:
- running tenant clusters behind NAT / firewall (e.g. LB cluster in a public cloud and tenant clusters on premises or on different cloud providers)
- encryption of the traffic between the LB and the tenant clusters (may be helpful especially if TLS sessions are terminated in the LB cluster and / or if public internet is used to interconnect LB and tenant clusters)

docs/proposals/kubelb.md
```diff
@@ -0,0 +1,197 @@
...
+Evaluation for failing node detection, so it is fast enough and meet our requirements. If not we need to do some active health checks.
```
rastislavs: Especially in case of multi-cluster load-balancing, in may be useful to take the number of active endpoints in individual clusters into account, so that we do not send too much traffic into clusters with too little endpoints (e.g. due to having many nodes compared to other clusters).

In case of single-cluster load-balancing, this may make sense as well when using `service.spec.externalTrafficPolicy` == `Local`.

docs/proposals/kubelb.md
mate4st: Thanks for pointing out this as a feature. If i have some time left i will have a look at this. I am a big fan of Wireguard which could be a potential easy implementation for this feature.

docs/proposals/kubelb.md
mate4st: I updated the proposal, the main change here is that i want to use envoy as a loadbalancer instead of relying on Kubernetes services, to be more flexible in general and use some health check mechanism which enables the use of `service.spec.externalTrafficPolicy` == `Local`.

PR #6122, opened 2020-11-04
Add Kubernetes 1.20

xrstf:
This adds support for Kubernetes 1.20 and makes 1.19 the default for new user clusters. Our Azure tests are defunct at the moment, the test failure is (probably) unrelated to Kubernetes 1.20. Likewise, the CSI stuff for Openstack is not supported in 1.20 anymore, so the tests can't pass. Handling Openstack compat should be done in a different PR.

1.19 has not been bumped to the most recent version because of two reasons:

* 1.19.4 has bugs in the conformance test, see https://github.com/kubernetes/kubernetes/issues/96577, https://github.com/kubernetes/kubernetes/issues/86418
* poseidon is not publishing kubelet Docker images for 1.19.4+

```release-note
Add Kubernetes 1.20
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -306,7 +306,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-      - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.22
+      - image: quay.io/kubermatic/e2e-kind:with-conformance-tests-v1.0.23
         command:
         - ./hack/ci/test-github-release.sh
         resources:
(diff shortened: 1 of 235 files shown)


irozzo-1A:
Closing at the moment as this was just a test.

xrstf:
Once more, then I'm convinced that it's not a fluke.

/retest

irozzo-1A:
@xrstf If you are actively working on this, I think you should update the mc first, it's maybe just a matter of updating the tested versions and the doc, but should be done as a prerequisite IMO e.g. https://github.com/kubermatic/machine-controller/pull/827

kron4eg (approved):
/lgtm
/approve

PR #6137, opened 2020-11-09
Update nodeport-proxy to use envoyv3

youssefazrak:
As part of the short-term strategy for the new expose strategy, we will use the existing Nodeport-proxy.
This one needs to be updated to use v3 packages.

```release-note
NONE
```


youssefazrak:
/override pre-kubermatic-spellcheck
See https://github.com/vmware-tanzu/velero/pull/3063

go.mod
```diff
@@ -99,7 +99,8 @@ replace (
 	github.com/go-logr/logr => github.com/go-logr/logr v0.2.1-0.20200730175230-ee2de8da5be6
 	github.com/go-logr/zapr => github.com/go-logr/zapr v0.2.0
 	// pin to prevent panics in DeepEqual() when comparing envoy objects
```
irozzo-1A: I would remove the entire comment as we are now using `reflect.DeepEqual` and this no longer applies.

youssefazrak: Good point, it's fixed now.

PR #6148, opened 2020-11-11
WIP - Update cert-manager to 1.0.4

xrstf:
This updates cert-manager to the first big stable release and paves the path for us to migrate away from v1alpha2, the version used everywhere in our Helm charts. As conversions between cert-manager CRDs are handled by the webhook, care must be taken to provide an upgrade path:

Just applying the 1.0.4 CRDs will fail, as they do not handle conversions for v1alpha2/3 CRs. Instead, the **non-legacy** 0.16.0 CRDs must be applied first, as they provide the necessary conversion from v1alpha2/3 to v1beta1. To complicate things, this involves the webhook, which needs to be named `cert-manager-webhook` and many moons ago I made the unwise decision to name it `webhook`. So the Helm release needs to be upgraded first. In simpler terms:

* `helm ugprade`
* `kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v0.16.0/cert-manager.crds.yaml`
* `k apply -f crd/` to install the 1.0.4 CRDs and the conversion from v1beta1 to v1

Alternative route: Delete all cert-manager resources, re-apply CRDs and then Helm upgrade.

**Documentation**:
WIP

```release-note
ACTION REQUIRED: Update cert-manager to 1.0.4; CRDs must be migrated in two steps, see upgrade notes.
```


xrstf:
This will be resurrected once #6150 has been merged.

PR #6150, opened 2020-11-11
Rename cert-manager Deployments/Service to conform with CRDs

xrstf:
The non-legacy cert-manager CRDs use a conversion webhook to serve multiple API versions. This ewbhook Service is called "cert-manager-webhook", yet when I adopted the chart many moons ago, I named the service just "webhook".
This will lead to all sorts of problems going forward, when we finally move away from v1alpha2. That's why this PR does the first necessary step and just renames the Deployments and Services.

```release-note
Fix cert-manager webhook Service naming
```


xrstf:
/cherrypick release/v2.15

Though not strictly necessary, having this backported into 2.15 will make the 2.16 upgrades much easier for customers, as they can rely on the webhook being setup properly. Kubernetes issues very scary and misleading warnings when the hook is not available.

kron4eg (approved):

PR #6154, opened 2020-11-12
Refactor nodeport proxy

irozzo-1A:
**What this PR does / why we need it**:
This PR prepares the terrain for #5804 

It includes the following changes:
* Moves the nodeport proxy controller logic in `pkg`.
* Build envoy cluster LB endpoints based on service endpoints instead of using pods.
* Small improvements

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6159, opened 2020-11-16
Fix nodeport proxy bugs introduced during refactoring

irozzo-1A:
This PR fixes a few bugs introduced with #6154:
* The envoy configuration is not set in the snapshot cache until the first reconcile request arrives, this causes the liveness probe to fail if no exposed services are present.
* Method creating the snapshot does not take into account the version provided in input.

Fixes #

**Documentation**:

```release-note
NONE
```


youssefazrak (approved):
/lgtm

PR #6161, opened 2020-11-16
Improve Version Handling

xrstf:
Previously, KKP used 2 different places to inject build-time versioning information:

* `pkg/resources` contained the `KUBERMATICCOMMIT` and `KUBEMATICGITTAG` variables, used as static global variables throughout the codebase.
* `pkg/operator/common` contained `KUBERMATICDOCKERTAG` and `UIDOCKERTAG`, but these two variables are wrapped up in a neat `Versions` struct that is then injected to where it is needed.

This gives us 4 variables in total, which **were** used like this:

* `KUBERMATICCOMMIT`
  * kubeletdnat-controller Docker tag (resources/vpnsidecar)
  * usercluster-ctrlmgr Docker tag
  * for nodeport-proxy image tag (for the custom per-cluster nodeport-proxy thingy, not the default one!)
  * for etcd-launcher image tag
  * for user-ssh-keys-agent image tag
  * for API image in openshift machine-controller init container
  * as status.kubermaticVersion in Cluster CRD
* `KUBERMATICGITTAG`
  * returned in the API `/v1/version` endpoint
* `KUBERMATICGITTAG+"-"+KUBERMATICCOMMIT`
  * used as KubermaticVersion in ClusterConditions
* `UIDOCKERTAG`
  * operator.common.Versions
    * dashboard Deployment in Operator
* `KUBERMATICDOCKERTAG`
  * operator.common.Versions
     * for addon image tag
     * for nodeport-proxy image tag
     * API Deployment in Operator
     * Master Ctrlmgr. Deployment in Operator
     * Seed Ctrlmgr. Deployment in Operator

This PR merges the two approaches, favoring the injection variant. As this lead to import cycles, the new location is `pkg/version/kubermatic`. It's not perfect, but good enough (as I could not think of a better package name, `pkg/version` is already used for **Kubernetes** Versions (i.e. the versions.yaml)).

After merging the variables into a single package, this PR then removes the old global variable access (`resources.KUBERMATICCOMMIT`) and replaces it with properly injected version variables. This means most controllers now get the KKP version handed in from the main() function. This also means that most of this PR is "chasing dependencies", i.e. adding just _one more_ parameter to many functions. ^.^

Besides merging the variables, the usage has also been changed: The nodeport-proxy, etcd-launcher, user-ssh-keys-agent and API do not use the commit hash anymore, but the version (which *is* the commit hash if nothing was tagged). This means that users who installed `quay.io/kubermatic/kubermatic:v1.2.3` will not get the same tag used for the user-cluster-controller-manager, instead of this image using the Git hash. This makes it more consistent and easier to create a Docker image mirror registry.

This is all done in preparation to make the KKP Installer be actually able to migrate from vX to vY.

```release-note
NONE
```


Makefile
```diff
@@ -21,18 +21,16 @@ REPO = $(DOCKER_REPO)/kubermatic$(shell [ "$(KUBERMATIC_EDITION)" != "ce" ] && e
 CMD = $(filter-out OWNERS nodeport-proxy kubeletdnat-controller, $(notdir $(wildcard ./cmd/*)))
 GOBUILDFLAGS ?= -v
 GOOS ?= $(shell go env GOOS)
-GITTAG = $(shell git describe --tags --always)
```
xrstf, line 24: Fun fact: we had some places that tried to override this variable from the outside, but that never had any effect.

Makefile
```diff
@@ -114,9 +112,6 @@ endif
 		$(DOCKER_BIN) push $(REPO):$$tag; \
 	done
 
-gittag:
-	@echo $(GITTAG)
-
```
xrstf, line 132: There was no point to this target.

xrstf:
/hold

Giving my team a chance to look at this.

xrstf:
@zreigz Can you please re-review? I had to rebase this.

/hold cancel

PR #6163, opened 2020-11-17
handle dockerizing spellchecker like any other tool

xrstf:
This removes the `spellcheck-in-docker` Make target and instead uses the same approach as we use for all other tools: Have a script and this script calls itself in Docker when necessary. And as the codespell image does not have bash, the PR also makes the lib.sh POSIX compatible again.

```release-note
NONE
```

### DIFF
--- a/hack/verify-spelling.sh
+++ b/hack/verify-spelling.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env sh
+
+# Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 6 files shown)


hack/verify-spelling.sh
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+cd $(dirname $0)/..
+. hack/lib.sh
+
+CONTAINERIZE_IMAGE=quay.io/kubermatic/codespell:1.17.1 containerize ./hack/verify-spelling.sh
```
irozzo-1A, line 22: Personally, I find this pattern quite hacky :-) but if this is what you decided to do, let's conform to it.

PR #6170, opened 2020-11-17
Fix Constraint match array

lsviben:
Fixes our kubermatic Constraints to fit the gatekeeper ones.

```release-note
NONE
```


lsviben:
for reference - https://github.com/open-policy-agent/gatekeeper#constraints . I accidentally made ours with string instead of string array.

Issue #5804, opened 2020-09-01
Implement Envoy LB for the new expose strategy


youssefazrak:
**User Story**
The goal of this issue is to implement the Envoy SNI LB for the new expose strategy. We will enhance the existing Nodeport-proxy. 

We are dealing with two types of traffic. The bootstrap one for the Kubelet and the rest.
Regarding the Kubelet traffic going to the APIServer of the Control Plane, we need to load balance the traffic hitting the Seed cluster to the right control plane.
This can be done using SNI as the traffic from the Kubelet to initiate the bootstrap process is using HTTPS and SNI.
We will be implementing this part on the seed by using Envoy as a generic ingress controller that will send the traffic to the right control plane.
Moreover, the Ingress Controller needs to support TLS passthrough. This is to prevent another of certificates management and have a "transparent" load balancing that doesn't terminate the TLS connection between the User cluster and the Seed cluster.
The Seed Controller Manager will take care of the provisioning and the lifecycle of this component.

**Acceptance criteria**
Envoy is deployed and is able to load balance TLS traffic using SNI and without termination to the right Control Plane.

irozzo-1A:
This is implemented.

PR #6182, opened 2020-11-19
Add nodeport proxy integration test

irozzo-1A:

This PR introduces two basic e2e tests for node-port proxy. As we are going to introduce some changes due to #5804, it's important to have a base to add new test cases and to make sure we are not introducing evident regressions.
The test cases introduced so far perform the following things. After the nodeport proxy is deployed in a dedicated namespace (using the resource functions of the operator), the following test scenarios are run:

1. First test case verifies that the services exposed by nodeport proxy are reachable passing from the service of type load-balancer.
   - Some services of type node port with the expose annotation matching some pods exposing an HTTP endpoint are deployed.
   - A pod with curl using the host network is deployed to shoot the service is deployed.
   - For each exposed service port a predefined number of HTTP requests is made targeting the node-port of the lb service of node-port proxy.
   - The test succeeds if all pods are reached, it fails otherwise.
2. The second test ensures that nodeport services not having the expose annotation are not exposed.
    - Some services of type node port without the expose annotation are created.
    - The test asserts that the lb service should not be updated.

Fixes #

The test uses Ginkgo framework and many helper functions are based on the k8s e2e framework. I did not use it directly as it is part of kubernetes repository. In case it will be externalized one day it could be more convenient to import it.
See https://github.com/kubernetes/kubernetes/issues/74352

**Documentation**:

```release-note
NONE
```


pkg/test/e2e/nodeport-proxy/const.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeportproxy
+
+// TODO(irozzo) make registries configurable
+const (
+	AgnosImage   = "k8s.gcr.io/e2e-test-images/agnhost:2.21"
+	NetexecImage = "gcr.io/kubernetes-e2e-test-images/netexec:1.1"
```
youssefazrak: Is there a specific use for using those two different images?
The commenter [here](https://github.com/kubernetes/kubernetes/issues/76342#issuecomment-495641423) says that the 1.1 version already contains netexec.

pkg/test/e2e/nodeport-proxy/services.go
```diff
@@ -0,0 +1,178 @@
...
+		// Set back namespace name in case it was generated
+		n.Namespace = ns.Name
+		n.Log.Debugw("Setting generated namespace to SeriviceJig", "namespace", n.Namespace)
+	}
+
+	labels := map[string]string{"apps": name}
+	// Finally create service
+	svc := n.newServiceTemplate(name, n.Namespace, annotations, labels)
+	svc.Spec.Type = corev1.ServiceTypeNodePort
+	svc.Spec.Ports[0].NodePort = nodePort
```
youssefazrak, line 66: Is there a reason for having those setup after the creation of the svc and not directly inside the `newServiceTemplate` method, with the `nodePort` as an additional argument?

pkg/test/e2e/nodeport-proxy/const.go
irozzo-1A: Good catch, the `agnhost` image contains all that we need for the test.

pkg/test/e2e/nodeport-proxy/services.go
irozzo-1A: you're right, at the moment it would be better to do as you suggested. I thought that keeping it generic will be beneficial as we will need also to expose services of type ClusterIp in a near future ;-)

pkg/test/e2e/nodeport-proxy/nodeport_proxy_suite_test.go
```diff
@@ -0,0 +1,72 @@
...
+*/
+
+package nodeportproxy
+
+import (
+	"flag"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
youssefazrak: Named import to be used.

PR #6184, opened 2020-11-20
Retry timed out request in api e2e tests

irozzo-1A:
This PR is an attempt to fix the following issue occurring during our e2e tests:

>     credentials_test.go:146: error reading response: Get "http://localhost:8080/api/v1/providers/azure/sizes": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/securecookie v1.1.1
 	github.com/gorilla/websocket v1.4.2
+	github.com/hashicorp/go-multierror v1.0.0
 	github.com/heptiolabs/healthcheck v0.0.0-20180807145615-6ff867650f40
 	github.com/hetznercloud/hcloud-go v1.15.1
 	github.com/huandu/xstrings v1.2.0 // indirect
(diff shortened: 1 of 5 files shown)


irozzo-1A:
```
E1121 10:37:41.740736   10974 portforward.go:385] error copying from local connection to remote stream: read tcp4 127.0.0.1:8080->127.0.0.1:41400: read: connection reset by peer
```

This seems to trigger the condition, and retrying the request does not seem to address the problem. I need to dig a bit more into this.

irozzo-1A:
> ```
> E1121 10:37:41.740736   10974 portforward.go:385] error copying from local connection to remote stream: read tcp4 127.0.0.1:8080->127.0.0.1:41400: read: connection reset by peer
> ```
> 
> This seems to trigger the condition, and retrying the request does not seem to address the problem. I need to dig a bit more into this.

After double-checking the timeout seems to be related to the Azure API endpoint used to list VM sizes takes long time to reply. The test sets the timeout to 10s, reducing it to 5s (even with retry) makes the timeout triggering systematically. It would be interesting to understand why it's taking so long to get an answer, but having `10s` seems to be effective to reduce failures.

irozzo-1A:
/hold

I saw we have already a client with retry logic

youssefazrak (approved):

PR #6192, opened 2020-11-24
Envoy Clusters / Listeners for New expose strategy

youssefazrak:
Introduces functions to the "nodeport-proxy" to create Envoy's clusters and listeners for the new expose strategy.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/nodeport-proxy/envoymanager/controller.go
+++ b/pkg/controller/nodeport-proxy/envoymanager/controller.go
@@ -151,7 +151,8 @@ func (r *Reconciler) sync() error {
 			serviceLog.Debug("skipping service: it has no running pods")
 			continue
 		}
-		l, c := r.makeListenersAndClustersForService(&service, &eps)
+		l := r.makeListenersForNodePortService(&service)
+		c := r.makeClusters(&service, &eps)
 		listeners = append(listeners, l...)
 		clusters = append(clusters, c...)
 
--- a/pkg/controller/nodeport-proxy/envoymanager/resources.go
+++ b/pkg/controller/nodeport-proxy/envoymanager/resources.go
@@ -27,22 +27,264 @@ import (
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/types/known/anypb"
 	corev1 "k8s.io/api/core/v1"
 
+	//	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+
+	envoyaccesslogv3 "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	envoyclusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	envoycorev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoyendpointv3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	envoylistenerv3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	envoyroutev3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
+	envoylistenerlogv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/file/v3"
 	envoyhealthv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/health_check/v3"
 	envoyhttpconnectionmanagerv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	envoytcpfilterv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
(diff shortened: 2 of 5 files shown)


pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
 	envoytcpfilterv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
 	envoycachetype "github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	envoycachev3 "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	envoywellknown "github.com/envoyproxy/go-control-plane/pkg/wellknown"
 )
 
+const (
+	HTTPSAltPort  = 8443
+	HTTPProxyPort = 8080
```
irozzo-1A: I would put those in the Reconciler Options later on, we will get their values with specific flags.

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
+			},
+		},
+	}
+	return virtualhosts
+}
+
+//TODO(youssefazrak) Currently we have a 1:1 bind between the listener itself
+// and the virtual_hosts. Each "tunnel" listener of the new expose strategy
+// should have two virtual_hosts, one for OpenVPN and one for the KAS.
+func (r *Reconciler) makeTunnelingListener(service *corev1.Service) *envoylistenerv3.Listener {
```
irozzo-1A: I would add the tunneling and SNI listener to the initial resources, as they are immutable. We should add them only if the listener ports are given.

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
 
+const (
+	HTTPSAltPort  = 8443
+	HTTPProxyPort = 8080
+	UpgradeType   = "CONNECT"
+)
+
+var stdoutAccessLog *anypb.Any
+
+func init() {
```
irozzo-1A: No need of declaring global variables and doing this init, if the reason you did that is to optimize it's not worthy ;-). I would move this into `makeAccessLog` directly.

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
+	}
+
+	stdoutAccessLog, err = ptypes.MarshalAny(f)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func makeAccessLog() []*envoyaccesslogv3.AccessLog {
+	accessLog := []*envoyaccesslogv3.AccessLog{
```
irozzo-1A, line 68: you can return this without declaring `accessLog` variable

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
+			ConfigType: &envoyaccesslogv3.AccessLog_TypedConfig{
+				TypedConfig: stdoutAccessLog,
+			},
+		},
+	}
+	return accessLog
+}
+
+func makeSNIFilterChainMatch(service *corev1.Service) *envoylistenerv3.FilterChainMatch {
+	filterChainMatch := &envoylistenerv3.FilterChainMatch{
```
irozzo-1A: same here

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -27,21 +27,259 @@ import (
...
+		// New annotation used for the new expose strategy. The "expose" one will stay
+		// for the default expose.
+		ServerNames:       []string{service.ObjectMeta.Annotations["nodeport-proxy.k8s.io/domainname"]},
```
irozzo-1A: in theory, the Annotations map should not be `nil` but I would add a check, also the specific annotation may be not present, in this case we are getting an empty string.

Maybe you should just add `serverName` as a function parameter. It will be the responsibility of the caller not to call this function if the annotation is not present.

pkg/controller/nodeport-proxy/envoymanager/controller.go
```diff
@@ -147,7 +151,8 @@ func (r *Reconciler) sync() error {
 			serviceLog.Debug("skipping service: it has no running pods")
 			continue
 		}
-		l, c := r.makeListenersAndClustersForService(&service, &eps)
+		l := r.makeListenersForNodePortService(&service)
+		c := r.makeClusters(&service, &eps)
```
irozzo-1A, line 155: You could avoid declaring `l` and `c` and call those in the `append` directly

pkg/controller/nodeport-proxy/envoymanager/resources.go
```diff
@@ -29,26 +29,223 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 
+	//	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
```
irozzo-1A, line 32: please remove this

irozzo-1A:
/lgtm

I'm going to address the remaining points with a follow-up PR.

Issue #5805, opened 2020-09-01
Implement Envoy Proxy for the new expose strategy


youssefazrak:
**User Story**
The goal of this issue is to implement the Envoy proxy on the Seed side for the new expose strategy. We will enhance the existing Nodeport-proxy. 

As mentioned in the #5804  we are dealing with two types of traffic. The bootstrap one for the Kubelet and the rest.
Regarding the Pods traffic going to the APIServer, but also the Kubelet traffic (after the bootstrap process) those are not using SNI and will be tunneled from the User cluster to the Seed cluster.
The LB will be done using HTTP2.0 Authority header. Envoy will be deployed on the Seed side.
Configuration of the Envoy will be managed using GO Api.
This will be using the same Envoy deployment as in #5804

**Acceptance criteria**
Envoy is deployed and is routing the tunneled traffic to the right control plane.

irozzo-1A:
The nodeport proxy should now be able to handle both SNI and HTTP/2 CONNECT traffic.

PR #6200, opened 2020-11-25
Add new annotations to control the way envoy-manager exposes the services

irozzo-1A:
This PR introduces new annotations that are processed by the `envoy-manager` to expose services with SNI and HTTP2 connect strategy. #5805

Before this  PR the only allowed "exposing" annotation (i.e. `"nodeport-proxy.k8s.io/expose"` by default) value was `true`. Now it is possible to add a comma-separated list of `ExposeType`s that don't just define if the service is exposed, but also how. Allowed values are `NodePort`, `SNI` and `HTTP2Connect`. The value `true` is still valid for backward compatibility and it is equivalent to `NodePort`.

Another annotation has been introduced specifically for the `SNI` expose type, `nodeport-proxy.k8s.io/port-mapping`, which contains a JSON encoded mapping between service port names and the FQDNs that will be used for the service.

This PR does not implement the new expose types, the implementation is deferred to a follow-up PR.

Fixes #

This PR does not fully implement the new expose types.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/nodeport-proxy/envoymanager/utils.go
+++ b/pkg/controller/nodeport-proxy/envoymanager/utils.go
@@ -28,22 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// ExposeType defines the strategy used to expose the service.
-type ExposeType string
-
-func ExposeTypeFromString(s string) (ExposeType, bool) {
-	switch s {
-	case NodePortType.String():
-		return NodePortType, true
-	case SNIType.String():
-		return SNIType, true
-	case HTTP2ConnectType.String():
-		return HTTP2ConnectType, true
-	default:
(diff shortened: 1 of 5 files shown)


pkg/controller/nodeport-proxy/envoymanager/utils.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
```
irozzo-1A, line 31: Move it after type declaration

Issue #6220, opened 2020-11-26
etcd-launcher should handle the member pod start with empty data directory


furkhat:
If data directory is empty the node should start as a fresh member.

PR #6221, opened 2020-11-26
[etcd] start as a fresh member if data dir is empty

furkhat:

When a member pod starts with empty directory etch crashes.

#6220 

**Special notes for your reviewer**: none

**Documentation**: none

```release-note
start as a fresh member if data directory is empty
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -18,27 +18,26 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
-	"log"
+	"io"
 	"net/url"
 	"os"
 	"os/exec"
-	"path"
+	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
+	"k8c.io/kubermatic/v2/pkg/resources"
 
 	"go.etcd.io/etcd/v3/clientv3"
 	"go.etcd.io/etcd/v3/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/v3/etcdserver/etcdserverpb"
 	"go.etcd.io/etcd/v3/pkg/transport"
 	"go.uber.org/zap"
-
-	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
-	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
-	"k8c.io/kubermatic/v2/pkg/resources"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -54,7 +53,7 @@ const (
 	initialStateNew          = "new"
 )
 
-type config struct {
+type etcdCluster struct {
 	namespace             string
 	clusterSize           int
 	podName               string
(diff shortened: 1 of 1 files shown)


multi-io:
Can we remove the etcdrestore - related parts (e.g. restoreDatadirFromBackupIfNeeded()) from this, as those are part of PR https://github.com/kubermatic/kubermatic/pull/5668 and current upstream master doesn't contain the CRDs?

cmd/etcd-launcher/main.go
```diff
@@ -66,145 +68,190 @@ type config struct {
...
+	if err != nil {
+		log.Panicw("failed to check cluster membership", zap.Error(err))
 	}
+	if thisMember != nil {
+		log.Infof("%s is a member", thisMember.Name)
 
+		if _, err := os.Stat(filepath.Join(e.dataDir, "member")); os.IsNotExist(err) {
+			client, err := e.getClusterClient()
+			if err != nil {
+				log.Panicw("can't find cluster client: %v", err)
```
xrstf: This needs a `zap.Error()` wrapping.

cmd/etcd-launcher/main.go
```diff
@@ -66,145 +68,190 @@ type config struct {
...
+			}
+			log.Warnw("No data dir, to ensure recovery removing and adding the member")
+			_, err = client.MemberRemove(context.Background(), thisMember.ID)
+			if err != nil {
+				close(client, log)
+				log.Panicw("remove itsely due to data dir loss", zap.Error(err))
```
xrstf: itsely? itself?

xrstf (commented):
minor oversights

/ok-to-test
/test-all

cmd/etcd-launcher/main.go
```diff
@@ -66,145 +68,190 @@ type config struct {
...
+			}
+			close(client, log)
+			joinCluster(e, log)
+		}
+	} else { // new etcd member, need to join the cluster
+		joinCluster(e, log)
```
moelsayed: Can we please return err here an evaluate it? I know the function will panic on error, it would be more readable if we do that here.

cmd/etcd-launcher/main.go
```diff
@@ -66,145 +68,190 @@ type config struct {
...
-	log := rawLog.Sugar()
+	return rawLog.Sugar()
+}
 
-	// here we find the cluster state
+func inClusterClient(log *zap.SugaredLogger) ctrlruntimeclient.Client {
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		log.Fatalw("failed to get in cluster config", zap.Error(err))
+		log.Panicw("failed to get in cluster config", zap.Error(err))
```
moelsayed: Why are we replacing `Fatal` with `Panic`?

cmd/etcd-launcher/main.go
```diff
@@ -297,6 +341,8 @@ func etcdCmd(config *config) []string {
 		fmt.Sprintf("--cert-file=%s", resources.EtcdCertFile),
 		fmt.Sprintf("--key-file=%s", resources.EtcdKetFile),
 		"--auto-compaction-retention=8",
+		"--heartbeat-interval=1000",
+		"--election-timeout=5000",
```
moelsayed: Can you please explain why those values are used?

moelsayed (changes requested):
Looks good. Thank you. Just a couple of clarifications and 1 change.

cmd/etcd-launcher/main.go
furkhat: semantically it is the same. The problem is that fatal does `os.Exit(1)` and ignores defers (if there are any). There are few other places in the code with similar problems, most of which can be identified with golint I think.

cmd/etcd-launcher/main.go
furkhat: Oh sorry it shouldn't be there, I deleted.

moelsayed:
@furkhat can you please update the fixtures?

furkhat:
@moelsayed I run `./hack/update-fixtures.sh` and no diffs. It seems there is nothing to update.

PR #6254, opened 2020-12-03
Extend retry parameters in http framework for e2e tests

irozzo-1A:
This PR makes the HTTP retry parameters for the HTTP framework used in e2e tests more generic, allowing to specifying the exact number of retries and configure an exponential backoff.

 

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/test/e2e/api/credentials_test.go
+++ b/pkg/test/e2e/api/credentials_test.go
@@ -141,10 +141,14 @@ func TestProviderEndpointsWithCredentials(t *testing.T) {
 				req.Header.Set("Location", tc.location)
 			}
 
-			// should be able to perform at least 4 calls in case request
-			// timeout is hit all the times
-			client := utils.NewHTTPClientWithRetries(t, 10*time.Second, 1*time.Second, 50*time.Second)
-
+			// with those settings the cumulative sleep duration is ~ 8s
+			// when all attempts are made.
+			client := &http.Client{
+				Transport: utils.NewRoundTripperWithRetries(t, 15*time.Second, utils.Backoff{
+					Steps:    4,
+					Duration: 1 * time.Second,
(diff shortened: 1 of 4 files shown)


xrstf:
Please let's not go back to exponential backoffs. They serve no purpose in testing local applications and only slow down tests needlessly. I specifically removed them all just a few weeks ago and replaced them with waiting loops and reduced the runtime of our API tests by minutes. Calling the locally running KKP API in test environments is not costly in terms of performance and does not incur bandwidth charges.

irozzo-1A:
> Please let's not go back to exponential backoffs. They serve no purpose in testing local applications and only slow down tests needlessly. I specifically removed them all just a few weeks ago and replaced them with waiting loops and reduced the runtime of our API tests by minutes. Calling the locally running KKP API in test environments is not costly in terms of performance and does not incur bandwidth charges.

@xrstf No worries, I did not see your previous PR but I'm not going back to anything. I agree exponential backoff is not very useful in many cases for the e2e tests. What I did makes just the retry configuration more flexible, but do not absolutely mandate to use of exponential backoff. I think it's more natural in many cases to specify the number of retries more than the timeout. 

I used the exponential backoff just [here](https://github.com/kubermatic/kubermatic/blob/f17e31d1b9f83412804c73394cf3e3e79251bb4d/pkg/test/e2e/api/credentials_test.go#L146) test where we are doing a request to Azure API (maybe we should mock it to have more predictable tests but this is another topic) that is taking a long time, and retrying every second for three minutes does not help. I'm doing 4 attempts with a total cumulative sleep of ~8s and I think it's reasonable. i.e. it is failing faster than before.
For the rest of the tests, I did not change much and used constant backoff. I think in the future we could probably better tune some parameters, because IMO in some cases we are doing way too many attempts.

PR #6255, opened 2020-12-03
added proper gatekeeper namespace

lsviben:
To find the gatkeeper config, gatkeeper audit check the given namespace and hardcoded "config" name (https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/controller/config/config_controller.go#L196-L198).
So this PR creates the `gatekeeper-system` namespace on the user cluster as a place where to hold the gatekeeper config CR and whatever gatekeeper will need in the future.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/namespace.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/namespace.go
@@ -23,7 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// NamespaceCreator creates the namespace for the Kubernetes Dashboard
+// NamespaceCreator creates the namespace for Gatekeeper
 func NamespaceCreator() (string, reconciling.NamespaceCreator) {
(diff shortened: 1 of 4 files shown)


moadqassem:
@lsviben OPA integration is enabled per cluster, thus why not saving these configs in the user cluster namespace ?

pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper/namespace.go
```diff
@@ -0,0 +1,31 @@
...
+package gatekeeper
+
+import (
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NamespaceCreator creates the namespace for the Kubernetes Dashboard
```
moadqassem: Please adjust the comment to reflect what this NamspaceCreator is doing :-)

PR #6272, opened 2020-12-08
Default to the latest available OpenStack CCM version

xmudrii:

Currently, we determine the OpenStack CCM version based on the Kubernetes version. In the case when we don't specify the OpenStack CCM version for a Kubernetes version, in-tree CCM would be used as a fall-back.

In this concrete case, we didn't specify the OpenStack CCM version for Kubernetes 1.19, so all newly-created 1.19 OpenStack clusters are using the in-tree cloud provider instead of the external CCM. This should be considered as a breaking change because we have deployed the external CCM for all 1.17 and 1.18 OpenStack clusters.

OpenStack clusters upgraded from 1.18 to 1.19 are not affected. Seed-controller-manager will log errors regarding that Kubernetes version is not supported, but the cluster will still use external CCM deployed for 1.18.

* The 1.19 OpenStack clusters that were created before this change will continue using the in-tree cloud provider instead of the external CCM
* Those clusters will have to use the CCM migration mechanism to migrate to the external CCM. This mechanism is still **not implemented**
* All newly-created 1.19+ OpenStack clusters created after this change will use the external CCM

In the case when OpenStack CCM is not available for the latest Kubernetes versions, the latest available OpenStack CCM version will be used.

The documentation on adding new Kubernetes versions support has been extended with guidelines for adding a new OpenStack CCM version for that Kubernetes version.

```release-note
Default to the latest available OpenStack CCM version. This fixes a bug where newly-created OpenStack clusters running Kubernetes 1.19 were using the in-tree cloud provider instead of the external CCM. Those clusters will remain to use the in-tree cloud provider until the CCM migration mechanism is not implemented. The OpenStack clusters running Kubernetes 1.19+ and created with KKP 2.15.6+ will use the external CCM.
```

/assign @irozzo-1A

### DIFF
--- a/pkg/resources/cloudcontroller/openstack.go
+++ b/pkg/resources/cloudcontroller/openstack.go
@@ -171,13 +171,19 @@ func getOSFlags(data *resources.TemplateData) []string {
 }
 
 func getOSVersion(version semver.Semver) (string, error) {
+	if version.Major() < 17 {
+		return "", fmt.Errorf("Kubernetes version %s is not supported", version.String())
+	}
+
 	switch version.Minor() {
 	case 17:
 		return "1.17.0", nil
 	case 18:
 		return "1.18.0", nil
+	case 19:
+		return "1.19.2", nil
 	default:
-		return "", fmt.Errorf("Kubernetes version %s is not supported", version.String())
+		return "1.19.2", nil
 	}
 }
 
--- a/docs/adding-kubernetes-versions.md
+++ b/docs/adding-kubernetes-versions.md
@@ -38,6 +38,10 @@ Once new Docker images are ready, KKP can be updated as well.
   minor (just copy an existing job and adjust accordingly). Make sure to change the Docker
   image tag for the e2e images to use the new tag you just created with the new test binaries.
 - Update the CSI addon manifests (`addon/csi/*.yaml`) to include the new minor version.
+- Update the OpenStack CCM manifest (`pkg/resources/cloudcontroller/openstack.go`) to
+  include the new minor version.
+  - The latest OpenStack CCM version can be found in the 
+  [`kubernetes/cloud-provider-openstack` repository](https://github.com/kubernetes/cloud-provider-openstack).
 - Ensure a kubelet ConfigMap for the new minor exists in `addons/kubelet-configmap/kubelet-configmap.yaml`.
 - Update `addons/rbac/allow-kubeadm-join-configmap.yaml` to include the new ConfigMap.
 - The conformance-tests runner (`cmd/conformance-tests/runner.go`) has a list of


pkg/resources/cloudcontroller/openstack.go
```diff
@@ -171,13 +171,19 @@ func getOSFlags(data *resources.TemplateData) []string {
...
+		return "", fmt.Errorf("Kubernetes version %s is not supported", version.String())
+	}
+
 	switch version.Minor() {
 	case 17:
 		return "1.17.0", nil
 	case 18:
 		return "1.18.0", nil
+	case 19:
+		return "1.19.3", nil
```
irozzo-1A: I don't think this version exists yet:
```
curl https://registry.hub.docker.com/v1/repositories/k8scloudprovider/openstack-cloud-controller-manager/tags
[{"layer": "", "name": "latest"}, {"layer": "", "name": "1.13.0"}, {"layer": "", "name": "1.13.1"}, {"layer": "", "name": "v0.1.0"}, {"layer": "", "name": "v0.2.0"}, {"layer": "", "name": "v0.3.0"}, {"layer": "", "name": "v1.14.0"}, {"layer": "", "name": "v1.15.0"}, {"layer": "", "name": "v1.16.0"}, {"layer": "", "name": "v1.17.0"}, {"layer": "", "name": "v1.18.0"}, {"layer": "", "name": "v1.18.0-cpo.2020.2.26"}, {"layer": "", "name": "v1.18.1"}, {"layer": "", "name": "v1.18.2"}, {"layer": "", "name": "v1.19.0"}, {"layer": "", "name": "v1.19.1"}, {"layer": "", "name": "v1.19.2"}]
```

pkg/resources/cloudcontroller/openstack.go
```diff
@@ -171,13 +171,19 @@ func getOSFlags(data *resources.TemplateData) []string {
 }
 
 func getOSVersion(version semver.Semver) (string, error) {
+	if version.Major() < 17 {
+		return "", fmt.Errorf("Kubernetes version %s is not supported", version.String())
```
irozzo-1A, line 175: nit: normally you don't capitalize the first error char, this case is particular because it's a name.

xmudrii: I've just copied the error message as it was before, I haven't changed anything regarding that.

pkg/resources/cloudcontroller/openstack.go
xmudrii: You're correct, the image is still not pushed. I'll update the PR to use 1.19.2 instead.

pkg/resources/cloudcontroller/openstack.go
```diff
@@ -171,13 +171,19 @@ func getOSFlags(data *resources.TemplateData) []string {
...
 	switch version.Minor() {
 	case 17:
 		return "1.17.0", nil
 	case 18:
 		return "1.18.0", nil
+	case 19:
+		return "1.19.3", nil
 	default:
-		return "", fmt.Errorf("Kubernetes version %s is not supported", version.String())
+		return "1.19.3", nil
```
irozzo-1A: @marko Wouldn't be better to return an error in case we are trying to install a version greater than 1.19? We would also need to change a bit the logic to prevent the cluster from being created in such a case to avoid forgetting about this when we will introduce support for 1.20. If you prefer we can do this in a follow-up PR

pkg/resources/cloudcontroller/openstack.go
irozzo-1A: true, BTW this is not a blocker ;-)

pkg/resources/cloudcontroller/openstack.go
xmudrii: @irozzo-1A Fixed, PTAL.

pkg/resources/cloudcontroller/openstack.go
xmudrii: No, it was like that before this change, but it's not a good idea, IMO. It's better to fall-back to the latest minor version available than to fall-back to the in-tree cloud provider.

> We would also need to change a bit the logic to prevent the cluster from being created in such a case to avoid forgetting about this when we will introduce support for 1.20.

This is the biggest reason for this change. As stated above, it's much better to fall-back to the previous minor version, which is going to work perfectly fine even with a new Kubernetes version. It might happen that it'll take a bunch of time for the OpenStack CCM team to release a minor version to match the latest Kubernetes version and in such a scenario:

- we don't want to block adding new Kubernetes version support
- we want to ensure that users can still deploy OpenStack clusters (i.e. returning an error is not an option)
- we want to ensure that we don't fall-back to the in-tree CCM

irozzo-1A: Sorry, I was not clear enough with my comment. I did not mean that we should fallback to in-tree cloud provider. I meant that we should block creating OpenStack cluster with minor version > `19` with 1.19 CCM. I think it's better to fail fast than succeeding in degraded mode (with old CCM in this case), because we risk missing this again (even if consequencese will be less important). 

I agree with you that returning an error with current logic will result in fall-back to in-tree, and this is definitely not what we want. That's why I also suggested we could handle this with a follow-up PR.

xmudrii: I think that blocking cluster creation is not a good idea. For example, what if it takes weeks before the OpenStack CCM is available? Blocking cluster creation for that long would block both us and users from testing OpenStack with the latest Kubernetes version.

> I think it's better to fail fast than succeeding in degraded mode (with old CCM in this case)

It's not running in a degraded mode. It would have all the features that were available in the latest minor version available (for example 1.19). We have checked with the OpenStack CCM team and SIG Cloud Provider about this and it's totally okay to do that. For example, in KubeOne, we unconditionally deploy OpenStack CCM 1.17.0, and it works flawlessly well for any Kubernetes version.

Of course, if you want the latest features, you have to deploy the latest CCM, but until it's not available, it's much better to use the previous version rather than block users from deploying clusters.

> because we risk missing this again (even if consequencese will be less important).

This can be solved by creating an issue to remind us to add the latest version.

irozzo-1A: Yep, but we have to remind to create the reminder each time :-). What you did is way better than what we have, but as we will be working explicitly on support for k8s `1.20`: having the OpenStack test failing is IMO a good reminder. It will force us to tackle the point. If no CCM will be available at this stage for 1.20 and the 1.19 will be proven to work fine we will use it.

irozzo-1A: That said, I think this is good enough to fix. I'll think about it and maybe create a follow-up PR to propose what I have in mind.

xmudrii: Sounds like a good plan. :+1:

irozzo-1A:
Sorry, wrong button :-)

PR #6273, opened 2020-12-08
Fix OpenStack CCM supported version validation

xmudrii:

Fix OpenStack CCM supported version validation. I've checked the major version instead of the minor version, so CCM doesn't get deployed at all.

```release-note
NONE
```


xrstf (approved):
/approve

PR #6293, opened 2020-12-10
WIP - migrate to v1 for our CRDs

xrstf:
v1beta1 has been deprecated (see https://github.com/kubernetes/kubernetes/issues/82022) and v1 has been available since Kubernetes 1.16.

Velero upstream is still using v1beta1 for compatibility reasons, cert-manager CRDs for v1.0 are using v1, but updating them is a bit more work.

```release-note
Kubermatic CRDs are now using apiextensions/v1.
```

### DIFF
--- a/charts/kubermatic/Chart.yaml
+++ b/charts/kubermatic/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubermatic
-version: 1.1.26
+version: 1.1.30
 appVersion: '__KUBERMATIC_TAG__'
 description: Kubermatic chart for master and/or seed clusters.
 keywords:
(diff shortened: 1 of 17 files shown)


xrstf:
Yeah, this was not as easy as thought. The k8s upstream docs are super confusing, as they do not agree on whether or not it's called `x-kubernetes-preserve-unknown-fields` or `x-preserve-unknown-fields` and where to put it (the doc comments in the Go code are misleading).

Also, in v1 Kubernetes began to protect its own API group: https://github.com/kubernetes/enhancements/pull/1111 -- so we must migrate our CRDs away from kubermatic.k8s.io. A move I very, very much welcome, though it will be a PITA.

PR #6326, opened 2020-12-22
Prevent SA for cluster creation

### LINKED ISSUE #6247: API created account not visible in UI
Problem statement:
Projects created are not visible in the UI, such projects have been created via API account.
Create an API account for the terraform provider and create only ressources - not bound to a user, therefore not visible.
API account should be traceable which should always be a user and it should be visible in the UI.

User Story:

As a user, I only want traceable accounts that can be audited so that I cannot create accounts that could pose a possible security threat.

zreigz:
**What this PR does / why we need it**: Prevent SA for cluster creation

Fixes #6247

```release-note
NONE
```


xrstf:
/override pre-kubermatic-e2e-aws-flatcar-1.17

etcd-launcher tests keep flaking, but are not related to this PR.

PR #6329, opened 2020-12-28
Prevent container orphans when using apiserver.IsRunningWrapper

### LINKED ISSUE #6328: InitContainers are not reconciled in User Clusters
**User Story**
When updating KKP, changes to the init containers (usually used for the copy-http-prober container) on the machine-controller, kube-state-metrics etc. are not properly reconciled. Instead, KKP only ever _adds_ new init containers.

Steps to reproduce:
1. Create a user cluster.
2. Edit for example the machine-controller Deployment and _rename_ the `copy-http-prober` init container to for example `foobar`.
3. Observe that KKP reconciles.
4. Check the final Deployment. You are now left with the `foobar` init container and a newly reconciled `copy-http-prober` init container.

This is most likely caused by the individual Creator() functions never setting `podSpec.initContainers = []`, so when we then call `apiserver.IsRunningWrapper()`, that thing will just add new containers.

Also, _old_ init containers are never removed because of this bug.

**Acceptance criteria**
- KKP properly updates all important pieces of usercluster resources (Deployments, StatefulSets, ...).

xrstf:
When the IsRunningWrapper() function is fiddling with initContainers, we must make sure we only accept and deal with a known state. Otherwise it can happen that old init containers are left behind or copies are never cleaned up.

Fixes #6328

```release-note
[ATTN] Fix orphaned apiserver-is-running initContainers in usercluster controlplane. This can cause a short reconciliation burst to bring older usercluster resources in all Seed clusters up to date. Tune the maxReconcileLimit if needed.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,6 @@ install:
 showenv:
 	@go env
 
-.PHONY: download-gocache
 download-gocache:
 	@./hack/ci/download-gocache.sh
 	@# Prevent this from getting executed multiple times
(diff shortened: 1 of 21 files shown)


kron4eg (commented):
/retest
/lgtm
/approve

PR #6380, opened 2021-01-12
stop syncing ssh keys if there was no keys attached to the user cluster

### LINKED ISSUE #6381: Do not override authorised SSH keys when no keys are provided
**User Story**
KKP is deploying in user clusters an agent (via DaemonSet) that takes care of synchronizing the SSH keys in all user homes found on the worker nodes using a secret as the source.

The agent is deployed even if no SSH keys are present in the secret, this means that the synchronization cannot be opt-out and that SSH keys provided with other means will be overridden.

https://github.com/kubermatic/kubermatic/blob/d652a660ab0176847969f0f6d2d86f1d76c1e6db/pkg/controller/usersshkeysagent/usersshkeys_controller.go#L172-L197

The goal of this task is to skip overriding the keys if no ssh key is present in the secret. This can be achieved by adding a check in the aforementioned logic.
As an optimization, it would also be possible to defer the deployment of the agent DaemonSet until some keys are added.

**Acceptance criteria**

authorized key files are not overridden when no SSH key is added to the user cluster.

moadqassem:
Disable the user-ssh-key-agent when there was no user ssh keys attached to the user cluster. Whatever ssh keys the cluster user would add to machines will remain there until a new user ssh key was attached to the cluster. In that case the ssh keys that were added manually will be overwritten by the ones which were attached.
   
Fixes #6381

**Documentation**:
Documentation for the user ssh keys agent will be added in a different PR

No

```release-note
Disable user ssh key sync when no user ssh keys are added.
```

### DIFF
--- a/pkg/controller/usersshkeysagent/usersshkeys_controller.go
+++ b/pkg/controller/usersshkeysagent/usersshkeys_controller.go
@@ -112,6 +112,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, fmt.Errorf("failed to fetch user ssh keys: %v", err)
 	}
 
+	// if there was no user ssh keys attached to the user cluster, the agent won't carry out any updates,
+	// however, if any ssh keys were attached to the cluster, then those newly attached keys will override any pre-existent ones
+	if len(secret.Data) == 0 {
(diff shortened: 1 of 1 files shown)


xrstf:
@moadqassem do we still need this?

moadqassem:
@xrstf I will update the issue and close it.

PR #6405, opened 2021-01-14
Use cloud init for flatcar on aws

moadqassem:
Flatcar support on aws stopped working due to an increase of the userdata. While using ignition, userdata size hits the limit on aws, however when coreos-cloud-config is used, then the userdata size will be accepted.

Fixes #

This is a temporary fix until the new operating system manager is finished and integrated with KKP, as we would like to keep using ignition as the default cloud-init for flatcar.
**Documentation**:

```release-note
Use Coreos-Cloud-Config for flatcar machines on aws 
```


xrstf (approved):
/approve

Issue #6385, opened 2021-01-12
Add the EnableUserSShKeyAgent field in the API


moadqassem:
**User Story**
As a user, I should be able to disable/enable the user ssh key agent via the KKP API.

**Acceptance Criteria**
-  EnableUserSShKeyAgent should be add in the API [Cluster spec ](https://github.com/kubermatic/kubermatic/blob/master/pkg/api/v1/types.go#L692) 
- The field cannot be edited once the cluster is created
- The api would reject any call to the Attach user ssh keys routes

moadqassem:
This has been Done

PR #6426, opened 2021-01-19
Disable/Enable UserSSHKey agent in user clusters

### LINKED ISSUE #6384: Enabling/Disabling the user ssh key agent in the user cluster
**User Story**
As a user I don't want the user ssh keys agent to be deployed in the user cluster, because I would like to manage the access of the worker nodes by myself.

**Acceptance Criteria**
- The User SSH Key agent is deployed by default
- If the agent is not deployed in the user cluster, the ssh keys should not be synced or attached to the machines 
- The user can add a default ssh key during the cluster creation(even if the agent is disabled), however, once the cluster is up and running it is not possible to add a new ssh key using KKP

moadqassem:
Adding the functionality to enable or disable the user ssh key agent, which responsible for syncing the attached ssh keys to worker nodes. The reason behind this, some users would like to take care of configuring the ssh keys on the nodes, directly without any KKP interception.

Fixes kubermatic/kubermatic#6384 kubermatic/kubermatic#6385

**Documentation**:
There is a ticket which responsible for documenting the agent #6430

```release-note
Disable/Enable user ssh key agent
```


moadqassem:
PTAL 
/assign @xrstf 
/assign @maciaszczykm

maciaszczykm:
Looks good to me, @zreigz can you confirm that?

pkg/api/v1/types.go
```diff
@@ -723,6 +723,12 @@ type ClusterSpec struct {
 	// If active the PodNodeSelector admission plugin is configured at the apiserver
 	UsePodNodeSelectorAdmissionPlugin bool `json:"usePodNodeSelectorAdmissionPlugin,omitempty"`
 
+	// EnableUserSSHKeyAgent control whether the UserSSHKeyAgent will be deployed in the user cluster or not.
+	// If it was enabled, the agent will be deployed and used to sync the user ssh keys, that the user attach
+	// to the created cluster. Otherwise, there will be user ssh keys sync(except the one that's added during cluster
+	// creation will be added) in the cluster. The agent is enabled by default
```
xrstf: > Otherwise, there will be user ssh keys sync (except the one that's added during cluster
> creation will be added) in the cluster.

What does this mean?

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -125,6 +125,13 @@ type ClusterSpec struct {
 
 	UsePodSecurityPolicyAdmissionPlugin bool `json:"usePodSecurityPolicyAdmissionPlugin,omitempty"`
 	UsePodNodeSelectorAdmissionPlugin   bool `json:"usePodNodeSelectorAdmissionPlugin,omitempty"`
+
+	// EnableUserSSHKeyAgent control whether the UserSSHKeyAgent will be deployed in the user cluster or not.
+	// If it was enabled, the agent will be deployed and used to sync the user ssh keys, that the user attach
+	// to the created cluster. Otherwise, there will be user ssh keys sync(except the one that's added during cluster
+	// creation will be added) in the cluster.
```
xrstf: Same question regarding the wording.

pkg/resources/usercluster/deployment.go
```diff
@@ -131,6 +131,7 @@ func DeploymentCreator(data userclusterControllerData, openshift bool) reconcili
 				"-version", data.Cluster().Spec.Version.String(),
 				"-cloud-provider-name", data.GetKubernetesCloudProviderName(),
 				"-owner-email", data.Cluster().Status.UserEmail,
+				fmt.Sprintf("-enable-ssh-key-agent=%t", data.Cluster().Spec.EnableUserSSHKeyAgent),
```
xrstf, line 134: Does this work for nil values?

pkg/api/v1/types.go
```diff
@@ -1699,4 +1707,5 @@ func ToInternalClusterType(externalClusterType string) kubermaticv1.ClusterType
 
 const (
 	InitialMachineDeploymentRequestAnnotation = "kubermatic.io/initial-machinedeployment-request"
+	UserSSHKeyAgentEnabledAnnotation          = "kubermatic.io/user-ssh-key-agent-enabled"
```
xrstf: What is this used for?

xrstf (commented):
Some questions

pkg/api/v1/types.go
moadqassem: by sync:
> the agent will be deployed and used to sync the user ssh keys, that the user attach to the created cluster.
as mentioned in the top 

If this wasn't clear enough, I can adjust it :)

pkg/crd/kubermatic/v1/cluster.go
moadqassem: Same answer as above.

pkg/resources/usercluster/deployment.go
moadqassem: Well, if the field in the API Cluster is nil, then this field will be defaulted to true:
https://github.com/kubermatic/kubermatic/pull/6426/files#diff-27c41fe06286392f13c14c7d60150fdb798847490bc8122cad718e44b810a589R181-R186

pkg/api/v1/types.go
moadqassem: Oh this should be removed, I was thinking of using the annotations(temporarily) to avoid `EnableUserSSHKeyAgent` being updated, but since I am gonna use the webhook instead, I will get rid of it

pkg/api/v1/types.go
xrstf: To me it reads "If you enable this, we deploy the agent and sync the keys. Otherwise, there will be user key sync [the exact same thing?!].", so yes, please clarify the wording :-)

zreigz:
code looks good
/approve

xrstf (approved):
/approve

xrstf:

This is only related to our vcenter and has nothing to do with this PR.

PR #6435, opened 2021-01-20
[Installer] Support setting up seed clusters

### LINKED ISSUE #5870: Kubermatic Installer: Install Seeds
**User Story**
While the Kubermatic Operator takes care of setting up the nodeport-proxy, the seed-ctrlmgr and Seed Validation Webhook, all the other components required for a full Seed are not:

* Kubermatic CRDs
* `kubermatic-fast` StorageClass
* Minio, s3-exporter Helm charts

The installer should provide a command to prepare a cluster to become a seed cluster by installing the components listed above. The installer can, if it wants, maybe, also apply the Seed resource on the master cluster, but maybe not, not sure.

**Acceptance criteria**
- [ ] The installer provides a mechanism to provision seed clusters.
- [ ] The installer provides hints on how to configure DNS for the seed, identically to how it does it already for the master cluster.

xrstf:
This PR extends the KKP installer to be able to provision seed clusters. While the KKP operator does most of the heavy lifting for KKP, the installer is still needed for Minio, S3-Exporter etc.

Fixes #5870

```release-note
Add `kubermatic-seed` stack target to the Installer
```

### DIFF
--- a/cmd/kubermatic-installer/cmd_deploy.go
+++ b/cmd/kubermatic-installer/cmd_deploy.go
@@ -32,11 +32,14 @@ import (
 	"k8c.io/kubermatic/v2/pkg/install/stack/kubermatic"
 	"k8c.io/kubermatic/v2/pkg/log"
 	"k8c.io/kubermatic/v2/pkg/util/edition"
+	"k8c.io/kubermatic/v2/pkg/util/yamled"
 	kubermaticversion "k8c.io/kubermatic/v2/pkg/version/kubermatic"
 
 	certmanagerv1alpha2 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha2"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlruntimeconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
@@ -87,9 +90,10 @@ var (
 
 func DeployCommand(logger *logrus.Logger, versions kubermaticversion.Versions) cli.Command {
 	return cli.Command{
-		Name:   "deploy",
-		Usage:  "Installs or upgrades the current installation to the installer's built-in version",
-		Action: DeployAction(logger, versions),
+		Name:      "deploy",
+		Usage:     "Installs or upgrades the current installation to the installer's built-in version",
+		Action:    DeployAction(logger, versions),
+		ArgsUsage: "[STACK=kubermatic-master]",
 		Flags: []cli.Flag{
 			deployForceFlag,
 			deployConfigFlag,
@@ -101,6 +105,16 @@ func DeployCommand(logger *logrus.Logger, versions kubermaticversion.Versions) c
 			deployStorageClassFlag,
 		},
 	}
+}
+
+type deployOptions struct {
+	cli                 *cli.Context
+	subLogger           *logrus.Entry
+	helmClient          helm.Client
+	kubeClient          ctrlruntimeclient.Client
+	helmValues          *yamled.Document
+	kubermaticConfig    *operatorv1alpha1.KubermaticConfiguration
+	rawKubermaticConfig *unstructured.Unstructured
 }
 
 func DeployAction(logger *logrus.Logger, versions kubermaticversion.Versions) cli.ActionFunc {
@@ -140,12 +154,6 @@ func DeployAction(logger *logrus.Logger, versions kubermaticversion.Versions) cl
 		}
 
 		logger.WithFields(fields).Info("🛫 Initializing installer…")
-
-		supportedProviders := kubermatic.SupportedStorageClassProviders()
-		chosenProvider := ctx.String(deployStorageClassFlag.Name)
-		if chosenProvider != "" && !supportedProviders.Has(chosenProvider) {
-			return fmt.Errorf("invalid storage class provider %q given (--%s)", chosenProvider, deployStorageClassFlag.Name)
-		}
 
 		// load config files
 		if len(kubeconfig) == 0 {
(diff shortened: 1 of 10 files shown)


cmd/kubermatic-installer/cmd_deploy.go
```diff
@@ -139,14 +144,25 @@ func DeployAction(logger *logrus.Logger, versions kubermaticversion.Versions) cl
...
+		case "kubermatic-master":
+			fallthrough
 
-		supportedProviders := kubermatic.SupportedStorageClassProviders()
-		chosenProvider := ctx.String(deployStorageClassFlag.Name)
-		if chosenProvider != "" && !supportedProviders.Has(chosenProvider) {
-			return fmt.Errorf("invalid storage class provider %q given (--%s)", chosenProvider, deployStorageClassFlag.Name)
+		default:
+			kubermaticStack = kubermaticmaster.NewStack()
 		}
```
xmudrii: What is a Kubermatic Stack supposed to represent? I'm not sure I understand this, but I have a few questions:

* What if a user provides some random string or doesn't provide any string?
* Is it expected that we always default to the `kubermatic-master` stack?

cmd/kubermatic-installer/cmd_deploy.go
```diff
@@ -139,14 +144,25 @@ func DeployAction(logger *logrus.Logger, versions kubermaticversion.Versions) cl
...
 
+		if kubermaticStack == nil {
+			return fmt.Errorf("unknown stack %q specified", stackName)
+		}
```
xmudrii: Can we ever reach enter this `if`? If we always default to the `kubermatic-master` stack in the switch above, do we need this?

pkg/install/stack/common/validation.go
```diff
@@ -0,0 +1,48 @@
...
+		secret, err := randomString()
+		if err == nil {
+			return fmt.Errorf("%s must be a non-empty secret, for example: %s", path, secret)
+		}
+
+		return fmt.Errorf("%s must be a non-empty secret", path)
+	}
+
+	return nil
+}
```
xmudrii, line 36: What is this supposed to do? If the value is not present, we are going to return an error in any case. Why do we generate a random string?

pkg/install/stack/common/validation.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+)
+
+func ValidateRandomSecret(value string, path string) error {
+	if value == "" {
+		secret, err := randomString()
+		if err == nil {
```
xmudrii, line 28: In which cases this can return an error? This function just generates a random string, it doesn't depend on `value` or `path`, so I don't see in which cases it can fail.

pkg/install/stack/kubermatic-seed/stack.go
```diff
@@ -0,0 +1,260 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	MinioChartName   = "minio"
+	MinioReleaseName = MinioChartName
+	MinioNamespace   = MinioChartName
```
xmudrii, line 46: Is it expected that those two values are the same?

pkg/install/stack/kubermatic-seed/stack.go
```diff
@@ -0,0 +1,260 @@
...
+
+	S3ExporterChartName   = "kube-system"
```
xmudrii: Is it expected that this chart is called `kube-system`?

pkg/install/stack/kubermatic-seed/stack.go
```diff
@@ -0,0 +1,260 @@
...
+	S3ExporterReleaseName = S3ExporterChartName
+	S3ExporterNamespace   = S3ExporterChartName
```
xmudrii: Is it expected that those two values are the same?

pkg/install/stack/kubermatic-seed/stack.go
xrstf: Yes, because the Secret it creates is shared by other components. It's ugly but it's the current state of things.

pkg/install/stack/kubermatic-seed/stack.go
xrstf: I could set both to `"minio"`, but that's not much different from using the same constant.

pkg/install/stack/common/validation.go
xrstf: Erm it's literally defined right below that:

```
func randomString() (string, error) {
	c := 32
	b := make([]byte, c)

	_, err := rand.Read(b) // <-- this can fail
	if err != nil {
		return "", err
	}

	return base64.RawURLEncoding.EncodeToString(b), nil
}
```

pkg/install/stack/common/validation.go
xrstf: To be nice to the user. To prevent us from having to formulate an awkward "Well you can use this random combination of shell utilities, but only on this particular Debian derivate. On Mac use this, and on Windows, oh boy, good luck. Make sure that the secret is between X and Y bytes long, contains at least 3 special characters and is also a palindrome when read in Hebrew." recommendation. Instead, if the user did not configure a secret, we offer them a strong random one right there for copy&pasting.

cmd/kubermatic-installer/cmd_deploy.go
xrstf: A "stack" in installer terms is "a combination of Helm charts meant to provide a common functionatlity". We have

* KKP master
* KKP seed
* monitoring
* logging
* (backup maybe)

stacks, as I see them currently.

W.r.t. to the other questions: That switch statement was a complete brainfart. I fixed it now, the original intent was to just make `./installer deploy` default to `./installer deploy kubermatic-master`.

xmudrii: Okay, that makes sense, thank you for explaining it to me! The switch statement looks okay now.

pkg/install/stack/kubermatic-seed/stack.go
xmudrii: It's okay like this, no need to change anything. I just wanted to double-check that it's correct.

pkg/install/stack/common/validation.go
xmudrii: I've seen this, but what confuses me is why we have `err == nil` instead of `err != nil` on L28. Considering https://github.com/kubermatic/kubermatic/pull/6435#discussion_r571913658, we want this code to generate the random string if there's no string provided, so I think that we should return an error only if that fails.

pkg/install/stack/common/validation.go
xmudrii: https://github.com/kubermatic/kubermatic/pull/6435#discussion_r571940212

pkg/install/stack/common/validation.go
xrstf: Oh. Ooooooohohohooo, I see. :-D

xrstf: No I don't. The current code is correct, the `err == nil` is what I wanted to express there. :)

xmudrii: We've discussed it internally and it makes sense now.

xmudrii (approved):
/lgtm
/approve

Issue #6225, opened 2020-11-30
[Proposal] Migration path for existing user clusters to external CCM and CSI


xmudrii:
**User Story**

Kubernetes has deprecated the in-tree cloud providers responsible for integrating with the cloud provider APIs in favor of external cloud controller managers and CSI drivers/plugins. The removal is supposed to happen in 1.23.

Affected cloud providers are AWS, Azure, GCE, OpenStack, and vSphere.

**Acceptance criteria**

* Explain the background of the change
* Explain the difference between in-tree and external implementations
* Explain the migration path
* Define tasks and efforts

irozzo-1A:
it's merged

Issue #5949, opened 2020-09-25
Cluster version skew rule violation


balous:
**User Story**
I've tried to upgrade user cluster by editing its resource and changing `version: 1.17.9` -> `version: 1.18.6`. I've observed that the control plane pods began to update instantly - apiserver, controller manager and scheduler in parallel.

This approach violates the [Kubernetes version and version skew support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-controller-manager-kube-scheduler-and-cloud-controller-manager) which states that apiserver must not be newer that the other control plane components.

Parallel upgrade leads to a situation that (for example) upgraded CM pod communicates with apiserver pod that is still waiting for upgrade. This can lead to anything between strange errors in logs and momentary cluster misfunction.

**Acceptance criteria**

The solution is to upgrade apiserver first, wait for completion and than proceeding to the other components.

kron4eg:
>which states that apiserver must not be newer that the other control plane components.

I'm confused. Quote from your link:
>`kube-controller-manager`, `kube-scheduler`, and `cloud-controller-manager` must not be newer than the `kube-apiserver`

The exact opposite.

The implementation of the upgrade is done so that all components are reconciled (thus restarted) at once, there are no "newer" / "older" components.

IMHO, this issue is invalid.

balous:
This is because api server and the other control plane components run in multiple instances and rolling update is configured for them. The update eventually comes to situation where a mixture of old and new versions is running - for all of the control plane components.

For example when we update 1.17 -> 1.18 in such a moment following component mix is running:

- apiserver 1.17
- apiserver 1.18
- controller manager 1.17
- controller manager 1.18

It is possible that, upon its startup, the controller manager 1.18 connects to apiserver 1.17 which is still waiting for update. Thus the version skew policy is violated.

Kubernetes are designed so that the cluster is functional even during the update. In order to achieve this the developers defined these rules, designed apiserver to be backwards compatible and they (hopefully :) test such scenarios.

The update must be done i a way that apiserver update is completed before the process advances to updating the rest of the cluster.

scheeles:
Agree with @balous currently we don't have a phase upgrade. 

We had a `MasterUpdatePhase` in the past https://github.com/kubermatic/kubermatic/blob/26ac0519f384ced6eb0d40c52827f296b103db5f/api/pkg/crd/kubermatic/v1/cluster.go#L44

[...]

PR #6437, opened 2021-01-21
Introduce task runner

irozzo-1A:
This PR introduces in the `reconciling` package a utility that can be used to define a graph of tasks and to run it respecting the dependencies.
This will be used for migrations (e.g. #6225) and upgrades (#5949) where the need for defining a state-machine naturally arises.

Fixes #

At the moment the tasks are executed in a single goroutine, but this can evolve in the future to allow some parallelization.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/codegen/reconcile/main.go
+++ b/codegen/reconcile/main.go
@@ -283,7 +283,9 @@ func {{ .APIVersionPrefix }}{{ .ResourceName }}ObjectWrapper(create {{ .APIVersi
 func Reconcile{{ .APIVersionPrefix }}{{ .ResourceNamePlural }}(ctx context.Context, namedGetters []Named{{ .APIVersionPrefix }}{{ .ResourceName }}CreatorGetter, namespace string, client ctrlruntimeclient.Client, objectModifiers ...ObjectModifier) error {
 	for _, get := range namedGetters {
 		name, create := get()
-		return Reconcile{{ .APIVersionPrefix }}{{ .ResourceName }}TaskFn(create, name, namespace, objectModifiers...)(ctx, client)
+		if err := Reconcile{{ .APIVersionPrefix }}{{ .ResourceName }}TaskFn(create, name, namespace, objectModifiers...)(ctx, client); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/codegen/reconcile/main.go
+++ b/codegen/reconcile/main.go
@@ -283,6 +283,16 @@ func {{ .APIVersionPrefix }}{{ .ResourceName }}ObjectWrapper(create {{ .APIVersi
 func Reconcile{{ .APIVersionPrefix }}{{ .ResourceNamePlural }}(ctx context.Context, namedGetters []Named{{ .APIVersionPrefix }}{{ .ResourceName }}CreatorGetter, namespace string, client ctrlruntimeclient.Client, objectModifiers ...ObjectModifier) error {
 	for _, get := range namedGetters {
 		name, create := get()
+		return Reconcile{{ .APIVersionPrefix }}{{ .ResourceName }}TaskFn(create, name, namespace, objectModifiers...)(ctx, client)
+	}
+
+	return nil
+}
+
+// Reconcile{{ .APIVersionPrefix }}{{ .ResourceName }}TaskFn will return a TaskFn to create or update
+// the {{ .ResourceName }} coming from the passed {{ .ResourceName }}Creator.
+func Reconcile{{ .APIVersionPrefix }}{{ .ResourceName }}TaskFn(create {{ .APIVersionPrefix }}{{ .ResourceName }}Creator, name, namespace string, objectModifiers ...ObjectModifier) TaskFn {
+	return func(ctx context.Context, client ctrlruntimeclient.Client) error {
 {{- if .DefaultingFunc }}
 		create = {{ .DefaultingFunc }}(create)
 {{- end }}
@@ -295,11 +305,11 @@ func Reconcile{{ .APIVersionPrefix }}{{ .ResourceNamePlural }}(ctx context.Conte
 		}
 
 		if err := EnsureNamedObject(ctx, types.NamespacedName{Namespace: namespace, Name: name}, createObject, client, &{{ .ImportAlias }}.{{ .ResourceName }}{}, {{ .RequiresRecreate}}); err != nil {
-			return fmt.Errorf("failed to ensure {{ .ResourceName }} %s/%s: %v", namespace, name, err)
+			return fmt.Errorf("failed to ensure {{ .APIVersionPrefix }}{{ .ResourceName }} %s/%s: %v", namespace, name, err)
 		}
-	}
 
-	return nil
+		return nil
+	}
 }
 
 `))
--- a/pkg/resources/reconciling/tasks_runner.go
+++ b/pkg/resources/reconciling/tasks_runner.go
@@ -107,7 +107,7 @@ func (tr *TasksGraphBuilder) AddTask(id TaskID, task TaskFn, dependencies ...Tas
 
 // Build creates a nes TaskGraph based on the TaskFn added so far.
 // It returns a non-nil error if some invariants are violated.
-// * All dependencies shoud be present.
+// * All dependencies should be present.
 // * No cycles between tasks.
 func (tr *TasksGraphBuilder) Build() (TasksGraph, error) {
 	tr.mark = make(map[*taskNode]struct{})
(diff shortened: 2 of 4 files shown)


xrstf:
This needs a detailed proposal to explain

* why we need a task runner
* why we're not using something that already exists
* why we're not re-using a 3rd party library
* why this is integrated in the reconciling framework
* how this works exactly

/hold

codegen/reconcile/main.go
```diff
@@ -295,11 +307,11 @@ func Reconcile{{ .APIVersionPrefix }}{{ .ResourceNamePlural }}(ctx context.Conte
...
 
 		if err := EnsureNamedObject(ctx, types.NamespacedName{Namespace: namespace, Name: name}, createObject, client, &{{ .ImportAlias }}.{{ .ResourceName }}{}, {{ .RequiresRecreate}}); err != nil {
-			return fmt.Errorf("failed to ensure {{ .ResourceName }} %s/%s: %v", namespace, name, err)
+			return fmt.Errorf("failed to ensure {{ .APIVersionPrefix }}{{ .ResourceName }} %s/%s: %v", namespace, name, err)
 		}
-	}
 
-	return nil
+		return nil
+	}
```
kron4eg, line 314: Nothing is changed in this file, here except adding a closure and increasing the call-stack, so what's the point of this change?

My guess it's for using the `reconciling.TaskFn`, since generated funcs conform the signature, but they are never used in this case, so I propose to ether use them like they should be used as `reconciling.TaskFn`, or for now simply drop that codegen change.

pkg/resources/reconciling/tasks_runner.go
```diff
@@ -0,0 +1,255 @@
...
+type TaskIDSet map[TaskID]struct{}
+
+// Has returns true if and only if item is contained in the set.
+func (s TaskIDSet) Has(item TaskID) bool {
+	_, contained := s[item]
+	return contained
+}
+
+// TaskFn represents a reconciliation task.
+type TaskFn func(context.Context, ctrlruntimeclient.Client) error
```
kron4eg, line 40:
```suggestion
type TaskFn func(context.Context) error
```

controller-runtime Client is a leaking abstraction that doesn't belong to the DAG library, we can always pass inside AddTask the wrapped in closure anything we want.

kron4eg (changes requested):
I really want to see how you want to use that in real code. 

Important observation: the lib is not thread-safe, maybe it's OK, maybe not, depends on how you're intended to use it.

Please add doc.go file with some description how does it work and how should it be used.

irozzo-1A:
xref: https://github.com/kubermatic/kubermatic/issues/2948
I ended up in this PR, it describes pretty well a use-case I had in mind for this lib, but that I did not mention explicitly at the moment.

pkg/resources/reconciling/tasks_runner.go
irozzo-1A: I agree with your comment in general, and I thought about this but I came up to the conclusion that the client will be probably needed on every task we will run with this lib, which is not really intended to be general-purpose. This could avoid unneeded wrapping in some cases. I will think about this again. Thx for the hint.

codegen/reconcile/main.go
irozzo-1A: This change reduces the impedance mismatch between the functions we use to define the resources and this lib. We can definitely defer this to the PR where this is firstly used, but I thought that this gives a hint on how this will be used.

Note I could have implemented this without adding a function call but in our context the impact is negligible, and I preferred to avoid repetitions.

irozzo-1A:
> Important observation: the lib is not thread-safe, maybe it's OK, maybe not, depends on how you're intended to use it.

I mentioned in the comment of `TasksGraphBuilder` struct that it is not thread-safe. It should not be a problem for the usage I want to do at the moment.
 
> Please add doc.go file with some description how does it work and how should it be used.

I will add some more details.

irozzo-1A:
Find a usage example in this POC https://github.com/irozzo-1A/kubermatic-1/compare/tasks-runner...irozzo-1A:migrate-poc?expand=1, I will better document once the API is stable.

kron4eg:
Here's the idea how to have both, "THE DAG" and reactivity (it requires small changes to the lib). Reactivity in the sense that there will be no blocking the loop and waiting.

Syntactically the construction of the DAG will look exactly the same. The tree of tasks, some of which are "mutating state" to the next decided state, some are just guard barriers to wait until certain conditions are met.

The DAG exists only in-memory and it's intended to be so, for easiness of reading, and changing without need to migrate any persisted data, A.K.A. CRD migrations. But the **key difference** from current implementation is that instead of directly advancing the state-machine from the beginning each and every time (e.g. each `Reconcile()` call) with “run every single task. one by one, slowly progressing trough them, possibly with short timeouts” work-loop, THE DAG should resolve the “current state and the next step (job) to run” from from `metav1.Object` annotations it will be given. The resolve happens by reading annotations and comparing to own DAG definition in memory.

Once _the next step (job)_ to take is found, execute **it and only it** and return control back to controller-runtime loop. 
Examples: 
* If next step (job) is "guard barrier" kind, that need to wait until some condition is met, then "RequeueAfter N seconds if condition is not true" and return. 
* If condition **is met**, update annotations and return (requeue will happen automatically).
* If step (or job) is successful, update annotations and return (requeue will happen automatically).
* If there’re other steps to take, return with Requeue.
* If there is an error, requeue the object with RequeueAfter and return.

That way we will have all in one: nicely visualised, easily to read, reactive DAG.

irozzo-1A:
Thx for the comment @kron4eg. I have the impression that there was some misunderstanding, probably related to the POC,  but AFAIU we are almost on the same page.

> Here's the idea of how to have both, "THE DAG" and reactivity (it requires small changes to the lib). Reactivity in the sense that there will be no blocking the loop and waiting.

Note that the library is completely agnostic in that respect, you can have tasks that wait or not. Referring to the POC note that I'm waiting for some conditions to be met, but this is not necessary for the graph to arrive at completion (in several sync iterations). I agree that in general wait should (and can) be avoided, especially if we're going to use this lib within the `cluster_controller`.
 
> Syntactically the construction of the DAG will look exactly the same. The tree of tasks, some of which are "mutating state" to the next decided state, some are just guard barriers to wait until certain conditions are met.

This is exactly what I had in mind. Some tasks could persist some state change (e.g. add conditions or annotations), some others can act as barriers. A thing I still need to add is the possibility of conditionally add a task i.e. if the condition is not met the NOP task is run instead. This will allow running some tasks only when some conditions are met without skipping children's tasks.
Something like:

```
// Ensure cloud-controllers in KCM deployment.
AddTask("ensure-kas",
    reconciling.ConditionalTask(ensureKasFn)
).
// Check that the rollout of the new KCM pods completed.
AddTask("kas-rollout-completed", // Barrier task
    // The task is replaced with a NOP task if hasUpgradeCompletedContion=true 
    reconciling.ConditionalTask(checkKCMReadyFn, !hasUpgradeCompletedConditionCondition),
    "ensure-kas",
).
// Ensure cloud controllers
AddTask("ensure-kcm",
    ensureKCMFn,
    "kas-rollout-completed",
).
// Ensure scheduler
AddTask("ensure-scheduler",
    ensureSchedulerFn,
    "kas-rollout-completed",
).
...
// Control-plane upgrade completed
AddTask("control-plane-upgrade-completed"
   setpUgradeCompletedConditionFn,
   "ensure-kcm", "ensure-scheduler", ...
).
```

Basically, the conditional task syntax will avoid wrapping conditions in `if (...) { }` statements, and having conditional dependencies (that is also an option to be considered).

> The DAG exists only in-memory and it's intended to be so, for easiness of reading, and changing without need to migrate any persisted data, A.K.A. CRD migrations. But the **key difference** from current implementation is that instead of directly advancing the state-machine from the beginning each and every time (e.g. each `Reconcile()` call) with “run every single task. one by one, slowly progressing trough them, possibly with short timeouts” work-loop, THE DAG should resolve the “current state and the next step (job) to run” from from `metav1.Object` annotations it will be given. The resolve happens by reading annotations and comparing to own DAG definition in memory.

As I mentioned before the idea is to run all the tasks that can be run at each sync loop. Tasks should be idempotent, tasks that should be run only when certain preconditions are met should have barrier tasks as parents or be conditional (as I mentioned before I still need to provide the syntax for this).
 
> Once _the next step (job)_ to take is found, execute **it and only it** and return control back to controller-runtime loop.
> Examples:
> 
> * If next step (job) is "guard barrier" kind, that need to wait until some condition is met, then "RequeueAfter N seconds if condition is not true" and return.
> * If condition **is met**, update annotations and return (requeue will happen automatically).
> * If step (or job) is successful, update annotations and return (requeue will happen automatically).
> * If there’re other steps to take, return with Requeue.
> * If there is an error, requeue the object with RequeueAfter and return.
> 
> That way we will have all in one: nicely visualised, easily to read, reactive DAG.

Again, I think that what you are describing is a possible usage of the library, and it does not require any particular API change, apart from the conditional tasks to avoid breaking the fluency of the API.

When I'll have a more definitive vision about the API I'll formalize it in a proposal.

PR #6301, opened 2020-12-11
Enforce TunnelingExposeStrategy feature gate

irozzo-1A:
**What this PR does / why we need it**:
This PR enforces the `TunnelingExposeStrategy` feature gate. If TunnelingExposeStrategy feature gate is not enabled on the following operations won't be allowed:

- `Seed`s with Tunneling expose strategy as default cannot be created.
- `Cluster`s with Tunneling expose strategy cannot be created.

xref: #4942 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:
Installations performed without the Kubermatic operator are out of scope.

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6438, opened 2021-01-21
Fix Cluster Admission Webhook

xrstf:
This fixes the newly introduced Cluster Admission Webhook (from #6301). In the previous code,

* the webhook creators were (already) defunct and so the KKP Operator never actually reconciled anything anymore,
* the *masterctrl* handled Cluster admissions, even though Clusters are handled by the seedctrlmgr. This was most likely done because the existing seed-validation-webhook and its certificate were re-used and those use the masterctrlmgr when they can (https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/common/resources.go#L329-L332)

With this PR, the following changes have been made:

* cluster- and seed validation webhooks use different target Services now. The cluster validation always happens inside the seedctrlmgr.
* All webhooks share the same certificate, because the masterctrlmgr _can_ do both cluster and seed validation, so it needs a cert that is valid for both endpoints.
* The code for the cluster admission webhook reconciling logic has been moved into the seed-operator package (previously it was in operator/common).
* The deprecated CLI flag names are not used anymore. The `WebhookOpts` struct has been fixed and extended, previously it was not actually able to handle the new CLI flags.

```release-note
NONE
```


xrstf:

vCenter is down.

PR #6443, opened 2021-01-22
Validating ssh key agent deployment

### LINKED ISSUE #6429: Reject enabling/disabling the user ssh key agent on a running cluster
**User Story**
When the user decides to disable or enable the user ssh key agent, there should be a rule added in the kubermatics-cluster validation webhook that would simply rejects any update for that field. The reason behind that is, if the user enable/disable the agent after the cluster creation, unintended behaviour might occur like, previous ssh keys will be gone and removed from the authorized_keys files on enabling.

**Acceptance Criteria**
- User ssh key agent can be enabled/disabled only upon cluster creation 
- The webhook should reject any updates for the field `enableUserSSHKey`

moadqassem:
Adding validation for user ssh key agent enabling. The user ssh key agent can be enabled/disabled only during the creation of the user cluster, once the cluster is created, the agent status cannot be changed.

Fixes kubermatic/kubermatic#6429

```release-note
Disabling/Enabling User SSH Key agent 
```

### DIFF
--- a/pkg/validation/cluster/handler.go
+++ b/pkg/validation/cluster/handler.go
@@ -28,7 +28,10 @@ import (
 	"k8c.io/kubermatic/v2/pkg/features"
 
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
@@ -38,12 +41,14 @@ type AdmissionHandler struct {
 	log      logr.Logger
 	decoder  *admission.Decoder
 	features features.FeatureGate
+	client   client.Client
 }
 
 // NewAdmissionHandler returns a new cluster.AdmissionHandler.
-func NewAdmissionHandler(features features.FeatureGate) *AdmissionHandler {
+func NewAdmissionHandler(client client.Client, features features.FeatureGate) *AdmissionHandler {
 	return &AdmissionHandler{
 		features: features,
+		client:   client,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/validation/cluster/handler.go
```diff
@@ -68,11 +73,18 @@ func (h *AdmissionHandler) Handle(_ context.Context, req webhook.AdmissionReques
...
 		validationErr := h.validateCreateOrUpdate(cluster)
 		if validationErr != nil {
 			h.log.Info("cluster admission failed", "error", validationErr)
 			return webhook.Denied(fmt.Sprintf("cluster validation request %s rejected: %v", req.UID, validationErr))
 		}
+
+		if err := h.rejectEnableUserSSHKeyAgentUpdate(ctx, cluster); err != nil {
+			h.log.Info("cluster admission failed", "error", err)
+			return webhook.Denied(fmt.Sprintf("cluster validation request %s rejected: %v", req.UID, err))
+		}
```
xrstf: Why is this a dedicated function and not part of the existing `validateCreateOrUpdate`?

pkg/validation/cluster/handler.go
```diff
@@ -95,3 +107,26 @@ func (h *AdmissionHandler) validateCreateOrUpdate(c *kubermaticv1.Cluster) error
 func (h *AdmissionHandler) SetupWebhookWithManager(mgr ctrl.Manager) {
 	mgr.GetWebhookServer().Register("/validate-kubermatic-k8s-io-cluster", &webhook.Admission{Handler: h})
 }
+
+func (h *AdmissionHandler) rejectEnableUserSSHKeyAgentUpdate(ctx context.Context, cluster *kubermaticv1.Cluster) error {
```
xrstf: This not just rejects Enablements, but also Disablements. The function could maybe be called `rejectUserSSHKeyAgentChanges`?

pkg/validation/cluster/handler.go
```diff
@@ -95,3 +107,26 @@ func (h *AdmissionHandler) validateCreateOrUpdate(c *kubermaticv1.Cluster) error
...
+	var (
+		oldCluster = &kubermaticv1.Cluster{}
+		nName      = types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}
```
xrstf: Clusters are non-namespaced, do not specifiy a namespace here in the key.

pkg/validation/cluster/handler_test.go
```diff
@@ -47,9 +49,10 @@ func TestHandle(t *testing.T) {
 		req         webhook.AdmissionRequest
 		wantAllowed bool
 		features    features.FeatureGate
+		client      client.Client
 	}{
 		{
-			name: "delete cluster succeess",
+			name: "Delete cluster succeess",
```
xrstf: If you change the capitalization, you can also fix the typo in `succeess' ;-)

xrstf (commented):
A few comments

xrstf (approved):
/approve

PR #6445, opened 2021-01-22
Deploy envoy-agent for tunneling expose strategy

irozzo-1A:
This PR introduces the deployment of the tunneling agent when tunneling expose strategy is used.

* Each agent is creating a dummy interface and assigning the IP chosen by the user with `host` scope through init containers.
* Envoy agent is deployed and configured to bind to such IP.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #6392, #6445

**Special notes for your reviewer**:
This is not intended to be prod ready.

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Add Tunneling expose strategy (tech preview).
```

PR #6446, opened 2021-01-23
Adding Envoy agent to the user cluster side

### LINKED ISSUE #5810: Create an agent on the User cluster for the new expose strategy
**User Story**
The goal of this issue to have an agent (Envoy) running on the User cluster side.

The agent will include an Envoy container that will be configured statically by means of a ConfigMap. It will be configured as TCP proxy with two listeners: one for the KAS, the other for OpenVPN server. The cluster will be configured to use the DNS entry pointing to the nodeport proxy.

The `user-controller-manager` should apply the agent manifests to the user cluster (DaemonSet, ConfigMap).

The ports used by the agent for KAS and OpenVPN server should be configurable.
 
**Acceptance criteria**
Agents are deployed on each node, tunneling the TCP requests using HTTP2 to the right target.

youssefazrak:
This PR is to create the resources for the new Agent that will be used
in the new expose strategy.

Fixes #5810 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -26,6 +26,7 @@ import (
 	controllermanager "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/controller-manager"
 	coredns "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns"
 	dnatcontroller "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/dnat-controller"
+	envoyagent "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/envoy-agent"
 	"k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper"
 	kubestatemetrics "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/kube-state-metrics"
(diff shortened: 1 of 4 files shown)


pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -631,7 +633,11 @@ func (r *reconciler) reconcileDeployments(ctx context.Context) error {
 		return fmt.Errorf("failed to reconcile Deployments in namespace %s: %v", kubernetesdashboard.Namespace, err)
 	}
 
-	kubeSystemCreators := []reconciling.NamedDeploymentCreatorGetter{coredns.DeploymentCreator()}
+	kubeSystemCreators := []reconciling.NamedDeploymentCreatorGetter{
+		coredns.DeploymentCreator(),
+		envoyagent.DeploymentCreator(),
```
irozzo-1A: We should deploy the agent conditionally, only when the cluster uses the tunneling expose strategy.

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -486,6 +487,7 @@ func (r *reconciler) reconcileConfigMaps(ctx context.Context, data reconcileData
 		creators = append(creators, []reconciling.NamedConfigMapCreatorGetter{
 			coredns.ConfigMapCreator(),
 			nodelocaldns.ConfigMapCreator(r.dnsClusterIP),
+			envoyagent.ConfigMapCreator(r.namespace, r.clusterURL.Hostname()),
```
irozzo-1A, line 491: Same here, the config for the agent should be included when the tunneling expose strategy is used only.

irozzo-1A:
Closed in favor of #6445 that includes this.

PR #6448, opened 2021-01-25
Extend global settings for resource quota

### LINKED ISSUE #6370: Extend global settings for resource quota

zreigz:
**What this PR does / why we need it**: Extend global settings for resource quota

Fixes #6370

```release-note
Extend global settings for resource quota
```


pkg/crd/kubermatic/v1/settings.go
```diff
@@ -73,6 +75,19 @@ type CleanupOptions struct {
 	Enforced bool
 }
 
+type MachineDeploymentVMResourceQuota struct {
```
moadqassem, line 78: Shouldn't this follow the same resources handling and types as k8s?

zreigz: no, we are going to use common parameters for all providers such as a number of vCPU and Memory. This is what we have from the provider clients and will be easy to use to filter out VMs

moadqassem: Alright, I mean as long as the user can actually override those then I think we are on the good side here.

pkg/crd/kubermatic/v1/settings.go
```diff
@@ -56,6 +56,8 @@ type SettingSpec struct {
 	RestrictProjectCreation     bool           `json:"restrictProjectCreation"`
 	EnableExternalClusterImport bool           `json:"enableExternalClusterImport"`
 
+	MachineDeploymentVMResourceQuota MachineDeploymentVMResourceQuota `json:"clusterVMResourceQuota"`
```
moadqassem: json tag is different than the field name, is that done on purpose?

zreigz: mistake, thanks

zreigz: done

zreigz:
@moadqassem PTAL

PR #6449, opened 2021-01-25
[Installer] Update Helm releases when values have changed

### LINKED ISSUE #6105: Kubermatic Installer Does not Update Dex Configuration
**User Story**

When trying to update the DEX configuration (e.g. adding an additional IdP) through the kubermatic-installer's `apply` option, no changes are applied to DEX. Using the `--force` flag changes the configuration but breaks certs issued by https://letsencrypt.org. This was observed with the community edition, cluster was bootstrapped by kubeone, Nginx Loadbalancer in front of nodes with ingress-controlers, cluster worked perfectly before trying to make changes :D. Manually changing the DEX configuration through the according HELM chart is not working as HELM complains about a non existing release. 

**Acceptance criteria**

- Changed configurations in the `values.yml` should be applied when running `./kubermatic-installer apply`

### LINKED ISSUE #6434: Reconcile Helm charts instead of force installing them
**User Story**
Currently, admins have to make a choice: Either run the KKP installer normally or run it with `--force`. Without `--force`, Helm charts are _not_ redeployed, as the installer currently has no knowledge of the configuration and cannot determine whether the user reconfigured something.

However, Helm stores the effective configuration in the cluster, so the installer _could_ get that, compare to the current `values.yaml` and then determine "Hey, the user changed a settings, let's update the Helm chart!" without requiring `--force`.

**Acceptance criteria**
- [ ] After changing the configuration for a Helm chart and running the Installer again, the chart will be updated automatically, even without `--force`.

xrstf:
Previously, the installer only used the chart's version number to detect if a deployed Helm release is up to date. While simple, this left much to be desired, as Helm values often change and then an explicit `--force` was requird to actually update a release.

This PR improves the situation by fetching the Helm values from the cluster and then comparing them to what was given to the installer. Note that because we usually use a single `values.yaml` for all charts, this means that a single change will make the installer re-install _all_ charts.

Fixes #6434, fixes #6105

```release-note
KKP Installer will update Helm releases automatically if the values have changed (no need for `--force` in most cases).
```


hack/ci/run-lint.sh
```diff
@@ -1,30 +0,0 @@
...
-### into the codebase.
-
-set -euo pipefail
-
-cd $(dirname $0)/../..
-
-lint_output="$(mktemp)"
-golangci-lint run -v --build-tags "$KUBERMATIC_EDITION" --print-resources-usage ./pkg/... ./cmd/... ./codegen/... 2>&1 | tee $lint_output
-
-if egrep -q 'compilation errors|Packages that do not compile' $lint_output; then
```
xrstf, line 27: This didn't do anything, golang-ci never outputs messages that would match this expression. Also, thanks to `set -e`, if compiling fails golang-ci will return 3 and the grep won't be executed anyway.

kron4eg (approved):

PR #6467, opened 2021-01-29
Relax memory limit for openvpn container

kron4eg:
Some of our customers run into the issue with openvpn container that described in the following post 
https://www.cloudfoundry.org/blog/the-garden-team-and-the-strange-case-of-a-connection-being-reset/

This PS is an attempt to fix the issue, but given the race-condition nature of the issue it may fail too.

Fixes #

**Documentation**:

```release-note
Relax memory limit for openvpn container
```


xrstf (approved):
/approve

PR #6477, opened 2021-02-01
Add Multus-CNI for user clusters.

### LINKED ISSUE #6024: Use Multus-CNI as a meta plug-in.
**User Story**
We are currently deploying Canal CNI for the user clusters.
This can't be changed but also it might limit the scope of operation on those user clusters.

Using a `meta` plug-in (plug-in that calls other plug-ins) like [Multus](https://github.com/intel/multus-cni) will allow PODs to have multiple network interfaces. One primary CNI is used for the "PODs network". 

This brings the following advantages:

- [Live CNI migration](https://www.blog.jetstack.io/blog/cni-migration/) 
- Splitting the control/data plane traffic, mostly applicable to VNF/CNF applications (KubeVirt)
- High-performance networking (using SR-IOV for example)
- Multitenant networks

**Acceptance criteria**

- Multus is acting as a meta plug-in
- A default CNI is up and running (Calico)
- Document upgrades for existing user clusters. TBD: downtime needed?

### LINKED ISSUE #6368: Create Multus-CNI addon

youssefazrak:
Add Multus-CNI for user clusters.
WIP, the description will be updated.

Fixes #6368

```release-note
Multus has been added as an addon. 
```


youssefazrak:
I've removed the configMap.
Users can later install whatever configMap is needed in their clusters and no reconciliation will overwrite the CM.
The `00-multus*` has been used for the daemonSet and the specific argument to prevent the automatic creation of the conflist file. This way the custom CM will be in the first position alphabetically speaking, and thus the kubelet will take it into account.

@irozzo-1A good for review :)

mfranczy:
/lgtm

@youssefazrak don't forget to mention in the documentation about multus CM.

Additionaly I am going to work on the new addons implementation proposal, be aware that they way how you build addons might change a bit.

mfranczy:
@youssefazrak btw do you think that would make sense to allow users to pass CM from the UI?

youssefazrak:
@mfranczy cheers noted. I will create a new issue for the documentation.

Regarding the CM from the UI, I'm not convinced it would bring an advantage. Do you have a specific case where we can benefit from this?

mfranczy:
@youssefazrak it should make users life easier, but with our current addons it might be not so convinient. Ok forget the UI part then.

PR #6489, opened 2021-02-03
credential-xxx-clusterid (sometimes) have empty "project-id" label

furkhat:

In our installation we have multiple secrets with empty project id label. The cluster I recently created had the same issue. This causes errors in machine controller. I added additional check to ensure project id label is not empty.

This is however does not solve the problem, hence I don't completely understand where and how it gets empty project id value.

I hope to identify the root cause of the problem by having discussions here in the PR.

 
Fixes #

**Documentation**:

```release-note
NONE
```


xrstf:
I have edited the PR and fixed the missing release-note.

PR #6501, opened 2021-02-03
Compute CoreDNS version based on Kubernetes version

### LINKED ISSUE #6023: Tailor CoreDNS version to the right K8s version for user clusters
**User Story**
We are currently using CoreDNS 1.3.1 for all our user clusters, without taking into account the Kubernetes version of the cluster.
This is done as a way to simplify the versioning but could introduce issues as some features of CoreDNS are deprecated for example.

**Acceptance criteria**

- Follow for the [official CoreDNS <-> Kubernetes](https://github.com/coredns/deployment/blob/master/kubernetes/CoreDNS-k8s_version.md) version mapping
- CoreDNS is successfully deployed in the user clusters and is using the correct version depending on the Kubernetes version.

kron4eg:
And result to deploy both, in-seed and in-user-cluster CoreDNS

Fixes #6023 

```release-note
CoreDNS version now based on Kubernetes version
```

### DIFF
--- a/cmd/etcd-launcher/Dockerfile
+++ b/cmd/etcd-launcher/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM busybox
+FROM busybox:1.32.1
 LABEL maintainer="support@kubermatic.com"
 
 COPY ./_build/etcd-launcher /
--- a/pkg/controller/user-cluster-controller-manager/resources/controller.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/controller.go
@@ -25,6 +25,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/heptiolabs/healthcheck"
 	"go.uber.org/zap"
 
@@ -96,6 +97,11 @@ func Add(
 		opaIntegration:                opaIntegration,
 		userSSHKeyAgent:               userSSHKeyAgent,
 		versions:                      versions,
+	}
+
+	var err error
+	if r.clusterSemVer, err = semver.NewVersion(r.version); err != nil {
+		return err
 	}
 
 	if r.openshift {
@@ -228,6 +234,7 @@ type reconciler struct {
 	seedClient                    client.Client
 	openshift                     bool
 	version                       string
+	clusterSemVer                 *semver.Version
 	cache                         cache.Cache
 	namespace                     string
 	clusterURL                    *url.URL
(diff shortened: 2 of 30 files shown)


multi-io:
While the cluster dns was still an addon we upgraded it to coredns 1.7.0 -- for all K8s versions. That's what e.g. kubeadm uses too, and we never had problems with that (we figured that the list at https://github.com/coredns/deployment/blob/master/kubernetes/CoreDNS-k8s_version.md may be primarily a documentation of history, not of compatibility).

In the Corefile we also removed `upstream` (which you've done in the PR now too -- but will that work with coredns 1.3.1?) and replaced `proxy . /etc/resolv.conf` with `forward . /etc/resolv.conf` because I think the proxy plugin is no longer supported in later coredns versions. Also, the `cluster.local` name shouldn't be hardcoded but set to the value of `cluster.spec.clusterNetwork.dnsDomain` if the user overrode that. -- which would require the usercluster-controller-manager to access the cluster resource, which currently isn't supported AFAIK, so for now I guess I'll be passing dnsDomain as an additional optional command line parameter to the user cluster controller manager if the user overrode it. We can also make a PR for this.

kron4eg:
This is what kubeadm do.

https://github.com/kubernetes/kubernetes/blob/v1.17.0/cmd/kubeadm/app/constants/constants.go#L337
```
CoreDNSVersion = "1.6.5"
```

https://github.com/kubernetes/kubernetes/blob/v1.18.0/cmd/kubeadm/app/constants/constants.go#L338
```
CoreDNSVersion = "1.6.7"
```

https://github.com/kubernetes/kubernetes/blob/v1.19.0/cmd/kubeadm/app/constants/constants.go#L332
```
CoreDNSVersion = "1.7.0"
```

https://github.com/kubernetes/kubernetes/blob/v1.20.0/cmd/kubeadm/app/constants/constants.go#L350
```
CoreDNSVersion = "1.7.0"
```

https://github.com/kubernetes/kubernetes/blob/v1.21.0-alpha.2/cmd/kubeadm/app/constants/constants.go#L350
```
CoreDNSVersion = "v1.8.0"
```

kron4eg:
> but will that work with coredns 1.3.1

Yes, because there will be no coredns 1.3.1 :) Oldest version will be 1.6.5

kron4eg:
> and replaced `proxy . /etc/resolv.conf` with `forward . /etc/resolv.conf` because I think the proxy plugin is no longer supported in later coredns versions.

Yeah, replaced that too.

youssefazrak (approved):
/lgtm

youssefazrak:
@kron4eg @multi-io  FYI `upstream` is the default behavior since CoreDNS 1.4 and later. Before those versions, this is indeed needed.
Regarding the `proxy`, Agree here, deprecated this 1.5.0 (https://coredns.io/2019/04/06/coredns-1.5.0-release/)

PR #6525, opened 2021-02-05
use kubermatic spec for kubermatic CT to make it easier for extension

lsviben:
Changes Kubermatic Constraint Template to use its own Spec instead of the Gatekeeper one to make it easier to extend. 
As the yaml structure stays the same, no changes should be needed for the UI, its only a backend change.

```release-note
NONE
```


xrstf (approved):
In this case I very much welcome less dependency on OPA upstream, as it seems to be gradding behind w.r.t to its dependencies.

/approve

Issue #6309, opened 2020-12-15
[Openstack] Add option to specify use-octavia flag in cloud-config


irozzo-1A:
**User Story**
When using the external OpenStack CCM, the Octavia API is used by default.

On the other hand, the in-tree cloud provider for OpenStack relies on the Neutron API (deprecated).
In order not to have problems during the migration between in-tree cloud-provider to external CCM it is important to provide the possibility of disabling the use of Octavia in order to minimize the risk, as the OpenStack may not support it. 

https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer

**Acceptance criteria**
It is possible to disable the use of Octavia for LB in a user cluster.

Happy2C0de:
Hi all
Thanks for this issue, this is great! 
There are other Openstack cloud.conf options not configurable via the Seed configuration.
ATM, one can change `manage-security-groups` but e.g. use-octavia are not configurable via the Seed k8s resource. Especially, the `floating-network-id` and `subnet-id` is not configured based on the network from the user cluster. This is different in the KubeOne project where the whole cloud.conf can be defined.

There are Openstack clusters that don't use Octavia yet because the Neutron LBaaS was supported until Openstack Train.
What's the reason behind this? 
Or is there a possibility to change the cloud.conf after user-cluster creation? (even though this is bad user-experience) The reconciliation (I think by the seed-controller!?), pushes the configuration if one changes it.

Additionally, I opened another issue about Application Credential support for Openstack.
https://github.com/kubermatic/kubermatic/issues/6527

irozzo-1A:
Hi @Happy2C0de, thx for your comment.

> There are Openstack clusters that don't use Octavia yet because the Neutron LBaaS was supported until Openstack Train.
> What's the reason behind this?

Yeah, the main reason for this is supporting Openstack installation where Octavia API is not yet available, and to reduce the risk during migrations between in-tree cloud provider to CCM, as the first one uses Neutron API as a default, while the second relies on Octavia.

> Or is there a possibility to change the cloud.conf after user-cluster creation? (even though this is bad user-experience) The reconciliation (I think by the seed-controller!?), pushes the configuration if one changes it.

Actually changing `cloud.conf` after the cluster creation is not feasible as it requires 'pausing' the cluster.

PR #6529, opened 2021-02-07
Add support for use-octavia in openstack-providerspecs

Happy2C0de:

Add support for the UseOctavia setting in Openstack provider specs in a given datacenter.

Is related to https://github.com/kubermatic/kubermatic/issues/6309

Please review the given PR and run it against your test infra.

**Documentation**:

Special docs for Kubermatic are pending.
Cloud-provider-openstack docs can be found in the official docs:
https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer

```release-note
Add support for "use-octavia" setting in Openstack provider specs. It defaults to "true" but leaves the possibility to set it to "false" if your provider doesn't support Octavia yet but Neutron LBaaSv2
```

### DIFF
--- a/docs/zz_generated.seed.yaml
+++ b/docs/zz_generated.seed.yaml
@@ -139,8 +139,6 @@ spec:
           # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
           # This setting defaults to true.
           manage_security_groups: true
-          # Optional: Gets mapped to "use-octavia" setting in the cloud config. This setting defaults to true.
-          use_octavia: true
           node_size_requirements:
             # MinimumMemory is the minimum required amount of memory, measured in MB
             minimum_memory: 0
@@ -151,6 +149,10 @@ spec:
           # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
           # This setting defaults to false.
           trust_device_path: false
+          # Optional: Gets mapped to the "use-octavia" setting in the cloud config.
+          # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
+          # This setting defaults to true.
+          use_octavia: true
         packet:
           # The list of enabled facilities, for example "ams1", for a full list of available
           # facilities see https://support.packet.com/kb/articles/data-centers
--- a/docs/zz_generated.seed.yaml
+++ b/docs/zz_generated.seed.yaml
@@ -139,6 +139,8 @@ spec:
           # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
           # This setting defaults to true.
           manage_security_groups: true
+          # Optional: Gets mapped to "use-octavia" setting in the cloud config. This setting defaults to true.
+          use_octavia: true
           node_size_requirements:
             # MinimumMemory is the minimum required amount of memory, measured in MB
             minimum_memory: 0
--- a/docs/zz_generated.seed.yaml
+++ b/docs/zz_generated.seed.yaml
@@ -150,8 +150,8 @@ spec:
           # This setting defaults to false.
           trust_device_path: false
           # Optional: Gets mapped to the "use-octavia" setting in the cloud config.
-          # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
-          # This setting defaults to true.
+          # use-octavia is enabled by default in CCM since v1.17.0, and disabled by
+          # default with the in-tree cloud provider.
           use_octavia: true
         packet:
           # The list of enabled facilities, for example "ams1", for a full list of available
--- a/pkg/crd/kubermatic/v1/datacenter.go
+++ b/pkg/crd/kubermatic/v1/datacenter.go
@@ -206,6 +206,10 @@ type DatacenterSpecOpenstack struct {
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
 	// This setting defaults to true.
 	ManageSecurityGroups *bool `json:"manage_security_groups"`
+	// Optional: Gets mapped to the "use-octavia" setting in the cloud config.
+	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
+	// This setting defaults to true.
+	UseOctavia *bool `json:"use_octavia"`
 	// Optional: Gets mapped to the "trust-device-path" setting in the cloud config.
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
 	// This setting defaults to false.
--- a/pkg/crd/kubermatic/v1/datacenter.go
+++ b/pkg/crd/kubermatic/v1/datacenter.go
@@ -208,7 +208,7 @@ type DatacenterSpecOpenstack struct {
 	ManageSecurityGroups *bool `json:"manage_security_groups"`
 	// Optional: Gets mapped to the "use-octavia" setting in the cloud config.
 	// use-octavia is enabled by default in CCM since v1.17.0, and disabled by
-	// default with the in-tree cloud provider. This setting defaults to true.
+	// default with the in-tree cloud provider.
 	UseOctavia *bool `json:"use_octavia"`
 	// Optional: Gets mapped to the "trust-device-path" setting in the cloud config.
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
--- a/pkg/crd/kubermatic/v1/datacenter.go
+++ b/pkg/crd/kubermatic/v1/datacenter.go
@@ -207,8 +207,8 @@ type DatacenterSpecOpenstack struct {
 	// This setting defaults to true.
 	ManageSecurityGroups *bool `json:"manage_security_groups"`
 	// Optional: Gets mapped to the "use-octavia" setting in the cloud config.
-	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
-	// This setting defaults to true.
+	// use-octavia is enabled by default in CCM since v1.17.0, and disabled by
+	// default with the in-tree cloud provider. This setting defaults to true.
 	UseOctavia *bool `json:"use_octavia"`
 	// Optional: Gets mapped to the "trust-device-path" setting in the cloud config.
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
--- a/pkg/resources/cloudconfig/configmap.go
+++ b/pkg/resources/cloudconfig/configmap.go
@@ -126,6 +126,7 @@ func CloudConfig(
 
 	case cloud.Openstack != nil:
 		manageSecurityGroups := dc.Spec.Openstack.ManageSecurityGroups
+		useOctavia := dc.Spec.Openstack.UseOctavia
 		trustDevicePath := dc.Spec.Openstack.TrustDevicePath
 		openstackCloudConfig := &openstack.CloudConfig{
 			Global: openstack.GlobalOpts{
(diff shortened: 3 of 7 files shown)


Happy2C0de:
This is an initial PR to make the Openstack cloud.conf more configurable via Kubermatic.
The current cloud.conf modifications are very limited and require more changes. Especially, the `LoadBalancer` and `Block Storage` requires more fine-grained modifications. E.g. `node-volume-attach-limit` (https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#block-storage) is required to let k8s know the maximal possible pvc mounted to one worker node. This is not possible at the moment and it leads to problems that are not easy to debug for the user.
Furthermore, the `floating-network-id` and `subnet-id` in LoadBalancer section (https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer) are required to get a working CCM that supports LoadBalancer service resource. These UUIDs depend on the user cluster and have to be fetched dynamically.

/assign @kgroschoff @youssefazrak

irozzo-1A:
Hi @Happy2C0de,

Thx for contributing! I'll review your PR ASAP.

Cheers,
Iacopo

pkg/resources/cloudconfig/configmap.go
```diff
@@ -144,6 +145,7 @@ func CloudConfig(
 			},
 			LoadBalancer: openstack.LoadBalancerOpts{
 				ManageSecurityGroups: manageSecurityGroups == nil || *manageSecurityGroups,
+				UseOctavia: useOctavia == nil || *useOctavia,
```
irozzo-1A:
```suggestion
				UseOctavia: useOctavia,
```

`UseOctavia` expects a bool pointer ;-). You can avoid declaring `useOctavia` and use `dc.Spec.Openstack.UseOctavia` directly.

docs/zz_generated.seed.yaml
```diff
@@ -139,6 +139,8 @@ spec:
           # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
           # This setting defaults to true.
           manage_security_groups: true
+          # Optional: Gets mapped to "use-octavia" setting in the cloud config. This setting defaults to true.
```
irozzo-1A: Did you update this by running `hack/update-docs.sh`? I'm asking because the validation seems to fail.

pkg/resources/cloudconfig/configmap.go
irozzo-1A: I would also add a test case in `pkg/resources/cloudconfig/configmap_test.go`.

irozzo-1A:
@Happy2C0de overall LGTM, just a few things to adjust ;-)

pkg/resources/cloudconfig/configmap.go
Happy2C0de: I see, I updated the code accordingly. I am not sure if it makes sense to have
```
	ManageSecurityGroups bool         `gcfg:"manage-security-groups"`
	UseOctavia           *bool        `gcfg:"use-octavia"`
```
in the machine-controller. See: https://github.com/kubermatic/machine-controller/blob/af516ec83c7f27c930462ea708275cc58294cb57/pkg/cloudprovider/provider/openstack/types/cloudconfig.go#L60
Mainly, `ManageSecurityGroups` is a bool and `UseOctavia` is a boolptr?

docs/zz_generated.seed.yaml
Happy2C0de: No, I didn't. Where is this documented?
The new push includes the corrected docs.

pkg/resources/cloudconfig/configmap.go
Happy2C0de: Afaik, there is only `TestGetVsphereCloudConfig` and for Openstack it would require a complete new Test. This is somehow out of scope for this tiny change, isn't it?
But, I agree there should be a `TestGetOpenstackCloudConfig`!

Happy2C0de:
@irozzo-1A 
Can you review the fixed line and verify my comments?

Mainly, it would be great to get something like https://github.com/kubermatic/machine-controller/blob/af516ec83c7f27c930462ea708275cc58294cb57/pkg/cloudprovider/provider/openstack/provider.go#L751
also in the kubermatic. ATM, all config options are top-level in the seed.yaml, this is not ideal. What about introducing:

```
openstack:
  images: XXX
  ...
  cloudconfig:
    global:
      auth_url: XXX
      ...
    loadbalancer:
      useOctavia: XXX
      manageSecurityGroups: XXX
      ...
    blockstorage:
      node-volume-attach-limit: XXX
      ...
```
This would be far more transparent and follows the implementation in the backend/machine-controller.

pkg/resources/cloudconfig/configmap.go
irozzo-1A: > Mainly, ManageSecurityGroups is a bool and UseOctavia is a boolptr?

The reason why UseOctavia is a pointer is that for OpenStack we still have old clusters using the in-tree cloud provider and newer ones using CCM and CSI. In the first case "use-octavia" is `false` by default, in the second is `true`. Having a pointer gives the possibility to omit it and keep using the respective default unless explicitly specified.
I should have put a commit to make this explicit, thx for pointing out.

Happy2C0de: Thanks for the explanation! That makes sense and I wasn't aware of that. 👍 Good point! :)

docs/zz_generated.seed.yaml
irozzo-1A: Good question :-), this codebase used to be closed source and we are aware we have room for improvement regarding contributors experience.  I'll open an issue to write a contributing guide ;-)

pkg/resources/cloudconfig/configmap.go
irozzo-1A: Yep, you are right. What I would recommend is to:
* Rename the test to `TestCloudConfig`
* Call the function `CloudConfig` instead of `getVsphereCloudConfig`.
* Add the testcase for openstack.
It should not be too complicated unless I missed something, it would be great if you can give it a try!

irozzo-1A:
> @irozzo-1A
> Can you review the fixed line and verify my comments?
> 
> Mainly, it would be great to get something like https://github.com/kubermatic/machine-controller/blob/af516ec83c7f27c930462ea708275cc58294cb57/pkg/cloudprovider/provider/openstack/provider.go#L751
> also in the kubermatic. ATM, all config options are top-level in the seed.yaml, this is not ideal. What about introducing:
> 
> ```
> openstack:
>   images: XXX
>   ...
>   cloudconfig:
>     global:
>       auth_url: XXX
>       ...
>     loadbalancer:
>       useOctavia: XXX
>       manageSecurityGroups: XXX
>       ...
>     blockstorage:
>       node-volume-attach-limit: XXX
>       ...
> ```
> 
> This would be far more transparent and follows the implementation in the backend/machine-controller.

Your point is legitimate and I think that reflecting the structure of the cloud-config would have advantages. But take into account that those structures are evolving especially for some providers. I'm not sure it would be easy to have a stable API. What do you think @zreigz?

zreigz:
All changes for our internal structures cause huge complications (backward compatibility, the stability of controllers). It is problematic from the API perspective but also from the client perspective when migration is needed and we want to avoid this.

Happy2C0de:
Thanks for the addition @zreigz 
I understand your points, and agree, it would have more implications.
It would just be nice to have the same structure in seed.yaml as in the machine-controller. Otherwise, this leads to regular mismatches (not only related to Openstack but in general) when transferring information from the KKP yamls to the other repos. But that's another topic and you have more background there. 👍

docs/zz_generated.seed.yaml
irozzo-1A: BTW I forgot to mention, but you have also to run `hack/update-swagger.sh` ;-)

Happy2C0de: I noticed already, see last commit.
But, `./hack/verify-api-client.sh` is also required ;-) Would be better to call `make verify` directly, but one has to know that.

irozzo-1A: Indeed, I'll put all this in the issue as promised ;-)

Happy2C0de: 👍

pkg/resources/cloudconfig/configmap.go
Happy2C0de: About this change, this requires a bit more effort I guess.
The CloudConfig function returns a string what is really helpful. I would move this to a new issue where it makes sense to have CloudConfig using individual `getXCloudConfig` functions. e.g. `getOpenstackCloudConfig`. This allows to simplify the CloudConfig function and gives the flexibility to use the CloudConfig from other providers within `configmap_test.go`.
What do you think?

irozzo-1A: Yes, what I would do is to have the test verifying the marshaled version of the cloud-config instead, I agree this requires some more effort. I agree this can be tackled in a follow-up PR.

Happy2C0de:
@irozzo-1A 
I think, all tests look good now. Thanks for your work!
I added a follow up Issue that tracks the change of the `configmap_test.go`.

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -206,6 +206,10 @@ type DatacenterSpecOpenstack struct {
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
 	// This setting defaults to true.
 	ManageSecurityGroups *bool `json:"manage_security_groups"`
+	// Optional: Gets mapped to the "use-octavia" setting in the cloud config.
+	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
+	// This setting defaults to true.
```
irozzo-1A: nit: sorry I overlooked this, the link seems to be broken and as we discussed the setting defaults to true only when CCM is used:

https://github.com/kubernetes/cloud-provider-openstack/blob/e5e97a92de1cb69d503bf9b4891405ce5f077efd/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer

I would remove the link and mention that, WDYT?

irozzo-1A:
> @irozzo-1A
> I think, all tests look good now. Thanks for your work!
> I added a follow up Issue that tracks the change of the `configmap_test.go`.

@Happy2C0de thx to you!
Just a small comment regarding a comment :-) and it's ready to go.

Happy2C0de:
I resolved the comment, I hope it's fine now.
/retest

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -206,6 +206,10 @@ type DatacenterSpecOpenstack struct {
 	// See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancer
 	// This setting defaults to true.
 	ManageSecurityGroups *bool `json:"manage_security_groups"`
+	// Optional: Gets mapped to the "use-octavia" setting in the cloud config.
+	// use-octavia is enabled by default in CCM since v1.17.0, and disabled by
+	// default with the in-tree cloud provider. This setting defaults to true.
```
irozzo-1A: Looks better, but I would remove `This setting defaults to true.` if you don't mind as AFIK no default is given.

docs/zz_generated.seed.yaml
```diff
@@ -149,6 +149,10 @@ spec:
           # See https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#block-storage
           # This setting defaults to false.
           trust_device_path: false
+          # Optional: Gets mapped to the "use-octavia" setting in the cloud config.
```
irozzo-1A, line 152: Probably you forgot `hack/update-docs.sh` this time ;-)

Happy2C0de: Yes, I am on it. Absolutely a pain! ;-)
`make update` would be nice. `make verify` needs then a `verify-docs.sh`, or so. :)

irozzo-1A: There is room for improvement :-) don't hesitate to open issues or create PRs.

irozzo-1A:
Need approval from api and cluster-management sigs

/assign @zreigz
/assign @moadqassem

PR #6533, opened 2021-02-08
Add components override field to set nodeportrange for apiserver

### LINKED ISSUE #6530: Direct node connection without load balancing
### Value Statement
As a Kubermatic customer
I want to use node ports with privileged ports
so that I can connect to my nodes directly instead of using loadbalancers

#### Situation
Our customer A would like to be able to configure the node port range in their clusters. They run a streaming platform. With their own software client to establish the connection to the meeting rooms. The client receives a hostname on the first connection and then establishes a direct connection to the streaming server. The client must use port 443 to accomplish this. Other ports are not allowed.
#### Problem
The regular kubernetes node port range is 30000-32767. Customer A cannot use this port range as corporate firewalls often block these ports. The standard ssl port is commonly open. Using a Cloud load balancer is not productive for customer A. It creates and additional hop which is not necessary.  It also requires the use of snat which is not helpful when running huge amounts of persistent connections.

The customer cannot set the node port range themselves since it is configured in the apiserver which runs in the seed cluster.

#### Solution
We can create a componentsOverride field for setting a custom node port range for a cluster. This would give us as the metakube team the ability to set a custom port range for individual clusters. It could be possible to extend the api in the future to allow the customer to set this field by themselves.
Having its own componentsOverride field gives us the ability to validate the node port range before applying it to the apiserver.
This change has a good change to be accepted upstream as well because there is already a componentsOverride field for EndpointReconcilingDisabled.

#### Alternatives
One alternative would be to try and find an alternative for setting a custom node port range together with Customer A.
It would not be clear however if we would find a good alternative and if Customer A would be satisfied with it.
Another alternative would be to add a generic componentsOverride field for overriding flags for the apiserver.
This would give us the option to change other values with less work in the future but is more error prone and harder to validate.

### Acceptance criteria
We can configure the node port range of the API Server for the customer.
The setting is persisted and will not be overwritten by an update.

phiphi282:
Allows overriding the nodeportrange for the apiserver in
componentsOverride.
If field is left empty nodeport will be set the same way as before.

Fixes #6530

**Documentation**:

```release-note
Add components override field to set nodeportrange for apiserver
```

### DIFF
--- a/pkg/resources/apiserver/deployment.go
+++ b/pkg/resources/apiserver/deployment.go
@@ -243,7 +243,10 @@ func DeploymentCreator(data *resources.TemplateData, enableOIDCAuthentication bo
 }
 
 func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, enableOIDCAuthentication, auditLogEnabled bool) ([]string, error) {
-	overrideFlags := getApiserverOverrideFlags(data)
+	overrideFlags, err := getApiserverOverrideFlags(data)
+	if err != nil {
+		return nil, fmt.Errorf("could not get components override flags: %v", err)
+	}
 
 	cluster := data.Cluster()
 
@@ -406,14 +409,16 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 
 // getApiserverOverrideFlags creates all settings that may be overridden by cluster specific componentsOverrideSettings
 // otherwise global overrides or defaults will be set
-func getApiserverOverrideFlags(data *resources.TemplateData) (settings kubermaticv1.APIServerSettings) {
+func getApiserverOverrideFlags(data *resources.TemplateData) (kubermaticv1.APIServerSettings, error) {
+	settings := kubermaticv1.APIServerSettings{}
 	// nodePortRange section
 	settings.NodePortRange = data.NodePortRange()
 	overrideNodePortRange := data.Cluster().Spec.ComponentsOverride.Apiserver.NodePortRange
 	if overrideNodePortRange != "" {
-		if _, err := net.ParsePortRange(overrideNodePortRange); err == nil {
-			settings.NodePortRange = overrideNodePortRange
+		if _, err := net.ParsePortRange(overrideNodePortRange); err != nil {
+			return kubermaticv1.APIServerSettings{}, fmt.Errorf("node port range format is invalid: %v", err)
 		}
+		settings.NodePortRange = overrideNodePortRange
 	}
 	if settings.NodePortRange == "" {
 		settings.NodePortRange = defaultNodePortRange
@@ -425,7 +430,7 @@ func getApiserverOverrideFlags(data *resources.TemplateData) (settings kubermati
 		settings.EndpointReconcilingDisabled = data.Cluster().Spec.ComponentsOverride.Apiserver.EndpointReconcilingDisabled
 	}
 
-	return
+	return settings, nil
 }
 
 func getVolumeMounts() []corev1.VolumeMount {
--- a/pkg/resources/apiserver/deployment.go
+++ b/pkg/resources/apiserver/deployment.go
@@ -18,6 +18,7 @@ package apiserver
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/net"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


xrstf:
/retest

Transient errors are to be expected, we're cleaning up the AWS account used for testing.

pkg/resources/apiserver/deployment.go
```diff
@@ -18,6 +18,7 @@ package apiserver
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/net"
```
kron4eg: Please move this import to k8s.io import group

kron4eg:
I'd like to discuss an alternative approach to what's described in #6530.

In my view it's already possible and no changes are necessary since there is [hostPort support](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-hostport).

/hold

phiphi282:
Ah yes this might work for our usecase as well. We will definitely try.
You might want to add that to your documentation ;)

phiphi282:
Looks like hostPorts unfortunately aren't enough for us.

We have a use case where we would like to use multiple ports on a nodePort Service to enable rolling Updates.
The hostPort field uses the pod directly which only allows one port to exist.

Additionally it not recommended in the kubernetes docu to use hostPorts and one should use nodePort instead if possible because of similar reasons: https://kubernetes.io/docs/concepts/configuration/overview/#services

So in the end we would still need these changes.

kron4eg:
@moadqassem PTAL

pkg/resources/apiserver/deployment.go
```diff
@@ -410,6 +404,30 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
...
 }
 
+// getApiserverOverrideFlags creates all settings that may be overridden by cluster specific componentsOverrideSettings
+// otherwise global overrides or defaults will be set
+func getApiserverOverrideFlags(data *resources.TemplateData) (settings kubermaticv1.APIServerSettings) {
+	// nodePortRange section
+	settings.NodePortRange = data.NodePortRange()
+	overrideNodePortRange := data.Cluster().Spec.ComponentsOverride.Apiserver.NodePortRange
+	if overrideNodePortRange != "" {
+		if _, err := net.ParsePortRange(overrideNodePortRange); err == nil {
```
moadqassem: Why ignoring the error completely? If the default NodePortRange will be the first fallback, I guess it makes sense at least log the error.

phiphi282: Ah good catch.
Thinking about it, it is probably better to just throw an error here to make it a bit clearer.
And I can't think of a reason why someone would like to have the default on if the custom port range is invalid 🤔

kron4eg: >Thinking about it, it is probably better to just throw an error here to make it a bit clearer.

Than please remove the naked return along the way ;)

moadqassem: @phiphi282  You are returning the error when err == nil :-) 

Edit: 
NVM I saw the old hash

kron4eg (commented):
/lgtm
/test all

pkg/resources/apiserver/deployment.go
```diff
@@ -410,6 +407,32 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
...
+	// nodePortRange section
+	settings.NodePortRange = data.NodePortRange()
+	overrideNodePortRange := data.Cluster().Spec.ComponentsOverride.Apiserver.NodePortRange
+	if overrideNodePortRange != "" {
+		if _, err := net.ParsePortRange(overrideNodePortRange); err != nil {
+			return kubermaticv1.APIServerSettings{}, fmt.Errorf("node port range format is invalid: %v", err)
+		}
+		settings.NodePortRange = overrideNodePortRange
+	}
+	if settings.NodePortRange == "" {
```
moadqassem, line 423: So if the NodePortRange has been specified by the controller, and the overrideNodePortRange was specified by the user, what will happen is, the overriding value will be ignored. IMO if the user specify the NodePortRange via the override field, then this value should have the upper hand, not the default value.

phiphi282: So first the "kubermatic wide" value will be set, which can be overridden with the cluster specific one.
If both of them are empty the default range will be set.

So the order is clusterOverride > kubermaticOverride > default I think

moadqassem: What I meant is, if the "kubermatic wide" node port was empty then the override node port value will be ignored. Even if the overrideNodePortRange was not empty.

phiphi282: Regardless of if the "kubermatic wide" value will be set, the settings.NodePortRange will be set to `overrideNodePortRange` as long as it is not empty and a valid port range in line 421

I am not quite sure what I am overlooking here

moadqassem: Ahh sorry mate, I am the one who overlooked stuff. You are right.

PR #6535, opened 2021-02-09
Add CCM/CSI migration logic

### LINKED ISSUE #5799: [OpenStack] Migration path for existing user clusters to external CCM and CSI
**User Story**
User clusters created with KKP having version => 1.15 make use of out-of-tree CCM/CSI, but existing user clusters still using in-tree cloud provider are not migrated automatically. The goal of this issue is to implement a migration path for such clusters. 

**Acceptance criteria**
A cluster administrator is enabled to migrate a user cluster from in-tree cloud provider to out of tree CCM/CSI with a documented (and possibly automated) procedure.

irozzo-1A:

This PR introduces some logic to migrate existing Openstack clusters to external CCM/CSI.

The migration can be triggered by adding the `externalCloudProvider` feature to the `Cluster` resource in addition to the following annotations:

`csi-migration.k8c.io/migration-needed`
`ccm-migration.k8c.io/migration-needed`

Note that those will be added in the future by a mutating webhook that detects the need for migration.

Find below a brief description of the migration process.

The migration is driven by the two aforementioned annotations plus a cluster condition ( i.e. `ClusterConditionCSIKubeletMigrationCompleted`) that will be added by a controller running on user controller manager, indicating that all CSI migration feature gates are present on all kubelets.

1. The `externalCloudProvider` feature triggers the deployment of CSI drivers.
1. `csi-migration.k8c.io/migration-needed` annotation presence and `ClusterConditionCSIKubeletMigrationCompleted` condition not being `true` triggers the activation of following feature gates: `CSIMigration=true", "CSIMigrationOpenStack=true", "ExpandCSIVolumes=true` are activated on kas, kcm, and mc deployments.
1. `ccm-migration.k8c.io/migration-needed`  and `csi-migration.k8c.io/migration-needed` annotation presence and `ClusterConditionCSIKubeletMigrationCompleted` not being `true` triggers the disabling of cloud controllers on kcm.
1. ccm deployment is not applied until in-tree cloud provider or just cloud controllers are deactivated in kcm.
1. migration is finalized when  `ClusterConditionCSIKubeletMigrationCompleted` is true. At this point in-tree cloud provider is deactivated and `"CSIMigrationComplete=true", "CSIMigrationOpenStackComplete=true"` are added to kas, kcm and mc deployments.

Fixes #5799 

This is still a work in progress, it requires some unit testing. e2e testing will be probably deferred to a follow-up PR.

**Documentation**:

```release-note
CCM/CSI migration
```


pkg/resources/data.go
```diff
@@ -352,15 +358,91 @@ func (d *TemplateData) GetGlobalSecretKeySelectorValue(configVar *providerconfig
...
 	return nil, nil
 }
 
-func GetKubernetesCloudProviderName(cluster *kubermaticv1.Cluster) string {
+func (d *TemplateData) GetCSIMigrationFeatureGates() []string {
+	return GetCSIMigrationFeatureGates(d.Cluster())
+}
+
+// KCMCloudControllersDeactivated return true if the KCM is ready and the
+// cloud-controllers are disabled.
```
irozzo-1A, line 517: make this comment clearer

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -260,7 +260,8 @@ func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuth
 	if data.Cluster().Annotations[kubermaticv1.AnnotationNameClusterAutoscalerEnabled] != "" {
 		deployments = append(deployments, clusterautoscaler.DeploymentCreator(data))
 	}
-	if flag := data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; flag {
+	if data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] &&
+		data.KCMCloudControllersDeactivated() {
```
irozzo-1A: Add condition on `ccm-migration.k8c.io/migration-needed`

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -203,8 +208,14 @@ const (
...
 
-	ReasonClusterUpdateSuccessful = "ClusterUpdateSuccessful"
-	ReasonClusterUpdateInProgress = "ClusterUpdateInProgress"
+	ClusterConditionCCMMigrationCompleted ClusterConditionType = "CCMMigrationCompleted"
+	// This condition is met when all the kubelet flags.
+	ClusterConditionCSIKubeletMigrationCompleted ClusterConditionType = "CSIMigrationKubeletFlags"
+
+	ReasonClusterUpdateSuccessful       = "ClusterUpdateSuccessful"
+	ReasonClusterUpdateInProgress       = "ClusterUpdateInProgress"
+	ReasonClusterCCMMigrationSuccesfull = "ClusterCCMMigrationSuccessful"
```
xmudrii: Nit:
```suggestion
	ReasonClusterCCMMigrationSuccesful = "ClusterCCMMigrationSuccessful"
```

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -46,6 +46,11 @@ const (
 	CredentialPrefix = "credential"
 )
 
+const (
+	CCMMigrationNeededAnnotation = "ccm-migration.k8c.io/migration-needed"
```
moadqassem, line 50: If I understand this correctly, me as a user of this cluster who would like to migrate ccm, all what I do is just add this annotation on the cluster object. I would assume that there is a check in the UI which triggers the process?

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -203,8 +208,14 @@ const (
 
 	ClusterConditionEtcdClusterInitialized ClusterConditionType = "EtcdClusterInitialized"
 
-	ReasonClusterUpdateSuccessful = "ClusterUpdateSuccessful"
-	ReasonClusterUpdateInProgress = "ClusterUpdateInProgress"
+	ClusterConditionCCMMigrationCompleted ClusterConditionType = "CCMMigrationCompleted"
+	// This condition is met when all the kubelet flags.
```
moadqassem: What does this actually mean? can you please elaborate more.

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: good catch @xmudrii, BTW at this stage I'm not using this. I'll either implement or remove it for the moment.

pkg/kubernetes/helper.go
```diff
@@ -79,3 +88,61 @@ func ValidateSecretKeySelector(selector *providerconfig.GlobalSecretKeySelector,
...
+		if err != nil {
+			return false, fmt.Errorf("cannot get the revision of deployment %q: %v", deployment.Name, err)
+		}
+		if revision != deploymentRev {
+			return false, fmt.Errorf("desired revision (%d) is different from the running revision (%d)", revision, deploymentRev)
+		}
+	}
+	if deployment.Generation <= deployment.Status.ObservedGeneration {
+		cond := GetDeploymentCondition(deployment.Status, appsv1.DeploymentProgressing)
+		if cond != nil && cond.Reason == "ProgressDeadlineExceeded" {
```
moadqassem, line 110: I think it makes sense to add the reason: `ProgressDeadlineExceeded` as part of the listed reasons in the pkg/crd/kubermatic/v1/cluster.go file

pkg/resources/data.go
```diff
@@ -380,6 +462,32 @@ func GetKubernetesCloudProviderName(cluster *kubermaticv1.Cluster) string {
 	}
 }
 
+func ExternalColoudProviderEnabled(cluster *kubermaticv1.Cluster) bool {
```
moadqassem:
```suggestion
func ExternalCloudProviderEnabled(cluster *kubermaticv1.Cluster) bool {
```

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: yes, I thought as mentioned in the description that the annotations could be put by a mutating webhook. Alternatively, we could use a validation webhook to make sure the annotations are present if `externalCloudProvider` feature is added to a cluster resource.

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: Good catch, this definitely does not make sense. I'll fix this.

pkg/kubernetes/helper.go
irozzo-1A: This is a reason related to `deployments`, not sure it makes sense to have it for `cluster` resources.

moadqassem:
@irozzo-1A any updates regarding this?

pkg/resources/data.go
```diff
@@ -388,6 +605,32 @@ func GetKubernetesCloudProviderName(cluster *kubermaticv1.Cluster) string {
...
+		// The following feature gates are always enabled when the
+		// 'externalCloudProvider' feature is activated.
+		if cluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] {
+			featureFlags = append(featureFlags, "CSIMigration=true", "CSIMigrationOpenStack=true", "ExpandCSIVolumes=true")
+		}
+		// The CSIMigrationNeededAnnotation is removed when all kubelets have
+		// been migrated.
+		if kubermaticv1helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue) {
+			featureFlags = append(featureFlags, "CSIMigrationOpenStackComplete=true")
+		}
```
xmudrii, line 629: FYI: The `CSIMigration*Complete` feature gates are removed in Kubernetes 1.21.

If someone would upgrade a migrated cluster from 1.20 to 1.21, it could break because the feature gate wouldn't exist. Maybe we should just not enable this feature gate for now. It's not that critical anyways.

irozzo-1A: Great! As it does not seem to be mandatory we can remove it, and avoid rolling out some components for nothing.

xmudrii: Alternatively, considering that we have the cluster object, we can check the Kubernetes version and:
* If the version is < 1.21, use the `CSIMigrationOpenStackComplete` feature gate,
* Otherwise, use the `InTreePluginOpenStackUnregister` feature gate

Reference: https://github.com/kubernetes/kubernetes/pull/98243

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -265,7 +265,12 @@ func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuth
 	if data.Cluster().Annotations[kubermaticv1.AnnotationNameClusterAutoscalerEnabled] != "" {
 		deployments = append(deployments, clusterautoscaler.DeploymentCreator(data))
 	}
-	if flag := data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; flag {
+	// If CCM migration is ongoing defer the deployment of the CCM to the
+	// moment in which cloud controllers or the full in-tree cloud provider
+	// have been deactivated.
+	if data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] &&
+		(!metav1.HasAnnotation(data.Cluster().ObjectMeta, kubermaticv1.CCMMigrationNeededAnnotation) ||
+			data.KCMCloudControllersDeactivated()) {
```
xmudrii, line 273: How's this going to behave in the following scenario:
* The ExternalCloudProvider feature is enabled (the first part of the `if` is `true`)
* There is no `CCMMigrationNeeded` annotation (this will make the whole `if` block `true`)
* The in-tree controllers are still not deactivated (`data.KCMCloudControllersDeactivated()` returns `false`)

In that case, the condition will be satisfied and we would deploy the external CCM, even if the in-tree plugins are still enabled.

I'm not sure if this is something that can actually happen, but if yes, then we might need to reconsider this condition.

I guess that if we implement the mutating webhook, the webhook will edit the Cluster object and apply the annotation before the seed-controller-manager will reconcile it. In that case, this condition wouldn't be satisfied (because the annotation is there) and we'll not deploy the external CCM. However, the question is how we should handle it until we don't implement the webhook?

irozzo-1A: I see your point, but consider that at the moment if the `externalCloudProvider`  feature is added to an existing cluster a hard migration is triggered anyway so this PR is not worsening the situation at least :-)

pkg/resources/data.go
xmudrii: The comment seems clear enough to me.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -46,6 +46,11 @@ const (
 	CredentialPrefix = "credential"
 )
 
+const (
+	CCMMigrationNeededAnnotation = "ccm-migration.k8c.io/migration-needed"
+	CSIMigrationNeededAnnotation = "csi-migration.k8c.io/migration-needed"
```
xmudrii, line 51: What is the lifecycle of those two annotations? Once the migration is fully-complete, is it safe to delete those two annotations, or those should stay for the whole lifecycle of the cluster?

irozzo-1A: yes, they don't harm but that's the idea ;-)

pkg/resources/data.go
```diff
@@ -388,6 +605,32 @@ func GetKubernetesCloudProviderName(cluster *kubermaticv1.Cluster) string {
...
+	// should not disable the in-tree cloud provider until all kubelets are
+	// migrated, otherwise we won't be able to use the volume API.
+	return cluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] &&
+		(kubermaticv1helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue) ||
+			!metav1.HasAnnotation(cluster.ObjectMeta, kubermaticv1.CSIMigrationNeededAnnotation))
+}
+
+func GetCSIMigrationFeatureGates(cluster *kubermaticv1.Cluster) []string {
+	var featureFlags []string
+	if metav1.HasAnnotation(cluster.ObjectMeta, kubermaticv1.CSIMigrationNeededAnnotation) {
```
xmudrii, line 619: Referring to https://github.com/kubermatic/kubermatic/pull/6535/files#r579296816.

If it's supposed to be possible to delete the `CSIMigrationNeeded` annotation, we need an additional condition here for once the annotation is gone. Those feature gates are supposed to be enabled for the whole lifecycle of the cluster, so we should add some condition related to the Cluster conditions (e.g. `CSIKubeletMigrationCompleted`).

pkg/crd/kubermatic/v1/cluster.go
xmudrii: What's the idea actually? That it should be possible to delete those or no? :sweat_smile:

pkg/resources/data.go
irozzo-1A: Well, those annotations are required until there are volumes that are created with the old provisioner, but yes I see your point. The annotations should not be removed until they are needed.

irozzo-1A: And there is nothing planned to remove them automatically if this is what you wanted to know

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -205,9 +210,16 @@ const (
...
+	// This condition is met when a CSI migration is ongoing and the CSI
+	// migration feature gates are activated on the Kubelets of all the nodes.
+	// When this condition is `true` CSIMigration{provider}Complete can be
+	// enabled.
+	ClusterConditionCSIKubeletMigrationCompleted ClusterConditionType = "CSIKubeletMigrationCompleted"
+
+	ReasonClusterUpdateSuccessful             = "ClusterUpdateSuccessful"
+	ReasonClusterUpdateInProgress             = "ClusterUpdateInProgress"
+	ReasonClusterCSIKubeletMigrationCompleted = "CSIKubeletMigrationSuccess"
+	ReasonClusterCCMMigrationInProgress       = "CSIKubeletMigrationInProgress"
```
xmudrii, line 222: Those 3 reasons are not used at all.

irozzo-1A: I introduced the last 2, and it's true they are not used yet. They will be used by: https://github.com/kubermatic/kubermatic/issues/6551

I don't have problems either in keeping or removing them.

xmudrii (approved):
/lgtm
/approve

PR #6538, opened 2021-02-09
Maintain a single CA bundle for all components

### LINKED ISSUE #6469: CA certs are not mounted to the controller-manager
**User Story**

Running Kubermatic 2.15.7 without the external CCM for Openstack, the `controller-manager` "crashloops". After further investigation, the image has been changed from `hyperkube` (2.14) to `kube-controller-manager`. This new image doesn't contain by default CA certs and those are supposed to be mounted from the Host itself. 
In the case of this reported bug, the nodes were running CentOS 7.9. 
This doesn't happen and thus triggers the error below, as the CA is not recognized as a valid root authority: 

`F0127 07:39:37.262651       1 controllermanager.go:230] error building controller context: cloud provider could not be initialized: could not init cloud provider "openstack": Post https://some.website.com: x509: certificate signed by unknown authority
`

**Acceptance criteria**
Have the CA store installed inside the container.

xrstf:
This PR introduces the concept of a "global CA" in KKP. This global CA (bundle) is a collection of CA certificates that we automatically  synchronize from the master cluster into all seeds and from there into all user clusters. It is then used for _all_ places where a custom CA can be configured (i.e. it replaces the old "OIDC CA File"), like calls to cloud provider APIs and other in-cluster CAs. Noteworthy is that it does not _extend_ the default CAs, but replaces them, i.e. we do not mount any certificates from the seed nodes into our pods anymore.

For this, the KubermaticConfiguration was changed to have a new `caBundle` field, which references a ConfigMap (because putting >200KB of PEM stuff into the KubermaticConfiguration CR is just ugly). The old `spec.auth.caBundle` has been removed, as it was too difficult to reconcile it so that it would be _always_ available.

The KKP Operator Helm chart ships a default CA bundle as a `ca-bundle` ConfigMap, so users who never touched CA stuff do not need to care about this change. Users who had previously added an OIDC CA file now need to _append_ it to the ConfigMap.

As of now, the CA is used in the following places:

* etcd backups (communication with the S3 storage that is possibly using a custom CA)
* machine-controller 1.26+ (for its vSphere/Openstack providers)
* KKP API (for talking to Dex, which might use a custom CA)
* kube-apiserver/scheduler/ctrlmgr, as these do not ship with their own CA certificates before 1.19 (1.17 and 1.18 did for example not set `SSL_CERT_FILE`) and without any external CA source, the internal AWS CCM for example is broken

Future improvements can be:

* Allow to configure a CA bundle per Seed. Currently we do it for the KubermaticConfiguration because master components (like the KKP API) need it.

Fixes #6469

**Documentation**:

```release-note
[ACTION REQUIRED] Overhauled CA handling, allow to configure a global CA bundle for every component. The OIDC CA ile has been removed, manual updates can be necessary.
```


xrstf:
/hold

To give both SIGs time for review

cmd/kubermatic-api/options.go
```diff
@@ -91,9 +93,9 @@ func newServerRunOptions() (serverRunOptions, error) {
 	flag.StringVar(&s.presetsFile, "presets", "", "The optional file path for a file containing presets")
 	flag.StringVar(&s.swaggerFile, "swagger", "./cmd/kubermatic-api/swagger.json", "The swagger.json file path")
 	flag.StringVar(&rawAccessibleAddons, "accessible-addons", "", "Comma-separated list of user cluster addons to expose via the API")
+	flag.StringVar(&caBundleFile, "ca-bundle", "", "The path to the certificate for the CA that signed your identity provider’s web certificate.")
```
kron4eg, line 94: Since this flag is mandatory, it might be good to provide some default value... Or a fallback (to read default PEM from some buffer). With fallback much more preferred as it will spread on all binaries where -ca-bundle is used (and is mandatory at the moment).

pkg/controller/operator/master/reconciler.go
```diff
@@ -50,6 +50,7 @@ type Reconciler struct {
 	scheme     *runtime.Scheme
 	workerName string
 	versions   kubermaticversion.Versions
+	namespace  string
```
kron4eg: This `namespace` is not used.

cmd/seed-controller-manager/controllers.go
```diff
@@ -153,13 +153,13 @@ func createKubernetesController(ctrlCtx *controllerContext) error {
...
 		ctrlCtx.runOptions.enableEtcdBackupRestoreController,
 		backupInterval,
-		ctrlCtx.runOptions.oidcCAFile,
 		ctrlCtx.runOptions.oidcIssuerURL,
 		ctrlCtx.runOptions.oidcIssuerClientID,
 		ctrlCtx.runOptions.kubermaticImage,
 		ctrlCtx.runOptions.etcdLauncherImage,
 		ctrlCtx.runOptions.dnatControllerImage,
 		ctrlCtx.runOptions.tunnelingAgentIP.String(),
+		ctrlCtx.runOptions.caBundle,
```
kron4eg, line 162: would be cool to push down the call stack ehter `[]byte` already read from the file (or other sources), or even x509.Pool directly. This is relevant for all controllers that read file on their own.

pkg/controller/seed-controller-manager/etcdbackup/resources.go
```diff
@@ -0,0 +1,43 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func caBundleConfigMapName(cluster *kubermaticv1.Cluster) string {
+	return fmt.Sprintf("cluster-%s-ca-bundle", cluster.Name)
+}
+
+func caBundleConfigMapCreator(name string, caBundle string) reconciling.NamedConfigMapCreatorGetter {
```
kron4eg: This function already defined in `pkg/controller/user-cluster-controller-manager/resources/resources/ca-bundle.ConfigMapCreator()`.

pkg/controller/seed-controller-manager/backup/resources.go
```diff
@@ -14,22 +14,23 @@ See the License for the specific language governing permissions and
...
 )
 
-// DexCACertificateCreator returns a function to create/update the secret with the certificate for TLS verification against dex
-func DexCACertificateCreator(getDexCA func() ([]*x509.Certificate, error)) reconciling.NamedSecretCreatorGetter {
-	return func() (string, reconciling.SecretCreator) {
-		return resources.DexCASecretName, certificates.GetDexCACreator(
-			resources.DexCAFileName,
-			getDexCA,
-		)
+func caBundleConfigMapCreator(name string, caBundle string) reconciling.NamedConfigMapCreatorGetter {
```
kron4eg: This function already defined in `pkg/controller/user-cluster-controller-manager/resources/resources/ca-bundle.ConfigMapCreator()`.

pkg/controller/user-cluster-controller-manager/resources/resources/ca-bundle/configmap.go
```diff
@@ -0,0 +1,38 @@
+/*
```
kron4eg, line 1: why `...resources/resources/ca-bundle`? I mean 2 times of `resources/resources`.

pkg/resources/apiserver/cabundle.go
```diff
@@ -0,0 +1,46 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type caBundleProvider interface {
+	CABundle() (string, error)
+}
+
+func CABundleCreator(data caBundleProvider) reconciling.NamedConfigMapCreatorGetter {
```
kron4eg, line 29: Another duplicate of this function with slight vadiation for `caBundleProvider`.

pkg/resources/apiserver/deployment.go
```diff
@@ -643,22 +651,20 @@ func GetEnvVars(data kubeAPIServerEnvData) ([]corev1.EnvVar, error) {
 	}
 	cluster := data.Cluster()
 
-	var vars []corev1.EnvVar
+	vars := []corev1.EnvVar{
+		// Kubernetes <1.19 did not ship any certificates in their Docker images,
```
kron4eg, line 655: It should be

> Kubernetes >1.19

or maybe

> Kubernetes 1.19+

pkg/resources/scheduler/deployment.go
```diff
@@ -119,10 +119,18 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
...
-					Name:         resources.SchedulerDeploymentName,
-					Image:        data.ImageRegistry(resources.RegistryK8SGCR) + "/kube-scheduler:v" + data.Cluster().Spec.Version.String(),
-					Command:      []string{"/usr/local/bin/kube-scheduler"},
-					Args:         flags,
+					Name:    resources.SchedulerDeploymentName,
+					Image:   data.ImageRegistry(resources.RegistryK8SGCR) + "/kube-scheduler:v" + data.Cluster().Spec.Version.String(),
+					Command: []string{"/usr/local/bin/kube-scheduler"},
+					Args:    flags,
+					Env: []corev1.EnvVar{
+						// Kubernetes <1.19 did not ship any certificates in their Docker images,
```
kron4eg, line 127: Kubernetes 1.19+

kron4eg (commented):
Awesome job!

P.S.
Left few notices for consideration.

cmd/kubermatic-api/options.go
xrstf: Don't you think having the files in the Helm charts is enough? I never optimized our CLI for ease of use :-D (except of course for the installer), but this at least reminds me that this PR needs to adjust the hack/run-* scripts.

cmd/seed-controller-manager/controllers.go
xrstf: In all places where I could, I already did that. But sometimes I would then have to re-encode a CertPool as PEM and this didn't seem better.

pkg/resources/apiserver/deployment.go
xrstf: \>1.19 is the opposite of <1.19 and from my tests, 1.17 and 1.18 failed because they were suddenly missing CAs, but with 1.19 they started shipping CAs.

cmd/seed-controller-manager/controllers.go
kron4eg: If you could pass the pathToFile string, you can pass PEM encoded certs for sure.

xrstf: You mean both?

kron4eg: No, I mean in exchange :)
readFile in main, 1 time, and then pass PEM around in form of string (or even better, x509.Pool)

pkg/controller/user-cluster-controller-manager/resources/resources/ca-bundle/configmap.go
xrstf: Because the controller is badly named ("resources"). It should be `pkg/controller/user-cluster-controller-manager/user-cluster-controller/resources`, but this is not the PR to move a ton of files around. As this piece of code is for a resource that lives inside the usercluster, `resources/resources/ca-bundle` is IMHO the right package.

pkg/resources/apiserver/cabundle.go
xrstf: I see what you mean, but I'm not 100% sure that merging the various implementations into one just to save a few lines of code is the right way to go here. All of these are different, some work of a PEM-encoded CA bundle, some read files.

For example, the user-cltrmgr does not use `resources.TemplateData`, so creating a single CABundleCreatorGetter based on this would not work over there. We would need to have multiple, and we would also need to make the name configurable, because for the backup jobs the name is not simply `ca-bundle`. Currently we have:

* `func CABundleConfigMapCreator(caBundle *corev1.ConfigMap) reconciling.NamedConfigMapCreatorGetter {` because this copies another ConfigMap
* `func CABundleCreator(data caBundleProvider) reconciling.NamedConfigMapCreatorGetter` because this uses TemplateData
* `func caBundleConfigMapCreator(name string, caBundle string) reconciling.NamedConfigMapCreatorGetter` for the etcd backup ConfigMaps

What we end up with is IMHO needlessly genericized code that saves maybe 6 lines of code (plus some boilerplate, so make it 20).

Additionally, this would make later improvements harder, as we tied together things that were purposefully left somewhat separated: the CA bundle for userclusters is by design a different, but similar thing from the CA bundle of the seed.

TL;DR: I don't think it's worth to merge these.

cmd/kubermatic-api/options.go
kron4eg: There is very wonderful function, [x509.SystemCertPool()](https://pkg.go.dev/crypto/x509#SystemCertPool) which can deliver us "the default system CAs" in case when -ca-bundle is not provided.

pkg/controller/user-cluster-controller-manager/resources/resources/ca-bundle/configmap.go
kron4eg: ouch... what a unfortunate naming. OK then 🤷‍♂️

zreigz:
code looks good, I am not able to check the functionality.
/lgtm
/approve

pkg/resources/apiserver/deployment.go
kron4eg: yeah, sorry for confusion 😅

pkg/resources/data.go
```diff
@@ -49,6 +48,13 @@ const (
 	cloudProviderExternalFlag = "external"
 )
 
+type CABundle interface {
+	CertPool() *x509.CertPool
+	String() string
+	Bytes() []byte
+	File() (string, error)
```
kron4eg:
```suggestion
```

I'm pretty sure we can do easy without those two.

kron4eg (approved):

PR #6544, opened 2021-02-11
fixed gatekeeper deployment annotations being overwritten

lsviben:
Fixes the reconcile loop that happens because gatekeeper deployment keeps overwriting its annotations.

```release-note
NONE
```


xrstf (approved):
/approve

PR #6545, opened 2021-02-11
add KubeVirt and Containerized-data importer addons

mfranczy:
Signed-off-by: Marcin Franczyk <marcin0franczyk@gmail.com>

It allows users to run VM workload side by side with containers

```release-note
KubeVirt addon allows to run VMs side by side with containers
```

### DIFF
--- a/addons/kubevirt/cdi-cr.yaml
+++ b/addons/kubevirt/cdi-cr.yaml
@@ -1,4 +1,4 @@
-# Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+# Copyright 2021 The Kubermatic Kubernetes Platform contributors.
 #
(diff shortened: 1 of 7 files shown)


mfranczy:
KubeVirt addon consists of two operators (KubeVirt, CDI) and two CRs, it's important to deploy the operators first.
I would like to rename yaml files from `kubevirt-operator.yaml` to `10-kubevirt-operator.yaml` (the same how udev rules are set - just to force order).

However, before doing that, is there any better way to force the order? Additionaly I think it would be good to reverse manifests order when those are being deleted, is there any known reason why it should not be changed? I can modify this in the PR. 

Thoughts?

mfranczy:
as we agreed on the sig-app meeting, first we improve the addon implementation

Issue #6195, opened 2020-11-24
Allow modification of CoreDNS ConfigMap in user clusters


rastislavs:
**User Story**

More complex k8s networking setups can benefit from extending the DNS resolution within the cluster, e.g. by forwarding specific zones to an external DNS server. An example of that may be a multi-cluster service discovery provided by [Submariner](https://submariner.io/). To do that, the CoreDNS ConfigMap in user clusters has to be modified/extended.

Currently, the content of CoreDNS ConfigMap cannot be modified in the user clusters, since its content is hard-coded in the [User Cluster Controller Manager](https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/configmap.go#L34-L50), and even if the ConfigMap is modified, the controller always reconciles it back to the hard-coded value.

**Acceptance criteria**

- CoreDNS ConfigMap can be modified in user clusters during runtime.

PR #6546, opened 2021-02-11
User cluster CoreDNS fixes and extensions

multi-io:

This brings features that the former dns addon had back into the CoreDNS that's now deployed into the user cluster by the usercluster-controller-manager. It also adds some new things to it. Specifically:

- if `.spec.clusterNetwork.dnsDomain` is overridden on the cluster resource, it gets copied into the dns configuration (`Corefile`) in the user cluster as the domain name for which records are looked up in the K8s api. We implement this by passing the overridden dnsDomain to the user-cluster-controller via an optional command line argument (TODO would it be better if user-cluster-controller grabbed this from the cluster resource directly?)

- we've added a dns-autoscaler deployment

- there's also an extension mechanism that allows the user to provide additional CoreDNS configuration by creating an optional configmap that contains corefiles which will then be used by the main Corefile. This will be used by the normal coredns (kubedns) as well as by the node-local-dns deployment. This may solve https://github.com/kubermatic/kubermatic/issues/6195.

The PR contains 3 commits for easier review -- the dnsDomain transfer is done by the first commit and the other two points by the last two. We can also maybe split this into more than one PR if you want.

Fixes #

**Documentation**:

```release-note
kube-dns supports custom cluster DNS domains as well as optional user-provided additional configuration
```

### DIFF
--- a/cmd/user-cluster-controller-manager/main.go
+++ b/cmd/user-cluster-controller-manager/main.go
@@ -61,6 +61,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
+const (
+	defaultClusterDNSDomain = "cluster.local"
+)
+
 type controllerRunOptions struct {
 	metricsListenAddr             string
 	healthListenAddr              string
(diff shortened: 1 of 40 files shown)


irozzo-1A:
Hi @multi-io, thx for your contribution.

In order to help us in the review process, to understand exactly which needs/problems this PR is addressing, and to have focused discussions I think it would make sense to have dedicated PRs for each of the points you mentioned.

e.g. do we need to scale CoreDNS linearly with the number of nodes (considering we are deploying node-local caches)? Should we have a dedicated auto-scaler? Should the config being hardcoded?

PR #6547, opened 2021-02-11
remove canary-ci-deployment job

xrstf:
The old Helm chart is going to be deleted in this release cycle (i.e. for 2.17), so soon we will not be able to use it for canary deployments anymore. Given that the chart was deprecated, the value in updating our CI cluster was minimal to begin with.

In the future, we should replace this with a kind-based job that uses the KKP Operator / installer instead.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1302,43 +1302,6 @@ presubmits:
             memory: 4Gi
             cpu: 2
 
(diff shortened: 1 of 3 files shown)


xmudrii (approved):
/lgtm

PR #6549, opened 2021-02-12
downgraded gatekeeper webhook admissionreview

lsviben:
Fixes the issue with gatkeeper not recognizing the admissionReview version by downgrading it to v1beta1. 

```release-note
Fix issue with gatkeeper not recongizing the AdmissionReview v1 version by changing the webhook to use v1beta1
```


lsviben:
Going with different approach

PR #6550, opened 2021-02-12
changed gatekeeper webhook admission version to v1beta1

lsviben:
Fix issue with gatekeeper not recognizing the AdmissionReview v1 version by changing the webhook to use v1beta1

```release-note
Fix issue with gatekeeper not recognizing the AdmissionReview v1 version by changing the webhook to use v1beta1
```


xrstf (approved):
/approve

PR #6565, opened 2021-02-16
Add UseOctavia flag in Cluster cloud spec

### LINKED ISSUE #6553: Add generic cloudconfig tests to configmap_test.go
**User Story**
In order to get a clean tests from the cloudconfig configmaps, it's necessary to add tests to the `configmap_test.go` from all providers (not only VSphere).

**Acceptance criteria**
Test run through with the new code structure. Allows to add tests in upcoming PRs.

irozzo-1A:
This PR completes the implementation of #6309 the allowing to set `use-octavia` cloud-config flag at `Cluster` level.

Fixes #6553

**Documentation**:

```release-note
OpenStack: Add support for "use-octavia" setting in Cluster Openstack cloud specs.
```

### DIFF
--- a/pkg/crd/kubermatic/v1/cluster.go
+++ b/pkg/crd/kubermatic/v1/cluster.go
@@ -582,7 +582,8 @@ type OpenstackCloudSpec struct {
 	//
 	// Takes precedence over the 'use_octavia' flag provided at datacenter
 	// level if both are specified.
-	UseOctavia *bool `json:"useOctavia"`
+	// +optional
+	UseOctavia *bool `json:"useOctavia,omitempty"`
 }
 
 // PacketCloudSpec specifies access data to a Packet cloud.
--- a/pkg/crd/kubermatic/v1/cluster.go
+++ b/pkg/crd/kubermatic/v1/cluster.go
@@ -575,6 +575,14 @@ type OpenstackCloudSpec struct {
 	FloatingIPPool string `json:"floatingIpPool"`
 	RouterID       string `json:"routerID"`
 	SubnetID       string `json:"subnetID"`
+	// Whether or not to use Octavia for LoadBalancer type of Service
+	// implementation instead of using Neutron-LBaaS.
+	// Attention:Openstack CCM use Octavia as default load balancer
+	// implementation since v1.17.0
+	//
+	// Takes precedence over the 'use_octavia' flag provided at datacenter
+	// level if both are specified.
+	UseOctavia *bool `json:"useOctavia"`
 }
 
 // PacketCloudSpec specifies access data to a Packet cloud.
--- a/pkg/resources/cloudconfig/configmap_test.go
+++ b/pkg/resources/cloudconfig/configmap_test.go
@@ -30,15 +30,15 @@ import (
 	"k8c.io/kubermatic/v2/pkg/semver"
 )
 
-func TestCloudConfig(t *testing.T) {
+func TestVSphereCloudConfig(t *testing.T) {
 	testCases := []struct {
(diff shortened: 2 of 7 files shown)


pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -575,6 +575,14 @@ type OpenstackCloudSpec struct {
...
 	RouterID       string `json:"routerID"`
 	SubnetID       string `json:"subnetID"`
+	// Whether or not to use Octavia for LoadBalancer type of Service
+	// implementation instead of using Neutron-LBaaS.
+	// Attention:Openstack CCM use Octavia as default load balancer
+	// implementation since v1.17.0
+	//
+	// Takes precedence over the 'use_octavia' flag provided at datacenter
+	// level if both are specified.
+	UseOctavia *bool `json:"useOctavia"`
```
kron4eg: is `,omitempty` missed here?
```suggestion
	UseOctavia *bool `json:"useOctavia,omitempty"`
```

pkg/resources/cloudconfig/configmap_test.go
```diff
@@ -97,26 +114,160 @@ func TestGetVsphereCloudConfig(t *testing.T) {
...
-			if err := tc.verify(cloudConfig); err != nil {
-				t.Error(err)
+			t.Logf("config: %v", cloudConfig)
+			var actual interface{}
+			switch {
+			case tc.cluster.Spec.Cloud.Openstack != nil:
+				actual = unmarshalINICloudConfig(t, &openstack.CloudConfig{}, cloudConfig)
+			case tc.cluster.Spec.Cloud.VSphere != nil:
+				actual = unmarshalINICloudConfig(t, &vsphere.CloudConfig{}, cloudConfig)
+			}
```
kron4eg: This switch calls for 2 dedicated Test functions.

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: yep, good catch

pkg/resources/cloudconfig/configmap_test.go
irozzo-1A: Not sure I got your point. I'm providing a different structs `&vsphere.CloudConfig{}` or `&openstack.CloudConfig{}`. I'm using the same function as both Openstack and vSphere use an INI file. When we'll add test cases for other providers we'll use different test functions.

kron4eg: I meant to have
`func TestGetVsphereCloudConfig(t *testing.T) {`
and
`func TestGetOpenstackCloudConfig(t *testing.T) {`

as a way to avoid empty `interface{}` and to provide each own test cases

irozzo-1A: Ok, I see. I think that both ways are legitimate. Personally, I don't see any fundamental difference and any problem in using `interface{}` in this context. 

If we have dedicated tests things are more compartmented but we have some more repetitions. I'm not strongly opinioned on this, I think it's largely a matter of taste.

irozzo-1A: @kron4eg As I had to do some changes anyway and I was neutral about this comment I split the tests per provider. Does it look better for you now?

kron4eg: Yeah, now it's beautiful! 👍

kron4eg (approved):

irozzo-1A:
/assign @zreigz

sig-api review is needed here ;-)

PR #6568, opened 2021-02-17
sync docs files into the correct directory

xrstf:
We changed where we put these files some time in the past.

```release-note
NONE
```


kgroschoff (commented):
/lgtm

PR #6580, opened 2021-02-19
Fixed CT seed controller issues with cluster enqueue and deletion

lsviben:
Fixes issues discovered on Constraint Template seed controller. One is that CT deletes got stuck if they were already deleted on user cluster, second is that when clusters were created with OPA integration already enabled, the CT controller enqueue was not recognizing properly that they had OPA integration enabled.

```release-note
Fixed a bug in OPA-integration where deleting a Constraint Template in the seed cluster, when the user cluster Constraint Template is already deleted caused the deletion to get stuck.
Fixed a bug in OPA-integration where creating a cluster with OPA-integration enabled didnt trigger the Constraint Template reconcile loop.
```


xrstf (approved):
/approve

Issue #6585, opened 2021-02-22
Support Hetzner CSI/CCM


xrstf:
**User Story**
I want to run my userclusters on Hetzner, because Hetzner is cheap and good. Currently KKP does not support the CSI/CCM stuff itself, so I have to manually install the components into the usercluster myself. That is annoying, time consuming and brittle.

Also, for CSI this works just fine, but for CCM, the cluster must already be _created_ with `-cloud-provider=external` on the kubelets, otherwise installing the Hetzner CCM will not work (see for example https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/70).

**Acceptance criteria**
* When creating a cluster on Hetzner, KKP automatically provisions CCM/CSI.
* All components that can run in the seed (i.e. the usercluster namespace) do run in the seed.
* A safe migration strategy for existing Hetzner clusters (with possibly manually installed CSI) exists.

PR #6588, opened 2021-02-22
Add support for Hetzner CCM

xrstf:
This is part of #6585.

```release-note
Add support for Hetzner CCM
```

### DIFF
--- a/cmd/conformance-tests/custom_tests.go
+++ b/cmd/conformance-tests/custom_tests.go
@@ -164,6 +164,9 @@ func (r *testRunner) testLB(ctx context.Context, log *zap.SugaredLogger, userClu
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "test",
 			Namespace: ns.Name,
+			Annotations: map[string]string{
+				"load-balancer.hetzner.cloud/location": "nbg1",
+			},
 		},
 		Spec: corev1.ServiceSpec{
 			Type:     corev1.ServiceTypeLoadBalancer,
--- a/cmd/conformance-tests/runner.go
+++ b/cmd/conformance-tests/runner.go
@@ -1346,7 +1346,8 @@ func supportsStorage(cluster *kubermaticv1.Cluster) bool {
 func supportsLBs(cluster *kubermaticv1.Cluster) bool {
 	return cluster.Spec.Cloud.Azure != nil ||
 		cluster.Spec.Cloud.AWS != nil ||
-		cluster.Spec.Cloud.GCP != nil
+		cluster.Spec.Cloud.GCP != nil ||
+		(cluster.Spec.Cloud.Hetzner != nil && cluster.Spec.Version.Minor() >= 18)
 }
 
 func (r *testRunner) printAllControlPlaneLogs(ctx context.Context, log *zap.SugaredLogger, clusterName string) {
--- a/hack/run-user-cluster-controller-manager.sh
+++ b/hack/run-user-cluster-controller-manager.sh
@@ -20,13 +20,14 @@ cd $(dirname $0)/..
 source hack/lib.sh
 
 KUBERMATIC_DEBUG=${KUBERMATIC_DEBUG:-true}
+PPROF_PORT=${PPROF_PORT:-6600}
 
 echodate "Compiling user-cluster-controller-manager..."
 make user-cluster-controller-manager
 
 # Getting everything we need from the api
 # This script assumes you are in your cluster namespace, which you can configure via `kubectl config set-context --current --namespace=<<cluster-namespace>>`
-NAMESPACE="${NAMESPACE:-$(kubectl config view --minify | grep namespace |awk '{ print $2 }')}"
+NAMESPACE="${NAMESPACE:-$(kubectl config view --minify | grep namespace | awk '{ print $2 }')}"
 CLUSTER_NAME="$(echo $NAMESPACE | sed 's/cluster-//')"
 CLUSTER_RAW="$(kubectl get cluster $CLUSTER_NAME -o json)"
 CLUSTER_URL="$(echo $CLUSTER_RAW | jq -r '.address.url')"
@@ -36,22 +37,22 @@ if [ -z "${OWNER_EMAIL:-}" ]; then
   exit 1
 fi
 
-# We can not use the `admin-kubeconfig` secret because the user-cluster-controller-manager is
+# We cannot use the `admin-kubeconfig` secret because the user-cluster-controller-manager is
 # the one creating it in case of openshift. So we just use the internal kubeconfig and replace
-# the apiserver uurl
+# the apiserver url
 KUBECONFIG_USERCLUSTER_CONTROLLER_FILE=$(mktemp)
-ADMIN_KUBECONFIG="$(kubectl get secret internal-admin-kubeconfig -o json \
-  | jq '.data.kubeconfig' -r \
-  | base64 -d \
-  | yaml2json \
-  | jq --arg url "$CLUSTER_URL" '.clusters[0].cluster.server = $url')"
-echo $ADMIN_KUBECONFIG > $KUBECONFIG_USERCLUSTER_CONTROLLER_FILE
+kubectl --namespace "$NAMESPACE" get secret internal-admin-kubeconfig -o json | \
+  jq '.data.kubeconfig' -r | \
+  base64 -d | \
+  yq r --tojson - | \
+  jq --arg url "$CLUSTER_URL" '.clusters[0].cluster.server = $url' \
+  > $KUBECONFIG_USERCLUSTER_CONTROLLER_FILE
 echo "Using kubeconfig $KUBECONFIG_USERCLUSTER_CONTROLLER_FILE"
 
-OPENVPN_SERVER_SERVICE_RAW="$(kubectl get service openvpn-server -o json)"
+OPENVPN_SERVER_SERVICE_RAW="$(kubectl --namespace "$NAMESPACE" get service openvpn-server -o json)"
 
-SEED_SERVICEACCOUNT_TOKEN="$(kubectl get secret -o json \
-  | jq '.items[]|select(.metadata.annotations["kubernetes.io/service-account.name"] == "kubermatic-usercluster-controller-manager")|.data.token' -r \
+SEED_SERVICEACCOUNT_TOKEN="$(kubectl --namespace "$NAMESPACE" get secret -o json \
+  | jq -r '.items[]|select(.metadata.annotations["kubernetes.io/service-account.name"] == "kubermatic-usercluster-controller-manager")|.data.token' \
   | base64 -d)"
 SEED_KUBECONFIG=$(mktemp)
 kubectl config view  --flatten --minify -ojson \
@@ -72,6 +73,7 @@ set -x
   -kubeconfig=${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE} \
   -metrics-listen-address=127.0.0.1:8087 \
   -health-listen-address=127.0.0.1:8088 \
+  -pprof-listen-address=":${PPROF_PORT}" \
   -namespace=${NAMESPACE} \
   -openvpn-server-port=${OPENVPN_SERVER_NODEPORT} \
   -cluster-url=${CLUSTER_URL} \
--- a/pkg/controller/seed-controller-manager/kubernetes/pending.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/pending.go
@@ -124,7 +124,8 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 
 	if cluster.Spec.ClusterNetwork.ProxyMode == "" {
 		setProxyMode := func(c *kubermaticv1.Cluster) {
-			// IPVS causes issues with Hetzner's LoadBalancers
+			// IPVS causes issues with Hetzner's LoadBalancers, which should
+			// be addressed via https://github.com/kubernetes/enhancements/pull/1392
 			if c.Spec.Cloud.Hetzner != nil {
 				c.Spec.ClusterNetwork.ProxyMode = resources.IPTablesProxyMode
 			} else {
--- a/pkg/controller/seed-controller-manager/kubernetes/pending.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/pending.go
@@ -124,7 +124,12 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 
 	if cluster.Spec.ClusterNetwork.ProxyMode == "" {
 		setProxyMode := func(c *kubermaticv1.Cluster) {
-			c.Spec.ClusterNetwork.ProxyMode = resources.IPVSProxyMode
+			// IPVS causes issues with Hetzner's LoadBalancers
+			if c.Spec.Cloud.Hetzner != nil {
+				c.Spec.ClusterNetwork.ProxyMode = resources.IPTablesProxyMode
+			} else {
+				c.Spec.ClusterNetwork.ProxyMode = resources.IPVSProxyMode
+			}
 		}
 		modifiers = append(modifiers, setProxyMode)
 	}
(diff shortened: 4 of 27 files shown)


cmd/conformance-tests/custom_tests.go
```diff
@@ -164,6 +164,9 @@ func (r *testRunner) testLB(ctx context.Context, log *zap.SugaredLogger, userClu
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "test",
 			Namespace: ns.Name,
+			Annotations: map[string]string{
+				"load-balancer.hetzner.cloud/location": "nbg1",
```
xrstf, line 168: Hetzner LBs always need an annotation like this.

cmd/conformance-tests/runner.go
```diff
@@ -1346,7 +1346,8 @@ func supportsStorage(cluster *kubermaticv1.Cluster) bool {
 func supportsLBs(cluster *kubermaticv1.Cluster) bool {
 	return cluster.Spec.Cloud.Azure != nil ||
 		cluster.Spec.Cloud.AWS != nil ||
-		cluster.Spec.Cloud.GCP != nil
+		cluster.Spec.Cloud.GCP != nil ||
+		(cluster.Spec.Cloud.Hetzner != nil && cluster.Spec.Version.Minor() >= 18)
```
xrstf, line 1350: Hetzner CCM is documented to be for 1.18+ -- older versions of the CCM might work for older Kubernetes versions, but this is for now safe enough IMHO.

hack/run-user-cluster-controller-manager.sh
```diff
@@ -36,22 +37,22 @@ if [ -z "${OWNER_EMAIL:-}" ]; then
...
-ADMIN_KUBECONFIG="$(kubectl get secret internal-admin-kubeconfig -o json \
-  | jq '.data.kubeconfig' -r \
-  | base64 -d \
-  | yaml2json \
-  | jq --arg url "$CLUSTER_URL" '.clusters[0].cluster.server = $url')"
-echo $ADMIN_KUBECONFIG > $KUBECONFIG_USERCLUSTER_CONTROLLER_FILE
+kubectl --namespace "$NAMESPACE" get secret internal-admin-kubeconfig -o json | \
+  jq '.data.kubeconfig' -r | \
+  base64 -d | \
+  yq r --tojson - | \
```
xrstf, line 47: yaml2json was a tiny go program from some corner of the net, nowadays we have yq mostly everywhere.

pkg/controller/seed-controller-manager/kubernetes/pending.go
```diff
@@ -124,7 +124,12 @@ func (r *Reconciler) ensureClusterNetworkDefaults(ctx context.Context, cluster *
 
 	if cluster.Spec.ClusterNetwork.ProxyMode == "" {
 		setProxyMode := func(c *kubermaticv1.Cluster) {
-			c.Spec.ClusterNetwork.ProxyMode = resources.IPVSProxyMode
+			// IPVS causes issues with Hetzner's LoadBalancers
+			if c.Spec.Cloud.Hetzner != nil {
+				c.Spec.ClusterNetwork.ProxyMode = resources.IPTablesProxyMode
+			} else {
+				c.Spec.ClusterNetwork.ProxyMode = resources.IPVSProxyMode
+			}
```
xrstf: Without this, LBs never become healthy, even though the nodeport is open. See

* https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/58#issuecomment-654347439
* https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/58#issuecomment-659170459
* https://github.com/metallb/metallb/issues/153

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -474,7 +474,13 @@ type DigitaloceanCloudSpec struct {
...
 	CredentialsReference *providerconfig.GlobalSecretKeySelector `json:"credentialsReference,omitempty"`
 
-	Token string `json:"token,omitempty"` // Token is used to authenticate with the Hetzner cloud API.
+	// Token is used to authenticate with the Hetzner cloud API.
+	Token string `json:"token,omitempty"`
+	// Network is the pre-existing Hetzner network in which the machines are running.
+	// While machines can be in multiple networks, a single one must be chosen for the
+	// HCloud CCM to work.
+	// If this is empty, the network configured on the datacenter will be used.
+	Network string `json:"network,omitempty"`
```
xmudrii, line 483: We should probably create a follow-up ticket to add this option to the dashboard.

pkg/resources/cloudcontroller/hetzner.go
```diff
@@ -0,0 +1,139 @@
...
+					Name:  HetznerCCMDeploymentName,
+					Image: data.ImageRegistry(resources.RegistryDocker) + "/hetznercloud/hcloud-cloud-controller-manager:v1.8.1",
+					Command: []string{
+						"/bin/hcloud-cloud-controller-manager",
+						"--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig",
+						"--cloud-provider=hcloud",
+						"--leader-elect=false",
+						"--allow-untagged-cloud",
+						"--allocate-node-cidrs=true",
+						"--cluster-cidr=10.244.0.0/16",
```
xmudrii: This should usually equal to the pod subnet of the cluster, which is `172.25.0.0/16` if I'm not mistaken. Should we update this flag to reflect the correct value?

pkg/resources/cloudcontroller/hetzner.go
```diff
@@ -0,0 +1,139 @@
...
+			dep.Spec.Template.Spec.Containers = []corev1.Container{
+				*openvpnSidecar,
+				{
+					Name:  HetznerCCMDeploymentName,
+					Image: data.ImageRegistry(resources.RegistryDocker) + "/hetznercloud/hcloud-cloud-controller-manager:v1.8.1",
+					Command: []string{
+						"/bin/hcloud-cloud-controller-manager",
+						"--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig",
+						"--cloud-provider=hcloud",
+						"--leader-elect=false",
```
xmudrii: Should we consider enabling the leader election? Since we might end up rolling out the CCM deployment (e.g. when upgrading it), there may be two CCM instances running (and reconciling) at the same time (for a very short period, though). The leader election should prevent that.

Note: I'm aware that the example from the Hetzner CCM repo doesn't have it enabled, but I'm not sure is there some concrete reason.

pkg/resources/cloudcontroller/hetzner.go
```diff
@@ -0,0 +1,139 @@
...
+						"--cluster-cidr=10.244.0.0/16",
+					},
+					Env: []corev1.EnvVar{
+						{
+							Name: "NODE_NAME",
+							ValueFrom: &corev1.EnvVarSource{
+								FieldRef: &corev1.ObjectFieldSelector{
+									FieldPath: "spec.nodeName",
+								},
+							},
```
xmudrii, line 111: Is that actually correct? Since we're running CCM in the seed cluster in the user cluster namespace, that's going to equal to a node in the seed cluster (which might not be a Hetzner node at all, it could be GKE, AWS, etc.).

I'm not sure how Hetzner CCM uses that value (it's required), but this might eventually lead to unexpected behavior.

pkg/resources/cloudcontroller/util.go
```diff
@@ -0,0 +1,92 @@
...
+	"net/url"
+
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/resources"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ExternalCloudControllerFeatureSupported checks if the
+func ExternalCloudControllerFeatureSupported(dc *kubermaticv1.Datacenter, cluster *kubermaticv1.Cluster) bool {
```
xmudrii, line 29: Tiny nit: we should eventually use `switch` instead of `if`s.

pkg/resources/cloudcontroller/util.go
```diff
@@ -0,0 +1,92 @@
...
+		//
+		// TODO(irozzo) This is a dirty hack to temporarily support OTC using
+		// Openstack provider, remove this when dedicated OTC support is
+		// introduced in Kubermatic.
+		return !isOTC(dc.Spec.Openstack) && OpenStackCloudControllerSupported(cluster.Spec.Version)
+	}
+
+	if dc.Spec.Hetzner != nil {
+		return cluster.Spec.Version.Minor() >= 18
+	}
```
xmudrii: Should we also check here if the network is provided? Do we check that in some other place already?

pkg/resources/cloudcontroller/hetzner.go
xrstf: I am not sure exactly. I can try, but leaving it on the default value from the hcloud-ccm did not seem to cause issues.

pkg/resources/cloudcontroller/hetzner.go
xrstf: > Since we might end up rolling out the CCM deployment (e.g. when upgrading it), there may be two CCM instances running

So maybe we should change the deployment strategy to simply prevent this?

pkg/resources/cloudcontroller/hetzner.go
xrstf: This is a fun one. As you mentioned, it's a mandatory env variable, which is then later used to determine the current zone from the machine the CCM is running on. _However_, this is currently not used anywhere in the CCM, and only the unit tests use the GetZone() functions.

So technically for now it doesn't matter what value the env variable has. However this might change and makes me wonder if we need to deploy the CCM into the usercluster instead. I could also imagine us making an upstream PR to allow a new `HCLOUD_ZONE` variable, which would remove the need for dynamically determining the zone based on a hostname.

pkg/resources/cloudcontroller/util.go
xrstf: The networks will be defaulted to what is configured for the datacenter by the KKP admin. I don't see a reason to validate anything more here in this function. If anything, Seeds/datacenters should be validated to ensure a network is set.

pkg/resources/cloudcontroller/hetzner.go
xmudrii: Let's try to change it and see are tests going to pass.

pkg/resources/cloudcontroller/hetzner.go
xmudrii: I think that the leader election is still a better way to handle this. I also think that we use it for OpenStack (could/should be confirmed).

pkg/resources/cloudcontroller/hetzner.go
xmudrii: > However, this is currently not used anywhere in the CCM, and only the unit tests use the GetZone() functions.

This seems to be true.

> However this might change and makes me wonder if we need to deploy the CCM into the usercluster instead. I could also imagine us making an upstream PR to allow a new HCLOUD_ZONE variable, which would remove the need for dynamically determining the zone based on a hostname.

For the beginning, can you make an issue in the Hetzner CCM repo? We could discuss this with them, and eventually, make a PR to add a new variable. If it doesn't work for them, we can try to deploy it in the user cluster.

pkg/resources/cloudcontroller/util.go
xmudrii: > If anything, Seeds/datacenters should be validated to ensure a network is set.

Can you create a follow-up issue for that?

pkg/resources/cloudcontroller/hetzner.go
xrstf: https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/159

cmd/conformance-tests/runner.go
kron4eg: I'd like to suggest to change the body of this function to switch. Will make it more readable.

hack/run-user-cluster-controller-manager.sh
```diff
@@ -20,13 +20,14 @@ cd $(dirname $0)/..
 source hack/lib.sh
 
 KUBERMATIC_DEBUG=${KUBERMATIC_DEBUG:-true}
+PPROF_PORT=${PPROF_PORT:-6600}
```
kron4eg, line 23: port 6600 is already occupied :D

```shell
$ grep PPROF_PORT hack/*.sh
hack/run-api.sh:PPROF_PORT=${PPROF_PORT:-6600}
hack/run-api.sh:  -pprof-listen-address=":$PPROF_PORT" \
hack/run-master-controller-manager.sh:PPROF_PORT=${PPROF_PORT:-6600}
hack/run-master-controller-manager.sh:  -pprof-listen-address=":${PPROF_PORT}" \
hack/run-seed-controller-manager.sh:PPROF_PORT=${PPROF_PORT:-6600}
hack/run-seed-controller-manager.sh:  -pprof-listen-address=":${PPROF_PORT}" \
hack/run-user-cluster-controller-manager.sh:PPROF_PORT=${PPROF_PORT:-6600}
hack/run-user-cluster-controller-manager.sh:  -pprof-listen-address=":${PPROF_PORT}" \
```

hack/run-user-cluster-controller-manager.sh
```diff
@@ -72,6 +73,7 @@ set -x
   -kubeconfig=${KUBECONFIG_USERCLUSTER_CONTROLLER_FILE} \
   -metrics-listen-address=127.0.0.1:8087 \
   -health-listen-address=127.0.0.1:8088 \
+  -pprof-listen-address=":${PPROF_PORT}" \
   -namespace=${NAMESPACE} \
   -openvpn-server-port=${OPENVPN_SERVER_NODEPORT} \
   -cluster-url=${CLUSTER_URL} \
```
kron4eg, line 79: some `shfmt -w hack/run-user-cluster-controller-manager.sh` will definitely benefit this file.

hack/run-user-cluster-controller-manager.sh
xrstf: And that is exactly why this line is there. So you can override it.

pkg/provider/cloud/hetzner/provider.go
```diff
@@ -50,7 +50,15 @@ func (h *hetzner) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 	}
 
 	client := hcloud.NewClient(hcloud.WithToken(hetznerToken))
-	_, _, err = client.ServerType.List(context.Background(), hcloud.ServerTypeListOpts{})
+
+	if spec.Hetzner.Network == "" {
+		// this validates the token
+		_, _, err = client.ServerType.List(context.TODO(), hcloud.ServerTypeListOpts{})
```
kron4eg: hcloud doesn't set own timeouts inside `hcloud.NewClient()` or in consequent calls. Better to use context with reasonable timeout.

hack/run-user-cluster-controller-manager.sh
xrstf: I tried, with many different options, but none have actually improved the formatting in this file.

pkg/resources/cloudcontroller/hetzner.go
xrstf: Tests seem to continue to work :)

pkg/resources/cloudcontroller/hetzner.go
xrstf: Enabled it, seems to work just fine.

pkg/crd/kubermatic/v1/cluster.go
xrstf: https://github.com/kubermatic/dashboard/issues/3124

pkg/resources/cloudcontroller/util.go
xrstf: https://github.com/kubermatic/kubermatic/issues/6614

cmd/conformance-tests/runner.go
xrstf: I honestly don't see the improvement a switch statement would bring in Go, where either all branches have to return or have an explicit fallthrough. This however is a nice, concise disjunction. :)

xmudrii (approved):
/lgtm

PR #6590, opened 2021-02-23
Gatekeeper webhook timeout

### LINKED ISSUE #6571: Increase gatekeeper webhook timeout
**User Story**

Current timeout for the gatkeeper webhook is 2 sec, which in some cases is low and causes the policies to be ignored. Increase the default value and make it configurable.

lsviben:
Increases the default gatekeeper webhook timeout from 2 to 10 seconds, and adds an option to configure it. 

Fixes #6571 

```release-note
Increased default gatekeeper webhook timeout from 2 to 10 seconds, and added option in cluster settings to configure it.
```


xrstf:
Where is the actual request to make this configurable come from? I saw the issue, but _you_ opened it, so was it your idea? ^^

I can see why a larger timeout might be necessary (though evil and annoying), but allowing our users to configure it? Isn't that a bit tooooo much detail/power for users?

lsviben:
> Where is the actual request to make this configurable come from? I saw the issue, but _you_ opened it, so was it your idea? ^^
> 
> I can see why a larger timeout might be necessary (though evil and annoying), but allowing our users to configure it? Isn't that a bit tooooo much detail/power for users?

I see what you mean. But as only admins and editors can set it for a specific cluster I dont think its an issue. 

The idea to make it configurable for users happened when I was testing some OPA constraints and noticed that the low timeout (at that point 2 secs) is causing 50% of checks to fail. So from my side I can just make a change and put a higher default timeout, but the users cant do that, and if they do it manually on the webhook, it will anyway get reset to default values on the next reconcile cycle. So you can consider this as a preemptive response to a possible user feature request for a configurable webhook timeout.

cmd/user-cluster-controller-manager/main.go
```diff
@@ -59,25 +59,26 @@ import (
...
+	overwriteRegistry            string
+	cloudProviderName            string
+	nodelabels                   string
+	seedKubeconfig               string
+	ownerEmail                   string
+	updateWindowStart            string
+	updateWindowLength           string
+	dnsClusterIP                 string
+	opaIntegration               bool
+	opaIntegrationWebhookTimeout int
```
xrstf: Can we call this "OPA webhook timeout" everywhere? the dditional "integration" sounds weird. ^^

xrstf (commented):
One suggestion

xrstf (approved):
/approve

PR #6592, opened 2021-02-23
Remove the CCM/CSI Migration annotations in favor of feature flags

### LINKED ISSUE #5799: [OpenStack] Migration path for existing user clusters to external CCM and CSI

xmudrii:

We've introduced the CCM/CSI Migration logic based on two annotations (`ccm-migration.k8c.io/migration-needed` and `csi-migration.k8c.io/migration-needed`) in #6535.

This PR refactors the logic to drop those two annotations in favor of two feature flags:

* `externalCloudProvider` (this feature flag has been existing before)
* `csiMigration` (this is a new feature flag)

This PR uses the following workflow:

* New OpenStack clusters are created with the `externalCloudProvider` feature flag by default
  * New OpenStack clusters will have in-tree cloud provider enabled, but CCM controllers will be disabled, and the external CCM will be deployed
    * This is different from what we were doing before (we were deploying clusters will in-tree cloud provider disabled)
  * In a follow-up, we should also enable the `csiMigration` feature flag by default, and disable the in-tree provider
* Existing OpenStack clusters with the external CCM
  * Those clusters will remain with the in-tree cloud provider disabled (i.e. no changes will be done to those clusters)
  * The CSI migration can be triggered by enabling the `csiMigration` feature flag
* Existing OpenStack clusters without the external CCM
  * The CCM migration is triggered by enabling the `externalCloudProvider` feature flag
  * The CSI migration is triggered by enabling the `csiMigration` feature flag

Fixes #5799

```release-note
NONE
```

/assign @kron4eg @moadqassem

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
@@ -201,35 +201,6 @@ func TestCloudControllerManagerDeployment(t *testing.T) {
 			wantCCMCreator: false,
 		},
 		{
-			name: "No CCM migration ongoing",
-			cluster: &kubermaticv1.Cluster{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "cluster-a",
-				},
-				Spec: kubermaticv1.ClusterSpec{
-					Features: map[string]bool{
-						kubermaticv1.ClusterFeatureExternalCloudProvider: true,
-					},
-				},
-				Status: kubermaticv1.ClusterStatus{
-					NamespaceName: "test",
-				},
-			},
-			kcmDeploymentConfig: KCMDeploymentConfig{
-				Flags: []string{},
-				Status: appsv1.DeploymentStatus{
-					ObservedGeneration: 1,
-					Replicas:           2,
-					AvailableReplicas:  1,
-					UpdatedReplicas:    1,
-				},
-				Generation: 1,
-				Replicas:   2,
-				Namespace:  "test",
-			},
-			wantCCMCreator: true,
-		},
-		{
 			name: "No external cloud-provider",
 			cluster: &kubermaticv1.Cluster{
 				ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/resources_test.go
@@ -52,14 +52,11 @@ func TestCloudControllerManagerDeployment(t *testing.T) {
 			cluster: &kubermaticv1.Cluster{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "cluster-a",
-					Annotations: map[string]string{
-						kubermaticv1.CCMMigrationNeededAnnotation: "",
-						kubermaticv1.CSIMigrationNeededAnnotation: "",
-					},
 				},
 				Spec: kubermaticv1.ClusterSpec{
 					Features: map[string]bool{
 						kubermaticv1.ClusterFeatureExternalCloudProvider: true,
+						kubermaticv1.ClusterFeatureCSIMigration:          true,
 					},
 				},
 				Status: kubermaticv1.ClusterStatus{
@@ -85,14 +82,11 @@ func TestCloudControllerManagerDeployment(t *testing.T) {
 			cluster: &kubermaticv1.Cluster{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "cluster-a",
-					Annotations: map[string]string{
-						kubermaticv1.CCMMigrationNeededAnnotation: "",
-						kubermaticv1.CSIMigrationNeededAnnotation: "",
-					},
 				},
 				Spec: kubermaticv1.ClusterSpec{
 					Features: map[string]bool{
 						kubermaticv1.ClusterFeatureExternalCloudProvider: true,
+						kubermaticv1.ClusterFeatureCSIMigration:          true,
 					},
 				},
 				Status: kubermaticv1.ClusterStatus{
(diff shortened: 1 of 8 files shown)


pkg/resources/controllermanager/deployment.go
```diff
@@ -205,8 +205,12 @@ func getFlags(data *resources.TemplateData) ([]string, error) {
 	controllers := []string{"*", "bootstrapsigner", "tokencleaner"}
 	// If CCM migration is enabled and all kubeletes have not been migrated yet
 	// disable the cloud controllers.
-	if metav1.HasAnnotation(data.Cluster().ObjectMeta, kubermaticv1.CCMMigrationNeededAnnotation) &&
-		!kubermaticv1helper.ClusterConditionHasStatus(data.Cluster(), kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue) {
+	// If in-tree cloud providers are deactivated (KCMCloudControllersDeactivated is true),
```
xmudrii:
```suggestion
	// If in-tree cloud providers are deactivated (KCMInTreeCloudProviderDisabled is true),
```

pkg/resources/data.go
```diff
@@ -609,19 +633,17 @@ func ExternalCloudProviderEnabled(cluster *kubermaticv1.Cluster) bool {
 	// If we are migrating from in-tree cloud provider to CSI driver, we
 	// should not disable the in-tree cloud provider until all kubelets are
 	// migrated, otherwise we won't be able to use the volume API.
+	// TODO: Properly handle existing clusters.
```
xmudrii: This is now properly handled, so this TODO should be removed.

pkg/resources/data.go
```diff
@@ -609,19 +633,17 @@ func ExternalCloudProviderEnabled(cluster *kubermaticv1.Cluster) bool {
...
-		// 'externalCloudProvider' feature is activated.
-		if cluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] {
-			featureFlags = append(featureFlags, "CSIMigration=true", "CSIMigrationOpenStack=true", "ExpandCSIVolumes=true")
-		}
+	// Currently, CSIMigration is supported only for OpenStack
+	if cluster.Spec.Cloud.Openstack != nil && cluster.Spec.Features[kubermaticv1.ClusterFeatureCSIMigration] {
+		featureFlags = append(featureFlags, "CSIMigration=true", "CSIMigrationOpenStack=true", "ExpandCSIVolumes=true")
 		// The CSIMigrationNeededAnnotation is removed when all kubelets have
 		// been migrated.
 		if kubermaticv1helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue) {
```
xmudrii: This needs to be fixed to work with 1.21.

pkg/resources/data.go
```diff
@@ -549,6 +550,29 @@ func (d *TemplateData) KCMCloudControllersDeactivated() bool {
 	return false
 }
 
+// KCMInTreeCloudProviderDisabled checks is the in-tree cloud provider disabled
+// on the KCM deployment. We don't care is the roll-out complete in this case.
```
irozzo-1A:
```suggestion
// on the KCM deployment. We don't care if the roll-out complete in this case.
```

pkg/resources/data.go
```diff
@@ -549,6 +550,29 @@ func (d *TemplateData) KCMCloudControllersDeactivated() bool {
...
+func (d *TemplateData) KCMInTreeCloudProviderDisabled() bool {
+	kcm := appsv1.Deployment{}
+	if err := d.client.Get(d.ctx, ctrlruntimeclient.ObjectKey{Name: ControllerManagerDeploymentName, Namespace: d.cluster.Status.NamespaceName}, &kcm); err != nil {
```
irozzo-1A, line 563: If there is an error when retrieving the deployment we consider the KCM as enabled? this could cause some flip behaviors. Having an explicit annotation or feature gate is more robust on that respect

xmudrii: This can be mitigated by adding some retry logic. I'll look into this.

pkg/controller/seed-controller-manager/kubernetes/resources_test.go
```diff
@@ -215,35 +200,6 @@ func TestCloudControllerManagerDeployment(t *testing.T) {
...
-					Replicas:           2,
-					AvailableReplicas:  1,
-					UpdatedReplicas:    1,
-				},
-				Generation: 1,
-				Replicas:   2,
-				Namespace:  "test",
-			},
-			wantCCMCreator: true,
-		},
```
xmudrii, line 246: This test case has been removed intentionally because it's failing, however, it's expected that it'll fail.

We deploy/reconcile the external CCM only if the KCM has finished rolling-out. For example, if we're upgrading the cluster, we wouldn't reconcile the external CCM until the KCM has not finished rolling-out. What's important is that the external CCM will not be removed while it's not reconciled, so this should be okay.

pkg/resources/data.go
irozzo-1A: Yes, you can check if you have an error because the object is missing (this should not happen, but not sure what's the better approach in this case) and retry with backoff otherwise, I still think it was better to explicitly flag the cluster explicitly when CCM migration is required, but it's your call.

xmudrii:
With this PR, we lose the ability to make a difference between the clusters that are migrated and the clusters that already have the external CCM enabled. We may need this ability to implement some other controllers, so we decided that we will not move forward with this one.

/close

Issue #6507, opened 2021-02-03
Extend autoscaler featureset with extrema for scaling


shaase-ctrl:
**User Story**
As a user I want to create extrema for my autoscaler to adapt to my defined limits so that I always make the best use of the available ressources. 

**Acceptance criteria**
- Research if we should extend the autoscaler featureset or create an addtional function
- Check for existing autoscaler addon and either replace or enhance (see above)
- Create input variables for min and max values in backend and UI
- Decide if to couple or decouple node type and node count (could be beneficial to adjust this seperately)
- Display last autoscaling action from logs in UI
- Version below 1.18 are not supported

Nice to have / create a follow up
- Consider making presets available for projects & admins alike(part of machine deployement templates) (p.e. standard deployment always has 1-10 nodes)

**Input from customer**
UI:
check if cluster-autoscaler deployment is present, and if so
display max and min replicas fields in machinedeployment creade/edit modals.
remove replicas field (as it has little use and [min; max] replicas fields should be enough)

API:
add code in the API side to create appropriate annotations if min/max replicas are set.

kron4eg:
xref: https://github.com/kubermatic/kubermatic/pull/6596

ahmedwaleedmalik:
Reference PR for dashboard: https://github.com/kubermatic/dashboard/pull/3194

PR #6596, opened 2021-02-25
Add min/max replicas fields to node/machine deployment API

stoewer:

This PR adds the fields `minReplicas` and `maxReplicas` to the [NodeDeploymentSpec](https://github.com/kubermatic/kubermatic/blob/fcfdea9cfbc210bd3817c0c85dc00fea49e90095/cmd/kubermatic-api/swagger.json#L16671) of the Kubermatic API.

When the `maxReplicas` is present and has a value > `0`, the corresponding annotations for the cluster autoscaler are applied to the respective machinedeployment.

This PR contributes to #6507 

```release-note
1. Support setting cluster autoscaler annotations from the Kubermatic API
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -2128,6 +2128,6 @@ func ToInternalClusterType(externalClusterType string) kubermaticv1.ClusterType 
 
 const (
 	InitialMachineDeploymentRequestAnnotation = "kubermatic.io/initial-machinedeployment-request"
-	AutoscalerMinSizeAnnotation = "cluster.x-k8s.io/cluster-api-autoscaler-node-group-min-size"
-	AutoscalerMaxSizeAnnotation = "cluster.x-k8s.io/cluster-api-autoscaler-node-group-max-size"
+	AutoscalerMinSizeAnnotation = "cluster.k8s.io/cluster-api-autoscaler-node-group-min-size"
+	AutoscalerMaxSizeAnnotation = "cluster.k8s.io/cluster-api-autoscaler-node-group-max-size"
 )
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -1883,9 +1883,9 @@ type NodeDeploymentSpec struct {
 	// required: false
 	DynamicConfig *bool `json:"dynamicConfig,omitempty"`
 	// required: false
-	MinReplicas   int32 `json:"minReplicas,omitempty"`
+	MinReplicas int32 `json:"minReplicas,omitempty"`
 	// required: false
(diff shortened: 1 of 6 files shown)


kron4eg:
Codegen check fails with

 

> [2021-03-01T11:06:31+00:00] /home/prow/go/src/k8c.io/kubermatic/pkg/test/e2e/utils/apiclient is out of date. Please run hack/gen-api-client.sh

stoewer:
/assign @moadqassem

@moadqassem to me it looks like the tests failed for reasons not directly related to my change. Sadly I can't run `/retest` as suggested by the bot. How do we best resolve this?

stoewer:
Still some tests failing. But if you look at the full test PR test history it looks like that none of the tests is consistently failing. This looks pretty flaky to me.

pkg/api/v1/types.go
```diff
@@ -1882,6 +1882,10 @@ type NodeDeploymentSpec struct {
 	Paused *bool `json:"paused,omitempty"`
 	// required: false
 	DynamicConfig *bool `json:"dynamicConfig,omitempty"`
+	// required: false
+	MinReplicas int32 `json:"minReplicas,omitempty"`
+	// required: false
+	MaxReplicas int32 `json:"maxReplicas,omitempty"`
```
xmudrii, line 1888: @zreigz Since those are non-required fields, is it eventually better to use pointers here (we already use pointers for `Paused` and `DynamicConfig`)?

xmudrii (commented):
I'm wondering should we have some validation at the API level. The following cases come to my mind:

* min (and max) replicas must be greater than zero (required by autoscaler)
* replicas must be in the range of min and max replicas, or otherwise, the autoscaler will misbehave
* we should double-check, but IIRC, you must have both min and max replicas set

Does it make sense to do this at the API level or it's enough to add some validation in the UI later?

zreigz:
In my opinion this validation can be done in the UI. @floreks what do you think?

pkg/api/v1/types.go
zreigz: Yes, good point. It can be a pointer.

floreks:
We can always do some validation in the UI. The question is if users that use API and not only UI should also get this validation.

moadqassem:
Please keep in mind that, these two new fields don't contradict with the field Replicas in the machine deployment. You could run into a lot of fun if for example the assigned replica is 2 and the max is 1. Also there should be validation on api and in the admission webhooks of kkp

stoewer:
I'm no longer interested to get this change merged. I'll leave the branch with my changes there, maybe @furkhat would like to take over?

PR #6602, opened 2021-02-26
[release/v2.16] Gatkeeper webhook configurable timeout and default value increased

lsviben:
Increases the default gatekeeper webhook timeout from 2 to 10 seconds, and adds an option to configure it.

```release-note
Increased default gatekeeper webhook timeout from 2 to 10 seconds, and added option in cluster settings to configure it.
```


lsviben:
Wrong base, closing

PR #6621, opened 2021-03-02
shfmt

kron4eg:
* format shell scripts in hack
* verify that format is maintained

To remove any arguing about shell style (what gofmt did to go)

Fixes #

**Documentation**:

```release-note
NONE
```


hack/changelog-gen.sh
```diff
@@ -49,31 +51,31 @@ if [ "$(git rev-parse --abbrev-ref HEAD)" == "master" ]; then
...
 fi
 
 # Create CHANGELOG.md if not exists in order for the cat not to fail
-[ -f $CHANGELOG_FILE ] ||  touch $CHANGELOG_FILE
+[ -f $CHANGELOG_FILE ] || touch $CHANGELOG_FILE
 
 # Generate changelog
 OUTPUT="$(gchl --for-version $NEW_RELEASE since $(git describe --abbrev=0 --tags) --release-notes)"
-echo "${OUTPUT}" | cat - $CHANGELOG_FILE > temp
+echo "${OUTPUT}" | cat - $CHANGELOG_FILE >temp
```
xrstf: Why does shfmt do The Right Thing by putting spaces around pipes, but does The Bad Thing by not having spaces for output redirects? Why is `a | b` right, but `a > b` is wrong and needs to be `a >b`?

kron4eg: https://github.com/mvdan/sh/issues/243#issuecomment-392019995

xrstf: Meh, I am not really convinced, but this is not a hill I want to die on. You can have your redirects, as long as we never ever indent shell scripts with tabs.

kron4eg: There is now a flag (`-sr`) specified both prow and Makefile

PR #6625, opened 2021-03-03
Do not enable external CCM if hetzner dc is improperly configured

xrstf:
This makes it so that when admins upgrade KKP, new Hetzner clusters won't break because the admin did not configure a network. Otherwise the new cluster will have crashlooping CCM and machine-controller.

/test pre-kubermatic-e2e-hetzner-ubuntu-1.20

```release-note
NONE
```


kron4eg (approved):

xmudrii (approved):
/lgtm

Issue #6555, opened 2021-02-12
Add mutating webhook for Cluster resources to detect CCM/CSI migration


irozzo-1A:
The goal of this issue is to create a mutating webhook for Cluster objects that does the following things:

* Add the CSI/CCM annotations signaling that migration is needed when a `Cluster` is updated with `externalCloudProvider` feature.
* Ensure that `UseOctavia` is set to `false` when a `Cluster` is updated with `externalCloudProvider` feature, the reason behind this is that in-tree cloud provider defaults it to `false` and CCM defaults it to `true`.

xmudrii:
This has been fixed.
/close

PR #6627, opened 2021-03-03
Move webhook related files to the pkg/webhook package

xmudrii:

This is the first stage of #6555. I'll be creating multiple PRs to make it easier to test and revert if something goes wrong.

This PR moves all webhook-related files to a new package called `pkg/webhook`. Currently, we only have validation webhooks that are living in the `pkg/validation` package. As part of #6555, we're implementing a mutating webhook as well.

We want to reuse the same flags, options, and pod as for the validation webhook. Therefore, it makes sense to move webhook-related files from `pkg/validation` to `pkg/webhook` to make it clear it's related to the webhooks and that we have a mutating webhook as well.

xref #6555

```release-note
NONE
```

/assign @xrstf @irozzo-1A


xrstf (approved):
/approve

PR #6635, opened 2021-03-04
Implement the MutatingWebhook for the Cluster objects

xmudrii:

This PR is the second stage of #6555. It implements the mutating webhook for the Cluster objects to apply the CCM and CSI migration-needed annotations. Currently, the mutating webhook affects only OpenStack clusters, in the following way:

* If the Cluster is updated to enable the `externalCloudProvider` feature gate, apply the annotations and enable the `UseOctavia` feature

The webhook is currently not wired with the other components, which will be done in the next stage. This way we can more easily revert it if something goes wrong.

xref #6555

```release-note
NONE
```

/assign @irozzo-1A @xrstf

### DIFF
--- a/pkg/webhook/cluster/mutation/mutation.go
+++ b/pkg/webhook/cluster/mutation/mutation.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mutation
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/go-logr/logr"
+
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/utils/pointer"
+	ctrlruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// AdmissionHandler for mutating Kubermatic Cluster CRD.
+type AdmissionHandler struct {
+	log     logr.Logger
+	decoder *admission.Decoder
+}
+
+// NewAdmissionHandler returns a new cluster mutation AdmissionHandler.
+func NewAdmissionHandler() *AdmissionHandler {
+	return &AdmissionHandler{}
+}
+
+func (h *AdmissionHandler) InjectLogger(l logr.Logger) error {
+	h.log = l.WithName("cluster-mutation-handler")
+	return nil
+}
+
+func (h *AdmissionHandler) InjectDecoder(d *admission.Decoder) error {
+	h.decoder = d
+	return nil
+}
+
+func (h *AdmissionHandler) Handle(ctx context.Context, req webhook.AdmissionRequest) webhook.AdmissionResponse {
+	cluster := &kubermaticv1.Cluster{}
+	oldCluster := &kubermaticv1.Cluster{}
+
+	switch req.Operation {
+	case admissionv1.Create:
+		// NOP
+	case admissionv1.Update:
+		if err := h.decoder.Decode(req, cluster); err != nil {
+			return admission.Errored(http.StatusBadRequest, err)
+		}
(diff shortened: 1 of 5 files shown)


pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -0,0 +1,116 @@
...
+			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("marshaling cluster object failed: %v", err))
+		}
+
+		return admission.PatchResponseFromRaw(req.Object.Raw, mutatedCluster)
+	case admissionv1.Delete:
+		// NOP
+	default:
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("%s not supported on cluster resources", req.Operation))
+	}
+	return webhook.Allowed(fmt.Sprintf("no mutation done for request %s", req.UID))
```
xmudrii, line 88: I'm not sure how correct is it to return `webhook.Allowed` in a mutating webhook. It should be okay. As an alternative, I can always return `admission.PatchResponseFromRaw`.

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -0,0 +1,116 @@
...
+			h.log.Info("cluster mutation failed", "error", err)
+			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("cluster mutation request %s failed: %v", req.UID, err))
+		}
+
+		mutatedCluster, err := json.Marshal(cluster)
+		if err != nil {
+			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("marshaling cluster object failed: %v", err))
+		}
+
+		return admission.PatchResponseFromRaw(req.Object.Raw, mutatedCluster)
```
xmudrii, line 82: The reason for using `admission.PatchResponseFromRaw` is that it's easier to use than `Patched`. It's also used in the `controller-runtime` examples.

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (h *AdmissionHandler) mutateUpdate(ctx context.Context, oldCluster, newCluster *kubermaticv1.Cluster) error {
+	// This part of the code handles the CCM/CSI migration. It currently works
+	// only for OpenStack clusters, in the following way:
+	//   * Add the CCM/CSI migration annotations
+	//   * Enable the UseOctaiva flag
+	switch {
+	case newCluster.Spec.Cloud.Openstack != nil:
+		if v, ok := newCluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; ok && v {
+			if oldV, oldOk := oldCluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; !oldOk || !oldV {
```
irozzo-1A: As the values in the feature map are bool and zero value is `false` you can avoid checking for the presence of the key:

```suggestion
		if v, oldV := newCluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider], oldCluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; v && !oldV {
```

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -0,0 +1,116 @@
...
+				useOctavia := true
+				if newCluster.ObjectMeta.Annotations == nil {
+					newCluster.ObjectMeta.Annotations = map[string]string{}
+				}
+
+				newCluster.ObjectMeta.Annotations[kubermaticv1.CCMMigrationNeededAnnotation] = ""
+				newCluster.ObjectMeta.Annotations[kubermaticv1.CSIMigrationNeededAnnotation] = ""
+				newCluster.Spec.Cloud.Openstack.UseOctavia = &useOctavia
```
irozzo-1A: nit: you could use
https://github.com/kubernetes/utils/blob/fddb29f9d0095a7a37582a1d86732ae83620f2e8/pointer/pointer.go#L77-L80
to avoid declaring `useOctavia` variable

pkg/webhook/cluster/mutation/mutation_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+var (
+	testScheme = runtime.NewScheme()
+)
+
+func init() {
+	_ = kubermaticv1.AddToScheme(testScheme)
+}
+
+func TestHandle(t *testing.T) {
```
irozzo-1A, line 44: +1 for the unit tests ;-)

pkg/webhook/cluster/mutation/mutation.go
irozzo-1A: yes, I think it's better

xmudrii:
@irozzo-1A Addressed comments, PTAL. :slightly_smiling_face:

irozzo-1A:
LGTM apart from a few nits, I very appreciated the unit test coverage.

I think that the deployment of the `mutatingwebhookconfiguration` is missing here, do you plan to do it in a follow-up PR?

xmudrii:
@irozzo-1A Are all the nits now addressed with the latest commit/push?

> I think that the deployment of the mutatingwebhookconfiguration is missing here, do you plan to do it in a follow-up PR?

Yes, that's the plan for a follow-up. Also, I need to register the webhook in the seed-controller-manager. The reason for doing this in a follow-up is that if anything goes wrong I can just revert the MutatingWebhookConfiguration without having to revert the whole webhook.

PR #6638, opened 2021-03-04
Make operator deploy the MutatingWebhookConfig for Clusters

xmudrii:

This PR is the final phase of #6555. It modifies the operator to deploy the MutatingWebhookConfiguration for the webhook and modifies the seed-controller-manager to register the webhook.

xref #6555
The issue will be closed once we're sure that the webhook works properly.

The webhook **is not tested**. This is because testing mutating webhook locally is complicated. We should get the first signal from the E2E tests and then we can do more detailed tests on the dev environment.

In the case something goes wrong, it's enough to just revert this PR and remove the MutatingWebhookConfiguration from the cluster.

```release-note
NONE
```

/assign @irozzo-1A @xrstf

### DIFF
--- a/pkg/controller/operator/seed/resources/kubermatic/admission.go
+++ b/pkg/controller/operator/seed/resources/kubermatic/admission.go
@@ -37,7 +37,7 @@ const (
 	ClusterAdmissionWebhookName = "kubermatic-clusters"
 )
 
-func ClusterAdmissionWebhookCreator(cfg *operatorv1alpha1.KubermaticConfiguration, client ctrlruntimeclient.Client) reconciling.NamedValidatingWebhookConfigurationCreatorGetter {
+func ClusterValidatingWebhookConfigurationCreator(cfg *operatorv1alpha1.KubermaticConfiguration, client ctrlruntimeclient.Client) reconciling.NamedValidatingWebhookConfigurationCreatorGetter {
 	return func() (string, reconciling.ValidatingWebhookConfigurationCreator) {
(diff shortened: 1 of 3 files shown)


pkg/controller/operator/seed/resources/kubermatic/admission.go
```diff
@@ -90,6 +90,58 @@ func ClusterAdmissionWebhookCreator(cfg *operatorv1alpha1.KubermaticConfiguratio
 	}
 }
 
+func ClusterMutatingWebhookConfigurationCreator(cfg *operatorv1alpha1.KubermaticConfiguration, client ctrlruntimeclient.Client) reconciling.NamedMutatingWebhookConfigurationCreatorGetter {
+	return func() (string, reconciling.MutatingWebhookConfigurationCreator) {
+		return ClusterAdmissionWebhookName, func(hook *admissionregistrationv1.MutatingWebhookConfiguration) (*admissionregistrationv1.MutatingWebhookConfiguration, error) {
+			matchPolicy := admissionregistrationv1.Exact
+			failurePolicy := admissionregistrationv1.Fail
+			sideEffects := admissionregistrationv1.SideEffectClassNone
```
xmudrii, line 98: I'm not sure is this the correct policy for our webhook. I've checked the [official docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#side-effects), but I'm not really sure I understand this at all.

I'm not sure what they mean by "out-of-band changes". To me, this reads as changing some other/different objects than the object that we're mutating, but I'm not sure. I can look a bit more into this, but suggestions and feedback are appreciated.

irozzo-1A: Yes, I also understand it that way. Our mutating webhook is "side-effects" free as we are not changing any state in the scope of the cluster admission. LGTM

PR #6641, opened 2021-03-05
Leader election config in ComponentSettings

### LINKED ISSUE #6558: Make leader election lease duration configurable for kube-controller-manager and kube-scheduler
**User Story**

We've observed in some environments that rolling out NodeDeployments might cause kube-controller-manager and kube-scheduler to lose the leader election. As a result, both components are stuck in the permanent CrashLoopBackOff.

The issue can occur when a new node joins the cluster and the old node starts draining, i.e. all pods running on the old node are being rescheduled to the new node. When draining starts, the time needed to renew the leader election for kube-controller-manager and kube-scheduler goes to 10-15 seconds instead of the usual 2 seconds.

By default, both components have a deadline of 10 seconds and if the leader election is not successfully renewed in those 10 seconds, the affected component would crash. If it takes more than 10 seconds for the leader election to be renewed after restarting, the component would get stuck in CrashLoopBackOff.

We believe that this issue depends on the environment where the seed cluster is running, the performance of the seed cluster, and the available resources. So far, we've managed to reproduce this issue in some environments where the seed cluster is running on OpenStack and the user clusters are also running on OpenStack. The affected user clusters have 150+ pods.

**Acceptance criteria**

After investigating, we found that various factors related to the performance of the seed cluster can contribute to this issue. One of the recommended solutions for this problem is to increase `--leader-elect-renew-deadline`.

* `--leader-elect-renew-deadline` is configurable
* `--leader-elect-lease-duration` is configurable
* `--leader-elect-retry-period` is configurable

irozzo-1A:
This PR provides the possibility of configuring the leader election settings for controller-manager and scheduler in the `componentsOverride` section of `Cluster` resources.

Fixes #6558 

**Documentation**:

```release-note
Provide the possibility of configuring leader election parameters for user cluster components.
```


cmd/seed-controller-manager/options.go
```diff
@@ -155,6 +159,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.IntVar(&c.schedulerDefaultReplicas, "scheduler-default-replicas", 1, "The default number of replicas for usercluster schedulers")
 	flag.IntVar(&c.concurrentClusterUpdate, "max-parallel-reconcile", 10, "The default number of resources updates per cluster")
 	flag.IntVar(&c.addonEnforceInterval, "addon-enforce-interval", 5, "Check and ensure default usercluster addons are deployed every interval in minutes. Set to 0 to disable.")
+	flag.IntVar(&c.controlPlaneLeaderElectLeaseDurationSeconds, "control-plane-leader-elect-lease-duration", 0, "The lease duration in seconds used by control plane components using leader election (i.e. controller-manager and schedule). The default value for the component is used when equal or less than 0.")
```
irozzo-1A: Should be more appropriate to call it `user-cluster-leader-elect-lease-duration`?

cmd/seed-controller-manager/options.go
```diff
@@ -88,6 +89,9 @@ type controllerRunOptions struct {
 	admissionWebhook                                 webhook.Options
 	concurrentClusterUpdate                          int
 	addonEnforceInterval                             int
+	controlPlaneLeaderElectLeaseDurationSeconds      int
+	controlPlaneLeaderElectRenewDeadlineSeconds      int
+	controlPlaneLeaderElectRetryPeriodSeconds        int
```
xmudrii: Nit: why we don't use `metav1.Time` as a type here? That way, the flags would have the same semantics as the flags used by kube-controller-manager and kube-scheduler.

cmd/seed-controller-manager/options.go
xmudrii: I don't have strong opinions. Maybe `user-cluster-leader-elect-lease-duration` would be better and more clear.

xmudrii: Nit: typo

```suggestion
	flag.IntVar(&c.controlPlaneLeaderElectLeaseDurationSeconds, "control-plane-leader-elect-lease-duration", 0, "The lease duration in seconds used by control plane components using leader election (i.e. controller-manager and scheduler). The default value for the component is used when equal or less than 0.")
```

cmd/seed-controller-manager/options.go
```diff
@@ -155,6 +159,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
+	flag.IntVar(&c.controlPlaneLeaderElectLeaseDurationSeconds, "control-plane-leader-elect-lease-duration", 0, "The lease duration in seconds used by control plane components using leader election (i.e. controller-manager and schedule). The default value for the component is used when equal or less than 0.")
+	flag.IntVar(&c.controlPlaneLeaderElectRenewDeadlineSeconds, "control-plane-leader-elect-renew-deadline", 0, "The lease renew deadline in seconds by control plane components using leader election (i.e. controller-manager and schedule). Should be smaller or equal than control-plane-leader-elect-lease-duration. The default value for the component is used when equal or less than 0.")
+	flag.IntVar(&c.controlPlaneLeaderElectRetryPeriodSeconds, "control-plane-leader-elect-retry-period", 0, "The duration in seconds that control plane components using leader election (i.e. controller-manager and schedule) should wait between attempting acquisition and renewal of a leadership. The default value for the component is used when equal or less than 0.")
```
xmudrii:
```suggestion
	flag.IntVar(&c.controlPlaneLeaderElectRenewDeadlineSeconds, "control-plane-leader-elect-renew-deadline", 0, "The lease renew deadline in seconds by control plane components using leader election (i.e. controller-manager and scheduler). Should be smaller or equal than control-plane-leader-elect-lease-duration. The default value for the component is used when equal or less than 0.")
	flag.IntVar(&c.controlPlaneLeaderElectRetryPeriodSeconds, "control-plane-leader-elect-retry-period", 0, "The duration in seconds that control plane components using leader election (i.e. controller-manager and scheduler) should wait between attempting acquisition and renewal of a leadership. The default value for the component is used when equal or less than 0.")
```

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -361,6 +361,11 @@ type APIServerSettings struct {
 	NodePortRange               string `json:"nodePortRange,omitempty"`
 }
 
+type ControllerSettings struct {
+	DeploymentSettings     `json:",inline"`
+	LeaderElectionSettings `json:"leaderElection,omitempty"`
```
xmudrii, line 366: Is it expected that this is an embedded field? I see that you put a JSON tag for it, so I think it might be better if it was a (named) field.

pkg/validation/cluster_test.go
```diff
@@ -158,3 +159,75 @@ func TestValidateUpdateWindow(t *testing.T) {
...
+			name: "valid leader election settings",
+			leaderElectionSettings: kubermaticv1.LeaderElectionSettings{
+				LeaseDurationSeconds: pointer.Int32Ptr(int32(10)),
+				RenewDeadlineSeconds: pointer.Int32Ptr(int32(5)),
+				RetryPeriodSeconds:   pointer.Int32Ptr(int32(10)),
+			},
+			wantErr: false,
+		},
+		{
+			name: "valid leader election settings",
```
xmudrii: Nit: typo.
```suggestion
			name: "invalid leader election settings",
```

pkg/validation/cluster_test.go
```diff
@@ -158,3 +159,75 @@ func TestValidateUpdateWindow(t *testing.T) {
...
+		},
+		{
+			name: "lease duration only",
+			leaderElectionSettings: kubermaticv1.LeaderElectionSettings{
+				LeaseDurationSeconds: pointer.Int32Ptr(int32(10)),
+			},
+			wantErr: true,
+		},
+		{
+			name: "lease duration only",
```
xmudrii: Nit: typo
```suggestion
			name: "renew duration only",
```

cmd/seed-controller-manager/options.go
irozzo-1A: Well, `metav1.Time` is to represent a point in time, while here we are representing time durations (in sec). If you take a look at k8s objects this is the way it is done.

cmd/seed-controller-manager/options.go
irozzo-1A: yeah, I have the same feeling

cmd/seed-controller-manager/options.go
xmudrii: Sorry, I've been talking about `metav1.Duration`, such as here: https://github.com/kubernetes/kubernetes/blob/66cbf0196bd5ed0f06d7c40fccba887aae5405d8/staging/src/k8s.io/component-base/config/types.go#L46-L61

irozzo-1A: ah ok, yes in this case it makes sense

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: yes, this is expected as I don't want to break backward compatibility

pkg/validation/cluster_test.go
irozzo-1A: nope, this is valid for real :-)

xmudrii: Is that really the case? This seems like an invalid case to me because `wantErr` is set to true and `RenewDeadlineSeconds` > `LeaseDurationSeconds`.

irozzo-1A: you are right, I looked at the wrong line ;-)

pkg/crd/kubermatic/v1/cluster.go
xmudrii: Why would adding a new field break backward compatibility? It's supposed to be compatible and I think that we have been doing that before.

irozzo-1A: adding a new optional field does not break backward compatibility indeed, but what I did here is to embed the `DeploymentSettings` that was already existing into a new structure `ControllerSettings` to add the `LeaderElectionSettings`. I could have added the `LeaderElectionSettings` to `DeploymentSettings` directly, but this would have added the field to `APIServerSettings` as well, and this does not make sense as kas is leaderless.

cmd/seed-controller-manager/options.go
irozzo-1A: I thought a bit about this, and I'm not sure it's necessary to use `metav1.Duration` here, it would be convenient if we had to represent a parameter that can assume very different time scales, but here we will be always dealing with a few or at most tenths of seconds. If you don't have a strong point against I would rather keep this as it is.

pkg/crd/kubermatic/v1/cluster.go
xmudrii: I agree with you regarding the `DeploymentSettings` field/struct. However, my comment was regarding the `LeaderElectionSettings` struct. This is a new struct and adding it as a field (rather than embedding it) makes more sense to me. I usually prefer to avoid embedded fields, but I'm not going to block on this for now.

xmudrii:
@irozzo-1A This PR should also update the operator to allow specifying those values on the seed level. As the seed-controller-manager is deployed by the operator, it's impossible to set the flags directly (unless you're using charts, which are removed in this release).

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: ok, I see what you mean. I prefer to have something more structured as those fields are related instead of flattening in the `DeploymentSettings`, but yeah this is largely a matter of taste

irozzo-1A:
> @irozzo-1A This PR should also update the operator to allow specifying those values on the seed level. As the seed-controller-manager is deployed by the operator, it's impossible to set the flags directly (unless you're using charts, which are removed in this release).

good point, note that this has to be backported down to 2.14 and already brings value.
BTW I was thinking that probably we could take a different approach with the flags, and provide a structured componentsettings configuration, this will bring even more value.

cmd/seed-controller-manager/options.go
```diff
@@ -155,6 +159,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
...
+	flag.IntVar(&c.userClusterLeaderElectLeaseDurationSeconds, "user-cluster-leader-elect-lease-duration", 0, "The lease duration in seconds used by user cluster control plane components having leader election (i.e. controller-manager and scheduler). The default value for the component is used when equal or less than 0.")
+	flag.IntVar(&c.userClusterLeaderElectRenewDeadlineSeconds, "control-plane-leader-elect-renew-deadline", 0, "The lease renew deadline in seconds used by user cluster control plane components having leader election (i.e. controller-manager and scheduler). Should be smaller or equal than control-plane-leader-elect-lease-duration. The default value for the component is used when equal or less than 0.")
+	flag.IntVar(&c.userClusterLeaderElectRetryPeriodSeconds, "control-plane-leader-elect-retry-period", 0, "The duration in seconds that user cluster control plane components having leader election (i.e. controller-manager and scheduler) should wait between attempting acquisition and renewal of a leadership. The default value for the component is used when equal or less than 0.")
```
xmudrii: I think that you have forgotten to update the flags. Also, I think we should s/should/must.

```suggestion
	flag.IntVar(&c.userClusterLeaderElectRenewDeadlineSeconds, "user-cluster-leader-elect-renew-deadline", 0, "The lease renew deadline in seconds used by user cluster control plane components having leader election (i.e. controller-manager and scheduler). Must be smaller or equal than user-cluster-leader-elect-lease-duration. The default value for the component is used when equal or less than 0.")
	flag.IntVar(&c.userClusterLeaderElectRetryPeriodSeconds, "user-cluster-leader-elect-retry-period", 0, "The duration in seconds that user cluster control plane components having leader election (i.e. controller-manager and scheduler) should wait between attempting acquisition and renewal of a leadership. The default value for the component is used when equal or less than 0.")
```

irozzo-1A: yeah, no worries. I will remove the flags for the moment to reduce the scope of this PR.

xmudrii:
> BTW I was thinking that probably we could take a different approach with the flags, and provide a structured componentsettings configuration, this will bring even more value.

That could be useful, but I think that we should try to keep it simple for now, especially if we need to cherry-pick this to the release branches.

xmudrii (approved):
/lgtm
/approve

PR #6651, opened 2021-03-08
update nginx-ingress-controller to 0.44.0

### LINKED ISSUE #6172: Annotation Option for nginx LoadBalancer Services
**User Story**
As an kubermatic user
I want to set a (cloud) specific annotations to the kubermatic ingress chart:

E.g. to deploy Kubermatic behind a private network at AWS a customer would need to set the annotation `service.beta.kubernetes.io/aws-load-balancer-internal: "true"`. Therefor currently he could use a patch command as workaround (or create the service upfront):
```
kubectl annotate svc nginx-ingress-controller -n nginx-ingress-controller service.beta.kubernetes.io/aws-load-balancer-internal=true
```

But this not following the infrastructure as code principal. So we would need a helm value option for the chart at this point:
https://github.com/kubermatic/kubermatic/blob/master/charts/nginx-ingress-controller/templates/service.yaml#L25

Other use case, are fore example a on-premise installation with metallb where some CIDRs need to get specified.

**Acceptance criteria**
- [ ] Custom [service annotations](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer) can be set
- [ ] Docs are updated

xrstf:
What it says on the tin.

Fixes #6172

```release-note
update nginx-ingress-controller to 0.44.0
```

### DIFF
--- a/charts/nginx-ingress-controller/Chart.yaml
+++ b/charts/nginx-ingress-controller/Chart.yaml
@@ -14,8 +14,8 @@
 
 apiVersion: v1
 name: nginx-ingress-controller
-version: 1.1.8
-appVersion: 0.34.1
+version: 1.1.9
+appVersion: 0.44.0
 description: nginx-ingress-controller
 keywords:
 - kubermatic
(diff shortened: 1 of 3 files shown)


toschneck:
@xrstf @jiachengxu can we backport this to v2.16 branch as well. Recently came up this is need for a customer setup. not high-prio but with next realeas would be awesome

toschneck:
awesome thanks!

PR #6664, opened 2021-03-08
[release/v2.15] Leader election config in ComponentSettings (#6641)

irozzo-1A:
This is an automated cherry-pick of #6641

/assign irozzo-1A

```release-note
Provide the possibility of configuring leader election parameters for user cluster components.
```


pkg/validation/cluster.go
```diff
@@ -464,3 +464,22 @@ func ValidateUpdateWindow(updateWindow *kubermaticv1.UpdateWindow) error {
 	}
 	return nil
 }
+
+func ValidateLeaderElectionSettings(l kubermaticv1.LeaderElectionSettings) error {
```
xmudrii: I don't see that this function is used anywhere? Is that as expected?

irozzo-1A: Yep, we had no cluster validation webhook in, 2.15. But you are right we can remove this.

pkg/validation/cluster_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"testing"
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	"k8s.io/utils/pointer"
```
xmudrii: This seems to be leftover.

xmudrii (approved):
/lgtm

PR #6669, opened 2021-03-09
Increase proxy buffer size for nginx by default

jan-walther:
Update proxy-buffer-size to prevent a 502 error when using Keycloak. This happens every time the header is too big in size e.g. after logging out from Kubermatic without closing the Keycloak session.

Idea is from https://www.ibm.com/support/pages/502-error-ingress-keycloak-response
nginx docs: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/nginx-ingress-controller/values.yaml
+++ b/charts/nginx-ingress-controller/values.yaml
@@ -19,7 +19,8 @@ nginx:
   image:
     repository: k8s.gcr.io/ingress-nginx/controller
     tag: v0.44.0
-  config: {}
+  config: 
+    proxy-buffer-size: "16k"
 #   load-balance: "least_conn"
   extraArgs: []
   resources:


xrstf:
Hi @jan-walther, thanks for your PR!

Can you please adjust the `Chart.yaml` and bump the chart version?

irozzo-1A:
Hey @jan-walther,

If I understood correctly your PR, is really specific to your environment. This could be a sensible default for you but not for others. Couldn't you just override this in the  `values.yaml` file that you are using for your KKP deployment?
I would favor documenting this rather than changing the default value.

jan-walther:
> If I understood correctly your PR, is really specific to your environment. This could be a sensible default for you but not for others. Couldn't you just override this in the `values.yaml` file that you are using for your KKP deployment?
> I would favor documenting this rather than changing the default value.

Hi @irozzo-1A,

I tend to agree with this assessment. However I opted for setting the default value because 

a) the error happens with a default Keycloak deployment leading to a hard to debug error for the person deploying KKP

b) the increased proxy-buffer-size ups it from one page (usually 4k or 8k) to 16k. It's not as drastic of a change that it'd lead to memory shortages on any system.

@xrstf: I'll do that depending on the result of the issue raised by @irozzo-1A :-)

xrstf:
I have two questions:

a) Is there any downside to setting this to a higher value? This PR is increasing from the default (4k) to 16k, which seems safe enough?

b) The same can be achieved by setting an annotation on Kubermatic's Ingress. Maybe that's a better option, compared to setting it globally? On the other hand, we generally use the same OIDC provider for the IAP, and hence the Ingresses for all IAPs would need to be adjusted as well. Maybe setting globally is either after all?

irozzo-1A:
> I tend to agree with this assessment. However I opted for setting the default value because
> 
> a) the error happens with a default Keycloak deployment leading to a hard to debug error for the person deploying KKP

Yeah, thx @jan-walther for taking the time to report this, really appreciated. Just to understand better the issue, you are using Keycloack as IDP and you configured an OIDC connector in DEX pointing to it, is it correct?
Could you please tell me a bit more about the flow that causes to header to overcome the buffer limit? Isn't because you have a particularly big JWT inside? If you made a capture would be great if you could share the header after (stripping sensitive information of course).
 
> b) the increased proxy-buffer-size ups it from one page (usually 4k or 8k) to 16k. It's not as drastic of a change that it'd lead to memory shortages on any system.

Well, I agree that in normal conditions the buffer increase should not harm, but as the value is configurable I would not change the default unless it results in being too small for "nominal" usage. Otherwise, I would favor documentation ;-)

jan-walther:
> a) Is there any downside to setting this to a higher value? This PR is increasing from the default (4k) to 16k, which seems safe enough?

I don't think so. 

> b) The same can be achieved by setting an annotation on Kubermatic's Ingress. Maybe that's a better option, compared to setting it globally? On the other hand, we generally use the same OIDC provider for the IAP, and hence the Ingresses for all IAPs would need to be adjusted as well. Maybe setting globally is either after all?

I'm frankly not familiar enough with the Kubermatic infrastructure to really have an informed opinion on this topic. 

> Just to understand better the issue, you are using Keycloack as IDP and you configured an OIDC connector in DEX pointing to it, is it correct?

Correct. I've followed [this](https://docs.kubermatic.com/kubermatic/v2.16/advanced/oidc_config/) guide to setup Keycloak. The error is easily reproducable:

- Setup Kubermatic CE
- Setup Keycloak as shown above (note: Keycloak is deployed into the same KubeOne cluster as Kubermatic)
- Log into Kubermatic and therefore also into Keycloak
- Log out of Kubermatic (this will keep the Keycloak sesison alive)
- Upon clicking "sign in" one will be redirected to Keycloak and shown a 502 error page.

The nginx ingress controller deployed Kubermatic will then generate the following error message:

```
2021/03/11 15:15:00 [error] 210#210: *5867556 upstream sent too big header while reading response header from upstream, client: X.X.X.X, server: auth.kubermatic-host, request: "GET /auth/realms/master/protocol/openid-connect/auth?response_type=id_token&client_id=kubermatic&redirect_uri=https://kubermatic-host/projects&scope=openid%20email%20profile%20roles&nonce=shUp-~nAv2VsRGcr9CMO-YHyYE-iBfth HTTP/2.0", upstream: "http://X.X.X.X:8080/auth/realms/master/protocol/openid-connect/auth?response_type=id_token&client_id=kubermatic&redirect_uri=https://kubermatic-host/projects&scope=openid%20email%20profile%20roles&nonce=shUp-~nAv2VsRGcr9CMO-YHyYE-iBfth", host: "auth.kubermatic-host", referrer: "https://kubermatic-host/"     
```

irozzo-1A:
Thx @jan-walther for the reproduction steps, I have a few additional questions would you mind reaching me out on our slack channel https://kubermatic-community.slack.com/archives/C0PGCH99P?

xrstf:
Where do we stand with this PR?

jan-walther:
> Thx @jan-walther for the reproduction steps, I have a few additional questions would you mind reaching me out on our slack channel https://kubermatic-community.slack.com/archives/C0PGCH99P?

I can't seem to sign up without a @loodse.com mail address.

scheeles:
@jan-walther you can create an account with this link https://kubermatic-community.slack.com/join/shared_invite/zt-ocohxmja-OiX7OKzBvmlud70xylHAWA#/shared-invite/email

Issue #6551, opened 2021-02-12
Add controller in user-cluster controller manager to notify when nodes are ready to complete CSI migration


irozzo-1A:
The goal of this task is to introduce a new controller in the user-cluster controller manager, which role is to handle a condition in `Cluster` resources where an in-tree cloud provider to CSI migration is progress, that signals when all the nodes have been migrated not to use the kubelet cloud provider functionalities (attach/detach). This will allow completing the migration.

The controller will:

- Reconcile `Cluster` resources having the CSI migration annotation (seed-cluster side).
- For each of the `Cluster`s check that all nodes have the CSI migration feature gates and Kubelet `cloud-provider` flag is unset or empty.
- If the check passes the `Cluster` condition to `true`, set to `false` otherwise.

irozzo-1A:
Actually, the Kubelet feature gates are not currently reflected in any resource (e.g. machine, node) they are only provided with a flag to the machine-controller.

I see two options:

- We move the Kubelet feature gates to the machine spec in a way it can be observed by the controller (requires a change in mc).
- We start by manually signal that the machines have been rolled out, it will be up to an operator to annotate the cluster resource for instance to signal that.

irozzo-1A:
xref: https://github.com/kubermatic/machine-controller/issues/897

irozzo-1A:
@xmudrii As we discussed  the controller will need to watch machines in user-cluster and update cluster in seed-cluster you will need to pass both seed and user mgrs when you create the controller e.g.:

https://github.com/kubermatic/kubermatic/blob/54533b205f31640d005c67ef3388e4fae5dc0814/cmd/user-cluster-controller-manager/main.go#L297

mlavacca:
I'm working on it and I'd like to discuss an aspect with you. This controller's goal is to set the annotation `v1.kubelet-flags.machine-controller.kubermatic.io/ExternalCloudProvider: "true"
` on all the machines of the user cluster; once it is done, it updates the cluster status setting the `ClusterConditionCSIKubeletMigrationCompleted` condition to true. There is the possibility to have an inconsistent status in which the cluster status is updated, but the machine controller hasn't updated the kubelet flags yet. I see a lack of feedback to communicate the effective rolling out of nodes in the machine API. What do you think @irozzo-1A @xmudrii?

irozzo-1A:
[...]

PR #6695, opened 2021-03-12
Update machine-controller to v1.27.1

kron4eg:

Fixes #

**Documentation**:

```release-note
Upate machine-controller to v1.27.1
```


pkg/resources/machinecontroller/webhook.go
```diff
@@ -53,6 +54,24 @@ var (
...
+			}
+
+			externalCloudProvider := data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]
+			if externalCloudProvider {
+				args = append(args, "-node-external-cloud-provider=true")
+			}
+
+			featureGates := data.GetCSIMigrationFeatureGates()
+			if len(featureGates) > 0 {
+				args = append(args, "-node-kubelet-feature-gates", strings.Join(featureGates, ","))
```
kron4eg, line 72: If you think this line is missing `"RotateKubeletServerCertificate=true"` then it's not. `"RotateKubeletServerCertificate=true"` defaulting to true is now part of the machine-controller (even when `-node-kubelet-feature-gates` is defined but without `RotateKubeletServerCertificate`.

xmudrii (changes requested):
Can you also update the [machine-controller dependency](https://github.com/kubermatic/kubermatic/blob/24376f2209aca78f52be35d7814ca7d7494c3b23/go.mod#L48), so that:

* we're sure everything works fine with the newer version of the dependency
* we can use the newly added types (such as the annotations)

kron4eg:
> Can you also update the [machine-controller dependency](https://github.com/kubermatic/kubermatic/blob/24376f2209aca78f52be35d7814ca7d7494c3b23/go.mod#L48), so that:
> 
> * we're sure everything works fine with the newer version of the dependency

done

> * we can use the newly added types (such as the annotations)

should be done in the followup PRs done under #6551

xmudrii (approved):
/lgtm

PR #6706, opened 2021-03-16
Gatekeeper upgrade to 3.1.3

### LINKED ISSUE #6630: Upgrade gatekeeper to 3.1.3
**User Story**
Currently gatkeeper version is `v3.1.0-beta.9`, and its on quay. Since then they moved to dockerhub and added a few more versions. 

Simply changing the version to `3.1.3` doesnt work as there are some deployment changes that needs to be done, as Gatekeeper added some more CRDs.

lsviben:
Upgrades Gatekeeper from `3.1.0-beta-9` to `3.1.3` . 

This change also brings in a Gatekeeper ConstraintPodStatus and ConstraintTemplatePodStatus which require Gatkeeper to be able to get its own Pod, which is not possible with our deployment in which Gatkeeper Controller and Audit controllers are deployed on the Seed cluster. So this PR also moves the Gatekeeper components from the Seed to the User cluster. In hindsight, it seems that this was the proper way to do it from the start. 

But this brings on some difficulties, as with this action the user clusters will need more RAM and CPU to run the Gatkeepeer, which we tried to avoid earlier by having them on the Seed side. This could cause already running user clusters with OPA Integration enabled to break if there wont be enough resources on the user cluster. 
Gatekeeper resources(both audit and controller are the same):
```
resources:
  limits:
    cpu: 1000m
    memory: 512Mi
  requests:
    cpu: 100m
    memory: 256Mi
```
Further more, there is a question of leftover Gatekeeper resources in the user cluster namespaces of the Seed cluster for the clusters in which OPA integration already exists. They wont break OPA as the webhook will target the new deployments on the user cluster, but they should be cleaned up in some way, either automatically or with a provided script.

Fixes #6630 

```release-note
Upgraded Gatekeeper from 3.1.0-beta-9 to 3.1.3. 
NOTICE: this change also moves the Gatekeeper deployment from the Seed to the User clusters. This means that the user clusters will need some additional resources to run the Gatekeeper Pods. Admins please refer to the upgrade guidelines in the documentation.
```


xrstf:
> user clusters will need more RAM and CPU to run the Gatkeepeer

OPA is optional, right? so we can/should just mention the resource requirements somewhere in our UI.

> leftover Gatekeeper resources in the user cluster namespaces

In your next PR, you can extend your controller to perform a cleanup/garbage-collect on each reconciliation. in 2.19 we can then remove this cleanup code again. (To be 100% fair, not all our controllers are super good at cleaning up when some resource was moved or renamed. But don't let that stop you from being better! ^^)

PR #6708, opened 2021-03-16
Implement Project-Sync Controller

### LINKED ISSUE #6688: Create a controller to propagate Project objects to Seed Cluster
MLA controller in seed controller manager will need to create dedicated Grafana Organization for every KKP Project, and map KKP users to Grafana users in the Organization. 
However, Project and User information is not available in the Seed Cluster.

**Acceptance criteria** 
- A controller in the master controller manager should be implemented to propagate `Project` objects from Master Cluster to Seed Cluster.
- The controller should take into consideration cases when Master and Seed clusters are on the same cluster, in that case, `Project` objects will not be duplicated.

jiachengxu:
This PR implements a sync controller in the master controller manager to propagate Project objects from Master Cluster to Seed Cluster.
This is needed for MLA (Monitoring, Logging, Alerting) of user clusters. MLA controller in seed controller manager will need to create dedicated Grafana Organization for every KKP Project, and map KKP users to Grafana users in the Organization.
However, Project and User information is not available in the Seed Cluster. This PR is the first step, it adds a controller to sync projects from Master cluster to Seed cluster. 
Fixes #6688 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/codegen/reconcile/main.go
+++ b/codegen/reconcile/main.go
@@ -166,6 +166,11 @@ func main() {
 				ImportAlias:      "kubermaticv1",
 				APIVersionPrefix: "KubermaticV1",
 			},
+			{
+				ResourceName:     "Project",
+				ImportAlias:      "kubermaticv1",
+				APIVersionPrefix: "KubermaticV1",
+			},
 		},
 	}
 
--- a/cmd/master-controller-manager/controllers.go
+++ b/cmd/master-controller-manager/controllers.go
@@ -24,17 +24,10 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 
-	externalcluster "k8c.io/kubermatic/v2/pkg/controller/master-controller-manager/external-cluster"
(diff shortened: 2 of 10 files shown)


codegen/reconcile/main.go
```diff
@@ -166,6 +166,11 @@ func main() {
 				ImportAlias:      "kubermaticv1",
 				APIVersionPrefix: "KubermaticV1",
 			},
+			{
+				ResourceName:     "Project",
+				ImportAlias:      "kubermaticv1",
+				APIVersionPrefix: "KubermaticV1",
```
lsviben, line 172: Dont think you need this prefix. I needed it because there were 2 kinds of ConstraintTemplates. 

On the other hand, it doesnt hurt, and also could help if we create a project V2 in the future. So IMO it can stay

jiachengxu: Yeah, I am also thinking that we can handle multiple versions of `Project` in the future, not sure which is better for now :)

PR #6709, opened 2021-03-16
decreased default opa webhook timeout from 10 to 3 sec

lsviben:
Long timeouts can cause leader election problems if gatekeeper is down. (https://github.com/open-policy-agent/gatekeeper/pull/871) so putting it to 3 secs.

```release-note
Changed default gatekeeper webhook timeout to 3 sec
```


xrstf (approved):
/approve

PR #6716, opened 2021-03-17
WIP: Fix operating system decoding on empty operating system spec

### LINKED ISSUE #6637: Failed to show MachineDeployments
**User Story**
In my usercluster, I opted to manage the MachineDeployments myself. I created two simple MD's, they are basically identical and look like this:

```yaml
apiVersion: cluster.k8s.io/v1alpha1
kind: MachineDeployment
metadata:
  name: worker-dedicated
  namespace: kube-system
spec:
  # do not specify replica count, as its going to be dynamically altered
  # for our custom scale up/down solution
  # replicas: 3
  selector:
    matchLabels:
      prow: worker-dedicated
  strategy:
    rollingUpdate:
      maxSurge: 5
      maxUnavailable: 0
    type: RollingUpdate
  template:
    metadata:
      labels:
        prow: worker-dedicated
    spec:
      metadata:
        annotations:
          # Make rotation quick. Nothing but test jobs is supposed to be on these nodes
          kubermatic.io/skip-eviction: "true"
        labels:
          # Required by the CSI drivers
          csi.hetzner.cloud/location: __HETZNER_LOCATION__
          # Mark dedicated nodes, so we can schedule only specific jobs here using nodeSelector.
          kubermatic.io/dedicated: "true"
      providerSpec:
        value:
          cloudProvider: hetzner
          cloudProviderSpec:
            location: __HETZNER_LOCATION__
            serverType: ccx31
            networks:
              - __HETZNER_NETWORK__
            labels:
              keep: "true"
          operatingSystem: ubuntu
          sshPublicKeys:
            - __SSH_PUBKEY__
      versions:
        kubelet: 1.20.2
```

The MD works just fine, but when I list my clusters in the KKP dashboard I get:

screenshot-2021-03-04-161032

Maybe because I don't have an `operatingSystemSpec`?

**Acceptance criteria**
- [ ] The error does not happen anymore, I guess?

moadqassem:

When machine deployments are created manually(without using the api) and the OperatingSystemSpec is left empty, KKP api will fail to decode the MachineDeployment spec. This PR returns an empty OperatingSystemSpec in case of dropping that field in the manually created MD.

Fixes #6637 

No
```release-note
None
```

### DIFF
--- a/pkg/machine/convert.go
+++ b/pkg/machine/convert.go
@@ -49,6 +49,10 @@ func GetAPIV1OperatingSystemSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv
 		return nil, fmt.Errorf("failed to get machine providerConfig: %v", err)
 	}
 
+	if decodedProviderSpec.OperatingSystemSpec.Raw == nil {
+		return &apiv1.OperatingSystemSpec{}, nil
+	}
+
 	operatingSystemSpec := &apiv1.OperatingSystemSpec{}
 
 	switch decodedProviderSpec.OperatingSystem {


pkg/machine/convert.go
```diff
@@ -49,6 +49,10 @@ func GetAPIV1OperatingSystemSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv
 		return nil, fmt.Errorf("failed to get machine providerConfig: %v", err)
 	}
 
+	if decodedProviderSpec.OperatingSystemSpec.Raw == nil {
+		return &apiv1.OperatingSystemSpec{}, nil
+	}
```
kron4eg, line 54: Down the callstack the empty OperatingSystemSpec{} will cause the "no operating system set" error:

https://github.com/kubermatic/kubermatic/blob/e34f174049c563225eb6fbafc66c5c74ca6426b7/pkg/machine/util.go#L86-L96

moadqassem: Well, I am not sure why this is not using the OS filed directly! At any case I am actually still testing it :)

moadqassem:
@kron4eg We need then to adjust this in the machine controller or simply ignore the whole issue.

kron4eg:
yep

PR #6721, opened 2021-03-17
Fix CA bundle handling in S3 uploader

xrstf:
The S3 uploader is not built automatically for each release, so the old version 0.1.4 doesn't know the `-ca-bundle` flag yet. This PR fixes that and improves the Helm chart.

```release-note
NONE
```


xmudrii (approved):
/lgtm

PR #6745, opened 2021-03-22
Tunneling Expose Strategy e2e

### LINKED ISSUE #6463: Add e2e test for tunneling expose strategy
The goal of this issue is to create an automated test for the tunneling expose strategy.

The test should make sure that:

- User cluster can be created, workers join the cluster successfully (SNI entry point works properly).
- Pods in the user cluster can successfully reach the KAS (CONNECT entry point works properly).

irozzo-1A:
This PR introduces e2e tests for the tunneling expose strategy.

Fixes #6463 

**Documentation**:

```release-note
NONE
```


pkg/controller/user-cluster-controller-manager/resources/resources/envoy-agent/daemonset.go
```diff
@@ -153,9 +153,9 @@ func getContainers() []corev1.Container {
 			SecurityContext: &corev1.SecurityContext{
 				Capabilities: &corev1.Capabilities{
 					Add: []corev1.Capability{
-						"CAP_CHOWN",
-						"CAP_SETGID",
-						"CAP_SETUID",
+						"CHOWN",
```
irozzo-1A, line 156: In k8s 1.20 I had to remove `CAP`, to be checked with older versions.

pkg/controller/seed-controller-manager/seedresourcesuptodatecondition/seedresourcesuptodatecondition_controller.go
```diff
@@ -182,5 +181,5 @@ func (r *reconciler) seedResourcesUpToDate(ctx context.Context, cluster *kuberma
 		}
 	}
 
-	return true, nil
+	return len(deployments.Items) > 0 || len(statefulSets.Items) > 0, nil
```
irozzo-1A, line 190: Add a comment here, and clarify that this is to avoid adding the resources up-to-date condition in the cluster before deployments end ss are applied.

pkg/semver/semver.go
```diff
@@ -18,12 +18,16 @@ package semver
...
 	"encoding/json"
+	"flag"
 	"fmt"
 	"strconv"
 
 	semverlib "github.com/Masterminds/semver/v3"
 )
 
+// ensure that Semver implements flag.Value interface.
+var _ flag.Value = &Semver{}
```
emmanuel-kubermatic, line 29: Consider wrapping in a function perhaps 

“ v, ok:=i.(Somether)”

Then an if statement to error based on that

pkg/test/e2e/expose-strategy/client.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (k KubeVersion) MajorVersion() uint64 {
+	if v, err := strconv.ParseInt(k.Major, 10, 64); err == nil {
+		return uint64(v)
+	}
+	return 0
+}
+
+func (k KubeVersion) MinorVersion() uint64 {
+	if v, err := strconv.ParseInt(k.Minor, 10, 64); err == nil {
```
emmanuel-kubermatic, line 62: The major and minor version strconv methods can be a single function used in both parts

pkg/semver/semver.go
irozzo-1A: This is just to have a compile-time check that `Semver` implements `flag.Value` interface.

pkg/test/e2e/expose-strategy/client.go
irozzo-1A: Do you mean something like:
```
func (k KubeVersion) MinorVersion() (major uint64, minor uint64)
```
?

hack/lib.sh
```diff
@@ -343,7 +343,8 @@ start_docker_daemon() {
 
   # Start Docker daemon
   echodate "Starting Docker"
-  dockerd > /tmp/docker.log 2>&1 &
+  # Set the MTU to 1400 to avoid issues with our CI environment.
+  dockerd --mtu 1400 > /tmp/docker.log 2>&1 &
```
youssefazrak, line 347: 1350 would be better (if an interface is set to 1400MTU for example).

youssefazrak:
@irozzo-1A all good for me (minus the MTU comment)

hack/lib.sh
irozzo-1A: Unfortunately, I had to launch the test many times :-) due to our current problem with CI, and the issue did not occur. We can decrease it later if we see it occur again if you don't mind ;-)

irozzo-1A:
/assign @jiachengxu 
Hey, I need an approval from app management here ;-)

PR #6750, opened 2021-03-23
[v2.16] Fix openstack ccm not deployed

### LINKED ISSUE #6748: OpenStack cloud provider is not deployed due to container name mismatch
**What happened**:

After creating a user cluster I am getting the following error(s):

```
failed  to ensure Deployment  cluster-znftrnqpbz/openstack-cloud-controller-manager: failed to  generate object: failed to add apiserver.IsRunningWrapper: pod has no  container named "cloud-controller-manager"

# and IMHO this is a follow up error b/c the cloud controller does not get deployed:

failed  to create initial MachineDeployment: Internal error occurred: failed  calling webhook "machine-controller.kubermatic.io-machinedeployments":  Post  "https://machine-controller-webhook.cluster-znftrnqpbz.svc.cluster.local./machinedeployments?timeout=10s":  Service Unavailable
```

This means that kubermatic cannot create the openstack cloud controller because it cannot find the name of the container name specified here:

https://github.com/kubermatic/kubermatic/blob/503de88224894a1932a05891ff5bac09b8e9885f/pkg/resources/cloudcontroller/deployment.go#L61-L66

**What is the expected behavior**:

OpenStack cloud controller manager should be successfully deployed.

**How to reproduce the issue**:

I think this is reproducible by using version 2.16.x and the external OpenStack cloud controller.

**Anything else we need to know?**

I think the issue is that `cloudcontroller/deployment.go` uses the variable here for the container name:

https://github.com/kubermatic/kubermatic/blob/503de88224894a1932a05891ff5bac09b8e9885f/pkg/resources/cloudcontroller/deployment.go#L33

However when the resource is created in `cloudcontroller/openstack.go`, it uses this variable:

https://github.com/kubermatic/kubermatic/blob/503de88224894a1932a05891ff5bac09b8e9885f/pkg/resources/cloudcontroller/openstack.go#L34

https://github.com/kubermatic/kubermatic/blob/503de88224894a1932a05891ff5bac09b8e9885f/pkg/resources/cloudcontroller/openstack.go#L101-L114

Funnily enough this is not the case in the master, where `ccmContainerName` is used instead of `osName`:

https://github.com/kubermatic/kubermatic/blob/16260d4a74785f68f492d0a77c5cc48fb95650be/pkg/resources/cloudcontroller/openstack.go#L101-L114

Looking at the blame view, the lines got edited by the same PR (https://github.com/kubermatic/kubermatic/pull/6588), so this might be some merge issue?

**Information about the environment**:
Kubermatic version: 2.16.5

irozzo-1A:
This PR fixes #6748.

Fixes #6748 

Review commit by commit ;-)

**Documentation**:

```release-note
Fix deployment of Openstack CCM.
```

### DIFF
--- a/pkg/resources/cloudcontroller/openstack.go
+++ b/pkg/resources/cloudcontroller/openstack.go
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	osName = "openstack-cloud-controller-manager"
+	OpenstackCCMDeploymentName = "openstack-cloud-controller-manager"
 )
 
 var (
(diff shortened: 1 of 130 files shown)


xrstf (approved):
/approve

xrstf:
@irozzo-1A Do we backport this to 2.16? I guess so?

xrstf:
Ah this targets 2.16 already, nevermind.

PR #6756, opened 2021-03-24
Fixed clusters clusterrolebinding naming and added cleanup

lsviben:
Fixed clusters clusterrolebinding naming and added cleanup.

```release-note
NONE
```


xrstf:
Can you describe this more? What does this do? Why do we need it?

lsviben:
> Can you describe this more? What does this do? Why do we need it?

Sure, sorry. The gatekeeper upgrade PR (https://github.com/kubermatic/kubermatic/pull/6706) added the creation of ClusterRole and ClusterRoleBinding for the `user-cluster-controller` so that it can report the health check results for gatekeeper deployment to the Cluster object. 

I forgot to change the naming for each ClusterRoleBinding, so now the `seed-cluster-controller` tries to create the same ClusterRoleBinding for different Clusters. Thats why the name is generated for each Cluster now. 

An alternative would be to just edit the Subjects list of the ClusterRoleBinding, but I was worried that it could lead to some update conflicts and possible weird bugs. Let me know what you think?

Also added removal when deleting Clusters so there is no trash leftover.

xrstf (approved):
/approve

PR #6757, opened 2021-03-24
added removal of old gatekeeper resources on seed cluster

lsviben:
Follow-up from #6706 . Cleans up the leftover Gatekeeper resources on the Seed cluster after upgrading Gatekeeper, in which Gatekeeper components are moved from the Seed cluster to the User cluster. This functionality should be removed in a later release.

```release-note
NONE
```


xrstf (approved):
sig-cluster consists of more than just me, would be nice if you could spread your PRs across different people ;-)

/approve

PR #6778, opened 2021-03-28
Add Kubernetes 1.21

xrstf:

Fixes #

```release-note
Add Kubernetes 1.21
```


addons/kubelet-configmap/kubelet-configmap.yaml
```diff
@@ -437,3 +437,84 @@ data:
...
+        cacheTTL: 2m0s
+        enabled: true
+      x509:
+        clientCAFile: /etc/kubernetes/pki/ca.crt
+    authorization:
+      mode: Webhook
+      webhook:
+        cacheAuthorizedTTL: 5m0s
+        cacheUnauthorizedTTL: 30s
+    cgroupDriver: cgroupfs
```
kron4eg, line 463: Using `cgroupDriver: cgroupfs` here (and in all other kubelet-config configMaps) can be troublesome, as basically all guides telling to switch to the systemd.

kron4eg: This is however is not blocking issue

irozzo-1A: That's a good point. The most important thing AFAIK is that Kubelet and Docker/containerd are aligned, but there are also :

> In the field, people have reported cases where nodes that are configured to use cgroupfs for the kubelet and Docker, but systemd for the rest of the processes, become unstable under resource pressure.

(https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers)

This is used for dynamic kubelet configuration that is disabled by default right? Because AFAIK in machine-controller we setup both docker/containerd and kubelet to use systemd as cgroup driver. In this case, I don't expect dynamic kubelet configuration to work properly with mc provisioned machines.

In any case, this PR did not introduce the issue but potentially perpetuated it.

kron4eg: IIRC, it used for kubeadm BYO worker nodes.

kron4eg (approved):

PR #6779, opened 2021-03-28
Use new Docker images

xrstf:
We have begun to cleanup our Docker images and this makes it so that we use the new images.

The way we start Docker has been cleaned up. All containers now should use the `start-docker.sh` provided in the Docker image. Other than that, there are no real changes in this PR.

```release-note
NONE
```

### DIFF
--- a/pkg/test/e2e/api/cluster_rbac_test.go
+++ b/pkg/test/e2e/api/cluster_rbac_test.go
@@ -21,6 +21,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/test/e2e/api/cluster_rbac_test.go
```diff
@@ -153,20 +161,16 @@ func getErrorResponse(err error) string {
 	return string(rawData)
 }
 
-func createProjectWithCluster(t *testing.T, testClient *utils.TestClient, dc, credential, version, location string, replicas int32) (*v1.Project, *v1.Cluster) {
+func createProjectWithCluster(t *testing.T, testClient *utils.TestClient, dc, credential, version, location string, replicas int32) (*v1.Project, *v1.Cluster, error) {
```
zreigz, line 164: :+1:

xrstf:
The changes of this PR will probably be rolled up into another one. This can be closed for now.

PR #6784, opened 2021-03-29
Add MLA ProjectController

### LINKED ISSUE #6488: Multi-tenancy support for Grafana in the MLA controller
**User Story**

For multi-tenancy support, MLA controller has to configure Grafana Orgs and Users accordingly.

* If KKP user1 share project to the other KKP user2 - user2 should be created and add to the corresponding organization in the Grafana with permission specified by KKP user1
* When KKP users don't want to share projects anymore, the corresponding users should be deleted from Grafana. 

**Acceptance criteria**

aborilov:
This PR adds ProjectController to the MLA seed-cluster-controller manager. ProjectController - Create/Update/Delete Grafana Organizations based on Kubermatic Projects. New annotation `mla.k8c.io/organization` added to the Project when Grafana Organization created to hold Grafana Organization ID. `kubermatic.io/mla` finalizer added to the Project by ProjectController

Fixes #6488 

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/project_controller.go
```diff
@@ -0,0 +1,231 @@
...
+	ctrlOptions := controller.Options{
+		Reconciler:              reconciler,
+		MaxConcurrentReconciles: numWorkers,
+	}
+	c, err := controller.New(ControllerName, mgr, ctrlOptions)
+	if err != nil {
+		return err
+	}
+
+	debugPredicate := predicate.ByLabel(kubermaticv1.WorkerNameLabelKey, workerName)
```
jiachengxu: This only works when you were implementing this controller and this would only reconcile Project with your worker label.
This should be removed to work with all projects in the system

aborilov: removed, thx

cmd/seed-controller-manager/options.go
```diff
@@ -166,6 +167,7 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.Var(&c.tunnelingAgentIP, "tunneling-agent-ip", "The address used by the tunneling agents.")
 	flag.StringVar(&c.grafanaURL, "grafana-url", "http://grafana.mla.svc.cluster.local", "The URL of Grafana instance which in running for MLA stack.")
 	flag.StringVar(&c.grafanaHeaderName, "grafana-header-name", "X-WEBAUTH-USER", "Grafana Auth Proxy HTTP Header that will contain the username or email")
+	flag.StringVar(&c.grafanaSecret, "grafana-secret-name", "mla/secret", "Grafana secret name that contains basic auth info")
```
xmudrii: Looking from the default value, is the value containing both namespace and secret name? If yes, we should eventually update the field name and/or flag description to make it clear.

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -17,28 +17,61 @@ limitations under the License.
 package mla
 
 import (
+	"context"
+	"fmt"
+	"strings"
+
 	"go.uber.org/zap"
 
+	grafanasdk "github.com/kubermatic/grafanasdk"
```
xmudrii, line 26: Nit: wrong import group, should be above with Zap.

pkg/controller/seed-controller-manager/mla/project_controller.go
```diff
@@ -0,0 +1,228 @@
...
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"go.uber.org/zap"
+
+	grafanasdk "github.com/kubermatic/grafanasdk"
```
xmudrii, line 27: Nit: wrong import group.

pkg/controller/seed-controller-manager/mla/project_controller_test.go
```diff
@@ -0,0 +1,302 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+
+	grafanasdk "github.com/kubermatic/grafanasdk"
```
xmudrii, line 31: Tiny nit: similar to other comments, wrong import group. :slightly_smiling_face:

pkg/controller/seed-controller-manager/mla/project_controller_test.go
```diff
@@ -0,0 +1,302 @@
...
+			requests: []request{
+				{
+					name:     "get org by name",
+					code:     404,
+					method:   "GET",
+					path:     "/api/orgs/name/projectName-delete",
+					response: `{"id":1,"name":"projectName-create","address":{"address1":"","address2":"","city":"","zipCode":"","state":"","country":""}}`,
+				},
+			},
+		},
```
xmudrii: Should we actually fail if we're deleting the project, but the organization doesn't exist? In such cases, it would make sense just to do nothing.

pkg/controller/seed-controller-manager/mla/project_controller.go
```diff
@@ -0,0 +1,228 @@
...
+			if status, err := r.grafanaClient.DeleteOrg(ctx, org.ID); err != nil {
+				log.Debugf("unable to delete organization: %w (status: %s, message: %s)",
+					err, pointer.StringPtrDerefOr(status.Status, "no status"), pointer.StringPtrDerefOr(status.Message, "no message"))
+			}
+			return err
+		}
+		return nil
+	}
+	expected.ID = uint(id)
+	if !reflect.DeepEqual(org, expected) {
```
xmudrii, line 206: I have some concerns regarding `DeepEqual`. The `Org` struct has three fields: ID, Name, and Address.

`expected` has ID and Name set. We expect that `org` will have ID and Name set, but what about the Address? Since `org` is coming from the API, isn't it going to have Address as well? If it has the Address, isn't it going to fail DeepEqual? In that case, we would do an update on each reconciliation loop, which can be avoided.

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: but it's kubermatic repo, I thought it should be near other kubermatic imports

xmudrii: @aborilov You're totally right! I read `grafanasdk` and thought it's an upstream dependency. Forget about my comments.

pkg/controller/seed-controller-manager/mla/project_controller.go
aborilov: only we are creating those orgs and we don't set address, so I think we should be fine here

pkg/controller/seed-controller-manager/mla/project_controller_test.go
aborilov: yeah, I was thinking about that,  but for that, I will need to add one more PR for grafanasdk, cause as for now, sdk doesn't differentiate `NotFound` error and other errors(like network connection). so we need to fix sdk first.

xmudrii (approved):
/lgtm
/approve

PR #6785, opened 2021-03-29
Fix CoreDNS upgrade issue

### LINKED ISSUE #6428: Coredns upgrade from 2.14 to 2.15 not working properly
When upgrading from 2.14 to 2.15 the dns addon is being replaced by coredns created from the user-cluster-controller.

Since the dns addon manifests got deleted from the repository already they won't be deleted from inside the cluster. 
The user-cluster-controller will not update the existing resources though because the deployment selector is immutable. This breaks the cluster since the coredns service can't find the respective pods anymore. 

This bug can be fixed by not deleting the manifests directly and only deleting it in 2.16.

moelsayed:
CoreDNS has a special upgrade case: CoreDNS was moved into a resource instead of a manifest addon. When KKP is upgraded
 from 2.14.x to 2.15.x, the CoreDNS manifest is actually removed from the KKP container, so it's resources are not cleaned up by `r.cleanupManfests()`. When the resources are applied later in the user-cluster-controller, it fails because the old manifest deployment still exists and it's has immutable fields that can't be updated.
Fixes #6428

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
@@ -56,13 +56,9 @@ func DeploymentCreator() reconciling.NamedDeploymentCreatorGetter {
 			dep.Labels = resources.BaseAppLabels(resources.CoreDNSDeploymentName, nil)
 
 			dep.Spec.Replicas = resources.Int32(2)
-			// The Selector is immutable, so we don't change it if it's set. This happens in upgrade cases
-			// where coredns is switched from a manifest based addon to a user-cluster-controller-manager resource
-			if dep.Spec.Selector == nil {
-				dep.Spec.Selector = &metav1.LabelSelector{
-					MatchLabels: resources.BaseAppLabels(resources.CoreDNSDeploymentName,
-						map[string]string{"app.kubernetes.io/name": "kube-dns"}),
-				}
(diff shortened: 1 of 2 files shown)


pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
```diff
@@ -56,13 +56,9 @@ func DeploymentCreator() reconciling.NamedDeploymentCreatorGetter {
...
 
 			dep.Spec.Replicas = resources.Int32(2)
-			// The Selector is immutable, so we don't change it if it's set. This happens in upgrade cases
-			// where coredns is switched from a manifest based addon to a user-cluster-controller-manager resource
-			if dep.Spec.Selector == nil {
-				dep.Spec.Selector = &metav1.LabelSelector{
-					MatchLabels: resources.BaseAppLabels(resources.CoreDNSDeploymentName,
-						map[string]string{"app.kubernetes.io/name": "kube-dns"}),
-				}
+			dep.Spec.Selector = &metav1.LabelSelector{
```
irozzo-1A, line 59: We should maybe port this to the master branch as well. WDYT?

xmudrii: Yes, I think that we should.

xmudrii (approved):
/lgtm

PR #6773, opened 2021-03-26
Update to kind 0.10

xrstf:
**What this PR does / why we need it**:
No actual reason, but kind 0.8.1 was pretty old already and 0.10 was a painless update.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #6814, opened 2021-04-01
Revert Update to kind 0.10 (#6773)

xmudrii:

This PR reverts update to kind 0.10 (#6773).

The reason for doing this is that we're observing a lot of test failures (about 80-85% of tests failing) that started happening between March 26th and 27th. One of the PRs that merged in that period is #6773 which updated kind to 0.10 and Kubernetes to v1.20.2.

To roll out updating kind as a possible reason for failures, we'll revert this PR and observe the situation. It's also the PR that is easiest to revert.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -332,7 +332,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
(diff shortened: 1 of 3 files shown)


xrstf (approved):
With a sad heart, but full of optimism that this will definitly solve all our problems, I

/approve

PR #6837, opened 2021-04-02
Add MLA UserProjectBinding controller

### LINKED ISSUE #6488: Multi-tenancy support for Grafana in the MLA controller

aborilov:
This PR adds UserProjectBindingController to the MLA seed-cluster-controller manager. UserProjectBinding - Create/Update/Delete Grafana Users to Grafana Organizations based on Kubermatic UserProjectBindings. `kubermatic.io/mla` finalizer added to the UserProjectBinding by UserProjectBindingController

Fixes #6488 

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	type response struct {
+		ID uint `json:"id"`
+	}
+	res := &response{}
+	defer resp.Body.Close()
+	decoder := json.NewDecoder(resp.Body)
+	if err := decoder.Decode(res); err != nil || res.ID == 0 {
+		return nil, fmt.Errorf("unable to decode response : %w", err)
+	}
```
kron4eg: is this https://pkg.go.dev/github.com/kubermatic/grafanasdk#Client.GetActualUser call you've been looking for?

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -35,11 +38,22 @@ import (
...
 	mlaFinalizer   = "kubermatic.io/mla"
+	defaultOrgID   = 1
+)
+
+var (
+	// groupToRole map kubermatic groups to grafana roles
+	groupToRole = map[string]models.RoleType{
+		rbac.OwnerGroupNamePrefix:  models.ROLE_ADMIN,
+		rbac.EditorGroupNamePrefix: models.ROLE_EDITOR,
+		rbac.ViewerGroupNamePrefix: models.ROLE_VIEWER,
```
kron4eg, line 50: are we importing grafana only for those 3 constants? (I don't see any other usage).

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -73,5 +87,8 @@ func Add(
 	if err := newProjectReconciler(mgr, log, numWorkers, workerName, versions, grafanaClient); err != nil {
 		return fmt.Errorf("failed to create mla project controller: %v", err)
 	}
+	if err := newUserProjectBindingReconciler(mgr, log, numWorkers, workerName, versions, grafanaClient, &http.Client{}, grafanaURL, grafanaHeader); err != nil {
```
kron4eg: http client here is not configured thus has 0 timeouts, this is really bad for anything.

kron4eg: and it looks like that we could even remove those params (&http.Client{}, grafanaURL, grafanaHeader) as there is a standard method that can be used from sdk instead of custom calls.

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	// delete user from default org
+	if status, err := r.grafanaClient.DeleteOrgUser(ctx, defaultOrgID, res.ID); err != nil {
```
jiachengxu: I have few concerns about this implementation:
1. Where is the code to add users to the default organization? I didn't find that, and only see the code to remove the user from the default organization.
2. If this is failing all the time, does it mean that all users will be in the same default org? In that case, the user with `admin` role can kick other users (including the admin user that we are using in this controller) out of the default organization, which will lead to permission escalations.

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: yep, only for roles

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
aborilov: 1. why do you need to add somebody to default org? default org - is the org where new auth users assigned by default, so we don't need to add anyone there, we only moving users from default org to the org where they should be. 
2. users that added to default org does not have admin right

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
aborilov: nope :) actual user - is the user that currently authorized, but we don't have any users authorized here, we need to add new user using oauth here, there are no methods for that in the sdk

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: that not the method that we need. we need to do `post` instead of `get`

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
aborilov: also, this method won't fail if user already removed from default org

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
kron4eg: But your code is doing `GET /api/user`, literally `GetActualUser()`. Did you meant to do something else instead of getting the current user?

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: oh, yeah, my bad, we need to `get`, but I need to set header, so again, can't use this method, right?

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
jiachengxu: > also, this method won't fail if user already removed from default org

But I think it can fail due to some connection issues?

> users that added to default org does not have admin right

This is good, but still IMO it is not good to have all KKP users in one single default organization, because I think that will leak customer information to each other. They can log in to the default organization and see all users?

Actaully this behavior is configurable: https://grafana.com/docs/grafana/latest/administration/configuration/#auto_assign_org
You can disable the auto assignment.

pkg/controller/seed-controller-manager/mla/mla.go
kron4eg: wrong thread? :)

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
kron4eg: > @aborilov 
> oh, yeah, by bad, we need to get, but I need to set header, so again, can't use this method, right?

https://pkg.go.dev/github.com/kubermatic/grafanasdk#NewClient takes http.Client, which is sdk.DefaultHTTPClient, which, in turn a default http.Client!! No timeouts! superbad. So I suggest to construct our own client with our own http.RoundTripper (which will set appropriate header).

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: of course :)

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
jiachengxu: Oh, actually the user role in the default organization is also configurable: https://grafana.com/docs/grafana/latest/administration/configuration/#auto_assign_org_role
and it is `Viewer` by default, and for `Viewer`, they can not manage users: https://grafana.com/docs/grafana/latest/permissions/organization_roles/#compare-roles
Does this mean that they even cannot see users? If yes, I think we are good :)

aborilov: if it fail due to some connection issue, we need to fail and try on the next time

> Actaully this behavior is configurable: https://grafana.com/docs/grafana/latest/administration/configuration/#auto_assign_org
> You can disable the auto assignment.

have you read it? :)
>When set to false, new users automatically cause a new organization to be created for that new user.

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
aborilov: hm, can try to do that. but can we make it so this header will be set only for one endpoint?

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
jiachengxu: > have you read it? :)

Yes, of course, but if we can not avoid leaking information of customers to each other, do we have a better choice?

And as I mentioned, because I see that user will have `Viewer` by default, if they cannot see each other, then I think we can put them in the single default org.

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
kron4eg: Sure, why not. RoundTripper has all control over the http.Request

aborilov: but is it worth it?
and also from doc
> RoundTrip should not modify the request, except for consuming and closing the Request's Body.

aborilov: I set the timeout to 15 sec, what do you think, is it a good value?

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	"k8s.io/utils/pointer"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+type userProjectBindingReconciler struct {
```
kron4eg:
```suggestion
type userGrafanaReconciler struct {
```

Maybe let's rename this controller as `userProjectBindingReconciler` doesn't reflect what it's doing.

aborilov: I thought the name should reflect the main object that this reconciler is watching, in our case, it's `UserProjectBinding`.

aborilov: for example, the next PR will have `clusterReconciler` which is watching for `Cluster` object and adds Grafana Datasources, but it also add deployments, configmaps, services

aborilov: done

pkg/controller/seed-controller-manager/mla/userprojectbinding_controller.go
kron4eg: 👍

kron4eg (approved):
/approve
/lgtm

PR #6851, opened 2021-04-07
Do not create initial Machine Deployments with 0 replicas

maciaszczykm:
**What this PR does / why we need it**: In the https://github.com/kubermatic/kubermatic/pull/6064 the check for node deployment creation has changed from `if body.NodeDeployment != nil && body.NodeDeployment.Spec.Replicas > 0` to just `if body.NodeDeployment != nil`. This pull request reverts that single change.

**Special notes for your reviewer**: @xrstf @zreigz @floreks What do you think? Should we restore the old behavior and perhaps add some note in the UI?

EDIT: Other solution could be adding some checkbox in the UI that disables initial machine deployment creation and then we would be sending `nil` to the API and this check would not be needed.

**Documentation**:

```release-note
Initial Machine Deployments when user specified 0 replicas will not be created anymore.
```

### DIFF
--- a/pkg/handler/common/cluster.go
+++ b/pkg/handler/common/cluster.go
@@ -135,7 +135,7 @@ func CreateEndpoint(ctx context.Context, projectID string, body apiv1.CreateClus
 	// the user cluster (in case errors happen and the controller needs to re-reconcile), we ensure that the MD
 	// has a proper name instead of relying on the GenerateName.
 	partialCluster.Annotations = make(map[string]string)
-	if body.NodeDeployment != nil {
+	if body.NodeDeployment != nil && body.NodeDeployment.Spec.Replicas > 0 {
 		isBYO, err := common.IsBringYourOwnProvider(spec.Cloud)
 		if err != nil {
 			return nil, errors.NewBadRequest("cannot verify the provider due to an invalid spec: %v", err)


xrstf:
While this might make it easier to run the dashboard tests, I don't think this is necessarily good, obvious behaviour.

What about changing the logic to "if _no name_ is given, do not create it"? That would be more obvious to me as a user and serve the same purpose (which I guess is easing e2e tests).

maciaszczykm:
> While this might make it easier to run the dashboard tests, I don't think this is necessarily good, obvious behaviour.

Yeah, after creating and speaking with @floreks I got the same feeling.

> EDIT: Other solution could be adding some checkbox in the UI that disables initial machine deployment creation and then we would be sending nil to the API and this check would not be needed.

@xrstf What do you think about this? This would be most clear way.

Until then:

/hold

xrstf:
A checkbox would be really nice, I was assuming this PR was mostly to "fix the dashboard tests". If you can add the checkbox instead, that would be da :bomb:

maciaszczykm:
@cschieder Do you agree with the checkbox solution here? Where it would be the best to put it?

maciaszczykm:
/close

Discussion moved to the dashboard's repo.

PR #6852, opened 2021-04-07
WIP - Embed get[provider]Credentials() vars to struct

### LINKED ISSUE #5557: Wrap variables from get[provider]Credentials() into struct
As mentioned in [here](https://github.com/kubermatic/kubermatic/pull/5535#discussion_r443447855) it would be nice to use a struct for variables from `get[provider]Credentials()`, as they are passed around a lot. This basically applies to all providers.

imharshita:

This PR wraps variables from `get[provider]Credentials()` to a struct and adds a method which return it.

Fixes #5557

This PR contains changes for only the `OpenStack` provider right now.
When these changes are approved, this could be replicated for all other providers in either same or other PRs

**Documentation**:

```release-note
none
```


pkg/handler/v1/provider/openstack.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2020 The Kubermatic Kubernetes Platform contributors.
-
```
irozzo-1A: You should remove those changes in the license header ;-)

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,9 +25,23 @@ import (
...
 )
 
+type Request struct {
+	OpenstackReq
+	OpenstackTenantReq
+	OpenstackSubnetReq
+}
+
+func getCredentials(userInfo *provider.UserInfo, request Request, presetsProvider provider.PresetProvider) (Request, error) {
+	var err error
```
irozzo-1A: This line is superfluous

imharshita: on it

pkg/handler/v1/provider/openstack.go
imharshita: on it!

pkg/handler/v1/provider/openstack.go
```diff
@@ -45,17 +56,17 @@ func OpenstackSizeEndpoint(seedsGetter provider.SeedsGetter, presetsProvider pro
 			return nil, fmt.Errorf("error getting dc: %v", err)
 		}
 
-		username, password, domain, tenant, tenantID, err := getOpenstackCredentials(userInfo, req.Credential, req.Username, req.Password, req.Domain, req.Tenant, req.TenantID, presetsProvider)
+		credential, err := getCredentials(userInfo, req, presetsProvider)
 		if err != nil {
-			return nil, fmt.Errorf("error getting OpenStack credentials: %v", err)
+			return nil, err
```
irozzo-1A: I would keep the added info to the error, it's always useful for debugging. You could consider using error wrapping: https://blog.golang.org/go1.13-errors

pkg/handler/v1/provider/openstack.go
```diff
@@ -45,17 +56,17 @@ func OpenstackSizeEndpoint(seedsGetter provider.SeedsGetter, presetsProvider pro
...
 		}
 
 		settings, err := settingsProvider.GetGlobalSettings()
 		if err != nil {
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
-		return providercommon.GetOpenstackSizes(username, password, tenant, tenantID, domain, datacenterName, datacenter, settings.Spec.MachineDeploymentVMResourceQuota)
+		return providercommon.GetOpenstackSizes(credential.Username, credential.Password, credential.Tenant, credential.TenantID, credential.Domain, datacenterName, datacenter, settings.Spec.MachineDeploymentVMResourceQuota)
```
irozzo-1A: You could also consider changing the signature of the function to pass credential object directly e.g.
```
providercommon.GetOpenstackSizes(credential, datacenterName, settings.Spec.MachineDeploymentVMResourceQuota)
```

pkg/handler/v1/provider/openstack.go
```diff
@@ -302,8 +309,8 @@ func DecodeOpenstackSubnetNoCredentialsReq(c context.Context, r *http.Request) (
 	return req, nil
 }
 
-// OpenstackTenantReq represent a request for openstack tenants
-// swagger:parameters listOpenstackTenants
+// // OpenstackTenantReq represent a request for openstack tenants
```
irozzo-1A: typo, double `//`

pkg/handler/v1/provider/openstack.go
imharshita: Since this is a public method. I didn't change it in order to avoid any breaking in other part of the code where these functions are being called.

pkg/handler/v1/provider/openstack.go
imharshita: Done

pkg/handler/v1/provider/openstack.go
imharshita: Done

pkg/handler/v1/provider/openstack.go
imharshita: I have not removed the added info, it is here 
https://github.com/kubermatic/kubermatic/pull/6852/files#diff-1ea1ef7250db578e0ba92bef24d40f20707fb444fb51c91f5fa05b14745a6bc5R39

pkg/handler/v1/provider/openstack.go
imharshita: ok

pkg/handler/v1/provider/openstack.go
irozzo-1A: This is a very good point @imharshita, note that you could address this by using a different function name as Goland does not support function overloading. e.g.

```
func GetOpenstackSizesWithTenantReq(... ) (...) {
    return GetOpenstackSizes(...)
}
```

Secondly, we are not as strict as we should be with API compatibility in Kubermatic repo as it is not supposed to be imported as a library. But still your attention to this is very much appreciated ;-)

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +29,45 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider"
 )
 
+type credential struct {
+	credential, username, password, tenant, tenantID, domain string
+}
+
+func auth(ctx context.Context, request interface{}, userInfoGetter provider.UserInfoGetter, presetsProvider provider.PresetProvider) (*provider.UserInfo, *credential, error) {
```
irozzo-1A: Any particular reason why you are using `interface` here?

pkg/handler/v1/provider/openstack.go
```diff
@@ -315,7 +288,7 @@ type OpenstackTenantReq struct {
 	// Domain OpenStack domain name
 	Domain string
 	// in: header
-	// DatacenterName Openstack datacenter na
+	// 	centerName Openstack datacenter na
```
irozzo-1A: typo

pkg/handler/v1/provider/openstack.go
imharshita: Yes, this method is called by different types of requests e.g., OpenstackReq, OpenstackTenantReq, etc.
So request interface is used to make the method generalized.

irozzo-1A: Would not be simpler to use the minimum common denominator? e.g. OpenstackTenantReq maybe

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +29,45 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider"
 )
 
+type credential struct {
+	credential, username, password, tenant, tenantID, domain string
+}
+
+func auth(ctx context.Context, request OpenstackReq, userInfoGetter provider.UserInfoGetter, presetsProvider provider.PresetProvider) (*provider.UserInfo, *credential, error) {
+	req := reflect.ValueOf(request)
```
irozzo-1A: You don't need to use `reflect` any longer, right?

imharshita: right

pkg/handler/v1/provider/openstack.go
imharshita: Regarding this comment, I can surely do what you suggested, but that will require us to apply the same 5-6 methods.

pkg/handler/v1/provider/openstack.go
imharshita: done

pkg/handler/v1/provider/openstack.go
```diff
@@ -315,7 +278,33 @@ type OpenstackTenantReq struct {
...
+	// Domain OpenStack domain name
+	Domain string
+	// in: header
+	// Tenant OpenStack tenant name
+	Tenant string
+	// in: header
+	// TenantID OpenStack tenant ID
+	TenantID string
+	// in: header
+	// DatacenterName Openstack datacenter name
```
irozzo-1A, line 307: This change does not seem to be required. Isn't it?

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +28,44 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider"
 )
 
+type credential struct {
+	credential, username, password, tenant, tenantID, domain string
```
irozzo-1A, line 32: `credential` does not seem to be necessary

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +28,44 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider"
 )
 
+type credential struct {
```
irozzo-1A, line 31: nit: I would rather call this `credentials`, you could also consider using OpenstackReq directly instead of defining a new struct

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +28,44 @@ import (
...
+	credential, username, password, tenant, tenantID, domain string
+}
+
+func auth(ctx context.Context, req OpenstackReq, userInfoGetter provider.UserInfoGetter, presetsProvider provider.PresetProvider) (*provider.UserInfo, *credential, error) {
+	var cred credential
+	userInfo, err := userInfoGetter(ctx, "")
+	if err != nil {
+		return nil, nil, common.KubernetesErrorToHTTPError(err)
+	}
+	cred.credential, cred.username, cred.password, cred.domain, cred.tenant, cred.tenantID = req.Credential, req.Username, req.Password, req.Domain, req.Tenant, req.TenantID
```
irozzo-1A, line 41: nit: I would rather use this syntax instead of defining a var and assigning values later

```
	cred := credential{
	    username: req.Username,
             ...
	}
```

pkg/handler/v1/provider/openstack.go
```diff
@@ -324,29 +313,29 @@ type OpenstackTenantReq struct {
...
 
-func getOpenstackCredentials(userInfo *provider.UserInfo, credentialName, username, password, domain, tenant, tenantID string, presetProvider provider.PresetProvider) (string, string, string, string, string, error) {
+func getCredentials(userInfo *provider.UserInfo, cred credential, presetProvider provider.PresetProvider) (credential, error) {
+	credentialName := cred.credential
 	if len(credentialName) > 0 {
 		preset, err := presetProvider.GetPreset(userInfo, credentialName)
 		if err != nil {
-			return "", "", "", "", "", fmt.Errorf("can not get preset %s for the user %s", credentialName, userInfo.Email)
+			var dummy credential
+			return dummy, fmt.Errorf("can not get preset %s for the user %s", credentialName, userInfo.Email)
```
irozzo-1A, line 330:
```suggestion
			return credential{}, fmt.Errorf("can not get preset %s for the user %s", credentialName, userInfo.Email)
```

pkg/handler/v1/provider/openstack.go
```diff
@@ -28,34 +28,44 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider"
 )
 
+type credential struct {
+	credential, username, password, tenant, tenantID, domain string
+}
+
+func auth(ctx context.Context, req OpenstackReq, userInfoGetter provider.UserInfoGetter, presetsProvider provider.PresetProvider) (*provider.UserInfo, *credential, error) {
```
irozzo-1A, line 35: I'm not very convinced by the name of this function, can we find one that better represents what it does?

imharshita: can you suggest any?

pkg/handler/v1/provider/openstack.go
imharshita: used here: https://github.com/kubermatic/kubermatic/pull/6852/files#diff-1ea1ef7250db578e0ba92bef24d40f20707fb444fb51c91f5fa05b14745a6bc5R325

pkg/handler/v1/provider/openstack.go
imharshita: will change it to credentials for the consistency

pkg/handler/v1/provider/openstack.go
irozzo-1A: personally, I would not have grouped the call to `userInfoGetter` and `getCretendials` in the first instance :-), but for this particular function I would probably use something like `getAuthInfo`

pkg/handler/v1/provider/openstack.go
imharshita: ok

pkg/handler/v1/provider/openstack.go
imharshita: I just placed OpenstackReq close to OpenstackTenantReq for readability. Should I undo it?

pkg/handler/v1/provider/openstack.go
imharshita: umm, I grouped it coz `userInfo` is only used to `getCretendials`.
renaming  `auth` to `getAuthInfo`

irozzo-1A: that's legitimate ;-)

pkg/handler/v1/provider/openstack.go
```diff
@@ -324,29 +313,29 @@ type OpenstackTenantReq struct {
...
 	req.Password = r.Header.Get("Password")
 	req.Domain = r.Header.Get("Domain")
 	req.DatacenterName = r.Header.Get("DatacenterName")
 	req.Credential = r.Header.Get("Credential")
-
 	return req, nil
 }
 
-func getOpenstackCredentials(userInfo *provider.UserInfo, credentialName, username, password, domain, tenant, tenantID string, presetProvider provider.PresetProvider) (string, string, string, string, string, error) {
+func getCredentials(userInfo *provider.UserInfo, cred credential, presetProvider provider.PresetProvider) (credential, error) {
```
irozzo-1A, line 324: I would rather do something like:

```
func getPresetCredentials(userInfo *provider.UserInfo, presetName string, presetProvider provider.PresetProvider) (*credentials, error) {
...
```

I would leave the caller to handle the case where a preset name is not given or credentials cannot be obtained. WDYT?

pkg/handler/v1/provider/openstack.go
irozzo-1A: I see, check my comment https://github.com/kubermatic/kubermatic/pull/6852/files#r611916775

pkg/handler/v1/provider/openstack.go
irozzo-1A: no, you can keep it

pkg/handler/v1/provider/openstack.go
imharshita: Yes that sounds good

pkg/handler/v1/provider/openstack.go
irozzo-1A: don't bother, it's not a blocker

pkg/handler/v1/provider/openstack.go
imharshita: ok, so I can remove the `credential` field form struct and just do `presetName= req.Credential`

irozzo-1A: yep

imharshita:
Creating another PR #6902  for the changes

PR #6855, opened 2021-04-08
[release/v2.16]Allow loose RPF for calico

### LINKED ISSUE #6853: Canal fails to deploy on VM with rp_filter sets to 2
**User Story**
Canal (Calico + Flannel) is deployed on user clusters.
If those VMs are running a newer version of systemd, the `rp_filter` will be set by default to 2 instead of 1 for all the interfaces ([See this change](https://github.com/systemd/systemd/blob/v240/sysctl.d/50-default.conf#L24-L25)).

This is technically not an issue. The problem comes from the version of Canal we are delivering on user clusters (v3.8.0). This version, up to the v3.11, doesn't support `rp_filter` being set to 2. Only to 0 or 1. This prevents the canal pods from being ready and thus having a non-working cluster.

The fix from Canal comes in the v3.12 ([see this issue and the linked PR](https://github.com/projectcalico/felix/issues/2082)). Long-term solution is to update Canal to v3.12 or later but for the moment, we can deploy a quick fix to solve this issue without having to do a tricky upgrade of the CNI.
The workaround is to remove the `rp_filter` check from Calico by setting an env variable `FELIX_IGNORELOOSERPF` to `true` in the Canal's DaemonSet. 

Canal is currently deployed using the Addons controller. 
You can find the Addons here: https://github.com/kubermatic/kubermatic/tree/master/addons
The Canal manifest is here: https://github.com/kubermatic/kubermatic/tree/master/addons/canal

**Acceptance criteria**
- Canal pod is ready in VMs having rp_filter set to 2.

moelsayed:

Fixes #6853

**Documentation**:

```release-note
Adds `FELIX_IGNORELOOSERPF=true` to `calico-node` container env to allow running on nodes with `net.ipv4.conf.*.rp_filter = 2` set.
```

### DIFF
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -438,6 +438,7 @@ spec:
         # container programs network policy and routes on each
         # host.
         - name: calico-node
+          # FIXME: Remove FELIX_IGNORELOOSERPF from env when this is v3.12.0+
           image: docker.io/calico/node:v3.8.0
           env:
             # Use Kubernetes API as the backing datastore.
@@ -460,6 +461,9 @@ spec:
             # Cluster type to identify the deployment type
             - name: CLUSTER_TYPE
               value: "k8s,canal"
+            # Run on systems with loose reverse path forwarding (RPF) (net.ipv4.conf.*.rp_filter = 2)
+            - name: FELIX_IGNORELOOSERPF
+              value: "true"
             # Period, in seconds, at which felix re-applies all iptables state
             - name: FELIX_IPTABLESREFRESHINTERVAL
               value: "60"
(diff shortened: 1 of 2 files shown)


irozzo-1A:
ref: https://github.com/kubermatic/kubermatic/issues/6853
@youssefazrak

youssefazrak:
btw @moelsayed why is this not pushed to the master branch?
We can then cherry-pick for specific releases.

youssefazrak:
@jiachengxu could you please have a look and lgtm accordingly?

moelsayed:
@youssefazrak I am assuming we prefer to upgrade calico in master?

moelsayed:
/hold
I want to do one more test.

irozzo-1A:
> @youssefazrak I am assuming we prefer to upgrade calico in master?

I think it's safer having this on master, we did not plan to update CNI for the 2.17 release.

youssefazrak:
@moelsayed Agree with Iacopo here, let's be on the safe side and have it on the master as well.

PR #6856, opened 2021-04-08
Set default update versions

### LINKED ISSUE #6683: Fix default version settings for operator
**User Story**

The default version settings used by the operator in master branch and `release/v2.16` have the following issues:
* Automatic update from 1.15 should not have a `*` in the `To` field as it is not supported but refer to a fixed version (e.g. 1.6.13). See https://github.com/kubermatic/kubermatic/blob/b115d108db0c3bd93f7511ee38ee0c05c53ff2d2/pkg/version/manager.go#L190-L194
* We shave automatic updates from 1.16 to 1.17 as it is deprecated.

https://github.com/kubermatic/kubermatic/blob/6f6157c6304e0ed37205472e0e57e15279fe1283/pkg/controller/operator/common/defaults.go#L214-L237

**Acceptance criteria**

The default version settings of the operator trigger automatic migration to 1.15 an 1.16 to 1.17.

emmanuel-kubermatic:
The PR fixes #6683 and allows default upgrades on startup

Fixes #6683

**Documentation**:

https://github.com/kubermatic/kubermatic/blob/master/docs/dev-getting-started.md

None


pkg/controller/operator/common/defaults.go
```diff
@@ -216,15 +216,15 @@ var (
...
-				To:        "1.16.*",
+				To:        "1.16.13",
 				Automatic: pointer.BoolPtr(true),
 			},
 
 			// ======= 1.16 =======
 			{
 				// Allow to change to any patch version
 				From: "1.16.*",
-				To:   "1.16.*",
```
irozzo-1A: Here you should keep the wildcard, only where we have automatic upgrade we need a fixed version in the `To`

emmanuel-kubermatic: Fixed.

pkg/controller/operator/common/defaults.go
```diff
@@ -216,7 +216,7 @@ var (
 			{
 				// Auto-upgrade unsupported clusters
 				From:      "1.15.*",
-				To:        "1.16.*",
+				To:        "1.16.13",
 				Automatic: pointer.BoolPtr(true),
 			},
```
irozzo-1A: @emmanuel-kubermatic Could you please add the auto-upgrade from 1.16 to 1.17?

irozzo-1A:
@emmanuel-kubermatic Note that there is no 1.16 version supported you can drop all this:

https://github.com/kubermatic/kubermatic/blob/f0bb52779fc58bb6b16d0f1e539fe315b271ed2a/pkg/controller/operator/common/defaults.go#L216-L234

Because an automatic update toward a version that is not supported will fail.

You should just add an automatic update from `1.16.*` to `1.17.16` instead of:

https://github.com/kubermatic/kubermatic/blob/f0bb52779fc58bb6b16d0f1e539fe315b271ed2a/pkg/controller/operator/common/defaults.go#L235-L239

PR #6857, opened 2021-04-08
use KUBERMATIC_DOMAIN in legacy tests

xrstf:
I forgot to update the legacy tests because we already wanted to have removed the legacy chart, but didn't so far. Without using KUBERMATIC_DOMAIN, the legacy job cannot possibly work on the new CI cluster.

```release-note
NONE
```


xmudrii (approved):
/lgtm

PR #6868, opened 2021-04-10
[release/v2.16] fix imagepullsecret for user-cluster-prometheus-config-validation

xrstf:
Forgot to update this in #6858.

The util image is public and doesn't need a pull secret.

```release-note
NONE
```


xrstf:
I guess this is not needed anymore?

PR #6869, opened 2021-04-11
Use new Docker images

xrstf:
SIG infra has reorganized the Docker images a bit, this PR adjust our jobs to use them.

```release-note
NONE
```


hack/verify-licenses.sh
```diff
@@ -19,7 +19,7 @@ set -euo pipefail
 cd $(dirname $0)/..
 source hack/lib.sh
 
-CONTAINERIZE_IMAGE=quay.io/kubermatic/wwhrd:0.4.0-2 containerize ./hack/verify-licenses.sh
+CONTAINERIZE_IMAGE=quay.io/kubermatic/build:go-1.16-node-14-0 containerize ./hack/verify-licenses.sh
```
xmudrii, line 22: Haven't we decided/discussed to leave `wwhrd` and `codespell` images, so folks don't have to download a huge image?

xrstf: That would be news to me. Also, I plan on merging many of our verify jobs into one big job, because most of these jobs need to compile large parts of KKP anyway, so doing it all in one big job is probably maybe faster. For that though I need to have a Docker image with all tools in it.

xmudrii (approved):
/lgtm

PR #6878, opened 2021-04-13
Fix hetzner network being optional

### LINKED ISSUE #6876: Hetzner provider allows empty networks
Kubermatic API and dashboard threats Hetzner network as optional because we have fallback to Datacenter's hetzner network. This whole logic is wrong since following it we could potentially place in the same private subnet nodes from unrelated clusters.

To fix this problem we should change API to require Hetzner network and remove the fallback.

Original issue: https://github.com/kubermatic/machine-controller/issues/943
Machine-controller part fix: https://github.com/kubermatic/machine-controller/pull/944

lsviben:
Fixes the issue of Hetzner network being optional in Kubermatic API. 

Fixes #6876 

```release-note
When creating Hetzner Clusters, specifying the network is now mandatory.
```


kron4eg:
I also suggest to remove https://github.com/kubermatic/kubermatic/blob/eea732a21311cc9269f137dd281b47cf99feef89/pkg/resources/cloudcontroller/hetzner.go#L85-L87

lsviben:
> I also suggest to remove
> 
> https://github.com/kubermatic/kubermatic/blob/eea732a21311cc9269f137dd281b47cf99feef89/pkg/resources/cloudcontroller/hetzner.go#L85-L87

Im worried removing this could have consequences on existing Hetzner Clusters that somehow work without having the network specified, as this changes the hetzner CCM in a reconcile loop. But TBH im not an expert in the cloudcontroller or Hetzner

kron4eg (approved):
/approve

PR #6880, opened 2021-04-13
Update ca bundle and move update script

lsviben:
Moves the ca-bundle update to a separate script from the kubermatic operator update to provide better flexibility and to avoid the need to update the ca-bundle for PRs that are not connected with it. 

```release-note
NONE
```


xrstf (approved):
/approve

PR #6899, opened 2021-04-16
Default upgrade

### LINKED ISSUE #6683: Fix default version settings for operator

emmanuel-kubermatic:
The PR fixes #6683 and allows default upgrades on startup

Fixes #6683

**Documentation**:

https://github.com/kubermatic/kubermatic/blob/master/docs/dev-getting-started.md

```release-note
Fix default version configuration to have automatic upgrade from Kubernetes 1.16 to 1.17.
```

### DIFF
--- a/charts/kubermatic/static/master/versions.yaml
+++ b/charts/kubermatic/static/master/versions.yaml
@@ -2,6 +2,8 @@
 
 versions:
 - type: kubernetes
+  version: 1.16.13
+- type: kubernetes
   version: 1.17.9
 - type: kubernetes
   version: 1.17.11
--- a/charts/kubermatic/static/master/versions.yaml
+++ b/charts/kubermatic/static/master/versions.yaml
@@ -2,8 +2,6 @@
 
 versions:
 - type: kubernetes
-  version: 1.16.13
-- type: kubernetes
   version: 1.17.9
 - type: kubernetes
   version: 1.17.11
--- a/pkg/controller/operator/common/defaults.go
+++ b/pkg/controller/operator/common/defaults.go
@@ -190,6 +190,8 @@ var (
 	DefaultKubernetesVersioning = operatorv1alpha1.KubermaticVersioningConfiguration{
 		Default: semver.MustParse("v1.19.3"),
 		Versions: []*semver.Version{
+			// Kubernetes 1.16
+			semver.MustParse("v1.16.13"),
 			// Kubernetes 1.17
 			semver.MustParse("v1.17.9"),
 			semver.MustParse("v1.17.11"),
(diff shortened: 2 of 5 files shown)


pkg/controller/operator/common/defaults.go
```diff
@@ -231,7 +228,8 @@ var (
 			{
 				// Allow to next minor release
```
irozzo-1A: Nit: I would put:
```
// Auto-upgrade unsupported clusters
```

irozzo-1A:
@emmanuel-kubermatic Overall LGTM. In order to fix the failing tests you should run:

```
hack/update-docs.sh
hack/update-kubermatic-chart.sh
```

and:

```
gofmt -s -w pkg/controller/operator/common/defaults.go
```

emmanuel-kubermatic:
Thanks a ton !

pkg/controller/operator/common/defaults.go
```diff
@@ -190,6 +190,8 @@ var (
 	DefaultKubernetesVersioning = operatorv1alpha1.KubermaticVersioningConfiguration{
 		Default: semver.MustParse("v1.19.3"),
 		Versions: []*semver.Version{
+			// Kubernetes 1.16
+			semver.MustParse("v1.16.13"),
```
irozzo-1A: Mmm, I was thinking that putting back a `v1.16` version just to put the auto-upgrade from `1.15` to `1.16` does not seem to be a good option. This version will be visible to users. I think that for the long term we should rework the auto-upgrade mechanism to cover the case where the destination minor version is not supported, but for the moment I would just remove this and the auto-upgrade `1.15` -> `1.16`, customers that still have 1.15 clusters will need to override the default settings to be able to upgrade. WDYT?

emmanuel-kubermatic: That's fine

If that cluster version is Out Of Life, no point for the customer using it

charts/kubermatic/static/master/versions.yaml
```diff
@@ -1,6 +1,8 @@
 # This file has been generated using hack/update-kubermatic-chart.sh, do not edit.
 
 versions:
+- type: kubernetes
+  version: 1.16.13
```
irozzo-1A: Remember to remove this too, and to regenerate the charts ;-)

pkg/controller/operator/common/defaults.go
irozzo-1A: Ok, you can remove this version then

emmanuel-kubermatic: I've removed those and updating the docs for the next push

emmanuel-kubermatic:
Done

PR #6902, opened 2021-04-16
Embed get[provider]Credentials() vars to struct

### LINKED ISSUE #5557: Wrap variables from get[provider]Credentials() into struct

imharshita:

This PR wraps variables from `get[provider]Credentials()` to a struct and adds a method which return it.

Fixes #5557

This PR contains changes for only the `OpenStack` provider right now.
When these changes are approved, this could be replicated for all other providers in either same or other PRs

**Documentation**:

```release-note
none
```


pkg/handler/v1/provider/openstack.go
```diff
@@ -324,29 +330,20 @@ type OpenstackTenantReq struct {
...
 }
 
-func getOpenstackCredentials(userInfo *provider.UserInfo, credentialName, username, password, domain, tenant, tenantID string, presetProvider provider.PresetProvider) (string, string, string, string, string, error) {
-	if len(credentialName) > 0 {
-		preset, err := presetProvider.GetPreset(userInfo, credentialName)
-		if err != nil {
-			return "", "", "", "", "", fmt.Errorf("can not get preset %s for the user %s", credentialName, userInfo.Email)
-		}
+func getPresetCredentials(userInfo *provider.UserInfo, presetName string, presetProvider provider.PresetProvider) *v1.Openstack {
+	preset, err := presetProvider.GetPreset(userInfo, presetName)
```
irozzo-1A: Please don't swallow the error here, we should distinguish the cases where the credentials are not defined for openstack from the case where an error occurs when fetching the preset at least for logging purposes. I would rather do the following (if you think it's worthy to define a function for this):

```
func getPresetCredentials(userInfo *provider.UserInfo, presetName string, presetProvider provider.PresetProvider) (*credentials, error) {
    p, err := presetProvider.GetPreset(userInfo, presetName)
    if err != nil {
        return nil, fmt.Errorf("can not get preset %s for the user %s", presetName, userInfo.Email)
    }
    if p.Spec.Openstack == nil {
        return nil, fmt.Errorf("credentials for OpenStack provider not present in preset %s for the user %s", presetName, userInfo.Email)
    }
    return &credentials{
			username: p.Spec.Openstack.Username,
			password: p.Spec.Openstack.Password,
			domain:   p.Spec.Openstack.Domain,
			tenant:   p.Spec.Openstack.Tenant,
			tenantID: p.Spec.Openstack.TenantID,
		}, nil
}
```

pkg/handler/v1/provider/openstack.go
```diff
@@ -34,33 +33,26 @@ type credentials struct {
...
+		cred = &credentials{
+			username: req.Username,
+			password: req.Password,
+			domain:   req.Domain,
+			tenant:   req.Tenant,
+			tenantID: req.TenantID,
 		}
 	}
-	return userInfo, &cred, nil
+	return userInfo, cred, nil
```
irozzo-1A: I would streamline a bit, something like:

```
// No preset is used
if req.Credential == "" {
    return userInfo, &credentials{
        username: req.Username,
	password: req.Password,
	domain:   req.Domain,
	tenant:   req.Tenant,
	tenantID: req.TenantID,
    }, nil
}
// Preset is used
cred, err := getPresetCredentials(userInfo, presetName, presetsProvider)
if err != nil {
    return nil, nil, fmt.Errorf("error getting preset credentials for OpenStack: %v", err)
}
return userInfo, cred, nil
```
Wdyt?

imharshita: updated

irozzo-1A:
/lgtm

@zreigz Can you please take a look when you have the time? thx

PR #6907, opened 2021-04-20
Use the custom ca_bundle for openstack and vsphere cluster validation

lsviben:
Fixes the issue where cluster validation was failing on certificate error because the validation provider was not using the provided custom CA Bundle. Openstack and vSphere providers now use the proper CA Bundle.

```release-note
Fixes issue where cluster validation was failing on certificate error because the validation provider was not using the provided custom CA Bundle. 
```


pkg/provider/cloud/openstack/provider.go
```diff
@@ -467,6 +468,11 @@ func getAuthClient(username, password, domain, tenant, tenantID, authURL string)
 	if err != nil {
 		return nil, err
 	}
+	if client != nil {
+		// use the provider util's HTTP client to benefit, among other things, from its CA bundle
+		client.HTTPClient = httpcautil.HTTPClientConfig.New(httpcautil.HTTPClientConfig{})
```
kron4eg: .New() takes no parameters

kron4eg (changes requested):
Please use existing facilities 
https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/certificates/ca-bundle.go#L63

xrstf (changes requested):
/hold

Please do not introduce global state. #6538 went into a lot of effort to ensure the CA bundle is properly injected where it is needed.

kron4eg (approved):
/lgtm

lsviben:
> /hold
> 
> Please do not introduce global state. #6538 went into a lot of effort to ensure the CA bundle is properly injected where it is needed.

Changed it to the solution with injecting. Initially I was following the `machine-controller` example as they use the same clients code, but you and @kron4eg are right, this is better.

xrstf:
Correct, the machine-controller uses global state, but only super super reluctantly: https://github.com/kubermatic/machine-controller/pull/895 -- we tried the sane way first, but in that particular case, global state was the better choice.

Now I have to go wash my mouth with soap for having said that.

xrstf (approved):
/approve

irozzo-1A:
cherry-pick release/v2.17

PR #6916, opened 2021-04-22
fix displayed GCP disk types

### LINKED ISSUE #6715: Match disk types for GCP image types
The balanced disk type doesn't work for some instance types. The same for the SSD type

zreigz:
**What this PR does / why we need it**: fix displayed GCP disk types. I've checked all GCP machine standards with `pd-balanced` disk type and all instances were failing. The only way was to filter out this disk type.

Fixes #6715

```release-note
NONE
```

### DIFF
--- a/pkg/handler/common/provider/gcp.go
+++ b/pkg/handler/common/provider/gcp.go
@@ -34,6 +34,8 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider/cloud/gcp"
 	kubernetesprovider "k8c.io/kubermatic/v2/pkg/provider/kubernetes"
 	"k8c.io/kubermatic/v2/pkg/util/errors"
+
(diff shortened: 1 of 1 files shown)


pkg/handler/common/provider/gcp.go
```diff
@@ -169,7 +171,8 @@ func GCPDiskTypesWithClusterCredentialsEndpoint(ctx context.Context, userInfoGet
 
 func ListGCPDiskTypes(ctx context.Context, sa string, zone string) (apiv1.GCPDiskTypeList, error) {
 	diskTypes := apiv1.GCPDiskTypeList{}
-
+	// TODO: There are some issues at the moment with local-ssd and pd-balanced, that's why it is disabled at the moment.
```
lsviben, line 174: do we have some issue or something to reference here?

PR #6927, opened 2021-04-23
filter AWS machine types by architecture

### LINKED ISSUE #6928: Filter AWS machine types by architecture

zreigz:
**What this PR does / why we need it**: filter AWS machine types by architecture
When creating user clusters in AWS, instance types should be grouped by architecture.

Fixes #6928

```release-note
NONE
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -118,12 +118,13 @@ type DatacenterMeta struct {
 // AWSSize represents a object of AWS size.
 // swagger:model AWSSize
 type AWSSize struct {
-	Name       string  `json:"name"`
-	PrettyName string  `json:"pretty_name"`
-	Memory     float32 `json:"memory"`
-	VCPUs      int     `json:"vcpus"`
-	GPUs       int     `json:"gpus"`
-	Price      float64 `json:"price"`
+	Name         string  `json:"name"`
+	PrettyName   string  `json:"pretty_name"`
+	Memory       float32 `json:"memory"`
+	VCPUs        int     `json:"vcpus"`
+	GPUs         int     `json:"gpus"`
+	Price        float64 `json:"price"`
+	Architecture string  `json:"architecture"`
 }
 
 // AWSSizeList represents an array of AWS sizes.
--- a/pkg/handler/common/provider/aws.go
+++ b/pkg/handler/common/provider/aws.go
@@ -274,7 +274,6 @@ func isARM64Architecture(physicalProcessor string) bool {
 }
 
 func isValidArchitecture(architecture, processorType string) bool {
-
 	if architecture == handlercommon.ARM64Architecture {
 		return isARM64Architecture(processorType)
 	}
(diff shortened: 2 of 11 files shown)


pkg/api/v1/types.go
```diff
@@ -118,12 +118,13 @@ type DatacenterMeta struct {
...
 // swagger:model AWSSize
 type AWSSize struct {
-	Name       string  `json:"name"`
-	PrettyName string  `json:"pretty_name"`
-	Memory     float32 `json:"memory"`
-	VCPUs      int     `json:"vcpus"`
-	GPUs       int     `json:"gpus"`
-	Price      float64 `json:"price"`
+	Name         string  `json:"name"`
+	PrettyName   string  `json:"pretty_name"`
```
maciaszczykm, line 122: Can you change to camelCase? I know it's mixed but it would be good if we would be targeting only one format.

pkg/handler/common/provider/aws.go
```diff
@@ -244,19 +245,47 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
...
 	return filterAWSByQuota(sizes, quota), nil
 }
 
+func isARM64Architecture(physicalProcessor string) bool {
+	// right now there is only one Arm-based processors: Graviton2
+	return strings.Contains(physicalProcessor, "Graviton")
+}
+
+func isValidArchitecture(architecture, processorType string) bool {
+
```
maciaszczykm: Remove empty line.

pkg/handler/common/provider/aws.go
```diff
@@ -244,19 +245,47 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
...
+			Architecture: machineArchitecture,
 		})
+
 	}
 
 	return filterAWSByQuota(sizes, quota), nil
 }
 
+func isARM64Architecture(physicalProcessor string) bool {
+	// right now there is only one Arm-based processors: Graviton2
```
maciaszczykm: I was hoping that https://instances.vantage.sh/ would have more info about architecture, but it seems that there is just basic info like:

```
[x86_64]
[x86_64 i386]
```

At the moment filtering by procesor names can work but we should think about finding a better way. The library was chosen some time ago, perhaps there are better ways right now.

maciaszczykm: However, we might also filter basing on the names. ARM instance names start with `A1`:

Zrzut ekranu 2021-04-26 o 11 33 57

https://aws.amazon.com/ec2/instance-types/a1/

I think it would be a better option for now.

irozzo-1A: I don't think it's not just `a1` family to use arm, I see also [c6](https://aws.amazon.com/ec2/instance-types/c6/), [r6](https://aws.amazon.com/ec2/instance-types/r6/), etc.

maciaszczykm: Okay then it seems easier to use the processor name.

pkg/api/v1/types.go
zreigz: it's already in use in the UI. This change will break displaying the name

pkg/handler/common/provider/aws.go
zreigz: done

zreigz:
@maciaszczykm PTAL

PR #6936, opened 2021-04-26
Enable HTTPS with mTLS between MLA Gateway and user-cluster MLA components

### LINKED ISSUE #6493: Enable mTLS Authentication on MLA Gateways
Enable authentication on MLA gateway to verify the client identity.

rastislavs:
Enables HTTPS with mTLS authentication between the user-cluster MLA (Monitoring, Logging & Alerting) components (Prometheus, Promtail) and seed-cluster MLA Gateway.

Fixes #6493

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/resources.go
```diff
@@ -397,3 +451,84 @@ func getOrgByProjectID(ctx context.Context, client ctrlruntimeclient.Client, gra
...
+			}
+			se.Data[resources.MLAGatewayCACertKey] = cert
+			se.Data[resources.MLAGatewayCAKeyKey] = key
+
+			return se, nil
+		}
+	}
+}
+
+type tlsServerCertCreatorData interface {
```
moadqassem: seems that this interface is only there to have an abstraction between the `TemplateData` type and the certificate creator. Since the `TemplateData` has the underlying method, and this interface is only use in here I would remove it and use a `func` type instead. See the comment below.  And where the method `GatewayCertificateCreator` is called, just call the (*TemplateData). GatewayCertificateCreator() directly.

pkg/controller/seed-controller-manager/mla/resources.go
```diff
@@ -397,3 +451,84 @@ func getOrgByProjectID(ctx context.Context, client ctrlruntimeclient.Client, gra
...
+	GetMLAGatewayCA() (*resources.ECDSAKeyPair, error)
+}
+
+// GatewayCertificateCreator returns a function to create/update a secret with the MLA gateway TLS certificate.
+func GatewayCertificateCreator(data tlsServerCertCreatorData, c *kubermaticv1.Cluster) reconciling.NamedSecretCreatorGetter {
```
moadqassem: Instead of creating a new interface which is private and will be only used here, I would inline this interfaces method like this: 
```suggestion
func GatewayCertificateCreator(fn func() (*resources.ECDSAKeyPair, error), c *kubermaticv1.Cluster) reconciling.NamedSecretCreatorGetter {
```

And in the line 498 would call that function. This way you keep the scope of this abstraction bounded to the corresponding method instead of introducing a new packaged-scope interface that is used in one position.

pkg/controller/seed-controller-manager/mla/resources.go
```diff
@@ -397,3 +451,84 @@ func getOrgByProjectID(ctx context.Context, client ctrlruntimeclient.Client, gra
...
+	return func() (string, reconciling.SecretCreator) {
+		return resources.MLAGatewayCertificatesSecretName, func(se *corev1.Secret) (*corev1.Secret, error) {
+			if se.Data == nil {
+				se.Data = map[string][]byte{}
+			}
+
+			ca, err := data.GetMLAGatewayCA()
```
moadqassem:
```suggestion
			ca, err := fn()
```

moadqassem (commented):
@rastislavs Looks good, I left small comments regarding code structuring, but in general looks good.

moadqassem (commented):
/approve
/lgtm

PR #6940, opened 2021-04-26
Filtered out ARM instances from AWS size list as KKP does not support it

### LINKED ISSUE #6928: Filter AWS machine types by architecture

lsviben:
Filters our ARM instances for the AWS size list, as KKP does not support it. 

Replaces https://github.com/kubermatic/kubermatic/pull/6927, as it seems we dont need to support ARM at all here.

Fixes #6928 

```release-note
ARM instances for AWS are now being filtered out from the instance size list as KKP does not support ARM. 
```

### DIFF
--- a/pkg/handler/common/provider/aws.go
+++ b/pkg/handler/common/provider/aws.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	ec2 "github.com/cristim/ec2-instances-info"
 
@@ -244,6 +245,11 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
 			continue
 		}
 
+		// We do not support ARM, so we filter them out here
+		if isARM64Architecture(i.PhysicalProcessor) {
+			continue
+		}
+
 		sizes = append(sizes, apiv1.AWSSize{
 			Name:       i.InstanceType,
 			PrettyName: i.PrettyName,
@@ -255,6 +261,11 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
 	}
 
 	return filterAWSByQuota(sizes, quota), nil
+}
+
+func isARM64Architecture(physicalProcessor string) bool {
+	// right now there is only one Arm-based processors: Graviton2
+	return strings.Contains(physicalProcessor, "Graviton")
 }
 
 func filterAWSByQuota(instances apiv1.AWSSizeList, quota kubermaticv1.MachineDeploymentVMResourceQuota) apiv1.AWSSizeList {
--- a/pkg/handler/common/provider/aws_test.go
+++ b/pkg/handler/common/provider/aws_test.go
@@ -17,28 +17,67 @@ limitations under the License.
 package provider_test
 
 import (
-	"reflect"
+	"strings"
 	"testing"
-
-	"k8s.io/apimachinery/pkg/util/diff"
 
 	v1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	"k8c.io/kubermatic/v2/pkg/handler/common/provider"
 )
 
-func TestAWSSize(t *testing.T) {
+func TestAWSSizeARMFiltering(t *testing.T) {
 	tests := []struct {
-		name          string
-		region        string
-		resourceQuota v1.MachineDeploymentVMResourceQuota
-		expectedNames []string
+		name                 string
+		region               string
+		resourceQuota        v1.MachineDeploymentVMResourceQuota
+		unexpectedNamePrefix []string
 	}{
 		{
 			name:          "test ARM filtering",
 			region:        "eu-central-1",
 			resourceQuota: genDefaultMachineDeploymentVMResourceQuota(),
-			// t4g.small is a instance with ARM which is filtered out here
-			expectedNames: []string{"t3a.small", "t3.small"},
+			// Instance List for ARM:
+			// a1.2xlarge
+			// a1.4xlarge
+			// a1.large
+			// a1.medium
+			// a1.metal
+			// a1.xlarge
+			// c6g.12xlarge
+			// c6g.16xlarge
+			// c6g.2xlarge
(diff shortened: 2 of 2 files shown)


pkg/handler/common/provider/aws.go
```diff
@@ -257,6 +263,11 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
 	return filterAWSByQuota(sizes, quota), nil
 }
 
+func isARM64Architecture(physicalProcessor string) bool {
+	// right now there is only one Arm-based processors: Graviton2
+	return strings.Contains(physicalProcessor, "Graviton")
```
maciaszczykm, line 268: I'd stick to names here like I said in the https://github.com/kubermatic/kubermatic/pull/6927. If you want to keep it then please adjust the comment above.

irozzo-1A: https://github.com/kubermatic/kubermatic/pull/6927#discussion_r620431596

pkg/handler/common/provider/aws.go
```diff
@@ -244,6 +245,11 @@ func AWSSizes(region string, quota kubermaticv1.MachineDeploymentVMResourceQuota
 			continue
 		}
 
+		// We do not support ARM, so we filter them out here
```
irozzo-1A, line 248: I would add a unit test to `AWSSizes` to validate/document this logic

pkg/handler/common/provider/aws.go
maciaszczykm: Ignore my comment.

pkg/handler/common/provider/aws_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/diff"
+
+	v1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/handler/common/provider"
+)
+
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
maciaszczykm: This should be on top of the file.

lsviben: indeed

pkg/handler/common/provider/aws_test.go
```diff
@@ -0,0 +1,72 @@
...
+		region        string
+		resourceQuota v1.MachineDeploymentVMResourceQuota
+		expectedNames []string
+	}{
+		{
+			name:          "test ARM filtering",
+			region:        "eu-central-1",
+			resourceQuota: genDefaultMachineDeploymentVMResourceQuota(),
+			// t4g.small is a instance with ARM which is filtered out here
+			expectedNames: []string{"t3a.small", "t3.small"},
```
irozzo-1A: What would be good is doing a negative test, e.g. increase the quotas and check that no instance types of `a1`, `c6`, `r6`, `t4g` families are returned (prefix matching). WDYT?

maciaszczykm:
I've removed my invalid remark from before.

@irozzo-1A PTAL. It looks good to me.

pkg/handler/common/provider/aws_test.go
irozzo-1A: > I've removed my invalid remark from before.
> 
> @irozzo-1A PTAL. It looks good to me.

LGTM as well, just a small suggestion to improve the test

lsviben: I though about it, but I wanted to leave this test like this so there could be a test as well here which tests the quotas, bad region, or pricing which are all in this AWSSize func.

But I can change it as well, let me know what you think

irozzo-1A: Yeah, good to cover all this, but I think that in the scope of this PR the most important thing to test is that no arm-based instance types are proposed. Ideally, there should be several cases or tests to cover what you mentioned.

lsviben: ok ill change it to prefix. Later we can take care of the other, bit surprising there are no tests for any of this tbh

irozzo-1A: agree, would be better to increase the test coverage a bit

lsviben: Ahh, but there is a problem. The `AWSSize` model does not contain info about instance type family. 

```
// AWSSize represents a object of AWS size.
// swagger:model AWSSize
type AWSSize struct {
	Name       string  `json:"name"`
	PrettyName string  `json:"pretty_name"`
	Memory     float32 `json:"memory"`
	VCPUs      int     `json:"vcpus"`
	GPUs       int     `json:"gpus"`
	Price      float64 `json:"price"`
}
```

Name and prettyName are just "t3a.small" and "T3A Small", so the model would have to change and this would be a bigger change as we need to add Family to it. I can do it but Ill need to test a bit if it doesnt mess something up. @maciaszczykm for UI it wont matter right?

irozzo-1A: What I would do is just simple prefix matching on `Name` at the moment

lsviben: Ok let me see for names, I guess they will all have this `g` in the name

maciaszczykm: Everything that UI doesn't use does not matter. However, we have to remember that this produces some additional network usage so usually we don't add too many things that we are not using. One property shouldn't change much.

lsviben: nvm disregard long comment up, i mixed up some fields

irozzo-1A: Something like this to give you the idea
```
armFamilies := []map[string]struct{}{"a1": struct{}{}, "c6": struct{}{}, "r6": struct{}{}, "tg4": struct{}{}}
for name := range names {
  family := strings.Split(name, ".")[0]
  if _, ok := armFamilies[family]; ok {
    //fail test
  }
}
```

lsviben: didnt notice above before sorry. Changed the test and listed out the ARM instances so its clear

PR #6942, opened 2021-04-26
add rbac for constraints owners,editors,viewers

### LINKED ISSUE #6941: Cluster owner is not allowed to manage Cluster Kubermatic Constraints
**User Story**
When Creating/Deleting/Updating a Kubermatic Constraint as a non-admin user we get:

```
Error 403 constraints.kubermatic.k8s.io "buggy-const" is forbidden: User "lovro@loodse.com" cannot update resource "constraints" in API group "kubermatic.k8s.io" in the namespace "cluster-4g76dhzgxz"
```
even though we are the cluster `owners` and should be in the `owners` group which is allowed to manage Constraints

lsviben:
Adds proper RBAC for owner/editors/viewers groups for Kubermatic Constraints.

Fixes #6941 

```release-note
Fix issue where Kubermatic non-admin users were not allowed to manage Kubermatic Constraints
```

### DIFF
--- a/pkg/controller/master-controller-manager/rbac/sync_resource_test.go
+++ b/pkg/controller/master-controller-manager/rbac/sync_resource_test.go
@@ -19,6 +19,7 @@ package rbac
 import (
 	"context"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


pkg/controller/master-controller-manager/rbac/sync_resource_test.go
```diff
@@ -1876,3 +1882,303 @@ func TestEnsureProjectClusterRBACRoleForNamedResource(t *testing.T) {
...
+			}
+
+			{
+				var roleBindings rbacv1.RoleBindingList
+				err = fakeMasterClusterClient.List(context.Background(), &roleBindings)
+				assert.NoError(t, err)
+
+			expectedRoleBindingsLoop:
+				for _, expectedRoleBinding := range test.expectedRoleBindings {
+					// double-iterating over both slices might not be the most efficient way
```
irozzo-1A: The nested iteration is not really a concern here, but if you want to avoid it and cover the creation of unwanted resources too: you could use put both actual and wanted roles in a map keyed by name, and use `deep.Equal` to compare them. This gives you a diff that could be useful to debug test failures. e.g.

https://github.com/kubermatic/kubermatic/blob/9f2af8e327595efbe90b96fd8f44bd51dcafa7ff/pkg/controller/nodeport-proxy/envoymanager/controller_test.go#L372-L374

Not a blocker, take it as a suggestion ;-)

pkg/controller/master-controller-manager/rbac/sync_resource_test.go
```diff
@@ -1876,3 +1882,303 @@ func TestEnsureProjectClusterRBACRoleForNamedResource(t *testing.T) {
...
+			}
+
+			{
+				var roles rbacv1.RoleList
+				err = fakeMasterClusterClient.List(context.Background(), &roles)
+				assert.NoError(t, err)
+
+			expectedRolesLoop:
+				for _, expectedRole := range test.expectedRoles {
+					// double-iterating over both slices might not be the most efficient way
```
irozzo-1A: Same here

pkg/controller/master-controller-manager/rbac/sync_resource_test.go
lsviben: thats a good suggestion. Let me check the diff. 

The nested iteration is a product of the other tests in the file. I wanted to keep it similar.

lsviben: Ok, i did something similar to what you suggest, but instead of comparing whole result, i just compare those we are interested about in the test(constraint ones).

lsviben:
> /hold @lsviben should we double-check if those failures are related to your changes somehow?

I think its unrelated.

in the history of this PR the tests that failed are:
1x pre-kubermatic-e2e-aws-flatcar-1.18
2x pre-kubermatic-api-e2e
2x pre-kubermatic-e2e-aws-ubuntu-1.21
1x pre-kubermatic-opa-e2e
1x pre-kubermatic-etcd-launcher-e2e

To me looks like our standard flaky tests. Ill check this last one to see if something is up

PR #6943, opened 2021-04-27
Add Alertmanager Configuration Endpoints

### LINKED ISSUE #6892: Alertmanager Config API
Add API endpoints for AlertmanagerConfig. and API will convert AlertmanagerConfig from user to a secret in the user-cluster namespace, and this secret will be referenced in AlertmanagerConfig CRD, and will be used by Alertmanager Config controller to config Alertmanager via HTTP requests.

jiachengxu:
This PR adds `GET`, `PUT` and `DELETE` endpoints for alertmanager configuration.
Fixes #6892 

**Documentation**:

```release-note
Add v2 endpoints to manage alertmanager configuration:
GET /api/v2/projects/{project_id}/clusters/{cluster_id}/alertmanager/config
PUT /api/v2/projects/{project_id}/clusters/{cluster_id}/alertmanager/config
DELETE /api/v2/projects/{project_id}/clusters/{cluster_id}/alertmanager/config
```


pkg/controller/master-controller-manager/rbac/sync_resource.go
```diff
@@ -574,6 +587,29 @@ func shouldSkipRBACRoleForClusterNamespaceResource(projectName string, cluster *
 	return false, generatedRole, nil
 }
 
+// shouldSkipRBACRoleForClusterNamespaceNamedResource will tell you if you should skip the generation of Role/Rolebinding of a named resource or not,
+// because for some groupPrefixes we actually don't create Role
+//
```
moadqassem: Nit: Please get rid of this empty line

pkg/controller/master-controller-manager/rbac/sync_resource.go
```diff
@@ -574,6 +587,29 @@ func shouldSkipRBACRoleForClusterNamespaceResource(projectName string, cluster *
...
+func shouldSkipRBACRoleForClusterNamespaceNamedResource(projectName string, cluster *kubermaticv1.Cluster, resourceName, policyAPIGroups, policyResource, kind, groupPrefix string) (bool, *rbacv1.Role, error) {
+	generatedRole, err := generateRBACRoleForClusterNamespaceNamedResource(
+		cluster,
+		GenerateActualGroupNameFor(projectName, groupPrefix),
+		policyAPIGroups,
+		policyResource,
+		kind,
+		resourceName,
+	)
+
```
moadqassem: Nit: Please get rid of this empty line

pkg/handler/middleware/middleware.go
```diff
@@ -444,3 +448,46 @@ func getConstraintProvider(clusterProviderGetter provider.ClusterProviderGetter,
...
+func getAlertmanagerProvider(clusterProviderGetter provider.ClusterProviderGetter, alertmanagerProviderGetter provider.AlertmanagerProviderGetter, seedsGetter provider.SeedsGetter, seedName, clusterID string) (provider.AlertmanagerProvider, error) {
+	seeds, err := seedsGetter()
+	if err != nil {
+		return nil, err
+	}
+
+	if clusterID != "" {
+		for _, seed := range seeds {
+			clusterProvider, err := clusterProviderGetter(seed)
+			if err != nil {
```
moadqassem, line 476: What if the error wasn't 404?

pkg/handler/v2/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,217 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
moadqassem:
```suggestion
Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```

pkg/handler/v2/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		alertmanagerProvider := ctx.Value(middleware.AlertmanagerProviderContextKey).(provider.AlertmanagerProvider)
+		alertmanager, config, err := alertmanagerProvider.Get(c, userInfo)
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
+		return convertInternalToAPIAlertmanager(alertmanager, config), nil
```
moadqassem: This chunk here isn't doing anything distinct, I think would be better if you wrap it instead of duplicating in the next 3 handlers

pkg/handler/v2/routes_v2.go
```diff
@@ -321,6 +322,19 @@ func (r Routing) RegisterV2(mux *mux.Router, metrics common.ServerMetrics) {
 		Path("/projects/{project_id}/clusters/{cluster_id}/addons/{addon_id}").
 		Handler(r.deleteAddon())
 
+	// Defines a set of HTTP endpoints for managing alertmanager
+	mux.Methods(http.MethodGet).
+		Path("/projects/{project_id}/clusters/{cluster_id}/alertmanager/config").
+		Handler(r.getAlertmanager())
+
+	mux.Methods(http.MethodPost).
```
moadqassem: Shouldn't be there a patch method as well? you how do you update the alermanager configs?

pkg/provider/kubernetes/alertmanager.go
```diff
@@ -0,0 +1,145 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
moadqassem:
```suggestion
Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```

pkg/provider/kubernetes/alertmanager.go
```diff
@@ -0,0 +1,145 @@
...
+		Namespace: cluster.Status.NamespaceName,
+	}, configSecret); err != nil {
+		return nil, nil, err
+	}
+	return alertmanager, configSecret, nil
+}
+
+// Create only updates an Alertmanager object and corresponding config Secret since Alertmanager and Secret will
+// be created by alertmanager configuration controller.
+func (p *AlertmanagerProvider) Create(expectedAlertmanager *kubermaticv1.Alertmanager, expectedSecret *corev1.Secret, userInfo *provider.UserInfo) (*kubermaticv1.Alertmanager, *corev1.Secret, error) {
```
moadqassem: I would say change this method name to update, since it is not really creating resources, it updates them

pkg/controller/master-controller-manager/rbac/sync_resource.go
jiachengxu: Removed

pkg/controller/master-controller-manager/rbac/sync_resource.go
jiachengxu: Removed

pkg/handler/middleware/middleware.go
jiachengxu: Actually, I didn't look into this in details, just copied from `getAddonProvider` now I think it makes more sense to return something like `KubernetesErrorToHTTPError` https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v1/common/error.go#L27
WDYT?

pkg/handler/v2/alertmanager/alertmanager.go
jiachengxu: Fixed, thanks

pkg/handler/v2/alertmanager/alertmanager.go
jiachengxu: Done

pkg/handler/v2/routes_v2.go
jiachengxu: So currently the alertmanager API struct only contains a config field which is a byte slice, and the user can POST the new configuration to the endpoint to update the config. But you are right, we will need a PATCH endpoint once we have more fields in alertmanger API, is that ok for you that we add that later?

pkg/provider/kubernetes/alertmanager.go
jiachengxu: Done

pkg/provider/kubernetes/alertmanager.go
jiachengxu: This is actually a good idea, I felt a bit weird when I implemented this `Create` method, WDYT @lsviben @rastislavs @aborilov ?
Also I feel like calling the endpoint `CreateEndpoint` is a bit weird, how about I rename the `CreateEndpoint` to something like `PostEndpoint`?

pkg/handler/middleware/middleware.go
lsviben: Yes, `KubernetesErrorToHTTPError` is better

pkg/handler/v2/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,214 @@
...
+	bodyMap := map[string]interface{}{}
+	if err := yaml.Unmarshal(req.Body.Spec.Config, &bodyMap); err != nil {
+		return fmt.Errorf("can not unmarshal yaml configuration: %w", err)
+	}
+	return nil
+}
+
+// deleteAlertmanagerReq defines HTTP request for deleting alertmanager
+// swagger:parameters deleteAlertmanager
+type deleteAlertmanagerReq struct {
```
lsviben: you can merge `deleteAlertmanagerReq` and `getAlertmanagerReq` into `alertManagerReq` as they are the same, you will save on less decode funcs, and `createAlertmanagerReq` can embed it.

pkg/handler/v2/alertmanager/alertmanager_test.go
```diff
@@ -0,0 +1,410 @@
...
+				test.GenTestSeed(),
+				test.GenDefaultCluster(),
+			),
+			ExistingAlertmanager: test.GenAlertmanager(test.GenDefaultCluster().Status.NamespaceName,
+				testAlertmanagerConfigSecretName),
+			ExistingAPIUser:    test.GenDefaultAPIUser(),
+			ExpectedHTTPStatus: http.StatusNotFound,
+		},
+		{
+			Name:      "scenario 4: user john can not get alertmanager that belongs to bob's cluster",
```
lsviben, line 110: you can add a test that Admin can get an alertManager that  belongs to bobs cluster

pkg/handler/v2/alertmanager/alertmanager_test.go
```diff
@@ -0,0 +1,410 @@
...
+		{
+			Name:      "scenario 1: get alertmanager that belongs to the given cluster",
+			ProjectID: test.GenDefaultProject().Name,
+			ClusterID: test.GenDefaultCluster().Name,
+			ExistingKubermaticObjects: test.GenDefaultKubermaticObjects(
+				test.GenTestSeed(),
+				test.GenDefaultCluster(),
+			),
+			ExistingAlertmanager: test.GenAlertmanager(test.GenDefaultCluster().Status.NamespaceName,
+				testAlertmanagerConfigSecretName),
```
lsviben: NIT: you can just put this up as its a KubermaticObject (disregard the weird formatting for some reason i couldnt make it nice)
```suggestion
			ExistingKubermaticObjects: test.GenDefaultKubermaticObjects(
				test.GenTestSeed(),
				test.GenDefaultCluster(),
				test.GenAlertmanager(test.GenDefaultCluster().Status.NamespaceName,
	                        testAlertmanagerConfigSecretName),
			),
```

pkg/handler/v2/alertmanager/alertmanager_test.go
```diff
@@ -0,0 +1,410 @@
...
+			Body:               generateRequestBody([]byte(testAlertmanagerConfig)),
+			ExistingAPIUser:    test.GenDefaultAPIUser(),
+			ExpectedHTTPStatus: http.StatusInternalServerError,
+		},
+		{
+			Name:      "scenario 5: user john can not get alertmanager that belongs to bob's cluster",
```
lsviben: NIT: `...can not get...` -> `...can not create...`

pkg/provider/kubernetes/alertmanager.go
lsviben: Maybe just make a `UpdateEndpoint` with PUT. You are right its weird to call it Create when it just updates.

pkg/provider/kubernetes/alertmanager.go
```diff
@@ -0,0 +1,145 @@
...
+	secret.Data = expectedSecret.Data
+	if err := impersonationClient.Update(ctx, secret); err != nil {
+		return nil, nil, err
+	}
+	return alertmanager, secret, nil
+}
+
+// Delete resets corresponding config Secret of Alertmanager object to the default config. Note that Delete will not remove
+// Alertmanager object, it will only delete the config secret, and alertmanager controller will create default config secret.
+func (p *AlertmanagerProvider) Delete(cluster *kubermaticv1.Cluster, userInfo *provider.UserInfo) error {
```
lsviben: maybe the endpoint which uses this should be called reset instead of delete. Although it does delete something the purpose of it is to reset the config to default. WDYT?

pkg/handler/v2/alertmanager/alertmanager.go
jiachengxu: Thanks for pointing out this, I put them in different requests because I am not sure if they will have different query parameters in the future.
What do you think? If we embed them now, and in the future, we need to split them, will it affect the user side?

pkg/handler/v2/alertmanager/alertmanager_test.go
jiachengxu: Currently, this endpoint doesn't work for admins because I didn't embed a `privilegedAlertmanagerProvider` there, and it only works for cluster owners/editors/viewers.

pkg/provider/kubernetes/alertmanager.go
jiachengxu: Thanks @lsviben, `PUT` is also an option. I used `POST` because I would like to keep consistent with cortex alertmanger API: https://cortexmetrics.io/docs/api/#set-alertmanager-configuration
And from user perspective, they are creating/deleting alertmanager configuration, it is just our underneath implementation is doing updating and reseting

pkg/handler/v2/alertmanager/alertmanager_test.go
lsviben: ahh ok. We should plan it as well for the future, because "admins" in KKP are like gods and we make little backdoors for them everywhere. Its not pretty and hopefully when we make our big RBAC/user management rehaul it will get removed

pkg/handler/v2/alertmanager/alertmanager.go
lsviben: It would affect the guys who use our generated client, otherwise the pure API should not be affected. 

But if we expect that they will have different params in the future, lets keep them separated

pkg/handler/v2/alertmanager/alertmanager_test.go
jiachengxu: Fixed, thanks!

pkg/handler/v2/alertmanager/alertmanager_test.go
jiachengxu: You are right, fixed

pkg/handler/middleware/middleware.go
jiachengxu: Done

pkg/provider/kubernetes/alertmanager.go
lsviben: I see. But if they dont Create anything, they will anyway GET a configuration, the default one, right? 

Same with delete, when they delete it, and then GET, they will get the default config?

pkg/handler/v2/alertmanager/alertmanager.go
jiachengxu: Thanks for explaining this, I would keep them separate for now :)

pkg/handler/v2/alertmanager/alertmanager_test.go
jiachengxu: Yeah, this definitely makes sense, I have created an issue for this: #6966.

pkg/provider/kubernetes/alertmanager.go
jiachengxu: Yeah you are right, I have replaced `POST` with a `PUT` and also renamed `Create` to `Update` and `Delete` to `Reset` :)

moadqassem (commented):
/approve
/lgtm

PR #6957, opened 2021-04-28
Check MLA gateway before deleting Alertmanager configuration.

jiachengxu:
This PR adds an MLA gateway check before deleting the Alertmanager configuration in alertmanager controller, otherwise, it returns no such host consistently.

Note that it can be the case that mla gateway service is deleted before the controller cleans alertmanager configuration up. For fixing that controller needs to send requests to cortex alertmanager endpoints with `X-Scope-OrgID` header directly. The issue is created here: #6959 

Fixes #

**Documentation**:

```release-note
NONE
```


xmudrii (approved):
/approve

PR #6722, opened 2021-03-18
Implement UserProjectBinding-Sync Controller

jiachengxu:
**What this PR does / why we need it**:
This PR implements a sync controller in the master controller manager to propagate `UserProjectBinding` objects from Master Cluster to Seed Cluster.
This is needed for MLA (Monitoring, Logging, Alerting) of user clusters. MLA controller in seed controller manager will map KKP User-Project to Grafana User-Organization.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #6707 

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6961, opened 2021-04-28
Add SIG App Management as OWNERS of MLA packages

xmudrii:

The SIG App Management folks are doing an amazing job with the new MLA controller, however, they need approval from SIG  Cluster Management for every PR.

This can be a big blocker and considering that they have created the controller and that they know how it should look and work like, it makes sense for them to own the MLA controller.

This PR adds SIG App Management as OWNERS to the following packages:

* `pkg/controller/seed-controller-manager/mla`
* `pkg/controller/user-cluster-controller-manager/resources/resources/mla`

They will still need approval for SIG Cluster Management for larger PRs, but I think that this will help at least for smaller PRs.

```release-note
NONE
```

/assign @shaase-ctrl @moadqassem @kron4eg @moelsayed 
cc @rastislavs @jiachengxu @aborilov 
/hold
for discussion

### DIFF
--- a/pkg/controller/seed-controller-manager/mla/OWNERS
+++ b/pkg/controller/seed-controller-manager/mla/OWNERS
@@ -1,0 +1,13 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+  - sig-app-management
+
+reviewers:
+  - sig-app-management
+
+labels:
+  - sig/app-management
+
+options:
+  no_parent_owners: true
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/mla/OWNERS
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/mla/OWNERS
@@ -1,0 +1,13 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+  - sig-app-management
+
+reviewers:
+  - sig-app-management
+
+labels:
+  - sig/app-management
+
+options:
+  no_parent_owners: true


pkg/controller/seed-controller-manager/mla/OWNERS
```diff
@@ -0,0 +1,13 @@
...
+  - sig-app-management
+
+reviewers:
+  - sig-app-management
+
+labels:
+  - sig/app-management
+
+options:
+  no_parent_owners: true
```
xmudrii, line 13: I'm not sure do we want to keep this option or not. IMO, it makes sense to keep it, but it means that the SIG App Management would have to approve every change to the MLA controller.

kron4eg: >SIG App Management would have to approve every change to the MLA controller.

Isn't that's the point?

xmudrii: The main point is that they can approve their PRs themselves. This option means that we (SIG Cluster Management) wouldn't be able to approve changes to the MLA packages. It makes sense because the SIG App Management folks will mostly like want to be aware of changes to the MLA packages, but it might complicate the review process (although I don't think that's a problem).

jiachengxu: IIRC this option means if third-party folks make changes in MLA packages, they will need approvals from both sig cluster management and sig app management.
and for folks in sig app management, we will still need approval from sig cluster management
For example, this project-sync controller is owned by sig app: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/master-controller-manager/project-sync/OWNERS
but if you check this message: https://github.com/kubermatic/kubermatic/pull/6722#issuecomment-804154076
It still needs approval from sig cluster management

jiachengxu: so I think this option will require additional approval from sig-app, but will still require approval from sig-cluster

kron4eg (approved):

pkg/controller/seed-controller-manager/mla/OWNERS
xmudrii: @jiachengxu To my knowledge, that option is used to restrict folks defined in parent OWNERS file(s) to approve changes in that package. I searched for some docs on that option to confirm this, but I wasn't able to find anything.

Regarding #6722, I believe that the SIG Cluster Management approval was required because the PR changed some files outside the `project-sync` package, owned by SIG Cluster Management (such as `pkg/resources/reconciling/zz_generated_reconcile.go`).

jiachengxu: @xmudrii You are right, thanks for explaining this to me :)

moadqassem (commented):
/approve
/lgtm

PR #6970, opened 2021-04-30
Add network configuration for user clusters in API

### LINKED ISSUE #6968: Make pod and service CIDRs configurable for user clusters
**User Story**

The goal of this PR is to make pod and service CIDR configurable when creating a user cluster. At the moment the values are defaulted:

https://github.com/kubermatic/kubermatic/blob/1fe410ffcd14a29dd1ce0636a5702105c8f47366/pkg/controller/seed-controller-manager/kubernetes/pending.go#L113-L125

But the API and dashboard do not provide a mean of specifying them.

This could bring problems in environments where the default IP ranges overlap with the ones used for other infrastructure. 
Another notable consequence is that this makes KKP in KKP not possible (i.e. using a KKP user cluster as seed cluster for another KKP installation), consider the following scenario:

1. OpenVPN client running on a worker node in the pod network (e.g. IP `172.25.2.19`) tries to connect to OpenVPN server passing from node-port proxy.
2. The envoy container running on seed cluster side using same pod CIDR (e.g. IP `172.25.0.24`) receives the connection from OpenVPN client and connects to OpenVPN server. (e.g. IP `172.25.1.10`)
3. OpenVPN server container receives the SYN packet but SYN/ACK never reach the destination (i.e. envoy container), the reason is the following routing table entry, that sends packets in the `172.25.0.0/16` to the `tun0` interface to be sent on the workers network.
```
172.25.0.0/16 via 10.20.0.2 dev tun0
```

**Acceptance criteria**
Having the possibility of specifying pod and service CIDRs when creating a cluster through KKP API.

irozzo-1A:
This PR is to extend the cluster API to make it possible to configure some user cluster network parameters:
- Pods CIDR
- Services CIDR
- Proxy mode

Fixes #6968

The dashboard will require a change as well.

**Documentation**:

```release-note
Add network configuration to user cluster API.
```


pkg/validation/cluster_test.go
```diff
@@ -22,9 +22,10 @@ import (
 	"strings"
 	"testing"
 
-	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
-
+	"k8s.io/apimachinery/pkg/util/validation/field"
```
zreigz, line 27: please swap imports: kubermatic then k8s

pkg/webhook/cluster/mutation/mutation_test.go
```diff
@@ -20,15 +20,18 @@ import (
...
 	"testing"
-	"text/template"
 
 	logrtesting "github.com/go-logr/logr/testing"
+	"github.com/go-test/deep"
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/resources"
 
+	jsonpatch "gomodules.xyz/jsonpatch/v2"
```
zreigz, line 29: move this import from k8s

PR #6971, opened 2021-04-30
Vault integration for ssh access [Proposal]

### LINKED ISSUE #6671: Add user ssh keys to worker nodes from Vault
**User Story**
As a user I would like to add user ssh keys to the worker nodes of a specific cluster by creating a secret in vault. Those ssh keys should be rotated on the worker nodes in a specific cluster. 

**Acceptance Criteria**
- User should provide some vault configurations where the agent can connect and read the secret from.
- User SSH Key agent is able to fetch the vault secret and add it to the worker nodes.
- If users specify that they wants to use vault instead of the UserSSHKeys list in KKP, ssh keys should be read from vault only.

emmanuel-kubermatic:
Proposes an integration pattern for using secret managers like Vault in the application

Fixes #6671

Rendered version: https://github.com/kubermatic/kubermatic/blob/d0fcc01ae2bbbfb028a7093bdaac76e822f8cedd/docs/proposals/secret-manager-store-api.md

**Documentation**:

```release-note
Adds support for vault and other secret manager integrations 
Support for frontend authentication for other secret managers
```

### DIFF
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -16,6 +16,10 @@ The proposed fix for the issue [6671](https://github.com/kubermatic/kubermatic/i
 
 A kubermatic user can assign ssh keys to clusters at creation time via the machine controller through cloud-init, but the user ssh agent also assists with updating the keys at runtime via the usersshkeys secret.
 
+The keys are updated mainly via the UserSSHKeys object in the master cluster, so not directly on the secrets. Once the list is updated in the master, users then can attach keys to user clusters which would then create secrets that have the keys in their data field. 
+
+Then those secrets are read from the seed by a user ssh key agent that would always rewrite the ssh keys on each UserSSHKey update.
+
 This approach is great but adds a limitation where the end user has to manually copy credentials to kubermatic usersshkey secrets or added via the API for each cluster's worker nodes authorized_keys.
 
 The new approach documented within this proposal provides the ability to use [SSH CA certificates](https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates), avoiding the burden of copying each user key individually.
@@ -51,7 +55,7 @@ Distribution of the keys is out of scope from the master to user controller, the
 
 Since the updates to use a TrustedCA require we update the SSH configuration and restart the SSH daemon, this would not be possible if the cluster is already created without the configuration.
 
-During a revocation / rotation, the user would update the `vault-ca` secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
+During a revocation / rotation, the user would up date the `vault-ca` secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
 
 The update would now be handled by the `user-ssh-key-agent` controller, instead of the machine controller and reconciled on any changes to the `vault-ca` secret in the `kube-system` namespace.
 
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -39,21 +39,19 @@ This includes the following process
     # Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
     CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
     ```
- - Restart the SSH daemon to enable configuration on the cluster
- - Distribute the public key to the workers
+ - Starts the SSH daemon with the above configuration on the cluster to use only keys signed by the stated TrustedCA.
+ - Distribute the public key to the workers upon further updates with the help of the `user-ssh-agent`.
 
 > NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
 > Management of the CA keys are the responsibility of the admin managing the KKP cluster.
 
-Since the signing public key is to be replicated across multiple instances, we'd apply the same approach used previously for the User SSH keys.
-
-The keys would be updated but would require an update to the SSHd config written by the machine-controller on cluster creation. 
-
-Only the master-controller-manager would write these configurations.
+The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, provided the CA key is specified on cluster creation.
 
 Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster.
 
-During a revocation / rotation, the user would update the vault-ca secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
+During a revocation / rotation, the user would update the `vault-ca` secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
+
+The update would now be handled by the `user-ssh-key-agent` controller, instead of the machine controller and reconciled on any changes to the `vault-ca` secret in the `kube-system` namespace.
 
 
 ## Alternatives considered
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -47,7 +47,9 @@ This includes the following process
 
 The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, provided the CA key is specified on cluster creation.
 
-Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster.
+Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster. The reason for this is to remove the issues around the secret being created after cluster creation.
+
+Since the updates to use a TrustedCA require we update the SSH configuration and restart the SSH daemon, this would not be possible if the cluster is already created without the configuration.
 
 During a revocation / rotation, the user would update the `vault-ca` secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
 
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -1,0 +1,69 @@
+# Secrets Store Proposal
+
+**Author**: Emmanuel Bakare (@emmanuel-kubermatic)
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+The proposed fix for the issue [6671](https://github.com/kubermatic/kubermatic/issues/6671) 
+
+- Define secret using Vault signed certificates to assist with SSH access
+
+## Non-Goals
+
+
+## Motivation and Background
+
+A kubermatic user can assign ssh keys to clusters at creation time via the machine controller through cloud-init, but the user ssh agent also assists with updating the keys at runtime via the usersshkeys secret.
+
+This approach is great but adds a limitation where the end user has to manually copy credentials to kubermatic usersshkey secrets or added via the API for each cluster's worker nodes authorized_keys.
+
+The new approach documented within this proposal supports the ability to hard-code authorized keys but rather, alongside another option to trust SSH CA public keys.
+
+This enables authentication via signed keys which can be provided via vault with a lease time configurable by the user.
+
+Vault provides the facility for storing and signing the keys used for authentication.
+
+
+## Implementation
+
+[Signed SSH Certificates](https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates) would be used to implement this.
+
+This includes the following process
+ - Generating a CA key pair 
+ - Configure the SSH daemon to trust the public key via TrustedUserCAKeys configuration, example:
+    ```text
+    .../etc/ssh/sshd_config
+    TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+    
+    # Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
+    CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+    ```
+ - Restart the SSH daemon to enable configuration on the cluster
+ - Distribute the public key to the workers
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+Since the signing public key is to be replicated across multiple instances, we'd apply the same approach used previously for the User SSH keys.
+
+The keys would be updated but would require an update to the SSHd config written by the machine-controller on cluster creation. 
+
+Only the master-controller-manager would write these configurations.
+
+Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster.
+
+During a revocation / rotation, the user would update the vault-ca secret configuration and that would be reconciled and updated in the clusters trusted ca config in ssh config at `/etc/ssh/TrustedCA.pem` via a hostPath mount point.
+
+
+## Alternatives considered
+
+No alternatives as this process solves all desired requirements.
+
+
+## Task & effort:
+*Specify the tasks and the effort in days (samples unit 0.5days) e.g.*
+* Define proposal and outline scope of the vault manager implementation - 3d
+* Test POC of the process documented in the proposal - 3d
+* Review changes and update process as desired - 1d
+* Add SSH configuration to machine-controller cloud-init template - 0.1d
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -45,7 +45,7 @@ This includes the following process
 > NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
 > Management of the CA keys are the responsibility of the admin managing the KKP cluster.
 
-The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, provided the CA key is specified on cluster creation.
+The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, the provided CA key is specified on cluster creation.
 
 Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster. The reason for this is to remove the issues around the secret being created after cluster creation.
 
--- a/docs/proposals/secret-manager-store-api.md
+++ b/docs/proposals/secret-manager-store-api.md
@@ -18,7 +18,7 @@ A kubermatic user can assign ssh keys to clusters at creation time via the machi
 
 This approach is great but adds a limitation where the end user has to manually copy credentials to kubermatic usersshkey secrets or added via the API for each cluster's worker nodes authorized_keys.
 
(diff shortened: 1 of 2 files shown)


docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+type K8SSecretStore struct{
+	Store
+	// other configs
+}
+```
+
+The approach would this can affect the way we reconcile secrets, even  on Kubernetes as API access to the secrets store could be modified to use those interfaces.
+
+A similar approach can be used already in here [user-ssh-agent](./pkg/controller/usersshkeysagent/usersshkeys_controller.go).
```
irozzo-1A: Nit: link is broken.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+Adds support for vault related read access within the kubermatic source
+
+## Non-Goals
+Adds support for vault access and a central store for accessing data
+
+## Motivation and Background
+
+Currently, there's a user ssh agent which manages credentials via fsnotify events and kubernetes secrets.
+
+This approach is great but adds a limitation where the end user has to manually copy credentials to ssh secrets or add them via the API.
```
irozzo-1A: Actually, the current flow is the following:

1. KKP users add secret with KKP UI or API, and [UserSSHKey](https://github.com/kubermatic/kubermatic/blob/9f2af8e327595efbe90b96fd8f44bd51dcafa7ff/pkg/crd/kubermatic/v1/sshkeys.go#L31-L50) get created on the master cluster
2. [master-controller-manager ensures](https://github.com/kubermatic/kubermatic/blob/9f2af8e327595efbe90b96fd8f44bd51dcafa7ff/pkg/controller/master-controller-manager/usersshkeyssynchronizer/controller.go#L199-L206) that secrets containing the certificates are present in the corresponding seed clusters.
3. [user-cluster-controller-manager](https://github.com/kubermatic/kubermatic/blob/9f2af8e327595efbe90b96fd8f44bd51dcafa7ff/pkg/controller/user-cluster-controller-manager/resources/reconciler.go#L69) sync those secrets from seed to user clusters
4. The agents running on each worker [fetch](https://github.com/kubermatic/kubermatic/blob/9f2af8e327595efbe90b96fd8f44bd51dcafa7ff/pkg/controller/usersshkeysagent/usersshkeys_controller.go#L154-L166) the certs from the secret and update the authorized keys.

The key thing to decide to me is at which stage do where do we want to integrate Vault.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+    init(config interface{}) error
+}
+
+type K8SSecretStore struct{
+	Store
+	// other configs
+}
+```
+
+The approach would this can affect the way we reconcile secrets, even  on Kubernetes as API access to the secrets store could be modified to use those interfaces.
```
irozzo-1A: Similar to my previous comment, where do we want to introduce this abstraction? It's not really clear to me.

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: I've seen that we do not need access via the master-controller-manager for now as we discussed.

Editing the cloud-init configuration is also not needed as we can create the file via a hostPath mount which would then use the VaultSecret named `vaultsignedca` to create a TrustedCA file in the `/etc/ssh` directory.

I have an idea that we can likewise overwrite the configuration via a secret mount instead of cloud-init?

I'd update that in the proposal when we resolve the ideas around this

pkg/controller/usersshkeysagent/usersshkeys_controller.go
```diff
@@ -215,32 +220,41 @@ func createBuffer(data map[string][]byte) (*bytes.Buffer, error) {
...
 	if err != nil {
-		return fmt.Errorf("failed describing the authorized_keys file in path %s: %v", userHome, err)
+		return fmt.Errorf("failed describing the directory in path %s: %v", fileDir, err)
 	}
 
 	uid := fileInfo.Sys().(*syscall.Stat_t).Uid
 	gid := fileInfo.Sys().(*syscall.Stat_t).Gid
 
-	if err := os.Chown(path, int(uid), int(gid)); err != nil {
+	if err := ChownR(path, int(uid), int(gid)); err != nil {
```
emmanuel-kubermatic: Is it necessary to update the permissions for the directory @irozzo-1A ?

Kindly give it a look if it's fine

irozzo-1A:
@emmanuel-kubermatic I would keep the proposal and the implementation PRs separated :-)

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,62 @@
...
+## Non-Goals
+Adds support for vault access and a central store for accessing data
+
+## Motivation and Background
+
+Currently, there's a user ssh agent which manages credentials via fsnotify events and kubernetes secrets.
+
+This approach is great but adds a limitation where the end user has to manually copy credentials to ssh secrets or add them via the API.
+
+The new approach documented within this proposal adds the ability to hard-code authorized keys but rather, add another option to lease out ssh certificates via vault which can be used to access the necessary clusters.
```
irozzo-1A: Actually, the solution is not strictly coupled with Vault. We could just mention that we will add support for [TrustedUserCAKeys](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-distributing_and_trusting_ssh_ca_public_keys), it's not something Vault specific, even if Vault provides a convenient way of storing CA cert and signing keys:
https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,62 @@
...
+.../etc/ssh/sshd_config
+TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+
+# Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+```
+
+Only the master-controller-manager would have access to the Vault instance deployed either within or outside the cluster and a PATH entry to the desired secret engine would be used to fetch it.
+
+Once fetched, it would be used to create a [UserSSHKey](https://github.com/kubermatic/kubermatic/pkg/crd/kubermatic/v1/sshkeys.go) CRD along with the list of clusters it should be synced to and from there, replicated as needed.
```
irozzo-1A: Don't think we should create a `UserSSHKey` to distribute the public CA key.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,62 @@
...
+
+```text
+.../etc/ssh/sshd_config
+TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+
+# Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+```
+
+Only the master-controller-manager would have access to the Vault instance deployed either within or outside the cluster and a PATH entry to the desired secret engine would be used to fetch it.
```
irozzo-1A: I think that in the first instance we could even consider providing the public CA certificate with a secret to the master/seed cluster, this way we don't need any integration with vault, and we could do that in a second iteration. WDYT?

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,62 @@
...
+## Alternatives considered
+
+No alternatives as this process solves all desired requirements.
+
+
+## Task & effort:
+*Specify the tasks and the effort in days (samples unit 0.5days) e.g.*
+* Implement Store Interface - 0.1d
+* Write Store APIs for Kubernetes Secrets Store - 1d
+* Implement Vault integration APIs (Using the agent injector might have limitations) - 2d
```
irozzo-1A: I think those are not up-to-date

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,62 @@
...
+
+The keys would be updated but an update to the SSHd config. 
+
+```text
+.../etc/ssh/sshd_config
+TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+
+# Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
+CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+```
```
irozzo-1A: I think it's quite important to clarify how do we intend to configure the SSH daemon on the workers e.g. using user data scripts and cloud-init directives through machine-controller.

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: I've updated that in the proposal

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: That was the previous approach

I've updated that now with the process you did provide

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: That could pose a risk if the keys aren't changed.

Better for them to provide their own keys from the start to avoid security concerns

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: They're not

I've  updated it now

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: Done.

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: I've removed it since we no longer interface with the user ssh agent

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+## Non-Goals
+
+
+## Motivation and Background
+
+A kubermatic user can assign ssh keys to clusters at creation time via the machine controller through cloud-init, but the user ssh agent also assists with updating the keys at runtime via the usersshkeys secret.
+
+This approach is great but adds a limitation where the end user has to manually copy credentials to kubermatic usersshkey secrets or added via the API for each cluster's worker nodes authorized_keys.
+
+The new approach documented within this proposal supports the ability to hard-code authorized keys but rather, alongside another option to trust SSH CA public keys.
```
irozzo-1A: I think this should be:
```suggestion
The new approach documented within this proposal provides the ability to use [SSH CA certificates](https://www.vaultproject.io/docs/secrets/ssh/signed-ssh-certificates), avoiding the burden of copying each user key individually.
```

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+    ```
+ - Restart the SSH daemon to enable configuration on the cluster
+ - Distribute the public key to the workers
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+Since the signing public key is to be replicated across multiple instances, we'd apply the same approach used previously for the User SSH keys.
+
+The keys would be updated but would require an update to the SSHd config written by the machine-controller on cluster creation.
```
irozzo-1A: To make things clearer, I would say that the machine controller will make sure of setting up the SSHd configuration and provisioning the initial public certificate, and it will be possible to rotate the certificate by using the ssh-agent.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+ - Distribute the public key to the workers
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+Since the signing public key is to be replicated across multiple instances, we'd apply the same approach used previously for the User SSH keys.
+
+The keys would be updated but would require an update to the SSHd config written by the machine-controller on cluster creation. 
+
+Only the master-controller-manager would write these configurations.
```
irozzo-1A: Not sure to understand your point here.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+
+Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster.
```
irozzo-1A: The problem with this approach is that if the secret is created after cluster creation, the initial machine deployment will be created without CA cert enabled, isn't it?

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: I've removed that part and changed some of the wordings based on your previous comment here: https://github.com/kubermatic/kubermatic/pull/6971#discussion_r638807366

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: Done

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: I've stated that in the proposal

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+The proposed fix for the issue [6671](https://github.com/kubermatic/kubermatic/issues/6671) 
+
+- Define secret using Vault signed certificates to assist with SSH access
+
+## Non-Goals
+
+
+## Motivation and Background
+
+A kubermatic user can assign ssh keys to clusters at creation time via the machine controller through cloud-init, but the user ssh agent also assists with updating the keys at runtime via the usersshkeys secret.
```
moadqassem: `The keys are updated mainly via the UserSSHKeys object in the master cluster, so not directly on the secrets. Once the list is updated in the master, users then can attach keys to user clusters which would then create secrets that have the keys in their data field. Then those secrets are read from the seed by a user ssh key agent that would always rewrite the ssh keys on each UserSSHKey update.`

This explains it a bit more, because the real canonical truth is always coming from the UserSSHKeys in the master. Would say to formulate the text above in your style and add it here.

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+ - Distribute the public key to the workers upon further updates with the help of the `user-ssh-agent`.
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, provided the CA key is specified on cluster creation.
+
+Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster. The reason for this is to remove the issues around the secret being created after cluster creation.
+
+Since the updates to use a TrustedCA require we update the SSH configuration and restart the SSH daemon, this would not be possible if the cluster is already created without the configuration.
```
moadqassem: `Since the updates to use a TrustedCA require ` could be formulated in a better way?

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,69 @@
...
+    # Needed for OpenSSH Server 8.1 and older, ssh-rsa is removed from the default supported CA algorithms
+    CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+    ```
+ - Starts the SSH daemon with the above configuration on the cluster to use only keys signed by the stated TrustedCA.
+ - Distribute the public key to the workers upon further updates with the help of the `user-ssh-agent`.
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, provided the CA key is specified on cluster creation.
```
moadqassem:
```suggestion
The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, the provided CA key is specified on cluster creation.
```

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,70 @@
...
+    ```
+ - Starts the SSH daemon with the above configuration on the cluster to use only keys signed by the stated TrustedCA.
+ - Distribute the public key to the workers upon further updates with the help of the `user-ssh-agent`.
+
+> NOTE: This allows the rollout of new CA key pairs for security reasons e.g rotation, revocation etc
+> Management of the CA keys are the responsibility of the admin managing the KKP cluster.
+
+The machine controller would create the CA pem file and configure the SSH daemon to add the `TrustedCAKeys` configuration, the provided CA key is specified on cluster creation.
+
+Distribution of the keys is out of scope from the master to user controller, the secret for the CA would be created on the cluster. The reason for this is to remove the issues around the secret being created after cluster creation.
```
irozzo-1A: The problem to me is that we should be able to provide the public certificate at cluster creation time as the initial machinedeployment should contain it.

> the secret for the CA would be created on the cluster

I'm not sure I understand which cluster you are referring to. If we can skip the distribution of the public keys to start simple it's fine, but take into account that we still need to configure the machines to trust the CA public key so we need it somehow when the md is deployed, isn't it?

This is the controller we use to deployment: https://github.com/kubermatic/kubermatic/blob/9ade0be3abcf1aa63086a533bd232c5d317060ed/pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go#L126

docs/proposals/secret-manager-store-api.md
```diff
@@ -0,0 +1,70 @@
+# Secrets Store Proposal
```
irozzo-1A: The title and the filename don't seem to be relevant any longer, could you please align with the actual content of the proposal?

docs/proposals/secret-manager-store-api.md
moadqassem: @irozzo-1A What about leveraging the user ssh key that we create in the user cluster, by adding a new field which contains the the public key? thus the machine creation wouldn't be dependent anymore on the cloud-init propagation. Once the public key is available in the secret, the agent will inject to the machines afterwards in async fashion.

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: Done

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: This has been done in the implementation

https://github.com/kubermatic/kubermatic/pull/7092

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: Done

docs/proposals/secret-manager-store-api.md
emmanuel-kubermatic: Updated this yesterday

docs/proposals/secret-manager-store-api.md
irozzo-1A: @moadqassem problem is that sshd has to be re-configured and restarted, it's not just a matter of dropping the public key somewhere in the fs. The ssh-agent will be used to rotate the public key, but for the initial setup, the cloud-init seems to be more appropriate.

moadqassem (commented):
/approve
/lgtm
/hold

@emmanuel-kubermatic if the others are ok with it, you can remove the hold

PR #6980, opened 2021-05-04
Proposal for Konnectivity implementation

### LINKED ISSUE #6188: Use Konnectivity as a drop-in replacement for OpenVPN
**User Story**
The goal of this story is to replace our current solution to allow control plane components to reach the cluster network based on OpenVPN with [Konnectivity](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/). The main gain would be simplifying our stack by:

- getting rid of `openvpn-client` and `dnat-controller` sidecar containers in several control-plane components.
- avoid custom "dnsConfig" in control plane components.
- getting rid of `openvpn-server` and `openvpn-client` deployments, in favor of Konnectivity Server and Agent.

I already did some experiments and the migration seems to be quite straightforward, but a few things should be considered:

- Apiserver extensions are supposed to be located in the "cluster" network, today we are deploying `metrics-server` in the "master" network (see https://github.com/kubernetes/kubernetes/pull/93103).
- Some components other than KAS are currently configured to reach cluster network (e.g. scheduler, controller-manager, openstack-cloud-controller-manager). While I'm not sure why this would be useful we still need to double check.
- beta since k8s 1.18

**Acceptance criteria**
Having replaced OpenVPN with Konnectivity :-).

### LINKED ISSUE #6199: Write proposal for OpenVPN replacement with Konnectivity
The goal of this user story is to formalize a proposal about the replacement of our current strategy to enable the communication between control plane components and worker nodes based on OpenVPN with [Konnectivity](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/).

Some aspects that should be taken into account are:

- How to set-up Konnectivity server (side-car container and UDS vs deployment and TCP).
- How to handle migration to this new solution (e.g. feature flag, migration of existing clusters).

youssefazrak:
Proposal to replace OpenVPN with Konnectivity.

Fixes #6199

**Release notes:**
```release-note
NONE
```

### DIFF
--- a/docs/proposals/konnectivity-openvpn.md
+++ b/docs/proposals/konnectivity-openvpn.md
@@ -1,0 +1,120 @@
+# Exposing control-plane services using SNI and HTTP tunneling
+
+Author: Rastislav Szabo (@rastislavs)
+
+Status: Draft proposal.
+
+
+## Goals
+
+The goal of this proposal is to replace KKP’s custom solution for communication between k8s control plane components (master/seed cluster) and worker nodes (user cluster) based on OpenVPN with a more standardized solution based on Konnectivity. The main driver for this change is the simplification of KKP’s stack, which should result in:
+
+* Better stability of the connections between the control plane cluster and the user cluster,
+* Easier debuggability, also for people who are not that familiar with KKP,
+* Lower resource consumption in the control plane cluster.
+
+
+## Non-Goals
+
+This proposal aims to solve only the connectivity initiated from the k8s control plane to the user cluster (e.g. API server to Kubelet communication), not the connectivity initiated in the opposite direction (e.g. from a pod in the user cluster to the API server), for which KKP uses different strategies.
+
+
+## Motivation and Background
+
+The current solution for communication between k8s control plane components and worker nodes is custom to KKP and quite complex. It relies on OpenVPN tunnels between the control plane cluster and user clusters, deployed as:
+
+* OpenVPN server pod in the control plane cluster,
+* OpenVPN client pod in user clusters,
+* Custom iptables configuration deployed as init containers in OpenVPN pods,
+* `openvpn-client` and `dnat-controller` sidecar containers in several control plane components.
+
+This complex solution may be prone to errors especially in large scale deployments. In case of issues, it is not easy to debug them without deep knowledge of this custom setup. With Konnectivity, all these components can be replaced with just Konnectivity Server in the control plane cluster and Konnectivity Agent(s) in the user cluster.
+
+As part of the implementation, we also propose moving the components that act as k8s API extensions (e.g. `metrics-server`) into the user cluster, which would allow us to simplify our stack even more, by getting rid of the custom `dnsConfig` in control plane pods and dns-resolver deployment in the control plane cluster.
+
+Note that Konnectivity is beta since k8s 1.18, so we need to support both OpenVPN and Konnectivity strategy for our users that use older k8s versions, or do not want to move to Konnectivity yet.
+
+
+## Implementation
+
+The implementation proposal follows [this documentation][kubernetes_konnectivity_doc]
+
+
+### Kube API Server Flags
+
+Konnectivity requires [Service Account Token Volume Projection][kubernetes_service_account] to be enabled with `system:konnectivity-server` in `api-audiences`. That means:
+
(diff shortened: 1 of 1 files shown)


docs/proposals/konnectivity-openvpn.md
```diff
@@ -0,0 +1,120 @@
...
+* In case that the service account token volume projection is not yet enabled (which may be already done as part of [this issue][issue_6191]), we need to add some more kube-apiserver flags if Konnectivity is enabled.
+* In case that service account token volume projection is already enabled, we need to add `system:konnectivity-server` into `api-audiences` list.
+
+Apart from that, the API Server needs the `--egress-selector-config-file` flag pointed to the file with `EgressSelectorConfiguration`, which defines the type of connection between the API Server and the Konnectivity Server. 
+
+### Konnectivity Server in Control Plane Cluster
+
+The Konnectivity server runs in the control plane cluster and acts as a proxy between the Kube API server running in the same cluster and Konnectivity Agents that are running in the user cluster and are connecting to the Konnectivity Server’s IP address, which needs to be reachable from the user cluster (for this connection, Kubermatic uses various [Expose Strategies][k8c_expose_strategies]).
+
+Konnectivity server will be deployed as a sidecar container for the Kubernetes API Server. Connection between the two will be done by using UDS.
```
irozzo-1A, line 56: We could mention that we will use an `emptyDir` volume to be able to use UDS between KAS and Konnectivity server.

docs/proposals/konnectivity-openvpn.md
```diff
@@ -0,0 +1,120 @@
...
+
+If there was a need for that going forward, we can potentially add a controller which would dynamically scale the number of replicas based on the number of nodes currently running in the cluster (e.g. if node auto-scaling is enabled).
+
+
+### Impact on cluster network reachability for other control plane components
+
+Currently, multiple control plane components (not only k8s API server, but also e.g. scheduler, controller-manager, openstack-cloud-controller-manager, etc.) are configured to be able to reach the user cluster network. The current solution is based on `openvpn-client` sidecar containers in the control plane pods, and custom `dnsConfig` pointing to a control-plane cluster-local `dns-resolver` deployment. With Konnectivity, this won’t be possible, since Konnectivity provides the cluster network connectivity only for the API server. However, we believe that given the k8s hub-and-spoke design, these components do not actually need the connectivity to anything else other than the k8s API server. If there is a component which currently runs in the control plane cluster and needs the connectivity to the cluster network, we will have to run it in the user cluster.
```
irozzo-1A, line 70: A notable example of components that should be moved on the workers side is metrics-server, because Konnectivity does not support aggregation layers:

https://github.com/kubernetes/kubernetes/issues/92374

This is not the only limitation, metrics-server should also be extended to "pass" from Konnectivity in order to reach kubelets.

docs/proposals/konnectivity-openvpn.md
```diff
@@ -0,0 +1,120 @@
...
+
+After all of the above mentioned changes, we can clean up these unnecessary components from the control plane cluster:
+
+* `openvpn-server`,
+* `openvpn-client` sidecars,
+* `dnat-controller` sidecars in control plane pods (e.g. in the apiserver deployment),
+* custom `dnsConfig` in control plane pods,
+* `dns-resolver` deployment,
+
+As well the `openvpn-client` in the user cluster.
```
irozzo-1A, line 90: Yesss, I like this :-)

docs/proposals/konnectivity-openvpn.md
youssefazrak: So what you are saying right now is that the metrics-server can't be deployed on the usercluster side which leads to deploy it and make it use the Konnectivity right?

irozzo-1A: Nope, I meant that at the moment metrics-server cannot be deployed on `seed` cluster side but only on `user` side, so we have no options but moving it.
Otherwise, we would have to reopen and try to push for kubernetes/kubernetes#92374, and contributing to metrics-server too in order to make it "Konnectivity" ready.

youssefazrak: Gotcha 👍🏻 
Pushed the fix.

docs/proposals/konnectivity-openvpn.md
youssefazrak: Pushed the fix

youssefazrak:
@irozzo-1A ready for another pass.

PR #6990, opened 2021-05-06
[OpenStack] Fix cluster finalizers race condition

### LINKED ISSUE #6759: Missing cluster finilazers due to possible data race
**Steps to reproduce:**
 * I create openstack cluster
 * I see cloud [openstack InitializeCloudProvider function](https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/openstack/provider.go#L158) created resources in openstack
 * When later I check cluster finalizers by running kubectl describe
 * I see [the finalizers](https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/openstack/provider.go#{L45:L58}) are missing
 
I debugged and found that finalizers were added and deleted shortly after. I found that finalizers added by [openstack InitializeCloudProvider function](https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/openstack/provider.go#L158) were deleted by following update here https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/kubernetes/pending.go#L90

I was able to fix the bug by reverting reverting changes introduced #4578

**Kubermatic version:** `v2.16.5`

irozzo-1A:

Fixes #6759

This PR provides the fix for OpenStack provider only. The other providers will be fixed by follow-up PRs.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
+++ b/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
@@ -183,12 +183,18 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 	}
 
 	if _, err := prov.InitializeCloudProvider(cluster, r.updateCluster); err != nil {
+		if kerrors.IsConflict(err) {
+			// In case of conflict we just re-enqueue the item for later
+			// processing without returning an error.
+			r.log.Infow("failed to add finalizer to cluster", "error", err)
+			return &reconcile.Result{Requeue: true}, nil
+		}
 		return nil, fmt.Errorf("failed cloud provider init: %v", err)
 	}
 
 	if _, err := r.updateCluster(cluster.Name, func(c *kubermaticv1.Cluster) {
 		c.Status.ExtendedHealth.CloudProviderInfrastructure = kubermaticv1.HealthStatusUp
-	}); err != nil {
+	}, false); err != nil {
 		return nil, fmt.Errorf("failed to set cluster health: %v", err)
 	}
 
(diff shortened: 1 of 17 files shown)


furkhat:
Do you think we should also change InitializeCloudProvider to check if resource is already created? For example if you look at openstack/provider.go when the network is created (func `createKubermaticNetwork`) the code doesn't check if network already exist, but it does when it creates security group. May be also rename func's to `getOrCreate..`?

irozzo-1A:
> Do you think we should also change InitializeCloudProvider to check if resource is already created? For example if you look at openstack/provider.go when the network is created (func `createKubermaticNetwork`) the code doesn't check if network already exist, but it does when it creates security group. May be also rename func's to `getOrCreate..`?

I agree we should observe the current state, the currently used pattern is to check if resource ID is present in spec and in this case assume the resource is present, otherwise create the resource and update the spec. This is not ideal because if a resource gets deleted by mistake it's not reconciled, and because if the spec update fails resources are leaked. That said I wouldn't go that far here :-), I would create another issue for this.

mfranczy (commented):
Hm, by seeing in how many places we have to add a flag for optimistic lock (especially the `false` value that doesn't change the default behaviour of the `updateCluster`), would it make sense to create another function called `updateClusterWithLock` instead of adding optional `optimisticLock` arg in the current function? 

Because of that you would not have to add `false` everywhere where optimistic lock is not needed. What do you think?

mfranczy:
> would it make sense to create another function called updateClusterWithLock instead of adding optional optimisticLock arg in the current function

I thought here to take common parts to one function and `updateCluster` and `updateClusterWithLock` could wrap it.

irozzo-1A:
> Hm, by seeing in how many places we have to add a flag for optimistic lock (especially the `false` value that doesn't change the default behaviour of the `updateCluster`), would it make sense to create another function called `updateClusterWithLock` instead of adding optional `optimisticLock` arg in the current function?
> 
> Because of that you would not have to add `false` everywhere where optimistic lock is not needed. What do you think?

Makes a lot of sense, don't know why I did not do it like this in the first instance ;-)

mfranczy:
> Makes a lot of sense, don't knkow why I did not do it like this in the first instance ;-)

Maybe because while discussing that we came to the conclusion that additional parameter makes sense, subconscious action :)

mfranczy (approved):
Nice, even better that you transformed update to variadic func

/lgtm

PR #7002, opened 2021-05-10
ensure resources concurrently to speed up reconciliation

furkhat:
Signed-off-by: Furkhat Kasymovgeniiuulu <vailodf@gmail.com>

We experienced that sometimes cluster creation is slow. If I run `watch kubectl get all -n cluster-CLUSTERNAME` with newly created cluster name, then I see resources appearing somewhat like this: one secret - ~5 seconds - then another secret - ~5 seconds - then some configmap - ~5 seconds - another configmap - ~5 seconds - etc. I suspect it is related to the [Wait until the object exists in the cache](https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/resources/reconciling/ensure.go#L87) in `EnsureNamedObject` function.
This changes improves speed roughly by 4 times in our environment.

### DIFF
--- a/pkg/controller/operator/master/reconciler.go
+++ b/pkg/controller/operator/master/reconciler.go
@@ -201,18 +201,25 @@ func (r *Reconciler) reconcileConfigMaps(ctx context.Context, config *operatorv1
 func (r *Reconciler) reconcileSecrets(ctx context.Context, config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
 	logger.Debug("Reconciling Secrets")
 
-	creators := []reconciling.NamedSecretCreatorGetter{
-		common.WebhookServingCASecretCreator(config),
-		common.WebhookServingCertSecretCreator(config, r.Client),
-		common.ExtraFilesSecretCreator(config),
+	creatorsOrdered := [][]reconciling.NamedSecretCreatorGetter{
+		{
+			common.WebhookServingCASecretCreator(config),
+		},
+		{
+			common.WebhookServingCertSecretCreator(config, r.Client),
+			common.ExtraFilesSecretCreator(config),
+		},
 	}
 
 	if config.Spec.ImagePullSecret != "" {
-		creators = append(creators, common.DockercfgSecretCreator(config))
+		creatorsOrdered[1] = append(creatorsOrdered[1], common.DockercfgSecretCreator(config))
 	}
 
-	if err := reconciling.ReconcileSecrets(ctx, creators, config.Namespace, r.Client, common.OwnershipModifierFactory(config, r.scheme)); err != nil {
-		return fmt.Errorf("failed to reconcile Secrets: %v", err)
+	modifiers := common.OwnershipModifierFactory(config, r.scheme)
+	for _, c := range creatorsOrdered {
+		if err := reconciling.ReconcileSecrets(ctx, c, config.Namespace, r.Client, modifiers); err != nil {
+			return fmt.Errorf("failed to reconcile Secrets: %v", err)
+		}
 	}
 
 	return nil
--- a/pkg/resources/reconciling/ensure.go
+++ b/pkg/resources/reconciling/ensure.go
@@ -19,12 +19,14 @@ package reconciling
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	v1 "k8s.io/api/apps/v1"
 	kubeerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilerror "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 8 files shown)


pkg/controller/operator/master/reconciler.go
```diff
@@ -201,18 +201,25 @@ func (r *Reconciler) reconcileConfigMaps(ctx context.Context, config *operatorv1
 func (r *Reconciler) reconcileSecrets(ctx context.Context, config *operatorv1alpha1.KubermaticConfiguration, logger *zap.SugaredLogger) error {
 	logger.Debug("Reconciling Secrets")
 
-	creators := []reconciling.NamedSecretCreatorGetter{
-		common.WebhookServingCASecretCreator(config),
-		common.WebhookServingCertSecretCreator(config, r.Client),
-		common.ExtraFilesSecretCreator(config),
+	creatorsOrdered := [][]reconciling.NamedSecretCreatorGetter{
+		{
+			common.WebhookServingCASecretCreator(config),
```
furkhat, line 206: There is a special case when ensuring objects concurrently. CA's are expected to be created before Certs.

pkg/resources/reconciling/ensure.go
```diff
@@ -60,6 +62,49 @@ func createWithName(rawcreate ObjectCreator, name string) ObjectCreator {
...
+	Name             types.NamespacedName
+	Creator          ObjectCreator
+	EmptyObj         runtime.Object
+	RequiresRecreate bool
+}
+
+// EnsureNamedObjectsConcurrent calls `EnsureNamedObject` for each specifiecation concurrently and returns aggregated errors if any occurred.
+func EnsureNamedObjectsConcurrent(ctx context.Context, client ctrlruntimeclient.Client, objs []EnsureObject) error {
+	var wg sync.WaitGroup
+	errChan := make(chan error)
```
kron4eg:
```suggestion
	errChan := make(chan error, len(objs))
```

To avoid blocking `errChan <- fmt.Errorf("ensure %+v: %v", obj.Name, err)`

pkg/resources/reconciling/ensure.go
```diff
@@ -60,6 +62,49 @@ func createWithName(rawcreate ObjectCreator, name string) ObjectCreator {
...
+	for _, obj := range objs {
+		wg.Add(1)
+		go func(obj EnsureObject) {
+			defer wg.Done()
+			err := EnsureNamedObject(ctx, obj.Name, obj.Creator, client, obj.EmptyObj, obj.RequiresRecreate)
+			if err != nil {
+				errChan <- fmt.Errorf("ensure %+v: %v", obj.Name, err)
```
kron4eg, line 83:
```suggestion
				errChan <- fmt.Errorf("ensure %+v: %w", obj.Name, err)
```

wrap the error with "new" (as of go1.13) error handling

pkg/resources/reconciling/ensure.go
```diff
@@ -60,6 +62,49 @@ func createWithName(rawcreate ObjectCreator, name string) ObjectCreator {
...
+			}
+		}(obj)
+	}
+
+	ret := make(chan error)
+	defer close(ret)
```
kron4eg: if we move `wg.Wait()` here, we woudn't need to launch a goroutine to gather errors.

kron4eg (commented):
Left few comments about implementation, but those are not really important. I'm more curious what we will achieve? Or at what scale (number of Cluster objects) we would feel the difference? What are numbers that you are trying to improve with this.

As old saying goes: without benchmarks you can't improve performance.

furkhat:
@kron4eg I don't have automated benchmark but for our installation we improved cluster creation speed from 9+ minutes to ~2 minutes.

pkg/resources/reconciling/ensure.go
furkhat: AFAIK %w is useful when client wants to [unwrap the error](https://golangbyexample.com/wrapping-and-unwrapping-error-golang/) and check the type/kind of the error for some special case. Because `EnsureNamedObject` uses `fmt.Errorf` I would change there to %w too. But that means the client would need to unwrap at least twice to get to some error he can check against. 
Sometimes to ensure clients do not use unwrap %v is used intentionally. Do you think it could be what we want?

kron4eg:
In the first place I suggest to find the real source of performance degradation. I never observed such a bad resources creation timing. In our dev cluster those resources are created almost instantly. Of course we don't have thousands of clusters in dev, but still, the process of reconciliation for a single cluster doesn't involve other clusters so it shouldn't be a problem.

I can only speculate that it's kube-apiserver is slow in your setup for whatever reason (slow etcd? too much load? not really sure).

furkhat:
@kron4eg as I have mentioned in PR description already, the "source of performance degragation" is in [wait until the object exists in the cache](https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/resources/reconciling/ensure.go#L87)

kron4eg:
I can't reproduce this issue in our cluster.

furkhat:
> he process of reconciliation for a single cluster doesn't involve other clusters so it shouldn't be a problem.

Yes resources are reconciled separately, but aren't they reconciled by the same controllers? For example in [kubernetes controller](https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/controller/seed-controller-manager/kubernetes/cluster_controller.go) if all workers are busy (because of slow cache sync I mentioned above) then other clusters will wait.

furkhat:
> I can't reproduce this issue in our cluster.

@kron4eg What would you suggest us to do? How do you proceed with bugs that you can't reproduce in your cluster?

kron4eg:
> > I can't reproduce this issue in our cluster.
> 
> @kron4eg What would you suggest us to do? How do you proceed with bugs that you can't reproduce in your cluster?

I'd suggest to find out why "the cache is slow" in your case. Do you have many thousands of clusters maybe? Or maybe it's kube-apiserver who fail to respond in timely manner.

furkhat:
If we find and improve cache sync speed that would be great.
Would you consider merging this anyway? Or what would you suggest?

kron4eg:
> Would you consider merging this anyway? Or what would you suggest?

From my PoV, it complicates the codebase for no good reason. I'd suggest to find the real source of slowness. Having no other evidence that this is useful I'd restrain from merging it.

furkhat:
This changes are useful for better design. Why resources should go one by one and block each other because of the cache?

Issue #6584, opened 2021-02-22
Use images with multi-arch support for KKP Components


rastislavs:
**User Story**
As a user, I would like to run KKP master/seed clusters and user clusters on the arm64 platform.

When deployed a KKP User Cluster on an arm64 machine, some KKP components would not start due to using images for an incorrect architecture:

```
NAMESPACE              NAME                                        READY   STATUS             RESTARTS   AGE
kube-system            canal-gr5z5                                 1/2     CrashLoopBackOff   57         4h9m
kube-system            coredns-856777c9f5-msrjk                    1/1     Running            0          24m
kube-system            coredns-856777c9f5-spfsc                    1/1     Running            0          21m
kube-system            kube-proxy-26bsj                            1/1     Running            0          4h30m
kube-system            logrotate-czh9x                             1/1     CrashLoopBackOff   32         4h23m
kube-system            node-local-dns-6zckq                        1/1     Running            0          4h23m
kube-system            openvpn-client-6c9dd998bc-qzc2m             1/2     CrashLoopBackOff   59         5h51m
kube-system            user-ssh-keys-agent-8rqfs                   0/1     CrashLoopBackOff   31         4h30m
kubernetes-dashboard   dashboard-metrics-scraper-975c84c89-fg79d   1/1     Running            1          5h51m
kubernetes-dashboard   dashboard-metrics-scraper-975c84c89-jrb5w   1/1     Running            0          5h51m
```

(Log message `standard_init_linux.go:211: exec user process caused "exec format error"`)

Ideally, all KKP components should use repositories with multi-arch manifests, so that the same repository and tag works for any architecture.

**Acceptance criteria**
- User Clusters nodes with arm64 architecture can successfully run all KKP components.
- All Master / Seed clusters components can run on arm64 clusters as well.

rastislavs:
Upgrade to canal with flannel v0.13.0 should resolve the issue for the canal pod (https://github.com/kubermatic/kubermatic/issues/6016)

mlavacca:
All the user-cluster components now use multi-arch images
/close

Issue #6016, opened 2020-10-12
Upgrade Canal as a CNI for user clusters.


youssefazrak:
**User Story**
We are currently using Canal as a CNI for user clusters. Till we implement #5589 and for clusters running older versions of Kubermatic, we should upgrade to the latest version: 3.16.3 from 3.8.0.

**Acceptance criteria**
- Canal in VXLAN mode is installed in user clusters.
- Having upgrade documentation for existing user clusters.

youssefazrak:
After some thinking, going with Cilium (eBPF) might be a good idea. This wouldn't change much on the difficulty compared to upgrading Calico to the latest version for existing installations.

HaveFun83:
any news here?

PR #7011, opened 2021-05-11
Flannel version update

mlavacca:
Flannel version has been updated to 0.13.0. The update of Flannel is required to support ARM architecture as seen in #6584.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -526,7 +526,7 @@ spec:
         # This container runs flannel using the kube-subnet-mgr backend
         # for allocating subnets.
         - name: kube-flannel
-          image: quay.io/coreos/flannel:v0.11.0
+          image: quay.io/coreos/flannel:v0.13.0
           command: [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr" ]
           securityContext:
             privileged: true


irozzo-1A:
Hey @mlavacca, this is in scope of #6016. Did you check if the upgrade between 0.11 and 0.13 is smooth?

mlavacca:
> Hey @mlavacca, this is in scope of #6016. Did you check if the upgrade between 0.11 and 0.13 is smooth?

I checked and didn't find any issue in using canal with calico 3.8 and flannel 0.13. Unfortunately, I have neither the certainty of the opposite nor found any compatibility matrix or similar. @kron4eg do you have something more regarding this?

kron4eg:
We've use such non-standard setup in KubeOne for a while (before upgrading calico). Since flannel is only responsible for VXLAN and calico is only responsible for network-policies it should be good. 

However in some foreseeable future we need to upgrade calico too as its said in #6016.

irozzo-1A:
> I checked and didn't find any issue in using canal with calico 3.8 and flannel 0.13. Unfortunately, I have neither the certainty of the opposite nor found any compatibility matrix or similar. @kron4eg do you have something more regarding this?

My point was more about the upgrade of flannel from 0.11 to 0.13 on live clusters, as it is more sensitive than fresh clusters creation.
BTW I will start working at #6016 today or the beginning of the next week, if there is no reason for integrating this right now I would tend to wait.

kron4eg:
@irozzo-1A I guess in that case flannel 0.13 will not interfere with your #6016.

irozzo-1A:
> @irozzo-1A I guess in that case flannel 0.13 will not interfere with your #6016.

I agree it won't interfere. I just wanted to know if the upgrade from flannel 0.11 to 0.13 was tested. This is something that has to be done in the scope of #6016. If it wasn't, I don't think integrating this makes real sense.

mlavacca:
@irozzo-1A @kron4eg I did some tests with [goldpinger](https://github.com/bloomberg/goldpinger) deployed when performing a live update of Flannel from 0.11 to 0.13. Actually, there are strange fluctuations in the node healthy after switching the version, as it is possible to see from the dashboard snapshot [here](https://snapshot.raintank.io/dashboard/snapshot/tb2UC66KlI5jIlNP0XcRrigZ1RBiWv3i). 
Since this work is in the scope of #6016 and updating only the flannel version could bring in some instability, I'll close the PR.

PR #7017, opened 2021-05-12
Use grafana orgid header

### LINKED ISSUE #6981: Fix race condition by handling Grafana datasources
Grafana API requires a user context switch to a particular Organization before calling an API call to add/modify a datasource in it.

As the Grafana datasource controller runs with multiple workers, and user context switches are global, this leads to race conditions in Grafana datasource controller, which may cause Grafana data sources to appear in a wrong Organization / KKP Project.

Alternatively to user context switches, this can be indicated in `X-Grafana-Org-Id` HTTP header when performing the API call.
More details: https://github.com/grafana/grafana/issues/5991

aborilov:

Fixes #6981 

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/datasource_grafana_controller.go
```diff
@@ -144,10 +148,22 @@ func (r *datasourceGrafanaReconciler) reconcile(ctx context.Context, cluster *ku
 	if cluster.Spec.MLA == nil {
 		cluster.Spec.MLA = &kubermaticv1.MLASettings{}
 	}
+	grafanaClient := grafanasdk.NewClient(r.grafanaURL, r.grafanaAuth, r.httpClient)
```
jiachengxu, line 154: So this means in every reconcilation loop, you will create a new grafana client.
Is this what we really want to do?
I am thinking that we could still pass the grafanaclient to this reconciler, and only set header in the reconciliation loop.
WDYT?

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -106,7 +107,7 @@ func Add(
 	}
 	// FIXME: Grafana API uses a global user context switches to manage datasources,
```
rastislavs: you should be also able to remove the FIXME comment and use numWorkers instead of 1

pkg/controller/seed-controller-manager/mla/datasource_grafana_controller.go
aborilov: yeah, that was my first idea, but in this way, we will have the same race condition, cause we are using one client for all workers.
creating a client does not do much, it's just creating structure, so it doesn't cost much for us.

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: yes, of course. I still plan to add some changes

PR #7036, opened 2021-05-13
Allow to override Machine Controller version

mfranczy:
Allow to override Machine Controller version for dev, tests and PoC purposes.
This PR doesn't change the way how MC is being shipped with KKP,  the introduced option is not supported for production setups.

Fixes #

After diving into the code I've decided to not change the way how we ship MC, that introduced option is only for PoC/tests/dev purposes. I've created another issue to create a proposal how to do versioning in KKP, this must be discussed.

**Documentation**:

```release-note
NONE
```


kron4eg:
@mfranczy please rebase this PR against the master, there is fix for that failing job.

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -103,6 +105,13 @@ type KubermaticAPIConfiguration struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 }
 
+// MachineControllerConfiguration configures Machine Controller
+type MachineControllerConfiguration struct {
```
xmudrii: Since other structs are prefixed with `Kubermatic`, it might make sense to prefix this one as well (for consistency).

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -49,6 +49,8 @@ type KubermaticConfigurationSpec struct {
 	Auth KubermaticAuthConfiguration `json:"auth"`
 	// FeatureGates are used to optionally enable certain features.
 	FeatureGates sets.String `json:"featureGates,omitempty"`
+	// MachineController configures the Machine Controller
+	MachineController MachineControllerConfiguration `json:"machineController,omitempty"`
```
xmudrii: I'm not sure if this is the right place for this struct. Since the machine-controller is a user cluster component, it might make sense to put this field in the `KubermaticUserClusterConfiguration` struct.

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -103,6 +105,13 @@ type KubermaticAPIConfiguration struct {
 	Replicas *int32 `json:"replicas,omitempty"`
 }
 
+// MachineControllerConfiguration configures Machine Controller
+type MachineControllerConfiguration struct {
+	// Image is used to override the Machine Controller image.
+	// It is only for development, tests and PoC purposes. This field must not be set in production environments.
+	Image string `json:"image,omitempty"`
```
xmudrii: I don't think that we have any API guidelines, but in other structs, we have used a combination of `DockerRegistry` and `DockerTag` fields. For example: 

https://github.com/kubermatic/kubermatic/blob/6785a3ab208c6eb410b53e74b843488e82d3fa85/pkg/crd/operator/v1alpha1/configuration.go#L117-L124

Maybe we should do the same here to be consistent.

pkg/crd/operator/v1alpha1/configuration.go
mfranczy: I would prefer to keep it simple, i doubt there is a benefit of adding `Kubermatic` prefix, Machine Controller term is unambiguous.

pkg/crd/operator/v1alpha1/configuration.go
mfranczy: I thought about that to move it to seed configuration (since seed manager deploys MC), let me take a look again, maybe your suggestion is better.

pkg/crd/operator/v1alpha1/configuration.go
mfranczy: Ok I can change, but it should be `ImageRepository` and `ImageTag` instead of `Docker*`, the name indicates that only docker is supported which is not true.

mfranczy:
@xmudrii i've addressed your comments, please take a look again.

xmudrii (approved):
/lgtm
/approve

kron4eg (commented):
/lgtm
/approve
/retest

PR #7037, opened 2021-05-13
[WIP] [OpenStack] Install the controller from CSI cinder addon to seed cluster

### LINKED ISSUE #6406: Cluster stuck in deletion Openstack when there are PersistentVolumes and no worker nodes
Steps to reproduce:
  1. Create Openstack cluster with worker nodes, wait for cluster to get healthy
  2. Then deploy some applications which would claim persistent storage, wait for them to be ready
  3. Then delete all worker nodes by deleting all node deployments, wait for it to finish
  4. Then delete the cluster.
  5. The cluster should be stuck in deletion.

Kubermatic deploys [csi addon](https://github.com/kubermatic/kubermatic/blob/master/addons/csi/controllerplugin.yaml#L44) on worker nodes, `csi-cinder-controllerplugin` is responsible for adding and deleting `external-attacher/cinder-csi-openstack-org` finalizer from PersistentVolumes and VolumeAttachments. When we delete all nodedeployments `csi-cinder-controllerplugin` will be down and won't be able to schedule because it needs worker nodes to run on.
When cluster is being deleted and cleaning up procedure issues deletion of in-cluster PersistentVolumes thay won't be deleted because they will have `external-attacher/cinder-csi-openstack-org` on them.

wozniakjan:
Cleanup logic for OpenStack clusters can leave an undesired residue when the cinder CSI plugin is deleted before it gets a chance to deprovision all PVs.

Fixes https://github.com/kubermatic/kubermatic/issues/6406

This PR proposes a change in addons handling. Previously addons got templated and deployed entirely to the user cluster, now it would be possible to define parts of the addon to get deployed to the seed cluster. And this is directly demonstrated by the cinder CSI plugin.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/addons/csi/openstack/controllerplugin-rbac.yaml
+++ b/addons/csi/openstack/controllerplugin-rbac.yaml
@@ -14,11 +14,13 @@
 
 {{ if .Cluster.Features.Has "externalCloudProvider" }}
 {{ if eq .Cluster.CloudProviderName "openstack" }}
+{{ if eq .AddonCluster "seed" }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: csi-cinder-controller-sa
-  namespace: kube-system
+  namespace: {{ .Cluster.Namespace }}
+{{ end }}
 
 ---
 # external attacher
--- a/addons/csi/openstack/controllerplugin.yaml
+++ b/addons/csi/openstack/controllerplugin.yaml
@@ -15,6 +15,7 @@
 {{ if .Cluster.Features.Has "externalCloudProvider" }}
 
 {{ if eq .Cluster.CloudProviderName "openstack" }}
+{{ if eq .AddonCluster "seed" }}
 {{ $version := "UNSUPPORTED" }}
 {{ if eq .Cluster.MajorMinorVersion "1.17" }}
 {{ $version = "v1.17.1 "}}
@@ -37,7 +38,7 @@ kind: Service
 apiVersion: v1
 metadata:
   name: csi-cinder-controller-service
-  namespace: kube-system
+  namespace: {{ .Cluster.Namespace }}
   labels:
     app: csi-cinder-controllerplugin
 spec:
@@ -52,7 +53,7 @@ kind: StatefulSet
 apiVersion: apps/v1
 metadata:
   name: csi-cinder-controllerplugin
-  namespace: kube-system
+  namespace: {{ .Cluster.Namespace }}
 spec:
   serviceName: "csi-cinder-controller-service"
   replicas: 1
@@ -71,6 +72,7 @@ spec:
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
+            - "--kubeconfig=/etc/kube/kubeconfig"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
@@ -78,12 +80,15 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+            - name: kubeconfig
+              mountPath: /etc/kube
         - name: csi-provisioner
           image: k8s.gcr.io/sig-storage/csi-provisioner:v2.1.0
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
             - "--default-fstype=ext4"
+            - "--kubeconfig=/etc/kube/kubeconfig"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
@@ -91,11 +96,14 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+            - name: kubeconfig
+              mountPath: /etc/kube
         - name: csi-snapshotter
           image: k8s.gcr.io/sig-storage/csi-snapshotter:v2.1.3
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
+            - "--kubeconfig=/etc/kube/kubeconfig"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
@@ -103,12 +111,15 @@ spec:
           volumeMounts:
             - mountPath: /var/lib/csi/sockets/pluginproxy/
               name: socket-dir
+            - name: kubeconfig
+              mountPath: /etc/kube
         - name: csi-resizer
           image: k8s.gcr.io/sig-storage/csi-resizer:v1.1.0
           args:
             - "--csi-address=$(ADDRESS)"
             - "--timeout=3m"
             - "--handle-volume-inuse-error=false"
+            - "--kubeconfig=/etc/kube/kubeconfig"
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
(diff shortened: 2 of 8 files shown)


addons/csi/openstack/controllerplugin-rbac.yaml
```diff
@@ -14,11 +14,13 @@
 
 {{ if .Cluster.Features.Has "externalCloudProvider" }}
 {{ if eq .Cluster.CloudProviderName "openstack" }}
+{{ if eq .AddonCluster "seed" }}
```
wozniakjan, line 17: For the purposes of splitting the installation of an addon between user and seed cluster, I would like to introduce `AddonCluster` field. The logic I was trying to implement is following:
* the sections that are contained within `if eq .AddonCluster "seed"` branch go only to seed cluster and not to user cluster
* the sections of the manifest not containing this branching are installed to the user cluster like they always had been

addons/csi/openstack/controllerplugin.yaml
```diff
@@ -169,7 +182,11 @@ spec:
         - name: socket-dir
           emptyDir:
         - name: cloud-config
+          configMap:
+            name: cloud-config
+        - name: kubeconfig
           secret:
-            secretName: cloud-config
+            secretName: internal-admin-kubeconfig
```
wozniakjan, line 189: the CSI sidecar containers are capable of being pointed to a different cluster. Should I create designated kubeconfig just for this CSI plugin or is it ok for it to re-use the `internal-admin-kubeconfig'?

pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -435,49 +452,95 @@ func (r *Reconciler) writeAdminKubeconfig(ctx context.Context, log *zap.SugaredL
...
 	done := func() {
 		kubeconfigDone()
 		manifestDone()
+		seedManifestDone()
+	}
+	return manifestInteraction{
+		kubeconfigFilename:   kubeconfigFilename,
+		userManifestFilename: userManifestFilename,
+		seedManifestFilename: seedManifestFilename,
+		done:                 done,
```
wozniakjan, line 524: the count of return parameters was getting a little too large I think so I decided to encapsulate that within a struct for easier consumption

pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -486,19 +549,36 @@ func (r *Reconciler) ensureIsInstalled(ctx context.Context, log *zap.SugaredLogg
...
+// ensureIsInstalled ensures proper installation of the addon to appropriate clusters - an addon can have part installed in seed and part in user cluster
+// NOTE: because previously addons were all installed only to user cluster, following code also ensures removal of the seed cluster addon part from the user cluster
+func (r *Reconciler) ensureIsInstalled(ctx context.Context, log *zap.SugaredLogger, addon *kubermaticv1.Addon, cluster *kubermaticv1.Cluster) error {
+	mi := r.setupManifestInteraction(ctx, log, addon, cluster)
+	if mi.err != nil {
+		return mi.err
+	}
+	defer mi.done()
+	var errs []error
+	if err := r.applyManifest(ctx, log, addon, cluster.Name, mi.seedManifestFilename, ""); err != nil {
```
wozniakjan, line 575: for the seed cluster, we don't need to pass a kubeconfig and use in cluster authentication instead

pkg/controller/seed-controller-manager/addon/addon_controller_test.go
```diff
@@ -418,7 +446,57 @@ func TestHugeManifest(t *testing.T) {
...
+	}
+	if svc.Namespace != cluster.Status.NamespaceName || svc.Namespace != clusterNS {
+		t.Fatalf("svc.Namespace[%v] != cluster.Status.NamespaceName[%v] != clusterNS[%v]", svc.Namespace, cluster.Status.NamespaceName, clusterNS)
+	}
+
+	manifests, err = controller.getAddonManifests(ctx, log, addon, cluster, templateAddonForUserCluster)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(manifests) != 0 {
```
wozniakjan, line 499: because the entire manifest is encapsulated within `{{ if eq .AddonCluster "seed" }}`, templating of the addon for user cluster should return no raw resources.

pkg/controller/seed-controller-manager/addon/addon_controller_test.go
```diff
@@ -418,7 +446,57 @@ func TestHugeManifest(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if len(manifests) != 1 {
+		t.Fatalf("invalid number of manifests returned. Expected 1, Got %d", len(manifests))
+	}
+	svc := v1.Service{}
+	if err := json.Unmarshal(manifests[0].Raw, &svc); err != nil {
+		t.Fatalf("failed to unmarshal raw service: %v", err)
+	}
+	if svc.Namespace != cluster.Status.NamespaceName || svc.Namespace != clusterNS {
```
wozniakjan, line 491: but for the seed cluster, it should template correctly including setting the namespace

wozniakjan (commented):
providing some explanation and rationale behind the code.

pkg/controller/seed-controller-manager/addon/addon_controller.go
```diff
@@ -260,12 +276,12 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, addo
 	}
 
 	// Reconciling
-	if err := r.ensureIsInstalled(ctx, log, addon, cluster); err != nil {
-		return nil, fmt.Errorf("failed to deploy the addon manifests into the cluster: %v", err)
-	}
 	if err := r.ensureFinalizerIsSet(ctx, addon); err != nil {
 		return nil, fmt.Errorf("failed to ensure that the cleanup finalizer exists on the addon: %v", err)
 	}
+	if err := r.ensureIsInstalled(ctx, log, addon, cluster); err != nil {
```
wozniakjan, line 282: I think ensuring finalizers should happen before the installation. It potentially could have been a culprit for leaving some residue behind if someone deleted an addon in the middle of an installation before the finalizers had a chance to be added.

wozniakjan:
I have re-added the WIP label, failing to reliably reproduce this fixes the issue after all today. It seemed to have worked ok earlier today but now I have very mixed feelings. Not sure if I'm testing this right
1) I patched my cluster with label: `worker-name: jwozniak`
2) running the controller locally with `KUBERMATIC_WORKERNAME=jwozniak ./hack/run-seed-controller-manager.sh`

it previously did install the addon in split fashion but now it seems to have a hard time deploying anything at all

wozniakjan:
TODO: it's now back to correctly splitting the installation but constantly toggles `Cluster` status condition `AddonControllerReconciledSuccessfully` between `true` and `false` and reports error during reconciliation
```
error retrieving RESTMappings to prune: invalid resource networking.k8s.io/v1, Kind=Ingress, Namespaced=true: no matches for kind \"Ingress\" in version \"networking.k8s.io/v1\"
```
I suspect this may be due to fact that the user cluster is running 1.19, kubectl baked in the controller probably running a similar version, but seed cluster is 1.17. Despite the fact this addon does not install ingress, `--prune` option probably doing resource discovery and seed has only older ingress API versions, missing the `networking.k8s.io/v1`
```
$ k api-resources | grep '\<ingresses\>'
ingresses                          ing             extensions/v1beta1                     true         Ingress
ingresses                          ing             networking.k8s.io/v1beta1              true         Ingress
```

wozniakjan:
it has been decided that this is not the right approach

wozniakjan:
https://github.com/kubermatic/kubermatic/pull/7413 has a better approach to solving this

wozniakjan:
Please don't reopen this, it brings more trouble than what it solves. Read https://github.com/kubermatic/kubermatic/pull/7037#issuecomment-886454800,

PR #7052, opened 2021-05-18
cleanup MLA resources when userClusterMLA disabled

### LINKED ISSUE #7019: Cleanup of MLA resources when MLA is disabled
Make sure that MLA-related resources are cleaned up when MLA is disabled on the Seed level, including:
 - MLA components in the user cluster
 - MLA components in the user cluster namespaces in Seed (e.g. MLA Gateway)
 - Grafana Org. ID annotations on Project resources
 - MLA finalizers

This story not cover the data cleanup - that is handled in a separate issue: https://github.com/kubermatic/kubermatic/issues/7021

aborilov:
**What this PR does / why we need it**: In this PR we changed the way MLA controller running, now they are running even when `userClusterMLA` disable, in that way they can clean up MLA resource when `userClusterMLA` switched from enabled to disabled.

Fixes #7019 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/seed-controller-manager/controllers.go
+++ b/cmd/seed-controller-manager/controllers.go
@@ -405,7 +405,7 @@ func createInitialMachineDeploymentController(ctrlCtx *controllerContext) error 
 }
 
 func createMLAController(ctrlCtx *controllerContext) error {
-	if !userClusterMLAEnabled(ctrlCtx) {
+	if !ctrlCtx.runOptions.featureGates.Enabled(features.UserClusterMLA) {
 		return nil
 	}
 	return mla.Add(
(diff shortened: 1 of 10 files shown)


jiachengxu (commented):
As mentioned in the issue #7019:
This clean up should also clean up:
- MLA components in the user cluster
- MLA components in the user cluster namespaces in Seed (e.g. MLA Gateway)

How does this PR handle those two points?
Also, I think this clean up also needs to clean up alertmanager related objects and configurations

aborilov:
> As mentioned in the issue #7019:
> This clean up should also clean up:
> 
> * MLA components in the user cluster
> * MLA components in the user cluster namespaces in Seed (e.g. MLA Gateway)
> 
> How does this PR handle those two points?
> Also, I think this clean up also needs to clean up alertmanager related objects and configurations

cause MLA controllers remain running even when MLA disable, those resources will be cleaned up. Only problem, that for datasource and alertmanager controller, that only checking cluster level MLA options we need to check seed level MLA flag.
as a workaround, I've added a check for seed level MLA enable flag, but we should add propagation controller, which will propagate seed level MLA options to the cluster.

PR #7054, opened 2021-05-18
project creation supports multiple owners as owner references 

### LINKED ISSUE #7053: Update sync project master controller
Update sync project controller to support multiple project owners during project creation. 
The endpoint to create a project by SA can pass a list of the human user owners:
```
POST http://localhost:8080/api/v1/projects


{"name":"test","users":["user1@kubermatic.com", "user2@kubermatic.com", "user3@kubermatic.com"]}
```
The sync project controller expects only one owner. Make all necessary changes to support multiple project owners.

zreigz:
**What this PR does / why we need it**: Update sync project controller to support multiple project owners during project creation. It's needed to create a project by service account. During this process, the endpoint expects a list of human owners. 

Fixes #7053

```release-note
NONE
```

### DIFF
--- a/pkg/controller/master-controller-manager/rbac/sync_project.go
+++ b/pkg/controller/master-controller-manager/rbac/sync_project.go
@@ -108,7 +108,7 @@ func (c *projectController) ensureProjectIsInActivePhase(ctx context.Context, pr
 
 // ensureProjectOwner makes sure that the owner of the project is assign to "owners" group
 func (c *projectController) ensureProjectOwner(ctx context.Context, project *kubermaticv1.Project) error {
-	var sharedOwnerPtr *kubermaticv1.User
+	var sharedOwnerPtrList []*kubermaticv1.User
 	for _, ref := range project.OwnerReferences {
 		if ref.Kind == kubermaticv1.UserKindName {
 			var sharedOwner kubermaticv1.User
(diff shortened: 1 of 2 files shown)


pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -108,34 +108,53 @@ func (c *projectController) ensureProjectIsInActivePhase(ctx context.Context, pr
...
+	var sharedOwnerPtrList []*kubermaticv1.User
 	for _, ref := range project.OwnerReferences {
 		if ref.Kind == kubermaticv1.UserKindName {
 			var sharedOwner kubermaticv1.User
 			key := types.NamespacedName{Name: ref.Name}
 			if err := c.client.Get(ctx, key, &sharedOwner); err != nil {
 				return err
 			}
-			sharedOwnerPtr = &sharedOwner
+			sharedOwnerPtrList = append(sharedOwnerPtrList, sharedOwner.DeepCopy())
```
moadqassem, line 122: So if I get this you have a list now, the resource not found error would actually stop the loop at some point. I would say filter by the error(line:116) and if the error is not found then continue.

pkg/controller/master-controller-manager/rbac/sync_project.go
```diff
@@ -108,34 +108,53 @@ func (c *projectController) ensureProjectIsInActivePhase(ctx context.Context, pr
...
 	var bindings kubermaticv1.UserProjectBindingList
 	if err := c.client.List(ctx, &bindings); err != nil {
 		return err
 	}
 
-	for _, binding := range bindings.Items {
-		if binding.Spec.ProjectID == project.Name && strings.EqualFold(binding.Spec.UserEmail, owner.Spec.Email) &&
-			binding.Spec.Group == GenerateActualGroupNameFor(project.Name, OwnerGroupNamePrefix) {
-			return nil
+	projectOwnerMap := map[string]bool{}
```
moadqassem: Nit: since you are only looking for keys here you use a struct pointer instead 
```suggestion
	projectOwnerMap := map[string]*struct
```

pkg/controller/master-controller-manager/rbac/sync_project.go
zreigz: good point, I've also added unit test for this

pkg/controller/master-controller-manager/rbac/sync_project.go
zreigz: changed to hash set

zreigz:
@moadqassem PTAL

moadqassem (commented):
/approve
/lgtm

PR #7060, opened 2021-05-19
User-ssh-keys-agent multiarch image with Buildah

### LINKED ISSUE #7059: Building user-ssh-keys-agent multi-arch image
**User Story**
As part of #6584, We would like to run KKP user clusters on the arm64 platform. This is part of that epic issue to have multi-arch images for the `user-ssh-keys-agent` component of KKP.

**Acceptance criteria**
The `user-ssh-keys-agent` component runs on arm64 machines as well as amd64 machines with the same image.

mlavacca:
Once Buildah will be included in the e2e docker image [infra-1367](https://github.com/kubermatic/infra/pull/1367), it will be possible to build multi-arch images. The image this PR is referred to is the `user-ssh-keys-agent`.

Fixes #7059 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/user-ssh-keys-agent/Dockerfile
+++ b/cmd/user-ssh-keys-agent/Dockerfile
@@ -12,9 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.13
+FROM docker.io/golang:1.16.4 as builder
+
+WORKDIR /go/src/github.com/kubermatic/kubermatic
+COPY . .
+ENV CGO_ENABLED=0
+RUN go build ./cmd/user-ssh-keys-agent
+
+FROM docker.io/alpine:3.13
 LABEL maintainer="support@kubermatic.com"
 
-COPY ./_build/user-ssh-keys-agent /usr/local/bin/user-ssh-keys-agent
-
+COPY --from=builder /go/src/github.com/kubermatic/kubermatic/user-ssh-keys-agent /usr/local/bin/user-ssh-keys-agent
 ENTRYPOINT ["/usr/local/bin/user-ssh-keys-agent"]
--- a/cmd/user-ssh-keys-agent/Makefile
+++ b/cmd/user-ssh-keys-agent/Makefile
@@ -20,5 +20,5 @@ build:
 	GOOS=$(GOOS) CGO_ENABLED=0 go build -o ./_build/user-ssh-keys-agent
 
 .PHONY: docker
-docker: build
-	docker build -t $(DOCKER_REPO)/user-ssh-keys-agent:$(TAG) .
+docker:
+	cd ../.. && docker build -t $(DOCKER_REPO)/user-ssh-keys-agent:$(TAG) -f cmd/user-ssh-keys-agent/Dockerfile .
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1323,7 +1323,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-      - image: quay.io/kubermatic/build:go-1.16-node-14-0
+      - image: quay.io/kubermatic/build:go-1.16-node-14-kind-0.8-2
         command:
         - "./hack/ci/run-offline-test.sh"
         # docker-in-docker needs privileged mode
--- a/hack/ci/push-images.sh
+++ b/hack/ci/push-images.sh
@@ -38,6 +38,7 @@ apt install time -y
 echodate "Logging into Quay"
 docker ps > /dev/null 2>&1 || start-docker.sh
 retry 5 docker login -u "$QUAY_IO_USERNAME" -p "$QUAY_IO_PASSWORD" quay.io
+retry 5 buildah login -u "$QUAY_IO_USERNAME" -p "$QUAY_IO_PASSWORD" quay.io
 echodate "Successfully logged into Quay"
 
 # prepare special variables that will be injected into the Kubermatic Operator;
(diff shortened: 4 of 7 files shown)


cmd/user-ssh-keys-agent/Dockerfile
```diff
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.13
-LABEL maintainer="support@kubermatic.com"
+FROM docker.io/golang:1.16.1 as builder
+
+WORKDIR /go/src/github.com/kubermatic/kubermatic
+COPY . .
+RUN CGO_ENABLED=0 GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) go build ./cmd/user-ssh-keys-agent
```
kron4eg: There is no point in `GOOS=$(go env GOOS) GOARCH=$(go env GOARCH)`.
```suggestion
ENV CGO_ENABLED=0
RUN go build ./cmd/user-ssh-keys-agent
```

cmd/user-ssh-keys-agent/Dockerfile
```diff
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.13
-LABEL maintainer="support@kubermatic.com"
+FROM docker.io/golang:1.16.1 as builder
```
kron4eg:
```suggestion
FROM docker.io/golang:1.16.4 as builder
```

cmd/user-ssh-keys-agent/Makefile
```diff
@@ -1,3 +1,4 @@
+
```
kron4eg:
```suggestion
```

cmd/user-ssh-keys-agent/Makefile
```diff
@@ -13,12 +14,46 @@
...
+.PHONY: push
+push:
+	MANIFEST=${DOCKER_REPO}/${IMAGE}:${TAG} ; \
+	buildah manifest push --all "$$MANIFEST" "docker://$$MANIFEST"
+
+.PHONY: tag
+tag:
+	OLD_MANIFEST=${DOCKER_REPO}/${IMAGE}:${PRIMARY_TAG} ; \
+	NEW_MANIFEST=${DOCKER_REPO}/${IMAGE}:${NEW_TAG} ; \
+	buildah tag $$OLD_MANIFEST $$NEW_MANIFEST
```
kron4eg: I think this needs no double $ (escape of $), isn't it?
```suggestion
	buildah tag $OLD_MANIFEST $NEW_MANIFEST 
```

cmd/user-ssh-keys-agent/Makefile
```diff
@@ -13,12 +14,46 @@
...
+docker:
+	cd ../.. ; \
+	docker build -t $(DOCKER_REPO)/user-ssh-keys-agent:$(TAG) -f cmd/user-ssh-keys-agent/Dockerfile .
+
+.PHONY: build
+build: build_multiarch
+	MANIFEST=${DOCKER_REPO}/${IMAGE}:${TAG} ; \
+	buildah manifest create "$$MANIFEST" ; \
+	for ARCH in ${ARCHITECTURES} ; do \
+		buildah manifest add "$$MANIFEST" "${DOCKER_REPO}/${IMAGE}-$$ARCH:${TAG}" ; \
```
kron4eg:
```suggestion
		buildah manifest add "$MANIFEST" "${DOCKER_REPO}/${IMAGE}-$ARCH:${TAG}" ; \
```

cmd/user-ssh-keys-agent/Makefile
```diff
@@ -13,12 +14,46 @@
...
+	done
+
+.PHONY: push
+push:
+	MANIFEST=${DOCKER_REPO}/${IMAGE}:${TAG} ; \
+	buildah manifest push --all "$$MANIFEST" "docker://$$MANIFEST"
```
kron4eg:
```suggestion
	buildah manifest push --all "$MANIFEST" "docker://$MANIFEST"
```

kron4eg (commented):
OK, so I've made a bunch of suggestions but scratch that. Since we don't exploit any of make features here (graph of dependencies, build automation of them), let's move logic from here to simple bash scripts ./hack directory.

cmd/user-ssh-keys-agent/Makefile
mlavacca: Yeah, they are needed. Make requires double $$ to access rule-scoped variables

mlavacca:
> OK, so I've made a bunch of suggestions but scratch that. Since we don't exploit any of make features here (graph of dependencies, build automation of them), let's move logic from here to simple bash scripts ./hack directory.

I used the makefile to keep using the [release-docker-images.sh](https://github.com/kubermatic/kubermatic/blob/master/hack/release-docker-images.sh) as we were doing before this change without introducing very verbose operations that would reduce its readability. with this approach, we can keep the verbosity and complexity of the operation outside that script that builds all the KKP images. If in the future we would like to build also other multiarch images with Buildah, centralizing those operations in a single script can result in a mess. Just an opinion though, if you think moving this to that script is a better solution I'll do it :)

kron4eg:
Bash scripts are subject to shfmt and shell-check, there is no such thing for makefile. No actual features of make are being used (DAG of build dependencies), make here is used as a shell, so let's use the original better shell that we can lint and format in automated way.

kron4eg (approved):
/approve
/lgtm

PR #7063, opened 2021-05-19
Implement User synchronizer controller

### LINKED ISSUE #7055: Sync KKP users from Master to Seed to let KKP admins access Grafana/Alertmanager of user MLA stack in Seed
For the Grafana and Alertmanager UI of the new user MLA stack seed, KKP admins should be able to access all Grafana organizations, datasourses, and alertmanager UI of any clusters.

However, it is impossible for now because `User` objects are only in Master cluster, but not in Seed, which means admin info is not available at the seed level.

For fixing this, a controller of master-controller-manager needs to be implemented to propagate `User` objects from Master to Seed.

Note: service account user (with prefix `serviceaccount-`) should be excluded.

jiachengxu:
This PR implements a user synchronizer controller to propagate KKP User objects from Master to all Seeds.
This is needed for:
- KKP admins to access the MLA stack at the Seed level. 
- Manage Users in Grafana based on KKP Users.

Fixes #7055 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/master-controller-manager/user-synchronizer/controller.go
+++ b/pkg/controller/master-controller-manager/user-synchronizer/controller.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package usersynchronizer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"go.uber.org/zap"
+
+	kubermaticapiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	kuberneteshelper "k8c.io/kubermatic/v2/pkg/kubernetes"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName = "user-synchronizer"
+	saPrefix       = "serviceaccount-"
+)
+
+type reconciler struct {
+	log          *zap.SugaredLogger
+	recorder     record.EventRecorder
+	masterClient ctrlruntimeclient.Client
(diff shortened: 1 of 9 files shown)


pkg/controller/master-controller-manager/user-synchronizer/controller.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	c, err := controller.New(ControllerName, masterManager, controller.Options{Reconciler: r, MaxConcurrentReconciles: numWorkers})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	serviceAccountPredicate := predicate.NewPredicateFuncs(func(object ctrlruntimeclient.Object) bool {
+		// We don't trigger reconciliation for service account.
+		return !strings.HasPrefix(object.GetName(), saPrefix)
```
zreigz: you can use existing public function for this: https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/kubernetes/sa.go#L364

jiachengxu: Thanks @zreigz, fixed

zreigz (approved):

pkg/controller/master-controller-manager/user-synchronizer/controller.go
```diff
@@ -0,0 +1,164 @@
...
+	if kuberneteshelper.HasFinalizer(user, kubermaticapiv1.SeedUserCleanupFinalizer) {
+		kuberneteshelper.RemoveFinalizer(user, kubermaticapiv1.SeedUserCleanupFinalizer)
+		if err := r.masterClient.Update(ctx, user); err != nil {
+			return fmt.Errorf("failed to remove user finalizer %s: %w", user.Name, err)
+		}
+	}
+	return nil
+}
+
+func (r *reconciler) syncAllSeeds(
```
irozzo-1A, line 145: Not sure this function brings a lot of benefits, with respect to looping directly over the sed clients, but not a blocker ;-)

irozzo-1A:
@jiachengxu LGTM from a code perspective.
I was thinking that as we are not triggering the reconcile loop when `user` resources on the seed side change, we could think to have a relatively short re-sync period to avoid drifts (not possible with controller-runtime thought if I recall correctly). WDYT?

pkg/controller/master-controller-manager/user-synchronizer/controller.go
jiachengxu: Thanks @irozzo-1A, this function is used in two different places:
one for creating User in all seeds: https://github.com/kubermatic/kubermatic/blob/69de3cb6213845a31cad054801c4cf2c097f0de1/pkg/controller/master-controller-manager/user-synchronizer/controller.go#L116-L118
another for deleting them when doing GC:
https://github.com/kubermatic/kubermatic/blob/69de3cb6213845a31cad054801c4cf2c097f0de1/pkg/controller/master-controller-manager/user-synchronizer/controller.go#L127-L132
I think it is convenient to have this and use it and you can focus on the `action` function :)

irozzo-1A: Yep, saw it. What I was pointing out is just that there is no much logic on it, but again this is largely a matter of taste and it's absolutely not a blocker.

jiachengxu:
Just to summarize discussion with @irozzo-1A here:
Previously, this controller watched `User` objects also on the seed to make sure the reconciliation loop will be triggered when seed side change happens. Now we decided to rely on the cache re-sync (every 2 seconds: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go#L100), this may give us some temporary drifts, but that is acceptable.

irozzo-1A:
> Just to summarize discussion with @irozzo-1A here:
> Previously, this controller watched `User` objects also on the seed to make sure the reconciliation loop will be triggered when seed side change happens. Now we decided to rely on the cache re-sync (every 2 seconds: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go#L100), this may give us some temporary drifts, but that is acceptable.

Thx for reporting here the discussion, we also said that this 2 seconds resync period should be reviewed because it's probably too short.

PR #7066, opened 2021-05-19
Support regular parameters for Kubermatic constraints Constraint

### LINKED ISSUE #6770: Allow for regular Constraint parameters instead of rawJSON
Currently the Kubermatic Constraints `parameters` field which should mirror the Gatkeeper Constraint parameters field is just a raw JSON representation of it:

```
spec:
  constraintType: K8sRequiredLabels
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Namespace"]
  parameters:
    rawJSON: '{"labels":["gatekeeper"]}'
```

```
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Namespace"]
  parameters:
    labels: ["gatkeeper"]
```

This was because there was some issue (i dont remeber correctly) with json/encoding and that because for the Gatkeeper Constraint we need to use Unstructured. But I think its time to take another look at this so it will be easier for users to create Kubermatic Constraints.

lsviben:
Adds support for regular Kubermatic OPA Constraint `parameters`. The old way with `rawJSON` holding the parameters in a JSON string is still supported and is used if present. 

This means that users can now more easily manage OPA Constraints in KKP as the Kubermatic Constraint spec will look the same as the Gatekeeper Constraints one.

Previously
```
kind: Constraint
metadata:
  name: ns-must-have-gk
  namespace: cluster-795pq9n2tc
spec:
  constraintType: K8sRequiredLabels
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Namespace"]
  parameters:
    rawJSON: '{"labels":["gatekeeper"]}'

```

With regular `parameters` support:
```
kind: Constraint
metadata:
  name: ns-must-have-gk
  namespace: cluster-795pq9n2tc
spec:
  constraintType: K8sRequiredLabels
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Namespace"]
  parameters:
    labels: ["gatkeeper"]

```

Fixes #6770 

```release-note
Kubermatic OPA Constraints now additionaly support using regular yaml `parameters` instead of `rawJSON`. `rawJSON` is still supported so no migration needed.   
```


pkg/util/deepcopy/deepcopy_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+			if diff := deep.Equal(tc.copyFrom, tc.expectedResult); diff != nil {
+				t.Errorf("Got unexpected difference in maps. Diff to expected: %v", diff)
+			}
+
+		})
+	}
+}
+
+type bob struct {
```
moadqassem: Nit: this can be inlined, as you are assigning to an interface and you are not doing any type assertion

pkg/util/deepcopy/deepcopy_test.go
```diff
@@ -0,0 +1,77 @@
...
+		copyFrom       map[string]interface{}
+		copyTo         map[string]interface{}
+		expectedResult map[string]interface{}
+		expectedErr    error
+	}{
+		{
+			name: "scenario 1: deep copy one map to another",
+			copyFrom: map[string]interface{}{
+				"key":      []string{"values", "values"},
+				"otherKey": bob{name: "Bob"},
```
moadqassem: Nit: check out the comment below:
```suggestion
				"otherKey": struct { name string }{ name: "bob" },
```

pkg/util/deepcopy/deepcopy_test.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			copyTo: map[string]interface{}{},
+			expectedResult: map[string]interface{}{
+				"key":      []string{"values", "values"},
+				"otherKey": bob{name: "Bob"},
```
moadqassem: Nit: check out the comment below:
```suggestion
				"otherKey": struct { name string }{ name: "bob" },
```

pkg/util/deepcopy/deepcopy.go
```diff
@@ -0,0 +1,32 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package deepcopy
+
+import "encoding/json"
+
+// StringInterfaceMapCopy copies one map[string]interface{} to another by using json to marshall/unmarshall it
+func StringInterfaceMapCopy(from map[string]interface{}, to map[string]interface{}) error {
```
moadqassem, line 22: Is this util function will be used across packages?

pkg/crd/kubermatic/v1/constraint.go
```diff
@@ -44,10 +48,44 @@ type ConstraintSpec struct {
...
+	//
+	// And with regular parameters:
+	//
+	// parameters:
+	//   labels: ["gatekeeper"]
+	//
 	Parameters Parameters `json:"parameters,omitempty"`
 }
 
+type Parameters map[string]interface{}
```
moadqassem, line 68: Why not having a concrete type here?

pkg/util/deepcopy/deepcopy.go
lsviben: I put it here just in case somebody needs a similar functionality. But I can just put it into pkg/crd/kubermatic/v1/constraint.go if needed

pkg/crd/kubermatic/v1/constraint.go
lsviben: Because the Gatekeeper Constraints CRDs are created dynamically, and its Parameters can hold anything. 

https://open-policy-agent.github.io/gatekeeper/website/docs/howto/#constraints

So it can be something easy like in the example above or it can whole objects. So we cant make a concrete type

moadqassem (commented):
/approve
/lgtm

PR #7081, opened 2021-05-20
Open NodePort range in openstack

### LINKED ISSUE #7080: Open NodePorts managed openstack security group

kron4eg:
Open NodePort range in openstack

Fixes #7080

**Documentation**:

```release-note
Open NodePort range in openstack
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
+++ b/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
@@ -69,12 +69,13 @@ var _ reconcile.Reconciler = &Reconciler{}
 type Reconciler struct {
 	ctrlruntimeclient.Client
 
-	log        *zap.SugaredLogger
-	recorder   record.EventRecorder
-	seedGetter provider.SeedGetter
-	workerName string
-	versions   kubermatic.Versions
-	caBundle   *x509.CertPool
+	log                *zap.SugaredLogger
+	recorder           record.EventRecorder
+	seedGetter         provider.SeedGetter
+	workerName         string
+	versions           kubermatic.Versions
+	caBundle           *x509.CertPool
+	seedNodePortsRange string
 }
 
 func Add(
@@ -85,15 +86,17 @@ func Add(
 	workerName string,
 	versions kubermatic.Versions,
 	caBundle *x509.CertPool,
+	nodePortsRange string,
 ) error {
 	reconciler := &Reconciler{
-		Client:     mgr.GetClient(),
-		log:        log.Named(ControllerName),
-		recorder:   mgr.GetEventRecorderFor(ControllerName),
-		seedGetter: seedGetter,
-		workerName: workerName,
-		versions:   versions,
-		caBundle:   caBundle,
+		Client:             mgr.GetClient(),
+		log:                log.Named(ControllerName),
+		recorder:           mgr.GetEventRecorderFor(ControllerName),
+		seedGetter:         seedGetter,
+		workerName:         workerName,
+		versions:           versions,
+		caBundle:           caBundle,
+		seedNodePortsRange: nodePortsRange,
 	}
 
 	c, err := controller.New(ControllerName, mgr, controller.Options{Reconciler: reconciler, MaxConcurrentReconciles: numWorkers})
@@ -147,7 +150,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 	if !found {
 		return nil, fmt.Errorf("couldn't find datacenter %q for cluster %q", cluster.Spec.Cloud.DatacenterName, cluster.Name)
 	}
-	prov, err := cloud.Provider(datacenter.DeepCopy(), r.getGlobalSecretKeySelectorValue, r.caBundle)
+	prov, err := cloud.Provider(datacenter.DeepCopy(), r.getGlobalSecretKeySelectorValue, r.caBundle, r.seedNodePortsRange)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create cloud provider: %v", err)
 	}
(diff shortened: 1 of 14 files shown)


pkg/controller/seed-controller-manager/cloud/cloud_controller.go
```diff
@@ -182,6 +185,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 		}
 	}
 
+	cluster.Spec.ComponentsOverride.Apiserver.NodePortRange = r.nodePortRange
```
irozzo-1A: I'm not very fond of this. We should rely on the value we get from cluster spec and not override the value here. The best strategy to me would be to move the defaulting of `ComponentsOverride` to the mutating webhook.

kron4eg: I've updated the PR to avoid making changes in cluster object

pkg/provider/cloud/openstack/provider.go
```diff
@@ -223,10 +236,23 @@ func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
 	}
 
 	if cluster.Spec.Cloud.Openstack.SecurityGroups == "" {
-		secGroupName, err := createKubermaticSecurityGroup(netClient, cluster.Name)
+		lowPort, highPort := resources.NewTemplateDataBuilder().
+			WithNodePortRange(osp.nodePortsRange).
+			WithCluster(cluster).
+			Build().
+			NodePorts()
```
xmudrii, line 243: Why we don't use the `ComputedNodePortRange` function here? My understanding is that the `NodePorts` function doesn't take component overrides into account, or if the provided range is an empty string.

pkg/validation/cluster.go
```diff
@@ -254,7 +275,7 @@ func ValidateUpdateCluster(ctx context.Context, newCluster, oldCluster *kubermat
 	}
 
 	secretKeySelectorFunc := provider.SecretKeySelectorValueFuncFactory(ctx, clusterProvider.GetSeedClusterAdminRuntimeClient())
-	cloudProvider, err := cloud.Provider(dc, secretKeySelectorFunc, caBundle)
+	cloudProvider, err := cloud.Provider(dc, secretKeySelectorFunc, caBundle, resources.DefaultNodePortRange)
```
xmudrii, line 278: Shouldn't we pass the "real" range here, i.e. the range provided to the seed-controller-manager? Passing the default range seems a little bit strange to me. If we would ever use that value in the context of validation, we would always validate the default value, which might not be used at all.

On that note, are we validating that value anywhere?

pkg/provider/cloud/openstack/provider.go
kron4eg: NodePorts() uses ComputedNodePortRange() to get string to parse (and return ints for low and high ports), so it does take override logic into account.

pkg/validation/cluster.go
kron4eg: No, it's OK here. It doesn't affect anything actually and an empty string could be passed as well.

NodePortRange parameter (just like caBundle) is a leaking API here, that indicates the need to refactor the cloud provider interface, but it's out of scope of this PR. And I've experienced that need first-hand.

The seed-manager flag `-node-port-range` is inaccessible in this context, as this function is part of web API, without deep refactoring it does not seem possible.

irozzo-1A:
@kron4eg I have two general comments.

First, I think that https://github.com/kubermatic/kubermatic/pull/7082 will simplify the picture. The problem here is that the default value is not explicitly set (and it is not immutable). This forces us to pass around the value that we use as a default.
Once the aforementioned PR is finished and merged the most of this won't be necessary any longer and we will be able to just rely on the value in the spec. The only reason I see why we should integrate this is maybe that it's easier to backport. WDYT?

Second, we are introducing a change in behaviour, I would recommend having a feature gate to switch it on.

xmudrii:
I'm generally okay with going forward with this PR.

@irozzo-1A Going forward with #7082 sounds okay too, but note that cherry-picking it will be much more complicated. That's because the webhook logic and structure got changed (quite) a bit between 2.16 and 2.17. The way we pass default values around in this PR is maybe not that elegant (see the comment I left above and @kron4eg's response), but it'll solve the problem more quickly (this PR is basically ready for merging).

> Second, we are introducing a change in behaviour, I would recommend having a feature gate to switch it on.

Can you please elaborate on this a little bit? The behavior seems pretty much the same to me, but maybe I missed something.

irozzo-1A:
> @irozzo-1A Going forward with #7082 sounds okay too, but note that cherry-picking it will be much more complicated. That's because the webhook logic and structure got changed (quite) a bit between 2.16 and 2.17. The way we pass default values around in this PR is maybe not that elegant (see the comment I left above and @kron4eg's response), but it'll solve the problem more quickly (this PR is basically ready for merging).

The PR I'm working on won't be backported to 2.16, if this is the target I agree we should integrate this first.

> > Second, we are introducing a change in behaviour, I would recommend having a feature gate to switch it on.
> 
> Can you please elaborate on this a little bit? The behavior seems pretty much the same to me, but maybe I missed something.

I was referring to opening the node port range to every source. This is probably ok for most of the users, but still could be interesting to have the choice to opt-in or opt-out. I would prefer the first option to keep the same behavior as before.

kron4eg:
OK then, since we clarified everything, let's LGTM it.

/cherrypick release/v2.17
/cherrypick release/v2.16

irozzo-1A:
Ok, as we discussed offline we can consider adding a switch later in case of need.

/lgtm
/approve

xmudrii (approved):
/lgtm
/approve

PR #7094, opened 2021-05-23
Adds seedconstraintsynchronizer controller 

### LINKED ISSUE #6804: Default Constraints - seed controller
Implement a controller which syncs Default constraints in the Seed cluster from the `kubermatic` namespace to the user cluster namespaces of Clusters which have OPA-integration enabled

imharshita:

Adds a seed controller for default constraints which syncs from to the seed kubermatic ns to user cluster ns.

Fixes #6804

**Documentation**: [Default Constraints Proposal](https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/default-constraints.md)

```release-note
none
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -2166,6 +2166,8 @@ const (
 	GatekeeperSeedConstraintCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-seed-constraint"
 	// GatekeeperConstraintCleanupFinalizer indicates that gatkeeper constraints on the user cluster need cleanup
 	GatekeeperConstraintCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-constraints"
+	// KubermaticUserClusterNsDefaultConstraintCleanupFinalizer indicates that kubermatic constraints on the user cluster namespace need cleanup
+	KubermaticUserClusterNsDefaultConstraintCleanupFinalizer = "kubermatic.io/cleanup-kubermatic-usercluster-ns-default-constraints"
 	// KubermaticConstraintCleanupFinalizer indicates that Kubermatic constraints for the cluster need cleanup
 	KubermaticConstraintCleanupFinalizer = "kubermatic.io/cleanup-kubermatic-constraints"
 	// SeedProjectCleanupFinalizer indicates that Kubermatic Projects on the seed clusters need cleanup
--- a/pkg/controller/seed-controller-manager/constraint-controller/controller.go
+++ b/pkg/controller/seed-controller-manager/constraint-controller/controller.go
@@ -1,0 +1,330 @@
+/*
+Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package seedconstraintsynchronizer
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	"go.uber.org/zap"
+
+	kubermaticapiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
+	controllerutil "k8c.io/kubermatic/v2/pkg/controller/util"
+	kubermaticpred "k8c.io/kubermatic/v2/pkg/controller/util/predicate"
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	kuberneteshelper "k8c.io/kubermatic/v2/pkg/kubernetes"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	"k8c.io/kubermatic/v2/pkg/util/workerlabel"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// This controller syncs the kubermatic constraints to constraint on the user cluster.
+	ControllerName = "constraint_syncing_controller"
+	finalizer      = kubermaticapiv1.KubermaticUserClusterNsDefaultConstraintCleanupFinalizer
+	Key            = "default"
+	Value          = "true"
+	AddAction      = "add"
+	RemoveAction   = "remove"
+)
+
+type reconciler struct {
+	log                     *zap.SugaredLogger
+	workerNameLabelSelector labels.Selector
+	recorder                record.EventRecorder
+	namespace               string
+	seedClient              ctrlruntimeclient.Client
+}
+
+func opaPredicate() predicate.Funcs {
+	return kubermaticpred.Factory(func(o ctrlruntimeclient.Object) bool {
+		cluster, ok := o.(*kubermaticv1.Cluster)
+		if !ok {
+			return false
+		}
+		return cluster.Spec.OPAIntegration != nil && cluster.Spec.OPAIntegration.Enabled
+	})
+}
+
+func withEventFilter() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldConstraint, ok := e.ObjectOld.(*kubermaticv1.Constraint)
+			if !ok {
+				return false
+			}
+			newConstraint, ok := e.ObjectNew.(*kubermaticv1.Constraint)
+			if !ok {
+				return false
+			}
+			return !reflect.DeepEqual(oldConstraint.Spec, newConstraint.Spec)
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
+func Add(ctx context.Context,
+	mgr manager.Manager,
+	log *zap.SugaredLogger,
+	workerName string,
+	namespace string,
+	numWorkers int) error {
+
+	workerSelector, err := workerlabel.LabelSelector(workerName)
+	if err != nil {
+		return fmt.Errorf("failed to build worker-name selector: %v", err)
+	}
+
+	reconciler := &reconciler{
+		log:                     log.Named(ControllerName),
+		workerNameLabelSelector: workerSelector,
+		recorder:                mgr.GetEventRecorderFor(ControllerName),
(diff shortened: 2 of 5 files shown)


pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+
+	return nil
+}
+
+func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, request reconcile.Request) error {
+
+	constraint := &kubermaticv1.Constraint{}
+
+	if err := r.seedClient.Get(ctx, request.NamespacedName, constraint); err != nil {
+		if controllerutil.IsCacheNotStarted(err) {
```
irozzo-1A: I don't see why we should check for this particular kind of error (I don't even think it should occur), anyway we are not doing anything different from returning it as we would do on line 157. Please remove this.

imharshita: done

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// Reconcile reconciles the kubermatic constraints in the seed cluster and syncs them to all user clusters namespace
+// which have opa integration enabled
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.With("request", request)
+	log.Debug("Reconciling")
+
+	err := r.reconcile(ctx, log, request)
+	if controllerutil.IsCacheNotStarted(err) {
```
irozzo-1A: I'm not sure this check is useful in any way, controller-runtime starts the informers and waits for the cache syncs. Is there any rationale behind it?

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return nil
+}
+
+func isOPAEnabled(userCluster *kubermaticv1.Cluster) bool {
+	if userCluster.Spec.OPAIntegration != nil && userCluster.Spec.OPAIntegration.Enabled {
+		return true
+	}
+	return false
```
lsviben: You can simplify this by changing to :

```suggestion
	return userCluster.Spec.OPAIntegration != nil && userCluster.Spec.OPAIntegration.Enabled 
```

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil
+		}
+
+		if err := r.cleanupConstraint(ctx, log, constraint); err != nil {
+			return err
+		}
+	}
+
+	if !kuberneteshelper.HasFinalizer(constraint, finalizer) {
+		if err := r.patchFinalizer(ctx, constraint, "add"); err != nil {
```
lsviben: NIT: you can put the "add" and "remove" actions as constants

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,284 @@
...
+
+	workerSelector, err := workerlabel.LabelSelector(workerName)
+	if err != nil {
+		return fmt.Errorf("failed to build worker-name selector: %v", err)
+	}
+
+	reconciler := &reconciler{
+		log:                     log.Named(ControllerName),
+		workerNameLabelSelector: workerSelector,
+		recorder:                mgr.GetEventRecorderFor(ControllerName),
```
irozzo-1A, line 116: I don't think recorder is used here

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,280 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// This controller syncs the kubermatic constraints to constraint on the user cluster.
+	ControllerName = "constraint_syncing_controller"
+	finalizer      = kubermaticapiv1.GatekeeperConstraintCleanupFinalizer
```
lsviben: this finalizer is used to mark the Kubermatic Constraints in the user cluster namespace on the seed cluster, that they need to cleanup the Gatekeeper Constraints on the user clusters before deletion. 

So I would suggest using a new finalizer not to mix that with this case, some DefaultConstraintCleanupFinalzier or something like that

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,280 @@
...
+		enqueueConstraints(reconciler.seedClient, reconciler.log, namespace),
+		workerlabel.Predicates(workerName),
+	); err != nil {
+		return fmt.Errorf("failed to create watch for clusters: %w", err)
+	}
+
+	if err := c.Watch(
+		&source.Kind{Type: &kubermaticv1.Constraint{}},
+		&handler.EnqueueRequestForObject{},
+		predicate.ByNamespace(namespace),
```
irozzo-1A: We should also reconcile when the `Constraints` change on the cluster namespaces, isn't it?

pkg/controller/seed-controller-manager/constraint-controller/controller_test.go
```diff
@@ -0,0 +1,149 @@
...
+	testCases := []struct {
+		name                 string
+		namespacedName       types.NamespacedName
+		requestName          string
+		expectedConstraint   *kubermaticv1.Constraint
+		expectedGetErrStatus metav1.StatusReason
+		seedClient           ctrlruntimeclient.Client
+	}{
+		{
+			name: "scenario 1: sync constraint to user cluster ns",
```
lsviben, line 64: not sure if this tests works correctly, in this first case you are getting and comparing to the same constraint which is generated in the `seedClient` as you are doing the GET below with the NamespacedName from it, you are not checking if the namespace is the proper one.

Similarly for tests 2 and 3 the reconcile doesnt even happen properly as you are giving it an empty NamespaceName.

You need to generate a cluster as well, a namespace for it and check if the sync happens properly

imharshita: done

pkg/controller/seed-controller-manager/constraint-controller/controller.go
lsviben: We discussed this with @imharshita , and for this case it seems to us that the default 10h resync will be ok. 

The reason is that users anyway wont be able to edit/delete these "default/global" constraints (at least from the API side, if they have kubeconfig access then its a different story), so there is no need to add active protection from that. Let us know if you see something we are missing

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	if !constraint.DeletionTimestamp.IsZero() {
+		if !kuberneteshelper.HasFinalizer(constraint, finalizer) {
+			return nil
+		}
+
+		if err := r.cleanupConstraint(ctx, log, constraint); err != nil {
+			return err
+		}
```
lsviben, line 227: you need to add `return nil` here so the reconcile finishes here if its being deleted

pkg/controller/seed-controller-manager/constraint-controller/controller.go
irozzo-1A: Well, of course in the happy path everything works fine :-). This would allow to automatically recover when `Constraints` get accidentally edited or removed. I don't think should be too complicated to put in place. Isn't it?

lsviben: no, acutally we could add a separate watch for Constraints on the user-cluster namespaces which have some default label, so we dont watch the "user" constraints changes. I just dont know anymore when to rely on the resync and when to watch actively in KKP :D. 

BTW @imharshita , this reminded me that it would be good to add the labeling of the "default" constraints in this PR, so we can distinguish them later in the API.

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,330 @@
...
+	return kubermaticpred.Factory(func(o ctrlruntimeclient.Object) bool {
+		cluster, ok := o.(*kubermaticv1.Cluster)
+		if !ok {
+			return false
+		}
+		return cluster.Spec.OPAIntegration.Enabled
+	})
+}
+
+func WithEventFilter() predicate.Predicate {
```
irozzo-1A: nit: you could use an unexported function (i.e. `withEventFilter`). It's preferable to be as restrictive as possible

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func enqueueConstraints(client ctrlruntimeclient.Client, log *zap.SugaredLogger, namespace string) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(a ctrlruntimeclient.Object) []reconcile.Request {
+		var requests []reconcile.Request
+
+		constraintList := &kubermaticv1.ConstraintList{}
+		err := client.List(context.Background(), constraintList, &ctrlruntimeclient.ListOptions{Namespace: namespace})
+
+		if err != nil {
```
irozzo-1A:
```suggestion
		if err := client.List(context.Background(), constraintList, &ctrlruntimeclient.ListOptions{Namespace: namespace}); err != nil {
```

pkg/controller/seed-controller-manager/constraint-controller/controller.go
irozzo-1A: @lsviben  You mean something like that?
```
		if err := r.cleanupConstraint(ctx, log, constraint); err != nil {
                          if kerrors.IsNotFound(err) {
                            return nil
                          } 
			return err
		}
```
In such case, I do agree.

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -0,0 +1,328 @@
...
+		&source.Kind{Type: &kubermaticv1.Constraint{}},
+		&handler.EnqueueRequestForObject{},
+		kubermaticpred.ByNamespace(namespace),
+	); err != nil {
+		return fmt.Errorf("failed to create watch for seed constraints: %v", err)
+	}
+
+	if err := c.Watch(
+		&source.Kind{Type: &kubermaticv1.Constraint{}},
+		&handler.EnqueueRequestForObject{},
```
irozzo-1A: Here we should not enqueue the object itself, but the owner, isn't it?

pkg/controller/seed-controller-manager/constraint-controller/controller.go
lsviben: I meant the `return nil` below which is already added. 

But your suggestion is also good. Although I think NotFound is already handled in cleanupConstraints func

pkg/controller/seed-controller-manager/constraint-controller/controller.go
imharshita: makes sense, on it!

pkg/controller/seed-controller-manager/constraint-controller/controller.go
irozzo-1A: ok, thx for the clarification

pkg/api/v1/types.go
```diff
@@ -2166,6 +2166,8 @@ const (
 	GatekeeperSeedConstraintCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-seed-constraint"
 	// GatekeeperConstraintCleanupFinalizer indicates that gatkeeper constraints on the user cluster need cleanup
 	GatekeeperConstraintCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-constraints"
+	// KubermaticUserClusterNsDefaultConstraintCleanupFinalizer indicates that kubermatic constraints on the user cluster namespace need cleanup
+	KubermaticUserClusterNsDefaultConstraintCleanupFinalizer = "kubermatic.io/cleanup-kubermatic-userclusterNS-default-constraints"
```
irozzo-1A: nit: I would not include `kubermatic` as it is redundant, and I would not use uppercase for consistency, what about:
```suggestion
	KubermaticUserClusterNsDefaultConstraintCleanupFinalizer = "kubermatic.io/cleanup-gatekeeper-default-constraints"
```

irozzo-1A:
LGTM Just a small nit and it's ready to go ;-)

PR #7100, opened 2021-05-25
add mla cleanup controller

### LINKED ISSUE #7064: Add MLA clean up controller
**User Story**

To make sure that all resources cleaned up when MLA disable and to not have to clean up on each reconciliation loop we need to add a separate controller that will only do clean up on startup of the controller if MLA is disabled.
We can easily add periodical clean-up later.
The main idea to enqueue a request at the startup of the controller and then properly reconcile it. We want to do a clean-up in the reconcile loop in case of some error, so we are able to retry properly.

**Acceptance criteria**

All resource created by MLA controllers should be removed when Seed level MLA settings disabled

aborilov:
refactored MLA controller, so they can be used by cleanup controller.

Fixes #7064 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/mla/alertmanager_controller.go
+++ b/pkg/controller/seed-controller-manager/mla/alertmanager_controller.go
@@ -55,18 +55,38 @@ const (
 	alertmanagerTenantHeaderName = "X-Scope-OrgID"
 )
 
-type alertmanagerReconciler struct {
+type alertmanagerController struct {
 	ctrlruntimeclient.Client
 	httpClient *http.Client
+
+	log                   *zap.SugaredLogger
+	cortexAlertmanagerURL string
+}
+
+func newAlertmanagerController(
+	client ctrlruntimeclient.Client,
+	log *zap.SugaredLogger,
+	httpClient *http.Client,
+	cortexAlertmanagerURL string,
+) *alertmanagerController {
+	return &alertmanagerController{
+		Client:     client,
+		httpClient: httpClient,
+
+		log:                   log,
+		cortexAlertmanagerURL: cortexAlertmanagerURL,
+	}
+}
+
+type alertmanagerReconciler struct {
+	ctrlruntimeclient.Client
 
 	log        *zap.SugaredLogger
 	workerName string
 	recorder   record.EventRecorder
 	versions   kubermatic.Versions
 
-	cortexAlertmanagerURL string
-
-	mlaEnabled bool
+	alertmanagerController *alertmanagerController
 }
 
 func newAlertmanagerReconciler(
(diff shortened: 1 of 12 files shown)


pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -102,20 +107,40 @@ func Add(
...
+		if err := newAlertmanagerReconciler(mgr, log, numWorkers, workerName, versions, alertmanagerController); err != nil {
+			return fmt.Errorf("failed to create mla alertmanager configuration controller: %w", err)
+		}
+		if err := newUserGrafanaReconciler(mgr, log, numWorkers, workerName, versions, userGrafanaController); err != nil {
+			return fmt.Errorf("failed to create mla user controller: %w", err)
+		}
 	}
-	if err := newUserGrafanaReconciler(mgr, log, numWorkers, workerName, versions, grafanaClient, httpClient, grafanaURL, grafanaHeader, mlaEnabled); err != nil {
-		return fmt.Errorf("failed to create mla user controller: %w", err)
+	cleanupController := newCleanupController(
```
jiachengxu: This means if `mlaEnabled`, we will still run `cleanupController`, what is the reason for that? Will it be flighting with other mla controllers?

pkg/controller/seed-controller-manager/mla/cleanup_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	mapFn := handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		return []reconcile.Request{
+			{NamespacedName: types.NamespacedName{
+				Name:      "identifier",
+				Namespace: "",
+			}}}
+	})
+
+	if err := c.Watch(&source.Kind{Type: &kubermaticv1.Cluster{}}, mapFn); err != nil {
```
jiachengxu: I guess you were mentioning that cleanup controller is watching for `Project`?

jiachengxu: and I guess this doesn't mean that cleanup will only trigger once? It means everything `Cluster` changes it will enqueue the `indentifier`?

pkg/controller/seed-controller-manager/mla/mla.go
aborilov: it won't be fighting cause there is a check for MLA flag in `cleanupController`, but I guess you are right, we can just not run this controller when MLA enabled

aborilov: done

pkg/controller/seed-controller-manager/mla/cleanup_controller.go
aborilov: yep, forget to push that

jiachengxu: but again, this doesn't mean that cleanup will only trigger once? It means every time the Project changes, it will enqueue the identifier?

aborilov: yeah, it will be triggered on each project change. will try to use the channel one more time

aborilov: done

pkg/controller/seed-controller-manager/mla/alertmanager_controller.go
```diff
@@ -55,18 +55,38 @@ const (
...
+	cortexAlertmanagerURL string,
+) *alertmanagerController {
+	return &alertmanagerController{
+		Client:     client,
+		httpClient: httpClient,
+
+		log:                   log,
+		cortexAlertmanagerURL: cortexAlertmanagerURL,
+	}
+}
```
jiachengxu: Can you please move line 58-79 to just above line 211? so we move the definition of `alertmangerController` near the member functions

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -102,20 +107,40 @@ func Add(
...
+		cleanupController := newCleanupController(
+			mgr.GetClient(),
+			log,
+			datasourceGrafanaController,
+			alertmanagerController,
+			orgUserGrafanaController,
+			orgGrafanaController,
+			userGrafanaController,
+		)
+		if err := newCleanupReconciler(mgr, log, numWorkers, workerName, versions, cleanupController); err != nil {
```
jiachengxu, line 142: Can you please add a comment about this cleanup reconciler to the `Add` function?

aborilov: sure.
done

pkg/controller/seed-controller-manager/mla/alertmanager_controller.go
aborilov: thx, done

PR #7102, opened 2021-05-25
Add proposal for user cluster templates

### LINKED ISSUE #7101: Cluster template proposal

zreigz:
**What this PR does / why we need it**: Add proposal for user cluster templates

Fixes #7101

```release-note
NONE
```

### DIFF
--- a/docs/proposals/cluster-templates.md
+++ b/docs/proposals/cluster-templates.md
@@ -1,0 +1,47 @@
+## Introduction
+
+A cluster template is a reusable cluster template object. The Kubernetes CRD will be used for this purpose.
+A cluster template allows you to specify a provider, node layout, and configurations to materialize a cluster instance
+via Kubermatic API or UI.
+
+During cluster creation or after successful creation the user can store cluster definition as a template.
+
+### Scope
+
+The cluster templates should be accessible from different levels. Should have global (admin), project or user scope.
+Template management will be available from project level.
+We also need a central template management place for admins.
+
+### Credentials
+
+For security reasons, the cluster template should contain a preset name to retrieve credentials. The user should be able
+to use their own credentials. In this case, the preset logic will be used. Credentials will copy to internal preset with a special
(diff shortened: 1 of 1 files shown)


docs/proposals/cluster-templates.md
```diff
@@ -0,0 +1,47 @@
...
+via Kubermatic API or UI.
+
+During cluster creation or after successful creation the user can store cluster definition as a template.
+
+### Scope
+
+The cluster templates should be accessible from different levels. Should have global (admin), project or user scope.
+Template management will be available from project level.
+We also need a central template management place for admins.
+
```
toschneck, line 14: Potential an enforcement of Templates should be respected as well

zreigz: Potentially it's possible but not so easy. I will prepare the whole implementation to support this in the future. The best will be to label the newly created clusters to indicate that they are created from the template

zreigz:
@jiachengxu @rastislavs the whole UI/API team has discussed this topic together with Sascha. Please take a look from your side if it makes sense or you need some extension.

docs/proposals/cluster-templates.md
```diff
@@ -0,0 +1,47 @@
...
+A cluster template is a reusable cluster template object. The Kubernetes CRD will be used for this purpose.
+A cluster template allows you to specify a provider, node layout, and configurations to materialize a cluster instance
+via Kubermatic API or UI.
+
+During cluster creation or after successful creation the user can store cluster definition as a template.
+
+### Scope
+
+The cluster templates should be accessible from different levels. Should have global (admin), project or user scope.
+Template management will be available from project level.
```
jiachengxu, line 12: Does this mean that the cluster template can not be shared across projects? For example, if a user creates a cluster in one project, then this user cannot use the same template in another project?

zreigz: can be share for the same user in different projects when the template has this user scope. Then the user can use the template in every assigned project

PR #7107, opened 2021-05-26
Implement rule group controller

### LINKED ISSUE #7087: Cortex Ruler Controller
Introduce a user-cluster-namespace-scoped CRD `Rule` or `AlertmanagerRule`
Introduce a Controller that reconciles `Rule` CRs from user-cluster namespace to Cortex Rules API endpoint via REST API.

jiachengxu:
This PR introduces a `RuleGroup` CRD and also a `RuleGroup` controller to reconcile `RuleGroup` objects to the rule groups in Cortex. Those rule groups will be used to generate alerts from metrics.
Ref: Cortex rule API: https://cortexmetrics.io/docs/api/#ruler

Fixes #7087 

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/rulegroup_controller.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+}
+
+func (r *ruleGroupController) reconcile(ctx context.Context, cluster *kubermaticv1.Cluster, ruleGroup *kubermaticv1.RuleGroup) (*reconcile.Result, error) {
+	requestURL, err := r.getRequestURL(ruleGroup)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get request URL: %w", err)
+	}
+
+	if !ruleGroup.DeletionTimestamp.IsZero() {
```
aborilov, line 202: why do we need this condition? 
you can call `handleDeletion` in the next one

jiachengxu: This is need for the case that user only wants to delete a `RuleGroup`, not because the cluster is being deleted or mla is disabled

aborilov: oh, I see, it's `DeletionTimestamp` on `ruleGroup` and the next one on the `cluster`

pkg/controller/seed-controller-manager/mla/rulegroup_controller.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	enqueueRuleGroupsForCluster := handler.EnqueueRequestsFromMapFunc(func(object ctrlruntimeclient.Object) []reconcile.Request {
+		cluster := object.(*kubermaticv1.Cluster)
+		if cluster.Status.NamespaceName == "" {
+			return nil
+		}
+		ruleGroupList := &kubermaticv1.RuleGroupList{}
+		if err := client.List(context.Background(), ruleGroupList, ctrlruntimeclient.InNamespace(cluster.Status.NamespaceName)); err != nil {
+			log.Errorw("failed to list ruleGroups for cluster", zap.Error(err), "cluster", cluster.Name)
```
aborilov, line 105: the error will be also logged by `HandleError`

jiachengxu: `HandleError` uses klog underneath, and I am not familiar it at all. Maybe let's just log it with our own zap logger, and we can make sure this is logged when bad thing happens here.

mfranczy (approved):
/lgtm

PR #7135, opened 2021-05-27
rerun of cloud initialization duplicates finalizers

furkhat:
Signed-off-by: Furkhat Kasymovgeniiuulu <vailodf@gmail.com>

**What this PR does / why we need it**: If `InitializeCloudProvider` fails, second run re-adds finalizers. For example we added finalizer for a network but could not create the network, thus spec.network is still empty, and on the next run we add finalizer again. And this can continue indefinitely. 

Continuation of #6759

```release-note
Fix finalizers duplication.
```

### DIFF
--- a/pkg/provider/cloud/openstack/provider_test.go
+++ b/pkg/provider/cloud/openstack/provider_test.go
@@ -29,6 +29,7 @@ import (
 	ostesting "k8c.io/kubermatic/v2/pkg/provider/cloud/openstack/internal/testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/provider/cloud/openstack/provider_test.go
```diff
@@ -24,6 +24,9 @@ import (
 	"github.com/go-test/deep"
 	"github.com/gophercloud/gophercloud"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
irozzo-1A:
```suggestion
```

pkg/provider/cloud/openstack/provider_test.go
```diff
@@ -24,6 +24,9 @@ import (
...
+	"k8s.io/apimachinery/pkg/util/sets"
```
irozzo-1A: Could you please move this after line 34?

PR #7136, opened 2021-05-27
Mark kubelet-configmap addon for continuous reconciliation

### LINKED ISSUE #4979: enforcement of default addons in user cluster

kron4eg:
Fixes #4979

**Documentation**:

```release-note
Mark kubelet-configmap addon for continuous reconciliation
```

### DIFF
--- a/charts/kubermatic/Chart.yaml
+++ b/charts/kubermatic/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubermatic
-version: 1.1.42
+version: 1.1.43
 appVersion: '__KUBERMATIC_TAG__'
 description: Kubermatic chart for master and/or seed clusters.
 deprecated: true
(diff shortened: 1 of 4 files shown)


xmudrii:
I'm a bit worried that this is a breaking change. What if users have customized their kubelet-configmaps? This is a valid use case if they want to adjust some Kubelet properties. There are two important questions that we should answer and document:

- Is this going to affect the existing clusters and nodes?
- Are users still able to adjust their Kubelet configuration without having to build a new addons image?

kron4eg:
@xmudrii very good point! Seems like mashing together kubelet-config and kubeadm-config was not so good idea and we are going to need to extract kubeadm config into own addon.

kron4eg:
I'm closing this PR as that will break existing expectations.

New issue #7141 has been created to address the problem at hand here, but using the other way around.

PR #7149, opened 2021-05-31
[ee] Adds Constraint cluster filtering

### LINKED ISSUE #6633: Default Constraints - filtering
Add filters similar to the ones for the Constraint Template, so admins can set Constraints to target only some clusters.

This is an EE feature

imharshita:
Implements constraint filters which are used in Kubermatic EE to allow users to filter out Constraints for different clusters.

Fixes #6633

```release-note
none
```


pkg/crd/kubermatic/v1/constraint.go
```diff
@@ -48,21 +48,18 @@ type ConstraintSpec struct {
 	ConstraintType string `json:"constraintType"`
 	// Match contains the constraint to resource matching data
 	Match Match `json:"match,omitempty"`
-	// Parameters specifies the parameters used by the constraint template REGO.
```
lsviben: why is this being removed?

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -216,15 +230,24 @@ func (r *reconciler) patchFinalizer(ctx context.Context, constraint *kubermaticv
...
 	// constraint deletion
 	if !constraint.DeletionTimestamp.IsZero() {
+
 		if !kuberneteshelper.HasFinalizer(constraint, finalizer) {
 			return nil
 		}
 
-		if err := r.cleanupConstraint(ctx, log, constraint); err != nil {
-			return err
+		if clusterList != nil {
```
lsviben: you dont need to check if a list is nil, the iteration in `syncAllClustersNS` will work fine.

Same for the checks below :)

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -180,9 +194,9 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 
 func addLabel(constraint *kubermaticv1.Constraint) *kubermaticv1.Constraint {
 	if constraint.Labels != nil {
-		constraint.Labels[Key] = Value
+		constraint.Labels[Key] = constraint.Name
 	} else {
-		constraint.Labels = map[string]string{Key: Value}
+		constraint.Labels = map[string]string{Key: constraint.Name}
```
lsviben, line 192: why is this change needed?

pkg/ee/constraint-controller/util.go
```diff
@@ -0,0 +1,178 @@
...
+	constraint *kubermaticv1.Constraint, workerNamesLabelSelector labels.Selector) ([]kubermaticv1.Cluster, []kubermaticv1.Cluster, error) {
+	var desiredClusterNames clusterNamesList
+
+	desiredList, err := getDesiredClusterListForConstraint(ctx, client, constraint, workerNamesLabelSelector)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed listing clusters: %w", err)
+	}
+
+	desiredClusterList := desiredList.Items
+
```
lsviben: I think a simpler solution here would be to get all clusters (simply with client like in the CE code) and then just do a diff of the desiredClusterList and allClustersList .

imharshita: But that list will contain clusters where constraint is not present. suppose we have 15 clusters and constraints are present on only 3 desired clusters then and 2 as unwanted. then instead of running the loop for 2 clusters (for deletion) we will be running the loop for all the 12 clusters unnecessarily and probably the logs will show that we are trying to delete the resource which is not present

pkg/controller/seed-controller-manager/constraint-controller/controller.go
imharshita: did this in order to get all the existing default constraints in user-cluster-ns for a request from kubermatic/constraint

pkg/controller/seed-controller-manager/constraint-controller/controller.go
imharshita: right

pkg/ee/constraint-controller/util.go
lsviben: Thats true, but I am not sure if that bothers us. We can eventually change the logs to log after a succesfull deletion, and just skip error log when not found.

But I think its a simpler and less bug-prone solution. @irozzo-1A WDYT?

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -216,15 +230,23 @@ func (r *reconciler) patchFinalizer(ctx context.Context, constraint *kubermaticv
...
+
 	// constraint deletion
 	if !constraint.DeletionTimestamp.IsZero() {
+
 		if !kuberneteshelper.HasFinalizer(constraint, finalizer) {
 			return nil
 		}
 
-		if err := r.cleanupConstraint(ctx, log, constraint); err != nil {
+		if err := r.cleanupConstraint(ctx, log, constraint, clusterList); err != nil {
```
irozzo-1A: Here you could delete all the constraints matching the label selector instead of issuing delete requests one by one.

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -216,15 +230,23 @@ func (r *reconciler) patchFinalizer(ctx context.Context, constraint *kubermaticv
 
 func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Constraint, log *zap.SugaredLogger) error {
 
+	clusterList, unwantedClusterList, err := r.getClustersForConstraint(ctx, constraint)
```
irozzo-1A: Listing all clusters is something that is needed both for CE and EE editions, I would factor it out. I would do something like:

```
	clusterList := &kubermaticv1.ClusterList{}
	if err := r.seedClient.List(ctx, clusterList, &ctrlruntimeclient.ListOptions{LabelSelector: r.workerNameLabelSelector}); err != nil {
		return nil, nil, fmt.Errorf("failed listing clusters: %w", err)
	}
	wanted, unwanted := r.filterClusters(clusterList, constraint)
```

WDYT?

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -235,19 +257,34 @@ func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Con
 		}
 	}
 
-	// constraint creation
+	if err = r.cleanupConstraint(ctx, log, constraint, unwantedClusterList); err != nil {
+		return err
+	}
+
+	if err = r.createConstraint(ctx, log, constraint, clusterList); err != nil {
```
irozzo-1A: I would use `ensure` rather than `create` as it does not always imply a creation.

pkg/ee/constraint-controller/util.go
irozzo-1A: I think that we can start simply as @lsviben is suggesting and favor readability and robustness. I'm not concerned about the logs, but definitely, we can save some API calls. I would rather do that in `cleanupConstraint` by providing the list of current constraints so that we are not calling Delete unconditionally (BTW this is an optimization and can be done in a follow-up PR).

Note that if we build on top of what I suggested before, you could simply implement a filtering function, that could go through the list of all clusters given in input and avoid additional API calls.

irozzo-1A (commented):
Overall LGTM, but I agree with @lsviben that we can simplify a bit.

pkg/ee/constraint-controller/util.go
```diff
@@ -0,0 +1,125 @@
...
+	existing := sets.NewString(existingClusterNames...)
+	desired := sets.NewString(desiredClusterNames...)
+
+	unwantedClusterNames := existing.Difference(desired)
+	unwantedClusterList := getClusterList(ctx, client, unwantedClusterNames.List())
+	unwantedList.Items = unwantedClusterList
+
+	return desiredList, unwantedList, nil
+}
```
lsviben: WDYT about something like this? It avoids additional calls and sets and just does the simple filtering on the existing list?
```suggestion
// FilterClustersForConstraint gets clusters for the constraints by using the constraints selector to filter out unselected clusters
func FilterClustersForConstraint(constraint *kubermaticv1.Constraint, clusterList *kubermaticv1.ClusterList) ([]kubermaticv1.Cluster, []kubermaticv1.Cluster, error) {
	constraintLabelSelector, err := v1.LabelSelectorAsSelector(&constraint.Spec.Selector.LabelSelector)
	if err != nil {
		return nil, nil, fmt.Errorf("error converting Constraint label selector (%v) to a kubernetes selector: %w", constraint.Spec.Selector.LabelSelector, err)
	}

	providersSet := sets.NewString(constraint.Spec.Selector.Providers...)

	var unwantedList []kubermaticv1.Cluster
	var desiredList []kubermaticv1.Cluster

	for _, cluster := range clusterList.Items {
		if !constraintLabelSelector.Matches(labels.Set(cluster.Labels)) {
			unwantedList = append(unwantedList, cluster)
			continue
		}
		
		name, err := provider.ClusterCloudProviderName(cluster.Spec.Cloud)
		if providersSet.Len() != 0 && err != nil && !providersSet.Has(name) {
			unwantedList = append(unwantedList, cluster)
			continue
		}
		desiredList = append(desiredList, cluster)
	}

	return desiredList, unwantedList, nil
}
```

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -235,29 +263,51 @@ func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Con
...
 			},
-		})
+		}
+
+		err := seedClient.Get(ctx, types.NamespacedName{Name: constraint.Name, Namespace: namespace}, constraint)
+		if err != nil && errors.IsNotFound(err) {
+			return nil
+		}
+
+		err = seedClient.Delete(ctx, constraint)
```
lsviben: I think just delete is fine, if its not there we get the same result.

irozzo-1A: This is to avoid doing API calls when is not needed, the `Get` is cached while delete will hit KAS.

lsviben: ahh good point

irozzo-1A: @harshita-sharma012 a comment is welcome as this is not obvious at first glance

imharshita: @lsviben do we need this part of code as with the new changes you suggested for filtering we are not getting any unnecessary clusters

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -235,38 +261,58 @@ func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Con
...
 				Namespace: namespace,
 			},
-		})
+		}
+
+		// to avoid performing API calls when not needed, the Get is cached while delete will hit the kubermatic api server
+		err := seedClient.Get(ctx, types.NamespacedName{Name: constraint.Name, Namespace: namespace}, constraint)
+		if err != nil || errors.IsNotFound(err) {
+			return nil
+		}
```
irozzo-1A: We should swallow the error here.

```suggestion
		if err != nil {
		         if errors.IsNotFound(err) {
		             return nil
		         }
			return err
		}
```

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -154,19 +152,31 @@ func Add(ctx context.Context,
 	return nil
 }
 
+// ByLabel returns a predicate func that only includes objects with the given label
+func ByLabel(key string) predicate.Funcs {
+	return kubermaticpred.Factory(func(o ctrlruntimeclient.Object) bool {
+		labels := o.GetLabels()
+		if labels != nil {
```
irozzo-1A: nit: when you read only from a map no need of doing `nil` check

pkg/controller/seed-controller-manager/constraint-controller/controller.go
```diff
@@ -216,15 +226,31 @@ func (r *reconciler) patchFinalizer(ctx context.Context, constraint *kubermaticv
 
 func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Constraint, log *zap.SugaredLogger) error {
 
+	clusterList := &kubermaticv1.ClusterList{}
+	if err := r.seedClient.List(ctx, clusterList, &ctrlruntimeclient.ListOptions{LabelSelector: r.workerNameLabelSelector}); err != nil {
+		return fmt.Errorf("failed listing clusters: %w", err)
+	}
+
+	desiredClusterList, unwantedClusterList, err := r.filterClustersForConstraint(ctx, constraint, clusterList)
```
irozzo-1A: nit: I tend to shorten a bit the name `List` is superfluous IMO, I would use `desiredClusters`, `unwantedClusters`

pkg/ee/constraint-controller/util.go
```diff
@@ -0,0 +1,71 @@
...
+			continue
+		}
+
+		name, err := provider.ClusterCloudProviderName(cluster.Spec.Cloud)
+
+		if err != nil || providersSet.Len() != 0 && !providersSet.Has(name) {
+			unwantedList = append(unwantedList, cluster)
+			continue
+		}
+		desiredList = append(desiredList, cluster)
```
irozzo-1A: Same comment as before, we should not swallow the error here.

```suggestion
		name, err := provider.ClusterCloudProviderName(cluster.Spec.Cloud)
                 if err != nil {
                     return nil, nil, err
                 }
		if providersSet.Len() != 0 && !providersSet.Has(name) {
			unwantedList = append(unwantedList, cluster)
			continue
		}
		desiredList = append(desiredList, cluster)
```

pkg/ee/constraint-controller/util.go
```diff
@@ -0,0 +1,71 @@
...
+
+// FilterClustersForConstraint gets clusters for the constraints by using the constraints selector to filter out unselected clusters
+func FilterClustersForConstraint(ctx context.Context, client ctrlruntimeclient.Client, constraint *kubermaticv1.Constraint, clusterList *kubermaticv1.ClusterList) ([]kubermaticv1.Cluster, []kubermaticv1.Cluster, error) {
+	constraintLabelSelector, err := v1.LabelSelectorAsSelector(&constraint.Spec.Selector.LabelSelector)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error converting Constraint label selector (%v) to a kubernetes selector: %w", constraint.Spec.Selector.LabelSelector, err)
+	}
+	providersSet := sets.NewString(constraint.Spec.Selector.Providers...)
+
+	var unwantedList []kubermaticv1.Cluster
```
irozzo-1A: nit: as before `unwanted` and `desired` seems enough to me ;-)

irozzo-1A:
@harshita-sharma012 Overall LGTM, just a few nits and some comments about error handling.

PR #7158, opened 2021-06-01
Logrotate removal

### LINKED ISSUE #7177: Remove Logrotate addon
**User Story**
Logrotate runs as a daemonset to perform log rotation with the following configuration:

* interval: hourly (hourly rotation)
* max-size: 50M (additional rotates, when the logfile exceeds 50MB)
* max-age: 60 (remove rotated logs older than 60 days)
* directories=/var/lib/docker/containers /var/log/docker (rotates only docker log files)

In addition to logrotate, the machine-controller docker config explicitly sets 100MB as the maximum size for log files. This means that actually the docker configuration is ignored, being the logrotate configuration stricter. Besides, logrotate does not handle log files when using Containerd as CRI (where kubelet flags can be used).

Logrotate should be removed in favor of the docker native log rotation.

**Acceptance criteria**
The log rotation for docker is performed only by the native Docker feature.

mlavacca:
Logrotate runs as a daemonset to perform log rotation with the following configuration:
* interval: hourly (hourly rotation)
* max-size: 50M (additional rotates, when the logfile exceeds 50MB)
* max-age: 60 (remove rotated logs older than 60 days)
* directories=/var/lib/docker/containers /var/log/docker (rotates only docker log files)

In addition to logrotate, the machine-controller docker config explicitly sets 100MB as the maximum size for log files. This means that actually the docker configuration is ignored, being the logrotate configuration stricter. Besides, logrotate does not handle log files when using Containerd as CRI (where kubelet flags can be used).

This PR removes logrotate in favor of the docker built-in log rotation. A PR on the machine controller will follow to set docker log rotation parameters as the same as the Kubelet defaults.

By removing logrotate we do one step further in the user-cluster multiarch image epic issue #6584 since it is one of the images missing the multiarch support.

Fixes #7177 

**Documentation**:

```release-note
NONE
```


kron4eg:
> Chart kubermatic was modified but its version (1.1.43) was not changed. Please adjust charts/kubermatic/Chart.yaml.

kron4eg (approved):

mlavacca:
> > Chart kubermatic was modified but its version (1.1.43) was not changed. Please adjust charts/kubermatic/Chart.yaml.

I tried simply incrementing the version and it isn't enough. I also run `hack/update-kubermatic-chart.sh` locally, but there is no change. Any hints @kron4eg?

kron4eg:
@mlavacca yeah, as always please rebase

pkg/controller/operator/common/defaults.go
```diff
@@ -838,12 +838,6 @@ items:
     name: pod-security-policy
     labels:
       addons.kubermatic.io/ensure: true
-- apiVersion: kubermatic.k8s.io/v1
-  kind: Addon
-  metadata:
-    name: logrotate
-    labels:
-      addons.kubermatic.io/ensure: true
```
kron4eg, line 846: It appears that removing this should also result to data migration (in kubermatic-installer).

mlavacca: I created a [PR](https://github.com/kubermatic/kubermatic-installer/pull/137) in kubermatic-installer, even if I'm not sure about it. Isn't it supposed to be used for kubermatic < 2.14?

kron4eg: Sorry, I meant this kubermatic-installer https://github.com/kubermatic/kubermatic/tree/master/cmd/kubermatic-installer

P.S.
We should get rid of that one in dedicated repo.

mlavacca: ok, gonna close the PR then :)

mlavacca: May I add data migration to this PR or does it worth a separate PR?

kron4eg: yeah, in here would be perfect

kron4eg (approved):

PR #7159, opened 2021-06-01
Be explicit about the need of the NET_RAW capability for OpenVPN Init…

### LINKED ISSUE #7147: Under CRI-O, openvpn-server doesn't initialize
**User Story**

While using a Seed cluster that uses CRI-O, openvpn-server doesn't initialize properly as iptables fails on the InitContainer. It works ok with containerd.
Upon investigation, the issue seems to be that the capability NET_RAW is needed but missing on the container. This may stem from the fact that CRI-O v1.18.0 had the following API change:

> CRI-O now runs containers without NET_RAW and SYS_CHROOT capabilities by default. This can result in permission denied errors when the container tries to do something that would require either of these capabilities. For instance, using ping requires NET_RAW, unless the container is given the sysctl net.ipv4.ip_forward. Further, if you have a container that runs buildah or configures RPMs, they may fail without SYS_CHROOT. Ultimately, the dropped capabilities are worth it, as the majority of containers don't need them. The fewer capabilities CRI-O gives out by default, the more secure it is by default. (#3119, @haircommander)

**Acceptance criteria**

`openvpn-server` pods are able to start when the seed cluster uses CRI-O as container runtime.

cns-ubiwhere:
…Container.

Explicitly adds NET_RAW capability to OpenVPN-server container, it seems to be required for the iptables InitContainer setup.

Fixes #7147 

I'm not familiar with Go at all, and haven't really tested compiling this change :/

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/resources/openvpn/deployment.go
+++ b/pkg/resources/openvpn/deployment.go
@@ -169,7 +169,10 @@ iptables -A INPUT -i tun0 -j DROP
 					},
 					SecurityContext: &corev1.SecurityContext{
 						Capabilities: &corev1.Capabilities{
-							Add: []corev1.Capability{"NET_ADMIN"},
+							Add: []corev1.Capability{
+								"NET_ADMIN",
+								"NET_RAW",
+							},
 						},
 						ProcMount: &procMountType,
 					},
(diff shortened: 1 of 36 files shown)


irozzo-1A:
Thanks @cns-ubiwhere for your contribution!

To make the validation pass you should update the fixtures used in the unit test, you can run `hack/update-fixtures.sh` and commit the changes.

PR #7174, opened 2021-06-03
Upgrade canal

### LINKED ISSUE #6016: Upgrade Canal as a CNI for user clusters.

irozzo-1A:
This PR upgrades canal addon.

Fixes #6016 

The old manifest was not removed from the addon to avoid forcing migration of already existing clusters, as the upgrade path has to be tested carefully. The upgrade is out of scope from this PR.

**Documentation**:

```release-note
NONE
```


mlavacca:
Fixes #7180

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -61,6 +61,11 @@ const (
 	MaxEtcdClusterSize     = 9
 )
 
+const (
+	AnnotationCNIPluginKind    = "cni-plugin.k8c.io/kind"
```
rastislavs: I may be missing something, but I do not see where these annotations are being used

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -96,6 +96,15 @@ func (h *AdmissionHandler) Handle(ctx context.Context, req webhook.AdmissionRequ
 }
 
 func (h *AdmissionHandler) applyDefaults(c *kubermaticv1.Cluster) {
+	// Add cni-plugin annotation if not present pointing to latest canal
```
rastislavs: actually this does not seem to be adding annotation as the comment suggests, but modifying the cluster spec

pkg/crd/kubermatic/v1/cluster.go
irozzo-1A: Good catch :-) I started using the annotations then I thought that extending the cluster CRD was cleaner.

rastislavs: yeah I thought so :)

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -152,7 +152,7 @@ type DatacenterSpec struct {
 
 	//nolint:staticcheck
 	//lint:ignore SA5008 omitgenyaml is used by the example-yaml-generator
-	Fake *DatacenterSpecFake `json:"fake,omitempty,omitgenyaml"`
+	Fake *DatacenterSpecFake `json:"fake,omitempty,omitempty"`
```
wozniakjan: nit: looks like duplicated `omitempty` parameter, is that intentional?

irozzo-1A: nope, this is totally unintentional. Thx for pointing out ;-)

PR #7183, opened 2021-06-04
Handle deletion of RuleGroup if the Cluster object is gone

jiachengxu:
I observed some behavior that when Cluster is deleted, the deletion of its namespace is blocked because RuleGroup controller cannot be triggered to handle deletion and remove finalizer from RuleGroup object.

This PR adds logic to also handle deletion of RuleGroup if Cluster object is gone, so RuleGroup will not block the deletion of cluster's namespace. 

Fixes #

**Documentation**:

```release-note
NONE
```


aborilov (commented):
/lgtm

jiachengxu:
@rastislavs I need a /lgtm again after merging with master :) 
/unhold

PR #7185, opened 2021-06-04
ExternalCloudProvider migration endpoint

### LINKED ISSUE #6689: Support CCM/CSI migration in Kubermatic API
In order to provide to KKP end user the possibility of migrating their clusters from in-tree cloud provider to external CCM/CSI we need to integrate this feature in the API.

It should be responsible for adding the `ExternalCloudProvider` feature flag in the corresponding Cluster resource.

mlavacca:
Addition of the `ExternalCloudProvider` flag in the cluster API and the logic to reflect that value in the cluster crd.

Fixes #6689 

**Documentation**:

```release-note
NONE
```


maciaszczykm:
@mlavacca Check @lsviben message in #testing channel. Perhaps OPA tests should be overwritten in your case.

mlavacca:
> @mlavacca Check @lsviben message in #testing channel. Perhaps OPA tests should be overwritten in your case.

Yeah, I guess so. Let's wait for sig/cluster lifecycle review, then I'll ask an admin to override it.

pkg/handler/common/cluster.go
```diff
@@ -185,6 +185,13 @@ func CreateEndpoint(ctx context.Context, projectID string, body apiv1.CreateClus
 		partialCluster.Spec.EnableUserSSHKeyAgent = body.Cluster.Spec.EnableUserSSHKeyAgent
 	}
 
+	partialCluster.Spec.Features = make(map[string]bool)
+	if body.Cluster.Spec.ExternalCloudProvider == nil {
+		partialCluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] = false
```
xmudrii: We also have some logic for enabling the `ClusterFeatureExternalCloudProvider` on L198-200. Should we merge this code with the code below to avoid confusion?

pkg/api/v1/types.go
```diff
@@ -800,6 +800,9 @@ type ClusterSpec struct {
 	// ClusterNetwork contains network settings.
 	ClusterNetwork *kubermaticv1.ClusterNetworkingConfig `json:"clusterNetwork,omitempty"`
 
+	// ExternalCloudProvider enables the automatic migration from in-tree cloud provider to external CCM/CSI
+	ExternalCloudProvider *bool `json:"externalCloudProvider,omitempty"`
```
xmudrii: I'm worried that someone might use this to disable the `ExternalCloudProvider` feature when it should be enabled. The external CCM should be enabled by default for all newly created clusters that support the external CCM (there's a function for checking that).

Should we extend validation to validate create requests and reject the request if the provider supports the external CCM, but it's disabled?

mlavacca: There is an aspect not clear to me: [`ExternalCloudControllerFeatureSupported`](https://github.com/kubermatic/kubermatic/blob/4e1b8c631a9f09cea31b666859b127d2c13fef56/pkg/resources/cloudcontroller/util.go#L29) automatically enables the `externalCloudProvider` feature if the provider supports it. My question is: should be the user allowed to choose if to enable it or not? If not, this flag appears to be useless to me.

xmudrii: > There is an aspect not clear to me: ExternalCloudControllerFeatureSupported automatically enables the externalCloudProvider feature if the provider supports it.

It's enabled by the following code for the new clusters: https://github.com/kubermatic/kubermatic/blob/cc2db9dad1faea9e65a3c5421987cdf050262ab6/pkg/handler/common/cluster.go#L191-L193

> My question is: should be the user allowed to choose if to enable it or not? If not, this flag appears to be useless to me.

For new clusters, they shouldn't be able to choose.

For existing clusters, they should be able to enable the `ExternalCloudProvider` feature if it isn't enabled. That would trigger the CCM/CSI migration. From that standpoint, this field might be useful, but you should check that with the SIG API.

maciaszczykm: As discussed, let's remove this and move the information to the cluster status. Changes (migration) can be done using the POST endpoint (to be implemented).

maciaszczykm:
/approve

@xmudrii @iacopo Can you give final LGTM?

pkg/api/v1/types.go
```diff
@@ -1045,9 +1045,21 @@ func newPublicAnexiaCloudSpec(internal *kubermaticv1.AnexiaCloudSpec) (public *P
...
+
+// ExternalCCMStatus represents the external CCM migration status
+// swagger:model ExternalCCMStatus
+type ExternalCCMStatus struct {
+	// the ClusterFeatureExternalCloudProvider in the cluster .spec.features
+	ExternalCCM bool `json:"externalCCM"`
+	// the logic AND of the CCMMigrationNeededAnnotation and CSIMigrationNeededAnnotation annotations
+	MigrationNeeded *bool `json:"migrationNeeded,omitempty"`
+	// reflect the ClusterConditionCSIKubeletMigrationCompleted condition of the cluster
+	MigrationCompleted *bool `json:"migrationCompleted,omitempty"`
```
xmudrii: Tiny nit: I think those should be proper Go comments (e.g. `ExternalCCM represents...`).

pkg/api/v1/types.go
```diff
@@ -1045,9 +1045,21 @@ func newPublicAnexiaCloudSpec(internal *kubermaticv1.AnexiaCloudSpec) (public *P
...
+	// CCM represents the externalCCM status
+	CCM ExternalCCMStatus `json:"ccm"`
+}
+
+// ExternalCCMStatus represents the external CCM migration status
+// swagger:model ExternalCCMStatus
+type ExternalCCMStatus struct {
+	// the ClusterFeatureExternalCloudProvider in the cluster .spec.features
+	ExternalCCM bool `json:"externalCCM"`
+	// the logic AND of the CCMMigrationNeededAnnotation and CSIMigrationNeededAnnotation annotations
```
xmudrii: This comment/description is a little bit confusing to me. Would it be better to say something along `MigrationNeeded reflects are the CCMMigrationNeeded and CSIMigrationNeeded annotations set`?

pkg/api/v1/types.go
```diff
@@ -1045,9 +1045,21 @@ func newPublicAnexiaCloudSpec(internal *kubermaticv1.AnexiaCloudSpec) (public *P
...
+	MigrationNeeded *bool `json:"migrationNeeded,omitempty"`
```
xmudrii, line 1060: This logic doesn't reflect is the migration needed. Instead, it reflects if the migration is in progress. To check if the migration is needed, you need to make sure that:

* ExternalCCM is `false`
* The provider supports the external CCM (using the function we already have)

The annotations are applied only when the migration has already started (by the mutating webhook).

xmudrii (changes requested):

pkg/api/v1/types.go
```diff
@@ -1045,9 +1045,22 @@ func newPublicAnexiaCloudSpec(internal *kubermaticv1.AnexiaCloudSpec) (public *P
...
+	CCM ExternalCCMStatus `json:"ccm"`
+}
+
+// ExternalCCMStatus represents the external CCM migration status
+// swagger:model ExternalCCMStatus
+type ExternalCCMStatus struct {
+	// ExternalCCM reflects the ClusterFeatureExternalCloudProvider in the cluster .spec.features
+	ExternalCCM bool `json:"externalCCM"`
+	// MigrationNeeded reflects if the CCMMigrationNeeded and CSIMigrationNeeded annotations are set and the cloud
+	// provider supports external CCM
```
xmudrii: Nit: this comment is not correct anymore.

pkg/handler/routes_v1.go
```diff
@@ -35,7 +35,6 @@ limitations under the License.
 //          in: header
 //
 // swagger:meta
-
```
maciaszczykm: Check the comment above: 

```
// NB: The following block has an empty line separating it from the
// package declaration. This is intentional, to not treat the Swagger
// spec as a godoc. During client generation, the empty line is
// temporarily removed.
```

xmudrii (approved):
/lgtm
/approve
/hold
for @maciaszczykm and @floreks to take a look as well

maciaszczykm:
LGTM.

PR #7189, opened 2021-06-07
Cluster deletion to leave dynamic PVs cleanup to provisioner

wozniakjan:
Cluster deletion cleanup leaves volumes in a cloud provider.
This avoids deleting `PV`s entirely, only waits for dynamically provisioned `PV`s with delete reclamation policy to be deleted by the provisioner.

touching https://github.com/kubermatic/kubermatic/issues/6406#issuecomment-848689649

The `DeleteVolume` CSI gRPC call for cinder plugin is responsible for cleanup of the OpenStack volumes. For CSI, deleting PV manually doesn't trigger `DeleteVolume` which can leave unexpected residue behind the cleanup logic. 

https://github.com/kubernetes/community/blob/9d56458d1d8aadcba53a664b0ccf080d07583e45/contributors/design-proposals/storage/container-storage-interface.md#deleting-volumes

Issue #6406 touched the topic of OpenStack volumes cleanup and will require a different fix but it also uncovered that during cluster cleanup we leak volumes that stay after the cluster no longer exist. An example can be found in https://cloud.syseleven.de/horizon/project/volumes/. The plan for #6406 is still to split the CSI installation between seed and user cluster, not using addon anymore but install similarly to `pkg/resources/openvpn/` (that will be another PR). Eventually, I would also like to examine the cinder CSI plugin and see if there is room for improvement.

<details><summary>screenshot</summary>
<p>

cinder-leaked-volumes

</p>
</details>

**Documentation**:

kubermatic/dashboard#3429 - informing better user that some volumes may not get deleted in cloud provider

```release-note
NONE
```

### DIFF
--- a/pkg/clusterdeletion/volumes.go
+++ b/pkg/clusterdeletion/volumes.go
@@ -55,7 +55,7 @@ func (d *Deletion) cleanupVolumes(ctx context.Context, cluster *kubermaticv1.Clu
 
 	allPVList := &corev1.PersistentVolumeList{}
 	if err := userClusterClient.List(ctx, allPVList); err != nil {
-		return deletedSomeResource, fmt.Errorf("failed to list PVs from user cluster: %v", err)
+		return false, fmt.Errorf("failed to list PVs from user cluster: %v", err)
 	}
 	pvList := &corev1.PersistentVolumeList{}
 	for _, pv := range allPVList.Items {
--- a/pkg/clusterdeletion/volumes.go
+++ b/pkg/clusterdeletion/volumes.go
@@ -30,6 +30,11 @@ import (
 
 const (
 	annotationKeyDescription = "description"
+
+	// AnnDynamicallyProvisioned is added to a PV that is dynamically provisioned by kubernetes
+	// Because the annotation is defined only at k8s.io/kubernetes, copying the content instead of vendoring
+	// https://github.com/kubernetes/kubernetes/blob/v1.21.0/pkg/controller/volume/persistentvolume/util/util.go#L65
+	AnnDynamicallyProvisioned = "pv.kubernetes.io/provisioned-by"
 )
 
 func (d *Deletion) cleanupVolumes(ctx context.Context, cluster *kubermaticv1.Cluster) (deletedSomeResource bool, err error) {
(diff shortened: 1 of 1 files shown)


irozzo-1A:
@wozniakjan is deleting PVs directly useful in some circumstances? When dynamic provisioning is used we are supposed to delete PVCs, if dynamic provisioning is not used deleting the PV resources won't have any effect on the provisioned volumes. Am I missing something? Otherwise, we could simply avoid deleting PVs.

wozniakjan:
> is deleting PVs directly useful in some circumstances?

good question, I'm not sure tbh. I wanted to be on the safe side to only leave `PV`s that are dynamically provisioned for the provisioner to take care of them. But I think there is a scenario that I didn't consider and that is when provisioner `StorageClass` has `ReclaimPolicy` equal to `retain`.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#retain

Given the text above, I would like to check if before the `PVC` cleanup we can modify the `PVC` `ReclaimPolicy` from `retain` to `delete` and then allow provisioners to perform the cleanup. Not sure how to deal with `PV` that are `released` but have `retain` policy, those will likely still create leaks in cloud provider volumes after cluster deletions.

irozzo-1A:
> Given the text above, I would like to check if before the `PVC` cleanup we can modify the `PVC` `ReclaimPolicy` from `retain` to `delete` and then allow provisioners to perform the cleanup. Not sure how to deal with `PV` that are `released` but have `retain` policy, those will likely still create leaks in cloud provider volumes after cluster deletions.

Personally, even if it is possible I'm not sure I would do it. I would rather clarify and document the limits of this feature. WDYT?

wozniakjan:
> Personally, even if it is possible I'm not sure I would do it. I would rather clarify and document the limits of this feature. WDYT?

that actually makes more sense, I agree

starting the effort to better inform users in: https://github.com/kubermatic/dashboard/pull/3429

pkg/clusterdeletion/volumes.go
```diff
@@ -48,10 +53,18 @@ func (d *Deletion) cleanupVolumes(ctx context.Context, cluster *kubermaticv1.Clu
 		return false, fmt.Errorf("failed to list PVCs from user cluster: %v", err)
 	}
 
-	pvList := &corev1.PersistentVolumeList{}
-	if err := userClusterClient.List(ctx, pvList); err != nil {
+	allPVList := &corev1.PersistentVolumeList{}
+	if err := userClusterClient.List(ctx, allPVList); err != nil {
 		return deletedSomeResource, fmt.Errorf("failed to list PVs from user cluster: %v", err)
```
irozzo-1A: @wozniakjan I know it's not you that introduced this, maybe I'm missing something but what's the point of returning `deletedSomeResource` here as it is never set before?

```suggestion
		return false, fmt.Errorf("failed to list PVs from user cluster: %v", err)
```

irozzo-1A:
LGTM, I just asked for a small clarification then it's ready to go.

pkg/clusterdeletion/volumes.go
wozniakjan: you are correct that for this particular `return` statement, `deletedSomeResource` is always going to be `false`. It's only the following `return` statements that may be `true` depending on `PVs` and `PVCs`. Your suggestion makes sense to me.

irozzo-1A:
/test pre-kubermatic-verify
/test pre-kubermatic-api-e2e

@wozniakjan #7230 should make `pre-kubermatic-verify` stable

PR #7199, opened 2021-06-09
CCM-CSI-migration controller

### LINKED ISSUE #6551: Add controller in user-cluster controller manager to notify when nodes are ready to complete CSI migration

mlavacca:
CCM-CSI migration controller in user-cluster-controller-manager

Fixes #6551 

**Documentation**:

```release-note
NONE
```


cmd/user-cluster-controller-manager/main.go
```diff
@@ -19,13 +19,14 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
```
moadqassem: Please re-group the imports here based on this grouping by our coding style https://github.com/kubermatic/kubermatic/blob/master/docs/code-style.md#imports

cmd/user-cluster-controller-manager/main.go
```diff
@@ -99,6 +101,7 @@ func main() {
 	flag.Var(&runOp.networks, "ipam-controller-network", "The networks from which the ipam controller should allocate IPs for machines (e.g.: .--ipam-controller-network=10.0.0.0/16,10.0.0.1,8.8.8.8 --ipam-controller-network=192.168.5.0/24,192.168.5.1,1.1.1.1,8.8.4.4)")
 	flag.StringVar(&runOp.namespace, "namespace", "", "Namespace in which the cluster is running in")
 	flag.StringVar(&runOp.clusterURL, "cluster-url", "", "Cluster URL")
+	flag.StringVar(&runOp.clusterName, "cluster-name", "", "cluster name")
```
moadqassem:
```suggestion
	flag.StringVar(&runOp.clusterName, "cluster-name", "", "Cluster name")
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
moadqassem:
```suggestion
Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ccmcsimigrator
+
+import (
+	"context"
```
moadqassem, line 20: Please re-group the imports here based on this grouping by our coding style https://github.com/kubermatic/kubermatic/blob/master/docs/code-style.md#imports

cmd/user-cluster-controller-manager/main.go
```diff
@@ -300,11 +307,17 @@ func main() {
...
+
 	if err := ownerbindingcreator.Add(rootCtx, log, mgr, runOp.ownerEmail); err != nil {
 		log.Fatalw("Failed to register ownerbindingcreator controller", zap.Error(err))
 	}
 	log.Info("Registered ownerbindingcreator controller")
 
+	if err := ccmcsimigrator.Add(rootCtx, log, seedMgr, mgr, versions, runOp.clusterName); err != nil {
+		log.Fatalw("failed to register ccmcsimigrator controller", zap.Error(err))
+	}
+	log.Info("registered ccmcsimigrator controller")
```
moadqassem:
```suggestion
	log.Info("registered ccm-csi-migrator controller")
```

cmd/user-cluster-controller-manager/main.go
```diff
@@ -300,11 +307,17 @@ func main() {
...
 	}
 	log.Info("Registered rolecloner controller")
+
 	if err := ownerbindingcreator.Add(rootCtx, log, mgr, runOp.ownerEmail); err != nil {
 		log.Fatalw("Failed to register ownerbindingcreator controller", zap.Error(err))
 	}
 	log.Info("Registered ownerbindingcreator controller")
 
+	if err := ccmcsimigrator.Add(rootCtx, log, seedMgr, mgr, versions, runOp.clusterName); err != nil {
+		log.Fatalw("failed to register ccmcsimigrator controller", zap.Error(err))
```
moadqassem:
```suggestion
		log.Fatalw("failed to register ccm-csi-migrator controller", zap.Error(err))
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
...
+		userClient:   userMgr.GetClient(),
+		seedRecorder: seedMgr.GetEventRecorderFor(controllerName),
+		versions:     versions,
+		clusterName:  clusterName,
+	}
+	c, err := controller.New(controllerName, userMgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %v", err)
```
moadqassem:
```suggestion
		return fmt.Errorf("failed to create controller %s: %v", controllerName , err)
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
...
+				},
+			})
+		}
+	}
+
+	// Watch for changes to Machines
+	if err = c.Watch(
+		&source.Kind{Type: &v1alpha1.Machine{}},
+		&handler.Funcs{
+			CreateFunc: func(e event.CreateEvent, queue workqueue.RateLimitingInterface) { commonHandler(e.Object, queue) },
```
moadqassem: Why this is needed? Since this is only a migration controller, the watch resources will not be created.

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
...
+		clusterName:  clusterName,
+	}
+	c, err := controller.New(controllerName, userMgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create controller: %v", err)
+	}
+
+	commonHandler := func(obj ctrlruntimeclient.Object, queue workqueue.RateLimitingInterface) {
```
moadqassem: Actually you can use predicate here instead of using a custom handler. Please check out this file https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/util/predicate/predicate.go and add create a new function to for annotations similar to this one: 
```go
func ByAnnotations(key string) predicate.Funcs {
	return Factory(func(o ctrlruntimeclient.Object) bool {
		annotations := o.GetAnnotations()
		if annotations != nil {
			if _, ok := annotations[key]; ok {
				return true
			}
		}
		return false
	})
}
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller_test.go
```diff
@@ -0,0 +1,181 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
moadqassem:
```suggestion
Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller_test.go
```diff
@@ -0,0 +1,181 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ccmcsimigrator
+
+import (
```
moadqassem, line 19: Please re-group the imports here based on this grouping by our coding style https://github.com/kubermatic/kubermatic/blob/master/docs/code-style.md#imports

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/doc.go
```diff
@@ -0,0 +1,22 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
moadqassem:
```suggestion
Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+	); err != nil {
+		return fmt.Errorf("failed to establish watch for the Machines %v", err)
+	}
+
+	return nil
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	log := r.log.With("Cluster", request.NamespacedName.String())
```
moadqassem:
```suggestion
	log := r.log.With("Request", request.NamespacedName.String())
```

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
mlavacca: Since we moved to the ccm-migration flag mechanism, I removed the filter (if the controller runs, it necessarily has to reconcile the machines for the given cluster)

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
mlavacca: In this case, using a simple helper like this one wouldn't fit because we need to check the annotations on the cluster resource, and not on the triggering one (the machine). BTW, I removed the filter because the ccm-migration flag is injected by the seed-controller-manager only when the migration is strictly required (this condition is checked a priori).

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// Watch for changes to Machines
+	if err = c.Watch(
+		&source.Kind{Type: &v1alpha1.Machine{}},
+		handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Name: clusterName,
```
xmudrii, line 82: Does it make sense to read this value from the Machine object itself instead of passing it via flag? Machine objects have a label called `system/cluster` in `.spec.metadata.labels`.

I'm not sure when this label was added, but I think I've seen it for a while. On the other side, the current approach might be easier.

@moadqassem @mlavacca WDYT about this? I don't have strong opinions. I'm just thinking out loud.

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
```diff
@@ -0,0 +1,173 @@
...
+	machines := &v1alpha1.MachineList{}
+	if err := r.userClient.List(ctx, machines); err != nil {
+		log.Debugw("error while listing machines in user oldCluster", zap.Error(err))
+		return err
+	}
+
+	// check all the machines have been migrated
+	var migrated = true
+	for _, machine := range machines.Items {
+		flag := common.GetKubeletFlags(&machine)[common.ExternalCloudProviderKubeletFlag]
```
xmudrii, line 126: Does it make sense to also check for the CSI migration feature gates?

xmudrii (commented):
Looks great to me! :100: I left two questions that we might want to think about, but the PR looks good even without those changes.

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
mlavacca: Yes, there is that label, and at the beginning, I used it to get the cluster name. But I moved into this approach because it is quite old and the key is hardcoded: https://github.com/kubermatic/kubermatic/blob/39d3d9e4a192b7af9d22f461feb82e44f3391d12/pkg/resources/machine/machinedeployment.go#L66
If someone in the future will remove or change it, there wouldn't be any reference to this controller. BTW, I guess it is a false problem, we can simply remove that hardcoding and refer to a common const. I think both approaches suit well, no strong opinion too.

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
mlavacca: I think that checking also CSI migration feature gates would add an unneeded complexity since all the providers have different feature gates, and we should distinguish between them. Of course, we can do it, but tbh I don't know if it worths, since the CSI feature flags are added along to the `ExternalCloudProviderKubeletFlag`: https://github.com/kubermatic/machine-controller/blob/3a115014a952d06d867ecee5e8d780acafefc289/pkg/admission/machines.go#L87

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
xmudrii: I don't have strong opinions about this, so I'm going to leave this up to you and @moadqassem. I'm also totally fine if we keep the current approach.

pkg/controller/user-cluster-controller-manager/ccm-csi-migrator/controller.go
xmudrii: > I think that checking also CSI migration feature gates would add an unneeded complexity since all the providers have different feature gates, and we should distinguish between them.

While this is true, there's also the `CSIMigration` feature gate that's common for all cloud providers. However, it increases the complexity a little bit, so I'm also not sure is it really worth it. I don't think we'll ever have a situation that Machine might have the external cloud provider enabled, but not the CSIMigration feature gates (for migrated clusters).

I think that we can ignore this for now.

xmudrii:
/lgtm
Will leave it to @moadqassem to take a look and approve the PR

moadqassem (commented):
/approve
/lgtm

xmudrii:
/hold
The vSphere tests are currently broken

PR #7204, opened 2021-06-10
[ee] Adds constraints active flip

### LINKED ISSUE #6806: Constraints - activate/deactivate flip
Activate/Deactivate flip on the Constraints. Instead of deleting Constraints, add a feature to deactivate the Constraint. Either on the admin level if its a default constraint, or on the cluster level for the cluster admins.

This will be an EE feature.

imharshita:

Activate/Deactivate flip on the Constraints. Instead of deleting Constraints, add a feature to deactivate the Constraint. Either on the admin level if its a default constraint, or on the cluster level for the cluster admins.

Fixes #6806

```release-note
none
```


pkg/controller/user-cluster-controller-manager/constraint-syncer/wrappers_ee.go
```diff
@@ -0,0 +1,76 @@
...
+			return fmt.Errorf("failed to set constraint finalizer %s: %v", constraint.Name, err)
+		}
+	}
+
+	// constraint creation
+	if constraint.Spec.Active {
+		if err := r.createConstraint(ctx, constraint, log); err != nil {
+			return err
+		}
+	}
```
lsviben: if its not active, we dont need a finalizer as the constraint wont get created on the user cluster

pkg/controller/user-cluster-controller-manager/constraint-syncer/controller.go
```diff
@@ -145,7 +112,26 @@ func (r *reconciler) reconcile(ctx context.Context, constraint *kubermaticv1.Con
...
+	toDelete.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   constrainthandler.ConstraintsGroup,
+		Version: constrainthandler.ConstraintsVersion,
+		Kind:    constraint.Spec.ConstraintType,
+	})
+	toDelete.SetName(constraint.Name)
+
+	if err := r.userClient.Delete(ctx, toDelete); err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete constraint: %v", err)
+	}
```
irozzo-1A, line 133: Why aren't you using `constraint` directly when you call `Delete`?

irozzo-1A: ok, I think I got it. The resource you are deleting is not the kubermatic `Constraint`. But why using `Unstructured`?

imharshita: coz we are creating/deleting unstructured gatekeeper Constraint object in user-cluster with `constraints.gatekeeper.sh/v1beta1` , while we have `kubermatic.k8s.io/v1` constraint in user-cluster-ns

irozzo-1A: @lsviben yeah, but is there any reason why we are not using the structured version? Is it to avoid an import?

lsviben: The Constraint CRDs for Gatekeeper are dynamic so we need to use unstructured because we never know what we get. 

Gatekeeper-controller creates the CRD for Constraints depending on the Constraint Template the user creates: https://open-policy-agent.github.io/gatekeeper/website/docs/howto/#constraint-templates

irozzo-1A: ok, interesting. Thanks for clarifying.

pkg/controller/user-cluster-controller-manager/constraint-syncer/wrappers_ce.go
```diff
@@ -0,0 +1,68 @@
...
+			return nil
+		}
+
+		if err := r.cleanupConstraint(ctx, constraint, log); err != nil {
+			return err
+		}
+
+		oldConstraint := constraint.DeepCopy()
+		kuberneteshelper.RemoveFinalizer(constraint, kubermaticapiv1.GatekeeperConstraintCleanupFinalizer)
+		if err := r.seedClient.Patch(ctx, constraint, ctrlruntimeclient.MergeFrom(oldConstraint)); err != nil {
```
irozzo-1A, line 47: If we are adding finalizers to the constraint from both user-cluster-controller and seed-cluster-controller we risk to end-up in a race condition like here: #6759
Maybe there is no consequence in this case, but to be safe it would be better to use optimistic locking.

pkg/controller/user-cluster-controller-manager/constraint-syncer/wrappers_ce.go
```diff
@@ -0,0 +1,68 @@
...
+			return fmt.Errorf("failed to remove constraint finalizer %s: %v", constraint.Name, err)
+		}
+		return nil
+	}
+
+	// add finalizer
+	if !kuberneteshelper.HasFinalizer(constraint, kubermaticapiv1.GatekeeperConstraintCleanupFinalizer) {
+		oldConstraint := constraint.DeepCopy()
+		kuberneteshelper.AddFinalizer(constraint, kubermaticapiv1.GatekeeperConstraintCleanupFinalizer)
```
irozzo-1A, line 56: Same here

pkg/controller/user-cluster-controller-manager/constraint-syncer/wrappers_ce.go
imharshita: we are only adding finalizers to constraints in clsuter-ns in  user-cluster-controller

irozzo-1A: ok, in such a case we are safe.

imharshita: yeah seed-controller adds finalizers to constraints in kubermatic ns and removes while cluster-deletion

irozzo-1A: Well, to be more precise we are safe if one single controller is adding removing finalizers, but it should be the case, right?

imharshita: right

imharshita: although in case of cluster-deletion we set the deletion-timestamp(to avoid finalizer getting added again by user-cluster-controller) and remove the finalizer 
user-cluster-controller also has code of removing the finalizer when deletion timestamp is set, do you think it can cause problem?

PR #7208, opened 2021-06-10
added option for Azure Load Balancer SKU in KKP API

### LINKED ISSUE #7195: Add Azure LB SKU type to API

lsviben:
Adds option for users to select Load Balancer SKU when creating Azure clusters. 

Fixes #7195 

```release-note
Added option to set the Load Balancer SKU when creating Azure clusters
```


pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -533,6 +533,8 @@ type AzureCloudSpec struct {
 	RouteTableName    string `json:"routeTable"`
 	SecurityGroup     string `json:"securityGroup"`
 	AvailabilitySet   string `json:"availabilitySet"`
+	// LoadBalancerSKU sets the LB type that will be used for the Azure cluster, possible values are "basic" and "standard", if empty, "basic" will be used
+	LoadBalancerSKU string `json:"loadBalancerSKU"`
```
mfranczy: Because we have clearly defined possible values, could you create a load balancer type for this? For instance
```
type LBSku string

const (
	StandardLBSku LBSku  = "standard"
	BasicLBSku LBSku  = "basic"
)
```

and later use it
```
LoadBalancerSku LBSku `json:"loadBalancerSku"`
```

it's easier to use API and it is less prone to errors

lsviben: good point, ill add it

PR #7213, opened 2021-06-10
Adds constraint finalizer removal

### LINKED ISSUE #6934: Constraints getting stuck in deletion
Sometimes after OPA-enabled cluster deletion, with active constraints, the constraints dont get deleted properly, meaning they are left with finalizers and thus cause the user cluster namespaces on the seed cluster to be stuck as well. 

Investigate and fix the reason for this.

```
$ k get ns
NAME                                 STATUS        AGE
namespace/cert-manager               Active        517d
namespace/cluster-4wzv54h8xr         Terminating   6d1h
namespace/cluster-52rlpk7cvm         Active        38m
namespace/cluster-9mz4mwvdt8         Active        3h51m
namespace/cluster-fpcm4m7x8b         Active        9d
namespace/cluster-gngb9dbgpz         Active        85m
namespace/cluster-gpv45s4qdx         Active        38m
namespace/cluster-lbpdsvnfgn         Terminating   6d3h
namespace/cluster-ldkqklqlmc         Active        7d3h
namespace/cluster-lv7smcfnp4         Active        3h50m
namespace/cluster-snipoc             Active        320d
namespace/cluster-vbw2sgswnl         Active        3h51m
namespace/default                    Active        3y185d
namespace/envoy                      Active        331d
...
$ k -n cluster-4wzv54h8xr get constraints.kubermatic.k8s.io
NAME   AGE
test   6d
$ k -n cluster-lbpdsvnfgn get constraints
NAME   AGE
test   6d1h
```

imharshita:

Fixes #6934

```release-note
none
```

### DIFF
--- a/pkg/clusterdeletion/constraints.go
+++ b/pkg/clusterdeletion/constraints.go
@@ -27,6 +27,10 @@ import (
 )
 
 func (d *Deletion) cleanupConstraints(ctx context.Context, cluster *kubermaticv1.Cluster) error {
+	if cluster.Status.NamespaceName == "" {
+		return nil
+	}
+
 	if err := d.seedClient.DeleteAllOf(ctx, &kubermaticv1.Constraint{}, ctrlruntimeclient.InNamespace(cluster.Status.NamespaceName)); err != nil {
 		return err
 	}
--- a/pkg/clusterdeletion/constraints.go
+++ b/pkg/clusterdeletion/constraints.go
@@ -18,6 +18,7 @@ package clusterdeletion
 
 import (
 	"context"
+	"fmt"
 
 	kubermaticapiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
@@ -48,7 +49,9 @@ func (d *Deletion) cleanupConstraints(ctx context.Context, cluster *kubermaticv1
 	for _, constraint := range constraintList.Items {
 		oldConstraint := constraint.DeepCopy()
 		kuberneteshelper.RemoveFinalizer(&constraint, kubermaticapiv1.GatekeeperConstraintCleanupFinalizer)
-		d.seedClient.Patch(ctx, &constraint, ctrlruntimeclient.MergeFrom(oldConstraint))
+		if err := d.seedClient.Patch(ctx, &constraint, ctrlruntimeclient.MergeFrom(oldConstraint)); err != nil {
+			return fmt.Errorf("failed to remove constraint finalizer %s: %v", constraint.Name, err)
+		}
 	}
 
 	oldCluster := cluster.DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/clusterdeletion/constraints.go
```diff
@@ -31,6 +31,18 @@ func (d *Deletion) cleanupConstraints(ctx context.Context, cluster *kubermaticv1
 		return err
 	}
 
+	// removing fianlzers from all constraints
+	constraintList := &kubermaticv1.ConstraintList{}
+	if err := d.seedClient.List(ctx, constraintList, ctrlruntimeclient.InNamespace(cluster.Status.NamespaceName)); err != nil {
```
jiachengxu, line 45: Can `cluster.Status.NamespaceName` be empty `""`?

imharshita:
- Tested that finalizer is being removed and not added again.
- Constraints gets created again (but without finalizer), coz of user-cluster-controller watch for constrains in user-cluster-ns
- Couldn't ensure cluster deletion coz of other reasons

pkg/clusterdeletion/constraints.go
```diff
@@ -31,6 +31,18 @@ func (d *Deletion) cleanupConstraints(ctx context.Context, cluster *kubermaticv1
 		return err
 	}
 
+	// removing fianlzers from all constraints
```
jiachengxu: Can we document in detail the reason that we would like to remove finalizers?
```suggestion
	// `kubermaticapiv1.GatekeeperConstraintCleanupFinalizer` is added by user-cluster-controller-manager/constraints-syncer.
	// It could be the case that during cluster deletion, user-cluster-controller-manager is deleted before it removes
	// the finalizer from constraints object, in this case, the user-cluster namespace will get stuck on deletion.
	// So here we just remove the finalizer from constraints so that user-cluster namespace can be garbage-collected.
	// Ref:https://github.com/kubermatic/kubermatic/issues/6934
```

pkg/clusterdeletion/constraints.go
imharshita: every cluster has a namespace in seed so ` cluster.Status.NamespaceName` would not be empty

jiachengxu: It could be the case that this `Cluster` is deleted right after it was created, or the controller who updates `cluster.Status.NamespaceName` is faulty. In those cases, cluster.Status.NamespaceName can be empty.
so I would suggest to check if this is empty before make API calls, for example, we have some other controllers who are doing similar thing:
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/addon/addon_controller.go#L134-L136

imharshita: sure will add it

imharshita: done

xmudrii (approved):
/approve

PR #7219, opened 2021-06-11
Mount gocache when building user-ssh-keys-agent image

### LINKED ISSUE #7222: Mount GOCACHE to multi-stage user-ssh-keys-agent image build
**User Story**
Since #7060  has been merged, the user-ssh-keys-agent is distributed as a multi-arch image; to do so, the image has to be built in a multi-stage environment, in which the binary is built in the first stage, and then copied to the destination stage. On the other hand, the compile-time of that image has been increased, since it is not possible to rely on the go-cache uploaded to the MinIO bucket.
To speed up the compile-time of the binary in the first stage of image building, the go-cache should be imported and used by the `docker build`.

**Acceptance criteria**
user-ssh-keys-agent multiarch binary built in the first step of `docker build` by using the go-cache downloaded from the MinIO bucket.

mlavacca:
gocache mount while building user-ssh-keys-agent container

Fixes #7222 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/user-ssh-keys-agent/Dockerfile
+++ b/cmd/user-ssh-keys-agent/Dockerfile
@@ -14,13 +14,21 @@
 
 FROM docker.io/golang:1.16.4 as builder
 
+ARG HOST_GOCACHE
 WORKDIR /go/src/github.com/kubermatic/kubermatic
-COPY . .
+
+COPY ./cmd/user-ssh-keys-agent ./cmd/user-ssh-keys-agent
+COPY ./cmd/http-prober/api ./cmd/http-prober/api
+COPY ./pkg ./pkg
+COPY ./go.mod ./go.mod
+COPY ./go.sum ./go.sum
+
 ENV CGO_ENABLED=0
-RUN go build ./cmd/user-ssh-keys-agent
+ENV GOCACHE=/go/src/github.com/kubermatic/kubermatic/cmd/user-ssh-keys-agent/$HOST_GOCACHE
+RUN go build -v -o ./_build/user-ssh-keys-agent ./cmd/user-ssh-keys-agent
 
 FROM docker.io/alpine:3.13
 LABEL maintainer="support@kubermatic.com"
 
-COPY --from=builder /go/src/github.com/kubermatic/kubermatic/user-ssh-keys-agent /usr/local/bin/user-ssh-keys-agent
+COPY --from=builder /go/src/github.com/kubermatic/kubermatic/_build/user-ssh-keys-agent /usr/local/bin/user-ssh-keys-agent
 ENTRYPOINT ["/usr/local/bin/user-ssh-keys-agent"]
--- a/cmd/user-ssh-keys-agent/Makefile
+++ b/cmd/user-ssh-keys-agent/Makefile
@@ -14,11 +14,24 @@
 
 DOCKER_REPO ?= "quay.io/kubermatic"
 GOOS ?= $(shell go env GOOS)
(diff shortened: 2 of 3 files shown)


cmd/user-ssh-keys-agent/Makefile
```diff
@@ -14,11 +14,25 @@
...
 
 .PHONY: build
 build:
-	GOOS=$(GOOS) CGO_ENABLED=0 go build -o ./_build/user-ssh-keys-agent
+	docker run \
+		-e GOCACHE=/go/.cache/go-build \
+ 		-e CGO_ENABLED=0 \
+		-v $(GOPATH)/src/github.com/kubermatic/kubermatic/:/go/src/github.com/kubermatic/kubermatic \
+		-v $(GOCACHE):/go/.cache/go-build \
+		-v $(GOPATH)/src/github.com/kubermatic/kubermatic/cmd/user-ssh-keys-agent/_build:/go/_build \
```
xrstf: Why is it not enough to mount the repo two lines above this one?

cmd/user-ssh-keys-agent/Dockerfile
```diff
@@ -14,13 +14,21 @@
...
 
+ARG HOST_GOCACHE
 WORKDIR /go/src/github.com/kubermatic/kubermatic
-COPY . .
+
+COPY ./cmd/user-ssh-keys-agent ./cmd/user-ssh-keys-agent
+COPY ./cmd/http-prober/api ./cmd/http-prober/api
+COPY ./pkg ./pkg
+COPY ./go.mod ./go.mod
+COPY ./go.sum ./go.sum
```
xrstf, line 24: Isn't one `COPY . .` much easier and future proof? Why do we copy everything individually?

mlavacca: Since the entire docker context is quite big, I copied only the needed stuff to decrease the requested time. Copying everything is easier and future-proof, but also heavier, and since with this PR we want to decrease the time needed to build this image, I thought it may worth it.

cmd/user-ssh-keys-agent/Makefile
mlavacca: You are right, gonna remove it.

xrstf (approved):
/approve

xrstf:
What's the status on this one?

mlavacca:
> What's the status on this one?

The PR itself should be good (if you prefer using `COPY . .` instead of the single dirs, I have no problem with it). I put it in hold status because the job `post-kubermatic-upload-gocache` uses `golang` as image, but now that job needs also docker to be properly executed. To have this merged without any problem (hopefully), we should first change that job image to `quay.io/kubermatic/build:go-1.16-node-14-kind-0.8-2`. If it works for you I can do it.

xrstf:
Sounds good (updating the job to use our mega monster image).

xrstf:
@mlavacca I think this is now good to be merged?

mlavacca:
> @mlavacca I think this is now good to be merged?

I think so
/unhold

xrstf:
@mlavacca is our current gocache primed already? In the current 1.19 e2e test, it still took 34 minutes to build just the user-ssh-key image... shouldn't this be faster already?

mlavacca:
> @mlavacca is our current gocache primed already? In the current 1.19 e2e test, it still took 34 minutes to build just the user-ssh-key image... shouldn't this be faster already?

@xrstf No, it isn't primed yet. The gocache will be uploaded by the post-submit job.

PR #7230, opened 2021-06-15
Fix verify-swagger flakiness

irozzo-1A:

This PR tentatively fixes the flakiness of `hack/verify-swagger.sh` script.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/hack/verify-swagger.sh
+++ b/hack/verify-swagger.sh
@@ -33,6 +33,11 @@ TMP_SWAGGER="${SWAGGER_FILE}.tmp"
 
 cd cmd/kubermatic-api/
 go run github.com/go-swagger/go-swagger/cmd/swagger generate spec --scan-models -o ${TMP_SWAGGER}
-diff -Naup ${SWAGGER_FILE} ${TMP_SWAGGER}
-
+# The parameters order in the generated swagger spec json file is not
+# deterministic, sort in order to avoid flake results.
+# The sorting here is applied only to first level arrays, nested arrays are not
+# sorted.
+curr="$(jq --argfile f ${SWAGGER_FILE} -n '($f | (..  | arrays) |= sort)')"
+exp="$(jq --argfile f ${TMP_SWAGGER} -n '($f | (..  | arrays) |= sort)')"
+diff -Naup <(echo "${curr}") <(echo "${exp}")
 go run github.com/go-swagger/go-swagger/cmd/swagger validate ${SWAGGER_FILE}
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -48,7 +48,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-      - image: golang:1.16.1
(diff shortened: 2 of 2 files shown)


hack/verify-swagger.sh
```diff
@@ -33,6 +33,11 @@ TMP_SWAGGER="${SWAGGER_FILE}.tmp"
...
 cd cmd/kubermatic-api/
 go run github.com/go-swagger/go-swagger/cmd/swagger generate spec --scan-models -o ${TMP_SWAGGER}
-diff -Naup ${SWAGGER_FILE} ${TMP_SWAGGER}
-
+# The parameters order in the generated swagger spec json file is not
+# deterministic, sort in order to avoid flake results.
+# The sorting here is applied only to first level arrays, nested arrays are not
+# sorted.
+curr="$(jq --argfile f ${SWAGGER_FILE} -n '($f | (..  | arrays) |= sort)')"
+exp="$(jq --argfile f ${TMP_SWAGGER} -n '($f | (..  | arrays) |= sort)')"
```
wozniakjan, line 41: https://github.com/stedolan/jq/issues/893#issuecomment-129708258
there appears to be a nifty `jq` magic with `walk` that works also recursively for nested arrays if you'd like to give it a shot
```
jq --argfile f ${FILE} -n '($f | walk(if type == "array" then sort else . end))'
```

wozniakjan (approved):
/lgtm

hack/verify-swagger.sh
irozzo-1A: hey @wozniakjan, I would keep it simple as this is enough at the moment. But thx for the hint ;-)

irozzo-1A:
@xrstf could you please override `pre-kubermatic-opa-e2e` ?

hack/verify-swagger.sh
xrstf: Would it make sense to fix the generation script to produce a neatly sorted JSON file instead of just sorting for the diff?

irozzo-1A: @xrstf You mean the generation tool? Of course, we could try to contribute, but I don't think it worths the effort.

xrstf: Yes and no. I meant can't we

```
go run github.com/go-swagger/go-swagger/cmd/swagger generate spec --scan-models -o ${TMP_SWAGGER}
cat $TMP_SWAGGER | jq magic > $TMP_SWAGGER
```

(pseudo code)

xrstf: Similar to how we format and "prettyprint" Grafana dashboards: https://github.com/kubermatic/kubermatic/blob/master/hack/lib.sh#L182

irozzo-1A: @xrstf yes we can, but I don't see any substantial gain. Note this is a hack to overcome a current limitation of the generation tool.

PR #7239, opened 2021-06-17
Allow preset filtering by domains and emails

Happy2C0de:

It improves the preset filtering but keeps things simple and backwards compatible. 
I introduced RequiredEmails as an option on the preset type. It gives users the possibility to limit certain presets to a list of domains and even on specific emails. Both in parallel. This includes the RequiredEmailDomain functionality but the variable is kept in place for backward compatibility.

The RequiredEmailDomain is appended to RequiredEmails such that both work in parallel and that the backward compatibility is assured. Added tests will take care of that.

Fixes #

I added several tests in order to verify the functionality.

**Documentation**:

```release-note
Adds support for RequiredEmails for improved presets filtering. RequiredEmailDomain is deprecated and might go away in a future release.
```

### DIFF
--- a/pkg/provider/kubernetes/preset.go
+++ b/pkg/provider/kubernetes/preset.go
@@ -195,43 +195,41 @@ func filterOutPresets(userInfo *provider.UserInfo, list *kubermaticv1.PresetList
 	var presetList []kubermaticv1.Preset
 
 	for _, preset := range list.Items {
-		requiredEmailDomain := preset.Spec.RequiredEmailDomain
-		// find preset for specific email domain (deprecated)
-		if requiredEmailDomain != "" {
-			userDomain := strings.Split(userInfo.Email, "@")
-			if len(userDomain) == 2 && strings.EqualFold(userDomain[1], requiredEmailDomain) {
-				presetList = append(presetList, preset)
-			}
-		} else {
-			// find preset based on domain and email list
-			// Example: [ "example.com", "foobar@example.com", "foo.bar@test.com"]
-			// --> all "example.com" emails and "foo.bar@test.com" will get the preset
-			requiredEmails := preset.Spec.RequiredEmails
-			if len(requiredEmails) != 0 {
-				userDomain := strings.Split(userInfo.Email, "@")
-				for _, emailItem := range requiredEmails {
-					domain := strings.Split(emailItem, "@")
+		// find preset based on domains and emails
+		// Example: [ "example.com", "foobar@example.com", "foo.bar@test.com"]
+		// --> all "example.com" emails and "foo.bar@test.com" will get the preset
+		requiredEmails := preset.Spec.RequiredEmails
 
-					// if it's a domain, we compare it against the userDomain,
-					// otherwise, it has to match the whole email
-					if len(domain) == 1 {
-						// domain provided
-						if len(userDomain) == 2 && strings.EqualFold(userDomain[1], domain[0]) {
-							presetList = append(presetList, preset)
-							break
-						}
-					} else {
-						// email provided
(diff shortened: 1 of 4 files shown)


Happy2C0de:
Question:
How easy is it to additionally implement a filtering based on projects. Somehow, the multi-tenancy is not completely working with the approach of limiting Presets based on emails. Are there any thoughts on this?
Thanks in advance.

Happy2C0de:
@zreigz I think the failures are due to some environment problems. They aren't related to this PR. Can you have a look on your side? Thanks

zreigz:
> @zreigz I think the failures are due to some environment problems. They aren't related to this PR. Can you have a look on your side? Thanks

probably our CI

Happy2C0de:
@zreigz LGTM, is there something hindering you from merging this PR?

pkg/provider/kubernetes/preset.go
```diff
@@ -195,12 +195,37 @@ func filterOutPresets(userInfo *provider.UserInfo, list *kubermaticv1.PresetList
...
-			if len(userDomain) == 2 && strings.EqualFold(userDomain[1], requiredEmailDomain) {
-				presetList = append(presetList, preset)
+			for _, emailItem := range requiredEmails {
+				domain := strings.Split(emailItem, "@")
+
+				// if it's a domain, we compare it against the userDomain,
+				// otherwise, it has to match the whole email
+				if len(domain) == 1 {
+					// domain provided
+					if len(userDomain) == 2 && strings.EqualFold(userDomain[1], domain[0]) {
```
kron4eg: Did you know that email can contain more then 1 @ symbol?

kron4eg: https://www.netmeister.org/blog/email.html

kron4eg:
In general it's MUCH better to create a preliminary issue where we could have an opportunity to discuss the need and execution of a new feature.

pkg/provider/kubernetes/preset.go
Happy2C0de: Yes, I knew but you don't allow it anyway.
See: https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/kubernetes/preset.go#L201
here, the current code is doing the exact same thing, except that it only allows domains. I don't see a reason why the current code is ok, but the new one gets blocked by this argument.
i.e.:
```
userDomain := strings.Split(userInfo.Email, "@")
if len(userDomain) == 2 && strings.EqualFold(userDomain[1], requiredEmailDomain) {
```
if a user has email `foo@bar@acme.com`. The current code is looking for `bar`!? The new code would allow you to set 
`requiredEmails: ["foo@bar@acme.com"]` and everybody would be happy.

Happy2C0de:
@kron4eg 
I agree in general. Discussions are very helpful and needed. Very good point 👍 
I opened several issues and as you see, I am also willing to participate and invest time. Nevertheless, unless some code is present nobody is interested to check things.

Important things to discuss:
- OIDC user creation is wrong, it binds on emails and not OIDC 'sub'. Emails are unique but not persistent. Identity providers allow to change emails in no time. (see 7291)
- Preset should bind to projects. (emails is fine too, if one would specify certain users within a project to be allowed to see the perset) I think, this is very important for the multi-tenancy that you strive for. (see 7260)
- IPV4-IPV6 Dual-Stack support (see 6610)
...

Nobody, gave me an answer to none of these issues even though the first two issues, are kind of relevant for your main points. OIDC integration and multi-tenancy.
I understand that all of you are busy, but you have to tell me if you don't like that externals participate.

kron4eg (approved):
/approve

Happy2C0de:
@zreigz @kron4eg LGTM now

Issue #7114, opened 2021-05-26
Introduce feature gate for Konnectivity deployment


irozzo-1A:
The goal of this task is to introduce a feature gate, to be used to enable the usage of Konnectivity instead of OpenVPN.
The feature gate should be added by default on newly created clusters, so that they make use of Konnectivity. 

https://github.com/kubermatic/kubermatic/blob/4061ff2771fd0261f1e5fce5cf06f3947ec4f0fc/pkg/features/features.go#L25-L56

PratikDeoghare:
The PR adds the feature gate but does not make it default for new installations as required by https://github.com/kubermatic/kubermatic/issues/7114

> The feature gate should be added by default on newly created clusters, so that they make use of Konnectivity.

This will be the last task after implementing Konnectivity.

almereyda:
Now that #7504 has been merged, and the feature gate seems present, @rastislavs left a last comment in https://github.com/kubermatic/kubermatic/pull/7504#issuecomment-912396775 to return here:

> The only thing I wish this would include is a per-user cluster option for enabling Konnectivity, as right now it will be enabled for whole KKP installation by the feature gate. That can be quite risky for existing KKP deployments :) But I guess this can be done in a future change as well.

Looking forward to the release of 2.18 and eager to test pre-releases, when they arrive.

moadqassem:
@PratikDeoghare Can you please review this?

rastislavs:
@PratikDeoghare I don't think we should keep this issue open until Konnectivity is finally enabled as default, as we have a whole epic for Konenctivity (#6188) and making it the default should be the last step in it. Therefore I created a separate issue under the Konenctivity epic for making it the default (#8129). Closing this one as the feature gate is already implemented.

PR #7249, opened 2021-06-21
Add feature gate to enable Konnectivity service deployment.

irozzo-1A:
This PR introduces a feature gate to be used to enable Konnectivity service deployment instead of OpenVPN.
Partly implements #7114 

Fixes #

I did not implement the default addition of the feature gate for newly created clusters, as this should be done as last step of #6188.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -53,6 +53,11 @@ const (
 
 	// UserClusterMLA if enabled MonitoringLoggingAlerting stack will be deployed with corresponding controller
 	UserClusterMLA = "UserClusterMLA"
+
+	// KonnectivityService enables the deployment of Konnectivity proxy for hte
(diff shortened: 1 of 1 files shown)


pkg/features/features.go
```diff
@@ -53,6 +53,11 @@ const (
 
 	// UserClusterMLA if enabled MonitoringLoggingAlerting stack will be deployed with corresponding controller
 	UserClusterMLA = "UserClusterMLA"
+
+	// KonnectivityService enables the deployment of Konnectivity proxy for hte
```
piotrkpc: `hte` - type-o :)

irozzo-1A: thx, fixed it ;-)

PR #7252, opened 2021-06-21
default component settings configuration

### LINKED ISSUE #5998: Configurable default control plane component settings
**User Story**
Cluster resource spec includes [componentsOverride](https://github.com/kubermatic/kubermatic/blob/979aceef489f54df2ef6db6b6e85383a78315017/pkg/crd/kubermatic/v1/cluster.go#L111) (probably the name is not appropriate), which includes settings for the control plane components (e.g. limits, number of replicas).
Today the default settings are [hardcoded](https://github.com/kubermatic/kubermatic/blob/979aceef489f54df2ef6db6b6e85383a78315017/cmd/seed-controller-manager/controllers.go#L87) and it's not possible to customize them.

**Acceptance criteria**
Having a way of specifying default settings for control plane components at master or seed level.

furkhat:
**What this PR does / why we need it**: Simple way to configure default components settings for newly created clusters. Also adds defaulting of missing fields in cluster mutation webhook and corresponding tests.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #5998

**Special notes for your reviewer**: #5756 can give some context on my previous implementation of cluster components defaulting and related discussions I had. @irozzo-1A probably knows most about it.

**Documentation**:

```release-note
* Default component settings can be configured in Seed resource.
* `nodeport-range` is deprecated. Seed's `.spec.defaultComponentSettings.apiserver.nodePortRange` should be used instead.
* `apiserver-default-replicas` is deprecated. Seed's `spec.defaultComponentSettings.apiserver.replicas` should be used instead.
* `apiserver-reconciling-disabled-by-default` is deprecated. Seed's `spec.defaultComponentSettings.apiserver.endpointReconcilingDisabled` should be used.
* `controller-manager-default-replicas` is deprecated. Seed's `spec.defaultComponentSettings.controllerManager.replicas` should be used.
* `scheduler-default-replicas` is deprecated. Seed's `spec.defaultComponentSettings.scheduler.replicas` should be used.
```


kron4eg:
@furkhat is this related to #5998?

furkhat:
@kron4eg yes this PR fixes #5998.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -426,7 +426,7 @@ type StatefulSetSettings struct {
 }
 
 type EtcdStatefulSetSettings struct {
-	ClusterSize  int                          `json:"clusterSize,omitempty"`
+	ClusterSize  *int32                       `json:"clusterSize,omitempty"`
```
kron4eg, line 429: What's the point of this change?

kron4eg (commented):
Since basically every cluster will get own settingsOverride, I fear that over time we will get Cluster objects that has different override settings (because they been created by different kubermatic versions) that could pose a great problem from data migration point of view.

WDYT @xrstf @moadqassem ?

pkg/crd/kubermatic/v1/cluster.go
furkhat: To allow distinguishing size unset and size set to zero.

kron4eg: While yes, technically this is correct, I meant: is this what we need, what's the point of having a Cluster of 0 etcds?

furkhat: I can't tell exact scenario. But sometimes users want to scale things down to zero. As an analogy - replicas count in deployments are also pointers and sometimes users scale them down to zero.

cmd/seed-controller-manager/controllers.go
```diff
@@ -93,24 +93,47 @@ func createSeedConditionUpToDateController(ctrlCtx *controllerContext) error {
...
-				Replicas: utilpointer.Int32Ptr(int32(ctrlCtx.runOptions.schedulerDefaultReplicas)),
-			},
-		},
+func defaultComponentSettings(ctrlCtx *controllerContext, seed *kubermaticv1.Seed) kubermaticv1.ComponentSettings {
+	// Copy default settings.
+	ret := seed.Spec.DefaultComponentSettings
+
+	if replicas := ctrlCtx.runOptions.apiServerDefaultReplicas; replicas != 0 {
+		if ret.Apiserver.Replicas != nil && replicas != int(*ret.Apiserver.Replicas) {
+			ctrlCtx.log.Panicf("cannot preceed due to conflicting settings, cli argument 'api-server-default-replicas'[%d] and 'seedResource.spec.defaultComponentSettings.apiserver.replicas'[%d] do not match", replicas, *ret.Apiserver.Replicas)
```
xrstf: No panic please. Please return an error all the way up the chain.

Also, "proceed" instead of "preceed".

Also we do not use the [] style of putting values in error messages, we use regular parenthesis ("... `foo.bar` (value here) do not match")

cmd/seed-controller-manager/main.go
```diff
@@ -191,7 +191,11 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		// Setup the validation admission handler for kubermatic Cluster CRDs
 		clustervalidation.NewAdmissionHandler(options.featureGates).SetupWebhookWithManager(mgr)
 		// Setup the mutation admission handler for kubermatic Cluster CRDs
-		clustermutation.NewAdmissionHandler(defaultComponentSettings(ctrlCtx)).SetupWebhookWithManager(mgr)
+		seed, err := ctrlCtx.seedGetter()
+		if err != nil {
+			log.Panicf("could not get seed resource: %v", err)
```
xrstf: `Fatalw` instead of `Panicf`.

pkg/crd/kubermatic/v1/cluster.go
xrstf: Why `int32` and not `int`?

pkg/resources/etcd/statefulset.go
```diff
@@ -394,6 +385,21 @@ func computeReplicas(data etcdStatefulSetCreatorData, set *appsv1.StatefulSet) i
 	return replicas
 }
 
+func getClusterSize(settings kubermaticv1.EtcdStatefulSetSettings) int {
+	if settings.ClusterSize == nil {
+		return kubermaticv1.DefaultEtcdClusterSize
+	}
+	if *settings.ClusterSize < kubermaticv1.MinEtcdClusterSize {
+		klog.V(2).Infof("cluster size [%d] is smaller than DefaultEtcdClusterSize [%d]. Falling back to DefaultEtcdClusterSize", *settings.ClusterSize, kubermaticv1.DefaultEtcdClusterSize)
```
xrstf: Can we get rid of these log statements? They do not add anything of value (IMHO) and are using the "wrong" logging mechanism.

xrstf:
> Since basically every cluster will get own settingsOverride, I fear that over time we will get Cluster objects that has different override settings (because they been created by different kubermatic versions) that could pose a great problem from data migration point of view.

That is true, but is it a) a problem and b) caused by this PR? If a user has a bunch of userclusters in their KKP with all different settings, that is probably just the nature of things, not a problem introduced by our configuration flexibility. Or to put it differently: Even if we didn't add this, the needs of userclusters do not suddenly vanish and the user will find other ways (crowbar.cpp) to achieve what they want.

@kron4eg can you describe further what you mean by "problem for data migration" and how this PR makes it worse than before?

cmd/seed-controller-manager/main.go
furkhat: Using panic ensures that defered calls are executed whereas using fatal does not. Do you think we don't need this?

furkhat: Though I can see that most of the places use Fatal. For the consistency I can change this.

pkg/crd/kubermatic/v1/cluster.go
furkhat: ClusterSize is the size of StatefulSet. Replicas field in StatefulSet is int32. I didn't like converting int to int32 because int is bigger.

kron4eg:
> > Since basically every cluster will get own settingsOverride, I fear that over time we will get Cluster objects that has different override settings (because they been created by different kubermatic versions) that could pose a great problem from data migration point of view.
> 
> That is true, but is it a) a problem and b) caused by this PR? If a user has a bunch of userclusters in their KKP with all different settings, that is probably just the nature of things, not a problem introduced by our configuration flexibility. Or to put it differently: Even if we didn't add this, the needs of userclusters do not suddenly vanish and the user will find other ways (crowbar.cpp) to achieve what they want.
> 
> @kron4eg can you describe further what you mean by "problem for data migration" and how this PR makes it worse than before?

I want to raise awareness of that we already defaulting bunch of `ComponentsOverride` at __creation time__ (so ComponentsOverride are persisted). Now we are changing that logic (that works only at creation time), so what about old Cluster objects that existed before deploying this code? They are not getting new defaults for ComponentsOverride.

xrstf:
> Now we are changing that logic (that works only at creation time), so what about old Cluster objects that existed before deploying this code? They are not getting new defaults for ComponentsOverride.

IMHO that is perfectly in-line with the feature. It is _supposed_ to only affect new clusters. I don't think we need to migrate anything, as this feature really isn't promising to the user that existing clusters would be affected. Am I overlooking something?

I don't think we have a good answer to "I want to change the number of apiservers for all my old user clusters!", except for "well, edit the `Cluster` objects manually."

kron4eg:
> > Now we are changing that logic (that works only at creation time), so what about old Cluster objects that existed before deploying this code? They are not getting new defaults for ComponentsOverride.
> 
> IMHO that is perfectly in-line with the feature. It is _supposed_ to only affect new clusters. I don't think we need to migrate anything, as this feature really isn't promising to the user that existing clusters would be affected. Am I overlooking something?
> 
> I don't think we have a good answer to "I want to change the number of apiservers for all my old user clusters!", except for "well, edit the `Cluster` objects manually."

Then what's the point of persisting (defaulting via webhook) all available ComponentsOverride? All we need to really persist is node-ports maybe. Everything else could be defaulted at runtime, and on-demand overriden for selected Clusters.

kron4eg:
The original issue #5998 didn't wanted to persist everything, it wanted to only have a way to customize if necessary on-demand.

xrstf:
> Then what's the point of persisting (defaulting via webhook) all available ComponentsOverride? All we need to really persist is node-ports maybe. Everything else could be defaulted at runtime, and on-demand overriden for selected Clusters.

Persisting the defaulted values should make consuming (reconciling) Cluster resources easier and less stressful, as you can assume that the Cluster object is self-contained and "in a finished state". This feature is not about "a single place to tweak the number of apiservers for all userclusters", this feature is about affecting a single, new cluster. Whether that's a useful thing is another question, but I would not outright replace the current system with a new one ^^

kron4eg:
> Persisting the defaulted values should make consuming (reconciling) Cluster resources easier and less stressful, as you can assume that the Cluster object is self-contained and "in a finished state".

But you can not assume this, since only new Clusters are defaulted.

xrstf:
/ok-to-test
/approve

I give the technical approval, @kron4eg you say whether you like the code (i.e. LGTM).

kron4eg (approved):

kron4eg:
@furkhat it needs a rebase

furkhat:
@kron4eg could you help me fix CI?

It looks like unrelated to my changes. I see similar failures in other PR's.

kron4eg:
@furkhat yeah... but I'm not sure what's wrong.

kron4eg:
@furkhat there is https://github.com/kubermatic/kubermatic/pull/7482 that is supposed to fix this flaky test.

kron4eg:
OK, now when #7482 please rebase against master again please.

kron4eg (approved):

PR #7254, opened 2021-06-21
Ensure non-colliding pprof ports when running locally

lrascao:

Ensures non-colliding pprof ports by default when running api, controller and operator locally

No

```release-note
NONE
```


xrstf (approved):
Thanks for your contribution! And sorry that it tooks this long to take a look.

/approve
/ok-to-test

PR #7266, opened 2021-06-23
add cluster template instance provider

### LINKED ISSUE #7240: Create provider for cluster template instances
The user should be able to set the number of cluster replicas created from templates

zreigz:
**What this PR does / why we need it**: add cluster template instance provider

Fixes #7240

```release-note
NONE
```


zreigz:
@lsviben @jiachengxu PTAL

pkg/provider/kubernetes/cluster_template_instance.go
```diff
@@ -0,0 +1,171 @@
...
+	instance := createClusterTemplateInstance(template, project, replicas)
+
+	err := r.privilegedClient.Create(context.Background(), instance)
+	return instance, err
+}
+
+func createClusterTemplateInstance(template *kubermaticv1.ClusterTemplate, project *kubermaticv1.Project, replicas int64) *kubermaticv1.ClusterTemplateInstance {
+	instance := &kubermaticv1.ClusterTemplateInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   GetClusterTemplateInstanceName(project.Name, template.Name),
```
lsviben, line 92: user can create multiple ClusterTemplateInstances on the same project with the same template right? Then this wont be unique anymore.

zreigz: in my opinion, should create only one. Doesn't make sense to create a different group of clusters because clusters will be the same from the same template. This solution will allow creating group of the clusters from the template which can be scale up/down

lsviben: right, so each create will act like an update and the only thing that can change is the replica count

zreigz: exactly

PR #7242, opened 2021-06-17
disable opa e2e tests temporarily

moadqassem:
**What this PR does / why we need it**:
The OPA e2e tests are super flakey that is failing almost all the time. This PR disables the OPA e2e test until the issue is fixed  
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
None
```

PR #7233, opened 2021-06-16
Stabilize opa e2e tests

lsviben:
**What this PR does / why we need it**:
OPA e2e tests got really flaky while waiting for the cluster to get healthy after enabling OPA. This PR tries to stabilize it. 

UPDATE:
It looks like this change: https://github.com/kubermatic/kubermatic/pull/7220/files#diff-bdb313299c08d80f20d961410fd103663692455b59c08f7bb1c77abeb563c781R41

changed the way we compare the old vs new version of resources and it causes endless loops on the `user-cluster-controller` because the new comparison is not semantic. This causes the Gatekeeper secret to never be deployed so the Gatekeeper Controller Pod is never Ready, which breaks the OPA test, otherwise we would not probably even notice this issue :/.  Well also it breaks the whole user-cluster deployment as all resources are not reconciled.

```
{"level":"error","time":"2021-06-17T12:17:48.030Z","caller":"resources/controller.go:231","msg":"Reconciling failed","error":"failed to reconcile Secrets in kube-system Namespace: failed to ensure Secret kube-system/usersshkeys: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
[Data: map[] != <nil map>]
{"level":"error","time":"2021-06-17T12:17:48.030Z","caller":"resources/controller.go:231","msg":"Reconciling failed","error":"failed to reconcile Secrets in kube-system Namespace: failed to ensure Secret kube-system/usersshkeys: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
[Data: map[] != <nil map>]

```

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #7278, opened 2021-06-25
Revert "disable opa e2e tests temporarily"

jiachengxu:
This PR reverts kubermatic/kubermatic#7242, since #7233 is merged, we can enable OPA tests again.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1240,8 +1240,8 @@ presubmits:
   #########################################################
 
   - name: pre-kubermatic-opa-e2e
+    run_if_changed: "(go.mod|go.sum|pkg/|.prow.yaml)"
     decorate: true
-    always_run: false
     clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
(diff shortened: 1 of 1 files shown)


lsviben:
I think as OPA tests were disabled, #7204 went in and broke the constraint handling for non-flagged constraints.

#7275 fixes that, maybe that PR should revert the opa test disabling? cc: @harshita-sharma012

lsviben:
> I think as OPA tests were disabled, #7204 went in and broke the constraint handling for non-flagged constraints.
> 
> #7275 fixes that, maybe that PR should revert the opa test disabling? cc: @harshita-sharma012

or you can add `Active: true` to the Constraint spec here: https://github.com/kubermatic/kubermatic/blob/master/pkg/test/e2e/opa/opa_test.go#L251

jiachengxu:
in this case, I think it is better to enable OPA tests in #7275
/close

Issue #7229, opened 2021-06-15
Openstack - LB name not unique enough when running multiple clusters in one tenant


lsviben:
**User Story**
The default naming schema in the Openstack Cloud Provider is not Unique enough when creating more then one cluster in one tenant.
In the OCCM we have following name generation
https://github.com/kubernetes/cloud-provider-openstack/blob/4d6ecb67ad4082a430333533b6644a86d64543ae/pkg/openstack/loadbalancer.go#L806

Which identifies a LB via ServiceName, Namespace and a Cluster name. The cluster name defaults to kubernetes.
So if you deploy to K8S cluster in one tenant and then the same services in the same namespace (f.e. Promethues in namespace monitoring), they will get the same names, and both controller try to reconcile the same OS LB.

For manual installations the OCCM has the option --cluster-name, but that is not configurable in KPP.

The fix could be to add --cluster-name option to the OCCM in https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/cloudcontroller/openstack.go

xmudrii:
Suppressed by #7330 and #7332 
/close

PR #7280, opened 2021-06-25
External CCM in vsphere

### LINKED ISSUE #5320: Use external CCM with vsphere
**User Story**
Analogous to #4610, but for vsphere

**Acceptance criteria**
- [ ] https://github.com/kubernetes-sigs/vsphere-csi-driver is used for new clusters instead of the in-tree provider

### LINKED ISSUE #7931: Use external CCM with vsphere
**User Story**
Analogous to #4610, but for vsphere

**Acceptance criteria**
- [ ] https://github.com/kubernetes-sigs/vsphere-csi-driver is used for new clusters instead of the in-tree provider

mlavacca:
External CCM in vsphere

Fixes #7931

**Documentation**:

```release-note
NONE
```


addons/csi/vsphere/csiController-rbac.yaml
```diff
@@ -0,0 +1,72 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: vsphere-csi-controller
+    namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: vsphere-csi-controller-role
+  apiGroup: rbac.authorization.k8s.io
+{{ end }}
+{{ end }}
```
moadqassem: Nit: please add a new line by the end of file

addons/csi/vsphere/csiController.yaml
```diff
@@ -0,0 +1,218 @@
...
+      targetPort: 2112
+      protocol: TCP
+    - name: syncer
+      port: 2113
+      targetPort: 2113
+      protocol: TCP
+  selector:
+    app: vsphere-csi-controller
+{{ end }}
+{{ end }}
```
moadqassem: Nit: please add a new line by the end of file

addons/csi/vsphere/csiNode-ds.yaml
```diff
@@ -0,0 +1,182 @@
...
+          hostPath:
+            path: /sys/devices
+            type: Directory
+      tolerations:
+        - effect: NoExecute
+          operator: Exists
+        - effect: NoSchedule
+          operator: Exists
+{{ end }}
+{{ end }}
```
moadqassem: Nit: please add a new line by the end of file

addons/csi/vsphere/csiNode-rbac.yaml
```diff
@@ -0,0 +1,47 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: vsphere-csi-node
+    namespace: kube-system
+roleRef:
+  kind: Role
+  name: vsphere-csi-node-role
+  apiGroup: rbac.authorization.k8s.io
+{{ end }}
+{{ end }}
```
moadqassem: Nit: please add a new line by the end of file

addons/csi/vsphere/driver.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: csi.vsphere.vmware.com
+spec:
+  attachRequired: true
+  podInfoOnMount: true
+  volumeLifecycleModes:
+    - Persistent
+    - Ephemeral
+{{ end }}
+{{ end }}
```
moadqassem: Nit: please add a new line by the end of file

addons/csi/vsphere/csiController.yaml
```diff
@@ -0,0 +1,218 @@
...
+{{ $version := "UNSUPPORTED" }}
+{{ if eq .Cluster.MajorMinorVersion "1.18" }}
+{{ $version = "v2.2.1 "}}
+{{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.19" }}
+{{ $version = "v2.2.1 "}}
+{{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.20" }}
+{{ $version = "v2.2.1 "}}
+{{ end }}
```
xmudrii, line 27: What about the latest Kubernetes version (e.g. currently 1.21)? If someone would create a cluster running 1.20 and upgrade it to 1.21, they would lose the CSI. I'm aware that the [compatibility matrix](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/docs/book/releases/v2.2.1.md#kubernetes-release) for the CSI driver says that the maximum supported version is 1.20, but we should somehow handle this case.

I think it should be safe to use CSI v2.2.1 with 1.21 as well, but we should test it.

pkg/resources/cloudcontroller/vsphere.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func getCPIContainerList(version string, data *resources.TemplateData) ([]corev1.Container, error) {
+	controllerManagerImage := fmt.Sprintf("%s/cloud-provider-vsphere/cpi/release/manager:v%s", data.ImageRegistry(resources.RegistryGCR), version)
+
+	containerList := []corev1.Container{
+		{
+			Name:  vsphereCPIControllerManager,
+			Image: controllerManagerImage,
+			Args: []string{
```
xmudrii: We should also set the `--cluster-name` flag to avoid problems similar to #7229.

We can do it in two ways:

1) Set it unconditionally
2) Set it in the same way as for OpenStack (by respecting the `ccmClusterName` cluster feature): https://github.com/kubermatic/kubermatic/blob/43d0e7bbaa8c445ca9dd5931a3466e40ebf199a4/pkg/resources/cloudcontroller/openstack.go#L138-L140

I think we should go with the second approach. Currently, the vSphere CCM will be deployed only for the new user clusters, so the first approach should work. However, we'll also migrate existing clusters later, which don't have the cluster name set, so the second approach can help us avoid potential problems that we encountered on OpenStack.

Note that if you go with the second approach, you also need to update this function, so the cluster feature is enabled by default for new user clusters: https://github.com/kubermatic/kubermatic/blob/43d0e7bbaa8c445ca9dd5931a3466e40ebf199a4/pkg/resources/cloudcontroller/util.go#L55-L64

pkg/resources/cloudcontroller/vsphere.go
```diff
@@ -0,0 +1,162 @@
...
+					},
+				},
+			})
+
+			return dep, nil
+		}
+	}
+}
+
+func getCPIContainerList(version string, data *resources.TemplateData) ([]corev1.Container, error) {
```
xmudrii: Nit: I don't see that this function can return an error, so I don't think we need to return it.

addons/csi/vsphere/csiController.yaml
mlavacca: I'll test it and if it works I'll set CSI v2.2.1 for 1.21 as well

pkg/resources/cloudcontroller/vsphere.go
mlavacca: Thanks, I refactored this function a bunch of times and probably I forgot to remove it :smile:

addons/csi/vsphere/csiController.yaml
mlavacca: I made extensive testing and k8s 1.21 is actually well supported, gonna add it.

pkg/resources/data.go
```diff
@@ -688,12 +691,23 @@ func GetCSIMigrationFeatureGates(cluster *kubermaticv1.Cluster) []string {
...
 		if cluster.Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider] {
-			featureFlags = append(featureFlags, "CSIMigration=true", "CSIMigrationOpenStack=true", "ExpandCSIVolumes=true")
+			featureFlags = append(featureFlags, "CSIMigration=true", "ExpandCSIVolumes=true")
+		}
+		if cluster.Spec.Cloud.Openstack != nil {
+			featureFlags = append(featureFlags, "CSIMigrationOpenStack=true")
+		}
+		if cluster.Spec.Cloud.VSphere != nil {
+			featureFlags = append(featureFlags, "CSIMigrationvSphere=true")
 		}
```
xmudrii, line 701: Do we also need to enable the CSI migration here?

https://github.com/kubermatic/kubermatic/blob/5182eb92494a74f20d7e3941179bd63df99d537d/addons/csi/vsphere/csiController.yaml#L193

If yes, please take care of that in a follow-up.

xmudrii (commented):
Before we close #5320, we should make sure to:

* Update machine-controller to a version that includes changes to the vSphere cloud-config
* Add E2E tests
* Make sure users can provide the needed VM Storage Policy, and create a StorageClass based on it
* Document how to set everything up

All those tasks can/should be done in a follow-up, but please create follow-up issues.

Issue #7112, opened 2021-05-26
OpenStack cluster stuck in deletion with failed network cleanup


wozniakjan:
**User Story**
An OpenStack cluster created through kkp dev has been stuck in terminating for almost 6 days - https://dev.kubermatic.io/projects/7fztn48c97/clusters/lv2jljfmwf. There is a reoccurring error wrt network cleanup.

```
$ k logs -n kubermatic kubermatic-seed-controller-manager-86bf8ddfdb-j87ms | grep lv2jljfmwf | tail -n1
{"level":"error","time":"2021-05-26T12:23:55.849Z","logger":"kubermatic_cloud_controller","caller":"cloud/cloud_controller.go:139","msg":"Reconciling failed","worker-name":"","request":"/lv2jljfmwf","cluster":"lv2jljfmwf","error":"failed cloud provider cleanup: failed to delete network 'kubernetes-lv2jljfmwf': failed to get network 'kubernetes-lv2jljfmwf' by name: no network named 'kubernetes-lv2jljfmwf' with external=false found","errorCauses":[{"error":"failed cloud provider cleanup: failed to delete network 'kubernetes-lv2jljfmwf': failed to get network 'kubernetes-lv2jljfmwf' by name: no network named 'kubernetes-lv2jljfmwf' with external=false found"}]}
```

<details><summary>$ k describe cluster lv2jljfmwf</summary>
<p>

```yaml
Name:         lv2jljfmwf
Namespace:    
Labels:       project-id=7fztn48c97
Annotations:  <none>
Address:
  Admin Token:    *** redacted ***
  External Name:  lv2jljfmwf.europe-west3-c.dev.kubermatic.io
  Internal URL:   apiserver-external.cluster-lv2jljfmwf.svc.cluster.local.
  Ip:             192.168.30.10
  Port:           6443
  URL:            https://lv2jljfmwf.europe-west3-c.dev.kubermatic.io:6443
API Version:      kubermatic.k8s.io/v1
Kind:             Cluster
Metadata:
  Creation Timestamp:             2021-05-20T13:32:15Z
  Deletion Grace Period Seconds:  0
  Deletion Timestamp:             2021-05-20T14:04:46Z
  Finalizers:
    kubermatic.io/cleanup-credentials-secrets
    kubermatic.io/cleanup-openstack-network-v2
    kubermatic.io/cleanup-openstack-router-subnet-link-v2
    kubermatic.io/cleanup-openstack-router-v2
[...]

PR #7284, opened 2021-06-28
Fixes network not found issue

### LINKED ISSUE #7241: Openstack Cluster stuck in deletion when network not found
**User Story**
Openstack cluster stuck in deletion with error `failed cloud provider cleanup: failed to delete network 'kubernetes-xlkxth9pbp': failed to get network 'kubernetes-xlkxth9pbp' by name: no network named 'kubernetes-xlkxth9pbp' with external=false found`

When trying to delete network and is already deleted/missing then deletion should progress instead of throwing error
So as far as I could see it, we need to:
- wrap error here https://github.com/kubermatic/kubermatic/blob/6b5514d04b1f4eeba369fe5af414717be66bdfe9/pkg/provider/cloud/openstack/helper.go#L280 with %w, so that here https://github.com/kubermatic/kubermatic/blob/6b5514d04b1f4eeba369fe5af414717be66bdfe9/pkg/provider/cloud/openstack/provider.go#L360
we can unwrap the error and check if the error value is `ErrNotFound` using
`if errors.Is(err, ErrNotFound) {
    // something wasn't found
}`
and move on to removing finalizer

**Fixes**  #7112

**Acceptance criteria**
Not return error when a resource, (network in this case) is not found i.e, `failed to get network by name`, in order to complete cluster deletion

imharshita:

This PR fixes the issue of `failed cloud provider cleanup: failed to delete network 'kubernetes-xxx': failed to get network 'kubernetes-xxx' by name: no network named 'kubernetes-xlkxth9pbp' with external=false found` by wrapping  error and unwrapping it to verify if the error is of kind `not found`

Fixes #7241 #7112

Steps performed to test the changes:
- Create a openstack cluster and delete the network
- Check if the `seed-controller-manager` throws error
- Check if openstack cluster is deleted

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/provider/cloud/openstack/helper.go
+++ b/pkg/provider/cloud/openstack/helper.go
@@ -52,11 +52,11 @@ const (
 func getSecurityGroups(netClient *gophercloud.ServiceClient, opts ossecuritygroups.ListOpts) ([]ossecuritygroups.SecGroup, error) {
 	page, err := ossecuritygroups.List(netClient, opts).AllPages()
 	if err != nil {
-		return nil, fmt.Errorf("failed to list security groups: %v", err)
+		return nil, fmt.Errorf("failed to list security groups: %w", err)
 	}
 	secGroups, err := ossecuritygroups.ExtractGroups(page)
 	if err != nil {
-		return nil, fmt.Errorf("failed to extract security groups: %v", err)
+		return nil, fmt.Errorf("failed to extract security groups: %w", err)
 	}
 	return secGroups, nil
 }
--- a/pkg/provider/cloud/openstack/helper.go
+++ b/pkg/provider/cloud/openstack/helper.go
@@ -98,7 +98,7 @@ func getNetworkByName(netClient *gophercloud.ServiceClient, name string, isExter
 	case 1:
 		return candidates[0], nil
 	case 0:
-		return nil, fmt.Errorf("no network named '%s' with external=%v found", name, isExternal)
+		return nil, fmt.Errorf("network named '%s' with external=%v not found", name, isExternal)
 	default:
 		return nil, fmt.Errorf("found %d networks for name '%s' (external=%v), expected exactly one", len(candidates), name, isExternal)
 	}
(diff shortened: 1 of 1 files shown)


pkg/provider/cloud/openstack/helper.go
```diff
@@ -468,7 +468,7 @@ func getSubnetForNetwork(netClient *gophercloud.ServiceClient, networkIDOrName s
 }
 
 func isNotFoundErr(err error) bool {
-	if _, ok := err.(gophercloud.ErrDefault404); ok || strings.Contains(err.Error(), "not found") {
+	if _, ok := err.(gophercloud.ErrDefault404); ok || strings.Contains(errors.Unwrap(err).Error(), "not found") {
```
irozzo-1A: I'm not fond of the `string.Contains`, but this is not something you introduced. On the other hand, the type check should do the unwrapping and not the string match. You can use 

```suggestion
	var errNotFound gophercloud.ErrDefault404
	if errors.As(err, &errNotFound); ok || strings.Contains(errors.Error(), "not found") {
```

imharshita: sure

pkg/provider/cloud/openstack/helper.go
```diff
@@ -135,7 +135,7 @@ func validateSecurityGroupsExist(netClient *gophercloud.ServiceClient, securityG
 func deleteSecurityGroup(netClient *gophercloud.ServiceClient, sgName string) error {
 	results, err := getSecurityGroups(netClient, ossecuritygroups.ListOpts{Name: sgName})
 	if err != nil {
-		return fmt.Errorf("failed to get security group: %v", err)
+		return fmt.Errorf("failed to get security group: %w", err)
```
irozzo-1A, line 138: You should also use `%w` in `getSecurityGroups`

imharshita: done

PR #7287, opened 2021-06-28
Default to DockerRegistry only if provided repo has less than 3 parts

xmudrii:

This PR changes the way how we default the registry for Kubermatic API, etcd-launcher, and dnat-controller images.

Currently, we're splitting the provided image by `/` and checking the number of parts:

* if there are 3 parts, use the first element of the slice as the registry
* otherwise, default to `docker.io`

That's causing problems if the provided image has more than 3 parts, for example: `registry/part2/part3/part4/kubermatic-ee`. In that case, KKP will try to use `docker.io/registry/part2/part3/part4/kubermatic-ee` as an image for running UCCM (which is incorrect).

I think that we should default to `docker.io` only in case the number of parts (when split by `/`) is less than 3, otherwise, use the first element of the slice as the registry.

I'm not sure is this going to break the backward compatibility for some users. We don't have any unit tests to confirm if this function is going to work as expected. I can work on some unit tests in the meanwhile.

```release-note
Kubermatic API, etcd-launcher, and dnat-controller images are defaulted to the docker.io registry only if the provided custom image has less than 3 parts
```

### DIFF
--- a/pkg/resources/data.go
+++ b/pkg/resources/data.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/distribution/reference"
+
 	providerconfig "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
 	httpproberapi "k8c.io/kubermatic/v2/cmd/http-prober/api"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
@@ -331,16 +333,7 @@ func (d *TemplateData) EtcdDiskSize() resource.Quantity {
 }
 
 func (d *TemplateData) EtcdLauncherImage() string {
-	imageSplit := strings.Split(d.etcdLauncherImage, "/")
-	var registry, imageWithoutRegistry string
-	if len(imageSplit) < 3 {
-		registry = RegistryDocker
-		imageWithoutRegistry = strings.Join(imageSplit, "/")
-	} else {
-		registry = imageSplit[0]
-		imageWithoutRegistry = strings.Join(imageSplit[1:], "/")
-	}
-	return d.ImageRegistry(registry) + "/" + imageWithoutRegistry
+	return d.parseImage(d.etcdLauncherImage)
 }
 
 func (d *TemplateData) EtcdLauncherTag() string {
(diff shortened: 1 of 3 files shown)


pkg/resources/data_test.go
```diff
@@ -109,3 +109,231 @@ func TestGetCSIMigrationFeatureGates(t *testing.T) {
...
+			},
+			wantAPIImage: "custom-registry.kubermatic.io/kubermatic/kubermatic",
+		},
+		{
+			name: "custom image with 1 part",
+			templateData: &TemplateData{
+				kubermaticImage: "kubermatic",
+			},
+			wantAPIImage: "docker.io/kubermatic",
+		},
```
xmudrii: This case doesn't make much sense if `overwriteRegistry` is not used (since `docker.io/kubermatic` will never work). We should either handle it in the function or just ignore it.

pkg/resources/data.go
```diff
@@ -534,7 +534,7 @@ func (d *TemplateData) NodeLocalDNSCacheEnabled() bool {
 func (d *TemplateData) KubermaticAPIImage() string {
 	apiImageSplit := strings.Split(d.kubermaticImage, "/")
 	var registry, imageWithoutRegistry string
-	if len(apiImageSplit) != 3 {
+	if len(apiImageSplit) < 3 {
 		registry = RegistryDocker
```
xmudrii: I'm not sure should we actually default the registry ourselves. Usually, that's done by the container runtime, which may have a default registry that's not `docker.io`.

kron4eg (approved):

pkg/resources/data.go
```diff
@@ -532,33 +525,30 @@ func (d *TemplateData) NodeLocalDNSCacheEnabled() bool {
...
-		registry = RegistryDocker
-		imageWithoutRegistry = strings.Join(apiImageSplit, "/")
-	} else {
-		registry = apiImageSplit[0]
-		imageWithoutRegistry = strings.Join(apiImageSplit[1:], "/")
+	return d.parseImage(d.kubermaticImage)
+}
+
+func (d *TemplateData) parseImage(image string) string {
+	named, _ := reference.ParseNormalizedNamed(image)
```
xmudrii, line 532: I don't like that this error is ignored. This can lead to NPEs, but I'd expect that to be very rare since we validate the input here: 

https://github.com/kubermatic/kubermatic/blob/e2fd4a8a11f54e0e6eed4c9c5c15adb4e6c743ea/pkg/controller/operator/common/defaults.go#L543-L561

I came up with two ideas, but I don't like them:

* Change functions to return an error. We would have to change the interface and I don't like that.
* Log an error and return an empty string. I don't like it because such failure is silent.

@kron4eg do you have any opinions?

kron4eg (approved):

pkg/resources/data.go
kron4eg: well, since it's passed the validation, we don't have to double check errors again

kron4eg: I mean, it's OK

PR #7289, opened 2021-06-28
[release/v2.17] added option for Azure Load Balancer SKU in KKP API

lsviben:

This is an cherry-pick of #7208

```release-note
Added option to set the Load Balancer SKU when creating Azure clusters
```


moadqassem (commented):
/approve
/lgtm

PR #7295, opened 2021-06-29
Add aws spot instance market options

### LINKED ISSUE #7250: Add market options configs in KKP
**User Story**
Include the aws spot instance market options configurations in KKP.

**Acceptance Criteria**
- Import the latest machine controller version which supports aws spot instances
- Include all the configuration of the market options fields of the aws spot instance in KKP

moadqassem:
Support AWS spot instances market options in KKP 
Fixes #7250 

**Documentation**:

```release-note
Support spot instance market options
```

### DIFF
--- a/cmd/kubermatic-api/swagger.json
+++ b/cmd/kubermatic-api/swagger.json
@@ -14859,6 +14859,21 @@
           "type": "boolean",
           "x-go-name": "IsSpotInstance"
         },
+        "spotInstanceInterruptionBehavior": {
+          "description": "SpotInstanceInterruptionBehavior Sets the interruption behavior for the spot instance when capacity is no longer\navailable at the price you specified, if there is no capacity, or if a constraint cannot be met. Charges for EBS\nvolume storage apply when an instance is stopped.",
+          "type": "string",
+          "x-go-name": "SpotInstanceInterruptionBehavior"
+        },
+        "spotInstanceMaxPrice": {
+          "description": "SpotInstanceMaxPrice The maximum price you are willing to pay per instance hour. Your instance runs when\nyour maximum price is greater than the Spot Price.",
+          "type": "string",
+          "x-go-name": "SpotInstanceMaxPrice"
+        },
+        "spotInstancePersistentRequest": {
+          "description": "SpotInstancePersistentRequest Ensures that your request will be submitted every time your Spot Instance is terminated.\nFor information about Persistent requests and other customization options",
+          "type": "boolean",
+          "x-go-name": "SpotInstancePersistentRequest"
+        },
         "subnetID": {
           "description": "The VPC subnet to which the node shall be connected.",
           "type": "string",
--- a/cmd/kubermatic-api/swagger.json
+++ b/cmd/kubermatic-api/swagger.json
@@ -11486,13 +11486,6 @@
         "parameters": [
           {
             "type": "string",
-            "x-go-name": "Architecture",
-            "description": "architecture query parameter. Supports: arm64 and x64 types.",
-            "name": "architecture",
-            "in": "query"
-          },
-          {
-            "type": "string",
             "x-go-name": "ProjectID",
             "name": "project_id",
             "in": "path",
@@ -11504,6 +11497,13 @@
             "name": "cluster_id",
             "in": "path",
             "required": true
+          },
+          {
+            "type": "string",
+            "x-go-name": "Architecture",
+            "description": "architecture query parameter. Supports: arm64 and x64 types.",
+            "name": "architecture",
+            "in": "query"
           }
         ],
         "responses": {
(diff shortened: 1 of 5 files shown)


pkg/api/v1/types.go
```diff
@@ -1482,6 +1482,16 @@ type AWSNodeSpec struct {
 	AssignPublicIP *bool `json:"assignPublicIP"`
 	// IsSpotInstance indicates whether the created machine is an aws ec2 spot instance or on-demand ec2 instance.
 	IsSpotInstance *bool `json:"isSpotInstance"`
+	// SpotInstanceMaxPrice The maximum price you are willing to pay per instance hour. Your instance runs when
+	// your maximum price is greater than the Spot Price.
```
xmudrii: Tiny nit:

```suggestion
	// SpotInstanceMaxPrice is the maximum price you are willing to pay per instance hour. Your instance runs when
	// your maximum price is greater than the Spot Price.
```

pkg/api/v1/types.go
```diff
@@ -1482,6 +1482,16 @@ type AWSNodeSpec struct {
...
+	SpotInstanceMaxPrice *string `json:"spotInstanceMaxPrice"`
+	// SpotInstancePersistentRequest Ensures that your request will be submitted every time your Spot Instance is terminated.
+	// For information about Persistent requests and other customization options
```
xmudrii: I think that you're missing the part of the comment here. Do you want to point to the AWS docs here?

pkg/api/v1/types.go
```diff
@@ -1482,6 +1482,16 @@ type AWSNodeSpec struct {
...
+	SpotInstancePersistentRequest *bool `json:"spotInstancePersistentRequest"`
+	// SpotInstanceInterruptionBehavior Sets the interruption behavior for the spot instance when capacity is no longer
```
xmudrii: Tiny nit:

```suggestion
	// SpotInstanceInterruptionBehavior sets the interruption behavior for the spot instance when capacity is no longer
```

pkg/machine/convert.go
```diff
@@ -278,3 +284,17 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 
 	return cloudSpec, nil
 }
+
+func extractSpotInstanceConfigs(config *aws.RawConfig) (*string, *string, *bool) {
+	if config.IsSpotInstance != nil && *config.IsSpotInstance {
+		if config.SpotInstanceConfig != nil {
```
xmudrii: Tiny nit: I think those two `if`s can be combined into a single one.

pkg/machine/convert.go
```diff
@@ -278,3 +284,17 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
...
+			return &config.SpotInstanceConfig.MaxPrice.Value,
+				&config.SpotInstanceConfig.InterruptionBehavior.Value,
+				&config.SpotInstanceConfig.PersistentRequest.Value
```
xmudrii: Are we sure that this is not going to throw NPEs if some property is `nil`? For example, if `MaxPrice` is `nil`, accessing `MaxPrice.Value` will throw the NPE.

xmudrii (commented):
Mostly some nits, otherwise LGTM.

cmd/kubermatic-api/swagger.json
```diff
@@ -11290,6 +11290,13 @@
         ],
         "operationId": "listAWSSizesNoCredentialsV2",
         "parameters": [
+          {
+            "type": "string",
+            "x-go-name": "Architecture",
+            "description": "architecture query parameter. Supports: arm64 and x64 types.",
```
lsviben: NIT: try running swagger generation again, sometimes this gets messed up for some reason :(. But 90% of times it resolves as was before.

pkg/api/v1/types.go
moadqassem: Nah, I will just omit it :D

pkg/machine/convert.go
moadqassem: That's the problem with all pointers tbh as the config is a pointer as well and it could lead to NPE as well. One thing that we can do is, adding validation for all configs

moadqassem: How? the `MaxPrice.Value` is string and the `MaxPrice` is of a type `ConfigVarString` which is a struct not a pointer. `SpotInstanceConfig` is a pointer however there is already a nil check before it

pkg/machine/convert.go
```diff
@@ -278,3 +284,17 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 
 	return cloudSpec, nil
 }
+
+func extractSpotInstanceConfigs(config *aws.RawConfig) (*string, *string, *bool) {
+	if config.IsSpotInstance != nil &&
+		*config.IsSpotInstance &&
+		config.SpotInstanceConfig != nil {
+		if config.SpotInstanceConfig != nil {
```
xmudrii: Nit: you already check `config.SpotInstanceConfig != nil` above.

pkg/machine/convert.go
xmudrii: Oh, my bad. For some reason, I thought those three values are pointers. Ignore my comment.

pkg/machine/convert.go
moadqassem: 😂😬

xmudrii (approved):
/lgtm
/approve

PR #7300, opened 2021-06-29
fix resource requests and limits setting for seed components

multi-io:
The defaultResourceList() function erroneously always sets only the default values, even if the input list already contained overridden values. This leads to resource requests and limits not being properly transferred from the KubermaticConfiguration to the respective target objects.

For the master components, this problem is masked by https://github.com/kubermatic/kubermatic/blob/38cf6e1ab738d145fe53bb7e2d2ae4decb04bcfb/pkg/controller/operator/master/reconciler.go#L101, which reads the in-cluster KubermaticConfiguration back into "config", but for seed components, most prominently the seed controller manager, you can easily reproduce that it always receives the default resource requests and limits no matter what you set in KubermaticConfiguration.spec.seedController.resources.

Fixes #

**Documentation**:

```release-note
Fixed seed components resource requests and limits setting via KubermaticConfiguration
```

### DIFF
--- a/pkg/controller/operator/common/defaults.go
+++ b/pkg/controller/operator/common/defaults.go
@@ -578,15 +578,17 @@ func defaultResources(settings *corev1.ResourceRequirements, defaults corev1.Res
 }
 
 func defaultResourceList(list *corev1.ResourceList, defaults corev1.ResourceList, key string, logger *zap.SugaredLogger) error {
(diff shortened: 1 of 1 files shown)


mate4st:
Thanks for your PR @multi-io. 
I am closing this in favor of https://github.com/kubermatic/kubermatic/pull/7302

PR #7305, opened 2021-06-29
Whitelisted registry controller

### LINKED ISSUE #7236: WhitelistedRegistry controller

lsviben:
Adds whitelisted registry controller

Fixes #7236 

```release-note
Added whitelisted registries feature for EE, which allows users to set which image registries are whitelisted so only workloads from those registries can be deployed on user clusters. 
```

### DIFF
--- a/pkg/ee/whitelisted-registry-controller/reconcile.go
+++ b/pkg/ee/whitelisted-registry-controller/reconcile.go
@@ -50,15 +50,15 @@ const (
 	WhitelistedRegistryField  = "whitelisted_registry"
 )
 
-type reconciler struct {
+type Reconciler struct {
 	log          *zap.SugaredLogger
 	recorder     record.EventRecorder
 	masterClient ctrlruntimeclient.Client
 	namespace    string
 }
 
-func NewReconciler(log *zap.SugaredLogger, recorder record.EventRecorder, masterClient ctrlruntimeclient.Client, namespace string) *reconciler {
-	return &reconciler{
+func NewReconciler(log *zap.SugaredLogger, recorder record.EventRecorder, masterClient ctrlruntimeclient.Client, namespace string) *Reconciler {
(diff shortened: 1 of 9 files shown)


pkg/ee/whitelisted-registry-controller/reconcile.go
```diff
@@ -0,0 +1,229 @@
...
+	return reconcile.Result{}, err
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, whitelistedRegistry *kubermaticv1.WhitelistedRegistry) error {
+	if whitelistedRegistry.DeletionTimestamp != nil {
+		if !kuberneteshelper.HasFinalizer(whitelistedRegistry, kubermaticapiv1.WhitelistedRegistryCleanupFinalizer) {
+			return nil
+		}
+
+		// Ensure Constraint with registry data
```
jiachengxu, line 98: Do we also want to delete the constraint template for this WhitelistedRegistry if it is deleted?

lsviben: good question. 

For now i decided to leave it as by itself(without an active Constraint), the ConstraintTemplate doesnt do anything, but we could consider removing it as well. WDYT?

jiachengxu: Both look good to me, I was asking because I noticed this is created in the reconcile loop but we didn't remove it when WhiteListedRegistery is removed, so I was wondering if we missed anything in the controller logic :)
but if the ContraintTemplate doesn't do anything, I am fine with leaving it

mfranczy (approved):
/approve
/lgtm

fabiand:
Conscious language, what about allowlist? :)

lsviben:
> Conscious language, what about allowlist? :)

Good point, totally forgot about that, we will change the naming

PR #7330, opened 2021-07-02
Add ClusterFeatureCCMClusterName feature for OpenStack clusters

### LINKED ISSUE #7332: Implement a feature flag to enable the --cluster-name flag
Implement a feature flag on the Cluster object that would set the `--cluster-name` flag on the external CCM deployment for the supported providers. This should solve the problem described in #7331. The reason for implementing this as a feature flag is described in #7330.

xmudrii:

This PR introduces a cluster feature called `ccmClusterName`, which is used to apply the `--cluster-name` flag on the external CCMs. Currently, this flag is only applied on the OpenStack CCM. In the future, it will be applied for the other cloud providers that support the external CCM as well.

The `--cluster-name` flag is often used for naming cloud resources, such as Load Balancers. If it's not provided, the CCM would default it to `kubernetes`.

Not providing the `cluster-name` flag can cause conflicts as described in issue #7229.

It's not possible to easily change this flag for the existing user clusters, as it can cause the external CCM to lose track of the existing cloud resources. Therefore, we're introducing a feature flag with the following behavior:

* if it's enabled, add the `--cluster-name` flag on the external CCM deployment
* enable the flag by default for newly created clusters
* allow users to explicitly enable this flag for the existing clusters
  * in this case, it's up to the user to clean up any leftover resources
  * in the future, we might implement a migration mechanism that would automatically clean up leftover resources

Fixes #7332

```release-note
Add ClusterFeatureCCMClusterName feature for OpenStack clusters. This feature adds the --cluster-name flag to the OpenStack external CCM deployment. The feature gate is enabled by default for newly created clusters. Enabling this feature gate for existing clusters can cause the external CCM to lose the track of the existing cloud resources, so it's up to the users to clean up any leftover resources.
```

/assign @kron4eg

### DIFF
--- a/pkg/crd/kubermatic/v1/cluster.go
+++ b/pkg/crd/kubermatic/v1/cluster.go
@@ -198,6 +198,10 @@ const (
 	// not be set if its not supported.
 	ClusterFeatureExternalCloudProvider = "externalCloudProvider"
 
+	// ClusterFeatureCCMClusterName sets the cluster-name flag on the external CCM deployment.
+	// The cluster-name flag is often used for naming cloud resources, such as load balancers.
+	ClusterFeatureCCMClusterName = "ccmClusterName"
+
 	// ClusterFeatureRancherIntegration enables the rancher server integration feature.
 	// It will deploy a Rancher Server Managegment plane on the seed cluster and import the user cluster into it.
 	ClusterFeatureRancherIntegration = "rancherIntegration"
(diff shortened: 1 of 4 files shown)


kron4eg (approved):
/approve

kron4eg:
@xmudrii looks like we also need someone from sig-api

PR #7353, opened 2021-07-06
CCM migration status

mlavacca:
A new field about external CCM migration has been added to the cluster status API.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/handler/common/cluster.go
```diff
@@ -1028,15 +1028,16 @@ func convertInternalCCMStatusToExternal(cluster *kubermaticv1.Cluster, datacente
...
 
 	if _, condition := helper.GetClusterCondition(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted); condition != nil {
-		status.MigrationCompleted = pointer.BoolPtr(helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue))
-	} else {
-		status.MigrationCompleted = nil
+		if helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue) {
+			status.MigrationCompleted = pointer.BoolPtr(true)
+		} else {
+			status.MigrationInProgress = pointer.BoolPtr(true)
+		}
```
xmudrii: This might produce an incorrect result between starting the migration and the controller. For example:

* The ExternalCCM is not enabled, but it's supported. We'll set `MigrationNeeded` to `true`. Other fields will be `nil`.
* We enable the CCM/CSI migration. At this point, the feature is enabled, so we'll set `ExternalCCM` to `true`. We will not set `MigrationNeeded`.
* The cluster condition is not set because the controller is still not enabled. This will cause other fields to be `nil`, so we will have only `ExternalCCM` set to `true`. This is not expected and will report a wrong status in the UI.

I think that this can be solved by checking the annotations:

* `ExternalCCM` will be `true` if the cluster feature is enabled
* `MigrationNeeded` will be `true` if the cluster feature is disabled, but the provider supports the external CCM
  * It'll become `false` as soon as we start the migration
  * If it's `true`, we don't need to set the other field
 * `MigrationInProgress` will be `true` if the cluster feature is enabled, and there are the CCM/CSI migration annotations, and the cluster condition is `nil` or `false`. Otherwise, it'll be `false`.
 * `MigrationComplete` will be `true` if the cluster condition is true
   * I'm not sure if this field is really needed and useful, since I don't think that we need to show this information in the UI. Anyways, let's keep it for now and see later.

pkg/api/v1/types.go
```diff
@@ -1051,15 +1051,22 @@ type ClusterStatus struct {
 	CCM ExternalCCMStatus `json:"ccm"`
 }
 
+type ExternalCCMMigrationStatus string
+
+var (
+	ExternalCCMMigrationNotNeeded   ExternalCCMMigrationStatus = "notNeeded"
+	ExternalCCMMigrationSupported   ExternalCCMMigrationStatus = "supported"
+	ExternalCCMMigrationUnsupported ExternalCCMMigrationStatus = "unsupported"
+	ExternalCCMMigrationInProgress  ExternalCCMMigrationStatus = "inProgress"
```
xmudrii: Tiny nit: usually status strings start with an uppercase letter. I don't think it's important, but I'll leave it up to @maciaszczykm to decide should this be changed or not.

pkg/api/v1/types.go
```diff
@@ -1051,15 +1051,22 @@ type ClusterStatus struct {
...
+)
+
 // ExternalCCMStatus represents the external CCM status
 // swagger:model ExternalCCMStatus
 type ExternalCCMStatus struct {
 	// ExternalCCM reflects the ClusterFeatureExternalCloudProvider in the cluster .spec.features
 	ExternalCCM bool `json:"externalCCM"`
```
xmudrii: Do we need this field any longer? Previously, it was used to determine if the migration is needed, but now that's covered with the new status field.

pkg/handler/common/cluster.go
```diff
@@ -1047,17 +1047,30 @@ func getSSHKey(ctx context.Context, userInfoGetter provider.UserInfoGetter, sshK
...
-		status.MigrationCompleted = pointer.BoolPtr(helper.ClusterConditionHasStatus(cluster, kubermaticv1.ClusterConditionCSIKubeletMigrationCompleted, corev1.ConditionTrue))
-	} else {
-		status.MigrationCompleted = nil
+	if !externalCCMEnabled && externalCCMSupported {
+		status.ExternalCCMMigration = apiv1.ExternalCCMMigrationSupported
+	}
+
+	if !externalCCMEnabled && !externalCCMSupported {
+		status.ExternalCCMMigration = apiv1.ExternalCCMMigrationUnsupported
 	}
```
xmudrii: Tiny nit: would it look better if we would replace those 3 `if`s with `switch`?

pkg/handler/common/cluster_test.go
```diff
@@ -0,0 +1,218 @@
...
+		},
+		{
+			Name: "scenario 4: CCM migration unsupported",
+			Datacenter: &kubermaticv1.Datacenter{
+				Spec: kubermaticv1.DatacenterSpec{Openstack: &kubermaticv1.DatacenterSpecOpenstack{}},
+			},
+			Cluster: &kubermaticv1.Cluster{
+				Spec: kubermaticv1.ClusterSpec{
+					Cloud: kubermaticv1.CloudSpec{
+						AWS: &kubermaticv1.AWSCloudSpec{},
```
xmudrii: Tiny nit: I would replace this with a Fake or Kubeadm provider since AWS will support the CCM/CSI migration in the future.

pkg/api/v1/types.go
maciaszczykm: I agree, let's align it with other places.

xmudrii (approved):
I think it should be good to go now. @maciaszczykm PTAL again.

/lgtm
/approve

PR #7367, opened 2021-07-12
extend owner and editor roles for patch verb

### LINKED ISSUE #7360: Fix privileges for PATCH endpoint for cluster template instances
During PATCH of cluster template instance, the endpoint return 'FORBIDDEN' error. Looks like the RBAC controller missing some verbs

zreigz:
**What this PR does / why we need it**: extend owner and editor roles for patch verb. For most API providers, we patch API objects and call the UPDATE client method (the historical reason, we used goclient instead of controller runtime client). It's possible to directly call the PATCH method instead. It requires an extension of our RBAC roles for owners and editors with the "patch" verb.

Fixes #7360

```release-note
NONE
```


zreigz:
@irozzo-1A PTAL, I had to fix the unit tests

PR #7387, opened 2021-07-15
fix DeepEqual method for different types

### LINKED ISSUE #7386: Seed Reconciliation issues in weekly-2021-27
**User Story**
When  I try to upgrade the existing kkp installation to weekly-2021-27, my seed does not get reconciled and kubermatic-operator pod spews out continuous warnings (with log-debug=true). 

Note the first line of log below.. It talks about ENV being empty array vs not being present. This difference seems to be ignorable. But after this line, no further log comes for like a second and then it dumps the error that it timed out waiting for reconciliation.

```
{"level":"debug","time":"2021-07-14T11:46:09.528Z","caller":"reconciling/compare.go:60","msg":"Object differs from generated one","type":"*v1.Deployment","namespace":"kubermatic","name":"kubermatic-seed-controller-manager","diff":["Spec.Template.Spec.Containers.slice[0].Env: [] != <nil slice>"]}
{"level":"error","time":"2021-07-14T11:46:10.109Z","logger":"kubermatic-seed-operator","caller":"seed/reconciler.go:74","msg":"failed to reconcile","request":"kubermatic/vsphere-seed","error":"failed to reconcile Kubermatic Deployments: failed to ensure Deployment kubermatic/kubermatic-seed-controller-manager: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
{"level":"error","time":"2021-07-14T11:46:10.110Z","logger":"controller-runtime.manager.controller.kubermatic-seed-operator","caller":"controller/controller.go:252","msg":"Reconciler error","name":"vsphere-seed","namespace":"kubermatic","error":"failed to reconcile Kubermatic Deployments: failed to ensure Deployment kubermatic/kubermatic-seed-controller-manager: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
{"level":"debug","time":"2021-07-14T11:46:10.109Z","logger":"controller-runtime.manager.events","caller":"record/event.go:308","msg":"Warning","object":{"kind":"KubermaticConfiguration","namespace":"kubermatic","name":"kubermatic","uid":"bf809a0c-c492-4271-b30e-5ca236231e2e","apiVersion":"operator.kubermatic.io/v1alpha1","resourceVersion":"52819763"},"reason":"SeedReconcilingError","message":"vsphere-seed: failed to reconcile Kubermatic Deployments: failed to ensure Deployment kubermatic/kubermatic-seed-controller-manager: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
```

If I rollback to v2.17.2 - these warnings disappear and seed gets reconciled properly.

As Lovro pointed out.. it could be due to.. [this change](https://github.com/kubermatic/kubermatic/pull/7220/files#diff-bdb313299c08d80f20d961410fd103663692455b59c08f7bb1c77abeb563c781R41)

The reconciling stops after it fails here, so the resources that are next in line never get reconciled.

**Acceptance criteria**
The compare should ignore this difference in ENV and should reconcile the seed properly.

zreigz:
**What this PR does / why we need it**: fix DeepEqual method for different types

Fixes #7386

```release-note
NONE
```

### DIFF
--- a/pkg/resources/reconciling/compare.go
+++ b/pkg/resources/reconciling/compare.go
@@ -19,15 +19,13 @@ package reconciling
 import (
 	"encoding/json"
 	"fmt"
-	"reflect"
 
 	"github.com/go-test/deep"
 	"go.uber.org/zap"
 
-	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	k8cequality "k8c.io/kubermatic/v2/pkg/crd/equality"
 	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
 
-	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
@@ -40,17 +38,8 @@ func init() {
 
 // DeepEqual compares both objects for equality
 func DeepEqual(a, b metav1.Object) bool {
-	// The ClusterTemplate consist with ClusterSpec which has nested Semver type.
-	// The semverlib.Version doesn't implement Equality function. Using equality.Semantic.DeepEqual for this object cause
-	// an informative panic.
-	if _, ok := a.(*kubermaticv1.ClusterTemplate); ok {
-		if reflect.DeepEqual(a, b) {
-			return true
-		}
-	} else {
-		if equality.Semantic.DeepEqual(a, b) {
-			return true
-		}
+	if k8cequality.Semantic.DeepEqual(a, b) {
+		return true
 	}
 
 	// For some reason unstructured objects returned from the api have types for their fields
--- a/pkg/resources/reconciling/compare.go
+++ b/pkg/resources/reconciling/compare.go
@@ -24,8 +24,10 @@ import (
 	"github.com/go-test/deep"
 	"go.uber.org/zap"
 
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


zreigz:
@irozzo-1A approve needed from you

pkg/resources/reconciling/compare.go
```diff
@@ -38,8 +40,17 @@ func init() {
 
 // DeepEqual compares both objects for equality
 func DeepEqual(a, b metav1.Object) bool {
-	if reflect.DeepEqual(a, b) {
-		return true
+	// The ClusterTemplate consist with ClusterSpec which has nested Semver type.
+	// The semverlib.Version doesn't implement Equality function. Using equality.Semantic.DeepEqual for this object cause
+	// an informative panic.
+	if _, ok := a.(*kubermaticv1.ClusterTemplate); ok {
```
irozzo-1A: @zreigz BTW a more elegant solution would be to create our `Equalities` instance providing a comparison function for `Semver` type. See the following for reference:

https://github.com/kubernetes/apimachinery/blob/b63a0c883fbfc313249150449400788e5589ef23/pkg/api/equality/semantic.go#L29-L49

zreigz: I will try, thanks

lsviben: I tested it and you can add this to the init function of `compare.go`
Or we can add this to the Add method of the ClusterTemplate controller, but then it only works for the controller, and in future maybe something else will need it. 

```
func init() {
	// Kubernetes Objects can be deeper than the default 10 levels.
	deep.MaxDepth = 20
	deep.LogErrors = true

	// Add comparison function for semver.Version
	equality.Semantic.AddFunc(func(a, b *semver.Version) bool {
		return a.Compare(b) == 0
	})
}

```

zreigz: I had to add other objects to make it run our controller, seems to work fine
Thanks, @irozzo-1A

zreigz:
@irozzo-1A @lsviben PTAL

lsviben:
Looks like this solution is easily extendable if we need it. 
/lgtm

xrstf:
I cancelled the test, just to test if restarting it will generate proper URLs. Sorry for the spam.

/retest

PR #7394, opened 2021-07-19
Improve default visibility

mate4st:

Fixes #https://github.com/kubermatic/docs/issues/705

**Documentation**:

```release-note
Add "kubermatic-installer print" command to improve defaulting visibility.
```

### DIFF
--- a/cmd/kubermatic-installer/cmd_print.go
+++ b/cmd/kubermatic-installer/cmd_print.go
@@ -17,95 +17,41 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
-	"os"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
-	"go.uber.org/zap"
 
-	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
-	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
-	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
-	yamlutil "k8c.io/kubermatic/v2/pkg/util/yaml"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-var (
-	seedFlag = cli.BoolFlag{
-		Name:  "seed",
-		Usage: "Print Seed defaults",
-	}
-
-	kubermaticConfigFlag = cli.BoolFlag{
-		Name:  "config",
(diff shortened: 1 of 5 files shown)


kron4eg:
@WeirdMachine dumping result of yaml.Marshal removes all the comments structs had, so it's not very useful. In the docs directory we have generated files with all the comments and defaults, so we need to use them as a source to those new commands. To embed those files during the build time please use https://pkg.go.dev/embed and https://pkg.go.dev/io/fs.

mate4st:
Yes I like the idea, but to use embed we would need to upgrade the go version to 1.16. So should we leave it for now or introduce some other third party library?

kron4eg:
We already use go 1.16, no “upgrade” is required.

mate4st:
go.mod is still 1.14 or am I missing something here?

kron4eg:
Yes, but we build using 1.16 golabg image. 1.14 been working for us so far since we didn’t used any 1.16 compiler features. Bottom line is: simply bump go.mod version.

cmd/kubermatic-installer/cmd_print.go
```diff
@@ -0,0 +1,59 @@
...
+	return handleErrors(logger, setupLogger(logger, func(ctx *cli.Context) error {
+
+		arg := strings.ToLower(ctx.Args().First())
+
+		switch arg {
+		case "":
+			fallthrough
+		case "config":
+			fallthrough
+		case "kubermaticconfiguration":
```
kron4eg:
```suggestion
		case "", "config", "kubermaticconfiguration":
```

cmd/kubermatic-installer/cmd_print.go
```diff
@@ -0,0 +1,59 @@
...
+		Usage:       "Prints example configuration manifest",
+		UsageText:   "kubermatic-installer print <resource>",
+		Description: "Print an example configuration manifest with defaults for the given resource.\n   Supported resources are \"seed\" and \"kubermaticconfiguration\".",
+		ArgsUsage:   "Supported resources are \"seed\" and \"kubermaticconfiguration\".",
+		Action:      PrintAction(logger),
+	}
+}
+
+func PrintAction(logger *logrus.Logger) cli.ActionFunc {
+	return handleErrors(logger, setupLogger(logger, func(ctx *cli.Context) error {
```
kron4eg: isn't logger is not used? 

```suggestion
	return func(ctx *cli.Context) error {
```

mate4st:
/hold
wait for https://github.com/kubermatic/kubermatic/issues/7400

kron4eg (approved):
/approve

PR #7396, opened 2021-07-19
Apiserver network policies

irozzo-1A:
This PR introduces the deployment of NetworkPolicies to constraint the egress traffic of KAS.

Fixes #

It applies to newly created clusters, the feature can be activated on old clusters by adding the feature gate to the `Cluster` resource. The same feature gate can be used to disable reconciliation of the `NetworkPolicies`, which have to be manually deleted in case of need.

**Documentation**:

```release-note
Add Apiserver network policy.
```

### DIFF
--- a/pkg/resources/apiserver/networkpolicy.go
+++ b/pkg/resources/apiserver/networkpolicy.go
@@ -151,3 +151,67 @@ func OpenVPNServerAllowCreator(c *kubermaticv1.Cluster) reconciling.NamedNetwork
 		}
 	}
 }
+
+func MachineControllerWebhookCreator(c *kubermaticv1.Cluster) reconciling.NamedNetworkPolicyCreatorGetter {
+	return func() (string, reconciling.NetworkPolicyCreator) {
+		return "machine-controller-webhook-allow", func(np *networkingv1.NetworkPolicy) (*networkingv1.NetworkPolicy, error) {
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PolicyTypes: []networkingv1.PolicyType{
+					networkingv1.PolicyTypeEgress,
+				},
+				PodSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						resources.AppLabelKey: name,
+					},
+				},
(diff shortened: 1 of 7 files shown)


PratikDeoghare (approved):

pkg/resources/apiserver/networkpolicy.go
```diff
@@ -0,0 +1,217 @@
...
+				},
+				Egress: []networkingv1.NetworkPolicyEgressRule{},
+			}
+
+			return np, nil
+		}
+	}
+}
+
+// EtcdAllowCreator returns a func to create/update the apiserver
```
rastislavs, line 50: nit: the description seems to be a copy-paste, does not match with what the func does (also for `DNSAllowCreator` and `OpenVPNServerAllowCreator`)

pkg/resources/apiserver/networkpolicy.go
```diff
@@ -0,0 +1,217 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DenyAllPolicyCreator returns a func to create/update the apiserver
+// deny all egress policy.
+func DenyAllPolicyCreator() reconciling.NamedNetworkPolicyCreatorGetter {
```
rastislavs, line 30: Is this empty egress rule really necessary to deny what is not explicitly allowed?

> Pods become isolated by having a NetworkPolicy that selects them. Once there is any NetworkPolicy in a namespace selecting a particular pod, that pod will reject any connections that are not allowed by any NetworkPolicy.

irozzo-1A: Indeed

pkg/resources/apiserver/networkpolicy.go
irozzo-1A: Thx! will fix this with a follow-up PR

PR #7410, opened 2021-07-22
Add strictArp IPVS setting for kube-proxy to Cluster API

### LINKED ISSUE #7309: Improve support for MetalLB in L2 mode (Kubeproxy IPVS strict ARP)
**User Story**
[MetalLB](https://metallb.universe.tf/) in L2 mode relies on ARP: for each virtual service IP (LoadBalancer IP), one leader-elected node in the cluster should respond to an ARP request for a LoadBalancer IP. 

In case that IPVS mode is used for kube-proxy (which is the default in KKP), this concept is a bit broken. Each node in the cluster has an `kube-ipvs0` interface with all LoadBalancer service IPs applied on it. This causes that by default, every node (not only the leader-elected one) replies to an ARP request for a LoadBalancer IP.

This should not cause any big issues for short-lived requests, as any node in the cluster is capable of forwarding the packet to the target service backends. But in case when persistent connections are needed, this may cause connections drops, whenever an upstream router issues a new ARP request for a LoadBalancer IP - as it will use the first answer, the target node may change randomly.

This issue can be reproduced more frequently if the upstream router does not extend the timeout of entries in the ARP cache with each received frame, but re-issues ARP request periodically - which is often the case for firewalls, to prevent ARP spoofing.

**Acceptance criteria**
Allow to enable `strictARP` mode in the `KubeProxyConfiguration` in KKP, which should prevent the issue.

mate4st:
Fixes #7309

**Documentation**:

```release-note
Add strictARP setting for kube-proxy IPVS mode and make it configurable vie Cluster CRD or API.
```


pkg/addon/template.go
```diff
@@ -80,6 +81,11 @@ func NewTemplateData(
 		return nil, fmt.Errorf("failed to determine cloud provider name: %v", err)
 	}
 
+	// value is set by the mutating AdmissionHandler
+	if cluster.Spec.ClusterNetwork.IPVS.StrictArp == nil {
```
kron4eg: I'd say that nil pointer should be defaulted to `true`. Otherwise existing clusters that didn't got modified just yet will start error-ing. And please use some central point (like resources package) to get this default value in both places, here and in the webhook.

pkg/addon/template_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"go.uber.org/zap"
+	"k8s.io/utils/pointer"
```
kron4eg: Please run https://github.com/xrstf/gimps over this PR.

kron4eg (changes requested):

kron4eg (approved):
/approve

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -465,6 +465,9 @@ type ClusterNetworkingConfig struct {
 	// Defaults to ipvs.
 	ProxyMode string `json:"proxyMode"`
 
+	// IPVS defines kube-proxy ipvs configuration options
+	IPVS IPVSConfiguration `json:"ipvs"`
```
maciaszczykm: One thing I would consider here is adding `omitempty` param. If the proxy mode is different than `ipvs` then it would be nice to not return this at all. WDYT?

maciaszczykm:
@WeirdMachine I've added one minor remark. Besides that looks good from the API perspective.

pkg/crd/kubermatic/v1/cluster.go
kron4eg: omitempty doesn't work on structs, only on pointers to structs

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -136,6 +136,12 @@ func (h *AdmissionHandler) applyDefaults(c *kubermaticv1.Cluster) {
 		}
 	}
 
+	if c.Spec.ClusterNetwork.IPVS != nil {
```
kron4eg, line 139: what if it is == nil?

mate4st: It gets defaulted in the kube-proxy template.

kron4eg:
Seems like this one needs swagger spec regenerated.

xrstf:
/retest

That failure was me.

PR #7413, opened 2021-07-23
[WIP] [OpenStack] Move CSI Cinder plugin out of addons to reconciler managed resource

### LINKED ISSUE #6406: Cluster stuck in deletion Openstack when there are PersistentVolumes and no worker nodes

wozniakjan:
**Incomplete implementation**

During cluster deletion, we have a rare race condition. When no worker nodes allow `csi-cinder-controllerplugin` pods to run while there are still some PVs, the finalizer `external-attacher/cinder-csi-openstack-org` doesn't get deleted and this makes cluster cleanup stuck. We can split the installation of CSI Cinder plugin between seed and user cluster with the controller on the seed. That way even during cleanup, controller will always be available to properly clean the PVs.

CSI cinder plugin was initially installed through addons, this doesn't allow splitting the installation and effort to do this had many difficult obstacles https://github.com/kubermatic/kubermatic/pull/7037. Instead I would like to try to use the `reconciler` pattern from `pkg/resources`.

Fixes https://github.com/kubermatic/kubermatic/issues/6406

Custom resource reconcilers in KKP expect client-go resources. CSI Cinder plugin gets released in form of a helm chart which can be templated to raw kubernetes yaml manifests. This PR leverages https://github.com/wozniakjan/reverse-kube-resource, tool that generates go code from kubernetes yaml manifests. This should simplify adjustments in the following releases of CSI Cinder plugin.

**Documentation**:

```release-note
The CSI Cinder plugin is moving from addons to reconciler managed resource.
```


go.mod
```diff
@@ -87,9 +88,9 @@ require (
 	gopkg.in/square/go-jose.v2 v2.5.1
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
-	k8s.io/api v0.20.2
+	k8s.io/api v0.21.1
```
kron4eg: please lower version to original, by lowering it in github.com/wozniakjan/reverse-kube-resource

hack/gen-pkg-resources.sh
```diff
@@ -0,0 +1,34 @@
...
+tmp_dir=$(mktemp -d)
+trap '{ rm -rf -- "$tmp_dir"; }' EXIT
+
+(
+    cd "$tmp_dir"
+    git clone --branch=v1.21.1 --depth 1 https://github.com/kubernetes/cloud-provider-openstack
+    helm template cinder-csi --namespace=kube-system ./cloud-provider-openstack/charts/cinder-csi-plugin > "$tmp_dir/cinder-csi-resources.yaml"
+)
+
+go run ./vendor/github.com/wozniakjan/reverse-kube-resource/main.go \
```
kron4eg:
```suggestion
go run github.com/wozniakjan/reverse-kube-resource \
```

There is no need for vendor, `go run` can work with modules unless there is something super specific in reverse-kube-resource (which should be fixed in that case).

wozniakjan: that's a good idea. I can then remove it from `hack/tools.go` at which point it won't pollute the `go.mod`. Thanks!

kron4eg: oh no! Please keep it in hack/tools.go, that way we will manage actual version that will be run.

wozniakjan: ok :)

wozniakjan:
/remove-lifecycle stale
just currently has low priority because https://github.com/kubermatic/kubermatic/issues/6406 has not occurred since https://github.com/kubermatic/kubermatic/issues/6759 and https://github.com/kubermatic/kubermatic/pull/7189 merged

wozniakjan:
https://github.com/kubermatic/kubermatic/issues/6406#issuecomment-1066564070

PR #7423, opened 2021-07-27
Storage policy in vSphere

### LINKED ISSUE #7419: Storage policy API for vSphere
**User Story**
In #7931 the external cloud controller has been added for vSphere. The storage provisioning needs the VM Storage policy to be set. This field can be added in the datacenter API and should be settable by the user.

**Acceptance criteria**

mlavacca:
The external CSI driver in vSphere may need the storage policy to be set in the storage-class to correctly provision the volumes. This PR aims at introducing a new field in the sphere CloudSpec to allow customizing such field. Besides, its value should be defaulted by a field at the datacenter level. The storage-policy is then used in the default storage-class created.

This PR fixes also vsphere  post-job and periodic e2e tests.

Fixes #7419, #7442

**Documentation**:

```release-note
DefaultStoragePolicy field has been added to the vSphere datacenter spec and storagePolicy to the vsphere CloudSpec at cluster level
```


addons/default-storage-class/storage-class.yaml
```diff
@@ -44,6 +44,21 @@ volumeBindingMode: WaitForFirstConsumer
...
+  annotations:
+    storageclass.beta.kubernetes.io/is-default-class: "true"
+  labels:
+    kubernetes.io/cluster-service: "true"
+  name: vsphere-csi
+provisioner: csi.vsphere.vmware.com
+{{ if .Cluster.VsphereStoragePolicy }}
+parameters:
+  storagepolicyname: {{ .Cluster.VsphereStoragePolicy }}
+{{ end }}
```
xmudrii: Does it make sense to create the StorageClass if `.Cluster.VsphereStoragePolicy` is not provided? How that StorageClass would work?

pkg/addon/template.go
```diff
@@ -198,6 +206,11 @@ type CNIPlugin struct {
 	Version string
 }
 
+type VsphereConfiguration struct {
+	// StoragePolicy is the storage policy to use for vsphere csi addon
+	StoragePolicy string
+}
```
xmudrii: Where do we use this struct?

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -401,14 +401,20 @@ func (r *Reconciler) ensureClusterRoleBindings(ctx context.Context, c *kubermati
...
 		openvpn.ServerClientConfigsConfigMapCreator(data),
 		dns.ConfigMapCreator(data),
 		apiserver.AuditConfigMapCreator(),
 		apiserver.AdmissionControlCreator(data),
 		apiserver.CABundleCreator(data),
 	}
+
+	if data.Cluster().Spec.Cloud.VSphere != nil {
+		creators = append(creators, cloudconfig.ConfigmapVsphereCSICreator(data))
+	}
```
xmudrii, line 437: Why we don't create a Secret instead of a ConfigMap? The original cloud-config is also a Secret if I'm not mistaken.

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	types2 "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
```
xmudrii: * Please make sure the import order is correct, i.e. this import should go along with other Kubermatic imports.
* Please give it a more descriptive import alias, such as `mctypes` or something like that.

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -106,6 +106,7 @@ func Add(
 		versions:          versions,
 		caBundle:          caBundle,
 		userClusterMLA:    userClusterMLA,
+		cloudProvider:     types2.CloudProvider(cloudProviderName),
```
xmudrii: What is the reason for using a machine-controller type/function here? How is this going to behave with a provider that's supported by KKP, but not MC, for example, BringYourOwnCloud?

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -70,16 +71,24 @@ func (r *reconciler) reconcile(ctx context.Context) error {
...
 	data := reconcileData{
-		caCert:        caCert,
-		openVPNCACert: openVPNCACert,
-		userSSHKeys:   userSSHKeys,
-		cloudConfig:   cloudConfig,
+		caCert:                caCert,
+		openVPNCACert:         openVPNCACert,
+		userSSHKeys:           userSSHKeys,
+		cloudConfig:           cloudConfig,
+		vsphereCSICloudConfig: vsphereCSICloudConfig,
```
xmudrii: Nit: maybe we could give this field a generic name, such as `csiCloudConfig`.

addons/default-storage-class/storage-class.yaml
mlavacca: It turned out that specifying the `storagePolicy` is not mandatory. If not set, the default one is used.

pkg/addon/template.go
mlavacca: It's garbage left from a previous implementation that I forgot to delete. Will remove it.

pkg/controller/seed-controller-manager/kubernetes/resources.go
mlavacca: The cloud-config resides in two distinct data structures: a cm in the seed cluster (created by the seed-controller-manager) and a secret in the user-cluster (created by the user-cluster-controller-manager) created from the aforementioned cm.

mlavacca: here is the secret creation in the user-cluster: https://github.com/kubermatic/kubermatic/blob/6c29856f862cb6319575034bec202fb3694d235d/pkg/controller/user-cluster-controller-manager/resources/reconciler.go#L613

pkg/controller/user-cluster-controller-manager/resources/controller.go
xrstf: Try https://github.com/xrstf/gimps ^^

pkg/controller/user-cluster-controller-manager/resources/controller.go
mlavacca: The reason behind it is that we already had all the needed providers in that pkg. Btw, having BringYourOwn and casting it to CloudProvider is fine, the string is simply cast to `CloudProvider` type. This cast was needed to compare the current CloudProvider with the ones defined in the mc pkg. If you think this is a kind of unclean approach, I can re-declare those constants in a pkg in KKP and use them.

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -154,7 +154,7 @@ type DatacenterSpec struct {
 
 	//nolint:staticcheck
 	//lint:ignore SA5008 omitgenyaml is used by the example-yaml-generator
-	Fake *DatacenterSpecFake `json:"fake,omitempty,omitgenyaml"`
+	Fake *DatacenterSpecFake `json:"fake,omitempty,omitempty"`
```
xmudrii: `omitgenyaml` is a valid tag, please see the comment above.

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -112,7 +112,7 @@ type KubermaticUIConfiguration struct {
 	// ---
 	//nolint:staticcheck
 	//lint:ignore SA5008 omitgenyaml is used by the example-yaml-generator
-	DockerTag string `json:"dockerTag,omitempty,omitgenyaml"`
+	DockerTag string `json:"dockerTag,omitempty,omitempty"`
```
xmudrii: `omitgenyaml` is a valid tag, please see the comment above.

pkg/resources/cloudconfig/configmap.go
```diff
@@ -32,6 +32,7 @@ import (
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	gcp "k8c.io/kubermatic/v2/pkg/provider/cloud/gcp"
 	"k8c.io/kubermatic/v2/pkg/resources"
+	vsphere2 "k8c.io/kubermatic/v2/pkg/resources/cloudconfig/vsphere"
```
xmudrii: Please avoid import aliases such as `<name>{1..n}`. In this case, something like `vspherecloudconfig` would be a better import alias.

docs/zz_generated.kubermaticConfiguration.yaml
```diff
@@ -264,6 +264,9 @@ spec:
       }
     # DockerRepository is the repository containing the Kubermatic dashboard image.
     dockerRepository: quay.io/kubermatic/dashboard
+    # DockerTag is used to overwrite the dashboard Docker image tag and is only for development
```
xmudrii: This file needs to be regenerated once the `omitgenyaml` tags are back in place.

docs/zz_generated.seed.yaml
```diff
@@ -81,6 +81,7 @@ spec:
         # EnforcePodSecurityPolicy enforces pod security policy plugin on every clusters within the DC,
         # ignoring cluster-specific settings
         enforcePodSecurityPolicy: false
+        fake: null
```
xmudrii: This file needs to be regenerated once the `omitgenyaml` tags are back in place.

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -765,6 +778,8 @@ type reconcileData struct {
 	mlaGatewayCACert *resources.ECDSAKeyPair
 	userSSHKeys      map[string][]byte
 	cloudConfig      []byte
+	// CSICloudConfig is currently used only by vSphere, whose needs it to properly configure the external CSI driver
+	CSICloudConfig []byte
```
xmudrii: Tiny nit: is there a reason why this field is exported, while others aren't? This struct is not exported anyways, but I think it would be nice to be consistent.

pkg/resources/usercluster/deployment.go
```diff
@@ -307,3 +313,46 @@ func getLabelsArgValue(cluster *kubermaticv1.Cluster) (string, error) {
...
+		return provider.PacketCloudProvider, nil
+	}
+	if cluster.Spec.Cloud.BringYourOwn != nil {
+		return provider.BringYourOwnCloudProvider, nil
+	}
+	if cluster.Spec.Cloud.Fake != nil {
+		return provider.FakeCloudProvider, nil
+	}
+	return "", fmt.Errorf("provider unknown")
+}
```
xmudrii: Why are we now using this function instead of `data.GetKubernetesCloudProviderName()`? The original function has been handling some additional cases including CCM/CSI migration.

pkg/provider/cloud/vsphere/provider_test.go
```diff
@@ -28,6 +28,10 @@ import (
 	"k8c.io/kubermatic/v2/pkg/resources"
 )
 
+const (
+	fakeStoragePolicy = "fake_storage_policy"
+)
+
 func TestGetCredentialsForCluster(t *testing.T) {
```
xmudrii, line 35: Do we have any validation for the specified StoragePolicy? For example, I see that we validate and ensure that the specified Datastore actually exists. Should we do that for the StoragePolicy as well?

pkg/resources/cloudconfig/configmap.go
```diff
@@ -266,6 +297,7 @@ func getVsphereCloudConfig(
 			Datacenter:       dc.Spec.VSphere.Datacenter,
 			DefaultDatastore: datastore,
 			WorkingDir:       cluster.Name,
+			ClusterID:        dc.Spec.VSphere.Cluster,
```
xmudrii, line 300: Shouldn't this be set only in the case when we're generating the CSI config? Otherwise, if we set it for cloudConfig as well, it will cause Kubelet to fail.

pkg/resources/usercluster/deployment.go
mlavacca: Because the function you mentioned doesn't properly return the provider in which the cluster is running but instead returns "external" as cloud provider when the `externalCloudProvider` is used. This function, instead simply returns the name of the provider, and gives it to the `user-cluster-controller-manager` as a flag. Besides, That flag appeared to be not used by the `user-cluster-controller-manager` itself before I introduced these changes.

pkg/resources/cloudconfig/configmap.go
mlavacca: The data structure is the same, and the cluster-ID is assigned in both cases (simple `cloudConfig` and `CSICloudConfig`), but they are converted to a string by using two different templates: the machine controller one (that ignores the `clusterID`) for the original `cloudConfig` and [this](https://github.com/kubermatic/kubermatic/blob/b62408de16a3f0743f050481bce033b8522bdbf7/pkg/resources/cloudconfig/vsphere/csi-configmap.go#L30) one for the CSI config, that takes into account also the `clusterID`.

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
mlavacca: You are right, gonna make it unexported.

docs/zz_generated.seed.yaml
mlavacca: All the docs changes have been made by a `./hack/update-docs.sh` run after a rebase to master.

pkg/provider/cloud/vsphere/provider_test.go
mlavacca: I'll check it out

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -223,6 +223,7 @@ type reconciler struct {
 	versions          kubermatic.Versions
 	caBundle          resources.CABundle
 	userClusterMLA    UserClusterMLA
+	cloudProvider     mctypes.CloudProvider
```
xmudrii: I think it would be better to use CloudProvider consts defined in KKP, rather than MC types. The type and consts are defined here: https://github.com/kubermatic/kubermatic/blob/67e94dee12b1b471f6ad2563d64304e6088e9c32/pkg/crd/kubermatic/v1/preset.go#L27-L69

(thanks to @xrstf for finding this :slightly_smiling_face:)

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -71,16 +72,24 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to get userSSHKeys: %v", err)
 	}
-	cloudConfig, err := r.cloudConfig(ctx)
+	cloudConfig, err := r.cloudConfig(ctx, resources.CloudConfigConfigMapName)
 	if err != nil {
 		return fmt.Errorf("failed to get cloudConfig: %v", err)
 	}
+	var vsphereCSICloudConfig []byte
```
xmudrii: Tiny nit: we don't need the `vsphere` prefix, similar to the variable in the `reconcileData` struct.

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -603,7 +612,11 @@ func (r *reconciler) reconcileConfigMaps(ctx context.Context, data reconcileData
 func (r *reconciler) reconcileSecrets(ctx context.Context, data reconcileData) error {
 	creators := []reconciling.NamedSecretCreatorGetter{
 		openvpn.ClientCertificate(data.openVPNCACert),
-		cloudcontroller.CloudConfig(data.cloudConfig),
+		cloudcontroller.CloudConfig(data.cloudConfig, resources.CloudConfigSecretName),
+	}
+
+	if r.cloudProvider == mctypes.CloudProviderVsphere {
```
xmudrii: Tiny nit: it might be better to check is `data.csiCloudConfig != nil`. In that case, it might be easier to extend it for other providers in the future (if needed).

pkg/resources/resources.go
```diff
@@ -524,6 +524,8 @@ const (
 
 	KubevirtKubeConfig = "kubeConfig"
 
+	VsphereCSICloudConfigConfigMapName = "cloud-config-csi"
+	VsphereCSICloudConfigSecretName    = "cloud-config-csi"
```
xmudrii: Tiny nit: `Vsphere` prefix is not needed at all.

pkg/resources/cloudcontroller/vsphere.go
```diff
@@ -76,7 +76,6 @@ func vsphereDeploymentCreator(data *resources.TemplateData) reconciling.NamedDep
 				return nil, err
 			}
 			dep.Spec.Template.Spec.AutomountServiceAccountToken = pointer.BoolPtr(false)
-			dep.Spec.Template.Spec.HostNetwork = true
```
xmudrii, line 80: Why did we remove this? For reference, the upstream manifest has `hostNetwork: true`: 

https://github.com/kubernetes/cloud-provider-vsphere/blob/951214ce27ed7dc439a812c4ab27c16947f18544/releases/v1.21/vsphere-cloud-controller-manager.yaml#L249

pkg/resources/usercluster/deployment.go
xmudrii: Okay, this makes sense, but I'd propose that we add this function to the same place as the `data.GetKubernetesCloudProviderName()` function, so it can be easily reused. Also, make sure to update the interface here: https://github.com/kubermatic/kubermatic/blob/ecc36450f80ab81a6d6e658dd7cba94b116233db/pkg/resources/usercluster/deployment.go#L57-L72

pkg/resources/cloudcontroller/vsphere.go
mlavacca: I removed it because the CCM works with the user cluster but is deployed in the seed one, and having the CCM in the host network in the seed cluster doesn't make any sense. We do the same thing for [openstack](https://github.com/kubernetes/cloud-provider-openstack/blob/295f57c9f0af38d50b901db2414de2114d18208f/charts/openstack-cloud-controller-manager/templates/daemonset.yaml#L84). Besides if we set the pod in `hostNetwork`, and we configure the DNS as it is required to be, the entire dns configuration of the node messes up (this is the reason behind the periodic vsphere test failure).

pkg/controller/user-cluster-controller-manager/resources/controller.go
mlavacca: Cool, I looked around but I didn't find those consts, gonna use them instead.

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	v1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
```
xmudrii: This import should go down with other `k8c.io` imports. Also, avoid the `v1` import alias. We're using `kubermaticv1` as the import alias in other places.

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -71,16 +71,24 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to get userSSHKeys: %v", err)
 	}
-	cloudConfig, err := r.cloudConfig(ctx)
+	cloudConfig, err := r.cloudConfig(ctx, resources.CloudConfigConfigMapName)
 	if err != nil {
 		return fmt.Errorf("failed to get cloudConfig: %v", err)
 	}
+	var CSICloudConfig []byte
```
xmudrii, line 78: Tiny nit: style change

```suggestion
	var csiCloudConfig []byte
```

xmudrii (commented):
Two tiny nits, otherwise LGTM.

xmudrii (approved):
Great job, @mlavacca! :100: 

/lgtm
/approve

xmudrii:
@xrstf @lsviben Can you PTAL at this PR? It's ready from the SIG ClusterMgmt side.

lsviben:
Looks good from sig-api side

/lgtm
/approve

mlavacca:
@xmudrii @xrstf now that the tests run with K8S 1.22, they are failing because of this: https://github.com/kubermatic/kubermatic/blob/8745bec5f8eced425779fa66a0d925d03df26ae4/addons/csi/vsphere/csiNode-ds.yaml#L18-L27
The problem is that the CSI driver 2.2.1 does not officially support version 1.22 (https://vsphere-csi-driver.sigs.k8s.io/compatiblity_matrix.html). Besides, a new version has been just released yesterday, that officially introduces support for 1.21. I would update the latest version to 2.3.0 in any case, but what about the 1.22 support? Should we use the latest also for it even if not officially supported?

xmudrii:
> Should we use the latest also for it even if not officially supported?

@mlavacca No, we shouldn't. The vSphere CSI still uses `apiextensions/v1beta1` API which is removed from 1.22, so the CSI driver wouldn't work at all. The Kubernetes changelog also says:

> If CSIMigrationvSphere feature gate is enabled, user should not upgrade to Kubernetes v1.22. vSphere CSI Driver does not support Kubernetes v1.22 yet because it uses v1beta1 CRD APIs. Support for v1.22 will be added at a later release. Check the following document for supported Kubernetes releases for a given vSphere CSI Driver version.

This is going to be fixed by https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/1188, but it might take some time until the PR is merged and a new release is out.

We discussed this earlier, and I think it would be best if we would block creating Kubernetes 1.22 clusters on vSphere. For now, I think it should be safe and okay to downgrade vSphere tests to 1.21.

.prow.yaml
```diff
@@ -1029,7 +1029,7 @@ presubmits:
         - name: KUBERMATIC_EDITION
           value: ee
         - name: VERSIONS_TO_TEST
-          value: "v1.22.1"
+          value: "v1.21.0"
```
xmudrii: Let's use the latest patch release -- v1.21.4.

mlavacca: The latest 1.21 patch release supported by KKP in creation seems to be 1.21.3: https://github.com/kubermatic/kubermatic/blob/e56b93784e25f9eaa6c820e4a8f95fd235a8efaf/pkg/controller/operator/common/defaults.go#L207-L210
When using 1.21.4 I got a `400 - Bad Request` from our API server

xmudrii (approved):

.prow.yaml
xmudrii: My bad. 1.21.3 is okay as well.

xrstf (approved):
/approve

PR #7425, opened 2021-07-27
Fix prometheus chart to allow flux2 post-render

### LINKED ISSUE #7424: Helm chart error when deployed with Flux2
**User Story**
Kubermatic Prometheus Chart cannot be deployed with Flux2 because post-renderer fails with:

`Helm install failed: error while running post render on files: may not add resource with an already registered id: rbac.authorization.k8s.io_v1_RoleBinding|~X|prometheus`

A check on the repo tells me that there is a missing `namespace` metadata what leads to a fail in kustomize which is the renderer within Flux2.

**Acceptance criteria**
.

Happy2C0de:

Adds missing namespace metadata to `Role` and `RoleBinding` in prometheus chart to fix Flux2 post-rendering.

Fixes #7424

**Documentation**:

```release-note
Fix helm post-rendering problems within monitoring/prometheus chart due to duplicate resource definitions.
```

### DIFF
--- a/charts/monitoring/prometheus/Chart.yaml
+++ b/charts/monitoring/prometheus/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: prometheus
-version: 2.3.6
+version: 2.3.7
 appVersion: v2.25.0
 description: Prometheus Monitoring for Kubernetes
 keywords:
--- a/charts/monitoring/prometheus/templates/role.yaml
+++ b/charts/monitoring/prometheus/templates/role.yaml
@@ -49,6 +49,7 @@ rules:
   - list
   - watch
 
+{{ if and (ne .Release.Namespace "default") (ne .Release.Namespace "kube-system") }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
@@ -67,3 +68,5 @@ rules:
   - get
   - list
   - watch
+
+{{- end }}
--- a/charts/monitoring/prometheus/templates/role.yaml
+++ b/charts/monitoring/prometheus/templates/role.yaml
@@ -54,6 +54,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: '{{ template "name" . }}'
+  namespace: '{{ .Release.Namespace }}'
 rules:
 - apiGroups:
   - ""
--- a/charts/monitoring/prometheus/templates/rolebinding.yaml
+++ b/charts/monitoring/prometheus/templates/rolebinding.yaml
@@ -46,6 +46,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: '{{ template "name" . }}'
+  namespace: '{{ .Release.Namespace }}'
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
--- a/charts/monitoring/prometheus/templates/rolebinding.yaml
+++ b/charts/monitoring/prometheus/templates/rolebinding.yaml
@@ -41,6 +41,7 @@ subjects:
   name: '{{ template "name" . }}'
   namespace: '{{ .Release.Namespace }}'
 
+{{- if and (ne .Release.Namespace "default") (ne .Release.Namespace "kube-system") }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
@@ -54,3 +55,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: '{{ template "name" . }}'
+
+{{- end }}


xrstf:
Why exactly is this necessary? If namespacing is a problem in flux, then the StatefulSet, ConfigMaps and many other resources from the Helm chart would also fail to be installed.

Happy2C0de:
@xrstf 
Thanks for having a look at this. I didn't had the time to debug this further until now.

Flux2 is using `helm` (Version 3) and `kustomize` in combination. Mainly, it uses `kustomize` as a postRenderer in helm what is very handy if you would like to adjust helm-charts without forking the upstream repo. Either one would like to set a value which is not configurable via `values.yaml` or one would like to patch something special within the k8s manifests. I made not usage of that while running into this problem, but there is a legitimacy to have this combination.

Mainly, one can reproduce the above problem locally using `helm` and `kustomize`.

Used versions:
- Helm: v3.5.4
- kustomize: v4.2.0

Reproduction:
```
git clone https://github.com/kubermatic/kubermatic.git
cd kubermatic/charts/monitoring/prometheus

# Build helm templates locally
helm template prometheus --output-dir ./test .

# Create kustomization.yaml needed by kustomize
cat << EOF > test/prometheus/templates/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- clusterrole.yaml
- clusterrolebinding.yaml
- configmaps.yaml
- migration.yaml
- role.yaml
- rolebinding.yaml
- service.yaml
- serviceaccount.yaml
- statefulset.yaml
- thanos-compact-deployment.yaml
- thanos-query-deployment.yaml
- thanos-query-service.yaml
- thanos-secret.yaml
- thanos-store-deployment.yaml
- thanos-store-service.yaml
- thanos-ui-deployment.yaml
- thanos-ui-service.yaml
EOF

# Run kustomize build
kustomize build --output kustomize.yaml ./test/prometheus/templates
```
The above failes with:
`Error: accumulating resources: accumulation err='accumulating resources from 'role.yaml': may not add resource with an already registered id: rbac.authorization.k8s.io_v1_Role|~X|prometheus'`

This is the same error as specified in issue #7424 .

Reason:
- kustomize detects resources with `apiVersion`, `kind`, `metadata/name` and `metadata/namespace`
- Kubermatic has a Role and a RoleBinding which is defined in `namespace: default` and anther one that has no namespace specified.
- Kustomize defaults the namespace metadata to `default`.

Fixes (locally):
There are two ways how to fix that:
1. Replace the `namespace: default` in the given Role and RoleBinding to e.g. `namespace: test`
2. Add `namespace: monitoring` to the unspecified Role and RoleBinding
You can test this fix with rerunning:
```
kustomize build --output kustomize.yaml ./test/prometheus/templates
```

Fix in kubermatic charts:
- Define the target namespace of the undefined Role and RoleBinding resources

Answer to your input:
> If namespacing is a problem in flux, then the StatefulSet, ConfigMaps and many other resources from the Helm chart would also fail to be installed.

Namespacing in general is not a problem. It's only a problem if the resources share `apiVersion, kind, metadata/name, metadata/namespace` what's the case for the given Role and RoleBinding.

I hope this clarifies the problem 👍

xrstf:
That clarifies the issue, I am just surprised that anyone would install the chart into the default namespace anyway. i.e. in the `helm template` call realistically `--namespace` should be set, because everyone (I Declare To Be True) always installs the chart into a special namespace.

Can you even install it into the default namespace, given the naming conflict you mentioned?

What do you think about adding an `{{ if .Release.Namespace != "default" }}` around the 2 offending resources? They are identical to the default-namespace-variant anyway, so we could simply skip emitting the same resources twice. :thinking:

Happy2C0de:
> What do you think about adding an `{{ if .Release.Namespace != "default" }}` around the 2 offending resources? They are identical to the default-namespace-variant anyway, so we could simply skip emitting the same resources twice. 🤔

Sounds reasonable. I did the commit but I left the `namespace: {{ .Release.Namespace  }}` because it's important to have the uniqueness within the given resources. If we omit it, it's not unique because if one specifies `-n prometheus` with the helm templating, the resource is still not unique because the rendering doesn't add the `metadata/namespace` key.
Accordingly, there is a need to render the namespace explicitly in this case. 

Good discussion. 👍

xrstf (approved):
/approve

Happy2C0de:
I rebased the branch to a newer master version. Additionally, I added a new chart release, `2.3.6`
@xrstf Can you add a LGTM

xrstf (approved):
/approve

xrstf:
@Happy2C0de please fix your release note in the PR description.

Happy2C0de:
@xrstf I added a release note. 👍

Happy2C0de:
Changed chart to `2.3.7` because yesterday, there was `2.3.6` merged.

Happy2C0de:
Again, a rebase to newest master was required.

PR #7440, opened 2021-08-02
Fix cluster list endpoint and improve error codes

maciaszczykm:
**What this PR does / why we need it**: At the moment when user doesn't have access to at least one of the datacenters that are used by the clusters, error will be returned. After this change problematic clusters will be filtered out.

Related to https://github.com/kubermatic/dashboard/issues/2866

```release-note
Fix cluster list endpoint that was returning errors when user didn't have access to at least one cluster datacenter.
```

### DIFF
--- a/pkg/handler/common/cluster.go
+++ b/pkg/handler/common/cluster.go
@@ -246,7 +246,7 @@ func GenerateCluster(ctx context.Context, projectID string, body apiv1.CreateClu
 	return partialCluster, nil
 }
 
-func GetExternalClusters(ctx context.Context, userInfoGetter provider.UserInfoGetter, clusterProvider provider.ClusterProvider, projectProvider provider.ProjectProvider, privilegedProjectProvider provider.PrivilegedProjectProvider, seedsGetter provider.SeedsGetter, projectID string) ([]*apiv1.Cluster, error) {
+func GetClusters(ctx context.Context, userInfoGetter provider.UserInfoGetter, clusterProvider provider.ClusterProvider, projectProvider provider.ProjectProvider, privilegedProjectProvider provider.PrivilegedProjectProvider, seedsGetter provider.SeedsGetter, projectID string) ([]*apiv1.Cluster, error) {
 	project, err := common.GetProject(ctx, userInfoGetter, projectProvider, privilegedProjectProvider, projectID, nil)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/handler/common/cluster.go
```diff
@@ -246,7 +246,7 @@ func GenerateCluster(ctx context.Context, projectID string, body apiv1.CreateClu
 	return partialCluster, nil
 }
 
-func GetExternalClusters(ctx context.Context, userInfoGetter provider.UserInfoGetter, clusterProvider provider.ClusterProvider, projectProvider provider.ProjectProvider, privilegedProjectProvider provider.PrivilegedProjectProvider, seedsGetter provider.SeedsGetter, projectID string) ([]*apiv1.Cluster, error) {
+func GetClusters(ctx context.Context, userInfoGetter provider.UserInfoGetter, clusterProvider provider.ClusterProvider, projectProvider provider.ProjectProvider, privilegedProjectProvider provider.PrivilegedProjectProvider, seedsGetter provider.SeedsGetter, projectID string) ([]*apiv1.Cluster, error) {
```
maciaszczykm, line 249: The name was misleading before as we have introduced the external clusters functionality. In this case external was referring to the externally exposed API type.

pkg/provider/conversions.go
```diff
@@ -25,79 +25,82 @@ import (
...
-	return foundSeeds[0], &foundDatacenters[0], nil
+	return matchingSeed, &matchingDatacenter, nil
 }
 
-func skipFilterByDomain(userInfo *UserInfo, datacenter kubermaticv1.Datacenter) bool {
-	requiredEmailDomain := datacenter.Spec.RequiredEmailDomain
-	requiredEmailDomains := datacenter.Spec.RequiredEmailDomains
+// canAccessDatacenter returns information if user with provided email domain can access given datacenter
+// based on the required email domains that are set on it.
+func canAccessDatacenter(dc kubermaticv1.Datacenter, emailDomain string) bool {
```
maciaszczykm, line 77: Currently the method will check both `RequiredEmailDomain` and `RequiredEmailDomains` if both are set. Previously when `requiredEmailDomain` was set and the requirement was met `RequiredEmailDomains` was ignored. This seemed wrong, but anyways both fields should not be used at the same time, and the first one is already deprecated.

maciaszczykm:
/assign @zreigz @xrstf @moadqassem

Let me know if you have any suggestions.

pkg/provider/conversions.go
```diff
@@ -25,79 +25,82 @@ import (
...
+		}
 	}
-	if requiredEmailDomain == "" && len(requiredEmailDomains) == 0 {
-		return true
+
+	return true
+}
+
+func getUserEmailDomain(userInfo *UserInfo) (string, error) {
+	split := strings.Split(userInfo.Email, "@")
```
zreigz, line 101: The spec for valid email addresses permits multiple "@" signs in an email address. You want to use the last one as the domain part, not the first one: https://play.golang.org/p/cg4RqZADLml

maciaszczykm: This is code moved from the old version, let's keep it as agreed during the meeting.

PR #7444, opened 2021-08-03
[promtail] kubelet log capturing and kubernetes event capturing

dharapvj:
* kubelet log capturing and kubernetes event capturing only if [eventrouter](https://github.com/heptiolabs/eventrouter) is installed)
* Also marks initContainer for promtail to be enabled by default and marks no of inotify instances to 256 as default as we observed that this has positive impact on worker machines's stable.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

Issue #7454, opened 2021-08-04
Support working declaratively


xrstf:
**User Story**
We have identified a lot of related tasks and technical debt and this epic is meant to coordinate between them:

* #6471
* #6337
* #5971
* #6692
* #7372
* #7398
* #7403

All of these eventually aim towards a single goal: turning KKP from an API-driven system to a declarative one, where working with custom resources directly (e.g. via `kubectl`) is well supported.

Another main aspect is the new rule regarding the `*.k8s.io` namespace: we currently use it and need to stop doing so.

To prevent changing our custom resources twice (once to change their API group to move away from k8s.io, and once when we turn it into a stable API), I suggest we for now keep the existing CRDs (using the cheating annotation) and then perform one big refactoring.

**Road to Success**

* Step 1: Gather requirements for the new architecture. We need to know what we want to achieve before we make any design decisions.

xrstf:
cc @moadqassem

embik:
This is a random thought I had, but should `Clusters` (eventually) become namespaced resources if we move to declarative support via Kubernetes API and CRDs? Otherwise, transferring the concept of a project becomes difficult, especially when it comes to (RBAC) permission management.

PR #7455, opened 2021-08-04
draft requirements for CRD refactoring

xrstf:
This is step 1 of #7454.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/kkp3-requirements.md
+++ b/docs/proposals/kkp3-requirements.md
@@ -1,0 +1,52 @@
+# KKP 3.0 Requirements
+
+**Author**: Christoph Mewes (@xrstf)
+
+**Status**: Drafting
+
+## Motivation and Background
+
+This document describes the basic assumptions and requirements we
+set for "KKP 3" (not an official name). See https://github.com/kubermatic/kubermatic/issues/7454
+for more information.
+
+## Requirements
+
+### Allow users to work declaratively
+
+Currently, while KKP uses Custom Resources (CRs) to store its data,
+most* interactions must happen through the REST API (e.g. creating a cluster).
+The CRs were never documented or considered stable, though of course
+that didn't stop people from manually editing them. Which is understandable
+because many things _must_ be done by editing them directly (for example
+to pause a cluster).
+
+To become even more Kubernetes-native and allow a GitOps-based workflow,
+we want to allow users to always (and possibly only) interact with our
+CRs directly.
+
+TL;DR: Allow `kubectl create -f mycluster.yaml`.
+
+### Have a stable, versioned API
+
+Currently our CRDs are unspec'ed and not very user friendly. When we
+turn them into our primary API, we need to make sure they are properly
+spec'ed.
+
+This also includes immediate feedback to the user if they try to create
+a malformed resource. Such resources must be rejected immediately and not
+simply cause errors later during reconciling.
+
+TL;DR: Generate spec/schema for types automatically, do not use unspec'ed CRDs
+anymore.
+
+### Do not use `*.k8s.io` anymore
+
+Only Kubernetes-approved APIs may use this domain, so we must move away to
+our own domain.
+
+TL;DR: Turn `kubermatic.k8s.io` into `kubermatic.io` for example.
+
+### Keep Master-Seed-Usercluster architecture
+
+We want to keep the architecture unchanged, but the naming is up for discussion.


docs/proposals/kkp3-requirements.md
```diff
@@ -0,0 +1,52 @@
...
+### Do not use `*.k8s.io` anymore
+
+Only Kubernetes-approved APIs may use this domain, so we must move away to
+our own domain.
+
+TL;DR: Turn `kubermatic.k8s.io` into `kubermatic.io` for example.
+
+### Keep Master-Seed-Usercluster architecture
+
+We want to keep the architecture unchanged, but the naming is up for discussion.
```
shaase-ctrl, line 52: I think this can be done later / independently but I would like to strengthen the seeds in several ways, p.e. allow only certain user groups to certain seeds to have a better separation of users (groups) and to put emphasis on security and wilful division. Furthermore only certain seeds shall provision certain machine types, so that deployments automagically happen only on certain "areas". Afaik this is not related, but please correct me if I am wrong.

jiachengxu:
First of all, all valuable requirements and I am looking forward to seeing this Kubermatic 3 in the future.

I just have one potential requirement that would like to discuss here:
```
Have only one single entry point (API server or cluster) that users will be interacting with.
```
Let me elaborate more on this. Currently, Kubermatic API server is the only entry point that users interact with. No matter users would like to create Project (in master cluster), or Cluster, Addon (in seed cluster), they will send requests to the KKP API server (in master cluster), and API server will create them in master and seed cluster.

Things can be different if we go fully declarative. We have master-scoped object(objects are in master cluster, e.g., Project), and seed-scoped object(objects are in seed cluster, e.g., Cluster, Addon, etc), and users will need to interact (e.g., `kubectl apply -f`) with both master and seed cluster. This means:
- If we want to use extension apiserver, we will need to deploy the extension apiserver on both master and seed clusters, and let them serve different APIs.
- If we want to go with CRD + Webhook, we will need to configure User Permission (RBAC) + authentication on both master and seed clusters.

IMHO, setting up any of the above options on both master and seed clusters can be complicated, and it is difficult for users to use, for example, they will need to figure out where to create what objects, and never forget to switch kubeconfig. Also, letting them access both master and seed clusters may be a bit too much.

Thus, I am thinking about if we just have a single entry point(API server or Cluster) to serve all Kubermatic APIs/CRDs which users will be interacting with, and then we have some propagate controllers to create underneath objects on seed clusters if needed.

WDYT?

xrstf:
> users will need to interact (e.g., kubectl apply -f) with both master and seed cluster.

Will they? What prevents them from only ever interacting with the master and the master takes care of syncing down to the seed clusters?

> Thus, I am thinking about if we just have a single entry point(API server or Cluster) to serve all Kubermatic APIs/CRDs which users will be interacting with, and then we have some propagate controllers to create underneath objects on seed clusters if needed.

I don't fully understand your question. I don't think we can avoid the situation where we have custom resources (however they are implemented) and a REST API for our dashboard (where, from a technical standpoint, `kubectl` and our REST API perform the same functionality: interacting with the kubernetes cluster and managing resources in it).

jiachengxu:
> Will they? What prevents them from only ever interacting with the master and the master takes care of syncing down to the seed clusters?

Nothing prevents that, and this is what I want to add as a requirement, because this syncing logic is currently missing in KKP, using CRD + webhook directly can not achieve the goal of managing all custom resources from master cluster. 

> I don't fully understand your question. I don't think we can avoid the situation where we have custom resources (however they are implemented) and a REST API for our dashboard (where, from a technical standpoint, `kubectl` and our REST API perform the same functionality: interacting with the kubernetes cluster and managing resources in it).

I got your point. I didn't mean that we should only have one of CRs and REST API. The `single entry point` in my previous comment means that, no matter we go with extension API server or CRD + webhooks, we need to let the user only interact with the master cluster directly (master cluster is the single entry point), not seeds. This is what I meant by `single entry point` for working declaratively.

The reason that I wanted to add this as a requiremetn is that I don't think `kubectl create -f cluster.yaml` is sufficient as a requirement, IMHO it should be
```
kubectl create -f cluster.yaml --kubeconfig=/path/to/master/cluster
```

xrstf:
This has been superseded by a Google Doc and is outdated now.

ewassef:
We are building an example of this for our own implementation and would like to share our approach since the google doc is probably not shared externally:
basically, we are building a simple translator between the API objects and their yaml equivalents. Of course some items may be references (cloud specific credentials, seed, dc  etc) and we are strategically referencing other objects in the CRD. The implementation is just an operator that converts the CRD to a JSON and calls the API internally. We preferred this approach in case yall change something in the bowels, chances are you will version the API and we will have time to update our own CRDs to match. 
My point is this could be a way to achieve your goals very quickly without a full rewrite. If the "operator" lives alongside the API container it can bypass the dex authenticating step (painful for service-based accounts right now) it will benefit from going down the same workflow logic effectively eliminating any drift between feature-sets. You will be able to create clusters via UI, API or CRD and they all converge into a single workflow eliminating potential bugs

We are building this in C# but we are using a grossly simplified CRD with opinionated options for clusters, but it would be trivial to expand to the whole API object mapping

PR #7456, opened 2021-08-04
Remove non-existing field from kube-proxy configmap

kron4eg:
It's causing unmarshal errors during e2e tests

```
using lenient decoding as strict decoding failed: strict decoder error ReadObject: found unknown field: max, error found in #10 byte of ...|ck":{"max":null,"max|..., bigger context ...|/16","configSyncPeriod":"15m0s","conntrack":{"max":null,"maxPerCore":0,"min":524288,"tcpCloseWaitTim|...
```

```release-note
NONE
```


moadqassem (commented):
/approve
/lgtm

PR #7459, opened 2021-08-05
prevent kubevirt kubeconfig from confusing the conformance tester

xrstf:
Handing the multiline JSON kubeconfig as an unquoted CLI flag caused lots of trouble. I tried to `jq -c` the kubeconfig and use quotes, but to no avail. So this PR simply changes the logic to read the kubeconfig from a file.

```release-note
NONE
```

### DIFF
--- a/cmd/conformance-tests/main.go
+++ b/cmd/conformance-tests/main.go
@@ -158,11 +158,12 @@ type secrets struct {
 }
 
 var (
-	providers             string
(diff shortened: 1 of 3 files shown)


moadqassem:
@xrstf Can't you hand in the kubeconfig data base64 and then decode before the running of the test?

kron4eg (approved):

PR #7461, opened 2021-08-05
Add support for Kubernetes 1.22, remove 1.17 and 1.18

### LINKED ISSUE #7457: Support Kubernetes 1.22
**User Story**
KKP should support userclusters using Kubernetes 1.22.

**Acceptance criteria**

xrstf:

Fixes #7457

**Documentation**:

```release-note
Add Kubernetes 1.22
Remove Kubernetes 1.17 and 1.18
```


addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 {{ $version := "UNSUPPORTED" }}
-{{ if eq .Cluster.MajorMinorVersion "1.18" }}
-{{ $version = "v1.18.2 "}}
-{{ end }}
```
xmudrii, line 18: I think we should still keep 1.18 because it was supported in KKP 2.17. For example, if someone upgrades KKP while still having 1.18 clusters, the cluster-autoscaler might stop working, which is not expected. In the ideal scenario, this shouldn't happen, but it's possible.

addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 {{ $version := "UNSUPPORTED" }}
```
xmudrii, line 15: Please add support for cluster-autoscaler v1.22.0: https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.22.0

addons/csi/openstack/controllerplugin.yaml
```diff
@@ -31,6 +25,9 @@
 {{ if eq .Cluster.MajorMinorVersion "1.21" }}
 {{ $version = "v1.20.3 "}}
 {{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.22" }}
+{{ $version = "v1.21.1 "}}
```
xmudrii: Why are we one minor version behind? I think we should use 1.22.0 for 1.22 clusters (https://github.com/kubernetes/cloud-provider-openstack/releases/tag/v1.22.0).

addons/csi/openstack/nodeplugin.yaml
```diff
@@ -31,6 +25,9 @@
 {{ if eq .Cluster.MajorMinorVersion "1.21" }}
 {{ $version = "v1.20.3 "}}
 {{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.22" }}
+{{ $version = "v1.20.3 "}}
+{{ end }}
```
xmudrii: This should match the version used in `addons/csi/openstack/controllerplugin.yaml`.

addons/csi/vsphere/csiController.yaml
```diff
@@ -28,6 +25,9 @@
 {{ if eq .Cluster.MajorMinorVersion "1.21" }}
 {{ $version = "v2.2.1 "}}
 {{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.22" }}
+{{ $version = "v2.2.1 "}}
+{{ end }}
```
xmudrii: This is a bit more complicated because vSphere CSI doesn't support 1.22. We should block upgrades to 1.22 for clusters using vSphere CSI. This has been discussed internally, but we need to follow up on it.

addons/cluster-autoscaler/cluster-autoscaler.yaml
xrstf: But we cannot easily support 1.18 because tons of APIs aren't available anymore. We should maybe think more about removing 1.18 from KKP 2.17, as it's EOL for quite some time already.

pkg/resources/data.go
```diff
@@ -650,6 +650,7 @@ func GetCSIMigrationFeatureGates(cluster *kubermaticv1.Cluster) []string {
 			if cluster.Spec.Cloud.Openstack != nil {
 				featureFlags = append(featureFlags, "CSIMigrationOpenStackComplete=true")
 			}
```
xmudrii, line 653: TODO: OpenStack is affected as well.

pkg/resources/dns/dns.go
```diff
@@ -50,10 +50,6 @@ var (
 // source: https://github.com/kubernetes/kubernetes/blob/vX.YY.0/cmd/kubeadm/app/constants/constants.go
 func GetCoreDNSImage(kubernetesVersion *semver.Version) string {
 	switch fmt.Sprintf("%d.%d", kubernetesVersion.Major(), kubernetesVersion.Minor()) {
-	case "1.19":
-		return "coredns/coredns:v1.7.0"
-	case "1.20":
-		return "coredns/coredns:v1.7.0"
```
xmudrii: Nit: it might be safer to leave those cases in place and make default return `coredns/coredns:v1.8.4`. For example, if we for some reason forget to add a new case for a new Kubernetes version, it'll downgrade CoreDNS, which can be a bit problematic.

xmudrii (approved):
/lgtm
/approve

xrstf:
NB: I will update kube-state-metrics to 2.2.0 in a dedicated PR, as that version was just now released and has improved 1.22 support.

PR #7472, opened 2021-08-09
extend global settings for cluster template sizes

zreigz:
**What this PR does / why we need it**: extend global settings for cluster template sizes

Fixes #

```release-note
NONE
```


pkg/crd/kubermatic/v1/settings.go
```diff
@@ -102,6 +103,20 @@ type MlaOptions struct {
...
+	Sizes []ClusterTemplateSize `json:"clusterTemplateSizes"`
+}
+
+type ClusterTemplateSize struct {
+	// Size name. Can be for example: small, large, L, XL
+	Name string `json:"name"`
+	// Number of clusters
+	NumberOfClusters int64 `json:"numberOfClusters"`
+	// Number od machine deployments
+	NumberOfMachineDeployments int64 `json:"numberOfMachineDeployments"`
```
mfranczy, line 117: are we talking here about replicas of the same cluster and replicas of the same machine deployments?

pkg/crd/kubermatic/v1/settings.go
```diff
@@ -102,6 +103,20 @@ type MlaOptions struct {
...
 }
 
+type ClusterTemplateOptions struct {
+	// Predefined cluster template sizes
+	Sizes []ClusterTemplateSize `json:"clusterTemplateSizes"`
+}
+
+type ClusterTemplateSize struct {
+	// Size name. Can be for example: small, large, L, XL
+	Name string `json:"name"`
```
maciaszczykm, line 113: Are there any restrictions for the name? If it can be similar to the clusters (so anything alphanumeric) then I'd avoid giving examples. It can be misleading.

PR #7479, opened 2021-08-11
Allow Custom Scrape Configs for User Cluster MLA Prometheus

### LINKED ISSUE #7363: Allow extending Prometheus scrape targets in user clusters
Allow extending of user cluster Prometheus scrape targets.
That is needed e.g. to support blackbox exporter in user clusters.

jiachengxu:
This PR allows users to specify customized scrape configs for User Cluster MLA Prometheus by adding ConfigMaps with a pre-defined name prefix `prometheus-scraping`. For example,
User can create a configmap from a file as follows:
```bash
kubectl create configmap prometheus-scraping-test --from-file=custom-scrape-configs.yaml -n mla-system
```
and the custom-scrape-configs.yaml can look like following:
```yaml
- job_name: 'prometheus'
  static_configs:
  - targets: ["localhost:9090"]
- job_name: 'schnitzel'
  kubernetes_sd_configs:
  - role: pod
  relabel_configs:
  - source_labels: [__meta_kubernetes_pod_annotation_kubermatic_scrape]
    action: keep
    regex: true
```
User will have to make sure to provide valid configurations. For now, after the customized scrape configs are added, the user will need to restart Prometheus or curl the reload endpoint to trigger a configuration reload manually, and the automation of reloading will be implemented in next PR.

Fixes #7363 

**Documentation**:

```release-note
NONE
```


cmd/user-cluster-controller-manager/main.go
```diff
@@ -59,33 +61,34 @@ import (
...
+	dnsClusterIP                                            string
+	nodeLocalDNSCache                                       bool
+	opaIntegration                                          bool
+	opaWebhookTimeout                                       int
+	useSSHKeyAgent                                          bool
+	caBundleFile                                            string
+	mlaGatewayURL                                           string
+	userClusterLogging                                      bool
+	userClusterMonitoring                                   bool
+	userClusterPrometheusCustomScrapeConfigsConfigMapPrefix string
```
xrstf: Maybe it would be possible to find a somewhat shorter name for this.

jiachengxu: Naming is hard :)
Just shortened it to `prometheusScrapeConfigMapPrefix`

jiachengxu:
@xrstf Can you please take a look?

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -60,9 +60,10 @@ const (
 )
 
 type UserClusterMLA struct {
-	Logging       bool
-	Monitoring    bool
-	MLAGatewayURL string
+	Logging                               bool
+	Monitoring                            bool
+	MLAGatewayURL                         string
+	PrometheusCustomScrapeConfigMapPrefix string
```
xmudrii: This name is also quite long. Maybe just `CustomScrapeConfigPrefix` or something like that would be better.

jiachengxu: How about we drop `Custom` and `Map`, so it will be `PrometheusScrapeConfigPrefix`? I would like to have `Prometheus` in the name, because we may have the same thing for `Loki` (for logs) in the future, for example, `LokiScrapeConfigPrefix`

xmudrii: Sounds good. :+1:

jiachengxu: Fixed :)

xmudrii (approved):

jiachengxu:
@xmudrii We need an `/approve` from your side ;)

PR #7499, opened 2021-08-14
use simple amd64 images for user-ssh-key-agent during e2e tests

xrstf:
Currently, the multiarch code for the user-ssh-key-agent is running for everyone who uses `setup-kubermatic-in-kind`, i.e. most tests. We don't actually build anything but the amd64 version, so we waste a lot of time trying to setup a cache and copying 4+GB worth of gazillion of tiny files into the Docker context when building the image.

This PR kind of reverts that and returns the e2e test logic back to the amd64-only code. If also changes the release code to just ignore the gocache entirely. It's slower, but we also don't have to cache arm64 packages that would needlessly blow up the cache.

(EDIT: Actually, building without gocache takes 8min for both amd64 and arm64 -- that is like 20% of the old runtime where we tried to manage the gocache -- i.e. not having a cache for this is actually faster and simpler.)

In the future, we can see if we can maybe separate the caches for architectures, but for now we need to bring our tests back to normal runtimes.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -395,837 +395,837 @@ presubmits:
   # Base Kubernetes Tests (AWS, Flatcar)
   #########################################################
 
-  - name: pre-kubermatic-e2e-aws-flatcar-1.18
-    decorate: true
-    run_if_changed: "(cmd/|codegen/|hack/|pkg/|charts/kubermatic|addons/|.prow.yaml)"
-    clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
-    labels:
-      preset-aws: "true"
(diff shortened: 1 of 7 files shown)


xrstf:
/override pre-kubermatic-ccm-migration-e2e

Fails because of OpenStack quota issues, unrelated to this PR.

xmudrii (approved):

xrstf:
/override pre-kubermatic-ccm-migration-e2e
/test  @kubermatic-bot
pre-kubermatic-api-e2e

xrstf:

Flaking without any end in sight.

PR #7504, opened 2021-08-16
add konnectivity option for controlplane to cluster communication 

PratikDeoghare:

This PR implements part of 
-  Epic https://github.com/kubermatic/kubermatic/issues/6188
-  Proposal [Exposing control-plane services using SNI and HTTP tunneling
](https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/konnectivity-openvpn.md)

It deploys konnectivity-proxy-server in sidecar in the KAS pods. 
It also deploys two agents one in user-cluster and one in seed-cluster. 
- User-cluster agent 
   This agent handles traffic going from KAS to user-cluster. 
- Seed-cluster agent 

   This agent handles traffic going to `metric-server` allowing us to avoid moving `metric-server` to user-cluster. 
   PR deviates here from the original proposal. We had planned to move `metric-server` to user cluster. 
  
   This agent uses token stolen from user-cluster.  This will be made more efficient in future PRs. See https://github.com/kubermatic/kubermatic-installer/issues/139. 

References: 
- [Set up Konnectivity service
](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/ )

```release-note
Add konnectivity option for controlplane to cluster communication
```

### DIFF
--- a/hack/run-user-cluster-controller-manager.sh
+++ b/hack/run-user-cluster-controller-manager.sh
@@ -89,4 +89,5 @@ set -x
   -seed-kubeconfig=${SEED_KUBECONFIG} \
   -owner-email=${OWNER_EMAIL} \
   -dns-cluster-ip=10.240.16.10 \
+  -konnectivity-enabled=true \
   ${ARGS}
--- a/pkg/controller/seed-controller-manager/kubernetes/resources.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/resources.go
@@ -31,6 +31,7 @@ import (
 	"k8c.io/kubermatic/v2/pkg/resources/dns"
 	"k8c.io/kubermatic/v2/pkg/resources/etcd"
 	"k8c.io/kubermatic/v2/pkg/resources/gatekeeper"
+	"k8c.io/kubermatic/v2/pkg/resources/konnectivity"
 	kubernetesdashboard "k8c.io/kubermatic/v2/pkg/resources/kubernetes-dashboard"
 	"k8c.io/kubermatic/v2/pkg/resources/machinecontroller"
 	metricsserver "k8c.io/kubermatic/v2/pkg/resources/metrics-server"
@@ -45,6 +46,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (r *Reconciler) ensureResourcesAreDeployed(ctx context.Context, cluster *kubermaticv1.Cluster) error {
@@ -188,6 +190,7 @@ func (r *Reconciler) getClusterTemplateData(ctx context.Context, cluster *kuberm
 		WithInClusterPrometheusDefaultScrapingConfigsDisabled(r.inClusterPrometheusDisableDefaultScrapingConfigs).
 		WithInClusterPrometheusScrapingConfigsFile(r.inClusterPrometheusScrapingConfigsFile).
 		WithUserClusterMLAEnabled(r.userClusterMLAEnabled).
+		WithKonnectivityEnabled(r.features.Konnectivity).
 		WithCABundle(r.caBundle).
 		WithOIDCIssuerURL(r.oidcIssuerURL).
 		WithOIDCIssuerClientID(r.oidcIssuerClientID).
@@ -242,9 +245,14 @@ func GetServiceCreators(data *resources.TemplateData) []reconciling.NamedService
 		metricsserver.ServiceCreator(),
 	}
 
+	if data.IsKonnectivityEnabled() {
+		creators = append(creators, konnectivity.ServiceCreator(data.Cluster().Spec.ExposeStrategy, data.Cluster().Address.ExternalName))
+	}
+
 	if data.Cluster().Spec.ExposeStrategy == kubermaticv1.ExposeStrategyLoadBalancer {
 		creators = append(creators, nodeportproxy.FrontLoadBalancerServiceCreator())
 	}
+
 	if flag := data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureRancherIntegration]; flag {
 		creators = append(creators, rancherserver.ServiceCreator(data.Cluster().Spec.ExposeStrategy))
 	}
@@ -271,6 +279,11 @@ func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuth
 		usercluster.DeploymentCreator(data),
 		kubernetesdashboard.DeploymentCreator(data),
 	}
+
+	if data.IsKonnectivityEnabled() {
+		deployments = append(deployments, konnectivity.DeploymentCreator(data))
+	}
+
 	if data.Cluster().Annotations[kubermaticv1.AnnotationNameClusterAutoscalerEnabled] != "" {
 		deployments = append(deployments, clusterautoscaler.DeploymentCreator(data))
 	}
@@ -292,7 +305,7 @@ func (r *Reconciler) ensureDeployments(ctx context.Context, cluster *kubermaticv
 }
 
 // GetSecretCreators returns all SecretCreators that are currently in use
-func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconciling.NamedSecretCreatorGetter {
+func (r *Reconciler) GetSecretCreators(data *resources.TemplateData, userClusterClient ctrlruntimeclient.Client) []reconciling.NamedSecretCreatorGetter {
 	creators := []reconciling.NamedSecretCreatorGetter{
 		certificates.RootCACreator(data),
 		openvpn.CACreator(),
@@ -325,6 +338,14 @@ func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconcili
 		resources.ViewerKubeconfigCreator(data),
 	}
 
+	if data.IsKonnectivityEnabled() {
+		creators = append(creators, []reconciling.NamedSecretCreatorGetter{
+			konnectivity.AgentTokenCreator(userClusterClient),
+			konnectivity.ProxyKubeconfig(data),
+			konnectivity.TLSServingCertificateCreator(data),
+		}...)
+	}
+
 	if flag := data.Cluster().Spec.Features[kubermaticv1.ClusterFeatureExternalCloudProvider]; flag {
 		creators = append(creators, resources.GetInternalKubeconfigCreator(
 			resources.CloudControllerManagerKubeconfigSecretName, resources.CloudControllerManagerCertUsername, nil, data,
@@ -339,7 +360,17 @@ func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconcili
 }
 
 func (r *Reconciler) ensureSecrets(ctx context.Context, c *kubermaticv1.Cluster, data *resources.TemplateData) error {
-	namedSecretCreatorGetters := r.GetSecretCreators(data)
+	var userClusterClient ctrlruntimeclient.Client
+	var err error
+
+	if data.IsKonnectivityEnabled() {
+		userClusterClient, err = r.userClusterConnProvider.GetClient(ctx, c)
+		if err != nil {
+			return fmt.Errorf("failed to get user-cluster client: %v", err)
+		}
+	}
+
+	namedSecretCreatorGetters := r.GetSecretCreators(data, userClusterClient)
 
 	if err := reconciling.ReconcileSecrets(ctx, namedSecretCreatorGetters, c.Status.NamespaceName, r.Client, reconciling.OwnerRefWrapper(resources.GetClusterRef(c))); err != nil {
 		return fmt.Errorf("failed to ensure that the Secret exists: %v", err)
@@ -430,6 +461,10 @@ func GetConfigMapCreators(data *resources.TemplateData) []reconciling.NamedConfi
 		apiserver.AuditConfigMapCreator(),
 		apiserver.AdmissionControlCreator(data),
 		apiserver.CABundleCreator(data),
+	}
+
+	if data.IsKonnectivityEnabled() {
+		creators = append(creators, apiserver.EgressSelectorConfigCreator())
 	}
 
 	if data.Cluster().Spec.Cloud.VSphere != nil {
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -31,6 +31,7 @@ import (
 	dnatcontroller "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/dnat-controller"
 	envoyagent "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/envoy-agent"
 	"k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/gatekeeper"
+	"k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity"
 	kubestatemetrics "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/kube-state-metrics"
 	kubernetesdashboard "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/kubernetes-dashboard"
 	machinecontroller "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/resources/resources/machine-controller"
@@ -113,7 +114,7 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 		return err
 	}
 
-	if err := r.reconcileServiceAcconts(ctx); err != nil {
+	if err := r.reconcileServiceAccounts(ctx); err != nil {
 		return err
 	}
 
@@ -212,7 +213,7 @@ func (r *reconciler) ensureAPIServices(ctx context.Context, data reconcileData) 
 	return nil
 }
 
-func (r *reconciler) reconcileServiceAcconts(ctx context.Context) error {
+func (r *reconciler) reconcileServiceAccounts(ctx context.Context) error {
 	creators := []reconciling.NamedServiceAccountCreatorGetter{
 		userauth.ServiceAccountCreator(),
 		usersshkeys.ServiceAccountCreator(),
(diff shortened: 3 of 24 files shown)


PratikDeoghare:
Don't look yet. Might hurt eyes. :warning:

pkg/resources/resources.go
```diff
@@ -635,6 +635,14 @@ alertmanager_config: |
...
     - name: 'null'
 `
+
+	// Konnectivity
+	KonnectivityDeploymentName         = "konnectivity-agent"
+	UserClusterKonnectivityDaemonSetName          = "konnectivity-agent"
+	UserClusterKonnectivityClusterRoleBindingName = "system:konnectivity-server"
+	UserClusterKonnectivityServiceAccountName     = "konnectivity-agent"
+	KonnectivityAgentContainer                    = "konnectivity-agent"
+	KonnectivityProxyServiceName                  = "konnectivity-server"
```
xrstf: Just from looking at this I am not sure if any of these names are used inside the usercluster. If so, please add some prefix so we do not actually clash names with whatever the user wants to do in their cluster. Also please do not use "kubermatic" or "kkp" as the prefix, as we _try_ to be somewhat white-label.

If the names are only used in the controlplane (i.e. the `cluster-....` namespace on the seed), then the names are totally fine.

PratikDeoghare: Some of them are used in the user cluster. Didn't think about prefixing them. Will prefix them. Thanks for the suggestion. :pray:

jiachengxu: > Also please do not use "kubermatic" or "kkp" as the prefix, as we try to be somewhat white-label.

@xrstf This is a good point, now I am thinking maybe we should also drop the `kubermatic` prefix for user cluster MLA resources, currently, those resources are prefixed with `system:kubermatic:mla`, we can just have `system:mla`.

jiachengxu: > Also please do not use "kubermatic" or "kkp" as the prefix, as we try to be somewhat white-label.

@xrstf This is a good point, now I am thinking maybe we should also drop the `kubermatic` prefix for user cluster MLA resources, currently, those resources are prefixed with `system:kubermatic:mla`, we can just have `system:mla`.

hack/run-user-cluster-controller-manager.sh
```diff
@@ -44,7 +44,7 @@ KUBECONFIG_USERCLUSTER_CONTROLLER_FILE=$(mktemp)
 kubectl --namespace "$NAMESPACE" get secret internal-admin-kubeconfig -o json |
   jq '.data.kubeconfig' -r |
   base64 -d |
-  yq r --tojson - |
+  yq eval --tojson - |
```
PratikDeoghare: reminder to self: maybe undo

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/clusterrolebinding.go
```diff
@@ -0,0 +1,35 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// ClusterRoleBindingCreator returns a func to create/update the ClusterRoleBinding for Prometheus
+func ClusterRoleBindingCreator() reconciling.NamedClusterRoleBindingCreatorGetter {
+	return func() (string, reconciling.ClusterRoleBindingCreator) {
+		return resources.UserClusterKonnectivityClusterRoleBindingName, func(crb *rbacv1.ClusterRoleBinding) (*rbacv1.ClusterRoleBinding, error) {
+			crb.ObjectMeta.Labels = map[string]string{
+				"kubernetes.io/cluster-service":   "true",
```
PratikDeoghare, line 31: What do these do?

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -715,6 +731,11 @@ func (r *reconciler) reconcileDeployments(ctx context.Context) error {
 		coredns.DeploymentCreator(r.clusterSemVer),
 	}
 
+	if r.isKonnectivityEnabled {
+		kubeSystemCreators = append(kubeSystemCreators,
+			konnectivity.DeploymentCreator(strings.TrimPrefix(r.namespace, "cluster-")))
```
PratikDeoghare: Is that cool? Any way to get clusterID here that I missed?

mfranczy:
@PratikDeoghare please squash commits and put more descriptive comments.

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/clusterrolebinding.go
irozzo-1A: https://github.com/kubernetes/kubernetes/issues/72757

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -325,6 +339,18 @@ func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconcili
...
 	}
 
+	if data.IsKonnectivityEnabled() {
+		// defer getting the CA
+		getCA := func() (*triple.KeyPair, error) {
+			return resources.GetClusterRootCA(context.Background(), data.Cluster().Status.NamespaceName, r.Client)
+		}
+		creators = append(creators, []reconciling.NamedSecretCreatorGetter{
+			konnectivity.AgentTokenCreator(userClusterClient),
+			konnectivity.ProxyKubeconfig(getCA, data.Cluster().Address.URL), // TODO(pratik): use GetInternalKubeconfigCreator?
```
irozzo-1A: any particular reason why you extracted `getCA` function instead of doing:
```
konnectivity.ProxyKubeconfig(data)
```
?

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -339,7 +365,13 @@ func (r *Reconciler) GetSecretCreators(data *resources.TemplateData) []reconcili
 }
 
 func (r *Reconciler) ensureSecrets(ctx context.Context, c *kubermaticv1.Cluster, data *resources.TemplateData) error {
-	namedSecretCreatorGetters := r.GetSecretCreators(data)
+
+	userClusterClient, err := r.userClusterConnProvider.GetClient(ctx, c)
```
irozzo-1A: Do you think it is a good idea to `GetClient` at each reconciliation?
I don't have evidence it has a big impact at the moment, but in case we could consider caching clients or moving this logic in user-cluster-controller-manager later.

pkg/controller/seed-controller-manager/kubernetes/resources.go
PratikDeoghare: Idea was to never make the call to get the cluster CA if the secret already exists. I got it from the codebase itself. 

But that is not what I am doing in the code and GetCA is available on data itself. 

Thanks for catching that. Fixing.

PratikDeoghare: Fixed.

pkg/controller/seed-controller-manager/kubernetes/resources.go
PratikDeoghare: Not if it can be avoided. Lemme check if that is possible.

PratikDeoghare: On every reconcile is definitely too frequent. We only need this when token is about to expire.

PratikDeoghare: Created an issue to fix this as discussed. https://github.com/kubermatic/kubermatic-installer/issues/139

pkg/resources/resources.go
PratikDeoghare: No idea about what good prefix might be. How about `network:`?

pkg/resources/konnectivity/tls-cert.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+
+			inClusterIP, err := resources.InClusterApiserverIP(data.Cluster())
+			if err != nil {
+				return nil, fmt.Errorf("failed to get the in-cluster ClusterIP for the apiserver: %v", err)
+			}
+
+			altNames := certutil.AltNames{
+				DNSNames: []string{
+					fmt.Sprintf("konnectivity-server.%s.europe-west3-c.dev.kubermatic.io", data.Cluster().Name),
```
rastislavs: Hardcoded value `europe-west3-c.dev.kubermatic.io`? That does not seem to be working for anything else then our dev setup.

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/deployment.go
```diff
@@ -0,0 +1,99 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package konnectivity
+
+import (
```
rastislavs, line 19: please sort the imports according to https://github.com/kubermatic/kubermatic/blob/master/docs/code-style.md#imports (in all files)

pkg/resources/konnectivity/sidecar.go
```diff
@@ -0,0 +1,98 @@
...
+			fmt.Sprintf("--cluster-cert=/etc/kubernetes/pki/%s.crt", resources.KonnectivityProxyTLSSecretName),
+			"--uds-name=/etc/kubernetes/konnectivity-server/konnectivity-server.socket",
+			fmt.Sprintf("--kubeconfig=/etc/kubernetes/kubeconfig/%s", resources.KonnectivityServerConf),
+			fmt.Sprintf("--server-count=%d", serverCount),
+			"--mode=grpc",
+			"--server-port=0",
+			"--agent-port=8132",
+			"--admin-port=8133",
+			"--health-port=8134",
+			"--agent-namespace=kube-system",
```
rastislavs, line 55: shouldn't this be a user cluster namespace?

pkg/resources/konnectivity/sidecar.go
```diff
@@ -0,0 +1,98 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// ProxySidecar returns container that runs konnectivity proxy server as a sidecar in apiserver pods.
+func ProxySidecar(serverCount int32) (*corev1.Container, error) {
+	return &corev1.Container{
+		Name:            "konnectivity-server-container",
+		Image:           "us.gcr.io/k8s-artifacts-prod/kas-network-proxy/proxy-server:v0.0.24",
```
rastislavs: (in whole PR) Please do not hardcode the image registry, use the [ImageRegistry](https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/data.go#L403-L409) helper to allow overring it.

pkg/resources/konnectivity/proxyservice.go
```diff
@@ -0,0 +1,80 @@
...
+
+			const port = 8132
+
+			se.Spec.Ports[0].Name = "secure"
+			se.Spec.Ports[0].Port = 443
+			se.Spec.Ports[0].Protocol = corev1.ProtocolTCP
+			se.Spec.Ports[0].TargetPort = intstr.FromInt(port)
+
+			if exposeStrategy == kubermaticv1.ExposeStrategyTunneling {
+				se.Spec.Ports[0].NodePort = 0 // TODO(pratik): why?
```
rastislavs: Allows switching from other expose strategies.. which is not possible ATM anyway, but I would still prefer to keep it for consistency with the rest of the code

pkg/resources/apiserver/deployment.go
```diff
@@ -394,6 +426,19 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		)
 	}
 
+	if data.IsKonnectivityEnabled() {
+		flags = append(flags,
+			"--egress-selector-config-file",
+			"/etc/kubernetes/konnectivity/egress-selector-configuration.yaml",
+			"--service-account-issuer",
```
rastislavs: you cannot just append `--service-account-issuer`, `--service-account-signing-key-file` and `--api-audiences` flags like this, if they were already appended by the existing code 50 lines above

rastislavs (changes requested):

pkg/resources/apiserver/egressselectorconfigmap.go
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
mfranczy: Could you change to 2021 everywhere where you create a new file?

pkg/resources/apiserver/egressselectorconfigmap.go
```diff
@@ -0,0 +1,46 @@
...
+
+// EgressSelectorConfigCreator returns function to create cm that contains egress selection configuration for apiserver
+// to work with konnectivity proxy.
+func EgressSelectorConfigCreator() reconciling.NamedConfigMapCreatorGetter {
+	return func() (string, reconciling.ConfigMapCreator) {
+		return resources.KonnectivityKubeApiserverEgress, func(c *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+			c.Data = map[string]string{
+				"egress-selector-configuration.yaml": `
+apiVersion: apiserver.k8s.io/v1beta1
+kind: EgressSelectorConfiguration
```
mfranczy: could you build EgressSelectorConfiguration struct instead of putting here a string? 
That would be better for the struct validation. 

Then you would have to get the string representation of a struct.

https://github.com/kubernetes/apiserver/blob/0ff020160db9f12f71f1278528cb974c978ba750/pkg/apis/apiserver/types.go#L55

pkg/resources/data.go
```diff
@@ -87,6 +87,7 @@ type TemplateData struct {
 	inClusterPrometheusScrapingConfigsFile           string
 
 	userClusterMLAEnabled bool
+	konnectivityEnabled   bool
```
mfranczy: could you keep the same name pattern everywhere? Either choose `IsKonnectivtyEnabled` or `KonnectivityEnabled`.

pkg/resources/apiserver/deployment.go
```diff
@@ -357,7 +385,11 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 	if cluster.Spec.Cloud.GCP != nil {
 		flags = append(flags, "--kubelet-preferred-address-types", "InternalIP")
 	} else {
-		flags = append(flags, "--kubelet-preferred-address-types", "ExternalIP,InternalIP")
+		if data.IsKonnectivityEnabled() {
+			flags = append(flags, "--kubelet-preferred-address-types", "InternalIP,ExternalIP")
```
rastislavs, line 398: Why exactly is this needed and what are the potential outcomes / risks?

PratikDeoghare: > Why... 

Without this fetching logs for pods in user-cluster fails. The problem discussed [here](https://kubermatic.slack.com/archives/C01EWQZEW69/p1628769575001400). 

KAS tries to connect to kubelet via konnectivity-agent in the user-cluster. This request fails because of security policies disallow external traffic to the node.  

> risks... 

All egress traffic from KAS goes through the konnectivity-proxy and to the konnectivity-agents.  Agents are authenticated when they register themselves with konenctivity-proxy. 

Without this we had to modify current security policy of the node to allow external traffic. 

So no security issues that I can spot. 

This was done under guidance of  :) @youssefazrak and  @irozzo-1A

PratikDeoghare:
@rastislavs @mfranczy thanks for the reviews. Yes to all the comments. Fixing....

pkg/resources/apiserver/deployment.go
rastislavs: OK, understood, thx for the explanation.. maybe a comment in there would help others who will be reading the code later.

PratikDeoghare: Ack. Will add it. :+1:

irozzo-1A: I confirm, and agree with @rastislavs that a comment here could result useful.

pkg/resources/konnectivity/sidecar.go
PratikDeoghare: https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/47a9234435b115cc764d76c8336057ea27ed0db6/cmd/server/app/options/options.go#L96

Interesting proxy-server accepts only one value there but we have agents deployed in seed-cluster in `cluster-somename` and one in `kube-system` on the user-cluster. 

Let me check again how this is being used.

pkg/resources/konnectivity/sidecar.go
PratikDeoghare: User-cluster-controller doesn't seem to have way of overriding registry. I have hard coded it there right now.

PratikDeoghare: Seed cluster controller is fixed. Pushing in a moment.

pkg/resources/konnectivity/proxyservice.go
PratikDeoghare: I want to know how exactly that switching would work tho.

pkg/resources/apiserver/egressselectorconfigmap.go
PratikDeoghare: Done.

pkg/resources/konnectivity/sidecar.go
PratikDeoghare:
```
.... 
"status":{"authenticated":true,"user":{"username":"system:serviceaccount:kube-system:system-konnectivity-agent",
... 
```

It should be `kube-system`. 
We deploy user-cluster konnectivity-agent in kube-system namespace. 
If the token of associated service account validates successfully (at konnectivity-proxy) we get something like above in the status. 
This is what konnectivity-proxy compares with `--agent-namespace` value. See here https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/47a9234435b115cc764d76c8336057ea27ed0db6/pkg/server/server.go#L568

pkg/resources/apiserver/deployment.go
PratikDeoghare: Done.

mfranczy:
@PratikDeoghare from my perspecitve the PR looks good, please rebase code.

rastislavs:
@PratikDeoghare looks good from my perspective as well. The only thing I wish this would include is a per-user cluster option for enabling Konnectivity, as right now it will be enabled for whole KKP installation by the feature gate. That can be quite risky for existing KKP deployments :) But I guess this can be done in a future change as well.

rastislavs (approved):

mfranczy:
each test fails with
```
make: *** [Makefile:50: _build/conformance-tests] Error 2
Retry 1/1 exited 2, no more retries left.
```

PratikDeoghare:
@mfranczy that was a bad merge. I fixed it earlier. 
I guess praw doesn't automatically cancel already triggered tests? 

```
pkg/resources/konnectivity/seedagent.go:50:66: undefined: resources.RegistryUSGCR
```

mfranczy:
@PratikDeoghare ok

/retest

PR #7512, opened 2021-08-17
Extend CCM migration E2E tests to cleanup MachineDeployments and wait for resources to be deleted

xmudrii:

The CCM migration tests are currently not properly cleaning up cloud resources:

* The MachineDeployment object is not explicitly removed, so we leak one OpenStack instance per test run
* We don't wait for cluster deletion to be complete, so it happens that we also leak SecurityGroups, networks, and other resources

This is causing us to reach OpenStack limits very frequently, causing all other OpenStack tests to fail.

This PR addresses this problem by explicitly deleting MachineDeployment and waiting for MachineDeployment and the user cluster to get deleted.

This PR also renames machines from `test-machine` to `ccm-migration-e2e`, so that we're sure machines are related to the CCM migration tests.

```release-note
NONE
```

### DIFF
--- a/pkg/test/e2e/ccm-migration/cluster.go
+++ b/pkg/test/e2e/ccm-migration/cluster.go
@@ -46,7 +46,7 @@ const (
 	clusterReadinessCheckPeriod = 10 * time.Second
 	clusterReadinessTimeout     = 10 * time.Minute
 
-	machineDeploymentName      = "ccm-test-machine"
+	machineDeploymentName      = "ccm-migration-e2e"
 	machineDeploymentNamespace = "kube-system"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/test/e2e/ccm-migration/cluster.go
```diff
@@ -211,7 +217,84 @@ func (c *ClusterJig) createCluster(cloudSpec kubermaticv1.CloudSpec) error {
 
 // CleanUp deletes the cluster.
 func (c *ClusterJig) CleanUp() error {
-	return c.SeedClient.Delete(context.TODO(), c.Cluster)
+	clusterClientProvider, err := clusterclient.NewExternal(c.SeedClient)
+	if err != nil {
+		return errors.Wrap(err, "failed to get user cluster client provider")
+	}
+
+	userClient, err := clusterClientProvider.GetClient(context.TODO(), c.Cluster)
```
xrstf: Please put the context into a `ctx` variable, we use tons of TODO()s in this function.

xrstf (changes requested):
One minor thing, otherwise let's see how this works out :)

xrstf (approved):
/approve

Issue #7450, opened 2021-08-04
Add metering download link creation to API


mate4st:
**User Story**
As a user i want to be able to Download metering csv reports on S3 via KKP API.

**Acceptance Criteria**

- KKP API includes call to generate s3 download links for a given metering report.

Issue #7448, opened 2021-08-04
Add metering to KKP Operator 


mate4st:
**User Story**
Rolling out KKP metering tool using the operator, once the enableMeteringTool field in the seed is being enabled.

**Acceptance Criteria**
- Metering runs inside every Seed Cluster
- Metering is integrated with the operator

PR #7518, opened 2021-08-17
Metering list exports endpoint

### LINKED ISSUE #7449: Add list metering reports to API
**User Story**
As a user i want to be able to list all available metering csv reports on S3 via KKP API.

**Acceptance Criteria**

- KKP API includes call to list all s3 reports generated by metering

mate4st:
Adds Endpoint to list metering exports from s3 bucket

Fixes #7449 and #7450

```release-note
NONE
```


pkg/handler/v2/metering/metering.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"k8c.io/kubermatic/v2/pkg/handler/middleware"
+	"k8c.io/kubermatic/v2/pkg/provider"
+	utilerrors "k8c.io/kubermatic/v2/pkg/util/errors"
+)
+
+func ListReportsEndpoint(userInfoGetter provider.UserInfoGetter) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+
+		// Todo(WeirdMachine): make this only available if metering is enabled
```
floreks: Will this be:
- added before the release?
- available in the CE version?
- enabled by default for the EE version?
- possible to check if this feature is enabled/disabled so that we can also make it conditionally available in the UI?

mate4st: This should be included in the upcoming release and be available for EE. 
Enabling by default is not possible as the admin needs to provide some s3 configuration.
If this feature is enabled at all might be provided by #7448

mate4st:
@floreks ptal, I pushed some fixes

pkg/provider/kubernetes/metering.go
```diff
@@ -0,0 +1,175 @@
...
+		reports, err := p.getReportsForSeed(ctx, seed)
+		if err != nil {
+			return nil, err
+		}
+		combinedReports = append(combinedReports, reports...)
+
+	}
+
+	// In-place deduplicate
+	if len(combinedReports) > 1 {
```
floreks: This part seems a bit overcomplicated and the complexity depending on sort would be in the best scenario probably `O(n * log(n))` + `O(n)` to traverse the list. I'd simply do:

```go
result := make([]string, 0)
unique := make(map[string]struct{})

for _, report := range combinedReports {
	if _, exists := unique[report]; !exists {
		result = append(result, report)
	}

	unique[report] = struct{}{}
}

return result
```

This makes the complexity `O(n)`

mate4st:
@floreks needed to rebase due to merge conflict

PR #7519, opened 2021-08-17
change vault domain to kubermatic.com

mate4st:
Changes the vault domain, used by our scripts in the hack directory, to kubermatic.com

```release-note
NONE
```


xrstf:
For all of the jobs we use in Prow, the `VAULT_ADDR` is injected by the `preset-vault` label on the Prowjobs. Maybe this is a good opportunity to remove the `export` statements from all files in `hack/ci/`? This is different from the scripts in `hack/`, which are supposed to be used by humans.

xrstf (approved):
/approve

PR #7529, opened 2021-08-18
improve test-helm-charts.sh for batch jobs

mate4st:
In a batch process of Prow the `PULL_PULL_SHA` var is not set. It can be omitted and will than be treated as `HEAD`

```release-note
NONE
```


xrstf (approved):
/approve

Thanks!

PR #7543, opened 2021-08-19
Add external KubeVirt CCM

mfranczy:
**What this PR does / why we need it**:
It's required for KubeVirt cloud provider in order to expose load balancer services from the overkube clusters

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #7480

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Users are able to expose load balancer from the overkube clusters
```

PR #7548, opened 2021-08-20
Add Anexia CCM to controlplane 

kstiehl:

Anexia has developed an own kubernetes cloud controller manager. This change adds it to the control plane and makes sure that the machine controller is started with the correct flags.

```release-note
* Anexia CCM is now added to the control plane 
```

### DIFF
--- a/hack/ci/test-helm-charts.sh
+++ b/hack/ci/test-helm-charts.sh
@@ -27,7 +27,7 @@ source hack/lib.sh
 
 # find all changed charts (the pre-kubermatic-verify-charts job ensures
 # that all updated charts also change their Chart.yaml)
-changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
+changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA:-}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
 
 for chartDirectory in $changedCharts; do
   chartName="$(basename "$chartDirectory")"
--- a/hack/ci/test-helm-charts.sh
+++ b/hack/ci/test-helm-charts.sh
@@ -27,7 +27,7 @@ source hack/lib.sh
 
 # find all changed charts (the pre-kubermatic-verify-charts job ensures
 # that all updated charts also change their Chart.yaml)
-changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA:-}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
+changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
 
 for chartDirectory in $changedCharts; do
   chartName="$(basename "$chartDirectory")"
--- a/pkg/resources/cloudcontroller/anexia.go
+++ b/pkg/resources/cloudcontroller/anexia.go
@@ -1,109 +1,128 @@
+/*
+Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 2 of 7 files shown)


hack/ci/test-helm-charts.sh
```diff
@@ -27,7 +27,7 @@ source hack/lib.sh
 
 # find all changed charts (the pre-kubermatic-verify-charts job ensures
 # that all updated charts also change their Chart.yaml)
-changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
+changedCharts=$(git diff --name-status "${PULL_BASE_SHA}..${PULL_PULL_SHA:-}" 'charts/**/Chart.yaml' | awk '{ print $2 }' | xargs dirname | sort -u)
```
xrstf: How did this change from #7529 make it into your PR?

pkg/resources/machinecontroller/deployment.go
```diff
@@ -281,6 +285,10 @@ func getFlags(clusterDNSIP string, nodeSettings *kubermaticv1.NodeSettings, cri
 			flags = append(flags, "-node-pause-image", nodeSettings.PauseImage)
 		}
 
+		if isExternal {
+			flags = append(flags, "-node-external-cloud-provider", "true")
```
xrstf: Please remove the "true", it would cause parsing errors in Go's flag handling. @xmudrii and me just recently discovered a similar bug in our machine-controller code (see `-node-csr-approver` flag above, which is also incorrect, but soon fixed in #7461)

kstiehl: done

hack/ci/test-helm-charts.sh
kstiehl: To be honest I've no idea.

kstiehl:
Not entirely sure how the failing test is related to anything

xrstf:
@kstiehl The api-e2e tests are notoriously flaky, even if you didn't touch them at all.

pkg/resources/machinecontroller/deployment.go
```diff
@@ -281,6 +285,10 @@ func getFlags(clusterDNSIP string, nodeSettings *kubermaticv1.NodeSettings, cri
 			flags = append(flags, "-node-pause-image", nodeSettings.PauseImage)
 		}
 
+		if isExternal {
+			flags = append(flags, "-node-external-cloud-provider")
+		}
+
```
mfranczy: this part is not needed, if the cloud provider is external then machine deployment webhook will set appropriate kubelet flag. Look at https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/machinecontroller/webhook.go#L65-L68

pkg/resources/machinecontroller/deployment.go
```diff
@@ -193,6 +193,10 @@ func DeploymentCreatorWithoutInitWrapper(data machinecontrollerData) reconciling
 	}
 }
 
+func isExternalCloudProvider(data machinecontrollerData) bool {
```
mfranczy: This part is not needed. If you want to always deploy the CCM as external provider, I suggest that you do it over mutation webhook like here: https://github.com/kubermatic/kubermatic/pull/7543/files#diff-8bbc7dde93249c1e7069d06767c89f4476b29eaaa92d861111a5843a73d2c6daR229-R232

There is no need to change the core deployment functions.

mfranczy (changes requested):

pkg/resources/machinecontroller/deployment.go
kstiehl: thnx for the hint
Done

pkg/resources/machinecontroller/deployment.go
kstiehl: Done

kstiehl:
source

pkg/resources/cloudcontroller/anexia.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"fmt"
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	"k8c.io/kubermatic/v2/pkg/resources/vpnsidecar"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
```
mfranczy: one more nit.. please add new lines between group of packages, following the pattern:
> - sdk libs
> - third party libs
> - k8c libs
> - k8s libs

mfranczy:
@kstiehl only one comment https://github.com/kubermatic/kubermatic/pull/7548/files#r695510150 then we are getting this PR over CI pipeline again

mfranczy:
@kstiehl and please signed-off the commit cause the dco job fails:
> dco — Commits in PR missing Signed-off-by

pkg/resources/cloudcontroller/anexia.go
xrstf: You can use gimps ( https://github.com/xrstf/gimps ) to fix it automatically ^^

mfranczy (approved):

PR #7549, opened 2021-08-20
Support metering tool in the operator

### LINKED ISSUE #7448: Add metering to KKP Operator

moadqassem:
Integrate Kubermatic metering tool in KKP.
  
Fixes #7448

**Documentation**:

```release-note
KKP metering tool integration 
```


docs/zz_generated.seed.yaml
```diff
@@ -240,6 +240,8 @@ spec:
       replicas: null
       resources: null
       tolerations: null
+  # EnableClustersMetering enables the metering on user clusters across the seed.
+  enableClustersMetering: false
```
xrstf: Do we anticipate further options coming along the way? If so, we might want to plan a structure like

```
spec:
  metering:
    enabled: true|false
```

In any case, I would not call it "ClustersMetering", but "ClusterMetering" or even "UserclusterMetering".

pkg/controller/operator/seed/resources/metering/deployment.go
```diff
@@ -0,0 +1,323 @@
...
+
+			d.Spec.Strategy.Type = appsv1.RecreateDeploymentStrategyType
+
+			d.Spec.Template.Labels = map[string]string{
+				"role":    meteringToolName,
+				"version": "1",
+			}
+			d.Spec.Template.Annotations = map[string]string{
+				"kubermatic/scrape":      "true",
+				"kubermatic/scrape_port": "2112",
```
xrstf: Please use `prometheus.io/scrape` and `prometheus.io/port`

pkg/controller/operator/seed/resources/metering/deployment.go
```diff
@@ -0,0 +1,323 @@
...
+// DeploymentCreator creates a new metering tool deployment per seed cluster.
+func DeploymentCreator(_ *kubermaticv1.Seed) reconciling.NamedDeploymentCreatorGetter {
+	return func() (string, reconciling.DeploymentCreator) {
+		//TODO: Add custom values for the metering deployment fields such as seed, interval and output-rotation.
+		return meteringToolName, func(d *appsv1.Deployment) (*appsv1.Deployment, error) {
+			d.Spec.Replicas = pointer.Int32Ptr(1)
+			d.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"role":    meteringToolName,
+					"version": "1",
```
xrstf: What did you intend to achieve by putting a version number as a selector? This selector can never be updated, as it's immutable, so the Deployment could never be reconciled to a hypothetical version 2, or am I overlooking something? In general we never put version numbers in MatchLabels.

pkg/controller/operator/seed/resources/metering/pvc.go
```diff
@@ -0,0 +1,69 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// PersistentVolumeClaimCreator creates a pvc for the metering tool where the processed data is bening saved before
```
xrstf: *being*

pkg/controller/operator/seed/resources/metering/pvc.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if err := client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: meteringDataName}, pvc); err != nil {
+		if kerrors.IsNotFound(err) {
+			pvc.ObjectMeta.Name = meteringDataName
+			pvc.ObjectMeta.Namespace = namespace
+			pvc.ObjectMeta.Labels = map[string]string{
+				"app": meteringToolName,
+			}
+
+			pvc.Spec.StorageClassName = pointer.StringPtr("kubermatic-fast")
```
xrstf: I could very well imagine having both the storageclass name and the size as options for the metering (regarding my comment with the stucture in our YAML maybe allowing some more expansion).

xrstf (commented):
Commentatio

mate4st:
If we want this as an ee only feature, i think it should be done in the operator as well. So that the operator is just not deploying metering if its ce version. WDYT?

docs/zz_generated.seed.yaml
moadqassem: Actually yes, I wanted to add this structure in the following up PR that I am working on now, which is the integration in the API, because I would have a better overview of which fields and configurations that could be added in this spec. Regarding the naming, since the metering is simply on a seed level, I guess `MeteringSpec` is convenient.

pkg/controller/operator/seed/resources/metering/deployment.go
moadqassem: This is was mainly inspired by those templates so I just did a copy and paste https://github.com/kubermatic/metering/blob/master/config/kubermatic-metering/templates/kubermatic-metering-dep.yaml#L19-L20. I can check why they were left as Kubermatic instead of Prometheus.

pkg/controller/operator/seed/resources/metering/deployment.go
moadqassem: Didn't intend anything, was a 1:1 copy from the original templates which were used originally here https://github.com/kubermatic/metering/blob/master/config/kubermatic-metering/templates/kubermatic-metering-dep.yaml#L10

pkg/controller/operator/seed/resources/metering/pvc.go
moadqassem: I totally agree with that. I was planning to have all theses config in the next PR that I am opening soon but I will adjust those fields in this PR and change the structure.

pkg/controller/operator/seed/resources/metering/deployment.go
mate4st: Those metrics are not exported anymore, so you can omit it for now.

pkg/controller/operator/seed/resources/metering/deployment.go
```diff
@@ -0,0 +1,323 @@
...
+
+			d.Spec.Template.Spec.Containers = []corev1.Container{
+				{
+					Name:    "metering",
+					Command: []string{"/usr/local/bin/kubermatic-metering"},
+					Args: []string{
+						"-workdir",
+						"/" + meteringDataName,
+						"-output-rotation",
+						"MONTHLY",
```
mate4st: Can you add the seed name as `--seed` parameter ?

moadqassem:
@WeirdMachine is the latest update for the those templates are merged?

pkg/controller/operator/seed/resources/metering/deployment.go
moadqassem: @WeirdMachine Why those metrics are not exported anymore?

mate4st: I would like to define first what kind of metric we export and why. The initial exposed data seems not relevant to me.

mate4st:
> @WeirdMachine is the latest update for the those templates are merged?

Working on it, I will ping you when it is done.

pkg/controller/operator/seed/resources/metering/deployment.go
moadqassem: Yeah that's no problem, I think we can use the Prometheus tags for now and once you decide what things need to be adjusted we can simply adjust them.

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -93,6 +93,8 @@ type SeedSpec struct {
 	MLA *SeedMLASettings `json:"mla,omitempty"`
 	// DefaultComponentSettings are default values to set for newly created clusters.
 	DefaultComponentSettings ComponentSettings `json:"defaultComponentSettings,omitempty"`
+	// Metering enables the metering on user clusters across the seed.
```
xrstf: "configures" instead of "enables" would sound better IMHO.

docs/zz_generated.seed.yaml
```diff
@@ -274,6 +274,8 @@ spec:
   # Optional: Detailed location of the cluster, like "Hamburg" or "Datacenter 7".
   # For informational purposes in the Kubermatic dashboard only.
   location: ""
+  # Metering enables the metering on user clusters across the seed.
+  metering: null
```
xrstf: Please extend https://github.com/kubermatic/kubermatic/blob/master/cmd/example-yaml-generator/main.go#L133 with an empty dummy object so the generated YAML will contain the various sub fields below `metering`.

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -416,3 +418,10 @@ type SeedMLASettings struct {
 	// Optional: UserClusterMLAEnabled controls whether the user cluster MLA (Monitoring, Logging & Alerting) stack is enabled in the seed.
 	UserClusterMLAEnabled bool `json:"user_cluster_mla_enabled,omitempty"`
 }
+
+// Metering contains all the configurations for the metering tool.
+type MeteringConfigurations struct {
+	Enabled          bool   `json:"enabled"`
+	StorageClassName string `json:"storageClassName"`
+	StorageSize      string `json:"storageSize"`
```
xrstf: Can you add a doc comment to each field to describe their default values? Maybe even, when you extend the example-yaml-generator, include the default values there so it' in the generated YAML as the default value, WDYT?

xrstf:
Is the entire feature EE only? If so, the code should probably live in https://github.com/kubermatic/kubermatic/tree/master/pkg/ee (legal department said that EE code must be in its own _directory_ and only pkg/ee has the EE license).

This would mean the CE code in `pkg/controller/operator/....` would have wrappers that either do nothing (CE) or call code in `pkg/ee/operator/seed`.

pkg/crd/kubermatic/v1/datacenter.go
moadqassem: Yeah that sound good, I will add a comment as well.

moadqassem:
> Is the entire feature EE only? If so, the code should probably live in https://github.com/kubermatic/kubermatic/tree/master/pkg/ee (legal department said that EE code must be in its own _directory_ and only pkg/ee has the EE license).
> 
> This would mean the CE code in `pkg/controller/operator/....` would have wrappers that either do nothing (CE) or call code in `pkg/ee/operator/seed`.

This is exactly what I am doing. The code in the `wrappers_ee.go` is where the logic resides and the `wrappers_ce.go` is actually doing nothing. Anyway I will just move the code in the metering package from the operator package to the ee.

moadqassem:
@xrstf ptal

xrstf (approved):
/approve

In another PR, we should add some e2e test for this.

xrstf:
This is failing pretty consistently with

> err="failed to create machine deployment: admission webhook \"machine-controller.kubermatic.io-machinedeployments\" denied the request: validation failed: failed to validate instance profile: NoSuchEntity: Instance Profile kubernetes-4bbnxtx6fp cannot be found.\n\tstatus code: 404, request id: aacd668a-4e5e-40da-b9b1-94ddb87e1460"

moadqassem:
> This is failing pretty consistently with
> 
> > err="failed to create machine deployment: admission webhook "machine-controller.kubermatic.io-machinedeployments" denied the request: validation failed: failed to validate instance profile: NoSuchEntity: Instance Profile kubernetes-4bbnxtx6fp cannot be found.\n\tstatus code: 404, request id: aacd668a-4e5e-40da-b9b1-94ddb87e1460"

@xrstf  Problem is this is not relevant to this PR

xrstf:
I would still like to know why it's only one of the e2e tests and why now and why I cannot see any errors anywhere. Neither master, seed nor usercluster controllers log anything, the MC and its webhook also don't log any failures.

moadqassem:
> I would still like to know why it's only one of the e2e tests and why now and why I cannot see any errors anywhere. Neither master, seed nor usercluster controllers log anything, the MC and its webhook also don't log any failures.

Not quite sure about that either. This PR is blocking two other potential PRs 🤷‍♂️

xrstf:
I looked long and hard, watched all logs I could imagine, checked our AWS account, could not find anything out of the order and then it worked.

I therefore conclude: cosmic rays.

moadqassem:
> I looked long and hard, watched all logs I could imagine, checked our AWS account, could not find anything out of the order and then it worked.
> 
> I therefore conclude: cosmic rays.

This has been going on for almost two days now. The thing is, before the conflicts the PR was green(today in the morning) thus I am pretty much sure, it has nothing to do with the CI being unpredictable.

PR #7558, opened 2021-08-23
Sync KubermaticConfiguration into seed clusters

xrstf:
Currently, the seed-ctrlmgr is still using ancient file formats, like the `versions.yaml` and `addons.ỳaml` and whatnot. These files have long been deprecated in favor of the KubermaticConfiguration CRD.

To finally get rid of the `kubermatic` Helm chart, we need to migrate the seed-ctrlmgr to use this new CRD instead of our homegrown files. And the first step to achieve this is to make the KubermaticConfiguration available on seed clusters. This PR does exactly that.

This PR also contains a few tweaks to the KKP operator in order to significantly reduce its reconciliations. When updating to controller-runtime 0.9, the behaviour of Patch() seemingly also changed, so the defaulting the operator did had no real effect.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -110,29 +110,13 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, seed
 	seedRecorder := r.seedRecorders[seedName]
 
 	// find the owning KubermaticConfiguration
-	configList := &operatorv1alpha1.KubermaticConfigurationList{}
-	listOpts := &ctrlruntimeclient.ListOptions{
-		Namespace: r.namespace,
+	config, err := getKubermaticConfigurationForNamespace(ctx, r.masterClient, r.namespace, log)
+	if err != nil || config == nil {
+		return err
 	}
-
-	if err := r.masterClient.List(ctx, configList, listOpts); err != nil {
-		return fmt.Errorf("failed to find KubermaticConfigurations: %v", err)
-	}
-
-	if len(configList.Items) == 0 {
-		log.Debug("ignoring request for namespace without KubermaticConfiguration")
-		return nil
-	}
-
-	if len(configList.Items) > 1 {
-		log.Infow("there are multiple KubermaticConfiguration objects, cannot reconcile", "namespace", r.namespace)
-		return nil
-	}
-
-	config := configList.Items[0]
 
 	// create a copy of the configuration with default values applied
-	defaulted, err := common.DefaultConfiguration(&config, log)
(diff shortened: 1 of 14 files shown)


pkg/controller/operator/seed/reconciler.go
```diff
@@ -110,29 +110,13 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, seed
...
-	if len(configList.Items) == 0 {
-		log.Debug("ignoring request for namespace without KubermaticConfiguration")
-		return nil
-	}
-
-	if len(configList.Items) > 1 {
-		log.Infow("there are multiple KubermaticConfiguration objects, cannot reconcile", "namespace", r.namespace)
-		return nil
+	config, err := getKubermaticConfigurationForNamespace(ctx, r.masterClient, r.namespace, log)
+	if err != nil || config == nil {
```
xmudrii, line 114: Tiny nit: I would split this into two `if`s and write some descriptive error message for `if config == nil`. Currently, if err is nil, but config is also nil, it's just going to return nil, which doesn't mean much to someone reading the logs. I guess this case shouldn't be common, but it might be helpful if it ever happens.

xmudrii (commented):
One tiny nit, otherwise LGTM.

pkg/controller/operator/seed/reconciler.go
xrstf: So I was gonna rewrite the section:

```go
// find the owning KubermaticConfiguration
config, err := getKubermaticConfigurationForNamespace(ctx, r.masterClient, r.namespace, log)
if err != nil {
	return err
}
if config == nil {
	return fmt.Errorf("no KubermaticConfiguration found in namespace %q", r.namespace)
}
```

But then I checked what `getKubermaticConfigurationForNamespace` does:

```go
if len(configList.Items) == 0 {
	log.Debug("ignoring request for namespace without KubermaticConfiguration")
	return nil, nil
}

if len(configList.Items) > 1 {
	log.Infow("there are multiple KubermaticConfiguration objects, cannot reconcile", "namespace", namespace)
	return nil, nil
}
```

So there is already a note being issued, which is why we do not need to also return an error. So I changed it to

```go
// find the owning KubermaticConfiguration
config, err := getKubermaticConfigurationForNamespace(ctx, r.masterClient, r.namespace, log)
if err != nil {
	return err
}
if config == nil {
	return nil
}
```

But when the 2nd `if` branch is used, `err` must be `nil` because of the statement above. So we can combine the two and end up with what I wrote originally :-D

xmudrii (approved):

PR #7570, opened 2021-08-24
Use KKP installer to deploy dev environment

xrstf:
This makes our stack not require Helm 2 anymore.

```release-note
NONE
```


xrstf:
@shaase-ctrl Just FYI. We deprecated the `nodeport-proxy` (NPP) Helm chart in KKP 2.15 (https://docs.kubermatic.com/kubermatic/v2.16/upgrading/2.14_to_2.15/), as the NPP is now entirely managed by the KKP Operator. We only kept the chart for folks who used the old `kubermatic` Helm chart, which is also slated for removal (the only reason we still have it in the repo is because of some technical dependencies i'm working on resolving).

With the KKP 2.18 release, the deprecation phase ends and we get rid of non-operator based installations, effectively.

shaase-ctrl:
Thanks for the notification, please make sure to add this to the release notes so that we can communicate effectively.

xmudrii:
This PR seems to be far-reaching, so we should treat it as such:

1) I think the PR title should make it clear that we dropped some charts
2) As @shaase-ctrl said, the release note should include some details about this

Otherwise, I think the PR looks good.

xmudrii (approved):

PR #7571, opened 2021-08-24
update kube-state-metrics to 2.2.0

xrstf:
v2.2.0 brings (better?) 1.22 compatibility: https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.2.0

```release-note
update kube-state-metrics to 2.2.0
```

### DIFF
--- a/charts/monitoring/kube-state-metrics/Chart.yaml
+++ b/charts/monitoring/kube-state-metrics/Chart.yaml
@@ -14,8 +14,8 @@
 
 apiVersion: v1
 name: kube-state-metrics
-version: 1.1.2
-appVersion: v2.1.1
+version: 1.2.0
+appVersion: v2.2.0
 description: Kube-State-Metrics for Kubermatic
 keywords:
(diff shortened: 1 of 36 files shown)


mate4st:
@xrstf can you please update the `cluster-role.yaml` in the chart and move `ingresses` to `networking.k8s.io` api group? At the moment kube-state-metrics seems to be broken because of this or is at least complaining.

mate4st:
Just recognized that it is entirely outdated, not only the ingresses

xrstf:
@WeirdMachine What exactly do you mean? The ksm chart has no ingress.

mate4st:
The clusterrole that is used for the kube-state-metrics service account in the chart is outdated. 
There is an updated version available at:
https://github.com/kubermatic/kubermatic/blob/master/addons/kube-state-metrics/rbac.yaml

xrstf:
Fixed

xmudrii (approved):

PR #7572, opened 2021-08-24
remove ancient unused test files

xrstf:
These files do not serve any purpose.

```release-note
NONE
```


xmudrii (approved):

xrstf:
/override pre-kubermatic-test-helm-charts

The job is confused because `charts/test/` was not actually a chart and therefore has no `Chart.yaml`. Fixing the job is pointless, as after this PR this case doesn't happen anymore. :crossed_fingers:

PR #7576, opened 2021-08-25
Fix OpenStack CCM version for 1.20+ clusters

xmudrii:

This PR addresses the following problems related to the OpenStack CCM versioning:

* The v1.21.1 image doesn't actually exist, even though the GitHub tag exists. The latest available image is v1.21.0
* The OpenStack CCM team recommends matching the CCM minor version with the Kubernetes minor version. Therefore:
    * Kubernetes 1.20 clusters will use OpenStack CCM v1.20.2
    * Kubernetes 1.21 clusters will use OpenStack CCM v1.21.0 (upgraded from 1.20.2)
    * Kubernetes 1.22 clusters will use OpenStack CCM v1.22.0

```release-note
Use OpenStack CCM v1.21.0 for Kubernetes v1.21 clusters, and CCM v1.22.0 for Kubernetes v1.22 clusters
```

### DIFF
--- a/pkg/resources/cloudcontroller/openstack.go
+++ b/pkg/resources/cloudcontroller/openstack.go
@@ -141,20 +141,18 @@ func getOSFlags(data *resources.TemplateData) []string {
 	return flags
 }
 
-const latestOpenstackCCMVersion = "1.21.1"
-
 func getOSVersion(version semver.Semver) (string, error) {
 	switch version.Minor() {
 	case 19:
 		return "1.19.2", nil
 	case 20:
-		return latestOpenstackCCMVersion, nil
+		return "1.20.2", nil
 	case 21:
-		return latestOpenstackCCMVersion, nil
+		return "1.21.0", nil
 	case 22:
-		return latestOpenstackCCMVersion, nil
+		fallthrough
 	default:
-		return latestOpenstackCCMVersion, nil
+		return "1.22.0", nil
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/resources/test/fixtures/deployment-openstack-1.20.0-openstack-cloud-controller-manager-externalCloudProvider.yaml
```diff
@@ -65,7 +65,7 @@ spec:
         - '{"command":"/bin/openstack-cloud-controller-manager","args":["--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig","--v=1","--cloud-config=/etc/kubernetes/cloud/config","--cloud-provider=openstack"]}'
         command:
         - /http-prober-bin/http-prober
-        image: docker.io/k8scloudprovider/openstack-cloud-controller-manager:v1.21.1
+        image: docker.io/k8scloudprovider/openstack-cloud-controller-manager:v1.20.2
```
xmudrii, line 68: NB: this is not a downgrade. It seems like we bumped this to v1.21 by mistake when adding support for Kubernetes 1.22.

pkg/resources/cloudcontroller/openstack.go
```diff
@@ -141,16 +141,16 @@ func getOSFlags(data *resources.TemplateData) []string {
 	return flags
 }
 
-const latestOpenstackCCMVersion = "1.21.1"
+const latestOpenstackCCMVersion = "1.22.0"
```
xrstf: Do we even need the constant anymore? IIRC it was only added because a linter complained.

xmudrii: Let's try without it and see if linter is going to complain again.

xrstf: oh i didn't see it in the diff that the constant was still used twice.

xrstf (approved):
/approve

PR #7579, opened 2021-08-26
Adds telemetry deploy

imharshita:

Installs Telemetry Agents with Kubermatic installer

Fixes https://github.com/kubermatic/telemetry/issues/38

```release-note
Adds telemetry charts in kubermatic installer with flag --disable-telemetry to disable it
```

### DIFF
--- a/charts/telemetry/values.yaml
+++ b/charts/telemetry/values.yaml
@@ -37,7 +37,7 @@ telemetry:
 # uuid is the unique identifier of the client where the agent is running.
 # This field is required and will print an error message when that entry is missing.
 # You can generate uuid using command uuidgen on your linux machine
-  uuid: ""
+  uuid: "47cbc08b-a1a6-478c-afd6-f303accae7b9" #<--- you HAVE TO set this
 
   resources:
     limits:
--- a/charts/telemetry/values.yaml
+++ b/charts/telemetry/values.yaml
@@ -19,19 +19,19 @@ telemetry:
   kubermaticAgent:
     image:
       repository: quay.io/kubermatic/telemetry-kubermatic-agent
-      pullPolicy: IfNotPresent
+      pullPolicy: Always
       tag: "v0.1.0"
 
   kubernetesAgent:
     image:
       repository: quay.io/kubermatic/telemetry-kubernetes-agent
-      pullPolicy: IfNotPresent
+      pullPolicy: Always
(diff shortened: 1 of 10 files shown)


jiachengxu (approved):
/lgtm
/approve

charts/telemetry/values.yaml
```diff
@@ -19,19 +19,19 @@ telemetry:
   kubermaticAgent:
     image:
       repository: quay.io/kubermatic/telemetry-kubermatic-agent
-      pullPolicy: IfNotPresent
+      pullPolicy: Always
```
xrstf, line 22: Why do we always need to pull these? If this is purely to make development easier on our dev system, please instead override this by adjusting the `europe-west3-c-values.yaml` in https://vault.loodse.com/ui/vault/secrets/dev/show/seed-clusters/dev.kubermatic.io (you couldalso set the tag to "master" in Vault, so that we, for development, don't have to override the 0.1.0 again and again and again with new images).

charts/values.example.ce.yaml
```diff
@@ -60,3 +60,9 @@ dex:
 minio:
   storeSize: '500Gi'
   storageClass: minio-hdd
+
+telemetry:
+# uuid is the unique identifier of the client where the agent is running.
+# This field is required and will print an error message when that entry is missing.
+# You can generate uuid using command uuidgen on your linux machine
```
xrstf: Please indent the comments like the data, i.e. 2 spaces in this case. Same for the EE example file.

xrstf (commented):
One question, one comment, otherwise LGTM.

charts/telemetry/values.yaml
imharshita: Telemetry charts will be released with Kubermatic installer but as telemetry is itself dependent on kubermatic , we will have to update the telemetry code with new kubermatic release after the kubermatic release. 
It is kind of a deadlock situation.
Hence, as a workaround, after the release in order to update telemetry on client side we thought of building/tagging the new telemetry image with same tag as mentioned in chart  right now. 
That's only possible if the imagePullPolicy is set to Always.
Please let us know, if you can think of  an alternate approach :)

xrstf: That makes sense

xrstf (approved):
/approve

PR #7584, opened 2021-08-26
remove priorityClassName from metric-server deployment

PratikDeoghare:
This was added by mistake and now creating pods fail with following error. 

> Warning  FailedCreate  5m37s (x18 over 16m)  replicaset-controller  Error creating: insufficient quota to match these scopes: [{PriorityClass In [system-node-critical system-cluster-critical]}]

The priority class doesn't make sense for us either way, as the metrics-server runs on the seed. Its existence however is certainly not critical for the _seed_ cluster.

**Documentation**:

Discussion https://kubermatic.slack.com/archives/C01FJ10TBM2/p1629965851002700

```release-note
NONE
```


xrstf (approved):
/approve

PR #7590, opened 2021-08-26
Block vSphere on K8s 1.22

### LINKED ISSUE #7578: Block Kubernetes cluster 1.22 on vSphere
**User Story**
As reported in the CSI-driver documentation: 

> If `CSIMigrationvSphere` feature gate is enabled, users should not upgrade to Kubernetes v1.22. vSphere CSI Driver does not support Kubernetes v1.22 yet because it uses v1beta1 CRD APIs. Support for v1.22 will be added at a later release. Check the following document for supported Kubernetes releases for a given vSphere CSI Driver version.

Since #7280 introduced the external CCM for vSphere, and #7461 added the support for K8s v1.22, we need to avoid vSphere clusters using external CCM using K8S 1.22.
Below there are the needed steps to achieve this goal:
* Block upgrades for existing clusters to v1.22 if the `externalCloudProvider` feature is enabled
* Block CCM/CSI migration if the cluster version is already 1.22
* Block users to create 1.22 clusters on vSphere

**Acceptance criteria**
Creating clusters on vSphere with K8s v1.22 must not be possible.

mlavacca:
Since the necessity to block Kubernetes v1.22 on vSphere emerged under certain specific scenarios, this PR aims at implementing a generally available mechanism to block unwanted Kubernetes versions without the need to write custom code in the API server. The `extra-files` secret (where `versions.yaml` and `updates.yaml` reside) has been improved by introducing a new file: `provider-incompatibilities.yaml`, whose structure is the following:
```yaml
providerIncompatibilities:
- condition: always
  operation: CREATE
  provider: vsphere
  version: 1.22.*
- condition: externalCloudProvider
  operation: UPDATE
  provider: vsphere
  version: 1.22.*
```
Below are the parameters' details:
* condition: the condition that the cluster must match to trigger the incompatibility (e.g., the cluster using the external CCM)
* operation: the operation to be applied in (creation of a new cluster or update of an existing one)
* provider: the provider to which to apply the block
* version: the Kubernetes version to block (wildcards are allowed)

The conditions can be improved in the future, every field about the cluster or the datacenter is allowed (it's just a matter of a couple of lines of code).
Adding and removing constraints is very simple by using this mechanism: when in the future we'll allow using 1.22 clusters on vSphere, it will be enough to delete the above rules from the `defaults.go` [file](https://github.com/kubermatic/kubermatic/blob/06fa2a02e3fb09b70888a1795d24373f41199d6f/pkg/controller/operator/common/defaults.go#L265-L278), without dealing with ad-hoc conditions embedded in the core code.

Fixes #7578 

**Documentation**:

```release-note
NONE
```


charts/kubermatic/static/master/provider-incompatibilities.yaml
```diff
@@ -0,0 +1,13 @@
+# This file has been generated using hack/update-kubermatic-chart.sh, do not edit.
```
xrstf, line 1: I appreciate the effort, but we do not update this chart anymore. It's soon going to be deleted entirely, so you can skip everything needed to generate this file.

xrstf: NB: We _still_ required the seed-ctrlmgr to read data from files (instead of the `KubermaticConfiguration`), but work is being done to change that (i.e. we do not want to have a versions.yaml, updates.yaml etc. anymore, all we want is one single KubermaticConfiguration CRD and that's it).

So we might still need code to generate this file, but we do not update this Helm chart anymore. The KKP operator must create the file instead.

codegen/kubermatic_operator/main.go
```diff
@@ -64,6 +64,13 @@ func main() {
 	}
 
 	writeYAML(updatesYAML, "charts/kubermatic/static/master/updates.yaml")
+
+	providerIncompatibilitiesYAML, err := common.CreateProviderIncompatibilitiesYAML(versionCfg)
+	if err != nil {
+		log.Fatalf("Failed to encode updates as YAML: %v", err)
+	}
+
+	writeYAML(providerIncompatibilitiesYAML, "charts/kubermatic/static/master/provider-incompatibilities.yaml")
```
xrstf, line 73: i.e. this can go away

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -334,6 +338,13 @@ type Update struct {
 	AutomaticNodeUpdate *bool `json:"automaticNodeUpdate,omitempty,omitgenyaml"`
 }
 
+type Incompatibility struct {
+	Provider  kubermaticv1.ProviderType        `json:"provider,omitempty"`
+	Version   string                           `json:"version,omitempty"`
+	Condition version.IncompatibilityCondition `json:"condition,omitempty"`
+	Operation version.OperationType            `json:"operation,omitempty"`
```
xrstf: Please document these fields and then regenerate the docs.

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -312,6 +313,9 @@ type KubermaticVersioningConfiguration struct {
 	// updates as well. 'automaticNodeUpdate: true' implies 'automatic: true' as well,
 	// because Nodes may not have a newer version than the controlplane.
 	Updates []Update `json:"updates,omitempty"`
+
+	// ProviderIncompatibilities lists all the kubernetes version prohibited for a given provider. Wildcards are allowed, e.g. "1.22.*".
```
xrstf: > Wildcards are allowed, e.g. "1.22.*".

That is a statement that belongs on the "Version" field in the `Incompatibiliy` struct.

pkg/handler/v1/common/common.go
```diff
@@ -75,8 +75,9 @@ type OIDCConfiguration struct {
 // UpdateManager specifies a set of methods to handle cluster versions & updates
 type UpdateManager interface {
 	GetVersions(string) ([]*version.Version, error)
+	GetVersionsV2(string, kubermaticv1.ProviderType, ...version.IncompatibilityCondition) ([]*version.Version, error)
```
xrstf: Why did you decide to do it this way? Why not simply extend the existing `GetVersions` function? You were able to extend the `GetPossibleUpdates` already.

pkg/handler/v1/metering/metering.go
```diff
@@ -25,9 +25,10 @@ import (
 	"strconv"
 
 	"github.com/gorilla/mux"
-	"k8s.io/apimachinery/pkg/api/resource"
 
 	k8cerrors "k8c.io/kubermatic/v2/pkg/util/errors"
+
+	"k8s.io/apimachinery/pkg/api/resource"
```
xrstf: :heart:

pkg/handler/v1/common/common.go
mlavacca: I did in this way because GetVersions is consumed by [this](https://github.com/kubermatic/kubermatic/blob/489c811eebf2419af4e83c8b1607fcd5a360daba/pkg/handler/routes_v1.go#L1668-L1689) endpoint, which has no way to fetch provider information, therefore must be replaced. I think it shouldn't be deleted before the UI has been migrated to use the new endpoint instead ([this](https://github.com/kubermatic/kubermatic/blob/1cd656ec7264744f3752b832273a43cf497ab211/pkg/handler/v2/routes_v2.go#L3554-L3579) one). Once The new endpoint will be correctly used by the UI, I think we can get rid of the old endpoint and change the getVersions function signature as you suggested.

pkg/handler/v1/metering/metering.go
mlavacca: Since when I discovered gimps, my life has changed :grin:

pkg/crd/operator/v1alpha1/configuration.go
mlavacca: right, gonna change it

pkg/handler/v2/version/version.go
```diff
@@ -0,0 +1,98 @@
...
+			return nil, errors.NewBadRequest("invalid request")
+		}
+		err := req.Validate()
+		if err != nil {
+			return nil, errors.NewBadRequest(err.Error())
+		}
+
+		versions, err := updateManager.GetVersionsV2(req.Type, kubermaticv1.ProviderType(req.ProviderName))
+		if err != nil {
+			return nil, errors.NewBadRequest(err.Error())
```
xrstf: If we already validated the request beforehand, what possible error conditions can happen here and are they really the user's fault (HTTP 4xx)?

pkg/handler/v2/version/version_test.go
```diff
@@ -0,0 +1,207 @@
...
+		existingKubermaticObjs []ctrlruntimeclient.Object
+		apiUser                apiv1.User
+		provider               kubermaticv1.ProviderType
+		versions               []*version.Version
+		updates                []*version.Update
+		incompatibilities      []*version.ProviderIncompatibility
+		wantVersions           []*apiv1.MasterVersion
+	}{
+		{
+			name: "upgrade available and incompatibility with another provider",
```
xrstf: you probably meant to write "and _compatible_ with"

pkg/provider/types.go
```diff
@@ -752,6 +752,9 @@ type PresetProvider interface {
 	SetCloudCredentials(userInfo *UserInfo, presetName string, cloud kubermaticv1.CloudSpec, dc *kubermaticv1.Datacenter) (*kubermaticv1.CloudSpec, error)
 }
 
+type VersionProvider interface {
+}
+
```
xrstf: A quick ctrl-f in Firefox did not show any spots where this is being used. Am I wrong? :)

pkg/resources/cloudcontroller/vsphere.go
```diff
@@ -145,7 +145,7 @@ func getVsphereCPIVersion(version semver.Semver) (string, error) {
 	case 21:
 		return latestVsphereCPIVersion, nil
 	case 22:
-		return latestVsphereCPIVersion, nil
+		return "", fmt.Errorf("kubernetes v1.22 not supported by external CCM")
```
xrstf: This is kind of a conflict for me. One the one hand, we make it configurable what is supported and what not, but this hardcodes it again to the KKP codebase. So which wins?

Can we rewrite this to make use of the configured incompatibilities? WDYT @xmudrii ? Technically this then also, in the future, affects other providers where we currently have hardcoded `switch` statements that technically would need to be moved to this new configuration file, right?

pkg/version/helpers.go
```diff
@@ -0,0 +1,32 @@
...
+
+package version
+
+import (
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+)
+
+func CreateAndCheckNotConstrained(baseVersion *semver.Version, version string) (bool, error) {
```
xrstf: What does this function create, besides a temporary variable that I, the caller, do not care in the slightest about? ^^ This could be shortened to just `IsConstrained`, no?

pkg/version/manager.go
```diff
@@ -23,19 +23,43 @@ import (
...
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
 	"k8c.io/kubermatic/v2/pkg/validation/nodeupdate"
 )
 
+type IncompatibilityCondition string
+
+const (
+	AlwaysCondition                IncompatibilityCondition = "always"
+	ExternalCloudProviderCondition IncompatibilityCondition = "externalCloudProvider"
```
xrstf: Please document these and the other new constants. I am not even sure what exactly these two control ^^

xrstf (commented):
Very nice work, yet I have a few questions and comments :)

pkg/provider/types.go
mlavacca: No, you are right, probably it is garbage left from the previous idea that I had.

pkg/handler/v1/common/common.go
xrstf: I see. I am so scared that the two versions will live forever and we will never clean up... so scared.. can you calm me down by promising you will clean it up? ^^

mlavacca: Once the UI part will be implemented I'll clean it up, you have my word :grin:

pkg/resources/cloudcontroller/vsphere.go
mlavacca: I agree that having this hardcoded switch is in conflict with the goal of this PR. We can use the incompatibilities file also to check these kinds of conditions without too much effort if it worths it.

pkg/version/helpers.go
mlavacca: Definitely a smarter name, yes.

pkg/handler/v2/version/version.go
mlavacca: Actually, if an error occurs in that situation, there is some problem with the versions in the extra-files, therefore I replaced 400 with 500.

pkg/handler/common/cluster.go
```diff
@@ -1047,6 +1053,10 @@ func getSSHKey(ctx context.Context, userInfoGetter provider.UserInfoGetter, sshK
 }
 
 func convertInternalCCMStatusToExternal(cluster *kubermaticv1.Cluster, datacenter *kubermaticv1.Datacenter) apiv1.ExternalCCMMigrationStatus {
+	if check, _ := version.CheckUnconstrained(cluster.Spec.Version.Version, "1.22.*"); !check {
+		return apiv1.ExternalCCMMigrationUnsupported
+	}
```
xmudrii: Does this apply to all providers or only to vSphere? I think allowing CCM/CSI migration for OpenStack 1.22 clusters should be fine.

NB: The CCM/CSI migration actually might not work properly since 1.21 because some feature gates got changed. I can take a look at this later.

pkg/handler/v1/common/common.go
xmudrii: Would it make sense to add a comment shortly describing this, maybe even linking to this thread? It might help if someone is reading this code later. :slightly_smiling_face:

pkg/resources/cloudcontroller/vsphere.go
xmudrii: Yes, I agree as well.

pkg/version/manager.go
```diff
@@ -23,19 +23,49 @@ import (
...
+)
+
+// OperationType is the type defining the operations triggering the compatibility check (CREATE or UPDATE)
+type OperationType string
+
+const (
+	// CreateOperation represents the creation of a new cluster
+	CreateOperation OperationType = "CREATE"
+	// UpdateOperation represents the update of an already existing cluster
+	UpdateOperation OperationType = "UPDATE"
```
xmudrii: I find calling this `UPDATE` slightly confusing. As far as I get from the code, this is used to block cluster upgrades. Therefore, it might make sense to call it `UPGRADE` instead.

xrstf:
I

/approve

and @xmudrii can give their /lgtm

Also, I

/retest

mlavacca:
> I
> 
> /approve
> 
> and @xmudrii can give their /lgtm
> 
> Also, I
> 
> /retest

@xrstf  I'm implementing this part:
> Can we rewrite this to make use of the configured incompatibilities? WDYT @xmudrii ? Technically this then also, in the future, affects other providers where we currently have hardcoded switch statements that technically would need to be moved to this new configuration file, right?

should be ready in a while

charts/kubermatic/static/master/provider-incompatibilities.yaml
mlavacca: @xrstf I had to recreate the file since it is used (along `with versions.yaml` and `updates.yaml`) by the `run-api.sh` script. If we don't statically provide this file to it, that test script is unusable.

pkg/resources/cloudcontroller/vsphere.go
mlavacca: @xmudrii @xrstf I used the incompatibilities mechanism to block also the migration, this ad-hoc part for vSphere has been deleted.

pkg/handler/common/cluster.go
mlavacca: Solved by using the incompatibilities mechanism

xmudrii (approved):
/lgtm
/approve

xrstf:
/hold

temporarily. I'm doing something on CI. Nothing to worry about.

mlavacca:
@xmudrii LGTM lost after rebase (there was a conflict). Could you lgtm again please?

xmudrii (approved):

PR #7603, opened 2021-08-27
Add MLAAdminSetting CRD

rastislavs:
Adds MLAAdminSetting CRD representing cluster-specific administrator settings for KKP user cluster MLA (monitoring, logging & alerting) stack.

Fixes #

**Documentation**:

```release-note
Add MLAAdminSetting CRD.
```


pkg/crd/kubermatic/v1/mla_admin_settings.go
```diff
@@ -0,0 +1,90 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec MLAAdminSettingSpec `json:"spec"`
+}
+
+// MLAAdminSettingSpec specifies the cluster-specific administrator settings
+// for KKP user cluster MLA (monitoring, logging & alerting) stack.
+type MLAAdminSettingSpec struct {
+	// Cluster is the reference to the user cluster whose MLA settings are defined in this object.
+	Cluster corev1.ObjectReference `json:"cluster"`
```
rastislavs: Just realized we don't need this Cluster reference, as this is a namespaces CRD and the namespace defines the user cluster. I'm going to remove it (FYI @aborilov)

jiachengxu: I think this depends on if your controller logic needs to interact with `Cluster` objects, if yes, then better have a Cluster reference, because deriving cluster name from namespace name may break if we use a different naming pattern for namespaces in the future.

rastislavs: OK will keep it for now, I don't think I need this reference for my part but @aborilov let me know whether you need it. Also not sure whether we need full object reference or just a cluster ID.

jiachengxu: Yeah, you are right, we can just use the `Name` in the ObjectReference, and let other fields be empty, and we assgin them values if we need in the future

aborilov: I always can find Cluster from a namespace or using just `Name`. I think `ObjectReference` is useful when we need `MLAAdminSetting` to be enquired when `Cluster` object changed, but I don't think we need that in our case

rastislavs: OK, let's keep it

jiachengxu (approved):
/approve
/lgtm

mfranczy (approved):
/approve
/lgtm

PR #7604, opened 2021-08-30
Make user cluster MLA components resources requirements configurable

### LINKED ISSUE #7537: Configurable resource req/limits for user cluster MLA components
The resource requests & limits of user cluster Prometheus and Promtail should be configurable, so that they can meet any scale requirements.

jiachengxu:
This PR makes user cluster MLA components resources configurable via KKP API

Fixes #7537 

**Documentation**:

```release-note
NONE
```


xmudrii:
/assign
/hold
this might not work as expected, I'll double-check

cmd/user-cluster-controller-manager/main.go
```diff
@@ -265,6 +265,7 @@ func main() {
 			MLAGatewayURL:                runOp.mlaGatewayURL,
 			PrometheusScrapeConfigPrefix: runOp.prometheusScrapeConfigPrefix,
 		},
+		runOp.clusterName,
```
xmudrii, line 268: The `-cluster-name` flag is provided to the UCCM only if the cluster is being CCM/CSI migrated. Please see this: https://github.com/kubermatic/kubermatic/blob/880595020ad5c5853f7cf416e53a548bfbcc162f/pkg/resources/usercluster/deployment.go#L179-L182

You should:

1. Change that part to unconditionally provide the `-cluster-name` flag
2. Change the flag validation in the UCCM to require the flag to be provided unconditionally: https://github.com/kubermatic/kubermatic/blob/1023c026298f124a5e70ebbeba4d1070ab347b44/cmd/user-cluster-controller-manager/main.go#L149-L151

pkg/controller/user-cluster-controller-manager/resources/resources/mla/promtail/daemonset.go
```diff
@@ -229,6 +230,22 @@ func DaemonSetCreator() reconciling.NamedDaemonSetCreatorGetter {
...
+			}
+			var err error
+			if overrides == nil {
+				err = resources.SetResourceRequirements(ds.Spec.Template.Spec.Containers, defResourceRequirements, nil, ds.Annotations)
+			} else {
+				overridesRequirements := map[string]*corev1.ResourceRequirements{
+					containerName: overrides.DeepCopy(),
+				}
+				err = resources.SetResourceRequirements(ds.Spec.Template.Spec.Containers, defResourceRequirements, overridesRequirements, ds.Annotations)
+			}
```
xmudrii: Why do you have this check instead of just calling `SetResourceRequirements`? You could do something like this, which is a little bit simpler and easier to read:

```suggestion
			var overridesRequirements map[string]*corev1.ResourceRequirements
			if overrides != nil {
				overridesRequirements = map[string]*corev1.ResourceRequirements{
					containerName: overrides.DeepCopy(),
				}
			}
			err = resources.SetResourceRequirements(ds.Spec.Template.Spec.Containers, defResourceRequirements, overridesRequirements, ds.Annotations)
```

cmd/user-cluster-controller-manager/main.go
jiachengxu: Thanks, fixed

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -190,6 +193,27 @@ func Add(
 		}
 	}
 
+	// Watch cluster if user cluster MLA is enabled so that controller can get resource requirements for user cluster MLA components.
+	if r.userClusterMLA.Monitoring || r.userClusterMLA.Logging {
+		seedWatch := &source.Kind{Type: &kubermaticv1.Cluster{}}
+		if err := seedWatch.InjectCache(seedMgr.GetCache()); err != nil {
+			return fmt.Errorf("failed to inject cache in seed cluster watch for cluster: %w", err)
```
xrstf, line 200: Can you leave a comment explaining why you had to inject the cache? I might be dense, but I can't quite explain it off the top of my head.

jiachengxu: Hi @xrstf,
I actually don't know why I need to inject the cache, I just copied it from few lines above this function: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/user-cluster-controller-manager/resources/controller.go#L185-L187,
and we InjectCache for every type of seed watch.

xmudrii (approved):
/lgtm
/approve
/hold cancel

xrstf:
/hold

temporarily

PR #7606, opened 2021-08-30
[WIP] Platform extensions proposal: Extension operators

dermorz:

This PR introduces an alternate implementation proposal for KKP platform extensions. Compared to the first Helm based approach this is based on operators managing the extensions and both approaches feature some complementary concepts.

This draft is in a pretty early stage and its primary purpose is to serve as a base of discussion and evaluation.

Any input or thoughts are helpful.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/platform-extension-implementation-extension-operators.md
+++ b/docs/proposals/platform-extension-implementation-extension-operators.md
@@ -59,13 +59,13 @@ managed by [OLM](https://olm.operatorframework.io/).
 
 Reconciles `Extension` CRs and takes care of registering Operators for
 Extensions in the Extension catalog, making them available for installation.
-It basically translates `Extensions` to CRs needed by the OML operator either
+It basically translates `Extensions` to CRs needed by the OLM operator either
 wrapping OLM's catalog operator or implementing the necessary parts to fit our
 requirements.
 
 ### Extension Controller
 
-The extension controller reconciles ExtensionOperator CRs. It runs in the KPP
+The extension controller reconciles ExtensionOperator CRs. It runs in the KKP
 controller manager in the seed cluster and manages OLM CRs on all user
 clusters. It acts as a wrapping layer around the OLM operator which taking care
 of the whole lifecycle of application operators.
--- a/docs/proposals/platform-extension-implementation-extension-operators.md
+++ b/docs/proposals/platform-extension-implementation-extension-operators.md
@@ -1,0 +1,86 @@
+# Platform Extension Implementation: Extension Operators
+
+**Author**: Moritz Bracht
+
+**Status**: Draft proposal.
+
+## Goals
+
+This proposal explores an operator based approach for KKP extensions.
+
+## Motivation and Background
+
+An implementation of platform-extension based on Helm charts provides flexible
+yet simplistic way to manage the lifecycle of platform-extensions.
+
+In case of more complex extensions that can not simply be installed with Helm
+([like KubeVirt](https://kubevirt.io/user-guide/operations/installation/), the
+extensions could be managed by their own operators.
+
+To cover simple Helm based extensions either a Helm operator could be installed
+or simple operators could be [generated directly from Helm
+charts](https://sdk.operatorframework.io/docs/building-operators/helm/tutorial/).
+
+[Check this for requirements](./platform-extensions.md#requirements-for-implementations)
+
+## Implementation
+
+Every extension is managed by its own operator. These operators are being
+managed by [OLM](https://olm.operatorframework.io/).
+
+### Main Concept Points
+
+* Extensions provided by KKP devs
+  * Extension operators are bundled and provided via OLM catalog
+* Extensions provided by users
+  * Users can create and bundle their own extension operators and extend the
+	catalog
+* Life cycle management for extensions comes with OLM
+
+### Extension Registration
+
+`Extension` 
+* Reference to a
+  [ClusterServiceVersion](https://olm.operatorframework.io/docs/concepts/crds/clusterserviceversion/)
+* Reference to a
+  [CatalogSource](https://olm.operatorframework.io/docs/concepts/crds/catalogsource/)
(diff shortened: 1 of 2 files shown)


shaase-ctrl:
Why would we need OLM?

docs/proposals/platform-extension-implementation-extension-operators.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Implementation
+
+Every extension is managed by its own operator. These operators are being
+managed by [OLM](https://olm.operatorframework.io/).
+
+### Main Concept Points
+
+* Extensions provided by KKP devs
+  * Extension operators are bundled and provided via OLM catalog
```
themue: Small hint on how the OLM catalog is realized needed. So that a reader has not dive into OLM in the first step.

docs/proposals/platform-extension-implementation-extension-operators.md
```diff
@@ -0,0 +1,86 @@
...
+* Extensions provided by users
+  * Users can create and bundle their own extension operators and extend the
+	catalog
```
themue: Join lines?

docs/proposals/platform-extension-implementation-extension-operators.md
```diff
@@ -0,0 +1,86 @@
...
+* Life cycle management for extensions comes with OLM
```
themue: How transparent is OLM to integrate into KKP? Is it always an individual tool or does it provide an API we can address by KKP?

themue (commented):
Sounds good so far. Have to take a look into OLM. 

Open questions for me:

1. What happens if cluster user installs an OLM too? How are our operators and those of the "external" OLM separated (for managing the operators)?
2. Same questions when using Helm: How are by us Helm-installed extensions separated from external Helm-installed software?

dermorz:
@shaase-ctrl wrote:
> Why would we need OLM?

The proposed idea is to have an operator for each extension. OLM comes with built-in life-cycle management that can be tied to the Kubernetes version and also offers an easy way for customers to extend the catalog of operators.

So OLM as a "package manager" for operators that manage the life-cycles for extensions.

dermorz:
@themue write:
> Open questions for me:
> 
> 1. What happens if cluster user installs an OLM too? How are our operators and those of the "external" OLM separated (for managing the operators)?

One option would be to have OLM operators by KKP run on the seed clusters and have them install the extensions onto the user clusters. If the user installs OLM operators on their cluster, they would have no access to the CRs to the KKP-OLM and run logically separated, no?

> 2. Same questions when using Helm: How are by us Helm-installed extensions separated from external Helm-installed software?

You mean it could lead to an endless reconciliation loop if 2 rivaling operators would want to install the same piece of software into the same namespace but with different configurations? That would be a problem then, yes. Since this question is aimed towards all currently proposed ideas, we might want to talk about it in another refinment meeting.

docs/proposals/platform-extension-implementation-extension-operators.md
dermorz: As OLM is a direct extension of Kubernetes, its resources are directly available. As of KKP integration, that part of the proposal is still pretty much open. I would see the Extension Catalog Controller and the Extension Controller as a part of the KKP controller manager. They would reconcile KKP Extension related CRs and generate CRs reconciled by the OLM operators.

This sounds kind of layered and indirected, but maybe can result in a clean interface to benefit from OLM's toolkit.

docs/proposals/platform-extension-implementation-extension-operators.md
dermorz: Sure, I will extend this section and also the one of the Extension Catalog Controller.

sankalp-r:
> Sounds good so far. Have to take a look into OLM.
> 
> Open questions for me:
> 
> 1. What happens if cluster user installs an OLM too? How are our operators and those of the "external" OLM separated (for managing the operators)?
> 2. Same questions when using Helm: How are by us Helm-installed extensions separated from external Helm-installed software?

@themue 
Accirding to our proposal,Helm-manager will only act on Extension related CRs, and it will not act on externally installed helm charts.

docs/proposals/platform-extension-implementation-extension-operators.md
```diff
@@ -0,0 +1,96 @@
...
+## Goals
+
+This proposal explores an operator based approach for KKP extensions.
+
+## Motivation and Background
+
+An implementation of platform-extension based on Helm charts provides flexible yet simplistic way to
+manage the lifecycle of platform-extensions.
+
+In case of more complex extensions that can not simply be installed with Helm ([like
```
mfranczy, line 16: KubeVirt could be easily installed with helm chart, but deleting KubeVirt is a problem, there is a need for some customization.
> Note: The apiservice and the webhookconfigurations need to be deleted manually due to a bug.

docs/proposals/platform-extension-implementation-extension-operators.md
```diff
@@ -0,0 +1,96 @@
...
+KubeVirt](https://kubevirt.io/user-guide/operations/installation/), the extensions could be managed
```
mfranczy, line 17: we would end up with having an operator that manges the projects operator. we should avoid that

mfranczy:
To be honest I would avoid following the OLM approach for extensions. 

Helm is a general purpose manager, you can have collection of k8s manifests so this can be anything, entry level to build a new extension is low.

OLM solves different use case, it's operator focused thus we are tight to this what OLM offers and bring higher complexity to build your own addon (writing/building a new operator - using helm based operator you still have to build a new operator).

dermorz:
I think, you are right @mfranczy. This implementation proposal was merely meant as exploration of possibilities and now I think this proposal is one step too far for that. Actually getting a competitive overview in this problem space (https://github.com/kubermatic/kubermatic/pull/7819) makes more sense than this proposal.

PR #7636, opened 2021-09-01
Fix inconsistent seeds update

moadqassem:
At the moment we are updating the metering configurations using the seed client for each seed. This could lead into a precondition failure where the comparison for updating the cached seed with the updated one. This PR would simply fetch the seed object from the master cluster using the master client and then carry out the update. Then the seedsync controller would sync the seeds between master and seed clusters. 

Fixes #

**Documentation**:

```release-note
None
```

### DIFF
--- a/cmd/kubermatic-api/main.go
+++ b/cmd/kubermatic-api/main.go
@@ -75,6 +75,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 	"k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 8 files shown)


WeirdMachine:
Why is the master client not wrapped in a provider?

lsviben:
> Why is the master client not wrapped in a provider?

Agree, it would be cleaner to wrap it in a provider, and pass it through to the endpoint as a provider.

moadqassem:
> Why is the master client not wrapped in a provider?

Because this is only being using to update the configuration not all the metering related fields. Thus what would wrapping the master client be used for? It will lead to inconsistency on a code level, as you have some parts of metering using the master client and others don't. Plus you need wrapping for that provider to handle ee and ce areas. IMHO this is a complexity and abstraction that has no value.

lsviben:
Discussed it with @moadqassem , the approach is fine for now, as we are just passing the client, and not a provider. 

In the future, we could incorporate this into a seedProvider, as the metering endpoint is changing Seeds, but we would need to add EE wrapping which could take some time.

/lgtm
/approve

PR #7637, opened 2021-09-02
update OWNERS files

xrstf:
* irozzo-1A has left the organization
* removed "ownership" of TODOs
* removed sig-security, as the SIG was introduced, got ownership of some directories but then never really acted; sig-api was and now is the original owner and was maintaining the code anyway, so I remove sig-security from th OWNERS files

```release-note
NONE
```


moadqassem (commented):
/approve
/lgtm

PR #7645, opened 2021-09-02
Use release version for metering images and fix cronjob name

simon-wessel:
**What this PR does / why we need it**: Use minor version v0.5 instead of development tag for metering images.
Also the schedule for the metering reports has been changed from monthly to weekly in https://github.com/kubermatic/kubermatic/pull/7612, but the CronJob name is still called monthly.

```release-note
NONE
```

### DIFF
--- a/pkg/ee/metering/cronjob.go
+++ b/pkg/ee/metering/cronjob.go
@@ -114,8 +114,9 @@ mc mirror --newer-than "65d0h0m" s3/$S3_BUCKET /metering-data || true`,
 
 			job.Spec.JobTemplate.Spec.Template.Spec.Containers = []corev1.Container{
 				{
-					Name:  "kubermatic-metering-report",
-					Image: "quay.io/kubermatic/metering:v0.5",
+					Name:            "kubermatic-metering-report",
+					Image:           "quay.io/kubermatic/metering:v0.5",
+					ImagePullPolicy: corev1.PullAlways,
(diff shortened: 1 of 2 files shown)


pkg/ee/metering/cronjob.go
```diff
@@ -115,7 +115,7 @@ mc mirror --newer-than "65d0h0m" s3/$S3_BUCKET /metering-data || true`,
 			job.Spec.JobTemplate.Spec.Template.Spec.Containers = []corev1.Container{
 				{
 					Name:  "kubermatic-metering-report",
-					Image: "quay.io/kubermatic/metering:7a12117",
+					Image: "quay.io/kubermatic/metering:v0.5",
```
mate4st: NIT. would be nice to set `ImagePullPolicy: corev1.PullAlways`

simon-wessel: Added in 92f17f1dc5bc4f0cc3e7fc621a442614e5553a73 :+1:

moadqassem (commented):
@simon-wessel what about the deployment ?

simon-wessel:
> @simon-wessel what about the deployment ?

@moadqassem I am not sure what you are referring to. If it is about the ImagePullPolicy: That has already been set for the deployment.

moadqassem (commented):
/approve
/lgtm

PR #7665, opened 2021-09-05
Enable optional tls support for minio chart

moelsayed:
The new BackupRestore controller requires an s3-compatible storage backend that works over TLS. Our current minio deployment works in-cluster and doesn't have TLS enabled. This PR provide optional support for TLS. 

Fixes #

**Documentation**:

```release-note
- Adds  optional TLS support for minio chart. The user can define a TLS secret that minio will use for its server. The TLS certificates should be signed by the global Kubermatic CA documented in https://github.com/kubermatic/docs/pull/524/files
```

### DIFF
--- a/charts/minio/templates/deployment.yaml
+++ b/charts/minio/templates/deployment.yaml
@@ -67,6 +67,10 @@ spec:
         ports:
         - containerPort: 9000
         volumeMounts:
+        {{- if .Values.minio.certificatesSecret }}
+        - name: certificates-volume
+          mountPath: /root/.minio/certs
+        {{- end }}
         - name: storage
           mountPath: "/storage"
         resources:
@@ -78,7 +82,11 @@ spec:
         args:
         - /bin/sh
         - -c
+        {{- if .Values.minio.certificatesSecret }}
+        - mc --insecure config host add src https://127.0.0.1:9000 "$MINIO_ACCESS_KEY" "$MINIO_SECRET_KEY" && while true; do sleep 1h; done
+        {{- else }}
         - mc config host add src http://127.0.0.1:9000 "$MINIO_ACCESS_KEY" "$MINIO_SECRET_KEY" && while true; do sleep 1h; done
+        {{- end }}
         env:
         - name: MINIO_ACCESS_KEY
           valueFrom:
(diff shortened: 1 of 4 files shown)


charts/minio/values.yaml
```diff
@@ -17,7 +17,11 @@ minio:
     repository: docker.io/minio/minio
     tag: RELEASE.2021-08-20T18-32-01Z
   storeSize: 100Gi
-
+  
+  # The is required to enable the BackupRestore controller so it can backup
+  # and restore from the local minio deployment
+  minioCertificateSecret: minio-certificates # tls secret used by minio.
```
xrstf: Please remove the "minio" prefix for this key.

charts/minio/templates/deployment.yaml
```diff
@@ -78,7 +82,11 @@ spec:
         args:
         - /bin/sh
         - -c
+        {{- if .Values.minio.minioCertificates }}
+        - mc --insecure config host add src https://127.0.0.1:9000 "$MINIO_ACCESS_KEY" "$MINIO_SECRET_KEY" && while true; do sleep 1h; done
```
xrstf, line 86: You have the certificate mounted already, right? So why do we need `--insecure` here?

charts/kubermatic/static/store-container-new.yaml
```diff
@@ -8,6 +8,7 @@ command:
 - |
   set -e
   s3cmd \
+    --no-check-certificate \
```
xrstf: So now we configure both a CA, but also disable any certificte checks? What's the point of that then?

xrstf: Customers should put the CA they use for their cert into the global CA bundle as documented (and then blindly deleted) in https://github.com/kubermatic/docs/pull/524

charts/minio/templates/deployment.yaml
```diff
@@ -108,6 +116,18 @@ spec:
...
+      - name: certificates-volume
+        secret:
+          secretName: {{ .Values.minio.certificatesSecret }}
+          items:
+          - key: tls.crt
+            path: public.crt
+          - key: tls.key
+            path: private.key
+          - key: tls.crt
+            path: CAs/public.crt
```
xrstf, line 129: Why is the cert mounted twice?

charts/minio/values.yaml
```diff
@@ -17,7 +17,11 @@ minio:
     repository: docker.io/minio/minio
     tag: RELEASE.2021-08-20T18-32-01Z
   storeSize: 100Gi
-
+  
+  # The is required to enable the BackupRestore controller so it can backup
+  # and restore from the local minio deployment
+  certificateSecret: minio-certificates # tls secret used by minio.
```
xrstf: This is where we need to comment/document that the user needs to but the CA used for their cert into the global CA bundle. Nobody will ever read the release note ;)

charts/minio/templates/deployment.yaml
moelsayed: needs to go to two places: https://github.com/minio/minio/tree/master/docs/tls/kubernetes#3-update-deployment-yaml-file

moadqassem (commented):
/approve
/lgtm

jiachengxu (approved):

PR #7679, opened 2021-09-07
add option to enable konnectivity by user-cluster

PratikDeoghare:
Adds option to enable/disable konnectivity by user-cluster. 

Ref #7504 

```release-note
Users can now enble/disable konnectivity on their clusters. 
```

### DIFF
--- a/pkg/crd/kubermatic/v1/cluster.go
+++ b/pkg/crd/kubermatic/v1/cluster.go
@@ -481,6 +481,9 @@ type ClusterNetworkingConfig struct {
 	// NodeLocalDNSCacheEnabled controls whether the NodeLocal DNS Cache feature is enabled.
 	// Defaults to true.
 	NodeLocalDNSCacheEnabled *bool `json:"nodeLocalDNSCacheEnabled,omitempty"`
+
+	// KonnectivityEnabled enables konnectivity for controlplane to node network communication.
+	KonnectivityEnabled bool `json:"konnectivityEnabled,omitempty"`
 }
 
 // MachineNetworkingConfig specifies the networking parameters used for IPAM.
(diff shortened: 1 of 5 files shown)


pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -175,6 +175,9 @@ type ClusterSpec struct {
 
 	// CNIPlugin contains the spec of the CNI plugin to be installed in the cluster.
 	CNIPlugin *CNIPluginSettings `json:"cniPlugin,omitempty"`
+
+	// IsKonnectivityEnabled enables connectivity for controlplane to node network communication.
+	IsKonnectivityEnabled bool `json:"isKonnectivityEnabled,omitempty"`
```
rastislavs: I would prefer to put this into `ClusterNetwork` (`ClusterNetworkingConfig`).
Also, I would prefer to keep the naming consistent with other bool flags in the API - without the `Is` prefix (e.g.  `MonitoringEnabled` / `NodeLocalDNSCacheEnabled`).

rastislavs:
@PratikDeoghare Please run `hack/gen-api-client.sh`

pkg/resources/usercluster/deployment.go
```diff
@@ -68,6 +68,7 @@ type userclusterControllerData interface {
 	KubermaticDockerTag() string
 	GetCloudProviderName() (string, error)
 	UserClusterMLAEnabled() bool
+	IsKonnectivityEnabled() bool
```
maciaszczykm, line 71: Implementation of this method is missing in this pull request. Was it added before?

maciaszczykm:
It would be great to add some simple test to check if the value is saved successfully.

pkg/resources/usercluster/deployment.go
PratikDeoghare: It was already there just wasn't included in the interface.

mfranczy:
looks good to me, please squash commits and I will approve it

PR #7744, opened 2021-09-08
Fix for seed API using seed clients instead of master

### LINKED ISSUE #7667: Seed API fails updating Seeds on Seed clusters which are not also master clusters
**User Story**
It seems that the Seed API has a bug where it tries to use the seed cluster clients to upgrade the Seed objects(taken form the master cluster), instead of using the master client, which causes Precondition failures.

Furthermore there is no point to update Seeds on the seed clusters, as the changes will be reconciled with the Seed state from the master cluster. 

Use the master client instead of the seed clients to manage Seeds.

lsviben:
Fixes the issue where Seed API was using seed clients to update the Seeds on master cluster instead of using the master client. This was causing Seed API not to work on Seeds which were not also the master clusters.

Fixes #7667 

```release-note
Fixes the issue where Seed API was using seed clients to update the Seeds on master cluster instead of using the master client. This was causing Seed API not to work on Seeds which were not also the master clusters.
```

### DIFF
--- a/pkg/handler/routes_v1.go
+++ b/pkg/handler/routes_v1.go
@@ -1560,7 +1560,7 @@ func (r Routing) createDC() http.Handler {
 		endpoint.Chain(
 			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
 			middleware.UserSaver(r.userProvider),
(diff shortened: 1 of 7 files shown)


xrstf (approved):
/approve

PR #7747, opened 2021-09-08
fix allowedregistry list overwriting

### LINKED ISSUE #7746: Multiple AllowedRegistries produce just one item in the Default Constraint registry list
Create 2 different Allowedregistries with different registryPrefixes. 

Expected result is that the resulting default constraint will have 2 items in the list, but instead it has 1, they overwrite each other.

lsviben:
Fixes a bug where 2 or more allowedRegistries produced just one item in the allowedRegistry Default Constraint parameters. 
Which was happening because of a bad casting that made the set of registries always empty. 

Additionally, eases up the DefaultConstraint reconciliation, so that users can set filtering and matching without it being overwritten.

Fixes #7746 

```release-note
NONE
```


mfranczy (approved):
/lgtm
/approve

jiachengxu (approved):

PR #7749, opened 2021-09-08
Replace type of Mutation Field to Bool

### LINKED ISSUE #7748: OPA enable mutation causing error
**User Story**
OPA enable mutation flag `--enable-mutation`  causing User Controller Manager to crash as the argument type in the code is taken as Address not as Bool

**Acceptance criteria**
Argument Value is taken as bool in User Controller Manager so that the value remains static

imharshita:

Fixes issue caused by Address value of Bool Argument `--enable-mutation`

Fixes #7748

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/resources/usercluster/deployment.go
+++ b/pkg/resources/usercluster/deployment.go
@@ -170,8 +170,8 @@ func DeploymentCreator(data userclusterControllerData) reconciling.NamedDeployme
 				args = append(args, "-opa-webhook-timeout", fmt.Sprint(*data.Cluster().Spec.OPAIntegration.WebhookTimeoutSeconds))
 			}
 
-			if data.Cluster().Spec.OPAIntegration != nil && data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation {
-				args = append(args, "--enable-mutation", fmt.Sprint(&data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation))
+			if data.Cluster().Spec.OPAIntegration != nil && data.Cluster().Spec.OPAIntegration.Enabled {
+				args = append(args, fmt.Sprintf("-enable-mutation=%t", data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation))
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/resources/usercluster/deployment.go
```diff
@@ -171,7 +171,7 @@ func DeploymentCreator(data userclusterControllerData) reconciling.NamedDeployme
 			}
 
 			if data.Cluster().Spec.OPAIntegration != nil && data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation {
-				args = append(args, "--enable-mutation", fmt.Sprint(&data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation))
+				args = append(args, "-enable-mutation", fmt.Sprint(data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation))
```
mfranczy: https://pkg.go.dev/flag#hdr-Command_line_flag_syntax

```
-flag x  // non-boolean flags only
```

please change to 
```
fmt.Sprintf("-enable-mutation=%t", data.Cluster().Spec.OPAIntegration.ExperimentalEnableMutation)),
```

mfranczy: btw. should not be the condition the same as here [link](https://github.com/kubermatic/kubermatic/pull/7749/files#diff-fc8f73fda439bfb46949c7196ce280cfe5393c27b629249b8d342a5ff4f981f3R173)?

If OPAIntegration is disabled but experimental mutation is enabled, do we still want to run the mutation webhook? If yes, why?

mfranczy (changes requested):

pkg/resources/usercluster/deployment.go
imharshita: In that case we don't want mutation webhook, updated the changes.
fyi create a PR to remove mutation webhook when `enable-mutation:false`

mfranczy (approved):
/approve
/lgtm

PR #7755, opened 2021-09-09
Merge swagger:meta blocks into one

### LINKED ISSUE #7745: Swagger spec has some invalid/old data
**User Story**
Swagger spec: https://github.com/kubermatic/kubermatic/blob/master/cmd/kubermatic-api/swagger.json

- `version` inside is outdated
- `termsOfService` is in invalid format, it should be an URL (`Must be in the URL format`, see: https://swagger.io/docs/specification/api-general-info/)
- Kubermatic should be changed to KKP?
- dot in the `title` should be removed

I think it's worth updating because we show this later in the UI:

Zrzut ekranu 2021-09-8 o 15 32 30

The link is broken atm.

maciaszczykm:
**What this PR does / why we need it**: I'm opening it to update the outdated fields but also for a discussion if we can simplify our config. More info in the comments.

Fixes https://github.com/kubermatic/kubermatic/issues/7745

```release-note
NONE
```


cmd/kubermatic-api/main.go
```diff
@@ -14,24 +14,29 @@ See the License for the specific language governing permissions and
...
 //
+//     Security:
+//     - api_key:
+//
+//     SecurityDefinitions:
+//     api_key:
+//          type: apiKey
+//          name: Authorization
+//          in: header
+//
```
maciaszczykm, line 39: I think it is a good idea to keep it in the main `swagger:meta` and show this info in the `swagger.json`.

cmd/kubermatic-api/main.go
```diff
@@ -14,24 +14,29 @@ See the License for the specific language governing permissions and
...
-// This describes possible operations which can be made against the Kubermatic API.
-//
-// Terms Of Service:
-//
-// There are no TOS at this moment, use at your own risk we take no responsibility
-//
-//     Version: 2.11
+//     Schemes: https
+//     Host: dev.kubermatic.io
+//     Version: 2.18
```
maciaszczykm, line 23: Updated.

cmd/kubermatic-api/main.go
```diff
@@ -14,24 +14,29 @@ See the License for the specific language governing permissions and
...
 //
-// This describes possible operations which can be made against the Kubermatic API.
-//
-// Terms Of Service:
-//
-// There are no TOS at this moment, use at your own risk we take no responsibility
-//
-//     Version: 2.11
+//     Schemes: https
+//     Host: dev.kubermatic.io
```
maciaszczykm, line 22: This one is the difference from before. Previously this was only in `routes_v1.go` and used only during API client generation. Do we really want to hide that in our main `swagger.json`?

floreks (approved):
Looks fine to me, but I'd wait for @zreigz to approve it.

PR #7757, opened 2021-09-09
Fix metering namespace

mate4st:
Set metering resources namespace to equal seed namespace

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operator/seed/resources/metering/wrappers_ce.go
+++ b/pkg/controller/operator/seed/resources/metering/wrappers_ce.go
@@ -21,17 +21,13 @@ package metering
 import (
 	"context"
 
-	"go.uber.org/zap"
-
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
-	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
 
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ReconcileMeteringResources reconciles the metering related resources.
-func ReconcileMeteringResources(_ context.Context, _ ctrlruntimeclient.Client, _ string,
-	_ *operatorv1alpha1.KubermaticConfiguration, _ *kubermaticv1.Seed, _ *zap.SugaredLogger) error {
+func ReconcileMeteringResources(_ context.Context, _ ctrlruntimeclient.Client, _ *kubermaticv1.Seed) error {
 
 	return nil
 }
--- a/pkg/ee/metering/handler.go
+++ b/pkg/ee/metering/handler.go
@@ -47,15 +47,14 @@ import (
 )
 
 const (
-	AccessKey       = "accessKey"
-	SecretKey       = "secretKey"
-	Bucket          = "bucket"
-	Endpoint        = "endpoint"
-	SecretName      = "metering-s3"
-	SecretNamespace = "kubermatic"
+	AccessKey  = "accessKey"
+	SecretKey  = "secretKey"
+	Bucket     = "bucket"
+	Endpoint   = "endpoint"
+	SecretName = "metering-s3"
 )
 
-var secretNamespacedName = types.NamespacedName{Name: SecretName, Namespace: SecretNamespace}
+var secretNamespacedName = types.NamespacedName{Name: SecretName, Namespace: resources.KubermaticNamespace}
 
 type configurationReq struct {
 	Enabled          bool   `json:"enabled"`
--- a/pkg/ee/metering/reconcile.go
+++ b/pkg/ee/metering/reconcile.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/resources"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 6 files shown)


pkg/ee/metering/deployment.go
```diff
@@ -48,6 +48,8 @@ func deploymentCreator(seed *kubermaticv1.Seed) reconciling.NamedDeploymentCreat
 	return func() (string, reconciling.DeploymentCreator) {
 		//TODO: Add custom values for the metering deployment fields such as seed, interval and output-rotation.
 		return meteringToolName, func(d *appsv1.Deployment) (*appsv1.Deployment, error) {
+			d.Name = meteringToolName
+			d.Namespace = resources.KubermaticNamespace
```
simon-wessel: This should probably be dynamic to the seed namespace name. Maybe it would be better to use `d.Namespace = seed.Namespace` instead like in [pvc.go](https://github.com/kubermatic/kubermatic/blob/master/pkg/ee/metering/pvc.go#L49)?

simon-wessel (dismissed):

pkg/ee/metering/deployment.go
mate4st: I think it makes sense to use the namespace provided by the operator. It is the same that is used to deploy the Seed-Controller-Manager. I do not know what is in control of the seed namespace. Maybe it is the same?

pkg/controller/operator/seed/resources/metering/wrappers_ce.go
```diff
@@ -21,17 +21,13 @@ package metering
 import (
 	"context"
 
-	"go.uber.org/zap"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
```
moadqassem: Please regroup the imports

pkg/ee/metering/handler.go
```diff
@@ -175,19 +172,19 @@ func CreateOrUpdateCredentials(ctx context.Context, request interface{}, seedsGe
 		Endpoint:  []byte(req.Endpoint),
 	}
 
-	for _, client := range seeds {
-		if err := createOrUpdateMeteringToolSecret(ctx, client, data); err != nil {
+	for seed, client := range seeds {
+		if err := createOrUpdateMeteringToolSecret(ctx, client, seed.Namespace, data); err != nil {
```
moadqassem: Why you are not using the `k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace` ?

pkg/ee/metering/reconcile.go
```diff
@@ -28,46 +28,41 @@ import (
...
 
 	if err := persistentVolumeClaimCreator(ctx, client, seed); err != nil {
-		return fmt.Errorf("failed to reconcile metering pvc: %v", err)
+		return fmt.Errorf("failed to reconcile metering PVC: %v", err)
 	}
 
 	if err := reconciling.ReconcileServiceAccounts(ctx, []reconciling.NamedServiceAccountCreatorGetter{
 		serviceAccountCreator(),
-	}, metav1.NamespaceSystem, client); err != nil {
+	}, seed.Namespace, client); err != nil {
```
moadqassem: Why you are not using the k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace ?

pkg/ee/metering/reconcile.go
```diff
@@ -28,46 +28,41 @@ import (
...
 		return fmt.Errorf("failed to reconcile metering ServiceAccounts: %v", err)
 	}
 
 	if err := reconciling.ReconcileClusterRoleBindings(ctx, []reconciling.NamedClusterRoleBindingCreatorGetter{
-		clusterRoleBindingCreator(cfg.Namespace),
+		clusterRoleBindingCreator(seed.Namespace),
```
moadqassem: Why you are not using the k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace ?

moadqassem (commented):
@WeirdMachine Haven't we agreed on just simply use the Kubermatic default namespace: 
`k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace`

WeirdMachine:
> @WeirdMachine Haven't we agreed on just simply use the Kubermatic default namespace:
> `k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace`

Yes. The idea behind this is to keep the namespace dynamic and deploy the metering next to the seed. Maybe i got a little confused about it, as in the previous version different resources were used to set the namespace. If we can rely on kubermatic namespace existence in every seed cluster i am fine with hardcoding it.

moadqassem:
@WeirdMachine There is absolutely no way that KKP can actually work without the `kubermatic` ns. This is somewhat of a standard. Anything that is seed related should go to the `kubermatic` ns. Since this feature is seed exclusive please put it there as it used to be before.

moadqassem:
> > @WeirdMachine Haven't we agreed on just simply use the Kubermatic default namespace:
> > `k8c.io/kubermatic/v2/pkg/resources.KubermaticNamespace`
> 
> Yes. The idea behind this is to keep the namespace dynamic and deploy the metering next to the seed. Maybe i got a little confused about it, as in the previous version different resources were used to set the namespace. If we can rely on kubermatic namespace existence in every seed cluster i am fine with hardcoding it.

@WeirdMachine There is absolutely no way that KKP can actually work without the kubermatic ns. This is somewhat of a standard. Anything that is seed related should go to the kubermatic ns. Since this feature is seed exclusive please put it there as it used to be before.

WeirdMachine:
@moadqassem I updated the PR. All resources should now be deployed consistently to the Kubermatic namespace.

moadqassem (commented):
/approve
/lgtm

PR #7761, opened 2021-09-09
Enable backup restore based on seed config

floreks:
Backup restore config was not taken into account when using new per-seed config.

```release-note
NONE
```


moadqassem (commented):
/approve
/lgtm

moadqassem:
cc @moelsayed

PR #7763, opened 2021-09-09
add a finalizer and cluster annotation only when we access s3 and fin…

moadqassem:
Adds the finalizer on the EtcdRestore CR only when the etcd _estore controller can access s3 and find the right object. Otherwise if the credentials/configurations of  s3 are wrong the restore object will be stuck and never gets deleted unless the credentials has been fixed. This is not a good experience for the ui user as they should be able to delete a failing restore object.     
Fixes #

**Documentation**:

```release-note
None
```


moadqassem:
cc @floreks @lsviben

PR #7764, opened 2021-09-09
Use KubermaticConfiguration as source for versions/updates

xrstf:
Since #7558, we have the KubermaticConfiguration (KC) available on master and seed clusters, so now we can begin to actually use it! We already dynamically loaded Seeds, so now we do exactly the same for the KC (this means that changes to the config do not require a rollout again).

This PR is the first step and removes the versions.yaml, updates.yaml  and provider-incompatibilities.yaml files. Instead this data is now taken directly from the KC. No more YAML serializing needed.

There are now a metric ton of CLI flags that can all be removed over time from the seed-ctrlmgr. That's gonna be a fun refactoring!

From an admin standpoint, nothing changes. They still continue to configure everything in the KC and let the KKP Operator take care of spreading that to whereever it's needed.

```release-note
NONE
```


mlavacca (commented):
/approve
/lgtm

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -20,13 +20,34 @@ import (
...
+)
+
+// OperationType is the type defining the operations triggering the compatibility check (CREATE or UPDATE)
+type OperationType string
+
+const (
+	// CreateOperation represents the creation of a new cluster
+	CreateOperation OperationType = "CREATE"
+	// UpdateOperation represents the update of an existing cluster
+	UpdateOperation OperationType = "UPGRADE"
```
lsviben, line 46: NIT: should it be UPDATE? or the var should be called UpgradeOperation and the description updated

xrstf: We have many functions called "controlPlaneUpgrade", so we have 2 distinct names pretty much everywhere for this. I agree that it's ugly, but it's not this PR's job to clean that up ^^

lsviben: ok, leaving it then :)

PR #7792, opened 2021-09-13
Fix Etcd backups user rbacs

### LINKED ISSUE #7788: Non-admin user RBAC for etcdBackupConfig and etcdRestore missing
**User Story**
Non-admin users are missing RBACs for etcdBackupConfig and etcdRestore which are causing errors poping up in the UI plus of course they cant use these features normally.

lsviben:
Non-admins didnt have the RBACs needed to access EtcdBackupConfigs and EtcdRestores. This fixes it. 

Also as the project listing endpoint was using cluster-wide listing, instead of giving cluster wide access to users, we make the check if user has access to the project, and then list.

Fixes #7788 

```release-note
NONE
```


mfranczy (approved):
/approve
/lgtm

PR #7793, opened 2021-09-13
defaulting backup/restore containers templates

### LINKED ISSUE #7778: Enabling new Automatic Backup&Restore on Seed does not mount proper store and delete containers
When enabling the new Automatic Backup&Restore on the Seed, without enabling it also in the deprecated KubermaticConfiguration BackupRestore settings, the proper store and delete containers are not mounted to the seed-controller-manager.

The issue for delete container  is here ([backupContainerconfigMap](https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/seed/resources/kubermatic/kubermatic.go#L78-L94)):

```
func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedConfigMapCreatorGetter {
	return func() (string, reconciling.ConfigMapCreator) {
		return backupContainersConfigMapName, func(c *corev1.ConfigMap) (*corev1.ConfigMap, error) {
			if c.Data == nil {
				c.Data = make(map[string]string)
			}

			c.Data[storeContainerKey] = cfg.Spec.SeedController.BackupStoreContainer
			c.Data[cleanupContainerKey] = cfg.Spec.SeedController.BackupCleanupContainer
			if cfg.Spec.SeedController.BackupRestore.Enabled {
				c.Data[deleteContainerKey] = cfg.Spec.SeedController.BackupDeleteContainer
			}

			return c, nil
		}
	}
}
```
Which can be easily fixed by just removing the condition and having the DeleteContainer always present and also doing a similar thing in [defaults.go](https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/common/defaults.go#L322-L333)

The new store container is a bit more complicated as it shares the same key with the old store container [defaults.go](https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/common/defaults.go#L308-L315), so either we need to know KKP is using the new or old backups, or we need to make them use different keys.
```
	if copy.Spec.SeedController.BackupStoreContainer == "" {
		if copy.Spec.SeedController.BackupRestore.Enabled {
			copy.Spec.SeedController.BackupStoreContainer = strings.TrimSpace(DefaultNewBackupStoreContainer)
		} else {
			copy.Spec.SeedController.BackupStoreContainer = strings.TrimSpace(DefaultBackupStoreContainer)
		}
		logger.Debugw("Defaulting field", "field", "seedController.backupStoreContainer")
	}
```

moadqassem:
Enabling the backup and restore containers for Etcd in the new seed configurations

Fixes #7778

**Documentation**:

```release-note
None
```

### DIFF
--- a/pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
+++ b/pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
@@ -18,7 +18,11 @@ package kubermatic
 
 import (
 	"fmt"
+	"strings"
 
+	"go.uber.org/zap"
+
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
@@ -75,7 +79,7 @@ func ClusterRoleBindingCreator(cfg *operatorv1alpha1.KubermaticConfiguration, se
 	}
 }
 
-func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.NamedConfigMapCreatorGetter {
+func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfiguration, seed *kubermaticv1.Seed, log *zap.SugaredLogger) reconciling.NamedConfigMapCreatorGetter {
 	return func() (string, reconciling.ConfigMapCreator) {
 		return backupContainersConfigMapName, func(c *corev1.ConfigMap) (*corev1.ConfigMap, error) {
 			if c.Data == nil {
@@ -84,10 +88,18 @@ func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfigurat
 
 			c.Data[storeContainerKey] = cfg.Spec.SeedController.BackupStoreContainer
 			c.Data[cleanupContainerKey] = cfg.Spec.SeedController.BackupCleanupContainer
-			if cfg.Spec.SeedController.BackupRestore.Enabled {
-				c.Data[deleteContainerKey] = cfg.Spec.SeedController.BackupDeleteContainer
+
+			if cfg.Spec.SeedController.BackupStoreContainer == strings.TrimSpace(common.DefaultBackupStoreContainer) &&
+				seed.Spec.BackupRestore != nil {
+				c.Data[storeContainerKey] = strings.TrimSpace(common.DefaultNewBackupStoreContainer)
 			}
 
+			if cfg.Spec.SeedController.BackupDeleteContainer == "" {
+				cfg.Spec.SeedController.BackupDeleteContainer = strings.TrimSpace(common.DefaultNewBackupDeleteContainer)
+				log.Debugw("Defaulting field", "field", "seedController.backupDeleteContainer")
+			}
+
+			c.Data[deleteContainerKey] = cfg.Spec.SeedController.BackupDeleteContainer
 			return c, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


moadqassem:
cc @lsviben

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
```diff
@@ -84,10 +88,18 @@ func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfigurat
 
 			c.Data[storeContainerKey] = cfg.Spec.SeedController.BackupStoreContainer
 			c.Data[cleanupContainerKey] = cfg.Spec.SeedController.BackupCleanupContainer
-			if cfg.Spec.SeedController.BackupRestore.Enabled {
-				c.Data[deleteContainerKey] = cfg.Spec.SeedController.BackupDeleteContainer
+
+			if cfg.Spec.SeedController.BackupStoreContainer == strings.TrimSpace(common.DefaultBackupStoreContainer) &&
+				seed.Spec.BackupRestore != nil {
```
moelsayed, line 93: Does `BackupRestore` still have an `Enabled` field? should we use it here?

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
```diff
@@ -84,10 +88,18 @@ func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfigurat
...
+				c.Data[storeContainerKey] = strings.TrimSpace(common.DefaultNewBackupStoreContainer)
+			}
+
+			if cfg.Spec.SeedController.BackupDeleteContainer == "" {
```
moelsayed: Should we include this in the `seed.Spec.BackupRestore != nil` as well? also line 102?

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
```diff
@@ -84,10 +88,18 @@ func BackupContainersConfigMapCreator(cfg *operatorv1alpha1.KubermaticConfigurat
 
 			c.Data[storeContainerKey] = cfg.Spec.SeedController.BackupStoreContainer
 			c.Data[cleanupContainerKey] = cfg.Spec.SeedController.BackupCleanupContainer
-			if cfg.Spec.SeedController.BackupRestore.Enabled {
-				c.Data[deleteContainerKey] = cfg.Spec.SeedController.BackupDeleteContainer
+
+			if cfg.Spec.SeedController.BackupStoreContainer == strings.TrimSpace(common.DefaultBackupStoreContainer) &&
+				seed.Spec.BackupRestore != nil {
+				c.Data[storeContainerKey] = strings.TrimSpace(common.DefaultNewBackupStoreContainer)
+			}
```
moelsayed, line 96: nit: debug log is a good idea, we could add one here, too.

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
moadqassem: No the api doesn't have that nor the seed spec. Adding this field means adding the UI and the API in addition the seed spec. TBH from my side I guess it is not intuitive for the customers to have a hint that the new backup and restore is disabled just because the fields are empty, their a better UX for this. @lsviben @floreks WDYT ?

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
moadqassem: Why? it could be that the backup and restore has been enabled in the KubermaticConfigurations crd and was left out from the seed.

moelsayed: I made a mistake there, I am sorry. The `BackupStoreContainer` is used in both legacy backup and BackupRestore controllers (with slightly different configuration). The `BackupDeleteCintainer` is _only_ used in the BackupRestore controllers. 

Defaulting logic should respect that. So, `BackupDeleteContainer` should be only enabled when `seed.Spec.BackupRestore != nil`.

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
moadqassem: Anyway maybe for now we leave it as it is

pkg/controller/operator/seed/resources/kubermatic/kubermatic.go
moadqassem: @moelsayed ok understood.

moadqassem:
@moelsayed PTAL

Issue #7759, opened 2021-09-09
Cluster with Konnectivity enabled does not start


rastislavs:
**User Story**
In a KKP installation with `KonnectivityService` featureGate enabled in KubermaticConfiguration, I created a new cluster with Konnectivity enabed:

```
spec:
  clusterNetwork:
    konnectivityEnabled: true
```

The cluster does not come up at all and the following error can be found in the Events log:

```
failed to reconcile cluster: failed to get user-cluster client: Secret "internal-admin-kubeconfig" not found
```

No resources can be found in the cluster namespace in Seed.

**Acceptance criteria**
The user cluster with konnectivityEnabled should start and use Konnectivity for control-plane to cluster network connectivity.

Issue #7760, opened 2021-09-09
Enabing Konnectivity for an exiting cluster does not work


rastislavs:
**User Story**
In a KKP installation with the `KonnectivityService` featureGate enabled in KubermaticConfiguration, I created a new cluster with Konnectivity disabled. After all cluster components came up & verifying that everything works properly, I enabled Konnectivity in the Cluster CRD:

```
spec:
  clusterNetwork:
    konnectivityEnabled: true
```

Konnectivity was not enabled and the following error could be found in the cluster Events log:
```
failed to reconcile cluster: failed to ensure that the Secret exists: failed to ensure Secret cluster-wrl2bxq6nf/stolen-agent-ca: failed to generate object: failed to get konnectivity-service-account: serviceaccounts "system-konnectivity-agent" not found
```

**Acceptance criteria**
The user cluster should start using Konnectivity for control-plane to cluster network connectivity after switching `clusterNetwork.konnectivityEnabled` to `true`, no errors should be produced.

PR #7795, opened 2021-09-14
reorder konnectivity resources to reconcile after dependencies

PratikDeoghare:
Deployment of Konnectivity resources failed because we tried to deploy them before 
their dependencies were deployed. This PR carefully orders the Konnectivity resources to be deployed. 

PR also  fixes  `hack/run-user-cluster-controller-manager.sh`. If `-tunneling-agent-ip` is not set `envoy-agent` doesn't 
deploy. This makes call from `calico/node` to KAS fail preventing setup of pod networking. This shouldn't be an issue 
on the actual clusters (only during  the development) because on real clusters the flag is set.  

Fixes 
https://github.com/kubermatic/kubermatic/issues/7759
https://github.com/kubermatic/kubermatic/issues/7760

```release-note
NONE
```

### DIFF
--- a/hack/run-user-cluster-controller-manager.sh
+++ b/hack/run-user-cluster-controller-manager.sh
@@ -84,6 +84,7 @@ set -x
   -version=${CLUSTER_VERSION} \
   -log-debug=$KUBERMATIC_DEBUG \
   -log-format=Console \
+  -tunneling-agent-ip=192.168.30.10\
   -logtostderr \
   -v=4 \
   -seed-kubeconfig=${SEED_KUBECONFIG} \
--- a/hack/run-user-cluster-controller-manager.sh
+++ b/hack/run-user-cluster-controller-manager.sh
@@ -84,8 +84,8 @@ set -x
   -version=${CLUSTER_VERSION} \
   -log-debug=$KUBERMATIC_DEBUG \
   -log-format=Console \
-  -tunneling-agent-ip=192.168.30.10\
   -logtostderr \
+  -tunneling-agent-ip=192.168.30.10 \
   -v=4 \
   -seed-kubeconfig=${SEED_KUBECONFIG} \
   -owner-email=${OWNER_EMAIL} \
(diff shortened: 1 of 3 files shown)


hack/run-user-cluster-controller-manager.sh
```diff
@@ -85,6 +85,7 @@ set -x
   -log-debug=$KUBERMATIC_DEBUG \
   -log-format=Console \
   -logtostderr \
+  -tunneling-agent-ip=192.168.30.10 \
```
xrstf, line 88: Because I'm a noob: Where does this IP come from? Could I also use `192.168.30.9`?

xrstf (commented):
One question, otherwise nice

hack/run-user-cluster-controller-manager.sh
rastislavs: seems to be hard-coded in the seed-controller-manager: https://github.com/kubermatic/kubermatic/blob/3a93c3489da1d3c5249edac40a6cb749d4a7f296/cmd/seed-controller-manager/options.go#L122-L123

rastislavs: I think it in theory can be any IP from an IP range that does not collide with node subnet / pod CIDR / services CIDR? (cc @youssefazrak )

youssefazrak: It is indeed hardcoded. This is what we chose with Iacopo for the tech preview and the goal is to make this one configurable as of right now, there is a probability of colliding with an existing IP in the VPC, data center, whatever is called.
Regarding the `192.168.30.10` choice it is random, in one of the 3 private ranges, and not one of the used ones from a customer/client perspective like `192.168.0.0/24`. Moreover and as mentioned by Rastislav, it's not used from a KKP perspective in any of the Pod/Services/Nodes/OpenVPN(DNAT) ranges.

xrstf: > right now, there is a probability of colliding with an existing IP in the VPC, data center

Does this affect production usage or only when 2 people use the script at the same time for debugging stuff?

youssefazrak: @xrstf it could affect production. Depending on the networks tbh.
In all cases, this should be made configurable.

PR #7799, opened 2021-09-14
migration webhook deployed only when migrating

mlavacca:
The vSphere CSI `validatingWebhookConfiguration` has to be deployed only when performing the CCM/CSI migration. We didn't check the migration condition but deployed it in every condition when the external CCM was used. This PR ensures that it is deployed only when migrating the cluster, or once the migration has been completed.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #7800, opened 2021-09-14
Change the name of etcd backup and restore secret

moadqassem:
Change the name of the secret which is used for the new etcd backup and restore feature, due to some conflicts with the minio chart secret.
Fixes #

**Documentation**:

```release-note
None
```

### DIFF
--- a/charts/kubermatic/static/cleanup-container.yaml
+++ b/charts/kubermatic/static/cleanup-container.yaml
@@ -29,10 +29,10 @@ env:
 - name: ACCESS_KEY_ID
   valueFrom:
     secretKeyRef:
-      name: s3-credentials
+      name: s3
       key: ACCESS_KEY_ID
 - name: SECRET_ACCESS_KEY
   valueFrom:
     secretKeyRef:
-      name: s3-credentials
+      name: s3
       key: SECRET_ACCESS_KEY
--- a/charts/kubermatic/static/cleanup-container.yaml
+++ b/charts/kubermatic/static/cleanup-container.yaml
@@ -29,10 +29,10 @@ env:
 - name: ACCESS_KEY_ID
   valueFrom:
     secretKeyRef:
-      name: s3
+      name: s3-credentials
       key: ACCESS_KEY_ID
 - name: SECRET_ACCESS_KEY
   valueFrom:
     secretKeyRef:
-      name: s3
+      name: s3-credentials
       key: SECRET_ACCESS_KEY
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -471,9 +471,9 @@ const (
 	EtcdTLSKeySecretKey = "etcd-tls.key"
 
 	// EtcdRestoreS3CredentialsSecret names the secret expected in seed kube-system that must contain S3 credentials for etcd backup restores
-	EtcdRestoreS3CredentialsSecret     = "s3-credentials"
-	EtcdRestoreS3AccessKeyIDKey        = "ACCESS_KEY_ID"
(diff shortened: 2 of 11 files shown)


pkg/resources/resources.go
```diff
@@ -471,9 +471,9 @@ const (
 	EtcdTLSKeySecretKey = "etcd-tls.key"
 
 	// EtcdRestoreS3CredentialsSecret names the secret expected in seed kube-system that must contain S3 credentials for etcd backup restores
-	EtcdRestoreS3CredentialsSecret     = "s3-credentials"
-	EtcdRestoreS3AccessKeyIDKey        = "ACCESS_KEY_ID"
-	EtcdRestoreS3SecretKeyAccessKeyKey = "SECRET_ACCESS_KEY"
+	EtcdRestoreS3CredentialsSecret              = "s3"
```
xrstf: While we're at it, it might be the perfect time to put the purpose into the secret name? Maybe call it `backup-credentials`?

moadqassem: The thing is we are gonna have multiple storage providers so not only s3. Thus in the future we will have, s3, google-file storage and azure storage. That's why we agreed on regarding naming it.

xrstf: `backup-s3` then maybe? I can see your point, but we recently had a case where our infra had a "bucket-reader" account that was used by an unknown number of components and it became much easier to reason about once we renamed the account to "prow-deck". The current confusion as to who and what should use the secret should be example enough why we should include "backup" somewhere in the name. :)

Another point: What if another component also needs S3 credentials, but for something totally different? Would we then want to have "s3-2"? "s3-second"? ^^

lsviben: `backup-s3` sounds good to me

pkg/resources/resources.go
```diff
@@ -1247,10 +1247,10 @@ func GetEtcdRestoreS3Client(ctx context.Context, restore *kubermaticv1.EtcdResto
 			return nil, "", fmt.Errorf("BackupDownloadCredentialsSecret not set")
 		}
 
-		// create BackupDownloadCredentialsSecret containing values from kube-system/s3-credentials / kube-system/s3-settings
+		// create BackupDownloadCredentialsSecret containing values from kubermatic/s3 / kube-system/s3-settings
 
 		credsSecret := &corev1.Secret{}
-		if err := client.Get(ctx, types.NamespacedName{Namespace: metav1.NamespaceSystem, Name: EtcdRestoreS3CredentialsSecret}, credsSecret); err != nil {
+		if err := client.Get(ctx, types.NamespacedName{Namespace: KubermaticNamespace, Name: EtcdRestoreS3CredentialsSecret}, credsSecret); err != nil {
```
lsviben: this should stay as `metav1.NamespaceSystem`, edit the above comment as well

charts/kubermatic/static/cleanup-container.yaml
```diff
@@ -29,10 +29,10 @@ env:
 - name: ACCESS_KEY_ID
   valueFrom:
     secretKeyRef:
-      name: s3-credentials
```
lsviben: should we change this for the Cleanup container? Its used for the old backups which use Minio so we should not touch it. The new ones use the Delete container.

Similar goes for the old store container

moadqassem: I can revert this in the chart

lsviben: also in the operator resources

moadqassem: yep, done already

PR #7836, opened 2021-09-16
revert updating metering configs

moadqassem:
This is a quick fix to prevent attempting to update the PVC size. PVCs cannot simply be resized by changing a field, as they are (generally) immutable. Some CSI's might support it, but it's certainly not supported everywhere.

Fixes #

**Documentation**:

```release-note
None
```


xrstf (approved):
/approve

But for this we also then need to remove the option from the dashboard, right?

PR #7850, opened 2021-09-21
Allow cilium as CNI plugin

wozniakjan:

Currently, we support only canal as a CNI plugin. This allows to specify different cluster CNI and is a first initial step of adding cilium as an alternative CNI addon.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/cluster/validation/validation.go
+++ b/pkg/webhook/cluster/validation/validation.go
@@ -37,9 +37,10 @@ import (
 )
 
 var (
-	supportedCNIPlugins        = sets.NewString(kubermaticv1.CNIPluginTypeCanal.String())
+	supportedCNIPlugins        = sets.NewString(kubermaticv1.CNIPluginTypeCanal.String(), kubermaticv1.CNIPluginTypeCilium.String())
 	supportedCNIPluginVersions = map[kubermaticv1.CNIPluginType]sets.String{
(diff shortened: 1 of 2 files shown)


pkg/webhook/cluster/validation/validation.go
```diff
@@ -37,9 +37,10 @@ import (
 )
 
 var (
-	supportedCNIPlugins        = sets.NewString(kubermaticv1.CNIPluginTypeCanal.String())
+	supportedCNIPlugins        = sets.NewString(kubermaticv1.CNIPluginTypeCanal.String(), kubermaticv1.CNIPluginTypeCilium.String())
 	supportedCNIPluginVersions = map[kubermaticv1.CNIPluginType]sets.String{
-		kubermaticv1.CNIPluginTypeCanal: sets.NewString("v3.8", "v3.19"),
+		kubermaticv1.CNIPluginTypeCanal:  sets.NewString("v3.8", "v3.19"),
+		kubermaticv1.CNIPluginTypeCilium: sets.NewString("1.10.4"),
```
rastislavs: shouldn't it rather be `v1.10.4` to match with the tag? https://github.com/cilium/cilium/releases/tag/v1.10.4

wozniakjan: good catch!

PR #7859, opened 2021-09-22
Copy original APIs

kron4eg:
This is a copy of original kubermatic CRDs that will be changed further in the follow-up PRs, including changing Group (from GVK)

xref #7866

```release-note
NONE
```


Makefile
```diff
@@ -123,7 +123,6 @@ endif
 lint:
 	golangci-lint run \
 		--verbose \
-		--build-tags "$(KUBERMATIC_EDITION)" \
```
xrstf, line 126: This would make it impossible to lint CE or EE.

xrstf (commented):
Commentatio!

Makefile
kron4eg: Turns out I've forgot to add ce/ee tags :D

kron4eg: Fixed

xrstf (approved):
/approve

PR #7864, opened 2021-09-22
Rename copy of kubermatic.k8s.io to kubermatic.k8c.io

### LINKED ISSUE #7867: Add a linter for new CRD json tags
**User Story**
Currently we have a mishmash of camelCase, snake_case and so on in the CRDs json tags. Make camelCase standard, and add a linter so we avoid these things in the future.

As the REST API types in some places use the specs of the CRDs directly, this change could affect it. To avoid changing it, add REST API types and mapping where needed.

**Acceptance criteria**
- [ ] camelCase for JSON tags in CRDs
- [ ] Linter check for each PR
- [ ] REST API stays the same

kron4eg:
Along the way:

* fix the json tags
* introduce linter to keep them fixed

xref #7866 #7868

Fixes #7867

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubermatic/v1/datacenter.go
+++ b/pkg/apis/kubermatic/v1/datacenter.go
@@ -80,15 +80,15 @@ type SeedSpec struct {
 	Datacenters map[string]Datacenter `json:"datacenters,omitempty"`
 	// Optional: This can be used to override the DNS name used for this seed.
 	// By default the seed name is used.
-	SeedDNSOverwrite string `json:"seed_dns_overwrite,omitempty"`
+	SeedDNSOverwrite string `json:"seedDNSOverwrite,omitempty"`
 	// NodeportProxy can be used to configure the NodePort proxy service that is
 	// responsible for making user-cluster control planes accessible from the outside.
-	NodeportProxy NodeportProxyConfig `json:"nodeport_proxy,omitempty"`
+	NodeportProxy NodeportProxyConfig `json:"nodeportProxy,omitempty"`
 	// Optional: ProxySettings can be used to configure HTTP proxy settings on the
 	// worker nodes in user clusters. However, proxy settings on nodes take precedence.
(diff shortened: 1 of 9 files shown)


pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -412,22 +412,22 @@ type NodeSettings struct {
...
 	// makes use of this option.
-	HyperkubeImage string `json:"hyperkube_image,omitempty"`
+	HyperkubeImage string `json:"hyperkubeImage,omitempty"`
 }
 
 // SeedMLASettings allow configuring seed level MLA (Monitoring, Logging & Alerting) stack settings.
 type SeedMLASettings struct {
 	// Optional: UserClusterMLAEnabled controls whether the user cluster MLA (Monitoring, Logging & Alerting) stack is enabled in the seed.
-	UserClusterMLAEnabled bool `json:"user_cluster_mla_enabled,omitempty"`
+	UserClusterMLAEnabled bool `json:"userClusterMlaEnabled,omitempty"`
```
xrstf: As discussed in our meeting, let's turn this into an extendable struct, like

```yaml
seedMLA: # remove "Settings" from the key, as it's obvious we're talking about settings here
  userCluster: # also no need to call it "userClusterMLA" when it's already a child of `seedMLASettings` i guess
    enabled: true
```

kron4eg: This PR is only about renaming tags and group.

P.S.
The tag seems to be renamed incorrectly, shoulda be `userClusterMLAEnabled`.

kron4eg: And I don't think we can do API structure changes like this, before we start using this new API. Otherwise we will run into hell trying to synchronize "old" API with new. There has to be only 1 API :)

xrstf (approved):
/approve

PR #7865, opened 2021-09-22
Support EnabledOperatingSystems for Datacenters

### LINKED ISSUE #7842: Configure available OS images per Datacenter
**User Story**
As a KKP admin, I would like to configure the available Operating Systems per datacenter. I was not able to remove certain images from the available OS. For example, in Openstack datacenters RHEL is defined but this is not an Openstack requirement. Accordingly, there are datacenters that cannot serve selected operating systems.

Is there a way to configure available images in KKP?

If not, this should be implemented e.g. by filtering out Operating Systems that are not configured.

**Acceptance criteria**
Make available OS images configurable.

Happy2C0de:
Add support for EnabledOperatingSystems for Datacenters.

Currently, you filter OS on provider what is not very general. There are private Clouds which belong to a certain Provider but it doesn't serve your OS configurations. E.g. not all Openstack DCs server Flatcar images.

Accordingly, I introduce an option `datacenter.spec.enabledOperatingSystems` to configure available OS per Datacenter. This is more general and allows the KKP admins to define their own supported OSs.

Fixes #7842 

I keep the backwards compatibility and check the OS also while creating MDs. This is critical.
I added a new TestSeed `regular-do-limited-os` in order to write valuable checks for `EnabledOperatingSystems`

**Documentation**:

```release-note
Add support for EnabledOperatingSystems which configures available OS per datacenter.
```


Happy2C0de:
@moadqassem 
Thanks for having a look at this PR. Yes, you are right and your points are very valuable. 👍 

I thought about several implementations and ended up with this PR.

- Afaik, the first statement is not completely true. The machine-controller checks if the image is present when the manifest is applied, see: [provider.go#L448](https://github.com/kubermatic/machine-controller/blob/fa0c9e3195e8d87925e2e0b05248a4bcdcb70f31/pkg/cloudprovider/provider/openstack/provider.go#L448) 
That means, if KKP applies the configuration it leads to a valid cluster from KKP point of view, but actually it **heavily** fails and it produces a lot of errors. Actually, it's a /bug that I want to fix here. :)
- The second point is good, and somehow wanted, not? 
As an admin, you would like to control which images the users can use. First, it is a matter of operational costs because as an admin you would like to control the "supported" worker OS. Secondly, it also has security and compatibility implications. What if a user would like to deploy a cluster on Ubuntu 12.04? I am pretty sure that the KKP provisioning mechanism would fail on this. Also KKP has OS requirements in order to limit the supported versions. KKP Operators should have a mechanism to forward this requirement to the end users.
- Buttom-up, you mean with the machine-controller? I apply already fails in the machine-controller, what's good! Maybe, it would be worth thinking about defining the supported OS in the `machine-controller` (e.g. as a variable). But, I don't have a strong feeling about that.

I also thought about redefining the images key in the Seed object. See: [zz_generated.seed.yaml#L142](https://github.com/kubermatic/kubermatic/blob/50ee269602f89fcd2715f1f43bf53f373752ab49/docs/zz_generated.seed.yaml#L142), e.g. change type string to `[]string` and take the first element as the default image. This would be a different implementation approach.

How are we going to address this?

moadqassem:
Here are some thoughts about the points you mentioned and how maybe we can proceed. FYI: we are actually working on a solution that makes the operating system management more dynamic, resilient and can be customised, but that's a tail for another day.   

> * Afaik, the first statement is not completely true. The machine-controller checks if the image is present when the manifest is applied, see: [provider.go#L448](https://github.com/kubermatic/machine-controller/blob/fa0c9e3195e8d87925e2e0b05248a4bcdcb70f31/pkg/cloudprovider/provider/openstack/provider.go#L448)
That means, if KKP applies the configuration it leads to a valid cluster from KKP point of view, but actually it **heavily** fails and it produces a lot of errors. Actually, it's a /bug that I want to fix here. :)

The image is something and the os is something else. What I meant is, the current structure in the machine controller is mapping an OS that is supported by a cloud provider to the right image. Thus, you might add(for example) that RHEL isn't supported in Openstack however Machine controller says it does( at least the current state that reflects its capabilities) thus you can simply edit MD and make the os field RHEL and that will pass. And no this is not a bug, it was designed that way. Machine controllers job is to create, bootstrap and provision a machine, not to make sure that the user, uses the right image, this is something that documentations, defaulting and guiding can assist them. Again, please keep in mind that images and os have different notions(to the least in KKP).  

> * The second point is good, and somehow wanted, not?
>   As an admin, you would like to control which images the users can use. First, it is a matter of operational costs because as an admin you would like to control the "supported" worker OS. Secondly, it also has security and compatibility implications. What if a user would like to deploy a cluster on Ubuntu 12.04? I am pretty sure that the KKP provisioning mechanism would fail on this. Also KKP has OS requirements in order to limit the supported versions. KKP Operators should have a mechanism to forward this requirement to the end users.

Then what's the point of having the filed custom image? which is also wanted by users. Taking your example in consideration( users will use ubuntu 12.04), it is not the job of KKP and( the admin) to administer what kinda image is gonna be used by the users. A user should understand those vulnerabilities and comply by them and if they are not sure, then they should use the default images. Also we have documentation for what OS is supported and should be used that can be a reference for the users to decide. If the user decides to use a malfunction image for whatever reason, his cluster will simply won't successfully be provisioned and the machines eventually will be rotated by the machine controller up until the user uses the right image.
   
> * Buttom-up, you mean with the machine-controller? I apply already fails in the machine-controller, what's good! Maybe, it would be worth thinking about defining the supported OS in the `machine-controller` (e.g. as a variable). But, I don't have a strong feeling about that.

Well, that's what we do anyway, the os that we support are hardcoded in each cloud provider. So yeah that's already done and unfortunately it is not that easy to sync with what the admin specify in KKP and a) how we can translate that in the machine, b) how can they be synced and migrated, for older machines.

> I also thought about redefining the images key in the Seed object. See: [zz_generated.seed.yaml#L142](https://github.com/kubermatic/kubermatic/blob/50ee269602f89fcd2715f1f43bf53f373752ab49/docs/zz_generated.seed.yaml#L142), e.g. change type string to `[]string` and take the first element as the default image. This would be a different implementation approach.
I wouldn't strongly recommend this as we are moving away from hardcoded configs and moving towards declarative style more than imperative style which is similar to the idea that you pitch via this PR. 

> How are we going to address this?
As I said in the beginning of this comment, atm we are actively working on a solution which have a better approach of handling OSes. Also a very important note to have in mind is, how those configurations that you added in the seed would be treated by the user-data plugins: 
https://github.com/kubermatic/machine-controller/tree/master/cmd/userdata

shaase-ctrl:
@Happy2C0de 

We want to release KKP 2.19 on 20th of January 2022. This is not final, but gives you a clear timeline hopefully.

moadqassem:
> @moadqassem Great analysis! This is very appreciated and many thanks for your thoughts! 👍
> 
> I completely agree with your view on this. It's definitively not an easy task to do it properly and yes, images and OSs are not the same thing. The `bug` statement was more meant in the view of KKP, not the machine-controller. Because provisioning of a cluster fails when people select an image which is not present in the underlying provider. Yes, you can say that this is an end user problem, but a smoother experience would certainly be beneficial for the whole project. I really like to hear that you are working on that!
> 
> Is there a target release for the new OS management feature?
> 
> A low hanging fruit might be to implement another endpoint like `/api/v1/providers/openstack/images` that returns a list of images which can be displayed in the UI as a drop-down selection? Not sure, if this is worth the effort if you redesign the OS management anyway.

@Happy2C0de Glad that I was able to shed a light on the challenges regarding OS management in KKP. As what @shaase-ctrl mentioned we are working on a trade-off between flexibility and reliability by the operating system management feature. Thanks again for your contribution, highly appreciated :)

PR #7894, opened 2021-09-29
Added option to specify a static ip for nginx ingress loadBalancer

dharapvj:
Nginx-ingress controller assigns random IP address to the LoadBalancer. So we cannot pre-create DNS record for kkp management URL endpoint.

In enterprise environments where DNS entries can take time, its important that we have control on IP address in advance so that DNS requests can be made even before the installation.

This PR adds ability to optionally specify a static ip address to the loadbalancer. Upstream chart already has [this feature](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/templates/controller-service.yaml#L25-L27)

NA

**Documentation**:

Should this be documented?

```release-note
NONE
```


charts/nginx-ingress-controller/templates/service.yaml
```diff
@@ -26,6 +26,9 @@ metadata:
     {{- toYaml .Values.nginx.service.annotations | nindent 4 }}
 spec:
   type: LoadBalancer
+{{- if .Values.nginx.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.nginx.service.loadBalancerIP }}
```
wurbanski, line 30: IMHO put the IP in quotes, just in case :)

wurbanski (dismissed):

xrstf:
> In enterprise environments where DNS entries can take time, its important that we have control on IP address in advance so that DNS requests can be made even before the installation.

I don't quite understand that usecae. Can you explain further?

charts/nginx-ingress-controller/templates/service.yaml
xrstf: And you can avoid repeating the key by doing

```
{{- with .Values.nginx.service.loadBalancerIP }}
  loadBalancerIP: {{ . | quote }}
{{- end }}
```

dharapvj: I just copied [same code from upstream](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/templates/controller-service.yaml#L25-L27) so that it's easier to maintain, if the need arises later to match upstream chart. 

If you still insist, I can make that change.. but it seems nitpicking to me ATM.

dharapvj:
> I don't quite understand that usecae. Can you explain further?
Sure @xrstf .

We would like to setup a new KKP install with master-seed combo in Azure (private cloud) and another seed in on-prem infra (virtualized). Lets call them azure and house. So for both seed to have HA master setup via kubeone, we would need to request a single VIP and a DNS.

e.g. 
azure.acme.com A 10.1.2.3
and
house.acme.com A 10.99.2.3

Kubeone install fails if we don't do these DNS requests earlier on. With those DNS records in place, kubeone install on both seeds will be take care of. But to do KKP install on azure, we will need another two DNS records before KKP dashboard can be accessed

kkp.acme.com A 10.5.6.7
*.kkp.acme.com A 10.5.6.7

Now as with enterprise customers, all such DNS requests take time.. so we would like to send all 4 DNS requests in a batch and get them done in one go. For this, we need the 10.5.6.7 address to be known in advance to provide the DNS configuration team along with other ip addresses.

This PR allows me to pre-emptively set that IP address for nginx so that I can get all DNS requests in place and be confident that when I install kubeone and kkp, everything will just work.

Hope that clarifies.

charts/nginx-ingress-controller/templates/service.yaml
wurbanski: Ok, the IP was quoted in older version of the charts, now it isn't so perhaps it is safe ;-) I'll resolve and lgtm

PR #7896, opened 2021-09-29
Fixed issue where EtcdBackupConfigs ID was equal to its name

### LINKED ISSUE #7838: Backup Details show different cluster-id
**User Story**
The created cluster has automated backups `default-backup` shown, with the cluster id:

backup

but when we click on the backup for that backup detail,the backups are from a different cluster and it happens for all the other clusters
backup1

This only happens for the same named backup like `default-backup`  

**Acceptance criteria**
The backup detail of  `default-backup` for a cluster should show the backups for that cluster only.

lsviben:
EtcdBackupConfigs API set the EBC name as both the name and ID in the API object. While this worked well when managing EBCs for a single cluster, when we introduced Project-wide EBC listing, this caused issues as the expected behavior would be that IDs are unique. 

This PR introduces a fix for that where IDs are now `<clusterID>-<EBCName>`. 

Fixes #7838 

```release-note
Fix for issue with EtcdBackupConfigs from multiple clusters having the same ID, which was causing a bug in the UI
```

### DIFF
--- a/pkg/handler/v2/etcdbackupconfig/etcdbackupconfig_test.go
+++ b/pkg/handler/v2/etcdbackupconfig/etcdbackupconfig_test.go
@@ -201,6 +201,20 @@ func TestGetEndpoint(t *testing.T) {
 			ExpectedHTTPStatusCode: http.StatusOK,
 			ExpectedResponse:       test.GenAPIEtcdBackupConfig("test-1", test.GenDefaultCluster().Name),
 		},
+		{
+			Name:               "get etcdbackupconfig with a complex cluster id",
+			EtcdBackupConfigID: etcdbackupconfig.GenEtcdBackupConfigID("test-1", "test-ab-cd"),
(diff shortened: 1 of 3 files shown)


pkg/handler/v2/etcdbackupconfig/etcdbackupconfig_test.go
```diff
@@ -147,21 +148,21 @@ func TestGetEndpoint(t *testing.T) {
 	}{
 		{
 			Name:               "get etcdbackupconfig that belongs to the given cluster",
-			EtcdBackupConfigID: "id-1",
+			EtcdBackupConfigID: etcdbackupconfig.GenEtcdBackupConfigID("test-1", test.GenDefaultCluster().Name),
```
zreigz, line 151: Can you also check situation when cluster name contains multiple `-` like `test-abc` and `test-ab-cd`

PR #7933, opened 2021-10-04
Fix LB expose strategy for LBs with external DNS names instead of IPs

### LINKED ISSUE #7108: Expose strategy LoadBalancer does not work in AWS
**User Story**

I tried to create a user cluster in AWS (seed cluster in aws as well)  with expose strategy loadbalancer but the user cluster was never provisioned and seed-controller-manager seemed to be stuck waiting for cluster to enter healthy condition. I am running kubermatic version `quay.io/kubermatic/kubermatic:v2.17.0` . 

After a quick research, the problem seems to be in the way, aws is populating k8s service with the lb details:
```
apiVersion: v1
kind: Service
metadata:
 (...)
  name: front-loadbalancer
 (...)
spec:
  clusterIP: <ip>
  externalTrafficPolicy: Cluster
  ports:
  - name: secure
    nodePort: 32092
    port: 443
    protocol: TCP
    targetPort: 443
  selector:
    app: nodeport-proxy-envoy
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - hostname: <aws lb hostname>
```

The hostname from status is not supported by kubernetes controller from seed-controller-manager: ([code](https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/resources/address/address.go#L105-L119)) and as consequence seed-controller-manager is silently failing: ([code](https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/controller/seed-controller-manager/kubernetes/resources.go#L73-L74))

**Acceptance criteria**

The loadbalancer expose strategy works in AWS

rastislavs:
Fixes LoadBalancer expose strategy for cases where load-balancers provide an external DNS name instead of an external IP, such as in AWS.

Fixes #7108

**Documentation**:

```release-note
Fix LoadBalancer expose strategy for LBs with external DNS names instead of IPs
```

### DIFF
--- a/pkg/resources/address/address.go
+++ b/pkg/resources/address/address.go
@@ -109,9 +109,7 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
 		if err := m.client.Get(ctx, nn, frontProxyLoadBalancerService); err != nil {
 			return nil, fmt.Errorf("failed to get the front-loadbalancer service: %v", err)
 		}
-		// Use this as default in case the implementation doesn't populate the status
-		frontProxyLBServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
-		// Supposively there is only one if not..Good luck
+		frontProxyLBServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP // default in case the implementation doesn't populate the status
 		for _, ingress := range frontProxyLoadBalancerService.Status.LoadBalancer.Ingress {
 			if ingress.IP != "" {
 				frontProxyLBServiceIP = ingress.IP
@@ -119,6 +117,9 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
 			if ingress.Hostname != "" {
 				frontProxyLBServiceHostname = ingress.Hostname
 			}
+		}
+		if len(frontProxyLoadBalancerService.Status.LoadBalancer.Ingress) > 1 {
+			m.log.Debugw("Multiple ingress values in LB status, the following values will be used", "ip", frontProxyLBServiceIP, "hostname", frontProxyLBServiceHostname)
 		}
 	}
 
@@ -159,7 +160,7 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
 	case kubermaticv1.ExposeStrategyLoadBalancer:
 		if frontProxyLBServiceIP != "" {
 			ip = frontProxyLBServiceIP
-		} else {
+		} else if frontProxyLBServiceHostname != "" {
 			var err error
 			// Always lookup IP address, in case it changes
 			ip, err = m.getExternalIPv4(frontProxyLBServiceHostname)
--- a/pkg/resources/address/address.go
+++ b/pkg/resources/address/address.go
@@ -101,7 +101,8 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
 		subdomain = m.seed.Spec.SeedDNSOverwrite
 	}
 
-	frontProxyLoadBalancerServiceIP := ""
+	frontProxyLBServiceIP := ""
+	frontProxyLBServiceHostname := ""
 	if m.cluster.Spec.ExposeStrategy == kubermaticv1.ExposeStrategyLoadBalancer {
 		frontProxyLoadBalancerService := &corev1.Service{}
 		nn := types.NamespacedName{Namespace: m.cluster.Status.NamespaceName, Name: resources.FrontLoadBalancerServiceName}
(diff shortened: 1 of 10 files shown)


pkg/resources/address/address.go
```diff
@@ -101,27 +101,35 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
...
-		frontProxyLoadBalancerServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
+		frontProxyLBServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
 		// Supposively there is only one if not..Good luck
 		for _, ingress := range frontProxyLoadBalancerService.Status.LoadBalancer.Ingress {
 			if ingress.IP != "" {
-				frontProxyLoadBalancerServiceIP = ingress.IP
+				frontProxyLBServiceIP = ingress.IP
+			}
+			if ingress.Hostname != "" {
+				frontProxyLBServiceHostname = ingress.Hostname
```
PratikDeoghare: This could be  simplified to
```
				frontProxyLBServiceIP = ingress.IP
				frontProxyLBServiceHostname = ingress.Hostname
```

pkg/resources/address/address.go
```diff
@@ -101,27 +101,35 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
...
 	if m.cluster.Spec.ExposeStrategy == kubermaticv1.ExposeStrategyLoadBalancer {
 		frontProxyLoadBalancerService := &corev1.Service{}
 		nn := types.NamespacedName{Namespace: m.cluster.Status.NamespaceName, Name: resources.FrontLoadBalancerServiceName}
 		if err := m.client.Get(ctx, nn, frontProxyLoadBalancerService); err != nil {
 			return nil, fmt.Errorf("failed to get the front-loadbalancer service: %v", err)
 		}
 		// Use this as default in case the implementation doesn't populate the status
-		frontProxyLoadBalancerServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
+		frontProxyLBServiceIP = frontProxyLoadBalancerService.Spec.LoadBalancerIP
 		// Supposively there is only one if not..Good luck
```
PratikDeoghare: nit: Supposively -> Supposedly 

maybe check that there is exactly one and error if not

rastislavs: I don't think we should error in that case - as long as we are able to retrieve one external IP / hostname, we should be good. But at least a warning could be a good idea.

pkg/resources/address/address.go
rastislavs: This goes back to the previous question. If we assume that there may be multiple Ingress's set, the check for emptiness is necessary to not overwrite previously retrieved value with an empty value.

pkg/resources/address/address.go
```diff
@@ -101,27 +101,35 @@ func (m *ModifiersBuilder) Build(ctx context.Context) ([]func(*kubermaticv1.Clus
...
 		for _, ingress := range frontProxyLoadBalancerService.Status.LoadBalancer.Ingress {
 			if ingress.IP != "" {
-				frontProxyLoadBalancerServiceIP = ingress.IP
+				frontProxyLBServiceIP = ingress.IP
```
PratikDeoghare, line 115: nit: check after this loop if at least one of frontProxyLBServiceHostname,  frontProxyLBServiceIP is set.

PratikDeoghare (approved):
Looks good. Left some nits. :+1:

pkg/resources/address/address.go
PratikDeoghare: Right. 

With the checks we can end up with inconsistent values for ip and hostname  in case of multiple ingresses. 

ingress1: ip: nil, hostname: h1
ingress2: ip: ip2, hostname: nil 

We end up with `frontProxyLBServiceIP = ip2` and `frontProxyLBServiceHostname = h1`.

Unlikely but possible :sweat_smile:

PratikDeoghare: We could pick the values from the first ingress that has one of the values set. This will avoid the problem of overwriting and inconsistencies. Currently we are picking the last one after going through the loop.

rastislavs: The thing is that I would still like to prioritize IP over hostname if we have it..
But frankly, I think we are trying to address a corner case that I have never seen :)

rastislavs: BTW, I think the only really consistent solution would be to collect lists of IPs and hostnames, sort them alphabetically and then use the first one - but I am still not 100% sure if I should go that way

pkg/resources/address/address.go
rastislavs: Here is the issue - that situation actually always happens with LoadBalancer services right after creation, before a CCM controller allocates an external IP / hostname for it. So not sure if we should produce errors in such case - maybe that was the reason for silently ignoring that before my PR here: https://github.com/kubermatic/kubermatic/blob/9174aa67ef847947c22f4b540faebf4a6bbbfc0e/pkg/controller/seed-controller-manager/kubernetes/resources.go#L76

rastislavs: OK, maybe I should do RequeueAfter instead of returning an error there - will rework it a bit.

pkg/resources/address/address.go
PratikDeoghare: I agree. Let's keep it the way it is. :+1:

PratikDeoghare (approved):

mfranczy (approved):
/approve
/lgtm

embik:
/cherry-pick release/v2.18

I just hit this problem with a 2.18 setup.

PR #7943, opened 2021-10-05
run e2e locally

### LINKED ISSUE #5817: Provide some docs and guidance to run the e2e tests locally
**User Story**

As a kubermatic contributor
I want to be able to run and debug CI tests locally
So that I can debug and fix my pull requests if CI fails

**Acceptance criteria**
- [ ] Documentation on how to run CI jobs locally

zreigz:
**What this PR does / why we need it**: scripts to run e2e locally

Fixes #5817

```release-note
NONE
```

### DIFF
--- a/hack/ci/testdata/seed.yaml
+++ b/hack/ci/testdata/seed.yaml
@@ -112,10 +112,8 @@ spec:
       country: DE
       spec:
         kubevirt:
-          dns_policy: "None"
           dns_config:
             nameservers:
-            - 10.96.0.10
             - 8.8.8.8
     syseleven-dbl1:
       country: DE
--- a/hack/ci/testdata/seed.yaml
+++ b/hack/ci/testdata/seed.yaml
@@ -115,6 +115,7 @@ spec:
           dns_policy: "None"
           dns_config:
             nameservers:
+            - 10.96.0.10
             - 8.8.8.8
     syseleven-dbl1:
       country: DE
--- a/hack/local/README.md
+++ b/hack/local/README.md
@@ -1,0 +1,115 @@
+# Run e2e tests on your local environment
+
+E2E tests are designed for verifying the functional correctness by replicating end-user behavior from app deployment.
+This describes how to run e2e tests in local environment.
+
+## Prerequisites
+
+ - install [kind](https://kind.sigs.k8s.io/)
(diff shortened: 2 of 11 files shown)


hack/local/README.md
```diff
@@ -0,0 +1,115 @@
...
+E2E tests are designed for verifying the functional correctness by replicating end-user behavior from app deployment.
+This describes how to run e2e tests in local environment.
+
+## Prerequisites
+
+ - install [kind](https://kind.sigs.k8s.io/)
+ - install [jq](https://stedolan.github.io/jq/download/)
+
+### dnsmasq
+Install and configure `dnsmasq` to set up local domain for the KKP installation.
```
mfranczy, line 17: If you have another dns resolver like systemd-resolved this can be problematic. Let me try some stuff how to handle both resolvers without disabling systemd-resolved. Will go back to you.

hack/ci/testdata/seed.yaml
```diff
@@ -115,6 +115,7 @@ spec:
           dns_policy: "None"
           dns_config:
             nameservers:
+            - 10.96.0.10
```
mfranczy: Not sure if this is needed. We discovered that systemd-resolved was overwriting ubuntu dns entries, but without policy `None` the dns service from k8s should be added to vm (for Centos it should work). I will check that.

hack/local/README.md
mfranczy: Here is a nice snippet how to make dnsmasq running together with systemd-resolved https://gist.github.com/frank-dspeed/6b6f1f720dd5e1c57eec8f1fdb2276df#full-dnsmasq-support-under-systemd-resolved, will check how it performs and if it's ok maybe we could add this to the document.

zreigz: I've tested it, works fine

hack/ci/testdata/seed.yaml
zreigz: I have been checking this, thanks

hack/local/README.md
zreigz: I will update documentation

zreigz: done

hack/ci/testdata/seed.yaml
zreigz: yes, you are right

PR #7957, opened 2021-10-07
Add static dex users as KKP admin

### LINKED ISSUE #6481: First Kubermatic user should be an admin
In case it already works this way feel free to close.

As discussed in the `sig-ui` meeting the admin panel that grants a lot of management features for the Kubermatic admin might not be available by default to the first user after installing Kubermatic. By default, the first Kubermatic user should have an admin flag set to true.

mate4st:

Add static dex users as KKP admin on installation

Fixes #6481

Based on: https://github.com/kubermatic/kubermatic/pull/7944

I left validation to the  dex chart as data is used there before hitting our customer installer code.

```release-note
NONE
```

### DIFF
--- a/pkg/install/stack/kubermatic-master/stack.go
+++ b/pkg/install/stack/kubermatic-master/stack.go
@@ -18,13 +18,16 @@ package kubermaticmaster
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"math/rand"
 	"path/filepath"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
 	"k8c.io/kubermatic/v2/pkg/install/helm"
 	"k8c.io/kubermatic/v2/pkg/install/stack"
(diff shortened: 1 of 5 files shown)


pkg/install/stack/kubermatic-master/stack.go
```diff
@@ -507,3 +514,52 @@ func showLoadBalancerDNSSettings(ctx context.Context, logger *logrus.Entry, kube
 
 	logger.Info("")
 }
+
+func createFirstUser(ctx context.Context, logger *logrus.Entry, kubeClient ctrlruntimeclient.Client, opt stack.DeployOptions) error {
+
+	staticUsers, ok := opt.HelmValues.Get(yamled.Path{"dex", "staticPasswords"})
```
moadqassem, line 520: I would strongly recommend that you actually add a validation for the staticPasswords block. Otherwise KKP will end up with no user/admin.

pkg/install/stack/kubermatic-master/stack.go
```diff
@@ -507,3 +514,52 @@ func showLoadBalancerDNSSettings(ctx context.Context, logger *logrus.Entry, kube
...
+
+	//No user provided, skipping
+	if !ok {
```
moadqassem, line 523: If there was no user you should log that and return an error. You can add a validation part already in this file pkg/install/stack/kubermatic-master/validation.go

mate4st: The staticPasswords user is not required for installation and also not recommended for production env:
https://github.com/kubermatic/kubermatic/blob/master/charts/values.example.ce.yaml#L39

moadqassem: But then how they can create an admin user? only by editing the user manually via kubectl ?

mate4st: Yes, i think this is the intended way for the initial user. After this you can assign users as admin in the UI. Also KKP users are first created after they logged in once.

moadqassem: Then I honestly don't understand why this is needed in the first place :D. Because you can simply add any connector without providing a staticPasword, Login and then assign the user using kubeconfig. And if this practice isn't recommend in production as you say, this whole functionality is wrong!

mate4st: Indeed, but i would argue its an enhancement for people who want to try KKP for the first time

moadqassem (approved):
/approve
/lgtm

WeirdMachine:
Note: the api test fails for real. It uses the dex user as non admin user which is now not true anymore.

moadqassem (approved):
/approve
/lgtm

Issue #7897, opened 2021-09-29
Rework KKP Helm charts to use upstream charts wherever possible


wurbanski:
**User Story**
Most of the charts in KKP are based on upstream Helm charts for specific products, forked at some point in time, and extended with customizations. The upstream charts very often contain changes that we later need to backport, which is problematic due to the scope of our customizations applied. To make maintenance easier, we should try to use upstream charts as dependencies (helm mechanism) for our own charts, which would later be extended with our own customizations and correct default values. In case of changes not covered by templating, we should either use [kustomize post-renderer](https://helm.sh/docs/topics/advanced/#post-rendering) or try to upstream our changes.
This effort should also include CI for our helm charts (at least static, "golden master" testing).

**Acceptance criteria**
* CI exists in place to avoid mistakes while porting the changes.
* All charts have been evaluated in terms of compatibility with upstream ones and updated wherever possible.
* Upgrade in-place is possible and all required changes are documented on a per-chart basis.

PR #7970, opened 2021-10-12
add vmware-exporter chart

mihiragrawal:
**What this PR does / why we need it**: To add basic monitoring of vsphere provider. This PR adds a chart named `vmware-exporter`, which adds metrics for Prometheus. According to the upstream chart below information will be get. 

- Basic VM and Host metrics
- Current number of active snapshots
- Datastore size and other stuff
- Snapshot Unix timestamp creation date

Fixes #

**Documentation**:

```
NONE
```

### DIFF
--- a/charts/monitoring/vmware-exporter/templates/_helpers.tpl
+++ b/charts/monitoring/vmware-exporter/templates/_helpers.tpl
@@ -3,7 +3,7 @@
 Expand the name of the chart.
 */}}
 {{- define "vmware-exporter.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- default .Chart.Name .Values.vmware-exporter.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
@@ -12,10 +12,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 If release name contains chart name it will be used as a full name.
 */}}
 {{- define "vmware-exporter.fullname" -}}
-{{- if .Values.fullnameOverride -}}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- if .Values.vmware-exporter.fullnameOverride -}}
+{{- .Values.vmware-exporter.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
-{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- $name := default .Chart.Name .Values.vmware-exporter.nameOverride -}}
 {{- if contains $name .Release.Name -}}
 {{- .Release.Name | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
--- a/charts/monitoring/vmware-exporter/templates/_helpers.tpl
+++ b/charts/monitoring/vmware-exporter/templates/_helpers.tpl
@@ -1,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "vmware-exporter.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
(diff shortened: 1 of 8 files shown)


charts/monitoring/vmware-exporter/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "vmware-exporter.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
```
wurbanski: **Note: this applies to any instance where you use `.Values.something`**

Since you updated the values.yaml to be prefixed with `vmware-exporter`, all instances of `.Values.xxx` in templates have to be changed to `.Values.vmware-exporter.xxx`

charts/monitoring/vmware-exporter/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+    #  readOnlyRootFilesystem: true
+    #  runAsGroup: 65532
+    #  runAsNonRoot: true
+    #  runAsUser: 65532
+  
+  # Set the Pod security context
+  podSecurityContext:
+    {}
+    #  fsGroup: 65532
+
```
wurbanski: nitpicking, but since there are some changes to do already: can you fix this newline situation here? :)

wurbanski (changes requested):
@mihiragrawal Can you describe the scope of changes you made to the original chart?

charts/monitoring/vmware-exporter/templates/_helpers.tpl
mihiragrawal: Thanks for pointing it out. Fixed.

charts/monitoring/vmware-exporter/values.yaml
mihiragrawal: Fixed.

xrstf:
> - Inject license information.

Would not be needed if we just used the upstream chart directly. 

> - add parent value string `vmware-exporter` in `value.yaml` to adjust values in KKP value file.

That is nice.

> - Update image tag to latest docker image available on [dockerhub](https://hub.docker.com/r/pryorda/vmware_exporter/tags).

We could simply contribute a one-line change to the upstream chart :)

> - Enabled svc deployment.
> - Update replica count to 2.

That can just be done via the values.yaml, this is not really a "change".

> - cleanup non required files, like `createRelease.sh` and `owner`.....

Would not hurt us if we just used the upstream chart.

I don't think we should fork the chart, as we're on our way to get rid of most of our forks and use upstream instead. I would really not want to introduce new charts that are immediately deprecated. Nothing of this is added to the KKP installer, so this chart would have to be installed manually anyway.

For these reasons I would strongly vote towards just recommending to use the upstream chart. It does bring a bit of inconvenience, because the upstream chart is not ready to be used with our old "single values.yaml for everything"-technique, but that technique is also on its way out as we begin to use more upstream charts.

toschneck:
> For these reasons I would strongly vote towards just recommending to use the upstream chart. It does bring a bit of inconvenience, because the upstream chart is not ready to be used with our old "single values.yaml for everything"-technique, but that technique is also on its way out as we begin to use more upstream charts.

That's pretty new to  me. So do we have a concept how we manage the dependency of a upstream chart? 
For me it just should be easy to use a "predefined" best practices stack, what should be versioned. Versioned tied charts are stable and help us to bring up customer setups quickly. 

@xrstf  how you will ensure that we have a standard set of upstream components (charts / operators etc.) easy manageable for our customer?

xrstf:
> ow you will ensure that we have a standard set of upstream components (charts / operators etc.) easy manageable for our customer?

That can be as easy as a `map[string]string` in the KKP installer, listing the charts and their version. Or a config file somewhere.

Also, I would not want us to extend the set of charts we support/recommend. We are not in the business of being a Helm chart broker and shipping tons of charts; every new chart is another dependency we need to carefully consider.

See also https://github.com/kubermatic/kubermatic/issues/7897 and our #dev-hackathon-helm-squad Slack channel for more info. I am not in the team that manages the transition from custom charts to upstream charts ;)

shaase-ctrl:
How can we support that there are no configs we "forget"?

xrstf:
Can you rephrase that question?

PR #7972, opened 2021-10-12
Set first user as admin

### LINKED ISSUE #6481: First Kubermatic user should be an admin

mate4st:

Elect the first authenticated user as an Admin.

Fixes #6481

**Documentation**:
https://github.com/kubermatic/docs/pull/824

```release-note
NONE
```


xrstf (approved):
/approve

PR #7986, opened 2021-10-12
CRD Migration Command

### LINKED ISSUE #7866: Migration script and documentation for KKP3 CRDs
**User Story**

When we switch to KKP3, we will change the API groups of our CRDs so that we move from the forbidden *.k8s.io and also that we unify our api groups:

kubermatic.k8s.io + operator.kubermatic.io -> kubermatic.k8c.io 

After the change, we wont support the old ones anymore.

Add a migration script and documentation for the migration to the new KKP CRDs. 

**Acceptance criteria**
- [ ] User clusters are working during the migration
- [ ] Migration script is ready and tested
- [ ] Migration is documented

xrstf:
This PR adds a new, hidden, `migrate-crds` command to the KKP installer, meant to perform the CRD migration as mentioned in #7866. The command is clearly marked as dev-only, as running it on an existing KKP setup could be fatal (though the extensive preflight checks make it pretty much impossible to accidentally cause problems).

The new command will, in order:

1. Assemble Kubernetes clients for master and all seed clusters.
2. Run preflight checks:
   * Make sure the seed clients work.
   * Check that all KKP controllers everywhere are turned off. This includes even usercluster-controller-managers.
   * Check that the new CRD files are present on disk.
   * Check that no existing resource is currently in deletion.
3. Create a backup. This is a tar.gz archive that contains a YAML dump of all KKP resources in all clusters. For brevity sake's, the `managedFields` field is removed to make the YAML nicer. Since the preflight checks already forbid DeletionTimestamps to be set, there is no need to remove those when writing the YAML. There is no option to disable backups. This is on purpose.
4. Migrate resources:
   * Install/update new CRDs
   * Duplicate all KKP resources into new API group, while rewriting ownerRefs and objectRefs accordingly.
5. Remove the resources in the old API group, unless `--keep-resources` is set. The CRDs however are not deleted, just in case.
6. Party time!

I chose to code all of this using typed kinds instead of some JSON-marshalling-magic. I believe that any attempt to create a generic converter would eventually lead to less, but much harder to understand code with many exceptions and edgecases. With the explicit approach, it's more code to type, but a) I enjoyed it, b) you don't have to retype it and c) this code is not going to be maintained in the long term, but will only live for 1 very special release.

There is a script to test all of this, `hack/run-crd-migration-in-kind.sh`, which will setup 2 kind clusters, fill them with the resources in `hack/ci/testdata/crdmigration` (the script is not meant for CI, but I still reused this `testdata` directory) and then run the migrator. There are not yet enough test files for all types and not all files have all options set, because I copied them from our dev environment, where we currently didn't have, for example, an `AdmissionPlugin` ^^

In addition to the migration, this PR also performs a few changes to the new API kinds:

* `KubermaticConfiguration`:
  * `spec.versions.kubernetes` has been moved to `spec.versions` (i.e. one level in the YAML has removed). We don't support anything but Kubernetes and so this special level can go away (previous KKP releases had `openshift` as another option).
  * `spec.addons` was treated the same way.
  * `spec.featureGates` is now a simple `map[string]bool`, which leads to a much simpler YAML structure.
* `Seed`, `Preset`:
  * the old `RequiredEmailDomain` has finally been removed and is merged into `RequiredEmailDomains`, which was also renamed to `RequiredEmails` to be in-sync with the presets.
* `KubermaticSettings`:
  * `ClusterType` is gone, as it was always `kubernetes`.
* `User`:
  * `TokenBlackListReference` has been renamed to `InvalidTokensReference`.

screenshot-2021-10-12-205435

screenshot-2021-10-20-213325

Fixes #7866

Don't fear the diff stat, most of this is just testdata YAML :)

**Documentation**:
Will need to be written once this puppy is released.

```release-note
NONE
```


kron4eg (approved):
/approve
/lgtm

❤️

Issue #6563, opened 2021-02-16
Support Account Audits


shaase-ctrl:
**User Story**
As a customer I want to easily audit all used or connected accounts so that I can periodically audit my platform. 

**Acceptance criteria**
- show active user accounts in UI
- show inactive user accounts in UI
- display a printable list of user accounts
- show the relation of accounts (parent, sibling etc.)

PR #7989, opened 2021-10-13
Add endpoint to list users

floreks:
- Adds new endpoint to list users - `/api/v2/users`
- Adds a new `lastSeen` field to the user crd spec
- Updates `UserSaver` middleware to update `lastSeen` date

Related to #6563

**Documentation**:

```release-note
Add an endpoint to list users and extend the user object to allow checking the last seen date.
```


pkg/controller/master-controller-manager/user-synchronizer/controller_test.go
```diff
@@ -104,6 +104,9 @@ func TestReconcile(t *testing.T) {
 					t.Fatalf("failed to get user: %v", err)
 				}
 			} else {
+				seedUser = sanitize(seedUser)
```
floreks, line 107: @xrstf could you PTAL? I couldn't force the controller to make an equal copy of the `lastSeen` field and it was always failing due to differences in the internal `location` field. If you've got a better idea of how to make this work, I will change this.

pkg/handler/middleware/middleware.go
```diff
@@ -184,6 +188,15 @@ func UserSaver(userProvider provider.UserProvider) endpoint.Middleware {
...
 				}
 			}
+
+			user.Spec.LastSeen = &[]metav1.Time{metav1.NewTime(Now().UTC())}[0]
+			user, err = userProvider.UpdateUser(user)
+			if err != nil {
+				return nil, common.KubernetesErrorToHTTPError(err)
+			}
+
+			fmt.Println(user.Spec)
```
floreks: Remove this :slightly_smiling_face:

pkg/handler/middleware/middleware.go
```diff
@@ -184,6 +188,15 @@ func UserSaver(userProvider provider.UserProvider) endpoint.Middleware {
 					}
 				}
 			}
+
+			user.Spec.LastSeen = &[]metav1.Time{metav1.NewTime(Now().UTC())}[0]
+			user, err = userProvider.UpdateUser(user)
+			if err != nil {
```
floreks, line 200: Do we have a check for the "object has been modified" error as it is not critical here? We can assume that another call just updated the time.

floreks: Found and updated

xrstf (approved):
/approve

PR #7990, opened 2021-10-13
[release/v2.16] Fix issue where $$ in the password causes machine-controller fails to deploy machines

kron4eg:
This is a cherry-pick of #7984

```release-note
Fixes a bug where `$$` in the environment-variables for machine-controller was interpreted in the Kubernetes Manifest and caused machine-controller to be unable to deploy resources, when for e.g. the password contains two consecutive `$` signs
```


xrstf (approved):
/approve

PR #7991, opened 2021-10-13
Add devcontainer

cedi:

Adding a .devcontainer configuration to kubermatic

  As per a Discussion I had with @kron4eg I decided to submit a
  .devcontainer configuration for kubermatic.

  If you never worked with Devcontainers before, here is a good artical by
  [Aaron Powell](https://www.aaron-powell.com/about) on why you should care about Devcontainers:
  [your open source project needs a devcontainer - here's why](https://www.aaron-powell.com/posts/2021-03-08-your-open-source-project-needs-a-dev-container-heres-why/)

  Tl;Dr:
  Devcontainers are awesome and they help to define an isolated
  development environment which is standardised across all contributors
  setups and leverages the beauty of [GitHub Codespaces](https://github.com/features/codespaces) once, it's GA.

  Just try it out once, before dismissing it.
  Setting up your VScode to use the Devcontainer is easy:
  1. Make sure you have the following plugin installed: `ms-vscode-remote.remote-containers`
  2. Open VScode in the root-directory of the Repository
  3. Click "Reopen in Container" once VScode shows you the popup, telling
  you that it detected a devcontainer configuration.

**Documentation**:
https://www.aaron-powell.com/posts/2021-03-08-your-open-source-project-needs-a-dev-container-heres-why/
https://code.visualstudio.com/docs/remote/containers
https://github.com/qdm12/godevcontainer
https://github.com/features/codespaces

```release-note
NONE
```


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# https://hub.docker.com/r/qmcgaw/godevcontainer
+FROM qmcgaw/godevcontainer
```
xmudrii, line 2: Is using `godevcontainer` image a requirement or we can use any image here?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apk add make curl bash
+
+RUN export GO111MODULE=on
+RUN export CGO_ENABLED=1
```
xmudrii: I'm not really sure do we need `CGO_ENABLED` enabled by default. As far as I see, we only use it when running race tests because it's required for those.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN go get -u github.com/go-delve/delve/cmd/dlv@master
+RUN go get -u honnef.co/go/tools/cmd/staticcheck
+RUN go get -u golang.org/x/lint/golint
+RUN go get -u github.com/fzipp/gocyclo/cmd/gocyclo
+RUN go get -u github.com/amitbet/gorename
```
xmudrii: Can't we depend on the VSCode to download all tools automatically?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
+// For format details, see https://aka.ms/vscode-remote/devcontainer.json or this file's README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.137.0/containers/go
+{
+	"name": "Go",
```
xmudrii: My understanding is that this is a name for the container. Can we call container something more recognize-able, such as `kubermatic-devcontainer` or something like that?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+	"build": {
+		"dockerfile": "Dockerfile",
+	},
+	"runArgs": [
+		"--cap-add=SYS_PTRACE",
+		"--security-opt",
+		"seccomp=unconfined"
+	],
```
xmudrii, line 13: As far as I understand, this is required for delve to work, right? Maybe it would be useful to leave a comment if that's the case.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"go.toolsManagement.autoUpdate": false,
```
xmudrii: Does it make sense and is it possible to enable this? This is related to my earlier comment — if I get it right, VSCode should update the required tools automatically if this is enabled.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"go.toolsManagement.checkForUpdates": "off",
+		"go.gopath": "/go",
+		"go.lintOnSave": "package",
```
xmudrii: By default, `lintOnSave` uses `staticcheck`, however, we use `golangci-lint` in the CI. We should also set `go.lintTool` to `golangci-lint` so that lint recommendations are consistent with the CI.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"go.vetOnSave": "package",
+		"go.formatTool": "goimports",
```
xmudrii: Is there some specific reason why we don't use the default format tool?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"go.useLanguageServer": true,
+		"go.autocompleteUnimportedPackages": true,
```
xmudrii: As per [VSCode docs](https://github.com/golang/vscode-go/blob/master/docs/settings.md#goautocompleteunimportedpackages), this is not applicable when using the language server.

> Include unimported packages in auto-complete suggestions. Not applicable when using the language server.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+			"testPackage": false,
+			"generateTestForFunction": true,
+			"generateTestForFile": false,
+			"generateTestForPackage": false,
+			"addImport": true,
+			"testCoverage": true,
+			"playground": true,
+			"debugTestAtCursor": true,
+			"benchmarkAtCursor": false
+		},
```
xmudrii: What's the reason for overriding this instead of using default values?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"go.enableCodeLens": {
+			"references": true,
+			"runtest": true
+		},
```
xmudrii: Similar to above, what's the reason for overriding this instead of using default values?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"terminal.integrated.defaultProfile.linux": "zsh",
+	},
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		// Language Specific Plugins
+		"golang.Go",
+		"ms-azuretools.vscode-docker",
+		// Linters, Formaters, Highlighters, ...
+		"davidanson.vscode-markdownlint",
```
xmudrii: I'm not really sure do we need to lint markdown files, but it might be useful.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"shardulm94.trailing-spaces",
+		"redhat.vscode-yaml",
+		// Working with the Repository
+		"github.vscode-pull-request-github",
+		"github.remotehub",
+		"eamodio.gitlens",
+		// Utilities
+		"bierner.emojisense",
+		"redhat.vscode-yaml",
```
xmudrii: We already define this extension on L70.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,88 @@
...
+		"IBM.output-colorizer",
+		"lyyyuna.prow-sink",
```
xmudrii, line 43: TIL this extension exists.

xmudrii (changes requested):
Overall, this looks okay and I think this might be useful. I left a bunch of comments and questions, but I think we can merge this after comments are addressed.

.devcontainer/devcontainer.json
cedi: fixed

.devcontainer/Dockerfile
cedi: you can use any Docker Container. I just use `qmcgaw/godevcontainer` because it comes with go1.17 out of the box and already has many go tools already pre-installed

.devcontainer/devcontainer.json
cedi: I like it when I press save (with format on save enabled) and my imports are sorted and automatically imported as well

.devcontainer/devcontainer.json
cedi: I just like explicit configuration over "default". This is what works for me, but I'm happy to remove this config.

.devcontainer/devcontainer.json
cedi: there is a Readme.md and some docs are in markdown ;)

.devcontainer/Dockerfile
xmudrii: I'm wondering does it make sense to use [our build image](https://quay.io/repository/kubermatic/build) here because it contains all the binaries you may need to develop and run KKP. However, a potential problem is that the build image is huge — close to 1 GB. @xrstf Do you have any opinions on this?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+		"ms-kubernetes-tools.vscode-kubernetes-tools"
+	],
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [9000],
+	// Use 'postCreateCommand' to run commands after the container is created.
+	//"postCreateCommand": "make tools",
+	// Uncomment to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root.
+	//"remoteUser": "vscode"
+}
```
xmudrii: Please add a new line.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# https://hub.docker.com/r/qmcgaw/godevcontainer
+FROM qmcgaw/godevcontainer
+
+RUN apk add make curl bash
+RUN export GO111MODULE=on
```
xmudrii: Is there any specific reason why we don't use `ENV GO111MODULE=on` instead?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+	"settings": {
+		"go.toolsManagement.autoUpdate": true,
+		"go.toolsManagement.checkForUpdates": "off",
+		"go.gopath": "/go",
+		"go.lintOnSave": "package",
+		"go.lintTool": "golangci-lint",
+		"go.vetOnSave": "package",
+		"go.formatTool": "goimports",
+		"go.useLanguageServer": true,
+		"go.gotoSymbol.includeImports": true,
```
xmudrii: As per the [VSCode Go plugin docs](https://github.com/golang/vscode-go/blob/master/docs/settings.md#gogotosymbolincludeimports), this is not applicable when using the language server:

> If false, the import statements will be excluded while using the Go to Symbol in File feature. Not applicable when using the language server.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+		"go.gotoSymbol.includeGoroot": true,
```
xmudrii: As per the [VSCode Go plugin docs](https://github.com/golang/vscode-go/blob/master/docs/settings.md#gogotosymbolincludegoroot), this is not applicable when using the language server:

> If false, the standard library located at $GOROOT will be excluded while using the Go to Symbol in File feature. Not applicable when using the language server.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+		"codeQL.telemetry.enableTelemetry": false,
+		"codeQL.runningQueries.memory": 2048,
+		"codeQL.runningQueries.debug": true,
```
xmudrii: I don't see that we use [codeQL extension](https://marketplace.visualstudio.com/items?itemName=github.vscode-codeql). Does it make sense to set those values if we don't use the extension itself?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+		"git.autofetch": true,
+		"editor.renderFinalNewline": true,
+		"editor.codeLens": true,
+		"editor.insertSpaces": true,
+		"editor.formatOnSave": true,
+		"editor.formatOnPaste": true,
+		"editor.snippetSuggestions": "none",
```
xmudrii: Why did you disable snippets? I think snippets can be useful, so I think we should leave them on.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+		"editor.codeActionsOnSave": {
+			"source.organizeImports": true,
+			"source.fixAll": true,
+		},
```
xmudrii, line 26: Are those options provided by VSCode or some plugin? I can't find anything about those options in the VSCode docs.

.devcontainer/devcontainer.json
xmudrii: Of course we do have markdown files, but I'm not really fond of idea to lint markdown files, although, folks might have different opinions. :slightly_smiling_face:

xmudrii (changes requested):

.devcontainer/Dockerfile
cedi: none other than me just being lazy ;)

.devcontainer/devcontainer.json
cedi: I actually don't know it for sure either. I just know I have this enabled in all my Go-Projects and it works :D 
It sorts my imports and I like that a lot actually. But I have no feelings for that

.devcontainer/devcontainer.json
xmudrii: `go.lintOnSave` is `package` by default.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+		"seccomp=unconfined"
+	],
+	// Set *default* container specific settings.json values on container create.
+	"settings": {
+		"go.toolsManagement.autoUpdate": true,
+		"go.toolsManagement.checkForUpdates": "off",
+		"go.gopath": "/go",
+		"go.lintOnSave": "package",
+		"go.lintTool": "golangci-lint",
+		"go.vetOnSave": "package",
```
xmudrii: `go.vetOnSave` is `package` by default.

.devcontainer/devcontainer.json
xmudrii: I think we can use the default tool without specifying it.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+		"go.formatTool": "goimports",
+		"go.useLanguageServer": true,
```
xmudrii: `go.useLanguageServer` is `true` by default.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+		"git.autofetch": true,
+		"editor.renderFinalNewline": true,
```
xmudrii: `editor.renderFinalNewline` is `true` by default.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+		"editor.codeLens": true,
```
xmudrii: `editor.codeLens` is `true` by default.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,59 @@
...
+		"editor.insertSpaces": true,
```
xmudrii: `editor.insertSpaces` is `true` by default.

xmudrii (changes requested):
There are a few variables set explicitly to their default values. I think we shouldn't specify variables if default values are satisfying. This makes the config easier to maintain.

xmudrii:
@cedi There’s a conflict — we can’t merge the PR unless it’s rebased.

cedi:
@xmudrii rebased it

xmudrii (approved):
Thanks for the contribution, @cedi!

/lgtm
/approve
/test all

xmudrii:
```
[2021-11-29T08:34:51] Checking Kubermatic CE licenses...
time="Mon, 29 Nov 2021 08:34:51 UTC" level=warning msg="File is smaller than its boilerplate." file=.devcontainer/Dockerfile
time="Mon, 29 Nov 2021 08:34:52 UTC" level=error msg="1 out of 1469 checked files is invalid."
```

@cedi This is a legitimate failure. Can you please take a look into this?

cedi:
will have a look

PR #7992, opened 2021-10-13
[WIP] Proposal for seed cluster templates

mate4st:

Proposal for Seed Cluster Templates

```release-note
NONE
```


docs/proposals/seed-cluster-templates.md
```diff
@@ -0,0 +1,46 @@
...
+```
+	// DefaultClusterTemplate is the name of a cluster template that is used to default newly created cluster.
+	DefaultClusterTemplate string `json:"defaultClusterTemplate"`
+```
+
+* With the ClusterTemplate used we can easily extend defaulting for user clusters and be in sync with the cluster spec.
+* It is possible to extend this feature and provide a set of default templates for different providers.  
+* DefaultClusterTemplate is supposed to be optional, in that case we have hardcoded defaults which are used by the cluster webhook.
+* Enforcing settings could be archived by defining an enforcing ClusterTemplates.
+* DefaultComponentSettings can be deprecated and used from the DefaultClusterTemplate.
```
xrstf: Extra question, what happens if the ClusterTemplate is missing or broken?

mate4st: * DefaultClusterTemplate is supposed to be optional, in that case we have hardcoded defaults which are used by the cluster webhook.

xmudrii: I have three concerns related to this:

* For example, let's say someone deletes the ClusterTemplate. If we just use the hardcoded defaults, how is an admin going to know that the ClusterTemplate is missing? Everything could work just fine, but still, the settings could be wrong. I think this can cause huge problems later and be quite hard to debug.
* What about other settings? The `DefaultComponentSettings` struct has a bunch of other settings that are not present in ClusterTemplates or the UI?
* What about permissions? Right now, the Seed object can be only modified by KKP admins that have the seed kubeconfig. Who can modify Cluster Templates, i.e. is it only admins or users as well?

mate4st: - A finalizer could prevent deletion if it is used in the seed spec. Or a flag is provided that enforces the template to be present. Or we require the template to be present and remove hardcoded defaults.
- ComponentSettings are used in the cluster.spec therefore everything is available in a ClusterTemplate. 
```
	// Optional component specific overrides
	ComponentsOverride ComponentSettings `json:"componentsOverride"
```
- Only admins with kubeconf can edit. This can be ensured due to the ClusterTemplate scope. So the UI can differentiate between the different types.

xmudrii: > A finalizer could prevent deletion if it is used in the seed spec. Or a flag is provided that enforces the template to be present. Or we require the template to be present and remove hardcoded defaults.

I wouldn't require Cluster Templates unless we provide a default Cluster Template. Finalizers are nice, but they can easily be removed, and then we have the same problem that an admin might be using the template which doesn't exist.

We could do something like this:

* Make Cluster Template required but provide a default one that would include hardcoded defaults
* Leave Cluster Template as optional, but fail to create a cluster if it's provided but doesn't exist (I believe this is the flag approach you mentioned, although, I wouldn't make it configurable)

> ComponentSettings are used in the cluster.spec therefore everything is available in a ClusterTemplate.

> Only admins with kubeconf can edit. This can be ensured due to the ClusterTemplate scope. So the UI can differentiate between the different types.

Okay, this makes sense.

PR #7997, opened 2021-10-13
Proposal for backups destination management

### LINKED ISSUE #7996: Design and proposal for backup destination management

lsviben:
Proposal for backup destination management

Fixes #7996 

```release-note
NONE
```

### DIFF
--- a/docs/proposals/backup_destination_management.md
+++ b/docs/proposals/backup_destination_management.md
@@ -1,0 +1,121 @@
+**ETCD Backup Destination Management**
+
+**Author**: Lovro Sviben
+
+**Status**: Draft proposal. Prototype in progress
+
+## Goals
+
+Allow managing multiple etcd backup destinations per seed.
+
+## Non-Goals
+
+This does not include supporting other providers, our backups can still only be used with S3-compatible clients.
+
+## Motivation and Background
+
+Since 2.18, we support automatic etcd backups and restores to an S3-compatible bucket. The backup are enabled and its
+destination are set in the Seed Object. The destination consists of a bucket name, and an endpoint, paired with a 
+corresponding secret which contains the credentials. So all backups for Cluster within this seed end up in one bucket.
+
+The goal of this feature is to provide an option to:
+- allow admins to set up multiple destinations per Seed (with their credentials)
+- When creating Etcd Backup Configs, Cluster admins can choose destination. based on the Cluster Seed
+
+## Implementation
+
+### Seed extension
+Seed has the following Backup data:
+
+```go
+// SeedBackupRestoreConfiguration are s3 settings used for backups and restores of user cluster etcds.
+type SeedBackupRestoreConfiguration struct {
+	// S3Endpoint is the S3 API endpoint to use for backup and restore. Defaults to s3.amazonaws.com.
+	S3Endpoint string `json:"s3Endpoint,omitempty"`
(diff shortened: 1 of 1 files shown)


docs/proposals/backup_destination_management.md
```diff
@@ -0,0 +1,121 @@
...
+	S3BucketName string `json:"s3BucketName,omitempty"`
+}
+```
+
+As we want to avoid breaking compatibility with the older version, we can deprecate these fields and in addition add a 
+map of BackupDestinations. When giving users options for backup destinations, add the old ones as well for now.
+
+```go
+// SeedBackupRestoreConfiguration are s3 settings used for backups and restores of user cluster etcds.
+type SeedBackupRestoreConfiguration struct {
```
moadqassem, line 45: What about adding a new field a type map instead of having two types which are exactly the same. So just adding that new field in the seed type and do a transformation. When the new version is applied have a modifier, reading from the old filed and append it to the new one. Once we make sure that all users have already migrated to the new field we can get rid of the old one.

docs/proposals/backup_destination_management.md
```diff
@@ -0,0 +1,121 @@
...
+
+```
+
+### EtcdBackupConfig extension
+
+The EtcdBackupConfig also needs to be extended with the backup destination instead of just using the Seed one. 
+
+```go
+// EtcdBackupConfigSpec specifies details of an etcd backup
+type EtcdBackupConfigSpec struct {
```
moadqassem, line 73: Why not adding this type directly in the cluster object? Instead of referencing the cluster.

docs/proposals/backup_destination_management.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Controllers
+Furthermore, we will need to update the backup and restore controllers to support the destinations, but that should be 
+fairly simple.
+
+## Constraints
+
+* If an admin removes/modifies a destination, it doesn't affect already existing EtcdBackupConfigs, unless it also changes its secret. 
+This could make the backups fail.
+* Changing a backup destination for an etcd backup config should not be allowed, as it could break restoring from previously made backups.
```
moadqassem, line 112: So far, we didn't add an admission webhooks(IIRC) we can add some, in order to prevent any of unwanted behaviour similar to this one.

docs/proposals/backup_destination_management.md
lsviben: thats a good idea. 

My only concern is that we could need new config options for backups. I would suggest putting this type map in an struct so we can add more stuff if needed. 

So something like :

```
type EtcdBackupConfiguration struct {
    BackupDestinations map[string]*SeedBackupRestorationConfiguration
}
```
as a part of the seed type. Other stuff we can do as you suggest, wdyt?

docs/proposals/backup_destination_management.md
lsviben: Not sure what you mean exactly. The EtcdBackupConfigSpec already exists with all these fieds except for the Destination that I want to add.

docs/proposals/backup_destination_management.md
lsviben: @moadqassem  
Check this out, this is how I envision it: https://github.com/kubermatic/kubermatic/pull/8021

moadqassem (approved):
/approve
/lgtm

PR #8014, opened 2021-10-18
Set --kubelet-certificate-authority to validate kubelet serving certs

embik:
Kubelet exposes its own API that the Kubernetes API server uses to pull information from the kubelet (for example, accessing Pod logs when a user runs `kubectl logs`). In KKP, the TLS certificates for this kubelet API are generated from CSRs and thus are issued by the Kubernetes CA.

By default, `kube-apiserver` does not validate the TLS certificates that kubelets serve on this API. See https://kubernetes.io/docs/concepts/architecture/control-plane-node-communication/#apiserver-to-kubelet. This PR adds a flag to `kube-apiserver` to validate that kubelets only serve TLS certificates issued by the CA.

Fixes https://github.com/kubermatic/edgematic/issues/50

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #8015, opened 2021-10-18
use Recreate strategy for KKP operator deployment

xrstf:
During KKP upgrades, we currently had the case that multiple operator pods could be alive at the same time. Since there is no leader election (because the operator is supposed to run with a single pod only, as a bit of downtime in the operator is absolutely harmless for a running KKP setup), these 2 operators would fight each other.

This PR changes the Deployment to use the Recreate strategy, ensuring that Kubernetes kills old pods before it creates new ones.

```release-note
NONE
```


embik (approved):
/lgtm

PR #8021, opened 2021-10-18
Extend backups for destinations

### LINKED ISSUE #7938: Expand EtcdBackupConfig with a destination field
**User Story**
To support multiple backup destinations per Seed, we need to expand the EtcdBackupConfig with Destination information, so the controller knows where to store the backups

lsviben:
Extends Seed with the option of multiple destinations for backups, and adds a destination field for EtcdBackupConfig. 

Fixes #7938 

```release-note
seed.backupRestore is now deprecated in favor of seed.etcdBackupRestore which offers the option to manage multiple etcd backup destinations. The deprecated seed.backupRestore is still supported for now. 
```


moadqassem (approved):
/approve
/lgtm

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -98,19 +97,39 @@ type SeedSpec struct {
...
 	BackupRestore *SeedBackupRestoreConfiguration `json:"backupRestore,omitempty"`
+	// EtcdBackupRestore holds the configuration of the automatic etcd backup restores for the Seed
+	EtcdBackupRestore *EtcdBackupRestore `json:"etcdBackupRestore,omitempty"`
 }
 
-// SeedBackupRestoreConfiguration are s3 settings used for backups and restores of user cluster etcds.
+// SeedBackupRestoreConfiguration defines the bucket name and endpoint as a backup destination.
+// Deprecated: use EtcdBackupRestore
 type SeedBackupRestoreConfiguration struct {
-	// +kubebuilder:default=s3.amazonaws.com
```
kron4eg, line 106: is it not defaults anymore?

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -98,19 +97,39 @@ type SeedSpec struct {
...
+	// it enables automatic backup and restore for the seed.
+	Destinations map[string]*BackupDestination `json:"destinations,omitempty"`
+}
+
+// BackupDestination defines the bucket name and endpoint as a backup destination, and holds reference to the credentials secret.
+type BackupDestination struct {
+	// Endpoint is the API endpoint to use for backup and restore.
+	Endpoint string `json:"endpoint,omitempty"`
+	// BucketName is the bucket name to use for backup and restore.
+	BucketName string `json:"bucketName,omitempty"`
```
kron4eg: we can default Endpoint, but do we default BucketName? I mean, `,omitempty` here means 'not-required field' in terms of generated OpenAPI schema. Is this what we want to allow?

pkg/apis/kubermatic/v1/datacenter.go
lsviben: It never defaulted it anywhere, the comment "defaults to s3.amazonaws.com" was misleading unfortunately.

pkg/apis/kubermatic/v1/datacenter.go
lsviben: good point, but this messes up our PATCH endpoints as the fields which are not in the patch without omitempty get defaulted to "", and overwrite the old value. 

But I can easily create an API model to solve that, so Ill change it

kron4eg (approved):
/approve

PR #8050, opened 2021-10-20
Change constraint parameter type

### LINKED ISSUE #8020: Use map[string]json.RawMessage instead of map[string]inteface{} for Constraint Parameters
**User Story**
Gatekeeper constraints paramenters can be anything so we need flexibility for this field, as we want to use KKP Constraints to manage Gatekeeper constraints on the user cluster. 

Initially map[string]inteface was used, but this brings some complications as it messes up the code generation. 

Instead, a better option would be to use map[string]json.RawMessage .

lsviben:
Changes the KKP Constraint Parameter type from map[string]interface{} to map[string]json.RawMessage.

This change is needed as map[string]interface is not suitable for code gen tools, plus map[string]json.RawMessage is a better fit.

Fixes #8020 

```release-note
Changed KKP Constraint parameter type to map[string]json.RawMessage. 
```


pkg/crd/kubermatic/v1/constraint.go
```diff
@@ -77,27 +76,6 @@ type ConstraintSelector struct {
 	LabelSelector metav1.LabelSelector `json:"labelSelector,omitempty"`
 }
 
-type Parameters map[string]interface{}
```
kron4eg, line 80: This is REST API breaking change (see swagger file).

if instead you'd replace it with `type Parameters map[string]json.RawMessage`, that will work without breaks.

lsviben: Thanks, ill check it out

PR #8054, opened 2021-10-21
Use strict TLS cipher suites for kubelet and kube-apiserver

ahmedwaleedmalik:

Kubernetes, by default, supports a vast number of TLS cipher suites including some with known security concerns which weakens the protection provided by them.

Hence, we are limiting the allowed cipher suites for `kubelet` and `kube-apiserver` to the recommended and most secure algorithms:

TLS 1.3 cipher suites:
```text
TLS_AES_128_GCM_SHA256
TLS_AES_256_GCM_SHA384
TLS_CHACHA20_POLY1305_SHA256
```

TLS 1.0 - 1.2 cipher suites:
```text
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
```

Fixes https://github.com/kubermatic/edgematic/issues/65

**Documentation**:

```release-note
NONE
```


pkg/resources/apiserver/deployment.go
```diff
@@ -310,6 +310,7 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		"--audit-log-maxsize", "100",
 		"--audit-log-path", "/var/log/kubernetes/audit/audit.log",
 		"--tls-cert-file", "/etc/kubernetes/tls/apiserver-tls.crt",
+		"--tls-cipher-suites", "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 ,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 ,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256",
```
embik: There's two spaces in here (probably from my ticket, sorry about that), one after `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` and one after `TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384`, and I think that's breaking the test fixtures generated (it shouldn't be multiple lines with cipher suites).

embik (changes requested):

pkg/resources/apiserver/deployment.go
ahmedwaleedmalik: No, I added them mistakenly somehow. It was resolved in my last commit.

addons/kubelet-configmap/kubelet-configmap.yaml
```diff
@@ -91,6 +91,16 @@ data:
     staticPodPath: /etc/kubernetes/manifests
     streamingConnectionIdleTimeout: 4h0m0s
     syncFrequency: 1m0s
+    tlsCipherSuites:
```
embik, line 94: My understanding is limited, but I think this is only setting the cipher for kubelets that use the dynamic configuration feature/addon, no? And the dynamic configuration feature is deprecated in upstream Kubernetes, so it will go away eventually. I'm not sure where non-dynamic-configuration kubelets would set this, but probably in machine-controller?

ahmedwaleedmalik: Let me check, I'm not sure about this one.

ahmedwaleedmalik: @embik Just confirmed you are right and we need to modify the default `kubelet` config in `machine-controller`. Created a PR for that as well. 

https://github.com/kubermatic/machine-controller/pull/1077

embik (approved):
/lgtm

pkg/resources/apiserver/deployment.go
```diff
@@ -310,6 +310,7 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		"--audit-log-maxsize", "100",
 		"--audit-log-path", "/var/log/kubernetes/audit/audit.log",
 		"--tls-cert-file", "/etc/kubernetes/tls/apiserver-tls.crt",
+		"--tls-cipher-suites", "TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305",
```
kron4eg: Please extract this list to the var []string{}, for visibility

ahmedwaleedmalik: Forgot to do that here, I'll fix this :+1:

pkg/resources/resources.go
```diff
@@ -515,6 +515,11 @@ const (
 	CloudConfigSecretKey = "config"
 )
 
+const (
+	// List of allowed TLS cipher suites for kube-apiserver
+	KubeAPIServerTLSCypherSuites = "TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305"
```
kron4eg: I meant real
```
[]string{
    "TLS_AES_128_GCM_SHA256",
    "TLS_AES_256_GCM_SHA384",
    ...
}
```
It can be also exposed to the addons templates as a function so we can have consistent list.

ahmedwaleedmalik: Handled, PTAL.

kron4eg (approved):
/approve

PR #8065, opened 2021-10-22
Support TLS peer connections and live migration in etcd-launcher

embik:
Our etcd clusters currently use unencrypted peering connections via plaintext HTTP. It's recommended by various standards to use TLS-encrypted peer connections via HTTPS instead.

This PR introduces (when using etcd-launcher) TLS peer connection setups for fresh clusters and an automated migration for existing etcd clusters via etcd-launcher and seed-controller-manager.

The migration is a two-step process:

1. Deploys `etcd-launcher` with a new, "mixed" mode where both plaintext and TLS peer ports are available. During this first rollout, "old" peers can still communicate with the "new" peers via the plaintext port, while the TLS port is made available and the necessary certificate information is passed. Pods are annotated to show they support TLS now.
2. As second step, `seed-controller-manager` supervises the rollout / status of etcd Pods, and updates the StatefulSet to switch to a "strict" mode, where only the TLS port is enabled. Peers from the previous rollout can switch from the plaintext port to the TLS port once a peer comes up again after the upgrade.

This also supports undoing the second step of the migration, so going from "strict" TLS-only peers to "mixed" peers with both plaintext and TLS endpoints. That would be a manual operation on the StatefulSet though, it's not an expected automated operation.

Fixes https://github.com/kubermatic/edgematic/issues/55

Still Work-in-Progress.

**Documentation**:

```release-note
- etcd-launcher uses TLS peer connections for new etcd clusters and automatically upgrades existing etcd clusters
- etcd-launcher cannot be disabled as a feature on a cluster to go back to plain etcd
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -340,13 +340,26 @@ func etcdCmd(config *etcdCluster) []string {
 		fmt.Sprintf("--advertise-client-urls=https://%s.etcd.%s.svc.cluster.local:2379,https://%s:2379", config.podName, config.namespace, config.podIP),
 		fmt.Sprintf("--listen-client-urls=https://%s:2379,https://127.0.0.1:2379", config.podIP),
 		fmt.Sprintf("--listen-metrics-urls=http://%s:2378,http://127.0.0.1:2378", config.podIP),
-		fmt.Sprintf("--listen-peer-urls=http://%s:2380", config.podIP),
 		fmt.Sprintf("--initial-advertise-peer-urls=http://%s.etcd.%s.svc.cluster.local:2380", config.podName, config.namespace),
 		"--client-cert-auth",
 		fmt.Sprintf("--trusted-ca-file=%s", resources.EtcdTrustedCAFile),
 		fmt.Sprintf("--cert-file=%s", resources.EtcdCertFile),
 		fmt.Sprintf("--key-file=%s", resources.EtcdKetFile),
 		"--auto-compaction-retention=8",
+	}
+
+	if config.initialState == initialStateNew {
+		cmd = append(cmd, []string{
+			fmt.Sprintf("--listen-peer-urls=https://%s:2381", config.podIP),
+			fmt.Sprintf("--peer-cert-file=%s", resources.EtcdCertFile),
+			fmt.Sprintf("--peer-key-file=%s", resources.EtcdKetFile),
+			fmt.Sprintf("--peer-trusted-ca-file=%s", resources.EtcdTrustedCAFile),
+			"--peer-client-cert-auth",
+		}...)
+	} else {
+		cmd = append(cmd, []string{
+			fmt.Sprintf("--listen-peer-urls=http://%s:2380", config.podIP),
+		}...)
 	}
 
 	if config.enableCorruptionCheck {
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -101,13 +101,15 @@ func main() {
 
 	// in addition, if "strict" mode is enforced, set it up for existing clusters too
 	if os.Getenv(envPeerTLSMode) == peerTLSModeStrict {
-		log.Info("peer-tls-mode: strict")
 		e.usePeerTLSOnly = true
 	}
 
 	log.Info("initializing etcd..")
 	log.Infof("initial-state: %s", e.initialState)
 	log.Infof("initial-cluster: %s", strings.Join(initialMemberList(e.clusterSize, e.namespace, e.usePeerTLSOnly), ","))
+	if e.usePeerTLSOnly {
+		log.Info("peer-tls-mode: strict")
+	}
 
 	// setup and start etcd command
 	etcdCmd, err := startEtcdCmd(e, log)
@@ -602,10 +604,16 @@ func (e *etcdCluster) removeDeadMembers(log *zap.SugaredLogger, unwantedMembers 
 	defer close(client, log)
 
 	for _, member := range unwantedMembers {
+		log.Infow("removing cluster member", "member-name", member.Name)
+
 		if member.Name == e.podName {
 			continue
 		}
 		if err = wait.Poll(1*time.Second, 15*time.Second, func() (bool, error) {
+			if len(member.ClientURLs) == 0 {
+				return false, nil
+			}
+
 			// we use the cluster FQDN endpoint url here. Using the IP endpoint will
 			// fail because the certificates don't include Pod IP addresses.
 			return e.isHealthyWithEndpoints(member.ClientURLs[len(member.ClientURLs)-1:], log)
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -57,6 +57,11 @@ const (
 	initialStateNew          = "new"
 	envPeerTLSMode           = "PEER_TLS_MODE"
 	peerTLSModeStrict        = "strict"
+
+	timeoutListMembers    = time.Second * 5
+	timeoutAddMember      = time.Second * 15
+	timeoutRemoveMember   = time.Second * 30
+	timeoutUpdatePeerURLs = time.Second * 10
 )
 
 type etcdCluster struct {
@@ -273,7 +278,10 @@ func joinCluster(e *etcdCluster, log *zap.SugaredLogger) error {
 
 	peerURLs = append(peerURLs, fmt.Sprintf("https://%s.etcd.%s.svc.cluster.local:2381", e.podName, e.namespace))
 
-	if _, err := client.MemberAdd(context.Background(), peerURLs); err != nil {
+	ctx, cancelFunc := context.WithTimeout(context.Background(), timeoutAddMember)
+	defer cancelFunc()
+
+	if _, err := client.MemberAdd(ctx, peerURLs); err != nil {
 		close(client, log)
 		return errors.Wrap(err, "add itself as a member")
 	}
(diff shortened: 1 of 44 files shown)


embik:
/hold

I found another case that doesn't work yet: Upgrading from an older version of etcd-launcher to this version can trigger a mismatch in the initial advertised peer URLs and the peer URLs in the `--initial-cluster` flag. I'm looking into it.

I still welcome any review comments.

cmd/etcd-launcher/main.go
```diff
@@ -93,9 +101,30 @@ func main() {
 		log.Panicw("failed to set initialState", zap.Error(err))
 	}
 
+	// "strict" mode (only TLS for peer traffic) will be used for new clusters
+	e.usePeerTLSOnly = e.initialState == initialStateNew
```
moelsayed: This doesn't strictly mean it's a freshly created cluster. It's possible to have a "new" cluster with existing data, IIRC in case of a backup restore. Is this case tested?

pkg/controller/seed-controller-manager/kubernetes/launching.go
```diff
@@ -20,10 +20,17 @@ import (
 	"context"
 
 	"go.uber.org/zap"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
xrstf: This import should be grouped with the other k8s imports.

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -120,7 +104,9 @@ func (r *Reconciler) ensureResourcesAreDeployed(ctx context.Context, cluster *ku
 	}
 
 	// check that all StatefulSets are created
-	if err := r.ensureStatefulSets(ctx, cluster, data); err != nil {
+	if ok, err := r.statefulSetHealthCheck(ctx, cluster); !ok || err != nil {
+		r.log.Info("Skipping reconcile for StatefulSets, not healthy yet")
```
xrstf, line 108: Does this affect all StatefulSets? Even Prometheus?

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -532,9 +518,9 @@ func (r *Reconciler) ensureConfigMaps(ctx context.Context, c *kubermaticv1.Clust
 }
 
 // GetStatefulSetCreators returns all StatefulSetCreators that are currently in use
-func GetStatefulSetCreators(data *resources.TemplateData, enableDataCorruptionChecks bool) []reconciling.NamedStatefulSetCreatorGetter {
+func GetStatefulSetCreators(data *resources.TemplateData, enableDataCorruptionChecks bool, enableTSOnly bool) []reconciling.NamedStatefulSetCreatorGetter {
```
xrstf: There's an L missing in `enableTSOnly`

pkg/resources/resources.go
```diff
@@ -105,6 +105,8 @@ const (
 	EtcdStatefulSetName = "etcd"
 	// EtcdDefaultBackupConfigName is the name for the default (preinstalled) EtcdBackupConfig of a cluster
 	EtcdDefaultBackupConfigName = "default-backups"
+	// EtcdTLSEnabledAnnotation is the annotation assigned to etcd Pods that run with a TLS peer endpoint
+	EtcdTLSEnabledAnnotation = "etcd.k8c.io/tls-peer-enabled"
```
xrstf: This should probably be `etcd.kubermatic.k8c.io`

cmd/etcd-launcher/main.go
```diff
@@ -198,24 +225,31 @@ func startEtcdCmd(e *etcdCluster, log *zap.SugaredLogger) (*exec.Cmd, error) {
 }
 
 func deleteUnwantedDeadMembers(e *etcdCluster, log *zap.SugaredLogger) (bool, error) {
-	containsUnwantedMembers, err := e.containsUnwantedMembers(log)
+	unwantedMembers, err := e.getUnwantedMembers(log)
 	if err != nil {
-		log.Warnw("failed to list members ", zap.Error(err))
+		log.Warnw("failed to get unwanted members ", zap.Error(err))
```
xrstf: trailing space in warning text

xrstf (changes requested):
Leave a comment

pkg/controller/seed-controller-manager/kubernetes/resources.go
embik: No, Prometheus is managed by a separate controller loop. The StatefulSets created by this `ensureStatefulSets` are defined here: https://github.com/kubermatic/kubermatic/blob/4c0ac9c42f60a54fbf27da8b03086abd4f10e154/pkg/controller/seed-controller-manager/kubernetes/resources.go#L535. It's only etcd and - optionally - rancher-server.

cmd/etcd-launcher/main.go
```diff
@@ -93,9 +101,30 @@ func main() {
...
 	log.Info("initializing etcd..")
 	log.Infof("initial-state: %s", e.initialState)
-	log.Infof("initial-cluster: %s", strings.Join(initialMemberList(e.clusterSize, e.namespace), ","))
+	log.Infof("initial-cluster: %s", strings.Join(initialMemberList(e.clusterSize, e.namespace, e.usePeerTLSOnly), ","))
+	if e.usePeerTLSOnly {
+		log.Info("peer-tls-mode: strict")
+	}
+
+	// if the cluster already exists, try to connect and update peer URLs that might be out of sync.
+	// etcd might fail to start if peer URLs in the etcd member state and the flags passed to it are different
```
moelsayed, line 112: Would it be possible to check if all the members are available before we do this? Because if not, they won't be all updated and it might break the ring.

pkg/controller/seed-controller-manager/kubernetes/launching.go
```diff
@@ -40,3 +47,39 @@ func (r *Reconciler) clusterIsReachable(ctx context.Context, c *kubermaticv1.Clu
...
+
+	if err != nil {
+		return false, err
+	}
+
+	useStrictTls := true
+
+	for _, pod := range pods.Items {
+		if _, ok := pod.Annotations[resources.EtcdTLSEnabledAnnotation]; !ok {
+			useStrictTls = false
```
moelsayed: Should we break here or do we need to check the rest of the list?

moelsayed (commented):
Brilliant solution! Thank you!
I have just a few questions. Also, I want to know if you managed to test/handle the following cases:
- Restore from backup, both when the backup is from before TLS was enabled and after. 
- Upgrade cases from older versions. There might be cases where users are upgrading from a version where etcd-launcher is not enabled. 

That's all. Great job.

pkg/controller/seed-controller-manager/kubernetes/launching.go
embik: You're right, I don't think we need to. We could even `return false, nil` in here and get rid of `useStrictTls` completely by returning `true, nil` in the last line of this function.

embik:
@moelsayed: Thanks for the review and the kind words!

> Restore from backup, both when the backup is from before TLS was enabled and after.

That's a very good question. I don't think I was able to test that because the restore didn't work on my test clusters that I labelled with `worker-name=marvinmbp` to get them out of the main KKP's control. Do you think we should include this as a test case maybe? It seems quite important that this works, now and in the future.

> Upgrade cases from older versions. There might be cases where users are upgrading from a version where etcd-launcher is not enabled.

I tested upgrades from older `etcd-launcher` versions manually (which actually caused a bit of a headache as the "remove dead member" logic in `etcd-launcher` obviously didn't account for this new scenario; but it wasn't causing any problems in the latest iterations anymore). There's also `pre-kubermatic-etcd-launcher-e2e` which includes enabling `etcd-launcher` on a normal etcd cluster, so that's tested in CI thankfully.

cmd/etcd-launcher/main.go
embik: Since this happens before the etcd node this `etcd-launcher` instance manages even starts, I'm not sure that's possible, there's going to be one unavailable by logic - or do you mean check the remaining members (so, 2 out of 3, 4 out of 5, etc) are available?

cmd/etcd-launcher/main.go
embik: Good question. As per https://github.com/kubermatic/kubermatic/pull/8065#issuecomment-966175175 I'm actually not sure, I don't think I managed to set that up as a scenario manually. Mirroring my question in the comment, do you think that should be a test case part?

moelsayed: I am afraid so. This particular case is very important to test.

cmd/etcd-launcher/main.go
moelsayed: No checking the remaining ones wouldn't make sense. They also might not be available.

moelsayed: Do you think it's possible to move it after the start command? maybe add a couple retries to make sure they are all up?

embik: I moved it to before the start command because etcd can fail to start if the peer URLs in the cluster state of the node is different from what is passed via flags/environment variables. So the idea was to make sure the in-cluster state is in sync with the intention of the starting etcd node. Maybe run the peer URL update a second time after starting?

embik:
@moelsayed @xrstf this is finally in a state for review again. As per @moelsayed's suggestion, I've added a completely new test case for taking a backup with plain etcd (which required setting up a different variant of KKP that includes minio with self-signed TLS certs and all that), enabling etcd-launcher and then restoring that backup. It uncovered a few issues that are now fixed, so thanks for the great suggestion.

xrstf (approved):
/approve

PR #8070, opened 2021-10-22
Markup kubermatic.k8c.io types with defaults and ENUMs

kron4eg:
Make new CRDs schema more precise

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
+++ b/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
@@ -304,6 +304,9 @@ spec:
                         description: LoadBalancerSKU sets the LB type that will be
                           used for the Azure cluster, possible values are "basic"
                           and "standard", if empty, "basic" will be used
+                        enum:
+                        - standard
+                        - basic
                         type: string
                       resourceGroup:
                         type: string
@@ -800,6 +803,7 @@ spec:
                     description: IPVS defines kube-proxy ipvs configuration options
                     properties:
                       strictArp:
+                        default: true
                         description: StrictArp configure arp_ignore and arp_announce
                           to avoid answering ARP queries from kube-ipvs0 interface.
                           defaults to true.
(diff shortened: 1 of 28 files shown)


charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
```diff
@@ -1550,30 +1593,74 @@ spec:
...
                 type: string
               extendedHealth:
                 description: ExtendedHealth exposes information about the current
                   health state. Extends standard health status for new states.
                 properties:
                   apiserver:
+                    enum:
+                    - 0
+                    - 1
+                    - 2
```
xrstf: Are we sure we don't want to turn those into a bit more usable values? Like strings? What's the reason why we have numbers here?

xrstf (commented):
Numbers tho

charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
kron4eg: Because the exposed type is int 🤷‍♂️

Should we drift further away from original types? Integration those new types probably would be crazy.

xrstf: For what it's worth, I hate these numbers in our CRDs. If we don't change them now, we will be stuck with them for-e-ver.

There are also very few k8s APIs I can think of where numbers are being used.

xrstf: Do we need a 3rd opinion as a tie breaker?

kron4eg: Fixed

pkg/apis/kubermatic/v1/settings.go
```diff
@@ -22,6 +22,15 @@ import (
...
 
+// +kubebuilder:validation:Enum=0;1
+
+type ClusterType int8
+
+const (
+	ClusterTypeAll ClusterType = iota
+	ClusterTypeKubernetes
+)
+
```
xrstf: My PR explicitly removed cluster types, because we only have one type. This was in agreement with the SIG API.

kron4eg: probably the artifact of rebase

xrstf (approved):
/approve

PR #8085, opened 2021-10-25
Remove versions/addons/prometheus CLI flags

xrstf:
With #7764 we started to use the KubermaticConfiguration more directly to get access to the various configuration options. Previously, we "tunneled" them through the help of tons and tons of CLI flags.

This PR continues this march by removing some leftover flags (like `-versions`) from the last PR and removing even more flags, most specifically related to addon handling and usercluster Prometheus stuff. Instead of reading addons from YAML files in the `extra-files` directory, we read them _live_ from the KubermaticConfiguration.

I am 85,4% sure that the old "Defaults" field, where an admin would just configure a flat list of addon names, is actually _deprecated_ and admins should use "DefaultManifests" instead. But I kept the logic as it is right now.

* `kubermatic-api`
  * `-master-resources` was removed (leftover from the last PR)
  * `-versions` same.
  * `-updates` same.
  * `-provider-incompatibilities` same.
  * `-accessible-addons` is now read from the KubermaticConfiguration under `spec.api.accessibleAddons`.
  * `accessible-addons` same.
* `kubermatic-seed-ctrl-mgr`
  * `-datacenter-name` was renamed to `-seed-name`, which is much clearer.
  * `-kubernetes-addons-path` was renamed to `-addons-path`, because we only ever support Kubernetes (the old "kubernetes-" prefix was added when we temporarily also supported Openshift).
  * `-kubernetes-addons-list` and `-kubernetes-addons-file` were removed and are now read directly from the KubermaticConfiguration.
  * `-in-cluster-prometheus-rules-file` same.
  * `-in-cluster-prometheus-disable-default-rules` same.
  * `-in-cluster-prometheus-disable-default-scraping-configs` same.
  * `-in-cluster-prometheus-scraping-configs-file` same.
  * `-monitoring-scrape-annotation-prefix` same.

As you can see, this PR kind of escalated a bit. :grin: 

```release-note
NONE
```


pkg/handler/v1/addon/addon.go
```diff
@@ -24,13 +24,12 @@ import (
 
 	"github.com/go-kit/kit/endpoint"
 	"github.com/gorilla/mux"
+	"k8s.io/apimachinery/pkg/util/sets"
```
kron4eg: sounds like you need to rerun [gimps](https://github.com/xrstf/gimps)! :D

xrstf: Judging from the many tests, I think gimps is the least of this PR's problems :D

kron4eg (approved):
/approve

PR #8093, opened 2021-10-26
Add support for preset deletion

Happy2C0de:
This PR adds support to delete providers from a given preset AND if there is only a single provider left in the preset, the preset gets fully deleted.

This is a very smooth and user friendly implementation how one can delete a preset from KKP via the API. This can also be added to the Dashboard in an upcoming PR if needed.

None

- Added DeletePreset function and the whole swagger models.
- Because the API endpoint looks as follows: `/api/v2/providers/<provider>/presets` it makes sense to keep the provider based functionality. Accordingly, if one calls `DELETE /api/v2/providers/<provider>/presets/<preset_name>` the implementation will only remove the given provider (in the case when there is more than one provider configured). If there is only one provider, the preset gets fully deleted.  This makes sense because a preset with no provider is forbidden.

**Documentation**:

```release-note
Add support to remove a preset provider via the REST API
```


kron4eg (commented):
/approve

Happy2C0de:
@kron4eg 
Can we backport this to release/v2.18 ?

Happy2C0de:
@kron4eg 
Thank you for the feedback.
I see your point. ATM, users cannot update the `requiredEmails` nor can users recreate a preset via the API what makes the Preset API not really usable. The two merged PRs fix this issue(/bug?) such that presets can be dynamically adjusted.
That's why I asked, but discuss it internally and decide how you would like to proceed. 👍 
Thanks a lot!

zreigz:
@Happy2C0de thanks for your contribution. We have to discuss this extension with our product manager to decide if it's should be in the current release or later.
Definitely from the API perspective, it's a very valuable functionality.

Happy2C0de:
Thank you for the feedback @zreigz 
I see your point with feature vs bugfix. Discuss it internally and decide what's reasonable in this situation. 👍 
In either case, I am looking forward to see this API endpoint in KKP. 🚀

PR #8101, opened 2021-10-27
Introduce Cloud Provider Reconciling / PoC for AWS

### LINKED ISSUE #8136: Add field and logic to trigger periodic cloud provider reconcilings

### LINKED ISSUE #8139: Implement Provider Reconciling for AWS

xrstf:

This PR significantly improves (IMHO) the AWS cloud provider in KKP. But it also includes 2 other important changes.

### Introduction of `LastProviderReconciliation`

To fix #8136 (the original intent behind this PR), a new timestamp is introduced: `LastProviderReconciliation`. This timestamp on the Cluster's status contains the time we last performed a cluster reconciliation.

Once enough time has passed (or the timestamp isn't set yet), KKP will reconcile the cloud provider the next time it reconciles the Cluster object. Reconciling the provider is then meant to fix missing resources, restore settings etc.

The minimal interval that needs to pass in between reconciliations is configured on each *Datacenter* (not Seed). This may be too fine grained, WDYT? The value is defaulted to 6 hours. I pulled this number out of a hat.

### Fixed integration tests

Since we updated to Go 1.17, the build constraint comments have changed from `// +build integration` to `//go:build integration`. This means that the regex we use in the Makefile to find integration test packages didn't work anymore and found 0 packages. So for quite a few weeks, our integration tests were effectively dead.

Since the AWS provider now has a bunch of integration tests, this PR needed to fix the tests as well. I updated the test binaries (which were on 1.18!) and the envtest stuff to make them work again, then added a whole bunch more for AWS. The logic for running the integration tests has been moved to a script, because we now need to start a localstack container and that would have been too much for a Makefile.

### `ReconcilingCloudProvider`

There is a new interface that can be implemented by cloud providers, requiring them to add a `ReconcileCluster()` function. Any cloud provider that implements this interface will then reconcile regularly, whereas "old" providers stay with the "init once, then forget" method.

### AWS Provider

This PR demonstrates the reconciling method using the AWS provider, which has been greatly rewritten. Here a re a few notes that might make the PR easier to follow:

* The `provider.go` is really only the provider implementation, the code for all actual resources (like security groups) is in their own files now.
* The provider calls `reconcile....` functions, which in turn do whatever they need to reconcile.
* The migration code for MultiAZ and ICMP have been removed. Now that we reconcile all the time, those kind of fixes will happen automatically without requiring a dedicated migration code. One of the nice side effects from having reconciliation :)
* Most of the `api.go` is kept without changes, the functions were just moved there to signify that this is where the package's public API lives.
* The reconciling loops for AWS try their best to find orphaned resources. Like when it created a security group and someone/something removes its ID from the cluster spec, the provider will not simply create a new one, but try to find an existing one (this is both to be nice, but also because names are unique and creating another SG with the same name would fail).
* In the tests, most names have a random suffix appended. This is to allow to re-use the same localstack container between tests.

Previously, the old provider would create a finalizer for each object it created. This was used to a) reduce the number of API calls when cleaning up, and b) to know if the resource was created by KKP or if the user specified the name of a pre-existing resource.

Now with the reconciling, there is no need for dedicated small finalizers. In fact, it makes things more complicated, as reconciling a cluster is now a single thing that happens. Like even without a finalizer, because we _want to_ reconcile, we have to check all resources anyway. So the finalizer will not save much work.

For this reason, the finalizers have been combined into a single `kubermatic.io/cleanup-aws`.

**But** this means that we must keep track of resource ownership in a different way. KKP must not delete a resource that it was told to re-use. To achieve this, we put an owner tag (`owned-by.kubermatic.k8c.io/<clusterid>`) on the AWS resource.

There is a migration, called "backfilling", that is run when the new code encounters an old cluster. In this migration we look at the finalizers to figure out if KKP created for example the security group. And if KKP did, we create the owner tag and remove the finalizer. After the backfilling, no finalizer except `kubermatic.io/cleanup-aws` is left.

In general there are 2 tags: The "ownership tag" (only of importance for KKP itself) and used for EC2/IAM resources, and there is the "cluster tag", which is required for AWS CCM to work properly (only used for EC2 resources).

### Testing

Most of this PR is just test code. Be warned though, those are integration tests and they are purposefully not table-driven. Instead each test tells its own story, so you can read a test from start to finish and understand what it's doing, instead of having a testcase struct with 20 anonymous functions that are called in a magic, arcane order, making it impossible to read and understand the table of tests.

You can use `hack/run-integration-tests.sh`, but this will probably lead to failed envtest and vsphere tests, though the AWS provider tests should work just fine. You can also perform the same setup as in the script and simply do `go test -tags integration ./pkg/provider/cloud/aws`.

Fixes #8139, fixes #8136

**Documentation**:

```release-note
Introduce periodic reconciling for cloud resources used by userclusters; currently implemented for AWS-based userclusters
```


embik (commented):
This sounds good to me, but I wonder if it should offer some way to force-sync a cluster by setting the `LastProviderReconciliation` field to nothing/nil?

hack/images/integration-tests/Dockerfile
```diff
@@ -12,17 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.17.1
+FROM quay.io/kubermatic/build:go-1.17-node-14-kind-0.11-1
 LABEL maintainer="support@kubermatic.com"
 
-ENV KUBEBUILDER_VERSION=2.3.1 \
```
xrstf, line 18: We didn't actually use kubebuilder, we just abused their archive to get kube-apiserver and etcd binaries.

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
```diff
@@ -176,46 +179,81 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
...
 	}
 
-	if cluster.Status.CloudMigrationRevision < awsHarcodedAZMigrationRevision {
-		if err := r.migrateAWSMultiAZ(ctx, cluster, prov); err != nil {
-			return nil, err
-		}
-	}
-
-	if _, err := prov.InitializeCloudProvider(cluster, r.updateCluster); err != nil {
+	hanndleProviderError := func(err error) (*reconcile.Result, error) {
```
embik:
```suggestion
	handleProviderError := func(err error) (*reconcile.Result, error) {
```

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
```diff
@@ -176,46 +179,81 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
...
+			}
+
+			cluster, err = r.updateCluster(cluster.Name, func(c *kubermaticv1.Cluster) {
+				now := v1.Now()
+				c.Status.LastProviderReconciliation = &now
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to set last reconcile timestamp: %w", err)
+			}
+		}
```
embik: I'm a little bit conscious of the way errors are handled when reconciling. Unless it's a conflict (as per the definition of `handleProviderError`), the seed-controller-manager would start a loop of trying to reconcile, failing, and immediately trying again, which could impact the API limits we're trying to stay clear of.

What's your thoughts on having a "LastProviderError" status timestamp that helps with a backoff period to a) wait for transient errors to clear and b) limit API access?

pkg/provider/cloud/aws/doc.go
```diff
@@ -0,0 +1,56 @@
...
+/*
+Package aws contains the Amazon Web Services (AWS) cloud provider implementation.
+
+This provider is responsible for managing a variety of resources in the
+AWS cloud, some of which can pre-exist, some are always created explicitly
+for a given usercluster. These resources are:
+
+* EC2: VPC
+  A VPC must already exist and currently KKP always chooses the default VPC
+  in the given AWS account. There is no option to influence this.
```
embik: Is this really true? When creating a Preset for AWS in the KKP UI, you get to choose a VPC.

hack/run-integration-tests.sh
```diff
@@ -0,0 +1,75 @@
...
+  containerName=kkp-localstack
+
+  docker run \
+    --name "$containerName" \
+    --rm \
+    --detach \
+    --publish 4566:4566 \
+    --publish 4571:4571 \
+    --env "SERVICES=iam,ec2" \
+    localstack/localstack:0.12.19
```
kron4eg: probably this image version should be more visible in the source

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
xrstf: I think using a RequeueAfter would be sufficient :thinking:

pkg/controller/operator/defaults/defaults.go
```diff
@@ -54,6 +55,10 @@ const (
 	DefaultMaximumParallelReconciles              = 10
 	DefaultS3Endpoint                             = "s3.amazonaws.com"
 
+	// DefaultCloudProviderReconciliationInterval is the time in between deep cloud provider reconciliations
+	// in case the user did not configure a special interval for the given datacenter.
+	DefaultCloudProviderReconciliationInterval = 6 * time.Hour
```
kron4eg, line 60: Isn't that too long?

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
embik: Fine by me, I think having the last time a cloud operation / reconcile failed would be neat, but it's not necessary. A RequeueAfter in `handleProviderError` for all errors would probably help with not hitting APIs too much in error cases.

pkg/provider/cloud/aws/doc.go
xrstf: Yeah no, I wrote the doc.go when I just started refactoring. Gotta fix that.

pkg/controller/operator/defaults/defaults.go
xrstf: As I said, I pulled it our of where no sun is shining (damn, I would like it there (wow this turned ugly)). I was thinking that 6hr might even be too fast and it should maybe be 12hr instead.

But yeah, I am open to ideas. ^^ @embik WDYT?

embik: I wonder if there's a "right" setting for this at all. Personally, I think six hours is reasonable to start out with and see how the feature fares - We can always adjust later on.

But I'm also relatively indifferent to adjusting. I just think we should be greater than 2-3 hours. Customers that want a faster reconciliation can adjust it, and customers with a large KKP environment won't (hopefully) get hit with rate limiting when they roll out 2.19.

xrstf: Remind me to add some Prometheus metrics for number of reconcilings for userclusters. Maybe in a follow up PR.

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
xrstf: Thinking further about this, maybe you're right. I was wondering if we maybe should the `LastProviderReconciliation` _always_, even if an error occured. This would however dramatically slow down any fixes that could be made. However, consider the errors that can happen during reconciliation: It's pretty much all external stuff, and what are the chances that creating a security group works in 10s if it didn't work this time?

So waiting a bit/much longer in cases of errors could make sense. @kron4eg WDYT?

xrstf:
Fixed some feedback, added basic Prometheus metrics and change the VPC reconciling behaviour to never _change_ a VPC on a Cluster object.

xrstf:
Metrics seem to work fine:

screenshot-2021-11-08-191101

embik (approved):
/lgtm

kron4eg (approved):
/approve

xrstf:
/override pre-kubermatic-e2e-vsphere-ubuntu-1.21
/override pre-kubermatic-e2e-vsphere-ubuntu-1.21-datastore-cluster

vCenter is offline

embik (approved):
/lgtm

PR #8105, opened 2021-10-28
Set sig/crd label when pkg/crd is changed

kron4eg:
To automate detection of PRs that change pkg/crd, so we could backport change to new pkg/apis/kubermatic API.

```release-note
NONE
```


xrstf (approved):
/approve

PR #8116, opened 2021-10-30
fix api client usage

mate4st:
Fixes api client usage for defaultClusterTemplates. Usual manager client uses a cache that is not initialized at the point it is used. 

```release-note
NONE
```


xrstf (approved):
/approve

Issue #8098, opened 2021-10-27
Make it possible to get the status of KonnectivityService flag from API


PratikDeoghare:
Dashboard needs to be able to get the status of KonnectivityService flag from the API. Currently this is not possible. 

Causes: https://github.com/kubermatic/dashboard/issues/3672
Blocks: https://github.com/kubermatic/dashboard/issues/3807

PR #8121, opened 2021-11-01
add api endpoint to get feature gates

PratikDeoghare:
Currently there is no way to get status of feature gates from the api. 
This prevents dashboard from picking the right components to show. 

 https://github.com/kubermatic/kubermatic/issues/8098
 kubermatic/dashboard#3672
 kubermatic/dashboard#3807

```release-note
NONE
```


pkg/features/features.go
```diff
@@ -61,6 +61,7 @@ const (
 )
 
 // FeatureGate is map of key=value pairs that enables/disables various features.
+// swagger:model FeatureGate
 type FeatureGate map[string]bool
```
maciaszczykm: I wonder if we should expose that structure. A map is okay internally, but in the API it might be better to expose a new structure. With structure it is clear what feature flags are available and exposed, besides that we avoid exposing everything automatically.

maciaszczykm:
@PratikDeoghare Please avoid non-descriptive commit messages.

pkg/features/features.go
maciaszczykm: @zreigz WDYT about it?

maciaszczykm: @kubermatic/sig-api WDYT about it?

pkg/handler/v2/routes_v2.go
```diff
@@ -4676,6 +4681,36 @@ func (r Routing) listUser() http.Handler {
...
+//       200: FeatureGate
+//       401: errorResponse
+//       403: errorResponse
+func (r Routing) getFeatureGates() http.Handler {
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(
+			func(ctx context.Context, request interface{}) (interface{}, error) {
```
zreigz: please add this method to `kubermatic/pkg/handler/v2` as a new folder with file `feature_gates/feature_gates.go`

pkg/provider/kubernetes/featuregates_test.go
```diff
@@ -0,0 +1,108 @@
+/*
```
zreigz: this is the wrong place for testing API. It should be here: `kubermatic/pkg/handler/v2/feature_gates/feature_gates_test.go`

pkg/features/features.go
```diff
@@ -61,6 +61,7 @@ const (
 )
 
 // FeatureGate is map of key=value pairs that enables/disables various features.
+// swagger:model FeatureGate
```
zreigz: we hold models here: `kubermatic/pkg/api/v2/types.go`

zreigz (changes requested):

pkg/features/features.go
zreigz: I agree with @maciaszczykm. You should create some API model for this

pkg/handler/v2/routes_v2.go
```diff
@@ -4676,6 +4682,32 @@ func (r Routing) listUser() http.Handler {
...
+//       default: errorResponse
+//       200: FeatureGates
+//       401: errorResponse
+//       403: errorResponse
+func (r Routing) getFeatureGates() http.Handler {
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(featuregates.CreateEndpoint(r.featureGatesProvider)),
```
zreigz: it's GetEndpoint

pkg/handler/v2/routes_v2.go
```diff
@@ -4676,6 +4682,32 @@ func (r Routing) listUser() http.Handler {
...
+		func(_ context.Context, r *http.Request) (interface{}, error) {
```
zreigz: you can use `common.DecodeEmptyReq,`

zreigz (changes requested):

Issue #8047, opened 2021-10-20
Full proposal for Addon Improvement


SimonTheLeg:
As a Kubermatic developer I would like a full proposal that builds on the architecture proposal, so that future work can be planned around this topic. 

## Acceptance Criteria

- Proposal is based on [architecture proposal](https://drive.google.com/file/d/1drryGHt2MbCDY6wRz99u7lMSBYXgI2lH/view?usp=sharing)
- Proposal follows [Kubermatic structure](https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/example.md)

Issue #6000, opened 2020-10-07
Implement application creation during cluster creation


zreigz:
**User Story**
- user can pick Applications during cluster creation
- applications should be created in the background once the cluster is ready (similar to node deployment creation)

cschieder:
Hey :) This was planned for 2.18 but not implemented, is there anything you need?

SimonTheLeg:
The functionality described in this ticket will be added with the new Applications mechanism. Currently we do not plan to extend the current AddOn mechanism with this functionality

As a result I think it makes sense to re-use this ticket. I will rename it to applications and move it into the Applications epic.
Additionally I will also move and rename the UI ticket this story blocks.

csengerszabo:
Issue moved to [kubermatic/dashboard #4437](https://github.com/kubermatic/dashboard/issues/4437) via [**ZenHub**](https://www.zenhub.com/)

Issue #5499, opened 2020-05-27
Move the configuration of the AddonConfigs to the main repo


maciaszczykm:
**What would you like to be added? / User Story**:
The configs with logos etc. should be deployed with the application in all environments. Currently we keep the config for node-exporter in the infra repo which makes it private only for us.

maciaszczykm:
Current location: https://github.com/kubermatic/infra/blob/master/seeds/dev.kubermatic.io/addonconfigs.yaml

Docs: https://docs.kubermatic.com/kubermatic/v2.13/advanced/addons/

maciaszczykm:
It should be possible to override this config if anyone wants to.

maciaszczykm:
I have done some investigation of it and also discussed it with @xrstf on Slack and it seems to me that we should discuss it further before we decide what to do next. I would like to avoid taking any decisions here as I am not 100% aware of how currently we are handling the installation of Kubermatic, especially since we are using the Kubermatic Operator which is our new solution. It would be perfect if the cluster-lifecycle team could join the discussion and move it further. If it can be done without modifying the operator I could do it on my own, but otherwise, I'd rather discuss it here and avoid breaking anything.

Few more things from the discussion on Slack:
1. The configuration has to be shipped to the user, especially for the addons that are common like node-exporter. Without config, the user would see only the name of the addon in the UI. The configuration provides a logo, description, and optionally form configuration.
1. Perhaps we can ship the configuration files to the customer without putting them in the Kubermatic Operator so the user that will be installing Kubermatic can decide which AddonConfigs he wants to include in his installation. He can edit them before and adjust to his needs. This adds one more manual step but it makes the configuration more visible to the user.
1. At some point, we should make AddonConfigs namespaced. To do this we would also need to update the API. It should not be a big problem, we just have to synchronize it.

maciaszczykm:
cc @kdomanski for prioritization. This change was requested by @scheeles.

zreigz:
cc @xrstf any update?

PR #8125, opened 2021-11-02
Proposal for improving addons

SimonTheLeg:

Proposal for Improving Addons. Why we need it → see Motivation and Background Section + #8047 

```release-note
NONE
```


docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+
+- CoreAddonCatalogue → will be maintained by Kubermatic. Should not be directly edited by the customer. We think it makes sense to add this as a layer of protection against misconfiguration. This catalogue is rarely changed. Moreover, this catalogue will not be shown in the UI. This catalogue contains addons for the supported Kubernetes versions (eg KubeProxy will have 3 versions of the addons, one per K8s version. OpenVPN will have one version of the addons compatible with the 3 K8s versions)
+- CustomAddonCatalogue(s) →one or multiple catalogues that contain custom addons. This catalogue can be changed more often. Additionally, it is possible to have multiple CustomAddonCatalogues
+
+Furthermore, the AddonCatalogueDistributor ensures that there are no duplicate addons inside the AddonCatalogue. A duplicate is defined by two addons having the same `name` and `version`. In these situations, we propose the AddonCatalogueDistributor to stop its distribution and log an error.
```
embik: I'd suggest to validate the `AddonCatalogue` resources through a webhook instead, so it gets rejected and doesn't take down the AddonCatalogueDistributor as a whole.

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+```yaml
+apiVersion: kubermatic.k8c.io/v1
+kind: AddonInstallation
+metadata:
+  name: prometheus
+    namespace: cluster-47s2ddlfgj
+spec:
+  coreAddon: false #This can be omitted, false will be the default
+    targetNamespace: monitoring
+    createNamespace: true # flag to create the targetNamespace
```
embik: this in't valid YAML. I'm not sure if it's intentional or not, but it's also happening in the example under `CoreAddonController`, so I wanted to point it out. Probably looking for:

```suggestion
spec:
  coreAddon: false #This can be omitted, false will be the default
  targetNamespace: monitoring
  createNamespace: true # flag to create the targetNamespace
```

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+- Mandatory core components of user clusters are named “Default Addons” which implies that these can be deleted by Cluster Admins
+- Complex Addons can leave a trail of reconciliation errors until successfully deployed because all manifest files are applied at the same time
+- Creating a cluster is a two-step process; firstly you have to create a cluster and only  afterward you can select custom Addons
+
+*note: PS team report that some clients are using weekly KKP update in dev environments, consequently they need to rebuild the addon image every week which is painful and error-prone.*
+
+## Goals
+
+- Decouple custom addons from the KKP lifecycle
+- Handle dependency **inside** an Addon (e.g. CRD need to be created before CRs, namespace needs to be created before other objects)
```
embik, line 55: How is this goal achieved in the proposal? Maybe I missed it, but I couldn't find it. The only thing I found is `RequiredResourceTypes`, and my understanding is that this is actually defining outside dependencies (prerequisites for successfully installing the addon).

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+
+### AddonCatalogue CR
+
+- contains a list of all addons
+- each addon comes in multiple versions
+- each version has
+  - template
+    - a rendering method. Possible methods are (additional methods can easily be added in the future):
+      - go-template → same as what we currently offer in addons
+      - helm → rendering of helm charts
```
embik, line 114: Any plans to make the template data (see https://docs.kubermatic.com/kubermatic/v2.18/guides/addons/#manifest-templating; this could help generating per-cluster Ingress FQDNs or other resources based on cluster information) passed to the current addons (`go-template` rendering method in this proposal) accessible to helm-based addons?

embik (commented):
I've added a few (minor) comments.

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+
+This proposal introduces improvements to the current [Addon system](https://docs.kubermatic.com/kubermatic/v2.18/guides/addons/). It aims to make current addons more flexible and user-friendly. Additionally, it simplifies the usage of addons we ship with (almost) every cluster, so-called "Default Addons".
+
+## Motivation and Background
+
+Currently, KKP only supports one mechanism for addons: For this, all addons are stored in a single docker image that is configured in KKP. Afterwards, an initContainer is being run that starts the image and copies all addons onto the seed-controller-managers local file-system. From there, addon manifests are being rendered and rolled out to user clusters.
+
+The current implementation of Addons has some flaws that make the usage and maintenance of Addons cumbersome for cluster administrators:
+
+- To add “Custom Addons” the cluster-admin needs to bundle manifests and templates for these Addons in a Docker image that is based on the official addon-manifest docker image by Kubermatic.
```
maciaszczykm, line 42: Besides that in most of the cases user needs to create AddonConfig CRD for each custom/accessible addon, so form, logo and description will be visible in the dashboard.

maciaszczykm: See: https://github.com/kubermatic/kubermatic/issues/5499

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+- Creating a cluster is a two-step process; firstly you have to create a cluster and only  afterward you can select custom Addons
+
+*note: PS team report that some clients are using weekly KKP update in dev environments, consequently they need to rebuild the addon image every week which is painful and error-prone.*
+
+## Goals
+
+- Decouple custom addons from the KKP lifecycle
+- Handle dependency **inside** an Addon (e.g. CRD need to be created before CRs, namespace needs to be created before other objects)
+- Addons are correctly reconciled on changes. [kubermatic/kubermatic#6035](https://github.com/kubermatic/kubermatic/issues/6035)
+- Be able to choose which addons should be installed during the cluster creation or in the cluster template
```
maciaszczykm, line 57: Linking: https://github.com/kubermatic/kubermatic/issues/6000

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+apiVersion: kubermatic.k8c.io/v1
+kind: AddonCatalogue
+metadata:
+ name: custom-catalogue
+spec:
+ addons:
+     - prometheus:
+            description: "The Prometheus Node Exporter exposes a wide variety of hardware- and kernel-related metrics."
+            logoFormat: "svg+xml"
+            logo: # logo must be base64 encoded.
```
maciaszczykm: Size of it may be problematic if we will store the logos together. In the @kubermatic/sig-api @kubermatic/sig-ui we were considering a change here, perhaps we should aim to host these images somewhere.

docs/proposals/improve-addons.md
```diff
@@ -0,0 +1,350 @@
...
+
+*note: PS team report that some clients are using weekly KKP update in dev environments, consequently they need to rebuild the addon image every week which is painful and error-prone.*
+
+## Goals
+
+- Decouple custom addons from the KKP lifecycle
+- Handle dependency **inside** an Addon (e.g. CRD need to be created before CRs, namespace needs to be created before other objects)
+- Addons are correctly reconciled on changes. [kubermatic/kubermatic#6035](https://github.com/kubermatic/kubermatic/issues/6035)
+- Be able to choose which addons should be installed during the cluster creation or in the cluster template
+- Be compatible with private registries in preparation for future air-gapped environment stories. The current Policy regarding the air-gapped environment is to assume a private trusted registry is available. Addon Source (manifest of the workload to deployed) and workload images of the addon should be  configurable to point on a private registry
```
maciaszczykm, line 58: Just an idea, but if we are improving the addons then perhaps we can solve issue like https://github.com/kubermatic/dashboard/issues/3666 too. So make k8s/dashboard an addon that will be always deployed but the version used will differ based on the user cluster version.

kron4eg:
I'd like to show you alternative perspective on addons.

The name for addons is very unfortunate and doesn't signal clearly their purpose. Current addons represent the integral part of each KKP version. They could have been written in Go using types from kubernetes API, however since most of them are very much static it's easier to maintain them as YAML files. So the "flaws" are no the flaws but a feature that was intended this way for a reason.

Best look at addons is as they are a Go types that escaped/leaked to YAML. Please don't improve addons building on top of them but rather build some completely new workflow.

docs/proposals/improve-addons.md
SimonTheLeg: I just realized from re-reading this bullet, that choosing suitable addons before cluster-creation is something that will be possible with the components in this proposal, but is not included as a design decision

cc @vgramer: I think we need to think about whether we want to put this into this proposal or in a separate one/follow up User Story

docs/proposals/improve-addons.md
SimonTheLeg: Yes this would definitely be possible. You could even use constraints to ensure only a suitable dashboard is being shown to users. 

While it's not directly a goal of this proposal, I am going to add it to the "Non Goals section". This ensures that we have the topic ready on our roadmap :)

docs/proposals/improve-addons.md
SimonTheLeg: Ah interesting. Maybe I misunderstood our current implementation. I thought our current implementation would do exactly that: ensure that CRDs and namespaces are deployed before any other manifests.
cc @vgramer: This is something I think we have to look in the code of the current implementation. But yes, if the current system is not able to install CRDs before deploying resources, then we should include that capability in our proposal.

On the other side: For the helm rendering method, handling dependencies is done directly via helm; So theoretically this should come out of the box (of course during implementation we will have to see how well it works in practice)

docs/proposals/improve-addons.md
SimonTheLeg: that is actually a really good point. We would have to slightly adapt the proposal, but this makes sense to me

docs/proposals/improve-addons.md
SimonTheLeg: good point. I have added logos as a non-goal for now, so we can move forward with the topic. Good news is: With the current architecture, we can add logos later on without any problems; be it base64 encoded or hosted somewhere else

docs/proposals/improve-addons.md
SimonTheLeg: Good Spot! 
Yeah sometimes Nuclino Yaml hates you :D

docs/proposals/improve-addons.md
embik: Not sure we're talking about the same thing here :D I was referring to the duplicate detection in an (instance of) `AddonCatalogue`. Maybe I misunderstood this part of the proposal, but it looked like the validation of resources was supposed to happen in the distributor, while a webhook would help by rejecting invalid `AddonCatalogues` before they hit the cluster and can be processed by the AddonCatalogueDistributor.

docs/proposals/improve-addons.md
SimonTheLeg: Ah interesting. Maybe I misunderstood our current implementation. I thought our current implementation would do exactly that: ensure that CRDs and namespaces are deployed before any other manifests.
cc @vgramer: This is something I think we have to look in the code of the current implementation. But yes, if the current system is not able to install CRDs before deploying resources, then we should include that capability in our proposal.

On the other side: For the helm rendering method, handling dependencies is done directly via helm; So theoretically this should come out of the box (of course during implementation we will have to see how well it works in practice)

docs/proposals/improve-addons.md
SimonTheLeg: I commented on the wrong comment on accident. I have moved my comment since then. Will get back to you about the webhook soon. Just have to think about it for a minute

docs/proposals/improve-addons.md
embik: Ah, I was more thinking of a flexible dependency management system within the addon, like https://carvel.dev/kapp/docs/latest/apply-ordering/. That's probably fairly complex to pull off, but I misunderstood the goal then :)

SimonTheLeg:
@kron4eg 

Thank you for the alternative perspective.

In my opinion the term "addon" describes the situation at hand well. You have a cluster and want to add components to it. I think the nature that some components are mandatory, does not change the fact that they are added onto a cluster. Hence the name "addon". Furthermore I think the proposed architecture hits a nice compromise between generality and enforcement. On one side we can use the same generic mechanism for the installation of both mandatory and optional addons, which reduces complexity. On the other, we can use "brain" components like the "CoreAddonController" to enforce that mandatory components are being deployed into the cluster.

Additionally, in my opinion, this proposal does not only build on top of the existing addon mechanism, but rather builds a foundation for new directions. Yes for the beginning we are still going to support the old docker-image source and go-template rendering method for compatibility. But we are also building a mechanism that allows us to easily add new sources and methods. For example we propose git as a new source and helm as a new rendering method.

docs/proposals/improve-addons.md
vgramer: @SimonTheLeg yes the `RequiredResourceTypes` works as @embik explain. it's a pre-requisite to render that addon's manifests.

Concerning the `Handle dependency **inside** an Addon ` I think too that `helm` should do the job

docs/proposals/improve-addons.md
SimonTheLeg: Good point, I have referenced the PR in the proposal

docs/proposals/improve-addons.md
SimonTheLeg: @embik we have updated the proposal to reflect that a change to the go-template method will be made. You can find the addendum in the `AddonInstallationController -> go-template` section of this document

docs/proposals/improve-addons.md
SimonTheLeg: @maciaszczykm

very good point. We actually realized that there was required functionality missing in the architecture. You can find it in the AddonController section as a separate bulletpoint.

Additionally I think we also need to update #6000 later on, as choosing addons is a bit more complicated due to the fact that the compatible addons need to be computed based on the chosen cluster.

docs/proposals/improve-addons.md
SimonTheLeg: Good point with the WebHook, we have added it to the proposal.

docs/proposals/improve-addons.md
SimonTheLeg: we have added it into the proposal

embik (approved):
/lgtm

I'm however not sure I'm the right person to "greenlight" this, so it's a good thing that others are taking a look too.

themue (approved):
/approve
/lgtm

SimonTheLeg:
@maciaszczykm Can you have a look at the updated version, which addresses your comments?

vgramer (approved):
/lgtm

maciaszczykm:
LGTM

SimonTheLeg:
@maciaszczykm 
If I understand our prow bot correctly you have to do `/approve` or `/lgtm` and then it auto-merges. I don't think it reacts directly to `LGMT`

embik:
@SimonTheLeg judging from the status, the problem is actually that your description doesn't include release note information (which can be "NONE", but that part of the description has to be there, I think). The approvals/lgtms are all there.

PR #8149, opened 2021-11-03
Set apiVersion of kubeadm-config based on Kubernetes version

### LINKED ISSUE #8039: Cannot join nodes to KKP cluster with kubeadm 1.22
The cluster creates fine in the KKP, but trying to join a node to the cluster fails with the following error message:

```
teo@kubermatic-testcluster:~$ sudo kubeadm join 6q5tmn8kv4.kubermatic.192-168-122-184.sslip.io:32157 --token [snip] --discovery-token-ca-cert-hash sha256:[snip] 
[preflight] Running pre-flight checks
[preflight] Reading configuration from the cluster...
[preflight] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -o yaml'
error execution phase preflight: unable to fetch the kubeadm-config ConfigMap: failed to decode cluster configuration data: no kind "ClusterConfiguration" is registered for version "kubeadm.k8s.io/v1beta1" in scheme "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/scheme/scheme.go:32"
To see the stack trace of this error execute with --v=5 or higher
```

Looking at the ConfigMap kube-system.kubeadm-config, we can verify that it is indeed stored as a ClusterConfiguration v1beta1:
```json
{
	"ClusterConfiguration": "apiServer:
		  extraArgs:
		    authorization-mode: Webhook
		  timeoutForControlPlane: 4m0s
		apiVersion: kubeadm.k8s.io/v1beta1
		certificatesDir: /etc/kubernetes/pki
		imageRepository: k8s.gcr.io
		kind: ClusterConfiguration
		kubernetesVersion: 1.22.2
		",
	"ClusterStatus": "apiVersion: kubeadm.k8s.io/v1beta1
		kind: ClusterStatus
		apiEndpoints:
		"
}
```

Looking at the release notes, support for kubeadm.k8s.io/v1beta1 was dropped in kubeadm 1.22 (in favour of v1beta2 and v1beta3).

Versions:
kubeadm version: &version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:37:34Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
child k8s cluster version: 1.22.2
kubermatic: v2.18.1 community edition

embik:

The v1beta1 API types have been removed from kubeadm in 1.22, thus kubeadm 1.22 cannot be used to join a cluster created via KKP's "kubeadm" provider type as the kubeadm configmap is using the removed API type. v1beta2 is available since kubeadm 1.15.0 [1], so v1beta2 should be save to use. v1beta3 can be used for Kubernetes 1.22+.

[1] https://pkg.go.dev/k8s.io/kubernetes@v1.15.0/cmd/kubeadm/app/apis/kubeadm/v1beta2

Fixes #8039

- Should this be backported to 2.18?

**Documentation**:

```release-note
kubeadm-config is updated based on Kubernetes version
```

### DIFF
--- a/addons/kubeadm-configmap/kubeadm-configmap.yaml
+++ b/addons/kubeadm-configmap/kubeadm-configmap.yaml
@@ -23,12 +23,20 @@ data:
       extraArgs:
         authorization-mode: Webhook
       timeoutForControlPlane: 4m0s
-    apiVersion: kubeadm.k8s.io/v1beta1
+    {{- if and (eq .Cluster.Version.Major 1) (le .Cluster.Version.Minor 21) }}
+    apiVersion: kubeadm.k8s.io/v1beta2
+    {{- else if and (eq .Cluster.Version.Major 1) (ge .Cluster.Version.Minor 22) }}
+    apiVersion: kubeadm.k8s.io/v1beta3
+    {{- end }}
     certificatesDir: /etc/kubernetes/pki
     imageRepository: k8s.gcr.io
     kind: ClusterConfiguration
     kubernetesVersion: {{ .Cluster.Version }}
   ClusterStatus: |
-    apiVersion: kubeadm.k8s.io/v1beta1
+    {{- if and (eq .Cluster.Version.Major 1) (le .Cluster.Version.Minor 21) }}
+    apiVersion: kubeadm.k8s.io/v1beta2
+    {{- else if and (eq .Cluster.Version.Major 1) (ge .Cluster.Version.Minor 22) }}
+    apiVersion: kubeadm.k8s.io/v1beta3
+    {{- end }}
     kind: ClusterStatus
     apiEndpoints:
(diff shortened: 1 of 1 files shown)


addons/kubeadm-configmap/kubeadm-configmap.yaml
```diff
@@ -23,12 +23,20 @@ data:
       extraArgs:
         authorization-mode: Webhook
       timeoutForControlPlane: 4m0s
-    apiVersion: kubeadm.k8s.io/v1beta1
+    {{- if and (eq .Cluster.Version.Major 1) (le .Cluster.Version.Minor 21) }}
```
xrstf, line 26: Wouldn't `.Cluster.Version.LessThan "1.22"` also work?

xrstf: Also, why do we need an `else if` and not just an `else`?

embik: I haven't worked with the semver library too much, but my understanding is that `LessThan` needs another `Version` (https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.LessThan), so I don't think passing a string (`"1.22"`) would work? Or is that incorrect?

The else is a good point, let me simplify it.

xrstf (approved):
/approve

PR #8055, opened 2021-10-21
Respect userCluster.overwriteRegistry from k8c config in user cluster controllers and addon manifests

dermorz:
**What this PR does / why we need it**:
The documentation of the field `userCluster.overwriteRegistry` states:
```
# OverwriteRegistry specifies a custom Docker registry which will be used for all images
# used inside user clusters (user cluster control plane + addons). This also applies to
# the KubermaticDockerRepository and DNATControllerDockerRepository fields.
``` 
But this was not the case. There were multiple components that directly used hardcoded registries without the possibility to override these registries.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #6562 

**Special notes for your reviewer**:
I am not yet sure about the `registry` package. If someone can show me a better place to put its content, I will adjust the PR.

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
The field `userCluster.overwriteRegistry` in the KubermaticConfiguration is now properly respected
by all provided addons and by more user cluster controllers: core-dns, envoy-agent, gatekeeper,
konnectivity, kubernetes-dashboard, mla-prometheus, mla-promtail, node-local-dns.
```

PR #8157, opened 2021-11-04
fix vsphere CSI addon manifests

xrstf:
Since #8055 we quote the docker images in the YAML, which unearthed the slightly broken version tags with a space at the end. This PR fixes that.

```release-note
NONE
```


xmudrii (approved):

moadqassem (approved):
/approve
/lgtm

PR #8160, opened 2021-11-04
Remove double implementation of registry override function

dermorz:
**What this PR does / why we need it**: This is a tiny refactoring follow-up PR to #8055. I expected there to be more potential for refactoring and clean up and I didn't want it in the initial PR where the fix was implemented to avoid distraction.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/resources/data.go
+++ b/pkg/resources/data.go
@@ -440,7 +440,7 @@ func (d *TemplateData) KubermaticAPIImage() string {
 func (d *TemplateData) parseImage(image string) string {
 	named, _ := reference.ParseNormalizedNamed(image)
 	domain := reference.Domain(named)
-	reminder := reference.Path(named)
+	remainder := reference.Path(named)
 
 	if d.OverwriteRegistry != "" {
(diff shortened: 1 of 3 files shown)


vgramer (commented):
/lgtm

pkg/resources/data.go
```diff
@@ -440,7 +440,7 @@ func (d *TemplateData) KubermaticAPIImage() string {
 func (d *TemplateData) parseImage(image string) string {
 	named, _ := reference.ParseNormalizedNamed(image)
 	domain := reference.Domain(named)
-	reminder := reference.Path(named)
+	remainder := reference.Path(named)
```
xrstf, line 443: But how will we now remember this?!

xrstf (approved):
/approve

pkg/resources/data.go
dermorz: This will forever *remin* a secret. 

:drum: *ba-dum-tss*

PR #8161, opened 2021-11-04
Add auditLogging.policyPreset that allows selecting from several audit policies

embik:

When enabling audit logging, KKP only offers one audit policy, which logs every request at the "Metadata" level. This PR adds a new field called `policyPreset` to the audit logging settings which allows to select from a list of pre-defined audit policy configurations:

- `metadata` is the same as the current policy, logging every request on the metadata level
- `minimal` is logging barely enough to satisfy the CIS Benchmark requirement. This logs request and response information for any change to a workload resource (Pods, Deployments, StatefulSets, etc), access to Pods via shells or network, and metadata for requests that access Pod logs, Secrets and ConfigMaps
- `recommended` logs the same items as `minimal` plus everything else on the metadata level. The CIS Benchmark _recommends_ logging "everything else" as metadata, but this setting has the highest output and as such, might not be suited for all users

Fixes https://github.com/kubermatic/edgematic/issues/72

- I'm sure I'm doing something wrong with the CRD/API integration, but I need some feedback on what's missing.

**Documentation**:

```release-note
User clusters now expose `spec.auditLogging.policyPreset`, which can be set to `metadata`, `minimal` or `recommended` to configure audit logging policy
```


pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -389,8 +389,19 @@ type OIDCSettings struct {
 	ExtraScopes   string `json:"extraScopes,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=metadata,recommended,minimal
```
kron4eg:
```suggestion
// +kubebuilder:validation:Enum=metadata;recommended;minimal
```

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -389,8 +389,19 @@ type OIDCSettings struct {
...
+const (
+	AuditPolicyMetadata    AuditPolicyPreset = "metadata"
+	AuditPolicyRecommended AuditPolicyPreset = "recommended"
+	AuditPolicyMinimal     AuditPolicyPreset = "minimal"
+)
+
 type AuditLoggingSettings struct {
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled      bool              `json:"enabled,omitempty"`
+	PolicyPreset AuditPolicyPreset `json:"policyPreset,omitempty"`
```
kron4eg, line 404: Because now `AuditPolicyPreset` from here and from crds package are two different types, you have to convert them manually in [clones.go](https://github.com/kubermatic/kubermatic/blob/61f27b5f1c46c90cc6d932b07c3919985389fa4b/pkg/install/crdmigration/clones.go#L409) otherwise it fails to compile like this:

```
 pkg/install/crdmigration/clones.go:409:70: cannot convert old.AuditLogging (type *"k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1".AuditLoggingSettings) to type *"k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1".AuditLoggingSettings 
```

Alternatively just don't sync APIs at all, we are doing this manually from time to time anyway.

embik: Let me fix that real quick, I think I know how.

pkg/apis/kubermatic/v1/cluster.go
embik: Good catch, oops. Thinking about it, should this maybe include `""` in the Enum as well, so you don't need set this field at all? That would make the API more backwards-compatible (if you don't set the field, you get `metadata`, which is the current behaviour).

embik: Or does that maybe happen already because of the `omitempty` on the field itself? Not sure how kubebuilder's validation interfaces with that.

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -389,8 +389,19 @@ type OIDCSettings struct {
 	ExtraScopes   string `json:"extraScopes,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=metadata;recommended;minimal
```
kron4eg: please rerun `./hack/update-codegen.sh` and also commit modified CRD files (in charts/kubermatic-operator), and don't forge to increment chart version in charts/kubermatic-operator/Chart.yaml

kron4eg: and `./hack/update-swagger.sh` too

kron4eg (approved):

embik:
@kron4eg thanks for the review, I had to bump the apiclient so the lgtm got removed, please take a quick look again when you have the time :-)

kron4eg (approved):

embik:
/retest

Third time's the charm.

PR #8174, opened 2021-11-08
Do not reconcile audit policy ConfigMap if policyPreset is empty

embik:
This is a follow-up to #8161 - after reading an old version of the audit logging documentation (https://docs.kubermatic.com/kubermatic/v2.16/tutorials/12-audit-logs/), it's clear that we advertised the feature of updating the audit policy ConfigMap after creating a cluster. The previous PR removed that functionality in favour of reconciling the ConfigMap at all times, but that might not be desirable.

This PR makes KKP only reconcile the ConfigMap if a policyPreset is set OR the ConfigMap is empty, which would be the case when initially creating it. After that, users can edit it with a custom policy, if they do not desire using a policy preset.

Fixes #

**Documentation**:

https://github.com/kubermatic/docs/pull/845

```release-note
NONE
```


kron4eg (approved):
/approve

PR #8193, opened 2021-11-09
Group alerts in Alertmanager and support multiple alerts in template

### LINKED ISSUE #8192: Alertmanager does not group alerts and only uses first alert in templates
**User Story**

The current Alertmanager configuration does not apply any grouping to alerts via `group_by`, which means that alerts are grouped into one big "ungrouped" group. This doesn't work well with the Slack templates we provide, as they select the first alert from the alert batch that is processed. Since all alerts are in one big group, the templates drop alerts if there is more than one firing at the moment.

**Acceptance criteria**

Alerts sent via Slack show all active alerts and are grouped sensibly.

embik:
No grouping in Alertmanager routing means that every alert is grouped into the same group. This PR provides a sensible top-level `group_by` setting for KKP environments and updates the Slack templates to support multiple alerts sent from a group in a batch (the current kubermatic template - if used, which isn't the default, I think - drops everything except the first alert).

Fixes #8192

**Documentation**:

```release-note
Alertmanager groups alerts by alertname, namespace, seed_cluster, cluster labels
```

### DIFF
--- a/charts/monitoring/alertmanager/Chart.yaml
+++ b/charts/monitoring/alertmanager/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: alertmanager
-version: 2.1.2
+version: 2.1.3
 appVersion: v0.22.2
 description: Alertmanager for Kubermatic
 keywords:
--- a/charts/monitoring/alertmanager/kubermatic.tmpl
+++ b/charts/monitoring/alertmanager/kubermatic.tmpl
@@ -2,40 +2,38 @@
 {* Capable of displaying the user-cluster from which alerts originate. *}
 {* Shows a pretty flag if the seed cluster name matches regex. *}
 
-{{ define "slack.kubermatic.pretty.runbook" }}{{ with .Annotations.runbook_url }}<{{ .}}|:notebook:>{{ end }}{{ end }}
+{{ define "slack.kubermatic.pretty.runbook" }}{{ with .Annotations.runbook_url }}<{{ . }}|:notebook:>{{ end }}{{ end }}
 {{ define "slack.kubermatic.titlelink" }}{{ end }}
 {{ define "slack.kubermatic.pretty.icon" }}{{ end }}
 {{ define "slack.kubermatic.color" }}{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}{{ end }}
 
 {{ define "slack.kubermatic.pretty.labels" -}}
-{{- with .Labels.seed_cluster -}}
+{{- with .CommonLabels.seed_cluster -}}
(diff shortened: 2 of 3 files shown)


xrstf:
Can you show how these alerts would look like? I remember that many moons ago I removed the grouping on purpose, because the alerts in slack were absolutely useless: https://github.com/kubermatic/kubermatic/pull/2650#issue-400645373

embik:
> Can you show how these alerts would look like? I remember that many moons ago I removed the grouping on purpose, because the alerts in slack were absolutely useless: [#2650 (comment)](https://github.com/kubermatic/kubermatic/pull/2650#issue-400645373) -- not only were they hard to group, but the alerts were gigantic in size. You could only fit like 3 on your Slack screen.

Ah, interesting! I didn't find that PR when looking for the reason. This does look like my PR reverts parts of it. I'll try to set something up for something visual.

In the end, it comes down to alert grouping being a core mechanism of Alertmanager. Without accounting for it, you lose information, like we did with the cert-manager alerts, as they were most likely sent at the same time as another alert. So basically this "improved" the alerting situation by hiding some alerts :-) The only per-alert information this template gives is the message, which will and can be different for every alert. I'm already omitting any label information though.

xrstf:
I still don't quite understand why _not grouping_ hides certain alerts.

> Without accounting for it, you lose information, like we did with the cert-manager alerts, as they were most likely sent at the same time as another alert.

So Alertmanager only sends 1 HTTP request to Slack per minute and ignores everything else? o.O

embik:
> I still don't quite understand why _not grouping_ hides certain alerts.
> 
> > Without accounting for it, you lose information, like we did with the cert-manager alerts, as they were most likely sent at the same time as another alert.
> 
> So Alertmanager only sends 1 HTTP request to Slack per minute and ignores everything else? o.O

Because not setting `group_by` isn't the same as _not grouping_ alerts. It makes Alertmanager aggregate all alerts into one group, the "ungrouped" group, see https://groups.google.com/g/prometheus-users/c/zYRIT_cKxmw.

If you wanted to disable aggregation entirely, this would be the setting (as per https://prometheus.io/docs/alerting/latest/configuration/#route):

```yaml
group_by: ['...']
```

That is discouraged by the Alertmanager devs though.

embik:
Screenshot 2021-11-09 at 13 08 29

How two alerts grouped together look like. I can move around the counter (the "[ 2 ]"), I'm not sure I like it there myself.

embik:
Updated screenshots:

Screenshot 2021-11-09 at 13 56 14

One of the two alerts getting resolved:

Screenshot 2021-11-09 at 13 55 26

Both alerts having resolved:

Screenshot 2021-11-09 at 13 58 44

xrstf (approved):
/approve

PR #8202, opened 2021-11-09
Pull dependencies for helm charts during installation.

### LINKED ISSUE #8163: Add features for handling charts with dependencies in kubermatic-installer

wurbanski:
To enable usage of upstream charts as dependencies in our charts, installer needs to handle downloading the dependencies up-to-spec, e,.g. using `helm dependency build` (not update).

Fixes #8163

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/install/helm/cli.go
+++ b/pkg/install/helm/cli.go
@@ -62,6 +62,9 @@ func (c *cli) BuildChartDependencies(chartDirectory string, flags []string) erro
 	}
 
 	chart, err := LoadChart(chartDirectory)
+	if err != nil {
+		return err
+	}
 
 	for idx, dep := range chart.Dependencies {
 		repoAddFlags := []string{
--- a/pkg/install/helm/cli.go
+++ b/pkg/install/helm/cli.go
@@ -68,14 +68,14 @@ func (c *cli) BuildChartDependencies(chartDirectory string, flags []string) (err
 
 	for idx, dep := range chart.Dependencies {
 		repoAddFlags := []string{
-			"repository",
+			"repo",
 			"add",
 			fmt.Sprintf("dep-%s-%d", chart.Name, idx),
 			dep.Repository,
 		}
 
 		repoRemoveFlags := []string{
-			"repository",
+			"repo",
 			"remove",
 			fmt.Sprintf("dep-%s-%d", chart.Name, idx),
 		}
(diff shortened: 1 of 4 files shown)


pkg/install/helm/cli.go
```diff
@@ -55,6 +55,47 @@ func NewCLI(binary string, kubeconfig string, kubeContext string, timeout time.D
 	}, nil
 }
 
+func (c *cli) BuildChartDependencies(chartDirectory string, flags []string) (err error) {
+	command := []string{
+		"dependency",
+		"build",
+	}
```
xrstf: Can you move this declaration down to where it's needed? I lost track of what "command" was (granted, I read code at like 9000% zoom level ^^)

pkg/install/helm/cli.go
```diff
@@ -55,6 +55,47 @@ func NewCLI(binary string, kubeconfig string, kubeContext string, timeout time.D
...
+
+		_, err = c.run("default", repoAddFlags...)
+	}
+
+	command = append(command, chartDirectory)
+
+	_, err = c.run("default", command...)
+
+	return err
+}
```
xrstf, line 101: Please add an empty line between functions :)

pkg/install/helm/cli.go
```diff
@@ -55,6 +55,47 @@ func NewCLI(binary string, kubeconfig string, kubeContext string, timeout time.D
...
+		}
+
+		defer func() {
+			_, removeErr := c.run("default", repoRemoveFlags...)
+			if err != nil {
+				err = removeErr
+			}
+		}()
+
+		_, err = c.run("default", repoAddFlags...)
```
xrstf: The defer should only happen after the `repo add` already succeeded, no?

Also, in case of an error, why not `break` out of the loop and skip needlessly adding and removing more repos? :)

xrstf (changes requested):
Small nitpicks

pkg/install/helm/cli.go
wurbanski: 1) I'd rather have the removal "scheduled" before I add the repo, since it can be "added" and then fail during update, which would lead to a repo created in helm anyway, just not working correctly. `repo remove` cleans it up anyway.

2) yeah, I will do that :)

pkg/install/helm/cli.go
wurbanski: Done. :)

pkg/install/helm/cli.go
wurbanski: Good note, done.

kron4eg (approved):
/approve

PR #8208, opened 2021-11-10
Do not deploy unnecessary resources in Konnectivity setup

### LINKED ISSUE #7489: Cleanup of unnecessary resources in Konnectivity setup
Make sure that when Konnectivity is enabled, unnecessary resources are not deployed in user cluster and seed, as described in the proposal: https://github.com/PratikDeoghare/kubermatic/blob/master/docs/proposals/konnectivity-openvpn.md#cleanup-of-unnecessary-resources

rastislavs:
When Konnectivity is used for control-plane to worker-nodes communication, several old control plane components in Seed become unnecessary, and should not be deployed:

- openvpn-server deployment,
- openvpn-client sidecar containers in controller-manager, scheduler, metrics-server,
- dnat-controller sidecar container in metrics-server,

The following will be removed as a separate change, after further verification (https://github.com/kubermatic/kubermatic/issues/8256):
- dns-resolver deployment,
- custom dnsConfig in apiserver, controller-manager, scheduler pods.

In the user cluster, openvpn-client deployment becomes unnecesary as well.

This PR makes sure they are not deployed if Konnectivity is enabled for the user cluster.

Fixes #7489 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/addons/openvpn/openvpn-client-dep.yaml
+++ b/addons/openvpn/openvpn-client-dep.yaml
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{ if not .Cluster.Network.KonnectivityEnabled }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -128,4 +127,3 @@ spec:
       - name: openvpn-client-config
         configMap:
           name: openvpn-client-config
-{{ end }}
--- a/addons/openvpn/openvpn-client-dep.yaml
+++ b/addons/openvpn/openvpn-client-dep.yaml
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{ if not .Cluster.Network.KonnectivityEnabled }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -127,3 +128,4 @@ spec:
       - name: openvpn-client-config
         configMap:
           name: openvpn-client-config
+{{ end }}
--- a/pkg/resources/cloudcontroller/kubevirt.go
+++ b/pkg/resources/cloudcontroller/kubevirt.go
@@ -70,18 +70,7 @@ func kubevirtDeploymentCreator(data *resources.TemplateData) reconciling.NamedDe
 				Labels: podLabels,
 			}
 
-			dep.Spec.Template.Spec.DNSPolicy, dep.Spec.Template.Spec.DNSConfig, err =
-				resources.UserClusterDNSPolicyAndConfig(data)
-			if err != nil {
-				return nil, err
-			}
-
 			dep.Spec.Template.Spec.AutomountServiceAccountToken = pointer.BoolPtr(false)
-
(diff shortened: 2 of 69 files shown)


PratikDeoghare:
nice! didn't know we had to update so many places. 

/lgtm

addons/openvpn/openvpn-client-dep.yaml
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{ if not .Cluster.Network.KonnectivityEnabled }}
```
mfranczy: could not we just skip the whole openvpn addon in the addont controller if Konnectivity is enabled? Instead of editing each manifest.

pkg/resources/cloudcontroller/kubevirt.go
```diff
@@ -110,9 +98,22 @@ func kubevirtDeploymentCreator(data *resources.TemplateData) reconciling.NamedDe
 			}
 
 			defResourceRequirements := map[string]*corev1.ResourceRequirements{
-				ccmContainerName:    kvResourceRequirements.DeepCopy(),
-				openvpnSidecar.Name: openvpnSidecar.Resources.DeepCopy(),
+				ccmContainerName: kvResourceRequirements.DeepCopy(),
 			}
+
+			if !data.IsKonnectivityEnabled() {
```
mfranczy, line 110: that can bring issues to user cluster if both k8s clusters (KubeVirt and user cluster) uses node local dns (a loop problem). Why if konnectivity is enabled it's not allowed to change DNS policy and add custom servers?

addons/openvpn/openvpn-client-dep.yaml
rastislavs: You are right, it probably would be nicer to skip it in the controller, I can change that.

pkg/resources/cloudcontroller/kubevirt.go
rastislavs: Not sure if I am getting the problem with the loop, but maybe it will help if I describe the rationale behind this change:

Before this change, the cloud controller pod in seed (as well as some other control plane components running in the seed) had the DNS policy set to `None` and instead of using the default k8s DNS of the seed cluster, it was configured with a custom DNS resolver running for the user cluster in the seed. I was under the impression that the sole role of this custom DNS resolver in seed was to resolve the user cluster services to pod IPs when control plane components (mainly apiserver) needs to contact them (e.g. extension apiservers, webhooks). 

This functionality is now built in Konenctivity for the apiserver, so the custom dns resolver in the seed for each user cluster is not necessary, at least for this purpose.

But I was not aware of any other role of this dns-resolver, e.g. preventing loops in case of KubeVirt as you are mentioning. If it is really needed, maybe we could still deploy it for KubeVirt user clusters only?

mfranczy (approved):
/approve
/lgtm

pkg/resources/cloudcontroller/kubevirt.go
rastislavs: The DNS change will be done as a separate change: https://github.com/kubermatic/kubermatic/issues/8256

mfranczy (approved):
/approve
/lgtm

PR #8210, opened 2021-11-10
ensure clusters are deleted after CCM migration e2e tests

xrstf:
We are currently leaking resources in this test, as the tests only deletes the kind cluster at the end, but not the userclusters. This PR fixes that. It also removes one ugly `./hack/run-ccm-migration-e2e-test-in-kind.sh: line XXX: type: ginkgo: not found` warning, which was misleading.

```release-note
NONE
```


mlavacca (approved):
/approve

Issue #6248, opened 2020-12-03
API Openstack field "tenant" is deprecated name


furkhat:
"tenant" is deprecated in OpenStack. This is what is included in the "OpenStack RC File v3". Can we rename it to "project" or "os_project"?

furkhat:
nope still true /remove-lifecycle stale

vgramer:
- [ ] update machine-controller to handle project auth vars -> kubermatic/machine-controller#1091
- [ ] bump machine controller in KPP
- [ ] update KPP to handle project auth vars 
- [ ] update dashboard

PR #8211, opened 2021-11-10
handle Openstack project auth variable

vgramer:

 In Openstack auth with tenant/TenantID are deprecated in favor of project/projectID.  But these variables contain the same value.

This PR:
* update  the crds in `pkg/crd/kubermatic/v1/preset.go` and `pkg/crd/kubermatic/v1/cluster.go` to add new optional fields `project` / `projectID`, make `tenant` fields optional. Instead of direct accesing fileds, we use the `GetProjectOrDefaultToTenant` and  `GetProjectIdOrDefaultToTenantId` that first try to read project variables and if it not defined fallback to tenant variable
* update crds `pkg/apis/kubermatic/v1/preset.go`  and `pkg/apis/kubermatic/v1/cluster.go` to have only the field  `project` and `projectID`. 
* update the migration tool (ie `pkg/install/crdmigration/clones.go`) accordingly.
* update others struct like `OpenstackCredentials` to use project variables 
* generate `credentials-openstack-<cluster-id>` secret with the `project` field. when we read this secret, use the same fallback mechanism to `tenant` variable
* update `OpenstackPlublicCloudSpec` to also return `project` variables
* update routes `/api/v1/providers/openstack/XXX` to handle auth `project` variables using the same fallback mechanism to `tenant` variables

ref #6248

a review by commit may be easier

**Documentation**:

```release-note
Add support for Openstack authentication with `project name` and  `project id`. `tenant name`  and `tenant id`  are deprecated
```


kron4eg (approved):
/approve

kron4eg (approved):

pkg/apis/kubermatic/v1/preset.go
```diff
@@ -326,11 +326,11 @@ type Openstack struct {
...
 
-	Username string `json:"username"`
-	Password string `json:"password"`
-	Tenant   string `json:"tenant"`
-	TenantID string `json:"tenantID"`
-	Domain   string `json:"domain"`
+	Username  string `json:"username"`
+	Password  string `json:"password"`
+	Project   string `json:"project"`
+	ProjectID string `json:"projectID"`
```
maciaszczykm, line 332: Remember to adjust presets so dev and CI will not break.

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -734,6 +736,26 @@ type OpenstackCloudSpec struct {
 	UseOctavia *bool `json:"useOctavia,omitempty"`
 }
 
+// GetProjectOrDefaultToTenant returns the the project if defined otherwise fallback to tenant
+// Deprecated: the tenant auth var is depreciated in openstack. In pkg/apis/kubermatic/v1/cluster.go we will only use Project
+func (s OpenstackCloudSpec) GetProjectOrDefaultToTenant() string {
```
maciaszczykm: I'd make the name simpler, i.e. `getProject` and then put info about defaulting in comment.

```go
// GetProject returns the the project if defined otherwise fallback to tenant which was deprecated.
func (s OpenstackCloudSpec) getProject() string {
```

pkg/crd/kubermatic/v1/cluster.go
```diff
@@ -734,6 +736,26 @@ type OpenstackCloudSpec struct {
...
+	if len(s.Project) > 0 {
+		return s.Project
+	} else {
+		return s.Tenant
+	}
+}
+
+// GetProjectIdOrDefaultToTenantId returns the the projectID if defined otherwise fallback to tenantID
+// Deprecated: the tenantID auth var is depreciated in openstack. In pkg/apis/kubermatic/v1/cluster.go we will only use ProjectID
+func (s OpenstackCloudSpec) GetProjectIdOrDefaultToTenantId() string {
```
maciaszczykm: As above.

pkg/crd/kubermatic/v1/preset.go
```diff
@@ -357,9 +380,10 @@ func (s Openstack) IsValid() bool {
 		return len(s.ApplicationCredentialSecret) > 0
 	}
 
+	projectOrTenant := (len(s.Project) > 0 || len(s.ProjectID) > 0) || (len(s.Tenant) > 0 || len(s.TenantID) > 0)
```
maciaszczykm: s/projectOrTenant/hasProjectOrTenant

So the logic is any of these four is enough? Then you can remove the brackets as they do not change anything.

pkg/handler/v1/provider/openstack.go
```diff
@@ -373,12 +378,18 @@ type OpenstackReq struct {
 	// Domain OpenStack domain name
 	Domain string
 	// in: header
-	// Tenant OpenStack tenant name
+	// Tenant OpenStack tenant name (dredecated use Project instead)
```
maciaszczykm: Typo.

pkg/handler/v1/provider/openstack.go
```diff
@@ -373,12 +378,18 @@ type OpenstackReq struct {
...
 	Tenant string
 	// in: header
-	// TenantID OpenStack tenant ID
+	// TenantID OpenStack tenant ID (dredecated use ProjectID instead)
```
maciaszczykm: As above.

pkg/provider/cloud/openstack/provider.go
```diff
@@ -746,31 +747,43 @@ func GetCredentialsForCluster(cloud kubermaticv1.CloudSpec, secretKeySelector pr
...
+		ProjectID:                   projectID,
 		Domain:                      domain,
 		ApplicationCredentialID:     applicationCredentialID,
 		ApplicationCredentialSecret: applicationCredentialSecret,
 	}, nil
 }
 
+// firstKeyOrFallback read the secret and return value for the firstkey. if the firstKey does not exist, tries with
+// fallbackKey. if the fallbackKey does not exist then return an error
+func firstKeyOrFallback(secretKeySelector provider.SecretKeySelectorValueFunc, configVar *providerconfig.GlobalSecretKeySelector, firstKey string, fallbackKey string) (string, error) {
```
maciaszczykm: s/firstKeyOrFallback/getFirstKey

I think we do not need to inform about fallback in the method name.

maciaszczykm:
I have some concerns about having 2 fields for the same thing. @kubermatic/sig-api please check it, do you have any remarks?

pkg/apis/kubermatic/v1/preset.go
vgramer: @maciaszczykm  beside `hack/ci/run-api-e2e.sh` there is any other adjustment to do?

pkg/crd/kubermatic/v1/cluster.go
maciaszczykm: It will be easier to switch to using project only once deprecation period will end.

pkg/validation/cluster.go
```diff
@@ -388,10 +388,12 @@ func validateOpenStackCloudSpec(spec *kubermaticv1.OpenstackCloudSpec, dc *kuber
...
 	var errs []error
-	if spec.Tenant == "" && spec.CredentialsReference != nil && spec.CredentialsReference.Name != "" {
+	if spec.GetProjectOrDefaultToTenant() == "" && spec.CredentialsReference != nil && spec.CredentialsReference.Name != "" {
+		errs = append(errs, kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackProject))
 		errs = append(errs, kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackTenant))
 	}
-	if spec.TenantID == "" && spec.CredentialsReference != nil && spec.CredentialsReference.Name != "" {
+	if spec.GetProjectIdOrDefaultToTenantId() == "" && spec.CredentialsReference != nil && spec.CredentialsReference.Name != "" {
+		errs = append(errs, kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackProjectID))
 		errs = append(errs, kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.OpenstackTenantID))
```
maciaszczykm: I'm wondering if we should validate both fields if only one will be used. Shouldn't second field be ignored if the first is specified?

pkg/apis/kubermatic/v1/preset.go
maciaszczykm: The presets that are defined in our Vault should be updated to use the new field once this merges. It should not break anything until there is fallback though.

pkg/crd/kubermatic/v1/cluster.go
vgramer: done :)

pkg/crd/kubermatic/v1/preset.go
vgramer: done :)

pkg/handler/v1/provider/openstack.go
vgramer: done :)

pkg/crd/kubermatic/v1/cluster.go
vgramer: done :)

pkg/handler/v1/provider/openstack.go
vgramer: done :)

pkg/provider/cloud/openstack/provider.go
vgramer: done :)

pkg/validation/cluster.go
vgramer: @maciaszczykm  i updated it to generate the following error

```go
fmt.Errorf("%q or %q cannot be empty", resources.OpenstackProject, resources.OpenstackTenant)
```
to stay compatible, I don't think I can do better :/

maciaszczykm: Okay. Last change would be updating the error message to something like "%q and %q cannot be empty at the same time". What do you think?

PR #8239, opened 2021-11-16
new endpoint endpoint to list External MD Metrics 

### LINKED ISSUE #8235: Crete a new endpoint listExternalClusterMachineDeploymentMetrics
**User Story**
Create a new endpoint to list external cluster MD metrics
Right now we have only for the whole cluster

zreigz:
**What this PR does / why we need it**: new endpoint to list External MD Metrics 

Fixes #8235

```release-note
new endpoint to list External MD Metrics:
/api/v2/projects/{project_id}/kubernetes/clusters/{cluster_id}/machinedeployments/{machinedeployment_id}/nodes/metrics
```


pkg/handler/v2/external_cluster/node.go
```diff
@@ -84,52 +84,56 @@ func ListNodesMetricsEndpoint(userInfoGetter provider.UserInfoGetter, projectPro
...
-			for _, n := range nodes.Items {
-				availableResources[n.Name] = n.Status.Allocatable
-			}
-
-			nodeDeploymentNodesMetrics := make([]v1beta1.NodeMetrics, 0)
-			allNodeMetricsList := &v1beta1.NodeMetricsList{}
-			if err := client.List(ctx, allNodeMetricsList); err != nil {
-				return nil, common.KubernetesErrorToHTTPError(err)
+		for _, m := range allNodeMetricsList.Items {
+			if _, ok := availableResources[m.Name]; ok {
```
imharshita, line 129: what is the need for this? I mean we are getting availableResources, metrics for the same pair of nodes, so why are we checking if availableResources has the node data

pkg/handler/v2/external_cluster/node.go
```diff
@@ -84,52 +84,56 @@ func ListNodesMetricsEndpoint(userInfoGetter provider.UserInfoGetter, projectPro
...
+	if isMetricServer {
+		client, err := clusterProvider.GetClient(cluster)
 		if err != nil {
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
-		cluster, err := getCluster(ctx, userInfoGetter, clusterProvider, privilegedClusterProvider, project.Name, req.ClusterID)
-		if err != nil {
-			return nil, common.KubernetesErrorToHTTPError(err)
+		nodes := &corev1.NodeList{}
+		if err := client.List(ctx, nodes); err != nil {
```
imharshita, line 114: Can we filter out nodes for the desired machinedeployment here before fetching the metrics like this https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/external_cluster/gke.go#L168

imharshita (changes requested):

pkg/handler/v2/external_cluster/node.go
zreigz: availableResources is a map with available resources for the node
allNodeMetricsList := &v1beta1.NodeMetricsList{} - taken from metrics server
when we combine those two structures then we get the whole metrics structure for the node

This whole block is taken from `kubectl top node` implementation

pkg/handler/v2/external_cluster/node.go
zreigz: this is a general method for all providers and also for the connected clusters from kubeconfig, so the whole implementation work with all provider. Adding filtering here will cause additional maintenance in the future when we will add a new provider. 
It's used in two places: one without filtering (get metrics for cluster nodes) and second when the filtering is done automatically without distinguishing the provider type (filter out MD nodes)

pkg/handler/v2/external_cluster/node.go
imharshita: Yes I got that, but both availableResources and allNodeMetricsList are for same set of nodes, right?
then we are running a loop to check if availableResources has the nodes from allNodeMetricsList

PR #8252, opened 2021-11-18
Use upstream helm chart for loki as a dependency

### LINKED ISSUE #7901: update chart to use upstream: logging/loki
(general description in the epic)

Acceptance criteria:
* evaluated if it's possible to use and extend the chart to use upstream as a dependency
* required changes are implemented
* CI configuration is added
* migration and changes required to upgrade are documented

wurbanski:
This PR switches our internal chart to use the upstream chart for Loki. Includes "golden master" test.

Fixes #7901

**Documentation**:

```release-note
loki chart has been updated to use upstream chart from https://grafana.github.io/helm-charts as a dependency
```


xrstf (approved):
/approve

charts/logging/loki/test/default.yaml
```diff
@@ -11,14 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
SimonTheLeg, line 13: I assume this file is still required so the `test-chart-rendering.sh` script is being triggered. If so I think it makes sense to add a "This file is required by test-chart-rendering.sh. Do not delete it!", because if someone deletes the file later in some sort of "cleanup repo", it would not be picked up by CI.

charts/logging/loki/values.yaml
```diff
@@ -80,23 +83,26 @@ loki:
...
 
   podLabels: {}
 
+  podAnnotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "3100"
+
+  rbac:
+    pspEnabled: false
+
```
SimonTheLeg, line 105: I think we should move these into a separate PR, as they change application behaviour and go beyond using the upstream helm chart. If that is too much of a hassle with changing all the test outputs, then I would at least place them in a separate commit. Even though we squash all commits, I think it still makes it easier to understand what the PR has changed.

SimonTheLeg (approved):
/lgtm

only minor comments

charts/logging/loki/test/default.yaml
wurbanski: Fair point, will add that

charts/logging/loki/values.yaml
wurbanski: I'm pretty sure it just puts everything to the way it was before switching to the upstream, e.g. to the "defaults" that we were using, [output diff](https://github.com/kubermatic/kubermatic/pull/8252/commits/254140d41667c03809ee1af334d9706a2b6e4c64#diff-eca8a05f6cbb885406e88aee2f5b8acbda5b4bd4f295790774a3157ef2049cd5) seems to confirm that unless I'm missing something?

SimonTheLeg: Okay maybe I misunderstood then, but my thinking was the following. Let's take the `persistence.enabled` as an example:
- I looked through our [current values](https://github.com/kubermatic/kubermatic/blob/master/charts/logging/loki/values.yaml), where it's not set to anything
- Then I looked up the [old upstream](https://github.com/grafana/helm-charts/blob/loki-2.3.0/charts/loki/values.yaml#L139) (which I assumed is the one we cloned) where it is set to `false`
- additionally I looked through the values of the new [upstream chart](https://github.com/grafana/helm-charts/blob/main/charts/loki/values.yaml#L142), where it is also set to `false`

And then I came to the conclusion that this would be set to the opposite. But maybe I overlooked something and we are setting the value to `true` somewhere in our old chart 🤔 ?

wurbanski: Our currently used chart is not directly copied from the upstream, so there are some differences based on assumptions in KKP. Our current chart does not need need to have `persistence.enabled` because it is enough to define a PVC to use it. In the upstream chart, you have to add `persistence.enabled` AND define a PVC for it to work. Our charts are basically a set of default configurations and optional additions to upstream charts, so to keep compatibility with older charts, some options have to be set differently than in the upstream.

The link to diff in test output is the best way to see what has been changed, as it shows the effect of installing the chart with default values.

PR #8273, opened 2021-11-23
Fetch chart dependencies on dev before installing logging and monitoring

wurbanski:
Fixes deployments on dev cluster after adding dependencies to the charts.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/hack/ci/deploy.sh
+++ b/hack/ci/deploy.sh
@@ -90,6 +90,8 @@ chmod 600 "$KUBECONFIG"
 echodate "Deploying ${DEPLOY_STACK} stack..."
 case "${DEPLOY_STACK}" in
 monitoring)
+  echodate "fetching chart dependencies..."
+  hack/ci/fetch-chart-dependencies.sh
   deploy "node-exporter" "monitoring" charts/monitoring/node-exporter/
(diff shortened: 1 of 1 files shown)


embik:
Do we need to think about how this affects the official KKP installation process? I assume customers would face the same error as the dev installation script when following the KKP documentation (as it doesn't include the actions taken by the new ci script). Do we need to vendor the chart dependencies into git maybe?

wurbanski:
Installer takes care of downloading the dependencies (since #8202) - is non-installer method still recommended? if yes, then we should add this to documentation.

embik:
Ah never mind my comment then, I wasn't aware of that!

xrstf:
Yes, we should document the manual installation, it's important for customers with non-standard requirements (e.g. those who do not want to use nginx, but .... öhm ..... IIS?).

xrstf (approved):
/approve

wurbanski:
> Yes, we should document the manual installation, it's important for customers with non-standard requirements (e.g. those who do not want to use nginx, but .... öhm ..... IIS?).

I'll create follow-up issue for that (#8275)

Issue #8221, opened 2021-11-12
Write UserStories for Applications


SimonTheLeg:
This issue is mainly used to track the current progress on writing User Stories for the Addon Proposal for dailys

SimonTheLeg:
Currently we have created a scratch-pad to track the different workstreams before writing the stories. 
It can be found here: https://app.nuclino.com/Kubermatic/Team-Caffeine-Overflow/Addon-UserStories-Scratchpad-c4c477dd-3fc0-4864-ba79-8d2a410b649f

Issue #7974, opened 2021-10-12
Introduce Applications


themue:
This epic targets the introduction of a new mechanism for users to install their applications. For improvements to the Addon-mechanism, which is used by Kubermatic to deploy mandatory components, see https://github.com/kubermatic/kubermatic/issues/6180. 

The current implementation of Addons has some flaws that make the usage and maintenance of Addons cumbersome for cluster administrators:

- Mandatory core components of user clusters are named “Default Addons” which implies that these are optional
- To add “Custom Addons” the cluster admin needs to bundle manifests and templates for these Addons in a Docker image that is based on the official addon-manifest docker image by Kubermatic
- It’s currently only possible to deploy multiple instances of the same Addon when using a workaround
- Complex Addons can leave a trail of reconciliation errors until successfully deployed because all manifest files are applied at the same time
- Addons are strictly tied to KKP versions due to the composition logic of the manifest image tag; for any provided manifest image the tag gets overwritten with `:<KKP VERSION>[-CUSTOM SUFFIX]`
- Creating a cluster is a two-step process; firstly you have to create a cluster and only  afterwards you can select custom Addons

## Default Addon

A *Default Addon* is installed on all user clusters based on configuration. The user can not delete *Default Addons*. These will remain unchanged after this proposal

- Shipped by Kubermatic,  tied to KKP version
- Tested by Kubermatic (Kubernetes versions)
- Supported by Kubermatic

## Application

An *Application* is an additional component installed in all user clusters on-demand, but not shipped with KKP. It is provided by the customer or a third party (e.g. helm community):

- Shipped by the customer
- Not tested nor supported by Kubermatic
- Could break clusters
- The customer is fully responsible

## Application by Kubermatic

An *OptionalApplication* is installed in a user clusters on demand during runtime or at cluster creation. The user can also uninstall and reinstall them. They are visible in the KKP UI.

- Shipped by Kubermatic, tied to KKP version (KKP version will be enforced using the aforementioned constraint feature)
- Tested by Kubermatic (Kubernetes versions and customizations)
- Supported by Kubermatic

*Note: From a technical perspective, Applications and Applications by Kubermatic will be handled the same way. However from a contractual perspective, it is important to differentiate: Applications by Kubermatic are supported by Kubermatic, while regular Applications are not.*

vgramer:
the first version of application feature has been delivered in KKP 2.21

PR #8276, opened 2021-11-23
Convert Addon Proposal to Application Proposal

### LINKED ISSUE #8304: Convert Addon-Proposal to Application-Proposal
In a separate meeting, we decided to move away from managing DefaultAddons due to the following reasons:

- Default Addons are distinct from Custom Addons and often require other components outside of the user cluster (e.g. kube-metrics)
- We decided that decoupling Default Addons from KKP versions is very difficult to support. This is because it allows the usage of all combinations of KKP and Addon versions, which are hard to test for us in advance. As a result we could run into a variety of unexpected errors

**User Story**

As a result we have to adapt our original proposal to adapt to these new changes. Additionally in order to avoid confusion with the existing Addon-mechanis, it was decided to rename the new approach into "Applications"

SimonTheLeg:

**What this PR does / why we need it**
Context for the change: After the inital merge in #8125, there have been concerns regarding tackling both Default and Custom Addons. In the end, we have decided to simplify the proposal and only focus on components a user should control in their cluster, not mandatory components provided by Kubermatic.
We decided to move a way from this due to the following factors:
  - Default Addons are distinct from Custom Addons and often require other components outside of the user cluster (e.g. kube-metrics)
  - We decided that decoupling Default Addons from KKP versions is very difficult to support. This is because it allows the usage of all combinations of KKP and Addon versions, which are hard to test for us in advance. As a result we 

**Side Notes**
The file is currently still called `improve-addons.md`. This is so git displays a proper diff, which makes it easier for reviewing. Shortly before merging, I am going to rename the file.

Pre-req for #7974, solves #8221
fix #8304

```release-note
None
```


docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
+Note on CRDs in general: The ApplicationDefinition strongly benefits from [OpenAPIv3's `oneOf` functionality](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/), which is supported by Kubernetes. For example in the yaml above, you could say a source must be `oneOf` `docker` or `git`. The cool thing about `oneOf` is that `docker` and `git` can have totally have different fields from each other, but still would be statically validated. However, syntax like this is [not yet supported in kubebuilder](https://github.com/kubernetes-sigs/controller-tools/issues/461), which we use to generate the CRDs. Nonetheless, we would still propose to use kubebuilder for now as the advantages of an automated generation outweigh the downside of not being able to do oneOf (yet).
```
vgramer:
```suggestion
Note on CRDs in general: The ApplicationDefinition strongly benefits from [OpenAPIv3's `oneOf` functionality](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/), which is supported by Kubernetes. For example in the yaml above, you could say a source must be `oneOf` `helm` or `git`. The cool thing about `oneOf` is that `helm` and `git` can have totally have different fields from each other, but still would be statically validated. However, syntax like this is [not yet supported in kubebuilder](https://github.com/kubernetes-sigs/controller-tools/issues/461), which we use to generate the CRDs. Nonetheless, we would still propose to use kubebuilder for now as the advantages of an automated generation outweigh the downside of not being able to do oneOf (yet).
```

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
 
-### AddonController
+### Updates to KKP API
 
-The AddonController resides in the seed cluster. Each seed cluster has its own instance of an AddonController. It has two main jobs:
+- adding an endpoint to list compatible Applications given a set of constraints (eg k8s version). This will be required to make it possible for users to select applications in the cluster wizard (see #6000)
+- creating ApplicationsInstallations for each user cluster. It is important to note that it only creates the CRs which Application should be installed in which cluster, but does not do the installation of the Application itself (this task will be done by the ApplicationInstallationController).
```
vgramer:
```suggestion
- adding an endpoint to install an application in a user cluster. It is important to note that it only creates the CRs which Application should be installed in which cluster, but does not do the installation of the Application itself (this task will be done by the ApplicationInstallationController).
```

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
+- create the seed-namespace right after the cluster object has been created. We need to make sure this early creation does not interfere with the seed-controller
+- create the application CRs after the namespace has been created
 
-In order to protect against accidental deletion of an AddonInstallation of type Core, we propose to make use of k8s [finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/). In this case the CoreAddonController would automatically add itself as a finalizer to any AddonInstallation it creates. Then whenever an AddonInstallation is being deleted, kubernetes will wait with the deletion until the finalizer is removed. Afterwards CoreAddonController is going to run a couple of checks internally:
+This option would keep the cluster-object in its current size. However we are currently not sure if there are any implications from creating the namespace early. We think this will have no implication for cluster templates, as they are based on the selected UI values, which would contain the Application information.
```
vgramer: In fact, it may have an impact because the cluster template is stored in `clustertemplate` and cluster are created from `clustertemplateinstances` CR.  So I have to check if we may encounter the same kind of problem regarding the size.

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
-  coreAddon: true
-  targetNamespace: openvpn
-  createNamespace: true
- addonRef:
-  name: openvpn
-  version: v1
+  targetNamespace: monitoring
+  createNamespace: true # flag to create the targetNamespace
+  applicationRef:
+    name: prometheus
```
vgramer:
```suggestion
    name: prometheus-node-exporter
```
to be coherent with the example of  `ApplicationDefinition` CR

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
 ```
 
-### AddonInstallationController
+### ApplicationInstallationController
 
-The AddonInstallationController resides in the seed cluster. Its main purpose is to watch AddonInstallation CRs and ensure that the requested addons are installed/modified/deleted in the corresponding user clusters. By extension of this, the AddonInstallationController needs to be able to fetch the manifests for addons.
+The ApplicationInstallationController resides in the seed cluster. Its main purpose is to watch ApplicationInstallation CRs and ensure that the requested Applications are installed/modified/deleted in the corresponding user clusters. By extension of this, the ApplicationInstallationController needs to be able to fetch the manifests for Applications.
 
-The functionality of the AddonInstallationController depends on the selected `source` and `method` in the AddonInstallation CR.
+The functionality of the ApplicationInstallationController depends on the selected `source` and `method` in the ApplicationInstallation CR. To start we propose the following sources and methods:
```
vgramer:
```suggestion
The functionality of the ApplicationInstallationController depends on the selected `source` and `method` in the ApplicationDefinition CR. To start we propose the following sources and methods:
```

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,184 @@ See below for a high level architecture view. Each component will be explained i
...
 
 **Sources**
```
vgramer: i would move this part, next to ApplicationDefinition CR

docs/proposals/improve-addons.md
SimonTheLeg: updated section with a new idea @vgramer and I developed

docs/proposals/improve-addons.md
SimonTheLeg: I think this would break this paragraph a bit. Because in the bullets we often describe what the ApplicationInstallationController currently is doing for the given source.

vgramer: fair enough

SimonTheLeg:
/hold

Add hold label, so this PR does not get auto-merged after the first person has approved it.

rastislavs:
/lgtm

Minor comment on the diagram "Cluster Administrator Creates A new Cluster with Addons", Addons / Addon Controller were not renamed, not sure if intentionally (as TBD is present in the title) or not.

SimonTheLeg:
@rastislavs ah good point! I have updated the diagram.

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
+  - the cluster where to install is specified by the namespace the ApplicationInstallation was created in (`metadata.namespace`) => the clusterID must be passed to k8c-api
 - contains the merged values that are being passed for the installation. This is needed as customers can set custom values for each installation.
 
 ```yaml
 apiVersion: kubermatic.k8c.io/v1
-kind: AddonInstallation
+kind: ApplicationInstallation
 metadata:
   name: prometheus
     namespace: cluster-47s2ddlfgj
```
kron4eg:
```suggestion
  namespace: cluster-47s2ddlfgj
```

kron4eg (commented):
/approve

micro nit in the comment

kron4eg:
I'm playing in the mind with the idea of custom resources definitions presented here would consist a dedicated group (for example, apps.kubermatic.k8c.io or any other).

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
+    - k8s-version → semVer range that describes compatible k8s versions
+  - values → These are values that can be used for overwriting defaults. We think this field will be required for Application if a customer is using a private registry. So application can be configured
 
-Note: The current implementation of go-templating allows to declare a `RequiredResourceTypes` field, to handle dependencies. We propose to keep this as is for go-templating, but not have it for helm. The reason is that helm offers its own dependency management and we think that building an additional dependency mechanism on top will drastically increase complexity. As a result, this field will be moved into under go-template method and not on the general addon level.
-
-An example CR could like this:
+An example CR could look like this:
 
 ```yaml
 apiVersion: kubermatic.k8c.io/v1
```
vgramer:
```suggestion
apiVersion: apps.kubermatic.k8c.io/v1
```

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
-- tells if an add-on is a CoreAddon or not → We propose to have this per installation, as different user clusters can have different CoreAddons. Furthermore, it should be noted that the AddonController will always set the `coreAddon` field to false, while the CoreAddonController will always set it to true
+### ApplicationInstallation CR
+
+- contains a mapping of a specific Application and cluster where to install
+  - the Application is specified by an ApplicationRef. Currently, this consists of a name and version
+  - the cluster where to install is specified by the namespace the ApplicationInstallation was created in (`metadata.namespace`) => the clusterID must be passed to k8c-api
 - contains the merged values that are being passed for the installation. This is needed as customers can set custom values for each installation.
 
 ```yaml
 apiVersion: kubermatic.k8c.io/v1
```
vgramer:
```suggestion
apiVersion: apps.kubermatic.k8c.io/v1
```

vgramer:
> I'm playing in the mind with the idea of custom resources definitions presented here would consist a dedicated group (for example, apps.kubermatic.k8c.io or any other).

make sense to me.

SimonTheLeg:
@vgramer is faster than the wind. I was about to commit the `apps` addition :D

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
 
-- watching for changes of an AddonCatalogue Custom Resource directly inside KKP master
-- watching a git repository → this idea was developed from PS feedback. It would enable customers to manage their addon catalogues in a GitOps way
+Furthermore, the ApplicationDistributor ensures that there are no duplicate Applications. A duplicate is defined by two Applications having the same `name` and `version`. We propose to protect against accidental duplicates using a ValidationWebhook.
+Lastly, we envisioned two possible sources for a ApplicationDefinitions:
 
-  *tbd:  Alternatively to achieve GitOps, it would also be possible to have a CI/CD pipeline directly update the AddonCatalogue CR in the master cluster. We are currently unsure if this is better or worse*
+- watching for changes of ApplicationDefinition Custom Resources directly inside KKP master
+- watching a git repository → this idea was developed from PS feedback. It would enable customers to manage their applications in a GitOps way
```
mfranczy: I would kill this idea, we are not building CI/CD, that should be handled by customers.

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
 
-- selecting compatible addons from the AddonCatalogue for each user cluster
-- managing AddonInstallations for each user cluster. It is important to note that it only tracks which addon should be installed in which cluster, but does not do the installation of the addon itself (this task will be done by the AddonInstallation controller).
-- providing a query-endpoint/function that takes in a clusterConstraints and returns a list of all available addons. We need this functionality to later on make it possible for users to select addons in the cluster wizard (see #6000). We think it makes sense to include this already in the proposal, so we build an architecture that will work for future requirements
+Alternatives considered: We considered the idea of creating CRs directly on the POST call in the api. This theoretically would have the advantage that you would not need to store all AddonRefs and their values in the cluster object. We decided to not go with this for the following reasons:
```
mfranczy: I think adding a new endpoint for creating application CR is better. You would not have to refactor the `CreateEndpoint`. I think UI should handle requests for creating clusters and later requests for applications when the cluster is ready. It could be done over some mechanims like promises. However, we have to ask about that somebody from the UI team. @maciaszczykm what do you think?

EDIT: although giving this a second thought, it might not be a bad idea to extend cluster. I don't have very strong opinion here.

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
+### ApplicationInstallationController
 
-1. Ask KKP api if the cluster was marked for deletion → if yes, remove finalizer and internal addon list; We need this check to ensure that CoreAddonController is not blocking the deletion of a cluster
-2. Check if the CoreAddon is still in internal list
-   - yes →keep CR as is to block deletion
-   - no →remove finalizer; CR will be deleted by k8s. This allows ClusterAdmins to remove CoreAddons if they really want to by removing them from CoreAddonControllers internal list, but still protects against accidental deletion
+The ApplicationInstallationController resides in the seed cluster. Its main purpose is to watch ApplicationInstallation CRs and ensure that the requested Applications are installed/modified/deleted in the corresponding user clusters. By extension of this, the ApplicationInstallationController needs to be able to fetch the manifests for Applications.
```
mfranczy: As I understand the installation controller would be deployed on the seed in the kubermatic namespace and watch all user clusters. I think we should change that. Instead of having relation one to many in that case it's better to have relation one to one.

It could be that we deploy application installation controller in the cluster namespace and it deploys ApplicationInstallation CR in the user cluster (similar like Machine Controller works and creates machines objects on the user cluster).

@moadqassem @kron4eg any opinion on this?

docs/proposals/improve-addons.md
mfranczy: btw please change `AddonsRef` to `ApplicationRef`

docs/proposals/improve-addons.md
```diff
@@ -80,230 +84,176 @@ See below for a high level architecture view. Each component will be explained i
...
+- contains a mapping of a specific Application and cluster where to install
+  - the Application is specified by an ApplicationRef. Currently, this consists of a name and version
+  - the cluster where to install is specified by the namespace the ApplicationInstallation was created in (`metadata.namespace`) => the clusterID must be passed to k8c-api
 - contains the merged values that are being passed for the installation. This is needed as customers can set custom values for each installation.
 
 ```yaml
-apiVersion: kubermatic.k8c.io/v1
-kind: AddonInstallation
+apiVersion: apps.kubermatic.k8c.io/v1
+kind: ApplicationInstallation
```
mfranczy: IMO should be deployed on the user cluster not on the seed.

docs/proposals/improve-addons.md
```diff
@@ -15,58 +15,62 @@
...
-- Generic Pull from object storage. It is definitely a feature that should be on roadmap for addons after the implementation of this proposal is done and should fit right in the current architecture. We have made this decision after consulting with PS. It was decided to go implement Git as an external source first, as this is going to reach the majority of customers
-- Reworking of the [dashboard into an addon](https://github.com/kubermatic/dashboard/issues/3666). In order to keep the size of this proposal manageable, we have decided to postpone this. It is definitely a feature that should be on roadmap for addons after the implementation of this proposal is done and should fit right in the current architecture
+- The intent is not to invent a whole new package manager. Therefore dependency management outside an Application is currently out of scope
```
mfranczy: I would change:
> Therefore dependency management outside an Application is currently out of scope
to
>  Therefore dependency management outside a supported source type is out of scope.

Or something similar to this, what I mean is that helm support dependencies. The sentence indicates that application has to bundle all dependencies (which is not true in case of helm).

docs/proposals/improve-addons.md
mfranczy: My reasoning here is that `ApplicationInstallation`  is user cluster specific those should be deployed there.

docs/proposals/improve-addons.md
kron4eg: It makes sense if we want to expose ApplicationInstallation CR to the end user, in the user-cluster API. A-la MachineDeployment. But the question is: is it the case?

mfranczy: I think it is, in my opinion this should be exposed to the end user (user cluster admin). @SimonTheLeg @vgramer what's your point of view on this?

mfranczy: and @shaase-ctrl ^

docs/proposals/improve-addons.md
maciaszczykm: For initial addons/applications (ones created with the cluster; from wizard) I'd say it would be the best to keep them as part of the cluster object (perhaps in the annotation like initial node deployment) and then create addons/application CR? For the existing clusters I'd use CRs only.

cc @kubermatic/sig-ui @kubermatic/sig-ui for more opinions on this

docs/proposals/improve-addons.md
shaase-ctrl: Sorry, I did not make it in time over the weekend. 
We need to have a user cluster admin be able to install apps for his cluster. KKP admins that need to configure everything per cluster is contrary to the self-service platform we built.

kron4eg: I don't think it's possible to expose it, as it's defined as

> contains a mapping of a specific Application and cluster where to install

vgramer: With the current design, KKP admin defined possible applications to install by creating an `ApplicationDefinition` CR in the master cluster. The `ApplicationDistributor` controller distributes these CRs to all seeds. 

the `ApplicationInstallation` CR  contains a reference to the `ApplicationDefinition` CR and the version of the application. this reference is used to know how to install the application.

if we move the `ApplicationInstallation` to the user-cluster and the `ApplicationInstallationController` into the cluster-namespace. That means we have to either:

1.  have a client to the seed cluster in the `ApplicationInstallationController` to be able to retrieve to the corresponding `ApplicationDefinition` CR. 
2. have a controller (ApplicationDistributor controller ?) that replicates `ApplicationDefinition` CR into each user cluster. So the `ApplicationInstallationController` can retrieve it.


Solution 1 seems pretty messy to me. especially if we consider deletion of an `ApplicationDefinition` CR  by a KKP admin. what happens to applications already installed.  

With solution 2, the deletion can be handled by an ownerRef set to `ApplicationInstalltion` CR inside the user-cluster.  When a KKP admin deletes an `ApplicationDefinition` CR, the deletion is eventually propagated to user-cluster and applications are removed.

SimonTheLeg: In addition to @vgramer comments, maybe a short background and two follow-up questions from my side

Background:
We had two initial ideas behind placing the ApplicationInstallationCRs inside the seed:
1. It would be tough for cluster admins to create invalid ApplicationInstallationCRs, as they can only use the UI to create them. This was born out of the idea that the majority of Application users would use the UI. Which sounds plausible to me, because if someone is more proficient in k8s and knows how to use helm, kustomized, etc., they would not really need the KKP Application feature. => now I think we can drop this issue, as we extend our ValidatingWebhook to ensure that no invalid ApplicationInstallationCRs can be created/edited (I think we just need to add that users cannot change the status field, otherwise Application state could get messy). @shaase-ctrl from a product-mgmt perspective: should the Cluster admin be able to edit ApplicationInstallations using the k8s api or would it be sufficient if they can do it via the UI?
2. Reduce the number of cross-cluster calls. We need to be aware of the fact that moving the ApplicationInstallationCRs into the user clusters will result in more cross-cluster calls. This is due to the fact that we need to keep ApplicationDefinitions in the seed cluster (otherwise cluster admins could just edit the definitions and make up their own Applications). This in turn means that for almost every interaction with the ApplicationInstallation we need to call the seed. From my previous experience I would always try to reduce cross cluster calls, but I am not sure how we feel about them at Kubermatic. Who would be the best person to weigh in on this?

docs/proposals/improve-addons.md
SimonTheLeg: @mfranczy changed to ApplicationRef

docs/proposals/improve-addons.md
SimonTheLeg: I have clarified the paragraph

docs/proposals/improve-addons.md
shaase-ctrl: Using the seeds for such features is the intention of how I would like to evolve KKP in general. We deliver additional features via the seeds in order to steer traffic, access etc. 

1. UI and API, it is as you reasoned, in the beginning more UI, later on more automation via API. For an MVP we could start with UI and go with API later (but it should be possible). The spread of the user base is quite large in the end.

2. Very good point for the cross cluster calls, I do not think that we optimized for this, but it is a good consideration. Overall this is a clash between control for KKP admin and flexibility for user cluster admin. My thoughts on this revolve around the possibility to use helm anyway in a user cluster for the user cluster admin, but helm could be delivered via the "applications" feature. Control and flexibility if you will. If I understood correctly what @vgramer  said, kkp admins could involuntarily delete apps that are in use, which could be quite unfortunate. In order to get to a result, I think we should go with the kkp admin approach you already have for "applications" and see if there is a huge demand for independently manage "applications" via the user cluster admin. Thanks @mfranczy @SimonTheLeg @vgramer and @kron4eg for your considerations.

mfranczy: @SimonTheLeg 

> It would be tough for cluster admins to create invalid ApplicationInstallationCRs

We should have a proper validation (on the client and server side), it should not matter if this is UI or API.

>  because if someone is more proficient in k8s and knows how to use helm, kustomized, etc., they would not really need the KKP Application feature. => now

Initially when I thought about this application mechanism I didn't mean to do it because some users don't know helm, it was for sake of automation and improved user experience (without forcing people to create their own automation on top to deploy standard apps they require - that's important with massive cluster creation)

> Reduce the number of cross-cluster calls. We need to be aware of the fact that moving the ApplicationInstallationCRs into the user clusters will result in more cross-cluster calls. This is due to the fact that we need to keep ApplicationDefinitions in the seed cluster (otherwise cluster admins could just edit the definitions and make up their own Applications).

Correct me if I am wrong, but IMO you won't reduce them here because the Application controller has to watch resources on the seed and user anyway. If `ApplicationInstallation` is deployed on the seed you have to check deployed resources on the user cluster to update the CR status. Or if `ApplicationInstallation` is on the user you have to get `ApplicationDefinition` CR.

@vgramer

> have a client to the seed cluster in the ApplicationInstallationController to be able to retrieve to the corresponding ApplicationDefinition CR.

Application controller has to be aware of seed and user clusters. Even with the current design while keeping the `ApplicationInstallation` CR on the seed you have to make a call to the user cluster to create a resource there over helm, k8s client or whatever source type you support.

Since the controller would run on the seed you have to get a client to user cluster (similar way how the current addon mechanism works).

If you move `ApplicationInstallation` CR to user you could have a custom informer to catch `ApplicationDefinition` events and map them to `ApplicationInstallation` CR events.


After all that's just my opinion, if you still feel different about that and have majority in this go ahead and continue with your idea.

SimonTheLeg:
@vgramer and I were thinking a bit about @mfranczy comments in regards to moving the ApplicationInstallationCRs into the user cluster. After a nice afternoon we made the following observations:

- @mfranczy is absolutely correct on the cross cluster calls :) Because the user-k8s-api still has to fulfil the requests (which requires calls between clusters, as user-k8s-api lives in the seed), we do not really not save cross-cluster-calls
- we cannot offload everything into the user-cluster, as the ApplicationInstallationController also needs to handle secrets to external ressources (like git credentials)

long story short: we made a small adjustment, which both moves the ApplicationInstallationCR into the user-cluster while keeping the secret handling in the seed. You can quickly see the diff by using this link [`b1c3fed` (#8276)](https://github.com/kubermatic/kubermatic/pull/8276/commits/b1c3fed747812f9abee907511373118a8ec85b86).

cc: @kron4eg @shaase-ctrl  who were also part of the discussion

kron4eg:
> @vgramer and I were thinking a bit about @mfranczy comments in regards to moving the ApplicationInstallationCRs into the user cluster. After a nice afternoon we made the following observations:
> 
>     * @mfranczy is absolutely correct on the cross cluster calls :) Because the user-k8s-api still has to fulfil the requests (which requires calls between clusters, as user-k8s-api lives in the seed), we do not really not save cross-cluster-calls
> 
>     * we cannot offload everything into the user-cluster, as the ApplicationInstallationController also needs to handle secrets to external ressources (like git credentials)
> 
> 
> long story short: we made a small adjustment, which both moves the ApplicationInstallationCR into the user-cluster while keeping the secret handling in the seed. You can quickly see the diff by using this link [`b1c3fed` (#8276)](https://github.com/kubermatic/kubermatic/pull/8276/commits/b1c3fed747812f9abee907511373118a8ec85b86).
> 
> cc: @kron4eg @shaase-ctrl who were also part of the discussion

:+1:

mfranczy (approved):
thanks for adjusting
/lgtm
/approve

SimonTheLeg:
all pendings resolved. files have been renamed. 

@vgramer handing over to you for the final approve to merge this.

Thanks everyone for your work! 🥇

PR #8282, opened 2021-11-24
network policies for user cluster kube-system namespace

mate4st:

Introduce network policies for user cluster in kube-system namespace

Fixes https://github.com/kubermatic/edgematic/issues/63

Done with: https://networkpolicy.io/

Feature can be disabled via feature flag **kubeSystemNetworkPolicies** at cluster spec. It is automatically added to new clusters, existing cluster won't be affected.

Attention: 
In case any custom applications are deployed to the user clusters kube-system namespace that need specific network access, a network policy needs to be added for that application to allow traffic other than dns.

```release-note
Deploy network policies with default deny to kube-system namespace in user cluster. Adding network policies for custom deployments in kube-system namespace maybe required.
```

### DIFF
--- a/addons/csi/hetzner/hcloud-csi-1.6.0.yaml
+++ b/addons/csi/hetzner/hcloud-csi-1.6.0.yaml
@@ -373,6 +373,8 @@ spec:
     - port: 9189
       name: metrics
       targetPort: metrics
+
+{{ if .Cluster.Features.Has "kubeSystemNetworkPolicies" }}
 ---
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
@@ -389,6 +391,7 @@ spec:
     - to:
         - ipBlock:
             cidr: 0.0.0.0/0
+{{ end }}
 {{ end }}
 {{ end }}
 {{ end }}
--- a/addons/csi/hetzner/hcloud-csi-1.6.0.yaml
+++ b/addons/csi/hetzner/hcloud-csi-1.6.0.yaml
@@ -373,6 +373,22 @@ spec:
     - port: 9189
       name: metrics
       targetPort: metrics
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: hcloud-csi
+  namespace: kube-system
+spec:
+  podSelector:
+    matchLabels:
+      app: hcloud-csi
+  policyTypes:
+    - Egress
+  egress:
+    - to:
+        - ipBlock:
+            cidr: 0.0.0.0/0
 {{ end }}
 {{ end }}
 {{ end }}
--- a/addons/openvpn/networkpolicy.yaml
+++ b/addons/openvpn/networkpolicy.yaml
@@ -1,0 +1,29 @@
+# Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: openvpn-client
+  namespace: kube-system
+spec:
+  podSelector:
+    matchLabels:
+      role: openvpn-client
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress: []
+  egress:
+    - {}
--- a/addons/openvpn/networkpolicy.yaml
+++ b/addons/openvpn/networkpolicy.yaml
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+{{ if .Cluster.Features.Has "kubeSystemNetworkPolicies" }}
+
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
@@ -27,3 +29,5 @@ spec:
   ingress: []
   egress:
     - {}
+
+{{ end }}
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -932,15 +932,15 @@ type reconcileData struct {
 	userSSHKeys      map[string][]byte
 	cloudConfig      []byte
 	// csiCloudConfig is currently used only by vSphere, whose needs it to properly configure the external CSI driver
-	csiCloudConfig         []byte
-	ccmMigration           bool
-	monitoringRequirements *corev1.ResourceRequirements
-	loggingRequirements    *corev1.ResourceRequirements
+	csiCloudConfig              []byte
+	ccmMigration                bool
+	monitoringRequirements      *corev1.ResourceRequirements
+	loggingRequirements         *corev1.ResourceRequirements
 	gatekeeperCtrlRequirements  *corev1.ResourceRequirements
 	gatekeeperAuditRequirements *corev1.ResourceRequirements
-	monitoringReplicas     *int32
-	clusterAddress         *kubermaticv1.ClusterAddress
-	k8sServiceApiIP        *net.IP
+	monitoringReplicas          *int32
+	clusterAddress              *kubermaticv1.ClusterAddress
+	k8sServiceApiIP             *net.IP
 }
 
 func (r *reconciler) ensureOPAIntegrationIsRemoved(ctx context.Context) error {
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -887,6 +887,11 @@ func (r *reconciler) reconcileNetworkPolicies(ctx context.Context, data reconcil
 		namedNetworkPolicyCreatorGetters = append(namedNetworkPolicyCreatorGetters, usersshkeys.NetworkPolicyCreator())
 	}
 
+	if r.isKonnectivityEnabled {
+		namedNetworkPolicyCreatorGetters = append(namedNetworkPolicyCreatorGetters, usersshkeys.NetworkPolicyCreator())
+
+	}
+
 	if err := reconciling.ReconcileNetworkPolicies(ctx, namedNetworkPolicyCreatorGetters, metav1.NamespaceSystem, r.Client); err != nil {
 		return fmt.Errorf("failed to ensure Network Policies: %v", err)
 	}
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -876,7 +876,7 @@ func (r *reconciler) reconcileDeployments(ctx context.Context, data reconcileDat
 func (r *reconciler) reconcileNetworkPolicies(ctx context.Context, data reconcileData) error {
 
 	namedNetworkPolicyCreatorGetters := []reconciling.NamedNetworkPolicyCreatorGetter{
-		// coredns.AllowAllDnsNetworkPolicyCreator(),
+		coredns.AllowAllDnsNetworkPolicyCreator(),
 		coredns.KubeDNSNetworkPolicyCreator(data.clusterAddress.IP, int(data.clusterAddress.Port)),
 	}
 
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -81,6 +81,10 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 			return fmt.Errorf("failed to get cloudConfig: %v", err)
 		}
 	}
+	clusterAddress, err := r.clusterAddress(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster address: %v", err)
+	}
 
 	data := reconcileData{
 		caCert:         caCert,
@@ -88,6 +92,7 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 		cloudConfig:    cloudConfig,
 		csiCloudConfig: CSICloudConfig,
 		ccmMigration:   r.ccmMigration || r.ccmMigrationCompleted,
+		clusterAddress: clusterAddress,
 	}
 
 	if !r.isKonnectivityEnabled {
@@ -872,7 +877,7 @@ func (r *reconciler) reconcileNetworkPolicies(ctx context.Context, data reconcil
 
 	namedNetworkPolicyCreatorGetters := []reconciling.NamedNetworkPolicyCreatorGetter{
 		// coredns.AllowAllDnsNetworkPolicyCreator(),
-		coredns.KubeDNSNetworkPolicyCreator(),
+		coredns.KubeDNSNetworkPolicyCreator(data.clusterAddress.IP, int(data.clusterAddress.Port)),
 	}
 
 	if r.userSSHKeyAgent {
@@ -918,13 +923,14 @@ type reconcileData struct {
 	userSSHKeys      map[string][]byte
 	cloudConfig      []byte
 	// csiCloudConfig is currently used only by vSphere, whose needs it to properly configure the external CSI driver
-	csiCloudConfig              []byte
-	ccmMigration                bool
-	monitoringRequirements      *corev1.ResourceRequirements
-	loggingRequirements         *corev1.ResourceRequirements
+	csiCloudConfig         []byte
+	ccmMigration           bool
+	monitoringRequirements *corev1.ResourceRequirements
+	loggingRequirements    *corev1.ResourceRequirements
 	gatekeeperCtrlRequirements  *corev1.ResourceRequirements
 	gatekeeperAuditRequirements *corev1.ResourceRequirements
-	monitoringReplicas          *int32
+	monitoringReplicas     *int32
+	clusterAddress         *kubermaticv1.ClusterAddress
 }
 
 func (r *reconciler) ensureOPAIntegrationIsRemoved(ctx context.Context) error {
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -890,7 +890,7 @@ func (r *reconciler) reconcileNetworkPolicies(ctx context.Context, data reconcil
 	}
 
 	if r.isKonnectivityEnabled {
-		namedNetworkPolicyCreatorGetters = append(namedNetworkPolicyCreatorGetters, metricsserver.NetworkPolicyCreator())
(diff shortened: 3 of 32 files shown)


pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// AllowAllDnsNetworkPolicyCreator allows egress traffic from all pods to kube-dns
+func AllowAllDnsNetworkPolicyCreator() reconciling.NamedNetworkPolicyCreatorGetter {
+	return func() (string, reconciling.NetworkPolicyCreator) {
+		return "allow-dns", func(np *networkingv1.NetworkPolicy) (*networkingv1.NetworkPolicy, error) {
+			dnsPort := intstr.FromInt(53)
```
PratikDeoghare: nit: Could be just `fmt.Sprint`.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
```diff
@@ -0,0 +1,191 @@
...
+							},
+							{
+								Protocol: &protoUdp,
+								Port:     &dnsPort,
+							},
+						},
+						To: []networkingv1.NetworkPolicyPeer{
+							{
+								IPBlock: &networkingv1.IPBlock{
+									CIDR: "169.254.20.10/32",
```
PratikDeoghare: `fmt.Sprintf("%s/32", somepackge.NodeLocalDNSCacheAddress)`

addons/csi/hetzner/hcloud-csi-1.6.0.yaml
```diff
@@ -364,6 +364,22 @@ spec:
...
+metadata:
+  name: hcloud-csi
+  namespace: kube-system
+spec:
+  podSelector:
+    matchLabels:
+      app: hcloud-csi
+  policyTypes:
+    - Egress
+  egress:
```
PratikDeoghare, line 390: Isn't this same as `egress: {}`?

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
```diff
@@ -0,0 +1,191 @@
...
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// AllowAllDnsNetworkPolicyCreator allows egress traffic from all pods to kube-dns
+func AllowAllDnsNetworkPolicyCreator() reconciling.NamedNetworkPolicyCreatorGetter {
```
PratikDeoghare: nit: `Dns -> DNS`

rastislavs:
A general question to this PR: are we really confident enough that this won't break anything to roll this out unconditionally to all existing user clusters of all existing KKP installations which upgrade to the new KKP version?

I would choose a bit safer approach, similar to the one chosen for apiserver network policies:

- Have a cluster API option (e.g. in `ClusterNetworkingConfig`) / cluster feature gate (like this one for [apiserver policies](https://github.com/kubermatic/kubermatic/blob/18c98aefc7ad2c9713e3bb34cd62f7e61d17ae5a/pkg/crd/kubermatic/v1/cluster.go#L231-L233)) for enabling these network policies,
- If these policies have to be enabled by default for each user cluster, then do it only for newly created clusters, e.g. in the mutation webhook like for [apiserver network policies](https://github.com/kubermatic/kubermatic/blob/18c98aefc7ad2c9713e3bb34cd62f7e61d17ae5a/pkg/webhook/cluster/mutation/mutation.go#L165-L171),
- In any case, it would be a good idea to have an option to disable these network polices - for special cases, if something breaks in production environment or for debugging purposes.

addons/csi/hetzner/hcloud-csi-1.6.0.yaml
rastislavs: same question :) If they are same, I would prefer to use the same expression everywhere, ideally the one which will be compatible with IPv6 ;)

addons/openvpn/networkpolicy.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```
rastislavs, line 1: we would need a similar policy for konnectivity-agent as well

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
```diff
@@ -0,0 +1,191 @@
...
+	return func() (string, reconciling.NetworkPolicyCreator) {
+		return "allow-dns", func(np *networkingv1.NetworkPolicy) (*networkingv1.NetworkPolicy, error) {
+			dnsPort := intstr.FromInt(53)
+			protoUdp := v1.ProtocolUDP
+			protoTcp := v1.ProtocolTCP
+
+			// dns access to node local dns cache
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PolicyTypes: []networkingv1.PolicyType{
+					networkingv1.PolicyTypeIngress,
```
rastislavs: I am not getting why the ingress part of the policy is needed here?

WeirdMachine:
It will definitely break something, but for dev this makes a lot of sense to figure out where we need to add or fix policies.

This should be enabled by default and I think it is okay to do so. It is relatively easy to add your own network policy. 

I agree with a setting in the cluster spec to to enable or disable network policies.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
mate4st: Since the pod selector is empty it applies to all pods. Therefore it denies Ingress and Egress (Except of dns) to all Pods that do not have an explicit network policy applied. I will move this to another package as this is more a default policy for all pods than a dns specific one.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
```diff
@@ -0,0 +1,191 @@
...
+						},
+					},
+				},
+				// api access
+				Egress: []networkingv1.NetworkPolicyEgressRule{
+					{
+						To: []networkingv1.NetworkPolicyPeer{
+							{
+								IPBlock: &networkingv1.IPBlock{
+									CIDR: fmt.Sprintf("%s/32", k8sApiIP),
```
rastislavs, line 116: I think we should include `kubernetes` service `ClusterIP` (`10.240.16.1` with default settings) here as well, to have full compatibility with all CNIs:

https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

> Cluster ingress and egress mechanisms often require rewriting the source or destination IP of packets. In cases where this happens, it is not defined whether this happens before or after NetworkPolicy processing, and the behavior may be different for different combinations of network plugin, cloud provider, Service implementation, etc.

> For egress, this means that connections from pods to Service IPs that get rewritten to cluster-external IPs may or may not be subject to ipBlock-based policies.

addons/openvpn/networkpolicy.yaml
mate4st: Yes, I was not sure if I can test it already as there is ongoing development. What do you think?

pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/networkpolicy.go
```diff
@@ -0,0 +1,55 @@
...
+package usersshkeys
+
+import (
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// NetworkPolicyCreator NetworkPolicy allows egress access to user ssh key agent.
```
rastislavs: actually, it allows egress access **from** user ssh agent, right?

addons/csi/hetzner/hcloud-csi-1.6.0.yaml
mate4st: No, this does not allow egress traffic to the cluster. But let me double check. If I am wrong I will let you know

pkg/controller/user-cluster-controller-manager/resources/resources/usersshkeys/networkpolicy.go
mate4st: Yes let me rephrase it

addons/openvpn/networkpolicy.yaml
rastislavs: Some tweaks still needed, but it is already working in kubermatic dev environment.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
PratikDeoghare: I think this policy selects all the pods from all the namespaces and allows traffic only to DNS. 
Wouldn't that block traffic between the pods?

PratikDeoghare (commented):
This should probably be behind a feature flag so that we can turn it off in case something goes wrong.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
PratikDeoghare: https://kubernetes.io/docs/concepts/services-networking/network-policies/#isolated-and-non-isolated-pods 

> Pods become isolated by having a NetworkPolicy that selects them. Once there is any NetworkPolicy in a namespace selecting a particular pod, that pod will reject any connections that are not allowed by any NetworkPolicy. (Other pods in the namespace that are not selected by any NetworkPolicy will continue to accept all traffic.)

This is what it looks like on my cluster. 

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  creationTimestamp: "2021-12-02T19:01:39Z"
  generation: 1
  name: allow-dns
  namespace: kube-system
  resourceVersion: "128709"
  uid: e009eb81-5f2a-4524-a576-5e1b7423d162
spec:
  egress:
    - ports:
        - port: 53
          protocol: TCP
        - port: 53
          protocol: UDP
      to:
        - podSelector:
            matchLabels:
              app.kubernetes.io/name: kube-dns
    - ports:
        - port: 53
          protocol: TCP
        - port: 53
          protocol: UDP
      to:
        - ipBlock:
            cidr: 169.254.20.10/32
  podSelector: {}
  policyTypes:
    - Ingress
    - Egress
```

PratikDeoghare:
@WeirdMachine 

Headsup: 

I have a cluster with 2 nodes. These pods for  `user-ssh-keys-agent, openvpn-client, envoy-agent`  are not coming up
for one of the nodes when I use the current branch. 

It could just be my hacks though.

WeirdMachine:
> @WeirdMachine 
> 
> Headsup: 
> 
> I have a cluster with 2 nodes. These pods for  `user-ssh-keys-agent, openvpn-client, envoy-agent`  are not coming up
> for one of the nodes when I use the current branch. 
> 
> It could just be my hacks though. 
> 
> 

Do they throw any error?

PratikDeoghare:
> 
Yes. Yes. OpenVPN health check fails. The other to fail to pull the image.

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
mate4st: Can this setting be changed? I did not found any usage of the ip address in the code, so i am not sure if i should start to hardcode it

rastislavs: It is the first address from the `cluster.Spec.ClusterNetwork.Services.CIDRBlocks[0]` range, which can be changed at cluster creation time. It appears as ClusterIP output of `kubectl get svc kubernetes` in the user cluster.

embik:
A general question: Are we not scraping any of the locked down components via Prometheus for metrics? That would need to be in the ingress rules I think, so I'm curious.

addons/csi/hetzner/hcloud-csi-1.6.0.yaml
PratikDeoghare: > No, this does not allow egress traffic to the cluster

I don't follow. Which cluster are we talking about?

rastislavs: So do we already know what is the difference between:
```
egress:
    - {}
```
and 
```
  egress:
    - to:
        - ipBlock:
            cidr: 0.0.0.0/0
```
?

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -857,6 +870,29 @@ func (r *reconciler) reconcileDeployments(ctx context.Context, data reconcileDat
...
+		kubesystem.DefaultNetworkPolicyCreator(),
+		coredns.KubeDNSNetworkPolicyCreator(data.clusterAddress.IP, int(data.clusterAddress.Port), data.k8sServiceApiIP.String()),
+	}
+
+	if r.userSSHKeyAgent {
+		namedNetworkPolicyCreatorGetters = append(namedNetworkPolicyCreatorGetters,
+			usersshkeys.NetworkPolicyCreator(data.clusterAddress.IP, int(data.clusterAddress.Port), data.k8sServiceApiIP.String()))
+	}
+
+	if r.isKonnectivityEnabled {
```
rastislavs, line 892: haven't you forgotten about `konnectivity.NetworkPolicyCreator` here?

PratikDeoghare: @WeirdMachine ^

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/networkpolicy.go
```diff
@@ -0,0 +1,46 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+)
+
+// NetworkPolicyCreator NetworkPolicy allows all egress traffic
+func NetworkPolicyCreator() reconciling.NamedNetworkPolicyCreatorGetter {
+	return func() (string, reconciling.NetworkPolicyCreator) {
+		return "konnectivity", func(np *networkingv1.NetworkPolicy) (*networkingv1.NetworkPolicy, error) {
```
PratikDeoghare, line 30: nit: a better name for this? maybe `konnectivity-allow-all`

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/networkpolicy.go
```diff
@@ -0,0 +1,46 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package konnectivity
+
+import (
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
PratikDeoghare: isn't this the other way round? :thinking: i.e. `k8s` imports are below `k8c` imports.

pkg/controller/user-cluster-controller-manager/resources/resources/node-local-dns/deamonset.go
```diff
@@ -30,6 +30,8 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+const CacheAddress = "169.254.20.10"
```
PratikDeoghare: This is called `NodeLocalDNSCacheAddress = "169.254.20.10"` in  pkg/resources/machinecontroller/deployment.go
Let us use the same name here too.

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/networkpolicy.go
mate4st: it does not allow all

pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/networkpolicy.go
mate4st: Yes it block traffic between the pods, thats the point of the network policy. The name was changed however to clarify that this is the default policy

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/networkpolicy.go
mate4st: It is blocking ingress and allow all egress

addons/csi/hetzner/hcloud-csi-1.6.0.yaml
mate4st: It is the same, unless you want to define ports, the networkpolicy editor is a bit confusing about it. What may makes sense here is to use cidr: 0.0.0.0/0 and use `except` and deny the cluster cidr. That way all egress to world would be allowed, but cluster access denied.

pkg/controller/user-cluster-controller-manager/resources/resources/node-local-dns/deamonset.go
```diff
@@ -92,7 +91,7 @@ func DaemonSetCreator(registryWithOverwrite registry.WithOverwriteFunc) reconcil
 					ImagePullPolicy: corev1.PullAlways,
 					Args: []string{
 						"-localip",
-						CacheAddress,
+						machinecontroller.NodeLocalDNSCacheAddress,
```
PratikDeoghare: I meant rename `CacheAddress` to `NodeLocalDNSCacheAddress`. 

Importing `"k8c.io/kubermatic/v2/pkg/resources/machinecontroller"` introduces dependency on the the machine controller.

pkg/controller/user-cluster-controller-manager/resources/resources/konnectivity/networkpolicy.go
PratikDeoghare: how about `konnectivity-allow-egress` then? :-)

kron4eg:
@WeirdMachine in release note it said "[ACTION REQUIRED]" (but missing this from the PR title), what action is required?

WeirdMachine:
> @WeirdMachine in release note it said "[ACTION REQUIRED]" (but missing this from the PR title), what action is required?

Adding network policies for custom deployments in kube-system namespace maybe required.

kron4eg:
> > @WeirdMachine in release note it said "[ACTION REQUIRED]" (but missing this from the PR title), what action is required?
> 
> Adding network policies for custom deployments in kube-system namespace maybe required.

Where to read about that, what policies exactly should be added?

WeirdMachine:
I thought the release note is actually enough information. Anyway I added some more information to the description.

WeirdMachine:
Updated description and removed action required as there is no effect on existing clusters.

PR #8283, opened 2021-11-24
added support for multiple destinations for etcd backups controller

### LINKED ISSUE #7939: Support EtcdBackupConfig Destinations in backup controllers
**User Story**
To support multiple backup destinations per Seed, the controller needs to support the new Destination information in the EtcdBackupConfigs, instead of using the default Seed one.

lsviben:
added support for multiple destinations for etcd backups controller.

The backup, and backup cleanup/delete job containers are defined in:
https://github.com/kubermatic/kubermatic/tree/master/charts/kubermatic/static

They use hardcoded env Vars for the secret and configmap(both are one per Seed) from which to take the credentials, and bucket info. 
As we want to introduce multiple destinations, this doesnt work for us as we support multiple pairs of credentials and bucket details.

So the approach is that if the destination is set, it will change the env vars of the backup job for the credentials and backup bucket name and endpoint. 

Fixes #7939 

```release-note
Etcd backups now support multiple destinations, which can be configured per Seed. If a destination is set for an EtcdBackupConfig, it will be used instead of the legacy `backup-s3` credentials secret and `s3-settings` backup bucket details. 
```

### DIFF
--- a/cmd/seed-controller-manager/controllers.go
+++ b/cmd/seed-controller-manager/controllers.go
@@ -247,6 +247,7 @@ func createEtcdBackupController(ctrlCtx *controllerContext) error {
 		ctrlCtx.runOptions.backupContainerImage,
 		ctrlCtx.versions,
 		ctrlCtx.runOptions.caBundle,
+		ctrlCtx.seedGetter,
 	)
 }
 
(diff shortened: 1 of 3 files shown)


moadqassem (approved):
/approve
/lgtm

PR #8288, opened 2021-11-24
fixes the user-deletion issue

### LINKED ISSUE #8287: Kubermatic users not getting deleted
**User Story**
users.kubermatic.k8s.io on the master cluster was not getting deleted, even after seed-users-cleanup, because of the reason the Update finalizer event was not sent to master-controller because of problematic controller event-filtering 

**Acceptance criteria**
users.kubermatic.k8s.io should get deleted after all seed's users deleted

imharshita:

Fixes #8287

```release-note
none
```

### DIFF
--- a/pkg/controller/master-controller-manager/user-synchronizer/controller.go
+++ b/pkg/controller/master-controller-manager/user-synchronizer/controller.go
@@ -19,7 +19,6 @@ package usersynchronizer
 import (
 	"context"
 	"fmt"
-	"reflect"
 
 	"go.uber.org/zap"
 
@@ -98,15 +97,7 @@ func withEventFilter() predicate.Predicate {
 			return true
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			oldUser, ok := e.ObjectOld.(*kubermaticv1.User)
-			if !ok {
-				return false
-			}
-			newUser, ok := e.ObjectNew.(*kubermaticv1.User)
-			if !ok {
-				return false
-			}
-			return !reflect.DeepEqual(oldUser.Spec, newUser.Spec)
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
 			return true


lsviben:
/lgtm

this eventPredicate was erroneously added in https://github.com/kubermatic/kubermatic/pull/7948, as one of the tries to fix the flaky test. It is unnecessary. 

@jiachengxu could you take a look as well

jiachengxu (approved):
/approve
/lgtm

pkg/controller/master-controller-manager/user-synchronizer/controller.go
```diff
@@ -84,39 +82,14 @@ func Add(
...
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			oldUser, ok := e.ObjectOld.(*kubermaticv1.User)
-			if !ok {
-				return false
-			}
-			newUser, ok := e.ObjectNew.(*kubermaticv1.User)
-			if !ok {
-				return false
-			}
-			return !reflect.DeepEqual(oldUser.Spec, newUser.Spec)
```
mfranczy, line 109: I am having a problem to understand how removing update predicate fixes the issue. Correct me if I am wrong, but if cluster admin creates a new user this user is being synchronized to all seeds, correct? So the communication is from master to seed for synchronization. If cluster admins delete a user then we should get a delete event from delete function. So how the update function was problematic?

pkg/controller/master-controller-manager/user-synchronizer/controller.go
```diff
@@ -84,39 +82,14 @@ func Add(
...
 
 	return nil
 }
 
-func withEventFilter() predicate.Predicate {
-	return predicate.Funcs{
-		CreateFunc: func(e event.CreateEvent) bool {
-			return true
-		},
-		UpdateFunc: func(e event.UpdateEvent) bool {
```
mfranczy, line 100: what about changing `UpdateFunc` to:
```
return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
this way we won't get events about `status` subresource update. 

I could imagine that we just compare `spec` without checking `metadata` and it's causing an issue, but still have a problem to understand how it happens like i described [here](https://github.com/kubermatic/kubermatic/pull/8288/files#r756157812)

mfranczy: ok, now I understand the update with `deletionTimestamp` as @lsviben pointed

pkg/controller/master-controller-manager/user-synchronizer/controller.go
lsviben: The issue is fixed because the deleting a resource with a finalizer actually does an update with deletionTimestamp, we ignore that in the event filter because we just check spec, so it doesnt get handled.

Good idea with comparing the generations, that should work. @imharshita could you change it?

lsviben:
/hold

we found an alternative solution

pkg/controller/master-controller-manager/user-synchronizer/controller.go
mfranczy: so @imharshita we should change the update func to 
```
return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
```
to filter out updates to status field. Please test that again with the new change.

PR #8294, opened 2021-11-25
Fix escaping of dots in user cluster dashboard

dakraus:
The dots in the PromQL query of the panels 'Minor Releases' and 'Minor Release History' in the user cluster dashboard were not escaped with double backslashes and therefore no data could be displayed.

```release-note
NONE
```


xrstf (approved):
thanks!

/approve

PR #8297, opened 2021-11-25
Fix OpenStack cloud initialization for subnets that are already attached to router

mfranczy:
Fixes cloud initialization at OpenStack when users use their own subnets that are already attached to router.

```release-note
NONE
```

### DIFF
--- a/pkg/provider/cloud/openstack/provider.go
+++ b/pkg/provider/cloud/openstack/provider.go
@@ -164,6 +164,28 @@ func validateExistingSubnetOverlap(networkID string, netClient *gophercloud.Serv
 	})
 }
 
+// We start by adding the finalizers, note that this is safe because the
+// clean-up is idempotent, if the cluster is deleted when resources associated
+// to the finalizer are not created yet, it does not fail.
+// The reason behind is that we have several controllers adding finalizers
+// to the Cluster resource at the moment, and to avoid race conditions we
+// need to use optimistic locking and return immediately in case of
+// conflicts to retry later.
+func ensureFinalizers(cluster *kubermaticv1.Cluster, finalizers []string, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
+	var err error
+
+	if len(finalizers) > 0 {
+		cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
+			kubernetes.AddFinalizer(cluster, finalizers...)
+		}, provider.UpdaterOptionOptimisticLock)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return cluster, nil
+}
+
 // InitializeCloudProvider initializes a cluster, in particular
 // creates security group and network configuration
 func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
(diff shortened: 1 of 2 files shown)


pkg/provider/cloud/openstack/provider.go
```diff
@@ -164,6 +164,21 @@ func validateExistingSubnetOverlap(networkID string, netClient *gophercloud.Serv
 	})
 }
 
+func ensureFinalizers(cluster *kubermaticv1.Cluster, finalizers []string, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
```
mfranczy, line 174: I started doing some small refactoring for the `InitializeCloudProvider` func. Without that lint would complain about cyclomatic complexity because of another condition added
```
pkg/provider/cloud/openstack/provider.go:169:1: cyclomatic complexity 31 of func `(*Provider).InitializeCloudProvider` is high (> 30) (gocyclo)
```

pkg/provider/cloud/openstack/provider.go
```diff
@@ -164,6 +164,21 @@ func validateExistingSubnetOverlap(networkID string, netClient *gophercloud.Serv
...
+	var err error
+
+	if len(finalizers) > 0 {
+		cluster, err = update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
+			kubernetes.AddFinalizer(cluster, finalizers...)
+		}, provider.UpdaterOptionOptimisticLock)
+		if err != nil {
+			return nil, fmt.Errorf("failed to add finalizers: %w", err)
+		}
```
xrstf: This can simply be `return update(.....)`. Otherwise, in case of an error, the full error message will be "failed to ensure finalizers: failed to add finalizers: ...", which is kinda redundant.

pkg/provider/cloud/openstack/provider.go
```diff
@@ -205,13 +220,9 @@ func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
 	// to the Cluster resource at the moment, and to avoid race conditions we
 	// need to use optimistic locking and return immediately in case of
 	// conflicts to retry later.
```
xrstf: Please move this comment as well, as it doesn't make sense here anymore.

pkg/provider/cloud/openstack/provider.go
```diff
@@ -314,7 +325,8 @@ func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
 	// reconciliations so far. This is to avoid hitting OpenStack API at each
 	// iteration.
 	// TODO: this is terrible, find a better way.
-	if cluster.Status.ExtendedHealth.CloudProviderInfrastructure != kubermaticv1.HealthStatusUp {
+	if cluster.Status.ExtendedHealth.CloudProviderInfrastructure != kubermaticv1.HealthStatusUp &&
+		kubernetes.HasFinalizer(cluster, RouterSubnetLinkCleanupFinalizer) {
```
xrstf, line 329: Isn't this the wrong way around? Attaching the subnet if the finalizer is already set?

xrstf (commented):
commentatione

pkg/provider/cloud/openstack/provider.go
mfranczy: it cannot :P because then in the `InitializeCloudProvider` function i would have to restore the `if len(finalizers)...` condition which will cause (in that function there are too many conditions already): 

```pkg/provider/cloud/openstack/provider.go:169:1: cyclomatic complexity 31 of func `(*Provider).InitializeCloudProvider` is high (> 30) (gocyclo)```

with the error message I agree

mfranczy: @xrstf what do you think to add gocyclo:ignore directive to this function and do a proper refactoring later? But as always who will have time to do such a thing later....

pkg/provider/cloud/openstack/provider.go
mfranczy: it's not because finalizers are added ahead before the resource is created
https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/openstack/provider.go#L185-L199

pkg/provider/cloud/openstack/provider.go
mfranczy: I fixed the message by not wrapping the error in the ensureFinalizers func

xrstf (approved):
/approve

We're rewriting this provider hopefully soon.

PR #8305, opened 2021-11-26
Deploy metrics-server in user cluster if Konnectivity is enabled

### LINKED ISSUE #7113: Migrating Metrics Server to User Cluster in Konnectivity Setup
API extension servers are not supported in the control networks when using Konnectivity, this means that the `metrics-server` has to be moved on the user-cluster side as a pre-requisite.

Find more information here: https://github.com/kubernetes/kubernetes/issues/92374

rastislavs:
Deploy metrics-server in user cluster if Konnectivity is enabled. This is required in Konnectivity setup, because metrics-server connects directly to kubelet running on the worker nodes. That is not possible from seed after openvpn tunnel has been removed.

Fixes #7113 

**Special notes for your reviewer**: Konnectivity agent deployment is seed that was connecting apiserver with the metrics-server running in seed was removed as well.

**Documentation**:

```release-note
NONE
```


PratikDeoghare (approved):

moadqassem:
@rastislavs So the code is full of `if konnectivity isEnabled; then`. What would be the draw back of simply deploy metrics server regardless if konnectivity is enabled or not?

rastislavs:
> @rastislavs So the code is full of `if konnectivity isEnabled; then`. What would be the draw back of simply deploy metrics server regardless if konnectivity is enabled or not?

The idea was to not affect existing clusters running on OpenVPN anyhow.

One of the reasons is that we are not 100% sure (only 99%) that moving the metrics-server does not impact anything, and so that we still have a way to switch back if this is not working fine in some edge scenario.

Another reason is that we would need to add some extra migration code for clusters already running metrics-server in the seed, e.g. the `metrics-server` service in the user cluster needs to be first deleted and then re-created (as it change the service type).

OFC, as Konnecitvity option in KKP matures and we finally deprecate and remove all OpenVPN-related code completely, all those `if konnectivity isEnabled` conditionals will be cleaned up as well.

moadqassem (approved):
/approve
/lgtm

PR #8321, opened 2021-11-29
Support multi-arch builds for kubeletdnat-controller

### LINKED ISSUE #8253: envoy-agent not starting on ARM-only user cluster nodes
**User Story**
I tried setting up a user cluster on AWS with ARM worker nodes (CentOS). The seed is using the `Tunneling` expose strategy. Unfortunately, the components in the user cluster didn't come up healthy.

The problem seems to be that `envoy-agent` is crashing with the container `create-dummy-interface` erroring out:

```
standard_init_linux.go:228: exec user process caused: exec format error
```

The image for that container is `quay.io/kubermatic/kubeletdnat-controller` (https://github.com/kubermatic/kubermatic/tree/master/cmd/kubeletdnat-controller).

#6584 was supposed to support multi-arch images for all components, but it didn't list `envoy-agent`, so I'm not sure if that component was introduced at a later date or just didn't make the list.

Since envoy-agent sets up the connectivity to the Kubernetes API, Canal can't start, and no other Pods come up successfully.

**Acceptance criteria**

`envoy-agent` starts successfully on ARM nodes.

embik:
`kubeletdnat-controller` isn't built as a multi-arch image. Because of that, a user cluster with ARM-only nodes will fail to come up as `envoy-agent` cannot start. `kubeletdnat-controller` is the container image that breaks this Pod (https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/resources.go#L586). This likely only affects the `Tunneling` expose strategy and might have flown under the radar because of that (ARM is already supported, sort of).

I've adapted the multiarch build strategy for `user-ssh-keys-agent`.

Fixes #8253

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/kubeletdnat-controller/Dockerfile.multiarch
+++ b/cmd/kubeletdnat-controller/Dockerfile.multiarch
@@ -1,0 +1,36 @@
+# Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


xrstf (approved):
We need the arm64 gocache sooner than later.

/approve

PR #8323, opened 2021-11-29
Update lastSeen and user watcher handling

floreks:
- throttle `lastSeen` field update in the `user` CRD to once a minute
- add cache to the user watcher (websocket)
- ignore `lastSeen` field updates when notifying user settings subscribers about changes

After all, I haven't touched the logic that uniquely identified users by email as all the comments in the code indicated that the email address should be unique and it would require a bigger refactoring.

Related to the https://github.com/kubermatic/dashboard/issues/3900. UI will be slightly adjusted aswell.

```release-note
NONE
```


lsviben:
can we move the lastSeen to a Status subresource?

floreks:
@lsviben it's already part of the API in the last official release, so it might break the compatibility.

maciaszczykm:
IMO it's more of a status than a spec. Spec shouldn't be updated that often. However, this may be breaking change at the moment and this makes it more problematic.

lsviben:
Im not sure if somebody is using the lastSeen for the User in some meaningful way except us. 

For breaking the CRD compatibility, we dont need to remove the lastSeen field, just deprecate and stop filling it, put a warning about it in the release notes. If needed, we can still support both for 1 release, and then deprecate it forever.

For API we are not breaking it anyway because we just fill the lastSeen field from the status instead of the spec.

WDYT? I think we need to do this at one point anyway

floreks:
If I understand correctly, you want to create a new `Status` structure that will be inside `User` or `UserSpec` CRD, right?

lsviben:
> If I understand correctly, you want to create a new `Status` structure that will be inside `User` or `UserSpec` CRD, right?

Yes, similarly to what we currently have on our Cluster resource for example (inside User, out of the spec). This is also what K8s suggests in docs, to have resource state information(like lastSeen) in the Status subresource.

floreks:
@lsviben @maciaszczykm PTAL

maciaszczykm:
Looks good to me, I'll let @lsviben to do the final check.

pkg/crd/kubermatic/v1/user.go
```diff
@@ -52,7 +58,6 @@ type UserSpec struct {
 	IsAdmin                 bool                                    `json:"admin"`
 	Settings                *UserSettings                           `json:"settings,omitempty"`
 	TokenBlackListReference *providerconfig.GlobalSecretKeySelector `json:"tokenBlackListReference,omitempty"`
-	LastSeen                *metav1.Time                            `json:"lastSeen,omitempty"`
```
lsviben, line 55: not sure if we can just remove this as it will break compatibility. I think we need to deprecate it, and remove it in a future User version. 

Although this field is for internal use only and 99% nobody uses it, so maybe we can do a breaking change, but we would need to inform users in the release notes. 

BTW, as the sig-crd guys are working on CRD changes, we need to mirror this change in https://github.com/kubermatic/kubermatic/blob/master/pkg/apis/kubermatic/v1/user.go and update the migration in
https://github.com/kubermatic/kubermatic/blob/master/pkg/install/crdmigration/clones.go

Saying this, as we are anyway migrating the CRD, maybe the new one can have just the lastSeen in Status? @xrstf ?

floreks: Keep in mind that the API version of this field hasn't changed. Only CRD will be different. I am fine with keeping it for now and deprecating until next release.

lsviben: Right, I misspoke, I meant the CRD compatibility will change. Although I think the chances are miniscule that somebody is integrating with our User CRD AND using the lastSeen, you never know. Plus we support integration through the API, not CRDs.

I would also be for removing it so we can forget about it, but would like some input from our SIG-CRD guys, anyway they need to approve this. 

Other then that, this PR LGTM

xrstf: > Saying this, as we are anyway migrating the CRD, maybe the new one can have just the lastSeen in Status? @xrstf ?

Your change in this PR looks totally fine to me, in fact I like moving this field into the `status` subresource. I don't quite understand the question, though ^^

lsviben: Sorry for the badly formulated question. 

Moving the field from `spec` to `status` is good for sure. The only concern is if we should remove the field from `spec` totaly, or should we leave a deprecated one there?

xrstf: Remove it :) Even if there was no migration, at worst we would lose this tiny piece of non-critical information.

PR #8326, opened 2021-11-30
Seccomp profiles for user cluster pods

embik:
CIS benchmark recommends that Pods have the "default runtime" seccomp profile enabled. This PR adds a seccomp profile to all Pods created by `user-cluster-controller-manager` and adds a test to conformance testing to make sure new components we introduce in the future also set the seccomp profile. 

Fixes https://github.com/kubermatic/edgematic/issues/64

**Documentation**:

```release-note
Components running on user clusters are set up with the runtime's default seccomp profile
```

### DIFF
--- a/addons/aws-node-termination-handler/aws-node-termination-handler.yaml
+++ b/addons/aws-node-termination-handler/aws-node-termination-handler.yaml
@@ -202,6 +202,9 @@ spec:
                     operator: NotIn
                     values:
                       - fargate
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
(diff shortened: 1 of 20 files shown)


rastislavs:
Could you please also take care of:
 - `cilium-operator` in `addons/cilium` (deployed instead of Canal in case that Cilium CNI is chosen)
 - `controller/user-cluster-controller-manager/resources/resources/metrics-server/deployment.go` - just recently added, deployed only when Konenctivity is enabled

embik:
@rastislavs great catch, thanks! I missed that cilium deploys two workloads, and I wasn't aware of the metric-server change.

PR #8327, opened 2021-11-30
Making source IP addresses configurable

### LINKED ISSUE #8151: Make NodePort security group ranges configurable
**User Story**

As per [this comment on the NodePort epic](https://github.com/kubermatic/kubermatic/issues/6583#issuecomment-946818242), there is a need to configure the IP range that is allowed to access NodePorts (the current implementation allows from anywhere).

That probably needs a new field on the `Cluster` resource somewhere.

**Acceptance criteria**

A user can configure the IP range that is allowed to access NodePorts via security groups created by KKP.

mlavacca:

Fixes #8151

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubermatic/v1/cluster.go
+++ b/pkg/apis/kubermatic/v1/cluster.go
@@ -659,15 +659,15 @@ type AzureCloudSpec struct {
 	ClientID       string `json:"clientID,omitempty"`
 	ClientSecret   string `json:"clientSecret,omitempty"`
 
-	ResourceGroup         string `json:"resourceGroup"`
-	VNetResourceGroup     string `json:"vnetResourceGroup"`
-	VNetName              string `json:"vnet"`
-	SubnetName            string `json:"subnet"`
-	RouteTableName        string `json:"routeTable"`
-	SecurityGroup         string `json:"securityGroup"`
-	AllowedIPRange        string `json:"allowedIPRange,omitempty"`
-	AssignAvailabilitySet *bool  `json:"assignAvailabilitySet"`
-	AvailabilitySet       string `json:"availabilitySet"`
+	ResourceGroup           string `json:"resourceGroup"`
+	VNetResourceGroup       string `json:"vnetResourceGroup"`
+	VNetName                string `json:"vnet"`
+	SubnetName              string `json:"subnet"`
+	RouteTableName          string `json:"routeTable"`
+	SecurityGroup           string `json:"securityGroup"`
+	NodePortsAllowedIPRange string `json:"nodePortsAllowedIPRange,omitempty"`
+	AssignAvailabilitySet   *bool  `json:"assignAvailabilitySet"`
+	AvailabilitySet         string `json:"availabilitySet"`
 	// LoadBalancerSKU sets the LB type that will be used for the Azure cluster, possible values are "basic" and "standard", if empty, "basic" will be used
 	LoadBalancerSKU LBSKU `json:"loadBalancerSKU"` //nolint:tagliatelle
 }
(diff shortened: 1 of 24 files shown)


pkg/provider/cloud/azure/provider.go
```diff
@@ -646,8 +646,13 @@ func (a *Azure) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update pr
 			Build().
 			NodePorts()
 
+		allowedIPRange := cluster.Spec.Cloud.Azure.AllowedIPRange
+		if allowedIPRange == "" {
+			allowedIPRange = "*"
+		}
```
embik: Do we need to convert `0.0.0.0/0` to asterisk or is that a valid value too?

pkg/provider/cloud/gcp/provider.go
```diff
@@ -336,6 +336,11 @@ func (g *gcp) ensureFirewallRules(cluster *kubermaticv1.Cluster, update provider
 		Build().
 		NodePorts()
 
+	allowedIPRange := cluster.Spec.Cloud.GCP.AllowedIPRange
+	if allowedIPRange == "" {
+		allowedIPRange = "0:0:0:0/0"
```
embik: Shouldn't this be `0.0.0.0/0`?

pkg/provider/cloud/openstack/provider.go
```diff
@@ -246,10 +246,16 @@ func (os *Provider) InitializeCloudProvider(cluster *kubermaticv1.Cluster, updat
 			Build().
 			NodePorts()
 
+		allowedIPRange := cluster.Spec.Cloud.Openstack.AllowedIPRange
+		if allowedIPRange == "" {
+			allowedIPRange = "0:0:0:0/0"
```
embik: Same here, shouldn't this be `0.0.0.0/0`? Or is this some ipv6 notation I'm not aware of?

pkg/provider/cloud/azure/provider.go
embik: Since we do a CIDR parsing in the spec validation, `*` wouldn't be a valid value for the cluster spec. That feels weird that we're defaulting to something that's not allowed.

embik (changes requested):

pkg/provider/cloud/gcp/provider.go
mlavacca: absolutely, it's just a typo, going to fix it

pkg/provider/cloud/azure/provider.go
mlavacca: `0.0.0.0/0` is represented as `*` in Azure, I agree the defaulting has to be changed

embik: Yeah but since the field is labeled as "IP range", my question is: Should we convert `"0.0.0.0/0"` user input to `"*"` internally here? Or is that confusing as an Azure user would expect to be using `"*"`?

mlavacca: Actually, in the Azure doc, the statement is "Asterisk '*' can also be used to match all source IPs", therefore defaulting to `0.0.0.0/0` and using the same notation everywhere is the best solution IMO (let's forget about this wildcard, which is simply another way to say `0.0.0.0/0`). I would replace this `allowedIPRange = "*"` with `allowedIPRange = "0.0.0.0/0"` WDYT?

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-03-01/network#SecurityRulePropertiesFormat.SourceAddressPrefix

embik: That sounds reasonable.

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -665,6 +665,7 @@ type AzureCloudSpec struct {
 	SubnetName            string `json:"subnet"`
 	RouteTableName        string `json:"routeTable"`
 	SecurityGroup         string `json:"securityGroup"`
+	AllowedIPRange        string `json:"allowedIPRange,omitempty"`
```
embik: Sorry for not raising this during the initial review I did, but I think we should either a) rename this field to include a reference to this being related to node ports (e.g. `NodePortsAllowedIPRange`) or b) add a description to the API object to clarify what's it for (that happens through a comment afaik, no?). Perhaps both? There are many things in a cloud provider spec that "allowed IP ranges" could refer to.

```suggestion
        // NodePortsAllowedIPRange defines the IP range that is allowed to access node ports via firewall/security groups. Defaults to "0.0.0.0/0" (access allowed from any IP)
	NodePortsAllowedIPRange        string `json:"nodePortsAllowedIPRange,omitempty"`
```

Obviously applies to the other additions to provider cloud specs too.

embik (changes requested):

embik (approved):
/approve
/lgtm

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -659,14 +659,15 @@ type AzureCloudSpec struct {
...
-	SecurityGroup         string `json:"securityGroup"`
-	AssignAvailabilitySet *bool  `json:"assignAvailabilitySet"`
-	AvailabilitySet       string `json:"availabilitySet"`
+	ResourceGroup           string `json:"resourceGroup"`
+	VNetResourceGroup       string `json:"vnetResourceGroup"`
+	VNetName                string `json:"vnet"`
+	SubnetName              string `json:"subnet"`
+	RouteTableName          string `json:"routeTable"`
+	SecurityGroup           string `json:"securityGroup"`
+	NodePortsAllowedIPRange string `json:"nodePortsAllowedIPRange,omitempty"`
```
kron4eg, line 668: Migration (copying from old types to new) is missing. You can follow one of two options:
* drop new API types changes (including generated CRDs and chart)
* add missing field (here, and for other changed types from this PR) to the migration (see https://github.com/kubermatic/kubermatic/blob/master/pkg/install/crdmigration/clones.go#L452)

kron4eg (commented):
/approve

PR #8344, opened 2021-12-02
Default service account token mount

mate4st:

Sets automountServiceAccountToken to false for default service accounts in user cluster namespaces provided by KKP

Fixes #https://github.com/kubermatic/edgematic/issues/61

```release-note
[ACTION REQUIRED] Set the default Service Accounts automountServiceAccountToken property to false for all KKP provided namespaces and kube-system.
```


embik (approved):
/approve
/lgtm

PR #8346, opened 2021-12-02
Refactor Cluster Defaulting & Validation

xrstf:

This PR refactors and rethinks how we handle `Cluster` defaulting and validation.

### Previously

In the old codebase, both aspects were split into two set functions:

* half of the implementation was in the webhooks
* the other half was scattered throughout our API

This meant that if you create a cluster via the KKP API, different defaulting and more importantly, different validation logic applied, compared to if you do a `kubectl apply -f mycluster.yaml` (which is technically still not allowed/supported). Previously you could, using `kubectl`, create a totally broken, incomplete `Cluster` object (all you had to set was the ExposeStrategy and that's it). Those incomplete clusters were not defaulted at runtime, so they are really unusable.

With our intent to move to declarative working, it is vital that we have one consistent set of rules that apply to `Cluster` objects, no matter how they are created. This also prevents bugs in our codebase if people do not put their validation in random places anymore (or worse, in multiple).

### Future

This PR consolidates the existing code. For example, **all Cluster validation happens in `pkg/validation`**, just as **all defaulting happens in `pkg/defaulting`**. The webhooks and the API are simply calling the functions on those packages.

Defaulting for `Seed`s and `KubermaticConfiguration` lives in `pkg/controller/operator/defaults`, but should probably be moved to the packages mentioned above, for consistency. This can be done in a follow up PR however.

### Webhook Behaviour

It is vital to understand that Kubernetes will always call the mutating (defaulting) webhooks first and call the validating webhooks last. This means the validation can always assume that defaulting has already happened, i.e. `pkg/validation.ValidateClusterSpec()` will throw an error if no NodeportRange is set in the Cluster. The NodeportRange is optional to the user, but the mutating webhook will default it and the validating webhook will then ensure it.

TL:DR; First mutation, then validation.

### Defaulting

KKP offers 5 different sources of default values for a Cluster, in order of importance:

* the default `ClusterTemplate`, configured on the Seed object (importantly, the ClusterTemplate is **optional**)
* the `Seed` object's `spec.defaultComponentSettings`
* the `KubermaticConfiguration` object's `spec.userCluster`
* the chosen Cloud Provider can also default (though none of our providers actually do anything)
* Go constants in `pkg/controller/operator/defaults`

This means that if you want to apply default values, you need all 5 of these. For example if the `ClusterTemplate` does not contain the number of APIserver replicas, we check the `Seed`, then the `KubermaticConfiguration`, then fallback to the constants.

However, in our codebase we invert this flow: Whenever someone uses a `KubermaticConfigurationGetter`, that thing will already apply the Go constants to the `KubermaticConfiguration`. So the config returned by the getter is already defaulted.

### Important to know:
When the `Cluster.spec` is extended with new values that have defaults assigned, all existing Clusters are updated to set the default value.
If no initial defaults are set, keep in mind that a change later will again trigger an update to all existing Clusters.

Default values in `Cluster` objects are persisted, so that changed defaults **do not**
affect existing clusters. This is different to the `KubermaticConfiguration`/`Seed`, where defaulting
happens only at runtime, because we do want new defaults (like new `spec.versions` in the `KubermaticConfiguration`)
to also apply to existing KKP installations. 

### Validation

Validation now means that we have a single set of validation rules that we use from everywhere in the codebase. We do make a distinction between "validation NEW clusters" and "validate cluster updates", so we can ensure immutability for certain fields.

### KKP API

The KKP API does something funky: During cluster creation, it will create a `ClusterSpec` and then run the validation logic on it. Any error there is then reported back up to the user. **This** is the one reason why we had separate validation logics: One was only for a few fields that are relevant in the KKP dashboard, the other was, well, other fields. However now that we only have 1 function for it, this means that the KKP API also has to do defaulting before it can validate. And this leads to the new situation that the KKP API will always call `seedClient.Create(ctx, cluster)` with a fully defaulted `Cluster` object. This has no visible or noticeable side effects, it's just something "good to know".

Technically the API could just call `seedClient.Create()`, get the answer (from the webhook) and take any errors from there and send them to the user, but this might require more response parsing and error handling. So for now, inside the KKP API we still manually run the defaulting/validation.

### Unit Tests

The refactoring has shown lots and lots of unit tests that were technically broken. For example the tests for the `PatchEndpoint()` started with incomplete clusters and then asserted incomplete patches. Now, because way more validation rules apply everywhere, a bunch of tests had to extended to provide a `Seed`, a `KubermaticConfiguration` etc.

Importantly, the tests for the webhooks break the mentioned "first mutate, then validate" rule: The validation webhook's tests sometimes test clusters and situations that can be misleading if not understood in the full context. For example there is a test that says "reject empty nodeport range", which is true: A `Cluster` must have a nodeport range. But in reality, the mutating webhook will always ensure that it is set. But just from reading the unit tests it could seem the opposite. I left a big fat comment to warn future devs.

### Other Goodies

* The mutating webhook now uses a `SeedGetter` to dynamically fetch the `Seed` at runtime. This means changes to the default `ClusterTemplate` are respected without the need to restart the entire seed-ctrlmgr. I would consider the old behaviour to be a bug.
* Some CLI flags were removed, because for proper defaulting (and because the `ClusterTemplate` is now fetched very late at runtime), I would have had to pipe the CLI flags through to the webhook and the defaulting functions, which was just ridiculous. The flags were deprecated and, more importantly, an implementation detail for which we never gave any stability guarantees.
* The generated `Seed` example YAML file is now much nicer and represents the actual defaults much better. Nice side effect :)
* Cloud providers are sorted by name, because some of our API tests actually rely on the order of items in `SupportedProviders`. A new `TestEnsureThatProvidersAreSorted` hopefully hints at whoever breaks the order and is greeted with lots of failing tests.
* From now on, inside KKP we can always assume that `Cluster` objects are actually valid and complete.

### Does this PR introduce a user-facing change?

```release-note
Cluster resources are now defaulted and validated a lot more thoroughly by the webhooks.
```


dermorz:
Can you please add parts of the issue description into a `docs/defaulting-and-validation.md`? Then there will be a guideline to avoid that the next developer implementing defaulting/validation puts it in a random place.

xrstf:
> Can you please add parts of the PR description into a docs/defaulting-and-validation.md? Then there will be a guideline to avoid that the next developer implementing defaulting/validation puts it in a random place.

Done

xrstf:
@floreks could you take a look please? :)

dermorz:
Looking at the test failures this doesn't seem like a flaky test. See one formatted example output:

```
cluster_test.go:928: Expected response body to be 
{
    "id": "mqqcndrvl5",
    "name": "keen-snyder",
    "creationTimestamp": "0001-01-01T00:00:00Z",
    "type": "kubernetes",
    "spec": {
        "cloud": {
            "dc": "fake-dc",
            "fake": {}
        },
        "version": "1.22.4",
        "oidc": {},
        "enableUserSSHKeyAgent": true,
        "containerRuntime": "containerd",
        "clusterNetwork": {
            "services": {
                "cidrBlocks": [
                    "10.240.16.0/20"
                ]
            },
            "pods": {
                "cidrBlocks": [
                    "172.25.0.0/16"
                ]
            },
            "dnsDomain": "cluster.local",
            "proxyMode": "ipvs",
            "nodeLocalDNSCacheEnabled": true
        }
    },
    "status": {
        "version": "1.22.4",
        "url": "",
        "externalCCMMigration": "Unsupported"
    }
}
            got 
{
    "id": "mqqcndrvl5",
    "name": "keen-snyder",
    "creationTimestamp": "0001-01-01T00:00:00Z",
    "type": "kubernetes",
    "spec": {
        "cloud": {
            "dc": "fake-dc",
            "fake": {}
        },
        "version": "1.22.4",
        "oidc": {},
        "enableUserSSHKeyAgent": true,
        "containerRuntime": "containerd",
        "clusterNetwork": {
            "services": {
                "cidrBlocks": [
                    "10.240.16.0/20"
                ]
            },
            "pods": {
                "cidrBlocks": [
                    "172.25.0.0/16"
                ]
            },
            "dnsDomain": "cluster.local",
            "proxyMode": "ipvs",
            "nodeLocalDNSCacheEnabled": true
        },
        "cniPlugin": {
            "type": "canal",
            "version": "v3.20"
        }
    },
    "status": {
        "version": "1.22.4",
        "url": "",
        "externalCCMMigration": "Unsupported"
    }
}
```
The spec now also contains the field `cniPlugin` and the tests don't expect it. Let's pause the triage bot for now.

/hold

rastislavs:
Not sure if is connected to the failing tests, but maybe it would help if I explain how `cniPlugin` defaulting worked before this change:
 - It was only defaulted upon cluster creation.
 - It was not expected to be defaulted upon cluster update. The reason is that for old user clusters which still have `cniPlugin == nil` that would cause unconditional CNI upgrade.

xrstf:
> It was not expected to be defaulted upon cluster update. The reason is that for old user clusters which still have cniPlugin == nil that would cause unconditional CNI upgrade.

That's not how this works. Either there are default values and we apply them, or there are no defaults and we reject incomplete Cluster objects.

WeirdMachine:
CNI are now omitted from defaulting when updating existing clusters.

pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -89,12 +103,29 @@ func (h *AdmissionHandler) Handle(ctx context.Context, req webhook.AdmissionRequ
...
+		// apply defaults to the new version
+		err := h.applyDefaults(ctx, cluster)
+		if err != nil {
+			h.log.Info("cluster mutation failed", "error", err)
+			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("cluster mutation request %s failed: %v", req.UID, err))
+		}
+
+		if preventCNIDefaulting {
+			cluster.Spec.CNIPlugin = nil
+		}
```
xmudrii, line 125: Why don't we default this to a proper value? Please correct me if I'm mistaken, but don't we reliably know that if `cluster.Spec.CNIPlugin == nil` then the CNI plugin is Canal v3.9? The purpose of defaulting is to populate as many fields as possible so we don't have nils -- this simplifies logic later significantly.

mate4st: There is custom logic for upgrading or setting CNI's. We do not want  this to happen automatically for all old clusters as we cannot ensure that this wont break existing clusters.  See: https://github.com/kubermatic/kubermatic/blob/0ab42f058ca9e693b506569c5737e42daa153142/pkg/validation/cluster.go#L710

pkg/validation/cluster.go
```diff
@@ -135,11 +135,6 @@ func ValidateClusterUpdate(ctx context.Context, newCluster, oldCluster *kubermat
 		allErrs = append(allErrs, errs...)
 	}
 
-	portRangeFld := field.NewPath("componentsOverride", "apiserver", "nodePortRange")
-	if err := ValidateNodePortRange(newCluster.Spec.ComponentsOverride.Apiserver.NodePortRange, portRangeFld); err != nil {
```
kron4eg: what happened with ValidateNodePortRange?

mate4st: It moved and was a duplication. Its now at: 
https://github.com/kubermatic/kubermatic/blob/0ab42f058ca9e693b506569c5737e42daa153142/pkg/validation/cluster.go#L104

kron4eg (approved):

PR #8392, opened 2021-12-06
Use upstream helm chart for cert-manager as a dependency

### LINKED ISSUE #7899: update chart to user upstream: cert-manager
(general description in the epic)

Acceptance criteria:
* evaluated if it's possible to use and extend the chart to use upstream as a dependency
* required changes are implemented
* CI configuration is added
* migration and changes required to upgrade are documented

wurbanski:

Fixes #7899

**Documentation**:

```release-note
cert-manager chart has been updated to use upstream chart from https://charts.jetstack.io as a dependency, version 1.5.2
ACTION REQUIRED: configuration for cert-manager should now be included under key `cert-manager` in values.yaml file (changed from `certManager`)
ACTION REQUIRED: The chart will no longer configure ClusterIssuers. Refer to the documentation of cert-manager for configuration guide: https://cert-manager.io/docs/configuration/
```


charts/cert-manager/test/test-certmanager.sh
```diff
@@ -0,0 +1,54 @@
...
+helm upgrade \
+  --install \
+  --namespace cert-manager \
+  --create-namespace \
+  --atomic \
+  --set cert-manager.clusterIssuers.letsencrypt-prod.email=dev@kubermatic.com \
+  --set cert-manager.clusterIssuers.letsencrypt-staging.email=dev@kubermatic.com \
+  cert-manager charts/cert-manager/
+
+echodate "Downloading cmctl..."
```
SimonTheLeg: I think it would be good to have a pre-check here if cmctl is already installed, similar to how we do it here https://github.com/kubermatic/kubermatic/blob/master/hack/update-prometheus-rules.sh#L25
Makes it a bit more pleasant running this for example locally. For this case the cmctl_cleanup{} also needs to be slightly adapted I think.

charts/cert-manager/test/test-certmanager.sh
```diff
@@ -0,0 +1,54 @@
...
+  exit $exitcode
+}
+trap cmctl_cleanup EXIT
+
+echodate "Testing cert-manager..."
+./cmctl check api --wait=2m
+exitcode=$?
+
+echodate "Deleting kind cluster..."
+kind delete cluster --name "$KIND_CLUSTER_NAME"
```
SimonTheLeg: One thing I noticed overall in the script: is there a specific reason why `set -euo pipefail` is not set? From my point of view I think it should be, because there is no real reason for the script to continue if it fails in the middle (e.g. at downloading cmctl).

If so of course we would need to move the kind cluster deletion into the cleanup, similar to how it is done here https://github.com/kubermatic/kubermatic/blob/master/hack/run-ccm-migration-e2e-test-in-kind.sh#L52

pkg/install/stack/kubermatic-master/certmanager.go
```diff
@@ -450,3 +469,97 @@ func deleteCertificate(ctx context.Context, kubeClient ctrlruntimeclient.Client,
...
+	mutatingWebhooksList := &unstructured.UnstructuredList{}
+	mutatingWebhooksList.SetGroupVersionKind(schema.FromAPIVersionAndKind("admissionregistration.k8s.io/v1", "MutatingWebhookConfigurationList"))
+
+	certManagerObjectsSelector := client.MatchingLabels{
+		"app.kubernetes.io/managed-by": "Helm",
+		"app.kubernetes.io/instance":   release.Name,
+	}
+
+	if err := kubeClient.List(ctx, deploymentsList, client.InNamespace(CertManagerNamespace), certManagerObjectsSelector); err != nil {
+		logger.Warn("Error querying API for the existing deployment, attempting to upgrade without removing it...")
```
SimonTheLeg: Out of curiosity: What kind of error are we talking about here where we say it is ok to ignore? 

What I am trying to say is: If it is about handling that there are no deployments, I think it does not return an error but only an empty list. In this case I would much rather prefer something like this: 

```go
...;  err != nil {
  return fmt.Errorf(....
}

if len(deploymentsList.Items) == 0 {
  logger.Warn("Could not find existing deployment, attempting to upgrade without removing it...")
}
```

I think this would be safer, in a situation like the following: Let's assume in the cluster there is a deployment running, however while this func makes the request the network connection fails (just this one time). In that case I think our code above would swallow this issue and just skip the deletion of the deployments. Afterwards though all requests go through and it runs the rest of the func and later deployment, which I think is not what we want

SimonTheLeg (changes requested):

pkg/install/stack/kubermatic-master/certmanager.go
wurbanski: I think you're right, I'm changing this.

charts/cert-manager/test/test-certmanager.sh
wurbanski: I made slight changes to this, please take a look now.

charts/cert-manager/test/test-certmanager.sh
wurbanski: Frankly, I just renamed it and changed the cmctl, didn't look close enough to nice it's not there :-) I added this now.

SimonTheLeg: Looks good so far. However the last line in the script which deletes the cluster now needs to be trapped as well. Because what can happen is that is that the script starts the cluster but fails in the middle. With the current code it would not run the cluster deletion. Additionally if I recall bash correctly you also need the `|| true` so the cleanup does not break the script, similar to here master/hack/run-ccm-migration-e2e-test-in-kind.sh#L52

wurbanski: Yeah, I added the `|| true` and exit with correct code. I captured it, but didn't use it. I think that's why there was no `-eo pipefail` before and I removed it as well, because the only thing that we care about is handled anyway.

SimonTheLeg: Okay I think this is ok now. You are right, we are doing it in other places and it also fits our [bash guidelines](https://github.com/kubermatic/kubermatic/blob/weekly-2021-50/docs/code-style.md#guidelines). I am resolving the conversation now.

SimonTheLeg (approved):

pkg/install/stack/types.go
```diff
@@ -40,7 +40,7 @@ type DeployOptions struct {
 	ChartsDirectory                   string
 	Logger                            *logrus.Entry
 	EnableCertManagerV2Migration      bool
-	EnableNginxIngressMigration       bool
+	EnableChartMigration              bool
```
xrstf: Please keep the "1 flag per migration" logic. Those flags are not meant to be convenient, but instead they should stop the admin and prevent them from making mistakes.

If we now introduce a "catch all" flag, the admin will just change their deployment script to include `--enable-chart-migrations` and forget about it. With the KKP release after that, we might ship some super critical migration that includes some downtime and the admin would be caught off guard and have the migration run without their consent.

To prevent this, the flags are there. If it's required, we might have 1 flag per chart, or even multiple (IMHO). ^^ I would have very little problem with requring `--enable-prometheus-migration --yes-really-migrate-prometheus --accept-monitoring-downtime=$(date)` if we had some super critical Prometheus migration and we really, really need the admin to read the release notes first.

The inconvenience of having to add yet another flag (temporarily) to the installer is insignificant compared to the safety win and peace of mind. :)

charts/cert-manager/Chart.yaml
```diff
@@ -28,3 +28,8 @@ sources:
 maintainers:
 - name: The Kubermatic Kubernetes Platform contributors
   email: support@kubermatic.com
+
+dependencies:
+- repository: https://charts.jetstack.io
+  name: cert-manager
+  version: 1.5.2
```
xrstf: Your editor is not adding trailing newlines to the end of files. 😁

wurbanski: It will now ;-) also fixed in the code.

xrstf (approved):
/approve
/hold

unhold when ready

embik:
@wurbanski IMHO, this PR makes an update to our install docs necessary (https://docs.kubermatic.com/kubermatic/master/installation/install_kkp_ce/). I don't see the new requirement for creating the CertIssuers yourself documented anywhere beyond the changelog here.

PR #8398, opened 2021-12-06
figure out weather to use https or http for minio access

mate4st:

Uses http or https prefix to figure out the connection type. If not set, by default uses https but gives the option to the user by configuring the url with http.

Fixes #https://github.com/kubermatic/metering/issues/79

```release-note
NONE
```


simon-wessel (approved):
Removing the scheme with the double slash seems not to be possible without using `Replace`, even when using `net/url` to parse/modify the URL. So this is fine.

PR #8408, opened 2021-12-07
Remove dns-resolver deployment and custom dnsConfig in Konnectivity setup

### LINKED ISSUE #8256: Remove dns-resolver deployment and custom dnsConfig in Konnectivity setup
When Konnectivity is used for control-plane to worker-nodes communication, custom dns-resolver for each user cluster running in seed is not required anymore for the apiserver to resolve user-cluster service names.

The aim of this task is to verify that the following can be removed in Konnectivity setup + perform the change:
- dns-resolver deployment in seed,
- custom dnsConfig in apiserver, controller-manager, scheduler pods in seed.

rastislavs:
When Konnectivity is used for control-plane to worker-nodes communication, custom dns-resolver for each user cluster running in seed is not required anymore for the apiserver to resolve user-cluster service names. This PR makes sure it is not deployed if Konenctivity is enabled, but at the same time does not change anything in the deployment of old OpenVPN setup.

Fixes #8256 

- Apiserver network policy needs to be modified as well, to allow all egress DNS traffic (as DNS requests from apiserver now go directly to the default DNS resolver of the seed cluster, which is platform-specific).
- `DNSResolverDown` alerting rule should not be active if dns-resolver is not deployed.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/resources/apiserver/deployment.go
+++ b/pkg/resources/apiserver/deployment.go
@@ -90,11 +90,17 @@ func DeploymentCreator(data *resources.TemplateData, enableOIDCAuthentication bo
 
 			etcdEndpoints := etcd.GetClientEndpoints(data.Cluster().Status.NamespaceName)
 
-			// Configure user cluster DNS resolver for this pod.
-			dep.Spec.Template.Spec.DNSPolicy, dep.Spec.Template.Spec.DNSConfig, err = resources.UserClusterDNSPolicyAndConfig(data)
-			if err != nil {
-				return nil, err
+			if !data.IsKonnectivityEnabled() {
+				dep.Spec.Template.Spec.DNSPolicy, dep.Spec.Template.Spec.DNSConfig, err = resources.UserClusterDNSPolicyAndConfig(data)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				// custom DNS resolver in not needed in Konnectivity setup
+				dep.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirst
+				dep.Spec.Template.Spec.DNSConfig = nil
 			}
+
 			dep.Spec.Template.Spec.Volumes = volumes
 			dep.Spec.Template.Spec.InitContainers = []corev1.Container{
 				etcdrunning.Container(etcdEndpoints, data),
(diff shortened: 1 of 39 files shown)


pkg/resources/cloudcontroller/vsphere.go
```diff
@@ -70,11 +70,18 @@ func vsphereDeploymentCreator(data *resources.TemplateData) reconciling.NamedDep
...
+					resources.UserClusterDNSPolicyAndConfig(data)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				// custom DNS resolver in not needed in Konnectivity setup
+				dep.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirst
+				dep.Spec.Template.Spec.DNSConfig = nil
 			}
+
```
mfranczy: that part of the code repeats quite often, could you make a function out of this? Maybe good place would be to create a new file under `pkg/resources/features.go` and have a place to check enabled feature gates? Or add this to `pkg/resources/resources.go` like we have for `UserClusterDNSPolicyAndConfig`?

mfranczy: Or even could be a good idea to refactor `resources.UserClusterDNSPolicyAndConfig` and check inside if konnectivity is enabled? What do you think?

rastislavs: good idea, let me try that

pkg/resources/apiserver/deployment.go
```diff
@@ -90,11 +90,17 @@ func DeploymentCreator(data *resources.TemplateData, enableOIDCAuthentication bo
...
-				return nil, err
+			if !data.IsKonnectivityEnabled() {
+				dep.Spec.Template.Spec.DNSPolicy, dep.Spec.Template.Spec.DNSConfig, err = resources.UserClusterDNSPolicyAndConfig(data)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				// custom DNS resolver in not needed in Konnectivity setup
+				dep.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirst
+				dep.Spec.Template.Spec.DNSConfig = nil
```
PratikDeoghare: nit: This `if...else...` would easier to read if it was the other way round.

`   `if data.IsKonnectivityEnabled() { ...} else {...}`

mfranczy (approved):
/approve
/lgtm

mfranczy:
oh I see that linter is complaining about missing fields.
```
pkg/handler/v2/etcdrestore/etcdrestore.go:157:62: req.ProjectID undefined (type listEtcdRestoreReq has no field or method ProjectID) (typecheck)
		erList, err := listEtcdRestore(ctx, userInfoGetter, c, req.ProjectID)
```

PR #8416, opened 2021-12-08
Kubevirt csi driver integration

mfranczy:
We need it to provision storage for user clusters created on top of KubeVirt

Fixes #

**Documentation**:

```release-note
Support for KubeVirt CSI driver
```

### DIFF
--- a/pkg/provider/cloud/kubevirt/client.go
+++ b/pkg/provider/cloud/kubevirt/client.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubevirt
+
+import (
+	"encoding/base64"
+
+	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func NewClientWithRestConfig(kubeconfig string) (ctrlruntimeclient.Client, *restclient.Config, error) {
+	config, err := base64.StdEncoding.DecodeString(kubeconfig)
+	if err != nil {
+		// if the decoding failed, the kubeconfig is sent already decoded without the need of decoding it,
+		// for example the value has been read from Vault during the ci tests, which is saved as json format.
+		config = []byte(kubeconfig)
+	}
+
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(config)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	client, err := ctrlruntimeclient.New(restConfig, ctrlruntimeclient.Options{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return client, restConfig, nil
+}
--- a/pkg/provider/cloud/kubevirt/storage.go
+++ b/pkg/provider/cloud/kubevirt/storage.go
@@ -1,0 +1,80 @@
+/*
+Copyright 2021 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubevirt
+
+import (
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+func csiServiceAccountCreator(name string) reconciling.NamedServiceAccountCreatorGetter {
+	return func() (string, reconciling.ServiceAccountCreator) {
+		return name, func(sa *corev1.ServiceAccount) (*corev1.ServiceAccount, error) {
+			return sa, nil
+		}
+	}
+}
+
+func csiRoleCreator(name string) reconciling.NamedRoleCreatorGetter {
+	return func() (string, reconciling.RoleCreator) {
+		return name, func(r *rbacv1.Role) (*rbacv1.Role, error) {
+			r.Rules = []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{"cdi.kubevirt.io"},
+					Resources: []string{"datavolumes"},
+					Verbs:     []string{"get", "create", "delete"},
+				},
+				{
+					APIGroups: []string{"kubevirt.io"},
+					Resources: []string{"virtualmachineinstances"},
+					Verbs:     []string{"list"},
+				},
+				{
+					APIGroups: []string{"subresources.kubevirt.io"},
+					Resources: []string{"virtualmachineinstances/addvolume", "virtualmachineinstances/removevolume"},
+					Verbs:     []string{"update"},
+				},
+			}
+
+			return r, nil
+		}
+	}
+}
+
+func csiRoleBindingCreator(name, namespace string) reconciling.NamedRoleBindingCreatorGetter {
+	return func() (string, reconciling.RoleBindingCreator) {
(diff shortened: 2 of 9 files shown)


mfranczy:
/hold

I have to pass csi driver kubeconfig in a bit different way

pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -44,30 +43,37 @@ func (k *kubevirt) DefaultCloudSpec(spec *v1.CloudSpec) error {
...
 	}
 
-	config, err := base64.StdEncoding.DecodeString(kubeconfig)
+	return nil
+}
+
+func (k *kubevirt) InitializeCloudProvider(cluster *v1.Cluster, update provider.ClusterUpdater) (*v1.Cluster, error) {
+	ctx := context.Background()
+	// soon the namespace will be changed to the cluster namespace
+	namespace := "default"
```
hdurand0710: <3

pkg/addon/template.go
```diff
@@ -229,6 +237,16 @@ type MLASettings struct {
 	LoggingEnabled bool
 }
 
+type CloudProviderData struct {
+	// KubeVirt specific data
+	KubeVirtData KubeVirtData
+}
+
+type KubeVirtData struct {
+	// StorageProvisionerKubeconfig represents a kubeconfig for user cluster to handle CSI driver requests
```
hdurand0710: user cluster is the Kubevirt infra cluster when not deployed via Addons right ?

pkg/provider/cloud/kubevirt/client.go
```diff
@@ -0,0 +1,90 @@
...
+
+type Clientset struct {
+	RestConfig *restclient.Config
+	// KubeVirt clientset
+	KubeVirt kv.Interface
+	// Kubernetes clientset
+	Kubernetes kubernetes.Interface
+}
+
+func NewClientset(cluster *v1.Cluster, secretKeyGetter provider.SecretKeySelectorValueFunc) (*Clientset, error) {
```
hdurand0710: As you said in your previous comments, I would need a slightly different signature to reuse it for the VmiPreset and StorageClass.

hdurand0710:
Looks good to me, just waiting for the ClientSet.

hdurand0710 (approved):

pkg/provider/cloud/kubevirt/client.go
```diff
@@ -0,0 +1,92 @@
...
+	// KubeVirt clientset
+	KubeVirt kv.Interface
+	// Kubernetes clientset
+	Kubernetes kubernetes.Interface
+}
+
+func NewClientset(kubeconfig string) (*Clientset, error) {
+	config, err := base64.StdEncoding.DecodeString(kubeconfig)
+	if err != nil {
+		// if the decoding failed, the kubeconfig is sent already decoded without the need of decoding it,
```
moadqassem, line 30: So the problem with that is, in the past we had some issues passing and parsing the Kubeconfig as a json or yaml from UI to API and vice versa. Thus we demanded that this should be base64 encoded so we avoid all the issues that comes a long parsing data as plaintext

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func EnsureInfraCSIAccess(ctx context.Context, kubeconfig string, clusterName string) ([]byte, error) {
+	// soon we are going to change the default namespace to cluster namespace
```
moadqassem: What is the issue of doing it now? I mean the name space of the cluster is provided already to the function ;) 
`clusterName string` => `cluster-[clusterName]`

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -0,0 +1,129 @@
...
+	namespace := "default"
+
+	cli, err := NewClientset(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	sa := &corev1.ServiceAccount{
```
moadqassem: Do we know already what kinda resources the client would access? I mean giving access to all resources might impose a security issues.

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -0,0 +1,129 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubevirt-csi",
+			Namespace: namespace,
+		},
+	}
+	_, err = cli.Kubernetes.CoreV1().ServiceAccounts(namespace).Create(ctx, sa, metav1.CreateOptions{})
```
moadqassem: Why not using the reconciling functions for creating those resources? Same goes for all the resources that you are creating. https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/reconciling/zz_generated_reconcile.go#L177-L178

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -0,0 +1,129 @@
...
+				Verbs:     []string{"list"},
+			},
+			{
+				APIGroups: []string{"subresources.kubevirt.io"},
+				Resources: []string{"virtualmachineinstances/addvolume", "virtualmachineinstances/removevolume"},
+				Verbs:     []string{"update"},
+			},
+		},
+	}
+	_, err = cli.Kubernetes.RbacV1().Roles(namespace).Create(ctx, role, metav1.CreateOptions{})
```
moadqassem: Why not using this reconciling functions instead accessing the api directly: 
https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/reconciling/zz_generated_reconcile.go#L476-L477

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     "kubevirt-csi",
+		},
+	}
+
+	_, err = cli.Kubernetes.RbacV1().RoleBindings(namespace).Create(ctx, roleBinding, metav1.CreateOptions{})
```
moadqassem: Same goes here, you can use this function type instead https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/reconciling/zz_generated_reconcile.go#L513-L514

moadqassem (changes requested):

pkg/provider/cloud/kubevirt/storage.go
mfranczy: I would have to change this for KubeVirt CCM, would be better to do it in the future PRs.

pkg/provider/cloud/kubevirt/storage.go
mfranczy: Good idea, was not aware of those.

pkg/provider/cloud/kubevirt/client.go
mfranczy: We should merge it in this way and adjust in your PR.

pkg/addon/template.go
mfranczy: User cluster is a new k8s cluster created on top of KubeVirt

pkg/provider/cloud/kubevirt/storage.go
mfranczy: I am not giving an acess to all resources here.

pkg/provider/cloud/kubevirt/storage.go
mfranczy: I was not aware of those, good idea.

pkg/provider/kubernetes/credentials.go
```diff
@@ -360,15 +361,25 @@ func createOrUpdatePacketSecret(ctx context.Context, seedClient ctrlruntimeclien
...
 
+	// ensure that CSI driver on user cluster will have an access to KubeVirt cluster
+	r, err := kubevirt.NewReconciler(spec.Kubeconfig, cluster.Name)
+	if err != nil {
+		return err
+	}
+	csiKubeconfig, err := r.ReconcileCSIAccess(ctx)
+	if err != nil {
+		return err
+	}
```
mfranczy, line 377: With the current design I am concerned about that part. There is only one request to create a cluster and potentially we might hit a race condition inside `ReconcileCSIAccess`. It is a small risk, although still possible. As far as I can see KubeVirt is the first provider that automatically generate secrets for user cluster components.

Refactoring of the whole mechanism would be much bigger than scope of the issue. I would like to talk about this code with @zreigz on Monday. If we see that this problem occurs often then we will have to do refactoring or implement a better mechanism to generate KubeVirt secrets.

Maybe we should implement a proper reconciler for KubeVirt cloud provider and from there inject a kubeconfig for CSI driver, although that opens a new way to interact with user clusters (we skip addon mechanism)

pkg/provider/cloud/kubevirt/storage.go
mfranczy: the same applies to the UI

pkg/provider/kubernetes/credentials.go
mfranczy: as discussed with @zreigz it's ok for now to leave it like this.

PR #8425, opened 2021-12-09
add tolerations to user-cluster system components

### LINKED ISSUE #7142: not all operational daemonsets are scheduled on the tainted nodes
**User Story**
I added a tainted nodepool to the user cluster (tunneling expose strategy), but noticed that no pod could run there. Even though the pods were scheduled, they were failing during setup of the container runtime. The reason for this was canal pods couldn't establish connection to the apiserver and were crash looping. It turned out, they can't connect to the apiserver because the envoy-agent is missing in the tainted nodes, envoy-agent daemonset is missing tolerations for the taints. As consequence the tainted nodes are unusable:

```
kubectl get ds canal -o go-template='{{ .spec.template.spec.tolerations }}'
[map[effect:NoSchedule operator:Exists] map[key:CriticalAddonsOnly operator:Exists] map[effect:NoExecute operator:Exists]]
```
vs. 
```
kubectl get ds envoy-agent -o go-template='{{ .spec.template.spec.tolerations }}'
<no value>
```

**Acceptance criteria**
All the required operational daemonsets are able to schedule pods on the tainted nodes.

aborilov:
**What this PR does / why we need it**: In this PR we added tolerations to user-cluster system components to make sure that they are deployed on each node if even user specify some taints

Fixes #7142 

**Documentation**:

```release-note
Add tolerations to user cluster system daemonsets.
```

### DIFF
--- a/addons/openvpn/openvpn-client-dep.yaml
+++ b/addons/openvpn/openvpn-client-dep.yaml
@@ -119,6 +119,11 @@ spec:
             memory: 50Mi
       restartPolicy: Always
       terminationGracePeriodSeconds: 5
+      tolerations:
+       - effect: NoSchedule
+         operator: Exists
+       - effect: NoExecute
+         operator: Exists
       volumes:
       - name: openvpn-client-certificates
         secret:
--- a/addons/openvpn/openvpn-client-dep.yaml
+++ b/addons/openvpn/openvpn-client-dep.yaml
@@ -119,11 +119,6 @@ spec:
             memory: 50Mi
       restartPolicy: Always
       terminationGracePeriodSeconds: 5
-      tolerations:
-       - effect: NoSchedule
-         operator: Exists
-       - effect: NoExecute
-         operator: Exists
       volumes:
       - name: openvpn-client-certificates
         secret:
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/core-dns/deployment.go
@@ -92,6 +92,16 @@ func DeploymentCreator(kubernetesVersion *semver.Version, registryWithOverwrite 
 			dep.Spec.Template.Spec.Volumes = volumes
 
 			dep.Spec.Template.Spec.Containers = getContainers(kubernetesVersion, registryWithOverwrite)
+			dep.Spec.Template.Spec.Tolerations = []corev1.Toleration{
+				{
+					Effect:   corev1.TaintEffectNoSchedule,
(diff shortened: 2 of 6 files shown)


addons/openvpn/openvpn-client-dep.yaml
```diff
@@ -119,6 +119,11 @@ spec:
             memory: 50Mi
       restartPolicy: Always
       terminationGracePeriodSeconds: 5
+      tolerations:
+       - effect: NoSchedule
+         operator: Exists
+       - effect: NoExecute
+         operator: Exists
```
embik: These "blanket" tolerations (for OpenVPN and all others in this PR) can be problematic because cordoning off a node doesn't work properly anymore (`kubectl cordon` and `kubectl drain` set the taint `node.kubernetes.io/unschedulable:NoSchedule`). This is fine for `DaemonSets` (envoy-agent, node-local-dns, promtail), but it can cause Deployment/StatefulSet Pods to go right back onto the node after being evicted (as no taint can prevent them from being scheduled), causing an endless loop that never fully drains a node (for example, during machinedeployment upgrades).

embik (changes requested):

addons/openvpn/openvpn-client-dep.yaml
aborilov: right we were going back and forth with whether to add that to deployment

aborilov: @embik but what about case when user add only one machine-deployment with some taints, those system components won't be ever scheduled

embik: @aborilov that's correct, but only creating one MachineDeployment and then adding taints to it doesn't make any sense -  The idea of taints is to exclude Pods from a set of nodes, so _they run elsewhere_. If there are no nodes without taints, all Pods need a toleration, and taints and tolerations offset each other.

What we could do is add a special toleration to system components (`kkp.k8c.io/system:NoSchedule` or something) and document that in the KKP documentation as necessary if you want to create your initial MachineDeployment as a "system-only" pool of nodes, you need to use that toleration. Alternatively, we could document that the initial MachineDeployment must not have custom taints.

The problem is essentially that we can't do "ignore all taints, except the ones coming from k8s".

aborilov: so, let's leave tolerations only for daemon sets, that should work, right?

embik: Yeah, that should do the trick!

embik (approved):
/approve

PR #8455, opened 2021-12-13
Add support for Kubernetes 1.23

### LINKED ISSUE #8429: Support Kubernetes 1.23
**User Story**

**Acceptance criteria**

* Kubernetes 1.23 is supported

### LINKED ISSUE #9389: Investigate Kubernetes 1.23 cluster creation failures on seeds running 1.23.4

xmudrii:

Add support for Kubernetes 1.23. This PR also updates relevant components, such as CCMs and CSI drivers, cluster-autoscaler, and more.

## 1.23 Support Policy

Kubernetes 1.23 is currently supported on:

* AMD64-based clusters with Canal and kube-proxy in the IPVS mode (IPVS is used by default for all clusters unless specified otherwise)
* All Canal clusters running kube-proxy in the iptables mode
* All Cilium clusters

In the other words, Kubernetes 1.23 is not supported on ARM64 clusters with Calico and kube-proxy in the IPVS mode. That's because Calico currently doesn't support this combination. See https://github.com/projectcalico/calico/issues/5717 and https://github.com/projectcalico/calico/issues/5011 for more details.

### Important Information about the Implementation

The current implementation allows creating ARM64-based Kubernetes 1.23 clusters with Canal and kube-proxy in the IPVS mode. We don't have a way to determine if the selected instance type is AMD64 or ARM64 based. In this case, the user will get a non-functional cluster, but they can just delete it and use 1.22, or switch to AMD64 instances.

Because the incompatibility validation is based on checking the Node objects, if there are no nodes in the cluster, KKP will allow upgrading to 1.23 even if MachineDeployments are using ARM64 instance types. Scaling up the cluster will create non-functional worker nodes. Users can replace those worker nodes with AMD64-based nodes or create a new cluster.

## CSIMigration

Starting with Kubernetes 1.23, CSIMigration is enabled by default for AWS, Azure (AzureDisk), and GCE. Those three providers don't support fallback to the in-tree volume providers. This means that we either have to deploy the CSI driver or disable the CSI migration.

To mitigate this, the appropriate CSIMigration feature gate (e.g. `CSIMigrationAWS`) will be disabled if:
* the cluster is running an in-tree cloud provider
* the cluster is running Kubernetes 1.23 or newer

An expectation to this rule is OpenStack, where we don't disable `CSIMigrationOpenstack` because OpenStack has a working fallback mechanism.

Fixes #8429
Fixes #9389

**Documentation**:

https://github.com/kubermatic/docs/pull/1024

```release-note
Add support for Kubernetes 1.23
Kubernetes 1.23 is currently not supported on ARM64 clusters running Canal and kube-proxy in the IPVS mode. KKP will allow the creation of new 1.23 clusters with the ARM64 nodes, Canal, and kube-proxy in the IPVS mode, but those clusters will not be usable. In this case, you can delete the cluster and create a 1.22 cluster, or switch to the AMD64 nodes. Upgrades of the existing clusters 1.22 clusters with ARM64 nodes, Canal, and kube-proxy in the IPVS mode is forbidden using the newly-added `nonAMD64WithCanalAndIPVS` incompatibility.
```


xmudrii:
/hold
Calico currently doesn't support Kubernetes 1.23. Please see https://github.com/projectcalico/calico/issues/5011#issuecomment-991121424 for more details.

caseydavenport:
Hm, surprising since the Calico team have e2e CI runs against Kubernetes v1.23.1 that pass. wonder if something else is going on.

xmudrii:
@caseydavenport This issue occurs only if you run kube-proxy in the IPVS mode (which we do by default). If you run it in the iptables mode (which is used by default if you don't explicitly specify another mode), everything works fine. If Calico E2E tests are testing only using the iptables mode, it's expected that the issue didn't occur.

caseydavenport:
@xmudrii thanks, I'll chase this one on our side and provide a fix once we've got one.

xmudrii:
@caseydavenport Thanks a lot! Please let us know if you need any additional info or anything else from our side.

caseydavenport:
FYI I believe @song-jiang is working on a fix for this in this PR: https://github.com/projectcalico/calico/pull/5485

xmudrii:
@caseydavenport I’ve seen that the PR is merged and the issue is closed. Is there any ETA when the next minor or patch release will be out? Also, as far as I understand PR, only amd64 is fixed, is that right?

xmudrii:
@caseydavenport We tested the latest Calico v3.22.0, but the issue is still present.

caseydavenport:
@xmudrii yep, the fix will be in v3.22.1 - the PR for this missed code-freeze but just a little bit, so will have to get it in a patch.

xmudrii:
@caseydavenport Is there any ETA when we might expect the Calico v3.22.1 patch release?

xrstf:
https://github.com/projectcalico/calico/releases/tag/v3.22.1 has been released on Friday :)

.prow.yaml
```diff
@@ -346,7 +346,49 @@ presubmits:
...
+      preset-docker-push: "true"
+      preset-vault: "true"
+      preset-repo-ssh: "true"
+      preset-e2e-ssh: "true"
+      preset-kubeconfig-ci: "true"
+      preset-kind-volume-mounts: "true"
+      preset-goproxy: "true"
+    spec:
+      containers:
+      - image: quay.io/kubermatic/build:go-1.17-node-16-kind-0.11-7
```
xrstf: please use `.....-kind-0.12-8`

xmudrii: @xrstf Fixed, PTAL.

moadqassem (approved):
/approve
/lgtm

pkg/resources/data.go
```diff
@@ -640,6 +642,23 @@ func GetCSIMigrationFeatureGates(cluster *kubermaticv1.Cluster) []string {
 				}
 			}
 		}
+	} else if !ccm && gte23.Check(cluster.Status.Versions.ControlPlane.Semver()) {
```
xrstf: Could this explode when no cluster status exists yet (see #9420 , #9427)?

xrstf (commented):
One question, otherwise LGTM

pkg/resources/data.go
xmudrii: I guess it could. Let me quickly push a fix.

xmudrii: @xrstf Fixed, PTAL.

Issue #8463, opened 2021-12-14
Fix method to check if AKS was imported 


imharshita:
**User Story**
Fix method to check if AKS was imported 

**Acceptance criteria**

Issue #8453, opened 2021-12-13
Fix method to check if GKE was imported


zreigz:
Extend check for region and name

PR #8466, opened 2021-12-15
fix isImport externalcluster

### LINKED ISSUE #8462: Fix method to check if EKS was imported
**User Story**
Fix method to check if EKS was imported 

**Acceptance criteria**

imharshita:

**What this PR does / why we need it**: fix isImport externalcluster gke, aks. eks

Fixes #8462, #8463, #8453 

```release-note
none
```

### DIFF
--- a/pkg/handler/common/provider/aws.go
+++ b/pkg/handler/common/provider/aws.go
@@ -361,8 +361,6 @@ func ListAWSRegions(credential EKSCredential) (apiv2.Regions, error) {
 	return regionList, nil
 }
 
-type listClusters func(EKSCredential, string) ([]*string, error)
-
 func listEKSClusters(cred EKSCredential, region string) ([]*string, error) {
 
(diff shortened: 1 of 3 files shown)


pkg/crd/kubermatic/v1/preset.go
```diff
@@ -346,9 +346,9 @@ func (s EKS) IsValid() bool {
 type AKS struct {
 	ProviderPreset `json:",inline"`
 
-	TenantID       string `json:"tenantId"`
-	SubscriptionID string `json:"subscriptionId"`
-	ClientID       string `json:"clientId"`
+	TenantID       string `json:"tenantID"`
```
zreigz: @maciaszczykm if we change it we have to update presets in the UI. We can also leave like it is right now, WDYT?

maciaszczykm: I'd keep it as it is to be honest. It's the same for Azure I think.

zreigz: OK, so @imharshita please revert those changes for the `preset.go`

PR #8467, opened 2021-12-15
add prometheus scraping if minio is running with TLS

mihiragrawal:
**What this PR does / why we need it**: `BackupRestore controller` is dependant on running minio with TLS according to [comment](https://github.com/kubermatic/kubermatic/blob/5328f8a481e62a3d7656a6990bbe971b1d12347f/charts/minio/values.yaml#L21-L24). Once minio runs with TLS, Prometheus scraping fails and gives an alert `PromScrapeFailed`. Example below.

```bash
up{app="minio", instance="10.244.37.21:9000", job="pods", namespace="minio", pod="minio-78cf8fcdbc-pdfqs", pod_template_hash="78cf8fcdbc"}
```
To fix that I have added Prometheus scraping job-specific for minio and updated pod job to skip it. 

Fixes #

**Documentation**:

<!--  Write your release note:
1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
2. If  no release note is required, just write "NONE".
```release-note
add Prometheus scraping if minio is running with TLS
```

### DIFF
--- a/charts/monitoring/prometheus/config/scraping/pods.yaml
+++ b/charts/monitoring/prometheus/config/scraping/pods.yaml
@@ -21,6 +21,13 @@ relabel_configs:
   regex: '{{ .Release.Namespace }};node-exporter'
   action: drop
 
+# drop minio, if it's running with TLS
+{{ if .Values.prometheus.scraping.minio.tls.enabled }}
+- source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_pod_label_app]
+  regex: '{{ .Values.prometheus.scraping.minio.namespace | default "minio" }};{{ .Values.prometheus.scraping.minio.appLabel | default "minio" }}'
+  action: drop
+{{ end }}
+
 # drop kube-state-metrics, as we have dedicated rules to properly map all the labels
 # to their original pods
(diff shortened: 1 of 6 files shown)


charts/monitoring/prometheus/config/scraping/pods.yaml
```diff
@@ -21,6 +21,13 @@ relabel_configs:
   regex: '{{ .Release.Namespace }};node-exporter'
   action: drop
 
+# drop minio, if it's running with TLS
+{{ if .Values.prometheus.scraping.minio.tls.enabled }}
+- source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_pod_label_app]
+  regex: '{{ .Values.prometheus.scraping.minio.namespace | default "minio" }};{{ .Values.prometheus.scraping.minio.appLabel | default "minio" }}'
```
wurbanski, line 27: What do you think about just setting the defaults in `values.yaml` instead of using `| default`?

mihiragrawal: Both should work. 
But I want less friction from the user side, so I added an option as a default and an opportunity to overwrite users if they use some other value.

PR #8471, opened 2021-12-15
handle external cluster status

zreigz:
**What this PR does / why we need it**: handle external cluster status

Fixes #3950

```release-note
NONE
```

### DIFF
--- a/pkg/api/v2/types.go
+++ b/pkg/api/v2/types.go
@@ -376,6 +376,13 @@ const (
 
 	// DELETING state indicates the cluster is being deleted.
 	DELETING ExternalClusterState = "DELETING"
+
+	// UNSPECIFIED Not set.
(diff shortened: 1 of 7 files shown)


pkg/api/v2/types.go
```diff
@@ -376,6 +376,13 @@ const (
 
 	// DELETING state indicates the cluster is being deleted.
 	DELETING ExternalClusterState = "DELETING"
+
+	// UNSPECIFIED Not set.
+	UNSPECIFIED ExternalClusterState = "UNSPECIFIED"
```
maciaszczykm: In other places we use "unknown". What do you think about aligning it?

zreigz: sounds good, I will change

zreigz: done

PR #8485, opened 2021-12-17
Fix for Kubesystem network policiies

mate4st:

- Fixes wrong feature gate condition
- Sets protocol for metrics-server networkpolicy that causes a fail of reconciliation otherwise
- Fix ingress premission for metric-server

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #8494, opened 2021-12-17
fix listing security groups

### LINKED ISSUE #8447: [AWS] Listing security groups does not take selected VPC into account
Currently the `/api/v1/providers/aws/{dc}/securitygroups` endpoint ignores VPC selected by the user and lists all available security groups. It often ends up in error during cluster creation if users actually selects the security group.

https://github.com/kubermatic/kubermatic/blob/18c98aefc7ad2c9713e3bb34cd62f7e61d17ae5a/pkg/handler/v1/provider/aws.go#L115-L125

image

zreigz:
**What this PR does / why we need it**: fix listing security groups

Fixes #8447

```release-note
NONE
```

### DIFF
--- a/cmd/kubermatic-api/swagger.json
+++ b/cmd/kubermatic-api/swagger.json
@@ -6775,6 +6775,11 @@
           },
           {
             "type": "string",
+            "name": "VPC",
+            "in": "header"
+          },
+          {
+            "type": "string",
             "x-go-name": "DC",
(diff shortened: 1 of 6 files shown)


zreigz:
@floreks the UI also needs to set the VPC in the header. I think right now it's missing

mfranczy (approved):
/approve
/lgtm

PR #8505, opened 2021-12-20
Vsphere ccm csi

### LINKED ISSUE #8426: Update CCM / CSI for Vsphere
**User Story**
Update Vsphere csi and ccm to latest version and support Kubernetes 1.22.

**Acceptance criteria**
Vsphere provider works with Kubernetes 1.22

mate4st:

- Update vSphere CCM and CSI
- Fixes a small bug with konnectivity deployments and vSphere ccm

Fixes #8426

```release-note
Support latest vSphere Cloud Controller Manager and  CSI driver for Kubernetes 1.22
```


pkg/resources/cloudcontroller/deployment.go
```diff
@@ -67,7 +67,11 @@ func DeploymentCreator(data *resources.TemplateData) reconciling.NamedDeployment
 					return nil, err
 				}
 
-				containerNames := sets.NewString(ccmContainerName, openvpnClientContainerName)
+				containerNames := sets.NewString(ccmContainerName)
+
+				if data.IsKonnectivityEnabled() {
```
rastislavs: should be `if !data.IsKonnectivityEnabled()`

mlavacca (commented):
The PR looks good to me, but we should also allow users to create/upgrade clusters to v1.22 because vsphere with K8s 1.22 is forbidden atm. You can achieve so by removing these fields:
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/defaults/defaults.go#L280-L299

Is 1.23 supported by this release? if not, you can replace 1.22 with 1.23 so that we can have the block in place.

WeirdMachine:
> The PR looks good to me, but we should also allow users to create/upgrade clusters to v1.22 because vsphere with K8s 1.22 is forbidden atm. You can achieve so by removing these fields: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/defaults/defaults.go#L280-L299
> 
> Is 1.23 supported by this release? if not, you can replace 1.22 with 1.23 so that we can have the block in place.

Thanks i was looking for this part already but did not find it. 
1.23 is no supported at the moment.

mlavacca (approved):
/lgtm

pkg/crd/kubermatic/v1/datacenter.go
```diff
@@ -252,7 +252,7 @@ type DatacenterSpec struct {
 
 	//nolint:staticcheck
 	//lint:ignore SA5008 omitgenyaml is used by the example-yaml-generator
-	Fake *DatacenterSpecFake `json:"fake,omitempty,omitgenyaml"`
+	Fake *DatacenterSpecFake `json:"fake,omitempty,omitempty"`
```
kron4eg: Ok, since you've removed omitgenyaml, you also should remove the comment about this tag, isn't it?

pkg/crd/operator/v1alpha1/configuration.go
```diff
@@ -134,7 +134,7 @@ type KubermaticUIConfiguration struct {
 	// ---
 	//nolint:staticcheck
 	//lint:ignore SA5008 omitgenyaml is used by the example-yaml-generator
-	DockerTag string `json:"dockerTag,omitempty,omitgenyaml"`
+	DockerTag string `json:"dockerTag,omitempty,omitempty"`
```
kron4eg: Same in this file too.

kron4eg (changes requested):
Some small changes requested

pkg/crd/kubermatic/v1/datacenter.go
mate4st: wow i have no idea where this comes from :o

mate4st: I reverted the crd change

PR #8538, opened 2021-12-22
Enhance CNI/networking addons handling

### LINKED ISSUE #8537: Enhance CNI addon handling
CNI addon handling has a few limitations:
 - when CNI upgrade is triggered, it may take up to 5 minutes until the CNI addon will start reconciliation, so the change is not immediate
 - all CNI addons are being installed into the user cluster namespace in Seed, even it is not necessary (e.g. there are always both canal and cilium addons. Which one of them will be finally installed into the user cluster is manged only via addon templating).

rastislavs:
Enhances handling of CNI addons to address the following issues:

- When CNI settings (e.g. version) change, it may take up to 5 minutes until the CNI addon will start reconciliation, so the change is not immediate. This PR makes sure that the reconciliation starts right after CNI settings change.
- All CNI addons are being installed into the user cluster namespace in Seed, even if they are not necessary (e.g. you would always find both `cilium` and `canal` Addon CRs there, even when only one of them is actually in use). Whether or not some addon finally will be applied into the user cluster is managed at various places, either in the addon controller or in the addon manifests. This PR unifies it - the addon installer controller is the right place, so that unused addon CRs are not even deployed into the user cluster namespace in the Seed.

Fixes #8537 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/addons/kube-proxy/clusterrolebinding.yaml
+++ b/addons/kube-proxy/clusterrolebinding.yaml
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{ if ne .Cluster.Network.ProxyMode "ebpf" }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -25,4 +24,3 @@ subjects:
 - kind: ServiceAccount
   name: kube-proxy
   namespace: kube-system
-{{ end }}
--- a/addons/kube-proxy/configmap.yaml
+++ b/addons/kube-proxy/configmap.yaml
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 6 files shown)


moadqassem (approved):
/approve
/lgtm

moadqassem:
@rastislavs I guess we need to redefine CNIs a bit from its current context as I don't believe those are addons anymore. For example, at the moment we install all addons CRs(including CNIs) in each created user cluster, probably this has to change, and instead we completely snag the CNI manifests from the addons folder and instead having them in the resources folder. I would also strongly recommend using the go embed package to take care of the portability and manifests managing. WDYT ?

rastislavs:
@moadqassem 

> I guess we need to redefine CNIs a bit from its current context as I don't believe those are addons anymore.

I agree that they are slowly evolving into something more sophisticated. So far the addons infra was sufficient for them, but it may not be in the future, where we want to allow more CNI customizations. Definitely something for consideration, and discussion. We could do that as part of https://github.com/kubermatic/kubermatic/issues/8540

PR #8545, opened 2021-12-23
Disable network policies for kube-system

mate4st:

Disable network policies for kube-system due to issues with cilium

```release-note
NONE
```


rastislavs:
@WeirdMachine it seems you need to update some cluster mutation tests

xrstf:
What were those issues with Cilium?

PR #8548, opened 2021-12-23
Add Hubble addon

### LINKED ISSUE #8547: Add Hubble addon
Allow installing Hubble relay and Hubble UI as a KKP addon, in a way that would work with Cilium CNI addon in KKP.

rastislavs:
This PR adds Hubble addon - network observability UI which works with Cilium CNI installed in KKP user clusters.

As installing Hubble from cilium CLI is not compatible with helm manifest -based Cilium installation (see https://github.com/cilium/cilium-cli/issues/515), this is the easiest way how KKP users can install Hubble into their KKP clusters with Cilium.

The PR also changes the way how we generate Hubble certificates - generating them using go template functions is not the best option, as the addon manifest is periodically reconciled, which may change certificates too often. Therefore we are switching to the more standard way using cronJob provided in the CIlium Helm chart (`hubble.tls.auto.method=cronJob` Helm option).

Fixes #8547 

**Documentation**:

```release-note
Add Cilium Hubble addon for CNI network observability.
```

### DIFF
--- a/addons/hubble/hubble.yaml
+++ b/addons/hubble/hubble.yaml
@@ -18,8 +18,15 @@
 #     --set kubeProxyReplacement=strict --set k8sServiceHost=CHANGEME --set k8sServicePort=CHANGEME
 #
 # Modifications:
-#   - Only hubble-relay and hubble-ui components are placed here, the rest is part of the "cilium" addon
+#   - Only templates/hubble, templates/hubble-relay and templates/hubble-ui components are placed here, the rest is part of the "cilium" addon
 
+---
+# Source: cilium/templates/hubble/tls-cronjob/serviceaccount.yaml
+apiVersion: v1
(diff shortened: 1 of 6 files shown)


addons/hubble/hubble.yaml
```diff
@@ -0,0 +1,517 @@
+# Copyright 2021 The Kubermatic Kubernetes Platform contributors.
```
moadqassem, line 1: Do you wanna deploy this, even if Cilium is not used ?

moadqassem (approved):
/approve
/lgtm

addons/hubble/hubble.yaml
rastislavs: Nope, that does not make sense. Will protect it in the manifest template.

PR #8528, opened 2021-12-21
deploy OSM CRDs using kubermatic operator

moadqassem:
**What this PR does / why we need it**:
Introducing OSMs CRDs deployment to the seed clusters in KKP, in order to complete the OSM integration with KKP.
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Deploy OSM CRDs in KKP seed clusters 
```

PR #8552, opened 2021-12-28
Improve CRD handling in Installer

xrstf:
The installer was built for a world were our CRDs never changed and it was only important that they exist at all. Nowadays, this isn't true anymore and we recently for example extended the additionalPrinterColumns for a CRD, so updating them is required. This PR changes the code to use `Update` instead of `Patch` semantics.

This PR also restructures the CRDs a bit, with the following thoughts/assumptions behind them:

* The OSM CRDs are supposed to be installed already (they can be seen as the pioneers of the new `k8c.io` API group).
* The CRDs for `apps.kubermatic.k8c.io` are **not** supposed to be installed already.
* To keep manual installation easy, it was my goal to allow users to do a single `kubectl apply -f charts/kubermatic-operator/crd/` without having to tell them to load the OSM CRDs from a different directory.
* Also, I changed the logic with OSM, the CRDs are now _always_ installed. #8528 said it would "install CRDs with the operator", but that's not what the PR did and the operator also never ever ever installs CRDs. Going further, if an admin enables the OSM feature flag, they would have to re-run the KKP installer (!) to get the CRDs. That is very surprising and something we should not require them to do. It's IMHO much safer and easier to always have the CRDs and at worst just don't use them.

With this in mind, this PR

* moves the OSM CRDs into `crd/k8c.io` because that seems to be where we store all `*.k8c.io` CRDs.
* creates symlinks for the 2 OSM CRDs so that the manual installation still just requires a single `kubectl apply`. This is easier for users and easier to document (also the documentation will not go out of sync so quickly, compared to listing individual files). The symlinks are so that we do not have to duplicate the files.
* The CRD install logic now always installs all CRDs, regardless of feature flags.
* The CRD install logic is now abstracted into `MasterStack.InstallKubermaticCRDs` and `SeedStack.InstallKubermaticCRDs`. The logic is identical between the two, but for a developer that should not be of interest.
* The KKP installer now also installs the CRDs on seed clusters, something that was utterly broken since forever and our documentation is also wrong. Docs PR: https://github.com/kubermatic/docs/pull/883
* There is one ugly duckling in this PR and that is `DeployOptions.KubermaticCRDDirectory`. The CRD migration needs to cheat and load CRDs from a different directory than normal and so it has a custom CRD directory field. Once the CRD migration is done, the temporary field can be removed. Because it is just temporary, I think the ugliness is okay. It was more important to me to keep the `MasterStack.InstallKubermaticCRDs` and `SeedStack.InstallKubermaticCRDs` abstraction intact.

```release-note
Fix kubermatic-installer not properly updating CRDs
```


charts/kubermatic-operator/crd/k8c.io/operatingsystemmanager.k8c.io_operatingsystemconfig.yaml
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This file has been copied from
+# https://github.com/kubermatic/operating-system-manager/blob/master/deploy/crd/crd-operating-system-config.yaml
```
embik, line 16: This is a minor nitpick, but should this perhaps be a permalink / link to a tag? That would make it easier to check "upstream" if the CRD changed since the last time it was copied to kubermatic/kubermatic.

pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -394,14 +394,16 @@ func (r *reconciler) networkingData(ctx context.Context) (address *kubermaticv1.
...
 	var serviceAccount corev1.ServiceAccount
 	err := r.Get(ctx, types.NamespacedName{
 		Namespace: namespace,
 		Name:      resources.DefaultServiceAccountName,
 	}, &serviceAccount)
-
 	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil
+		}
```
embik, line 408: This seems somewhat unrelated to the PR, and I'm not sure it should be added as the default ServiceAccount should always be there, no? If it's not there, something isn't correct.

charts/kubermatic-operator/crd/k8c.io/operatingsystemmanager.k8c.io_operatingsystemconfig.yaml
xrstf: The "check if upstream changed" would be "copy the new files, check git diff" IMHO. I'm not even sure there are stable, tagged versions of OSM CRDs we could use?

pkg/controller/user-cluster-controller-manager/resources/controller.go
xrstf: I don't quite understand what you mean. This change is definitely unrelated, but I just got triggered from the endless warnings in our CI logs _when the tests shut down_.

embik: Ah, gotcha. What I mean is that I expect a "default" ServiceAccount to be there at all times as Kubernetes sets it up for each namespace. So I wonder what would even trigger this situation. I guess it might be between "namespace creation" and kube-controller loops kicking in to create it.

charts/kubermatic-operator/crd/k8c.io/operatingsystemmanager.k8c.io_operatingsystemconfig.yaml
embik: Shouldn't the CRD version in the release tag corresponding to the OSM release we deploy be the "stable" version, since it's in sync with what the OSM version we deploy expects? The `master` CRD might already diverge from what the release needs. No big deal I think, just a thought I had.

embik:
lgtm, but needs a rebase / resolving the merge conflict. I'm happy to approve it afterwards.

embik (approved):
/lgtm

PR #8553, opened 2021-12-29
always use installer in deploy.sh

xrstf:
Since #8552 the installer can now properly setup seeds. We should use it to ensure it works.

This PR also fixes the broken deployment check for the IAP chart. Since we moved to oauth2-proxy for the IAP, there is no `discovery_url` at all anymore. The only reason we still deployed is because our own values for the dev system still contained the dicovery_url. :grin: 

```release-note
NONE
```


xrstf:
/hold

needs #8552 first

ahmedwaleedmalik (approved):
/lgtm
/approve

PR #8554, opened 2021-12-30
Fix creation of ClusterRole in cert-manager chart

dakraus:
Added a missing document separator between two consecutive ClusterRole
definitions, which prevented the creation of the ClusterRole
'cert-manager-edit'.

```release-note
Create missing ClusterRole 'cert-manager-edit' for cert-manager
```


xrstf (approved):
/approve

Thank you!

PR #8556, opened 2021-12-31
use embedded upstream CRDs for gatekeeper CRD installation

### LINKED ISSUE #7971: Replace OPA CRDs deployment from code to upstream manifests/charts as static files - Test Release 2.20
**User Story**
Currently, we keep CRDs deployment in the code instead of mirroring upstream manifests or using helm charts for deployment. Every change in the CRD must be tracked in the code and copied exactly as is in the upstream manifests. That's very prone to errors. 
Also we don't copy the validation schema, we just use field `preserveUnknownFields` to stop the API server pruning fields

**Acceptance criteria**
Using Gatekeeper CRDs Upstream Helm charts/manifests which will be easy to maintain and upgrade.

lsviben:
uses embedded upstream gatekeeper CRDs for gatekeeper installation on user clusters, instead of the current way of having them in code. This will make them easier to mantain

Fixes #7971 

```release-note
NONE
```

### DIFF
--- a/hack/verify-boilerplate.sh
+++ b/hack/verify-boilerplate.sh
@@ -31,7 +31,8 @@ boilerplate \
   -exclude pkg/resources/certificates/triple/triple.go \
   -exclude pkg/resources/etcd/testdata \
   -exclude pkg/ee \
-  -exclude charts/kubermatic-operator/crd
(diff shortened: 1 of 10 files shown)


xrstf (approved):
/approve

PR #8560, opened 2022-01-03
fix nodeport proxy e2e test

PratikDeoghare:
E2E test for nodeport proxy is broken. It did not run because `run_if_changed` is too strict in prow.yaml. This pr fixes it. 

```release-note
NONE
```


xrstf (approved):
/approve

PR #8566, opened 2022-01-04
Move webhooks into standalone Pod

### LINKED ISSUE #7871: Move cluster and seed webhooks from seed/master ctrlmngr to standalone deployments
**User Story**
Currently the webhooks are part of seed and master controller managers, this makes it difficult to develop and test them. Move them to standalone deployments

**Acceptance criteria**
- [ ] Cluster and seed webhooks are standalone deployments

xrstf:
Up until now, we always put webhook logic into our controller-managers. This was done because it was convenient, easy and cheap, but it has severe drawbacks:

* There is some magic in the KKP operator to figure out if Seed validation requests should be sent tothe master-ctrl-mgr or the seed-ctrl-mgr (to handle cases where master/seed are the same cluster).
* Shutting down the controller managers means you can effectively not work with KKP resources anymore.
* Scaling is not possible, because for a number of more or less valid reasons, we do not really recommend scaling up the controller-managers.

For these reasons, this PR extracts the webhook logic from both controller-managers and moves it into the brand spanking new `kubermatic-webhook`. This one is, just like before, deployed on master and seed clusters and is smart enough to handle shared master/seed and separate master/seed clusters.

The seed validation and cluster validation webhooks were also refactored to use ctrl-runtime 0.11's new CustomValidator and CustomDefaulter feature, which allows us to get rid of our own handling logic. Sadly, ctrl-runtime cannot (yet?) handle more complex defaulting, which is the reason why the cluster mutation is still using the old implementation.

Fixes #7871

```release-note
All webhooks have been moved from the controller-managers into a standalone webhook Deployment.
```


pkg/controller/operator/seed/reconciler.go
```diff
@@ -576,28 +608,31 @@ func (r *Reconciler) reconcileServices(ctx context.Context, cfg *kubermaticv1.Ku
...
+		return fmt.Errorf("failed to reconcile validating Validating Webhooks: %w", err)
 	}
 
 	mutatingWebhookCreators := []reconciling.NamedMutatingWebhookConfigurationCreatorGetter{
 		kubermaticseed.ClusterMutatingWebhookConfigurationCreator(cfg, client),
 	}
 
 	if err := reconciling.ReconcileMutatingWebhookConfigurations(ctx, mutatingWebhookCreators, "", client); err != nil {
-		return fmt.Errorf("failed to reconcile mutating AdmissionWebhooks: %w", err)
+		return fmt.Errorf("failed to reconcile mutating Mutating Webhooks: %w", err)
```
embik: Nit: _"[...] mutating Mutating Webhooks"_.

pkg/controller/operator/seed/reconciler.go
```diff
@@ -576,28 +608,31 @@ func (r *Reconciler) reconcileServices(ctx context.Context, cfg *kubermaticv1.Ku
...
+		validatingWebhookCreators = append(
+			validatingWebhookCreators,
+			kubermaticseed.OperatingSystemProfileValidatingWebhookConfigurationCreator(cfg, client),
+			kubermaticseed.OperatingSystemConfigValidatingWebhookConfigurationCreator(cfg, client),
+		)
 	}
 
 	if err := reconciling.ReconcileValidatingWebhookConfigurations(ctx, validatingWebhookCreators, "", client); err != nil {
-		return fmt.Errorf("failed to reconcile validating AdmissionWebhooks: %w", err)
+		return fmt.Errorf("failed to reconcile validating Validating Webhooks: %w", err)
```
embik: Second nit: _"[...] validating Validating Webhooks"_. The previous error was probably fine ("validating AdmissionWebhooks" is fairly accurate).

pkg/ee/cmd/webhook/wrappers.go
```diff
@@ -0,0 +1,46 @@
+//go:build ee
+
+/*
+                  Kubermatic Enterprise Read-Only License
+                         Version 1.0 ("KERO-1.0”)
+                     Copyright © 2020 Kubermatic GmbH
```
embik: Super nit: It's 2022.

pkg/webhook/seed/validation.go
```diff
@@ -25,95 +25,93 @@ import (
...
-		if _, exists := seedsMap[seed.Name]; !exists && op == admissionv1.Delete {
-			return nil
-		}
-		// in case of delete request the seed is empty
-		seed = seedsMap[seed.Name]
+		return fmt.Errorf("failed to list Seeds: %w", err)
 	}
 
-	client, err := v.seedClientGetter(seed)
+	// if isDelete {
```
mate4st: What about this? Do we need it or not?

embik (commented):
Some very minor nits but I think at least the error messages should be fixed :-)

pkg/webhook/seed/validation.go
xrstf: Need it.

PR #8568, opened 2022-01-04
do not take ownership of CA bundle created by our Helm chart

xrstf:
I noticed that during KKP depoloyments or during tests, the KKP operator would seemingly out of nowhere cause restarts to many Deployments like the master-ctrl-mgr or the API. After inspecting the ReplicaSets, it turns out that we increase the resourceVersion on the CABundle ConfigMap.

This is because the seed-operator needs to create the CABundle on the seed, and if master and seed are the same cluster, it will overwrite an annotation and set the owner reference, which will increase the version. And since many Deployments use the CABundle, they are all needlessly restarted.

This PR changes the OwnershipModifier to not modify anything if Helm owns it.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operator/common/util.go
+++ b/pkg/controller/operator/common/util.go
@@ -48,6 +48,10 @@ const (
 	// ManagedByLabel is the label used to identify the resources
 	// created by this controller.
 	ManagedByLabel = "app.kubernetes.io/managed-by"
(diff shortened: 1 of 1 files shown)


xmudrii (approved):
/lgtm
/approve

PR #8575, opened 2022-01-05
Only register scheme and webhooks for OSM if feature is enabled

ahmedwaleedmalik:

Only register scheme and webhooks for OSM if feature is enabled

Fixes #

**Documentation**:

```release-note
NONE
```


ahmedwaleedmalik:
> I don't think this is a good idea. Just because someone disables the feature, shouldn't mean the existing webhooks suddenly fail and people are unable to work with existing OSM resources.

You are right, my assumption was that it will remove those webhooks as well if the feature is disabled. But looking at the code [here](https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/seed/reconciler.go#L589), it won't. 

Thanks @xrstf, closing this PR since this is not required.

PR #8617, opened 2022-01-07
Improve handling of KUBECONFIG in http-prober

ahmedwaleedmalik:
This PR changes the behavior of `http-prober` wait for CRD functionality. Previously, it relied specifically on `KUBECONFIG` env variable for path to kubeconfig. with the current changes it will try to resolve kubeconfig in the following order, based on precedence:

```text
1. PROBER_KUBECONFIG env variable
2. KUBECONFIG env variable
3. -kubeconfig flag
4. In-cluster config if running inside a cluster
5. $HOME/.kube/config if exists.
```

We now need to use in-cluster and external cluster's kube config within the same container. For machine-controller webhooks, previously using `KUBECONFIG` env variable was sufficient since we only required the user cluster's kubeconfig. But now it also needs to load in-cluster config for seed/master cluster. By specifying `KUBECONFIG` at a container level we can't fetch in-cluster config using https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/config/config.go#L59

As a work-around we updated it to have a dedicated env variable `PROBER_KUBECONFIG` to avoid any conflicts if multiple configs are required within the same context.

**NOTE:** This shouldn't break anything for the existing workflows since `KUBECONFIG` env variable still provides the same functionality.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/http-prober/main.go
+++ b/cmd/http-prober/main.go
@@ -39,9 +39,11 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/rest"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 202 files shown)


xrstf (approved):
/approve

PR #8628, opened 2022-01-10
Fix default etcd backups for multiple backup destinations

### LINKED ISSUE #8557: Adjust default cluster backup with multiple backup destinations
**User Story**
When automatic backups are enabled for a Seed (either old backupRestore config is set, or etcdBackupRestore backup destination is set in the Seed config), for each user cluster, a default EtcdBackupConfig is created, which makes backups every 20 minutes.

Seed spec:
``` 
 backupRestore:
    s3BucketName: kkpbackuptest
    s3Endpoint: s3.amazonaws.com
...
  etcdBackupRestore:
    destinations:
      s3:
        bucketName: kkpbackuptest
        endpoint: s3.amazonaws.com
      test:
        bucketName: testbucket-test
        endpoint: bob.com
```

Default EtcdBackup
```
  spec:
...
    keep: 20
    name: default-backups
    schedule: '@every 20m'

```

With the introduction of multiple backup destinations, its not clear anymore which (if any) destination to pick if we have multiple destinations. Even if just 1 destination is set, it could just be a temp thing, and user could add more in some time.  

There are a few options on how to deal with this:

**1. Remove the default backups.**

This is the simplest solution, just remove the creation of default backups and let users create their own backups, with destinations that they want. 

**2. Create default backups only if the old backupRestore config is set.**

This solution means that the default backups will be created just for clusters in Seeds which have the deprecated backupRestore settings set, and will ignore the multiple destination setting. This seems confusing and error-prone

**3. Add a way to mark a destination as default**

Then this destination will be used for the default backups. And if no default destination is set, the default backups are not created.

But this complicates things because:
- what to do with existing default backups when the default destination changes?
- what when the default destination is deleted?

lsviben:
Fixes the issue where the default etcd backup, which is created for each cluster if automatic etcd backups are enable, didnt work when using new multiple backup destinations, as it didnt set a destination.

Added a field to the Seed.Spec.EtcdBackupRestore which selects which backup destination is default (if any). So that users can choose which destination they want to be default. Also added removal of etcdBackupConfigs if the default destination is removed. Includes validation in the seed validation webhook.

Fixes #8557 

```release-note
Added support for setting a default etcd backup destination when using the new multiple etcd backup destinations feature. The default destination is used for the default EtcdBackupConfig which is created for each user cluster when automatic etcd backups are configured for a Seed.
```


moadqassem (approved):
/approve
/lgtm

xrstf (approved):
/approve

PR #8670, opened 2022-01-10
Check usercluster compatibility prior to KKP upgrades

### LINKED ISSUE #7314: Installer: check if there are user clusters running unsupported Kubernetes versions before upgrading KKP
**User Story**

Historically, we've been automatically upgrading user clusters running an unsupported Kubernetes version to a supported one. For example, we've upgraded 1.17 to 1.18 user clusters automatically in KKP 2.17: https://github.com/kubermatic/kubermatic/blob/8ec6f46ffbf114ca10fedb3ccfc5cd9cc366d11f/pkg/controller/operator/common/defaults.go#L215-L222

This can cause many problems:

* Automatically upgrading clusters can affect the workload availability. The upgrade should be done explicitly at the user request
* Kubernetes has started deprecating and removing a lot of old APIs. If the user is still using one of the deprecated APIs, and the API gets removed in the target Kubernetes version, the user cluster could become entirely broken

Ideally, users should familiarize themselves with the changes in the target Kubernetes version (e.g. by reading the changelog) before they upgrade the user clusters.

We've briefly discussed this in the past, and we think that it might make sense to replace the automatic upgrades with a check in the installer that would reject upgrading KKP before upgrading unsupported user clusters.

**Acceptance criteria**

* There is a mechanism in the kubermatic-installer to verify if all user clusters are running a supported Kubernetes version
* The KKP upgrade should be forbidden if there is a user cluster running an unsupported version

xrstf:
This adds another stage to the installation process, a "preflight" stage. Previously we only validated the given configuration files, now we also probe the target cluster and ensure that it's good to go. The first check for this preflight stage is to ensure that all userclusters are still covered by the version configuration.

This check runs on the **master** cluster, even though userclusters of course live on seed clusters. But

1. when doing `kubermatic-installer deploy kubermatic-seed`, the installer does not know which Seed is the current one (this only applies to shared master+seed setups)
2. The installer doesn't actually install any KKP stuff into seeds, the KKP operator is responsible for this. And it does that from the master cluster. It therefore makes more sense to validate the master (and collect all Seeds there and check them) and not the seeds.

Fixes #7314

```release-note
KKP Installer will now check that all userclusters match the given versioning configuration prior to upgrading KKP. This ensures no userclusters are suddenly "orphaned" and incompatible with KKP.
```

### DIFF
--- a/cmd/kubermatic-installer/cmd_deploy.go
+++ b/cmd/kubermatic-installer/cmd_deploy.go
@@ -303,7 +303,7 @@ func DeployAction(logger *logrus.Logger, versions kubermaticversion.Versions) cl
 		opt.SeedsGetter = seedsGetter
 		opt.SeedClientGetter = provider.SeedClientGetterFactory(seedKubeconfigGetter)
 
-		logger.Info("🩺 Validating existing installation…")
+		logger.Info("🚦 Validating existing installation…")
 
 		if errs := kubermaticStack.ValidateState(appContext, opt); errs != nil {
 			logger.Error("⛔ Cannot proceed with the installation:")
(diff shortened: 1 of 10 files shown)


ahmedwaleedmalik (approved):
/lgtm
/approve

PR #8672, opened 2022-01-10
Prevent KKP PDBs from blocking node rotations

### LINKED ISSUE #7110: Impossible to evict kubermatic-master-controller-manager
error when evicting pod "kubermatic-master-controller-manager-c5896967c-7wcxt" (will retry after 5s): 
Cannot evict pod as it would violate the pod's disruption budget.
evicting pod kubermatic/kubermatic-master-controller-manager-c5896967c-7wcxt

The pod disruption budget does not tolerate any disruptions:
```
kubectl get poddisruptionbudgets.policy -n kubermatic kubermatic-master-controller-manager
NAME                                   MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
kubermatic-master-controller-manager   1               N/A               0                     20d
```

### LINKED ISSUE #8068: Improve the usage of PDB for Kubermatic core components
**User Story**
With KKP, we deliver some core components like seed-controller-manager (https://github.com/kubermatic/kubermatic/blob/master/charts/kubermatic/templates/kubermatic-controller-manager-dep.yaml) with 1 replica and at the same time we are defining PodDisruptionBudget `maxUnavailable: 1`.

This leads to situations that it's not possible to maintain the nodes where these components are running because during drain, you will receive error like this:
```
error when evicting pods/"kubermatic-master-controller-manager-7ffb788c69-q8xrg" -n "kubermatic" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
error when evicting pods/"kubermatic-seed-controller-manager-7ccd4bb846-n9htb" -n "kubermatic" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

Some of the workaround for this can be to either use `--disable-eviction=true` with drain command or update the number of replicas, drain and update replicas again back to 1.

**Acceptance criteria**
We should either document how properly maintain the KKP in the current deployment setup or consider whether we actually need to keep the PDB in place.
Any outcome which would improve the experience for end-users is welcome.

Extra question - what are the potential blockers of having all controllers and core components with 2 replicas?

Thank you!

xrstf:
With `replicas=1` (our default), a PDB with `minAvailable=1` will block node rotations. This PR adds a simple fix for that.

The underlying reason why we default to just 1 replica is that our webhooks are part of the controller-managers and only the leader starts the webhook server. That's why in #5616 we reduced the default to 1 replica.

Once #8566 is merged, the webhook will live in its own pod and the controller-managers can safely be scaled up by default. Given that users can still choose to use just 1 replica means that even then we still have to make the PDB work in that usecase. Hence why I applied this fix to the API/UI deployments as well, which were never part of the webhook-leader problem.

Fixes #8068, fixes #7110

```release-note
Fix PodDisruptionBudgets for master/seed-controller-manager blocking node rotations.
```

### DIFF
--- a/pkg/controller/operator/master/resources/kubermatic/api.go
+++ b/pkg/controller/operator/master/resources/kubermatic/api.go
@@ -166,7 +166,13 @@ func APIPDBCreator(cfg *operatorv1alpha1.KubermaticConfiguration) reconciling.Na
 
 	return func() (string, reconciling.PodDisruptionBudgetCreator) {
 		return name, func(pdb *policyv1beta1.PodDisruptionBudget) (*policyv1beta1.PodDisruptionBudget, error) {
+			// To prevent the PDB from blocking node rotations, we accept
+			// 0 minAvailable if the replica count is only 1.
+			// NB: The cfg is defaulted, so Replicas==nil cannot happen.
 			min := intstr.FromInt(1)
+			if cfg.Spec.API.Replicas != nil && *cfg.Spec.API.Replicas < 2 {
+				min = intstr.FromInt(0)
+			}
 
 			pdb.Spec.MinAvailable = &min
 			pdb.Spec.Selector = &metav1.LabelSelector{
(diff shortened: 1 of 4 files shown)


xmudrii (approved):
/lgtm
/approve

PR #8673, opened 2022-01-11
Remove metering Deployment/CronJob when feature is disabled

### LINKED ISSUE #7828: Disabling metering does not stop the metering pod from running and doing reports
**User Story**
Disabling metering does not stop the metering pod from running and doing reports. Expected behaviour would be that the metering stops.

xrstf:
This ensures no active components remain when metering is disabled. For simplicity (no, not because I'm lazy) we keep the metadata-only elements like RBAC in case the admin re-enables the feature in the future.

Fixes #7828

```release-note
Fix disabling metering not having any effect.
```

### DIFF
--- a/pkg/ee/metering/reconcile.go
+++ b/pkg/ee/metering/reconcile.go
@@ -35,6 +35,10 @@ import (
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 	"k8c.io/kubermatic/v2/pkg/resources/registry"
 
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/ee/metering/reconcile.go
```diff
@@ -83,3 +91,31 @@ func ReconcileMeteringResources(ctx context.Context, client ctrlruntimeclient.Cl
...
+
+	key.Name = meteringCronJobWeeklyName
+	if err := cleanupResource(ctx, client, key, &batchv1beta1.CronJob{}); err != nil {
+		return fmt.Errorf("failed to cleanup metering CronJob: %v", err)
+	}
+
+	return nil
+}
+
+func cleanupResource(ctx context.Context, client ctrlruntimeclient.Client, key types.NamespacedName, obj ctrlruntimeclient.Object) error {
```
moadqassem, line 111: So for the RBAC stuff I would leave it, but what about the secret and configurations? Such as the bucket endpoint, aws/minio credentials?

xrstf: I would also leave them. 😁

moadqassem (approved):
/approve
/lgtm

PR #8683, opened 2022-01-11
Recocile aks irrespective of state

imharshita:

**What does this PR do / Why do we need it**: Reconcile azure kubernetes cluster irrespective of cluster state.
This change is necessary in order to get updated kubeconfig for an already imported aks cluster, in case the cluster is starting from a stopped state or it becomes healthy after a failed state.
Since these state changes doesnt affect the cluster crd, so reconcile  requeue time needs to be set

Fixes #

```release-note
none
```


xrstf:
Why is this necessary? Can you please provide a description for this change?

zreigz:
> Why is this necessary? Can you please provide a description for this change?

Do you mean release note?

imharshita:
> Why is this necessary? Can you please provide a description for this change?

Done

xrstf:
Thank you!

PR #8687, opened 2022-01-11
Fix metric names for MinIO Grafana dashboard

dakraus:
Because the metrics exposed by MinIO were refactored as part of pull
request [8003](https://github.com/minio/minio/pull/8003), the Grafana dashboard for MinIO did no longer
display any metrics.

Therefore the dashboard was updated to use the new metrics as outlined
in their [migration guide](https://github.com/Praveenrajmani/minio/blob/9ff7af2dfbc14d971fd3ce8fd02ae32b4252e6d4/docs/metrics/prometheus/README.md#migration-guide-for-the-new-set-of-metrics). Since we run a standalone MinIO server, the
panel 'Network Traffic' only takes the values for the current server
instance into account.

```release-note
Fix Grafana dashboard for MinIO to display the correct Prometheus metrics
```

### DIFF
--- a/charts/monitoring/grafana/dashboards/minio/overview.json
+++ b/charts/monitoring/grafana/dashboards/minio/overview.json
@@ -68,7 +68,7 @@
       "pluginVersion": "6.5.2",
       "targets": [
         {
-          "expr": "minio_disk_storage_used_bytes / minio_disk_storage_total_bytes * 100",
+          "expr": "disk_storage_used{app=\"minio\"} / disk_storage_total{app=\"minio\"} * 100",
           "instant": true,
           "refId": "A"
         }
@@ -122,7 +122,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "minio_disk_storage_available_bytes",
+          "expr": "disk_storage_available{app=\"minio\"}",
           "interval": "$interval",
           "legendFormat": "available",
           "refId": "A"
@@ -232,12 +232,12 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "rate(minio_network_sent_bytes_total[$interval])",
+          "expr": "rate(s3_tx_bytes_total{app=\"minio\"}[$interval])",
           "legendFormat": "sent",
           "refId": "A"
         },
         {
-          "expr": "rate(minio_network_received_bytes_total[$interval])",
+          "expr": "rate(s3_rx_bytes_total{app=\"minio\"}[$interval])",
           "legendFormat": "received",
           "refId": "B"
         }
--- a/charts/monitoring/grafana/Chart.yaml
+++ b/charts/monitoring/grafana/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: grafana
-version: 1.5.6
+version: 1.5.7
 appVersion: 8.1.2
 description: Grafana for Kubermatic
 keywords:


ahmedwaleedmalik (approved):
Thanks for fixing it, seems like in the current state the grafana dashboard for minio is broken. Then it makes sense to ship this fix with KKP 1.19. WDYT @dakraus?

/lgtm
/assign @wurbanski 

@wurbanski hopefully you can take a look at this PR as well. IIRC you shared a change log for updates in charts somewhere. This change, if merged, can be included there.

dakraus:
@ahmedwaleedmalik sounds good. I intentionally did not add this PR to the milestone to not disturb the preparation of KKP 2.19. But you are right, the dashboard is currently broken, so it makes sense to ship this asap.

xrstf:
Lovely. Could you please extend the queries to make them a bit more specific, now that the `minio_` prefix has been removed? The full metric looks like

> s3_tx_bytes_total{app="minio", instance="10.44.29.44:9000", job="pods", namespace="minio", pod="minio-9db6db5fc-ndnmv", pod_template_hash="9db6db5fc"}

So maybe `...{app="minio"}` is safer?

dakraus:
@xrstf I've added a label matcher to all expressions as suggested. FYI: Since MinIO [RELEASE.2021-01-30T00-20-58Z](https://github.com/minio/minio/releases/tag/RELEASE.2021-01-30T00-20-58Z) there are two additional Prometheus endpoints (`minio/v2/metrics/cluster` and `minio/v2/metrics/node`).

charts/monitoring/grafana/dashboards/minio/overview.json
```diff
@@ -232,12 +232,12 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "rate(minio_network_sent_bytes_total[$interval])",
+          "expr": "rate(s3_tx_bytes_total{app=\"minio\"}[$interval])",
```
SimonTheLeg, line 235: for me something does not seem right with this part of the change. What I mean by that is: yes it is fixing our dashboard, but `s3_tx...` and `s3_rx..` are legacy metrics that are deprecated and should not be used anymore. `network_sent_bytes_total` and `network_received_bytes_total` are the current ones. So we are somehow moving from the current metrics to the old ones, which does not seem right to me 🤔 


This can be quickly verified by doing

```bash
export MINIO_PROMETHEUS_AUTH_TYPE=public
./minio server /tmp/data


# using the current endpoint, will yield no result
curl -s http://192.168.0.61:9000/minio/v2/metrics/cluster | grep rx 

# using the legacy endpoint, will result in the metric
❯ curl -s http://192.168.0.61:9000/minio/prometheus/metrics | grep rx
internode_rx_bytes_total 0
s3_rx_bytes_total 0
```

I am also a bit confused about the old change as from the looks of it our scraping path has not been changed in the past 2 years https://github.com/kubermatic/kubermatic/blob/98a76b89094ed3f53c68ba87485259577f6914b3/charts/minio/templates/deployment.yaml#L32
Has the original ever worked?

All in all this PR is fine in the sense that it does fix the dashboard, but I am wondering if we have a deeper issue in how we scrape minio, that has impact on this change.

dakraus: @SimonTheLeg yes, you are right. As mentioned in my [previous comment](https://github.com/kubermatic/kubermatic/pull/8687#issuecomment-1012025667), I also discovered that we use the "legacy" metrics endpoint of MinIO.

However we are not moving from the current metrics to the old ones: If you take a look at the [migration guide](https://github.com/Praveenrajmani/minio/blob/9ff7af2dfbc14d971fd3ce8fd02ae32b4252e6d4/docs/metrics/prometheus/README.md#migration-guide-for-the-new-set-of-metrics) the metric `minio_network_sent_bytes_total` was indeed renamed to `s3_tx_bytes_total` (or split into `s3_tx_bytes_total` and `internode_tx_bytes_total`) and is not part of the currently [exposed metrics](https://github.com/minio/minio/blob/RELEASE.2022-01-08T03-11-54Z/docs/metrics/prometheus/list.md).

I agree with you that we should use the current metrics endpoints (`minio/v2/metrics/cluster` and `minio/v2/metrics/node`). I don't have a strong opinion when we want to implement this change: I can take a look at the metrics exposed by the v2 metrics endpoints and implement the necessary changes right away or I can also create a follow-up issue to discuss this in more depth. What do you think?

xrstf: I would vote that we merge this and have any other discussion regarding changing the metrics endpoint in another PR/issue. @wurbanski is working hard on making our charts obsolete anyway, so just by virtue of using upstream we might also then use the new endpoint (and probably have to fix the dashboard again).

wurbanski:
> @wurbanski hopefully you can take a look at this PR as well. IIRC you shared a change log for updates in charts somewhere. This change, if merged, can be included there.

I don't think this change requires a note in the upgrade docs (no action required), but I think a release note in the PR would be sufficient (_something_ is changing).

xrstf (approved):
/approve

dakraus:
Test is failing because PR #8725 was merged bumping the Grafana Helm chart to the same version. I will add an additional commit.

PR #8710, opened 2022-01-12
fix MLA logging health status

### LINKED ISSUE #8691: MLA Logging Health status not updated for `Down Status`
### What happened?
Refer to #8635 
Scenario:
Cluster healthy created on AWS with Logging. 
Update the ds promtail in user cluster (refer to scenario #8635) to have a non healthy promtail. 

```
Update the ds promtail is the user cluster with an invalid value, like config file
 spec:
      containers:
      - args:
        - -config.file=/etc/promtail/promtail.yaml2
```

```
% k get ds
NAME       DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
promtail   1         1         0       1            0           <none>          3h57m

```
-> Status for Logging should be red, but it remains green.

And status in `cluster CR` is also wrong -> issue in the back-end.

### What should happen?
Logging Status should be red.

### Environment

- KKP Version: v2.19.0-alpha.3
- Master/seed cluster provider: AWS

sankalp-r:
Adding `DaemonSet` to resource-watch list, so that it gets reconciled.
As `Promtail` is a daemonset and if its gets down, the health-status gets updated after reconciliation.

Fixes #8691 

```release-note
None
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/controller.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/controller.go
@@ -179,6 +179,7 @@ func Add(
 		&appsv1.Deployment{},
 		&v1beta1.PodDisruptionBudget{},
 		&networkingv1.NetworkPolicy{},
+		&appsv1.DaemonSet{},
 	}
 
 	// Avoid getting triggered by the leader lease AKA: If the annotation exists AND changed on


hdurand0710 (approved):

moadqassem (approved):
/approve
/lgtm

PR #8725, opened 2022-01-13
Update Grafana dashboards

### LINKED ISSUE #8025: Outdated metric types in dashboard "Kubelets Overview"
**User Story**
Several panels in the "Kubernetes / Kubelets Overview" are based on the assumption that the underlying metric is of type Summary. That's however not the case for any of the supported K8s versions in KKP.

In specific, the rows "Operation Durations (90th Percentiles)" and "Request Durations" do not show any data at all in its panels. I've looked into this and it looks like the issue is that the metrics queried here used to be of type Summary, but have switched to Histograms some time before Kubernetes 1.19.

Screenshot 2021-10-18 at 17 17 11

I've checked this for one metric, the "PLEG Relist Duration" panel. Current PromQL `sum(rate(kubelet_pleg_relist_duration_seconds{instance=~"$node",quantile="0.9",operation_type=~"$operations"}[5m])) by (instance)` is based on the assumption there's a Summary metric that exposes data with the `quantile` label. The label `operation_type` doesn't seem to exist for the metric either.

A working query instead would be:

```
histogram_quantile(0.9, sum(rate(kubelet_pleg_relist_duration_seconds_bucket{instance=~"$node"}[5m])) by (instance, le))
```

This is just an example for one of the panels, but I think all of them have similar issues.

**Acceptance criteria**

All panels in the mentioned rows show data based on valid kubelet metrics.

xrstf:
This PR cleans up a few issues in our Grafana dashboards:

* scraping the kube-apiserver produces a go_info metric, but without namespace/pod name on GKE; those go_info metrics lead to an empty line in our go version table which is now hidden
* the node count was not federated to the seed level Prometheus, so the node count metric on the machine-controller dashboard was broken
* one dashboard still had openshift leftovers
* the kubelet dashboards were broken (#8025)

Fixes #8025

```release-note
NONE
```

### DIFF
--- a/charts/monitoring/grafana/dashboards/kubernetes/kubelets-overview.json
+++ b/charts/monitoring/grafana/dashboards/kubernetes/kubelets-overview.json
@@ -10,6 +10,11 @@
   "panels": [
     {
       "collapsed": false,
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {},
+        "overrides": []
+      },
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -32,12 +37,14 @@
       "description": "",
       "editable": true,
       "fill": 1,
+      "fillGradient": 0,
       "gridPos": {
         "h": 9,
         "w": 8,
         "x": 0,
         "y": 1
       },
+      "hiddenSeries": false,
       "id": 47,
       "legend": {
         "avg": false,
@@ -52,7 +59,11 @@
       "linewidth": 1,
       "links": [],
       "nullPointMode": "null as zero",
+      "options": {
+        "alertThreshold": true
+      },
       "percentage": false,
+      "pluginVersion": "8.1.2",
       "pointradius": 5,
       "points": false,
       "renderer": "flot",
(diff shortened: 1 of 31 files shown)


charts/monitoring/grafana/dashboards/kubernetes/kubelets.json
```diff
@@ -2295,5 +966,5 @@
   "timezone": "",
   "title": "Kubelets",
   "uid": "p4ynSSeik",
-  "version": 6
+  "version": 1
```
embik: I think this should be `7` as it's updating an existing dashboard. Grafana expects the `version` field to be incremented only, I can see this causing problems when loading the dashboard.

embik (changes requested):

charts/monitoring/grafana/dashboards/kubernetes/kubelets-overview.json
```diff
@@ -885,5 +939,5 @@
   "timezone": "",
   "title": "Kubelets Overview",
   "uid": "ZJ1iOSemz",
-  "version": 12
+  "version": 1
```
embik: Same here.

embik (changes requested):

embik (approved):

PR #8729, opened 2022-01-14
Fix reconciling using recreate strategy

### LINKED ISSUE #6912: Reconciliation of coredns PodDisruptionBudget failed when upgrading to KKP 2.17
The following error log was found in `user-controller-manager` after upgrading from KKP 2.16 to KKP 2.17:

```
{"level":"error","time":"2021-04-20T12:44:36.228Z","caller":"resources/controller.go:226","msg":"Reconciling failed","error":"failed to reconcile PodDisruptionBudgets: failed to ensure PodDisruptionBudget kube-system/coredns: failed to create object *v1beta1.PodDisruptionBudget \"kube-system/coredns\": resourceVersion should not be set on objects to be created"}
```

xrstf:
When using Delete+Create to reconcile a resource, we should ensure that metadata like UID/resourceVersion are not set, otherwise the kube-apiserver will reject the operation with something like

> EnsureObject returned an error while none was expected: failed to create object *v1.ConfigMap "default/test": resourceVersion can not be set for Create requests

Fixes #6912

```release-note
NONE
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,6 @@
 **What does this PR do / Why do we need it**:
 
-**Does this PR close any issues?**:
-<!-- optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged -->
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik (approved):
Makes sense to do this when we are re-creating something.

/lgtm
/appove

PR #8735, opened 2022-01-17
allow to enable debug logging of the usercluster-ctrl-mgr

### LINKED ISSUE #8734: Allow to toggle debug logging on user-cluster-ctrl-mgr
It's currently impossible to get debug logging on the user-ctrl-mgr, because the seed-cltr-mgr will never set the `-log-debug` flag and to manually set it, one needs to pause the cluster. But if the cluster is paused, the user-ctrl-mgr will do nothing, so enabling debug logging on it is pointless.

### Acceptance criteria
- [ ] Debug logging can be toggle globally via KubermaticConfiguration (just like we do for the API, master- and seed-ctrl-mgrs).
- [ ] (maybe) There is a controllerOverride thing in place to toggle debug logging just for single clusters, not globally.

xrstf:
This PR makes it possible to enable debug logging for individual userclusters. Since this is much more useful than having a global configuration flag, I would consider this to fully solve #8734. Since this controller-manager is not really a Kubernetes component, I chose to not use the ComponentsOverride stuff for this.

Fixes #8734

```release-note
Add `spec.debugLog` field to `Cluster` objects to toggle the verbose log on the usercluster-controller-manager.
```

### DIFF
--- a/pkg/crd/kubermatic/v1/cluster.go
+++ b/pkg/crd/kubermatic/v1/cluster.go
@@ -124,6 +124,9 @@ type ClusterSpec struct {
 	// PauseReason is the reason why the cluster is no being managed.
 	PauseReason string `json:"pauseReason,omitempty"`
 
+	// DebugLog enables more verbose logging by KKP's usercluster-controller-manager.
+	DebugLog bool `json:"debugLog,omitempty"`
+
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve
/lgtm

PR #8738, opened 2022-01-17
Fix reconcile loop for vSphere CSI ValidatingWebhookConfiguration

### LINKED ISSUE #8715: external-ccm-migration causes usercluster-controller to get stuck
### What happened?

After migration of user-cluster to external CCM (vsphere), the user-cluster-controller manager pod get stuck with below error.

```
{"level":"error","time":"2022-01-12T16:37:19.139Z","caller":"resources/controller.go:283","msg":"Reconciling failed","error":"failed to reconcile ValidatingWebhookConfigurations: failed to ensure ValidatingWebhookConfiguration /validation.csi.vsphere.vmware.com: failed waiting for the cache to contain our latest changes: timed out waiting for the condition"}
```

But if we check the cluster, it has required webhook config
```
$ k get validatingwebhookconfigurations -A
NAME                                          WEBHOOKS   AGE
validation.csi.vsphere.vmware.com             1          83m
```

user-cluster-controller pod does not pickup new config even after restarting

### What should happen?
CCM migration should successfully complete

### How to reproduce?
Have a user-cluster in vsphere created as part of kkp v2.17. 
Upgrade kkp to v2.18
do a CCM migration via UI

### Environment

- KKP Version: v2.18.4
- Master/seed cluster provider: vsphere

xrstf:
We had a reconcile loop for the webhook, at least because we forgot to set the service port.

> 2022-01-17T15:20:14.767+0100    debug   reconciling/compare.go:62       Object differs from generated one       {"type": "*v1.ValidatingWebhookConfiguration", "namespace": "", "name": "validation.csi.vsphere.vmware.com", "diff": ["Webhooks.slice[0].ClientConfig.Service.Port: <nil pointer> != int32"]}

This error was however invisible, even with enabled debug logging, because we forgot to update the global Logger instance based on the given CLI flags. That is why this PR also changes the seed-ctrl-mgr.

With regards to the other fields (NamespaceSelector, MatchPolicy), I am not 100% sure if we would need to specify them, but our other creators already do and this is the safe approach.

Fixes #8715

```release-note
Fix vSphere clusters getting stuck after CSI migration due to bad ValidatingWebhookConfiguration reconciling.
```

### DIFF
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -89,6 +89,9 @@ func main() {
 	// Set the logger used by sigs.k8s.io/controller-runtime
 	ctrlruntimelog.SetLogger(zapr.NewLogger(rawLog))
 
+	// make sure the logging flags actually affect the global (deprecated) logger instance
+	kubermaticlog.Logger = log
+
 	versions := kubermatic.NewDefaultVersions()
 	cli.Hello(log, "Seed Controller-Manager", logOpts.Debug, &versions)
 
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
Great work.

/lgtm
/approve

PR #8744, opened 2022-01-18
use owner references to cleanup seed-level RBAC instead of manual work

### LINKED ISSUE #7832: Cluster stuck in deletion because user-cluster-controller-manager doesn't have permission to work with Cluster CRs
Cluster stuck in deletion because user cluster controller manager doesn't have permission to work with Cluster CRs, and throws follow errors:
> E0911 08:54:09.147315       1 reflector.go:138] k8s.io/client-go@v12.0.0+incompatible/tools/cache/reflector.go:167: Failed to watch *v1.Cluster: failed to list *v1.Cluster: clusters.kubermatic.k8s.io is forbidden: User "system:serviceaccount:cluster-zgx26shkjx:kubermatic-usercluster-controller-manager" cannot list resource "clusters" in API group "kubermatic.k8s.io" at the cluster scope

## The reason
We have a clusterrolebinding for uccm service account to manage cluster resources: https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/usercluster/rbac.go#L123-L141
And during cluster deletion, if this clusterrolebinding is removed before the uccm deployment in cluster namespace is removed, then the service account doesn’t have permission to work with cluster resources anymore, and throw this kind of errors.

## Potential Fix:
What we can do is that we can add logic here: https://github.com/kubermatic/kubermatic/blob/master/pkg/clusterdeletion/clusterrolebindings.go#L32, and make sure only clean up clusterrolebinding if the cluster namespace is removed

xrstf:
During cleanup, we remove the ClusterRole/Binding too early, leading to the usercluster-ctrl-mgr logging errors (see linked issue). This PR changes the cleanup procedure to use owner references instead. Thankfully, namespaces are not namespaced (^^), so we can set owner refs to the cluster namespace. This means that the RBAC stuff is deleted only after the cluster namespace is gone, which should be when the usercluster-ctrl-mgr pod is also already gone.

Fixes #7832

```release-note
Fix clusters being occasionally stuck in deletion because seed-level RBAC resources were deleted too early.
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -2340,7 +2340,8 @@ const (
 	SeedUserProjectBindingCleanupFinalizer = "kubermatic.io/cleanup-seed-user-project-bindings"
 	// SeedUserCleanupFinalizer indicates that Kubermatic Users on the seed clusters need cleanup
 	SeedUserCleanupFinalizer = "kubermatic.io/cleanup-seed-users"
-	// ClusterRoleBindingsCleanupFinalizer indicates that the cluster ClusterRoleBindings on the seed cluster need cleanup
+	// ClusterRoleBindingsCleanupFinalizer indicates that the cluster ClusterRoleBindings on the seed cluster need cleanup.
+	// This finalizer is deprecated and should not be used anymore since we migrated to using owner references for cleanup.
 	ClusterRoleBindingsCleanupFinalizer = "kubermatic.io/cleanup-cluster-role-bindings"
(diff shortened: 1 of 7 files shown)


ahmedwaleedmalik (approved):
/lgtm
/approve

ahmedwaleedmalik (approved):
/approve

xrstf:

vCenter is down. And now this is actually the only error in the integration tests ^^

multi-io:
Wouldn't it be even easier to just use a RoleBinding in the cluster namespace instead of a ClusterRoleBinding with an owner reference to the cluster namespace?

PR #8745, opened 2022-01-18
Update flatcar controller to support removal of resources

### LINKED ISSUE #8708: flatcar update operator is not removed when no flatcar node in the cluster
https://test-env-kkp.ee.lab.kubermatic.io/
KKP version: v2.19.0-alpha.4
Refer to: #8664

### What happened?

- Cluster creation with flatcar.

- Create a new node deployment with a different OS in the cluster and delete the initial node deployment.

- Check if flatcar update operator does not exist after the procedure.

---------------------
After md with flatcar
```
% k get deploy -n kube-system
NAME                            READY   UP-TO-DATE   AVAILABLE   AGE
calico-kube-controllers         1/1     1            1           9m11s
coredns                         2/2     2            2           9m27s
flatcar-linux-update-operator   1/1     1            1           6m48s
openvpn-client                  1/1     1            1           9m13s
```

- Now add the md with `Ubuntu`

- Then remove the md Flatcar

** Issue found **
Deployment is still there:
```
% k get deploy
NAME                            READY   UP-TO-DATE   AVAILABLE   AGE
calico-kube-controllers         1/1     1            1           5h1m
coredns                         2/2     2            2           5h1m
flatcar-linux-update-operator   0/1     1            0           4h59m
openvpn-client                  1/1     1            1           5h1m
```
pod is `Pending`
```
% k get pod
NAME                                             READY   STATUS    RESTARTS   AGE
...
flatcar-linux-update-operator-7bcc7dfb69-qmqs8   0/1     Pending   0          166m
```
with status
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2022-01-12T10:54:32Z"
    message: '0/1 nodes are available: 1 node(s) didn''t match Pod''s node affinity/selector.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
```

### What should happen?
flatcar update operator should be removed.

### Environment

- KKP Version: v2.19.0-alpha.4
- Master/seed cluster provider: AWS

embik:
This PR updates the flatcar controller in `user-cluster-controller-manager` to remove `flatcar-update-{operator,agent}` related resources when no Flatcar node is in the cluster. That way, no leftover resources linger when a Flatcar `MachineDeployment` is added and removed later on.

Fixes #8708

- The predicate has been removed because we want to check the conditions against all nodes, not only the one caught by the controller loop. Node changes merely act as a trigger now.
- The check if a node is unschedulable has been removed as both DaemonSet and Deployment ignore ANY taints. Flatcar nodes that are marked as unschedulable can therefore still run the update-agent.
- The node selector has been removed from the Deployment because the update-operator does not _require_ to run on Flatcar nodes (it does not interact with the underlying node; it just coordinates the agent actions via the Kubernetes API) and blocks draining the last remaining Flatcar Node of a `MachineDeployment` (as any taint is ignored, that includes the taint repelling Pods from unschedulable Nodes). This way, the update-operator can be drained and started on another Node in the cluster, and then be removed when the last Flatcar Node is successfully removed.
- The "catch-all" tolerations have been removed from the `flatcar-update-operator` Deployment. Putting those on Deployments is problematic because it prevents node drains from succeeding. Other, similar resources like `calico-kube-controllers` don't have those tolerations either.

**Documentation**:

```release-note
Unused flatcar update resources are removed when no Flatcar `Nodes` are in a user cluster
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
+++ b/pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
@@ -83,7 +83,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	if err := r.List(ctx, &nodeList,
 		ctrlruntimeclient.MatchingLabels{nodelabelerapi.DistributionLabelKey: nodelabelerapi.FlatcarLabelValue},
 	); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to list nodes: %v", err)
+		return reconcile.Result{}, fmt.Errorf("failed to list nodes: %w", err)
 	}
 
 	// filter out any Flatcar nodes that are already being deleted
@@ -96,11 +96,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 
 	if len(nodes) == 0 {
 		if err := r.cleanupUpdateOperatorResources(ctx); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to clean up UpdateOperator resources: %v", err)
+			return reconcile.Result{}, fmt.Errorf("failed to clean up UpdateOperator resources: %w", err)
 		}
 	} else {
 		if err := r.reconcileUpdateOperatorResources(ctx); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to reconcile the UpdateOperator resources: %v", err)
+			return reconcile.Result{}, fmt.Errorf("failed to reconcile the UpdateOperator resources: %w", err)
 		}
 	}
 
--- a/pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
+++ b/pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
@@ -25,7 +25,6 @@ import (
 	"k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/flatcar/resources"
 	nodelabelerapi "k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/node-labeler/api"
 	controllerutil "k8c.io/kubermatic/v2/pkg/controller/util"
-	predicateutil "k8c.io/kubermatic/v2/pkg/controller/util/predicate"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 	"k8c.io/kubermatic/v2/pkg/resources/registry"
 
@@ -63,12 +62,7 @@ func Add(mgr manager.Manager, overwriteRegistry string, updateWindow kubermaticv
 		return err
 	}
 
-	predicate := predicateutil.Factory(func(o ctrlruntimeclient.Object) bool {
-		node := o.(*corev1.Node)
-		return o.GetLabels()[nodelabelerapi.DistributionLabelKey] == nodelabelerapi.FlatcarLabelValue && !node.Spec.Unschedulable
-	})
-
-	return c.Watch(&source.Kind{Type: &corev1.Node{}}, controllerutil.EnqueueConst(""), predicate)
+	return c.Watch(&source.Kind{Type: &corev1.Node{}}, controllerutil.EnqueueConst(""))
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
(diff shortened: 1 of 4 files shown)


pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
```diff
@@ -81,13 +75,30 @@ func (r *Reconciler) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
...
+	); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to list nodes: %v", err)
+	}
+
+	if len(nodes.Items) == 0 {
+		if err := r.cleanupUpdateOperatorResources(ctx); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to clean up UpdateOperator resources: %v", err)
+		}
+	} else {
+		if err := r.reconcileUpdateOperatorResources(ctx); err != nil {
```
ahmedwaleedmalik, line 102: I'd recommend checking here that there is at least one node that is ready, schedulable, or at least doesn't have `metadata.deletionTimestamp`. Because if there are no nodes in the ready state then there is no point in installing the update operator.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
```diff
@@ -64,12 +63,7 @@ func Add(mgr manager.Manager, overwriteRegistry string, updateWindow kubermaticv
 		return err
 	}
 
-	predicate := predicateutil.Factory(func(o ctrlruntimeclient.Object) bool {
-		node := o.(*corev1.Node)
-		return o.GetLabels()[nodelabelerapi.DistributionLabelKey] == nodelabelerapi.FlatcarLabelValue && !node.Spec.Unschedulable
-	})
-
-	return c.Watch(&source.Kind{Type: &corev1.Node{}}, controllerutil.EnqueueConst(""), predicate)
```
ahmedwaleedmalik: Umm, I am not a big fan of removing the predicate all together. Since that would result in a lot of redundant reconciliations.

For reate and delete events it's fine because they don't occur much often. But update is something that can happen quite frequently. I think for updates, we can still keep this predicate or at the very least keep `o.GetLabels()[nodelabelerapi.DistributionLabelKey] == nodelabelerapi.FlatcarLabelValue`

ahmedwaleedmalik (changes requested):

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
embik: What happens when the `user-cluster-controller-manager` is down while the last Flatcar node is removed? Then the removal logic will be never be triggered as when it starts back up, the predicate filters out all of the existing nodes.

Given that we only do `LIST` and `GET` within the reconcile loop and that will work with the cache afaik, the redundant reconciliations are not a problem IMHO.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
embik: Re: schedulable, from the PR notes:

> The check if a node is unschedulable has been removed as both DaemonSet and Deployment ignore ANY taints. Flatcar nodes that are marked as unschedulable can therefore still run the update-agent.

On the deletion timestamp, I agree. Let me add that.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
ahmedwaleedmalik: When `user-cluster-controller-manager` will start again, it will re-queue existing nodes and that should be sufficient :) 

> Given that we only do LIST and GET within the reconcile loop and that will work with the cache afaik, the redundant reconciliations are not a problem IMHO.

For every update on nodes whether it's flatcar or not, you'll be triggering a reconcile loop. Doesn't sound logically correct to me at least.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
ahmedwaleedmalik: > The check if a node is unschedulable has been removed as both DaemonSet and Deployment ignore ANY taints

But the pod will never be created in this case right? So IMO it doesn't make sense to create deployment/daemonset knowing that the pod is not schedulable.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
embik: > When user-cluster-controller-manager will start again, it will re-queue existing nodes and that should be sufficient :)

What I mean is the scenario that this PR is trying to address: The last Flatcar node is removed from the cluster while `user-cluster-controller-manager` is down. Since the node is gone after being removed, `user-cluster-controller-manager` will not be aware of it when starting again, and the predicate will keep the removal logic from being triggered.

embik: As per our chat, I kept a predicate in but limited the label check for `UPDATE` events. `CREATE` and `DELETE` events are triggering the reconcile checks unconditionally. Works with starting up `user-cluster-controller-manager` after removing all Flatcar nodes.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
embik: We discussed this, it's important to not ignore `unschedulable` Nodes because with a single Flatcar Node, it could be drained and rebooted by the update-operator/agent. That would make the node temporarily unschedulable and trigger the `user-cluster-controller-manager` to remove the operator and agent, something we don't want in that scenario.

PR #8746, opened 2022-01-18
make k8s dashboard version depend on usercluster version

### LINKED ISSUE #8196: Serve different k8s/dashboard versions depending on the control plane that is used?
**What would you like to be added? / User Story**: We have to find a way to avoid errors like below:

Zrzut ekranu 2021-09-8 o 15 03 28

It is caused by the version mismatch as KKP offers multiple versions of Kubernetes but just a single Kubernetes Dashboard version.

xrstf:
The dashboard depends on the cluster version. This PR makes it dynamic, even though there currently is only 1 realistic useful release (2.4.0). KKP doesn't support anything <1.20 and there is no dashboard release with 1.22+ support yet. So in the end, the dynamic version is currently always the same :D

Fixes #8196

```release-note
The version of the installed Kubernetes dashboard now depends on the usercluster version.
```

### DIFF
--- a/pkg/resources/kubernetes-dashboard/deployment.go
+++ b/pkg/resources/kubernetes-dashboard/deployment.go
@@ -23,6 +23,7 @@ import (
 	"k8c.io/kubermatic/v2/pkg/resources"
 	"k8c.io/kubermatic/v2/pkg/resources/apiserver"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	"k8c.io/kubermatic/v2/pkg/semver"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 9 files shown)


maciaszczykm:
Thanks! This is good base for handling it better even if there is only one supported k8s/dashboard version at the moment. I think we will release a new version soon, we wanted to have some time after moving from Gulp to Makefile to be able to test everything.

cc @floreks 

/lgtm

floreks:
Awesome! We will release a new version soon to officially support new K8S versions. This is a good base to handle it.

xrstf:

vCenter is down.

pkg/resources/kubernetes-dashboard/deployment.go
```diff
@@ -166,3 +177,12 @@ func getVolumes() []corev1.Volume {
...
 	}
 }
+
+func getDashboardVersion(clusterVersion semver.Semver) (string, error) {
+	// check the GitHub releases for find compat info on the dashboard:
+	// https://github.com/kubernetes/dashboard/releases
+
+	// v2.4.0 is the latest version and supports Kubernetes 1.20/1.21;
+	// there is no official release to support 1.22+ yet.
+	return "v2.4.0", nil
```
floreks: @xrstf FYI: We have released yesterday `v2.5.0`. It can be used for the `v1.22` and `v1.23` Kubernetes clusters.

xrstf: https://github.com/kubernetes/dashboard/releases/tag/v2.5.0 says it's for 1.23 only. Which one is right? ^^

xrstf:
@floreks Now this should be good to go :)

PR #8760, opened 2022-01-19
make error message more readable

xrstf:
This prevents the ugly

> {"level":"error","time":"2022-01-19T16:46:38.483Z","logger":"controller.kubermatic_mla_controller","caller":"controller/controller.go:317","msg":"Reconciler error","name":"test-terraform-nrcch","namespace":"","error":"unable to add grafana user : unable to decode response : %!w(<nil>)"}

I am not sure what exactly happens in the `ensureGrafanaUser` function at the beginning, but this change should be safe.

```release-note
NONE
```


xrstf:

vCenter is down.

PR #8762, opened 2022-01-20
Fix Konnectivity/OpenVPN switching with Tunneling expose strategy

rastislavs:
Fixes switching between Konnectivity and OpenVPN with Tunneling expose strategy by adding a config hash annotation to the envoy-agent daemonset, to ensure that the envoy-agent is restarted when the config changes.

Also fixes the envoy config when Konnectivity is enabled - it that case the config should not contain the listener for openvpn-server (before this change, it would contain it with the bind port == 0, which is not valid - even though it is harmless).

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -18,6 +18,7 @@ package resources
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"net"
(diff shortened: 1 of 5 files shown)


moadqassem (approved):
/approve
/lgtm

rastislavs:

vCenter is down.

xrstf:

vCenter is offline.

xrstf:

vCenter is down

PR #8780, opened 2022-01-25
change debug level for external cluster

### LINKED ISSUE #8732: Improve error handling when external cluster is deleted on provider side
Change controller to handle the situation when the external cluster was deleted on the provider side. Right now controller throws a lot of errors. We can handle this by events and warning instead of errors

zreigz:
**What does this PR do / Why do we need it**: change debug level for the external cluster

Fixes #8732

```release-note
NONE
```


xrstf:
/hold

temporarily

PR #8782, opened 2022-01-25
Drop unused test values from cert-manager

michalvanco:
Support for `clusterIssuers` was dropped from the helm chart thus it does not make sense to keep those in test values.

As you can see from the output, there is nothing being produced based on these values.

**Special notes for your reviewer**: N/A

**Documentation**: N/A

```release-note
NONE
```

### DIFF
--- a/charts/cert-manager/test/default.yaml
+++ b/charts/cert-manager/test/default.yaml
@@ -13,9 +13,3 @@
 # limitations under the License.
 
 cert-manager:
-  clusterIssuers:
-    letsencrypt-prod:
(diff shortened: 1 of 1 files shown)


wurbanski:
Hi, this is correct, but CI wants us to bump the chart version... @xrstf does it make sense to ignore the tests for determining if the chart has changed?

xrstf (approved):
/approve

xrstf:
/retest

Prow, hello? Merge pls?

xrstf:
/override pre-kubermatic-verify

just the chart version check

PR #8783, opened 2022-01-26
Update codebase to use kubermatic.k8c.io

xrstf:
## What does this PR do / Why do we need it

This PR rewrites the entire repository to use the new API group `kubermatic.k8c.io`. This means

* updating all import statements
* rewriting the code to use the new types (we renamed and moved some fields around)
* updating all example files, testdata
* updating all scripts

https://github.com/kubermatic/dashboard/pull/4131 migrates the dashboard accordingly.

### `status` subresource

Some of our resources have a `status` subresource. Previously, this wasn't marked as such, so we could update it like any other field in the `spec`, for example.

With the new types, `status` is a real subresource and must be updated separately from the rest of the resource. To that end, this PR updates all places where the status is modified to issue a separate `Patch()` call.

### RBAC

KKP maps users and projects to Kubernetes resources and creates RBAC rules to, for example, only allow users of a project to create a cluster. Within the KKP REST API there is then a distinction being made between admin users (using an admin kube client) and regular users (using an impersonation client).

I chose to _not_ allow regular users to modified the `status` subresource of anything (i.e. `clusters/status` is not in the ClusterRole). Status information is something that only KKP should ever touch and regular users should not be able to, for example, change the cluster namespace of a Cluster.

For this reason, the admin client is always used to modify the `status` of these resources.

### Actual Changes

* `userSshKeys` is not a required parameter when creating a ClusterTemplate anymore.
* Most occurences of "tenant" in the context of Openstack have been renamed to "project", except for in the API, which would have affected the dashboard.
* The UserWatcher and SettingsWatcher have been refactored. The old implementation used the clients we generated in `pkg/crd/client`, but for the new API groups we do not create such clients anymore. The new implementation uses controller-runtime.
* The KKP API logging has been extended. Instead of Apache-style logs we now always emit JSON via Zap. On the debug level, we also log the response bodies, which makes debugging api-e2e tests soooo much easier, because Swagger likes to just drop response bodies if it recognizes the status code.
* `RequiredEmailDomains` and `RequiredEmailDomain` have been merged into the new `RequiredEmails` in our REST API types.
* `Tenant` and `TenantID` have been removed from the `PublicOpenstackCloudSpec` in our REST API types.
* The Scheme from `pkg/crd` is not used anymore, so all tests have to manually load our kubermaticv1 scheme. No more free lunch.
* In the project-synchronizer, rbac and related controllers a few bugs have been fixed related to shared master/seed setups. On shared systems, the controllers tried to delete things and logged misleading errors.
* Bugfix: The project cleanup is not dependent on the `cleanup` finalizer anymore. Previously, the rbac controller would not clean up a project with a missing finalizer, but instead endlesstry to to reconcile it even after deletion.
* The backup controller was not using UTC for the filenames fof backups, leading to random unit test failures. This has been fixed.

### Now What?

Once this PR would merge, **there is no going back.** The next deployment would expect the development system to already be migrated. Because of this I propose we

* hold this PR
* review it
* once we're all ready to go, someone(TM) takes the kubermatic-installer from my branch, builds it and uses it to migrate dev
* once dev is migrated, we merge this PR

```release-note
Migrate API group `kubermatic.k8s.io` to `kubermatic.k8c.io`.
```


xrstf:
/hold

This must not be merged by accident.

cmd/kubermatic-installer/shared.go
```diff
@@ -26,11 +27,13 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
 	operatorv1alpha1 "k8c.io/kubermatic/v2/pkg/crd/operator/v1alpha1"
```
kron4eg, line 31: I thought we moved operator APIs too

xrstf: We did. But the installer needs to migrate the old config, so it needs the old types.

pkg/controller/operator/common/util.go
```diff
@@ -74,22 +74,27 @@ var (
...
-	for _, feature := range cfg.Spec.FeatureGates.List() {
-		features = append(features, fmt.Sprintf("%s=true", feature))
+// key=value list like "featureA=true,featureB=true,...". The list of feature gates is
+// sorted, so the output of this function is stable.
+func StringifyFeatureGates(cfg *kubermaticv1.KubermaticConfiguration) string {
+	// use a set to ensure that the result is sorted, otherwise reconciling code that
+	// uses this will end up in endless loops
+	features := sets.NewString()
+	for feature, enabled := range cfg.Spec.FeatureGates {
+		if enabled {
```
kron4eg: I know it wasn't in the original function, but what if some feature gate is being explicitly disabled? Looks like we simply skip (and destroy) this information.

pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -331,7 +335,32 @@ func (r *Reconciler) updateCluster(ctx context.Context, cluster *kubermaticv1.Cl
...
+		return nil
+	}
+
+	if !reflect.DeepEqual(oldCluster.Status, cluster.Status) {
+		if err := r.Status().Patch(ctx, cluster, ctrlruntimeclient.MergeFrom(oldCluster)); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
kron4eg, line 363: Looks like other controllers have exactly the same code to update the kubermaticv1.Cluster, maybe let's move it to some helpers

xrstf: No further refactoring please. This is waaaaaay to large of a PR already.

kron4eg (approved):
/approve

kgroschoff (commented):
/approve

kron4eg (approved):

kron4eg (approved):

wurbanski (approved):

xrstf:
/override pre-kubermatic-e2e-nutanix-centos-1.22
/override pre-kubermatic-e2e-nutanix-ubuntu-1.22

Nutanix VPN is not available.

/override pre-kubermatic-e2e-vsphere-ubuntu-1.21
/override pre-kubermatic-e2e-vsphere-ubuntu-1.21-customfolder
/override pre-kubermatic-e2e-vsphere-ubuntu-1.21-datastore-cluster

vCenter is also offline.

xrstf:

Docker rate limitting in effect. No changes since the last api-e2e test success, so I think it's okay to skip the test for once.

xrstf:
/hold cancel

Here we go

PR #8810, opened 2022-01-26
GCP resources reconciled

### LINKED ISSUE #8142: Implement Provider Reconciling for GCP

mlavacca:
Improves the cloud provider for GCP to be capable of periodic reconciling, a feature introduced with #8101. It implements the extended provider functionality that enables periodic reconciling.

Fixes #8142 

**Documentation**:

```release-note
GCP cloud resources are periodically reconciled
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
+++ b/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
@@ -225,7 +225,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 			}
 
 			// remember that we reconciled
-			cluster, err = r.updateCluster(cluster.Name, func(c *kubermaticv1.Cluster) {
+			cluster, err = r.updateClusterStatus(ctx, cluster, func(c *kubermaticv1.Cluster) {
 				now := metav1.Now()
 				c.Status.LastProviderReconciliation = &now
 			})
--- a/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
+++ b/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
@@ -225,7 +225,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 			}
 
 			// remember that we reconciled
-			cluster, err = r.updateClusterStatus(ctx, cluster, func(c *kubermaticv1.Cluster) {
+			cluster, err = r.updateCluster(cluster.Name, func(c *kubermaticv1.Cluster) {
 				now := metav1.Now()
 				c.Status.LastProviderReconciliation = &now
 			})
--- a/pkg/provider/cloud/gcp/firewall_rules.go
+++ b/pkg/provider/cloud/gcp/firewall_rules.go
@@ -22,6 +22,7 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/api/compute/v1"
+
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/crd/kubermatic/v1"
 	kuberneteshelper "k8c.io/kubermatic/v2/pkg/kubernetes"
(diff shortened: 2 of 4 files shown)


pkg/provider/cloud/gcp/firewall_rules.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	firewallService := compute.NewFirewallsService(svc)
+	tag := fmt.Sprintf("kubernetes-cluster-%s", cluster.Name)
+	selfRuleName := fmt.Sprintf("firewall-%s-self", cluster.Name)
+	icmpRuleName := fmt.Sprintf("firewall-%s-icmp", cluster.Name)
+	nodePortRuleName := fmt.Sprintf("firewall-%s-nodeport", cluster.Name)
+
+	// allow traffic within the same cluster
+	if !kuberneteshelper.HasFinalizer(cluster, firewallSelfCleanupFinalizer) {
```
embik: This condition means that once the firewall rules were created, the provider will not attempt to reconcile them on further runs. The new reconcile functionality is supposed to find resources that went missing from the cloud provider and re-create them if needed. I think this needs to be replaced with a query to GCP to check if the firewall service is there.

embik (changes requested):

pkg/controller/seed-controller-manager/cloud/cloud_controller.go
```diff
@@ -225,7 +225,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 			}
 
 			// remember that we reconciled
-			cluster, err = r.updateClusterStatus(ctx, cluster, func(c *kubermaticv1.Cluster) {
+			cluster, err = r.updateCluster(cluster.Name, func(c *kubermaticv1.Cluster) {
```
embik: This does not seem right, why is it being changed? The inline func updates a status field, so it should call `updateClusterStatus`. `updateCluster` is a no-op if it changes `c.Status` fields.

pkg/provider/cloud/gcp/firewall_rules.go
```diff
@@ -0,0 +1,247 @@
...
+		return fmt.Errorf("failed to create firewall rule %s: %w", icmpRuleName, err)
+	}
+
+	newCluster, err := update(cluster.Name, func(cluster *kubermaticv1.Cluster) {
+		kuberneteshelper.AddFinalizer(cluster, firewallICMPCleanupFinalizer)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to add %s finalizer: %w", firewallICMPCleanupFinalizer, err)
+	}
+	*cluster = *newCluster
```
embik: Why, not do `cluster, err := update(...)`? It doesn't seem necessary to do those reassignments later on.

pkg/provider/cloud/gcp/firewall_rules.go
```diff
@@ -0,0 +1,247 @@
...
+	_, err = firewallService.Insert(projectID, &compute.Firewall{
+		Name:    icmpRuleName,
+		Network: cluster.Spec.Cloud.GCP.Network,
+		Allowed: []*compute.FirewallAllowed{
+			{
+				IPProtocol: "icmp",
+			},
+		},
+		TargetTags: []string{tag},
+	}).Do()
```
embik: This (and other parts of the code) will still unconditionally run `firewallService.Insert(...)` without checking if that is even necessary. IMHO, we should eliminate any mutating API calls from the reconcile logic unless it was determined that they are necessary.

The same pattern is applied earlier in the code with this:

```go
firewallObject, err := firewallService.Get(projectID, selfRuleName).Do()
if isHTTPError(err, http.StatusNotFound) {
	// add rule for selfRuleName
}
```

That should be wrapped around any calls to GCP to change something (or similar; potentially it's not enough to check for `http.StatusNotFound`).

embik (changes requested):

embik (approved):
/approve
/lgtm

PR #8822, opened 2022-01-28
Make hack/update-docs.sh work on macOS

embik:

What it says on the title. Makes `hack/update-docs.sh` aware of `gsed` if present on the local system, which allows running the script on macOS. Volume mounts are either performing badly or do not work at all for containerisation solutions on macOS, so running the script without `NO_CONTAINERIZE=1` works well in general, and just needs a few little tweaks to be more cross-platform compatible.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/hack/update-docs.sh
+++ b/hack/update-docs.sh
@@ -28,14 +28,16 @@ dummy=kubermaticNoOmitPlease
 # temporarily create a vendor folder
 go mod vendor
 
+sed="sed"; [ "$(command -v gsed)" ] && sed="gsed"
+
 # remove omitempty tags from structs so that genyaml will not skip fields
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
Tried it on my local machine and it works as expected.

/lgtm
/approve

embik:
/hold

Waiting until vSphere tests are working again. This isn't important enough to bother Christoph with an override request.

xrstf:
Why do we need this, if the script runs in Docker? In 2 months time, someone will come along and remove this again because they don't know why we have macOS-hacks in a golang Docker container.

xrstf:
> I can close the PR and keep this as a local change around if it's not wanted.

This is not about wanting or not wanting. This is about documenting your reasons for introducing hacks again. Nowhere in the PR did it say "happens only on M1 macs", "cannot use Docker on Mac because ............." or even "this hack can be removed when .........".

PR #8829, opened 2022-01-28
Ensure CNI

mate4st:

Before .Spec.CNIPlugin was introduced, canal v3.8 was deployed to clusters referenced as `CanalCNILastUnspecifiedVersion`.
To apply defaulting and ensure the CNIPlugin field is not nil and proper initialized this PR ensure it is set for old clusters during crd migration. In other words, If .Spec.CNIPlugin is nil, it will be set to CanalCNILastUnspecifiedVersion which is equivalent.
This ensures that CNIPlugin is never nil due to the webhooks that apply defaulting for new clusters, therefore we can remove code that handled this specific case.

```release-note
Ensure existing cluster have `.Spec.CNIPlugin` initialized
```

### DIFF
--- a/pkg/webhook/cluster/mutation/mutation.go
+++ b/pkg/webhook/cluster/mutation/mutation.go
@@ -187,14 +187,6 @@ func (h *AdmissionHandler) mutateUpdate(oldCluster, newCluster *kubermaticv1.Clu
 		}
 	}
 
-	// For backward compatibility, if CNIPlugin is not set (possible only for clusters started before KKP v2.18 release),
-	// explicitly set it to the last Canal CNI version supported by KKP before v2.18.
-	if newCluster.Spec.CNIPlugin == nil {
-		newCluster.Spec.CNIPlugin = &kubermaticv1.CNIPluginSettings{
-			Type:    kubermaticv1.CNIPluginTypeCanal,
-			Version: cni.CanalCNILastUnspecifiedVersion,
-		}
-	}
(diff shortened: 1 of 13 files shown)


pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -196,31 +185,6 @@ func (h *AdmissionHandler) mutateUpdate(oldCluster, newCluster *kubermaticv1.Clu
...
-	if newCluster.Spec.CNIPlugin == nil {
-		newCluster.Spec.CNIPlugin = &kubermaticv1.CNIPluginSettings{
-			Type:    kubermaticv1.CNIPluginTypeCanal,
-			Version: cni.CanalCNILastUnspecifiedVersion,
-		}
-	}
-	// This part handles CNI upgrade from unsupported CNI version to the default Canal version.
-	// This upgrade is necessary for k8s versions >= 1.22, where v1beta1 CRDs used in old Canal version (v3.8)
-	// are not supported anymore.
-	if newCluster.Spec.CNIPlugin.Type == kubermaticv1.CNIPluginTypeCanal &&
```
rastislavs: I think we should still keep this part to support upgrade of k8s versions >= 1.22. In KKP 2.19 all clusters that previously had `cluster.Spec.CNIPlugin == nil` now have CNI set to `canal` of version `v3.8`.

WeirdMachine:
/hold
waiting for @xrstf crd migration to merge first

PR #8836, opened 2022-02-01
add e2e tests for cilium

### LINKED ISSUE #8813: E2E test for Cilium
Currently all our E2E tests use Canal.  We won't know if we break usercluster with Cilium CNI. 

Acceptance Criteria:
- There is E2E test to check networking in user cluster is working when Cilium is CNI.

PratikDeoghare:
Currently all our e2e tests run using Canal CNI. If we break clusters with Cilium CNI we won't know. This PR implements e2e tests for Cilium. 

Fixes #8813 

```release-note 
NONE
```


.prow.yaml
```diff
@@ -1055,8 +1055,91 @@ presubmits:
...
+              cpu: 2
+            limits:
+              memory: 6Gi
+  
   #########################################################
+  # Cilium e2e tests
+  #########################################################
+
+  - name: pre-kubermatic-cilium-e2e
+    # run_if_changed: "(.*(cilium|hubble|addon|connectivity-check|defaults|cni|validation|operator|webhook|crd).* | .prow.yaml)"
```
rastislavs: uncomment?

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,460 @@
...
+		}
+	})
+
+	// create a usercluster on aws
+	cluster, err := apicli.CreateAWSCluster(project.ID, seed, userclusterName,
+		secretAccessKey, accessKeyID, utils.KubernetesVersion(),
+		"aws-eu-central-1a", "eu-central-1a", 1, true, &models.CNIPluginSettings{
+			Version: "v1.11",
+			Type:    "cilium",
+		})
```
rastislavs, line 545: Not 100% sure, but should we consider using ebpf proxy mode for the cluster?

pkg/test/e2e/cilium/testdata/connectivity-check.yaml
```diff
@@ -0,0 +1,1039 @@
+# Automatically generated by Makefile. DO NOT EDIT
```
rastislavs, line 1: If we are copying this file, we may use the one with my node-local-dns fix: https://github.com/rastislavs/cilium/blob/connectivity-check-dns/examples/kubernetes/connectivity-check/connectivity-check.yaml
With that one all tests should pass. We just need to document its source somewhere :)

pkg/test/e2e/cilium/cilium_test.go
rastislavs: Actually both ebpf and non-ebpf variant would be interesting..

.prow.yaml
PratikDeoghare: Failed to run the test with it. Will fix after debugging is done. Is there way to check this locally?

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: Yep. Let’s add.

go.mod
```diff
@@ -103,137 +101,171 @@ require (
 )
 
 require (
-	cloud.google.com/go v0.81.0 // indirect
+	github.com/cilium/cilium v1.11.0-rc0
```
embik: This is a release candidate, can we use a stable version of the package?

.golangci.yml
```diff
@@ -74,6 +74,9 @@ issues:
   - cyclomatic complexity [0-9]+ of func `\(\*Provider\)\.InitializeCloudProvider` is high
   - cyclomatic complexity [0-9]+ of func `\(\*Reconciler\)\.ensureResourcesAreDeployed` is high
   - cyclomatic complexity [0-9]+ of func `GetAPIV2NodeCloudSpec` is high
+  - cyclomatic complexity [0-9]+ of func `TestKonnectivity` is high
+  # goconst
+  - string `(get|openshift-kube-scheduler|us-central1|quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:86255c4efe6bbc141a0f41444f863bbd5cd832ffca21d2b737a4f9c225ed00ad)` has [0-9]+ occurrences. make it a constant
```
embik: Why do this and not apply the recommendation?

PratikDeoghare: This I got from resolving a merge conflict. Let me see where it comes from.

go.mod
PratikDeoghare: Ouch... Thanks for finding that. I missed it totally.

go.mod
```diff
@@ -72,7 +71,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.2.0
 	google.golang.org/api v0.56.0
 	google.golang.org/genproto v0.0.0-20211102202547-e9cf271f7f2c
-	google.golang.org/grpc v1.42.0
+	google.golang.org/grpc v1.41.0-dev
```
PratikDeoghare: wut

go.mod
PratikDeoghare: Found it. 

`github.com/envoyproxy/go-control-plane v0.9.10-0.20210907150352-cf90f659a021`

This breaks `pkg/controller/nodeport-proxy/envoymanager`. 
If I downgrade it to 9.9 (with which envoymanager works), cilium package gets downgraded to rc0 above. 

I guess I have to fix `pkg/controller/nodeport-proxy/envoymanager` to work with `go-control-plane v0.9.10-0`.

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+	}
+
+	userClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		t.Fatalf("failed to create client: %s", err)
+	}
+
+	var nodeIP string
+
+	t.Logf("waiting for nodes to come up")
```
lsviben, line 157: you could use this for the node check: https://github.com/kubermatic/kubermatic/blob/master/pkg/test/e2e/utils/client.go#L814

lsviben, line 157: Could you put all of these test blocks into separate functions? Like this its hard to keep up with what the test is doing as its huge

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+		}
+	}
+
+	t.Logf("waiting for pods to get ready")
+	{
+		err := wait.Poll(30*time.Second, 30*time.Minute, func() (bool, error) {
+			t.Logf("checking pod readiness...")
+
+			pods, err := userClient.CoreV1().Pods("kube-system").List(
+				context.Background(), metav1.ListOptions{})
```
lsviben: the `metav1.ListOptions` contain a `LabelSelector` which you can use to filter by label

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+			return true, nil
+		})
+		if err != nil {
+			t.Fatalf("nodes never became ready: %v", err)
+		}
+	}
+
+	t.Logf("waiting for pods to get ready")
+	{
+		err := wait.Poll(30*time.Second, 30*time.Minute, func() (bool, error) {
```
lsviben: the 30 min wait time seems excessive. If something goes wrong, we will wait for 30 mins for the test to fail. The same goes for all the test waits below. 

Could you give some realistic estimate when you think the test is failed for sure. IMO after the nodes are ready, if the pods dont get ready in 5 min, then they wont get ready at all.

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+			t.Logf("checking pod readiness...")
+
+			pods, err := userClient.CoreV1().Pods("kube-system").List(
```
lsviben: we use `ctrlruntimeclient`. If you use the cluster creation methods I mentioned in a comment above, you will have it ready to use. 


Check the other e2e tests how they create clusters and get master/seed/user clients. For example to get the user cluster client, you can use https://github.com/kubermatic/kubermatic/blob/master/pkg/test/e2e/utils/client.go#L1556

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+				t.Fatalf("failed to create %q namespace: %v", ciliumTestNs, err)
+			}
+		}()
+		if err != nil {
+			t.Fatalf("failed to create %q namespace: %v", ciliumTestNs, err)
+		}
+
+		t.Logf("namespace %q created", ciliumTestNs)
+
+		s := makeScheme()
```
lsviben: Check how its done for OPA tests : https://github.com/kubermatic/kubermatic/blob/master/pkg/test/e2e/opa/opa_test.go#L56

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+
+			pods.Items = filterByLabel(pods.Items, "name", names...)
+
+			if len(pods.Items) == 0 {
+				t.Logf("no connectivity test pods found")
+				return false, nil
+			}
+
+			allHealthy := true
+			for _, pod := range pods.Items {
```
lsviben: do we really need to check all these to be sure that the pod is ready?

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+							t.Log("not container ready", pod.Name, c.Type, c.Status)
+						}
+					}
+				}
+
+				if !podHealthy {
+					t.Logf("%q not healthy", pod.Name)
+				}
+
+				allHealthy = allHealthy && podHealthy
```
lsviben: there are a few places where you are listing pods and checking if they are ready/healthy. Maybe extract that into a function so its not duplicated.

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+			return nil, err
+		}
+		objs = append(objs, obj)
+	}
+
+	return objs, nil
+}
+
+// creates a usercluster on aws.
+func createUsercluster(t *testing.T, proxyMode string) (string, string, func(), error) {
```
lsviben: I have some questions about this:
1. Second return value is not used, so it can be removed?
2. Why do you need to save the kubeconfig to a file?

pkg/test/e2e/konnectivity/konnectivity_test.go
```diff
@@ -264,13 +264,49 @@ func TestKonnectivity(t *testing.T) {
...
+							t.Log("not ready", pod.Name, c.Type, c.Status)
+						}
+					} else if c.Type == corev1.ContainersReady {
+						if c.Status != corev1.ConditionTrue {
+							allHealthy = false
+							t.Log("not container ready", pod.Name, c.Type, c.Status)
+						}
+					}
+				}
+				t.Log(pod.Name, pod.Status.Phase)
```
lsviben, line 298: Here you also have a similar code block as in the cilium tests: get a list of pods, and check if they are ready, you can deduplicate all of this if you extract that to a func.

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+	{
+		s := makeScheme()
+
+		objs1, err := resourcesFromYaml("./testdata/hubble-relay-svc.yaml", s)
+		if err != nil {
+			t.Fatalf("failed to read objects from yaml: %v", err)
+		}
+
+		objs2, err := resourcesFromYaml("./testdata/hubble-ui-svc.yaml", s)
+		if err != nil {
```
lsviben: NIT: maybe instead of objs1 and objs2 some better name

pkg/test/e2e/cilium/cilium_test.go
```diff
@@ -0,0 +1,660 @@
...
+		},
+	}
+
+	var mu sync.Mutex
+
+	for _, test := range tests {
+		proxyMode := test.proxyMode
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			mu.Lock()
```
lsviben, line 129: not sure if the mutex is needed. There should not be issues if we create/delete multiple clusters at the same time. Or am I missing something?

lsviben (dismissed):
Added some comments

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: Without it sometimes cluster creation fails with 

> [POST /api/v1/projects][409] createProjectConflict

https://kubermatic.slack.com/archives/C01FBQ53EN6/p1644331513565219

PratikDeoghare: I hadn't enabled kubermatic-api logs when it happened. So not sure why it fails.

lsviben: I see, ill take a look at the project creation code to see if I can figure out smt

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: 1. This function was copied from other test TestKonnectivity. It is used there. I will move this function to the util package for use in both the tests. 
2. Hmm weird. Makes not much sense. I did it at the time and don't remember why.

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: Good catch. That is leftover from debugging. Wanted to keep the things alive while I debugged. Will fix.

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: Sure.

PratikDeoghare:
@lsviben Thanks for the detailed review. Will fix things on Monday-ish.

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: Can I do 1. in another PR? Konnectivity tests are disabled. I need to debug them and I am planning to refactor this in that PR. 

Fixed 2.

pkg/test/e2e/cilium/cilium_test.go
PratikDeoghare: This I could not use because I have `TestReadyCluster` which can directly test a usercluster if kubeconfig is provided. It would be unnecessarily complex to get projectID, seed etc just to construct the testclient, just to get usercluster client. 

Currently `testUserCluster` only depends on usercluster config which I think is neat.

PratikDeoghare:
@lsviben This is ready for review again. Ptal. :pray:

pkg/test/e2e/cilium/cilium_test.go
lsviben: sure, you can do it when the konnectivity tests are being fixed

pkg/test/e2e/cilium/cilium_test.go
lsviben: I see what it is, but currently no idea why it happens. Its looks like a name conflict, but that shouldnt happen as we randomize the names

moadqassem (approved):
/approve 
/lgtm

PR #8843, opened 2022-02-02
Do not run ICMP rules addition in Azure if we do not own the NSG

embik:

The ICMP migration runs unconditionally on an Azure NSG if `status.CloudMigrationRevision == 0`, which usually should not be the case, but can happen. This means it adds the new rules to pre-provisioned NSGs as well. This PR adds a simple ownership check, and makes the ICMP rules addition a no-op on non-owned NSGs.

Fixes #

**Documentation**:

```release-note bugfix
ICMP rules migration only runs on Azure NSGs created by KKP
```


moadqassem (approved):
/approve
/lgtm

embik:
/hold

docker pull limits will not allow the last test to succeed. Will try to rerun and cancel hold later.

PR #8846, opened 2022-02-02
Dissolve temporary SIG CRD

xrstf:
SIG CRD was introduced in #8078 to oversee the migration. Now that the migration has been completed, we can get rid of this SIG again. Nobody should need to touch pkg/crd/ anymore (and after 2.20 is released, it can be removed alltogether).

```release-note
NONE
```


kron4eg (approved):

kron4eg:
/hold

to stop autoretests

kron4eg (approved):

Issue #8723, opened 2022-01-13
Split oidc secrets for dashboard authentication and for kubeconfig of user clusters


michalvanco:
### What happened?
today I was troubleshooting authentication issue with KKP and Dex and option of OIDC KubeConfig for user clusters...
I have discovered that only option to make this work is to use the same OIDC secret for both dex clients (kubermatic and kubermaticIssuer) which I believe is not optimal because the secret for kubeconfig issuer is then exposed in the dowloaded kubeconfig for user cluster... see the example snippet
```
...
users:
- name: michal@kubermatic.com
  user:
    auth-provider:
      config:
        client-id: kubermaticIssuer
        client-secret: T79tKCSXoNNedNmDxaPyu8JQjOa6pqng    # <-- this one!!!
[...]

PR #8851, opened 2022-02-03
Update example values and KKPConfig to respect OIDC settings

michalvanco:
There were a misleading steps in the example values around the configuration of dex clients and KKPConfiguration.

Value of `issuerClientSecret` belongs to the issuer == used for accessing kubeconfig of user clusters via OIDC authentication.

And thus is actually not needed if users don't use that option.

It was discovered during the investigation of related issue - https://github.com/kubermatic/kubermatic/issues/8723.

Right now the situation was misleading because `issuerClientSecret` is not connected with the usage of `kubermatic` client (that is used for access KKP dahboard).

Fixes #

**Documentation**:
No updates in docs needed.

```release-note
Update example values and KubermaticConfiguration to respect OIDC settings
```

### DIFF
--- a/charts/kubermatic.example.ce.yaml
+++ b/charts/kubermatic.example.ce.yaml
@@ -46,9 +46,10 @@ spec:
     skipTokenIssuerTLSVerify: true
     tokenIssuer: https://cluster.example.dev/dex
 
-    # this must match the secret configured for the kubermatic client from
-    # the values.yaml.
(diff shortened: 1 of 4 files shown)


zreigz:
perfecto :+1: 
/lgtm

someone from app-management has to approve

xrstf (approved):
/approve

PR #8855, opened 2022-02-03
Fix Azure cloud spec defaulting and required fields

### LINKED ISSUE #8857: Cannot create Azure clusters with some cloud spec settings
### What happened?

The new CRDs handle validation of cluster specs differently, and some restrictions are tighter than before. I'll use this ticket to gather problems with creating Azure clusters, either on dev or in e2e tests.

Case 1: `loadBalancerSKU: ""` is rejected because it is now validated as enum:

```
Cluster.kubermatic.k8c.io "tdj2c9jr4v" is invalid: spec.cloud.azure.loadBalancerSKU: Unsupported value: "": supported values: "standard", "basic"
```

Case 2: `assignAvailabilitySet: null` is rejected because the field is validated as string by the CRD: 

```
"Cluster.kubermatic.k8c.io \"kclfsxpfs4\" is invalid: spec.cloud.azure.assignAvailabilitySet: Invalid value: \"null\": spec.cloud.azure.assignAvailabilitySet in body must be of type boolean: \"null\""
```

### What should happen?

Azure clusters should be created successfully.

### How to reproduce?

Case 1: set `loadBalancerSKU: ""` in an Azure cloud spec and apply it.
Case 2: set `assignAvailabilitySet: null` in an Azure cloud spec and apply it.

### Environment

- KKP Version: `master` ([e9e2191](https://github.com/kubermatic/kubermatic/commit/e9e21912831361f4d8f88924fbd76355838cdaae))
- Shared or separate master/seed clusters?: Shared

embik:

The new CRDs do validation differently/better. The CRD migration covered setting the `LoadBalancerSKU` field here: https://github.com/kubermatic/kubermatic/blob/e9e21912831361f4d8f88924fbd76355838cdaae/pkg/install/crdmigration/clones.go#L798-L804

For new clusters, this field needs to be defaulted, because otherwise creation will fail with this error:

```
Cluster.kubermatic.k8c.io "x9hm4dws5j" is invalid: spec.cloud.azure.loadBalancerSKU: Unsupported value: "": supported values: "standard", "basic"
```

In addition, some other changes to the Azure spec are necessary.

Fixes #8857 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/kubermatic-operator/Chart.yaml
+++ b/charts/kubermatic-operator/Chart.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: v1
 name: kubermatic-operator
-version: 0.4.0
+version: 0.4.1
 appVersion: '__KUBERMATIC_TAG__'
 description: Helm chart to install the Kubermatic Operator
 keywords:
--- a/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
+++ b/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
@@ -361,7 +361,6 @@ spec:
                       vnetResourceGroup:
                         type: string
                     required:
-                    - assignAvailabilitySet
                     - availabilitySet
                     - loadBalancerSKU
                     - resourceGroup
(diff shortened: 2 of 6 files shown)


xrstf (approved):
/approve
/retest

PR #8865, opened 2022-02-04
Add Nutanix csi driver

### LINKED ISSUE #8678: Add Nutanix CSI driver
Same as with https://github.com/kubermatic/kubeone/issues/1726 (for kubeone), we need to support the Nutanix CSI driver (https://github.com/nutanix/csi-plugin).

Use what is provided in the Helm charts (https://github.com/nutanix/helm/tree/master/charts), as that seems to be newer than the repository linked above.

mate4st:

Adds support for Nutanix CSI driver

Fixes #8678

```release-note
Add Nutanix CSI driver
```


addons/csi/nutanix/csi-driver.yaml
```diff
@@ -0,0 +1,466 @@
...
+          volumeMounts:
+            - mountPath: /csi
+              name: socket-dir
+          image: {{ Registry "k8s.gcr.io" }}/sig-storage/livenessprobe:v2.3.0
+          imagePullPolicy: IfNotPresent
+          args:
+            - --csi-address=/csi/csi.sock
+            - --http-endpoint=:9807
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
embik: This needs to be removed, no? We don't have master nodes in a user cluster.

addons/csi/nutanix/snapshot-controller.yaml
```diff
@@ -0,0 +1,878 @@
...
+        - --leader-election=false
+        securityContext:
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          runAsUser: 1000
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
embik: Same again here, I don't think this works as no node should have that label in a user cluster.

addons/csi/nutanix/snapshot-controller.yaml
```diff
@@ -0,0 +1,878 @@
...
+        volumeMounts:
+          - name: snapshot-validation-webhook-certs
+            mountPath: /etc/snapshot-validation-webhook/certs
+            readOnly: true
+      volumes:
+        - name: snapshot-validation-webhook-certs
+          secret:
+            secretName: csi-webhook-certs
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
embik: Same here.

docs/zz_generated.seed.yaml
```diff
@@ -146,6 +146,10 @@ spec:
             rhel: ""
             sles: ""
             ubuntu: ""
+          # PeEndpoint to use for accessing Nutanix Prism Element. No protocol or port should be passed,
+          peEndpoint: ""
+          # Optional: PePort to use when connecting to the Nutanix Prism Element endpoint (defaults to 9440)
+          pePort: null
```
embik: Let's keep the generated value the default port, same as we did with `port`. Needs changes here: https://github.com/kubermatic/kubermatic/blob/4f4928f3501e5aea22b57d1ed292ade3ea6e5a6f/codegen/example-yaml/main.go#L165.

pkg/resources/cloudconfig/configmap.go
```diff
@@ -106,6 +106,36 @@ func ConfigmapVsphereCSICreator(data configMapCreatorData) reconciling.NamedConf
...
+
+			nutanixCsiConf := fmt.Sprintf("%s:%d:%s:%s", data.DC().Spec.Nutanix.PeEndpoint, pePort, credentials.Nutanix.PeUsername, credentials.Nutanix.PePassword)
+
+			cm.Labels = resources.BaseAppLabels(name, nil)
+			cm.Data[resources.CloudConfigConfigMapKey] = nutanixCsiConf
+
+			return cm, nil
+		}
+	}
+}
```
embik: Why are we adding this `ConfigMap`? It does not seem to be mounted in any of the addon resources that are being added and it contains a password and should probably be a secret because of that.

docs/zz_generated.seed.yaml
```diff
@@ -146,6 +146,10 @@ spec:
             rhel: ""
             sles: ""
             ubuntu: ""
+          # PeEndpoint to use for accessing Nutanix Prism Element. No protocol or port should be passed,
```
embik: Uber nit: Remove the trailing comma.

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -458,6 +458,12 @@ type DatacenterSpecNutanix struct {
 	Endpoint string `json:"endpoint"`
 	// Optional: Port to use when connecting to the Nutanix Prism Central endpoint (defaults to 9440)
 	Port *int32 `json:"port,omitempty"`
+
+	// PeEndpoint to use for accessing Nutanix Prism Element. No protocol or port should be passed,
+	PeEndpoint string `json:"peEndpoint"`
+	// Optional: PePort to use when connecting to the Nutanix Prism Element endpoint (defaults to 9440)
+	PePort *int32 `json:"pePort,omitempty"`
```
embik: I'm still thinking about this, but I believe this should be in the Preset / Cluster, not the Seed. Here's my reasoning:

Currently, we treat one Prism Central instance as a KKP datacenter. One Prism Central can manage multiple NTNX clusters. Our `Cluster` resources have the `NutanixCloudSpec.ClusterName` key, which means you can select the NTNX cluster you'd like your user cluster running on _during user cluster creation_.

But Prism Element is the API for one NTNX cluster. Defining this at the `Seed` level means that you already lock everything to a single NTNX cluster, as the CSI driver will always talk to that NTNX cluster. Deploying to a different NTNX cluster means this breaks in bad ways.

IMO we have four ways out of this:

1. Lock one KKP Nutanix datacenter to a specific NTNX cluster managed by the target Prism Central. I'm not a fan of this, because it might create a lot of repetition if you manage several NTNX clusters from a Prism Central. Defeats the purpose of that to a certain degree IMHO.
2. Change the `Seed` datacenter spec to contain a `map[string]PrismElementConfig{}` with `PrismElementConfig` having an `endpoint` and a `port` key. That way, you can define multiple PE endpoints for NTNX cluster names at the Seed level.
3. See if we can be smart and query this information from Prism Central for a certain NTNX cluster name. That would probably be awesome, but not sure that's going to work.
4. Move this to the `Cluster` spec and set it in a `Preset` together with username and password.

WDYT, @WeirdMachine?

embik (dismissed):

pkg/resources/cloudconfig/configmap.go
mate4st: It lives in the cluster namespace on the seed and will be mirrored into the usercluster by the user-cluster-cm. Its a pendant to the cloud-config of vsphere. Therefore it is possible to use the same logic that is used by vsphere.

embik: I get that, but why are we putting it in the user cluster? Where is it used? (I'm just trying to understand it)

pkg/apis/kubermatic/v1/datacenter.go
mate4st: Yes, lets move it to the cluster spec. This is also a requirement to query it from the nutanix api.

pkg/resources/cloudconfig/configmap.go
mate4st: https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/cloudconfig/configmap.go#L79

embik: Not really happy about that but I see now why we need it. Let's keep it if that is necessary.

addons/csi/nutanix/snapshot-controller.yaml
```diff
@@ -0,0 +1,851 @@
...
+        image: {{ Registry "k8s.gcr.io" }}/sig-storage/snapshot-validation-webhook:v4.2.1
+        imagePullPolicy: IfNotPresent
+        args:
+          - --tls-cert-file=/etc/snapshot-validation-webhook/certs/cert.pem
+          - --tls-private-key-file=/etc/snapshot-validation-webhook/certs/key.pem
+          - --v=2
+          - --port=8443
+        ports:
+        - containerPort: 8443
+        securityContext:
```
embik, line 701: Can we also add the seccomp profile `RuntimeDefault` to this?

addons/csi/nutanix/snapshot-controller.yaml
```diff
@@ -0,0 +1,851 @@
...
+    spec:
+      serviceAccount: snapshot-controller
+      containers:
+      - name: snapshot-controller
+        image: {{ Registry "k8s.gcr.io" }}/sig-storage/snapshot-controller:v4.2.1
+        imagePullPolicy: IfNotPresent
+        args:
+        - --v=5
+        - --leader-election=false
+        securityContext:
```
embik, line 654: Can we also add the seccomp profile `RuntimeDefault` to this?

pkg/controller/user-cluster-controller-manager/resources/resources/csi/webhook.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/certificates/triple"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+// ValidatingWebhookConfigurationCreator returns the ValidatingWebhookConfiguration for the machine controller.
```
embik: Nit: comment needs an update.

embik (changes requested):
Small nits, but should be good afterwards.

pkg/validation/cluster.go
```diff
@@ -640,6 +640,46 @@ func validateNutanixCloudSpec(spec *kubermaticv1.NutanixCloudSpec) error {
...
+
+	if spec.PePort == nil {
+		if spec.CredentialsReference == nil {
+			return errors.New("no prism element port or reference specified")
+		}
+
+		if err := kuberneteshelper.ValidateSecretKeySelector(spec.CredentialsReference, resources.NutanixPePort); err != nil {
+			return err
+		}
+	}
```
embik: Should be optional since we use a default value in the provider code, no? E2E also fails with:

```
[FAIL] - [Kubermatic] Create cluster
      {"Payload":{"error":{"details":null,"code":400,"message":"invalid cluster: cloud: Invalid value: \"\u003credacted\u003e\": no prism element port or reference specified"}}}
```

hack/run-seed-controller-manager.sh
```diff
@@ -68,7 +68,7 @@ set -x
   -ca-bundle=$CA_BUNDLE \
   -addons-path=addons \
   -feature-gates=OpenIDAuthPlugin=true,KonnectivityService=true \
-  -worker-name="$(worker_name)" \
+  -worker-name="matthias" \
```
embik: needs to be reverted I think :D

mate4st: :D

WeirdMachine:
CSI option is now optional for nutanix to not break the api or introduce some other shenanigans. 

/unhold

embik (approved):
/approve
/lgtm

embik (approved):

kron4eg (approved):
/approve

PR #8866, opened 2022-02-04
make shutdown command wait for pods to be gone

xrstf:
Previously, the shutdown command just scaled down all Deployments to 0 at the same time. But the KKP controllers and operator need some time to shutdown, and so the operator just reconciled and thereby restarted all controllers. This means that running the shutdown command once would only shutdown the operator eventually. Then you have to run it again to shut down the seed/master controllers and then once again to finally shutdown the usercluster-ctrl-mgrs.

This PR improves this. Unfortunately there is no real status field for terminating pods and I know for a fact that GKE loves to leave Terminated pods laying around, but a name match seems to be the only choice. At worst the user has to manually delete the Terminated pod if they are on GKE.

```release-note
NONE
```

### DIFF
--- a/pkg/install/crdmigration/shutdown.go
+++ b/pkg/install/crdmigration/shutdown.go
@@ -18,7 +18,10 @@ package crdmigration
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


embik (approved):
One minute timeout for the "pods be gone" check might be a bit optimistic in my view, but I don't have any objections to merging this.

xrstf:
> One minute timeout for the "pods be gone" check might be a bit optimistic in my view, but I don't have any objections to merging this.

I was thinking that most pods probably have a 30s internal timeout, but I was just guessing :D

embik:
> I was thinking that most pods probably have a 30s internal timeout, but I was just guessing :D

I mean, if it works, it works. I don't see this being different across environments.

PR #8872, opened 2022-02-04
Consider Hetzner network set by cluster spec

### LINKED ISSUE #8868: Hetzner machine spec does not consider network defined in cluster spec - Test Release 2.20
### What happened?

As reported on the Kubermatic community slack, when creating a cluster on Hetzner and passing the Hetzner network in the cloud spec and not on the datacenter level, the creation of the (initial) machines fails with:

```
failed to create machine at cloudprovider, due to network "" does not exist
```

### What should happen?

The network passed on the Hetzner cloud provider spec should propagate to the `MachineDeployments` created for the cluster.

### How to reproduce?

- Set up Hetzner datacenter in your `Seed` with `network: ""`
- Create a Hetzner cluster with `network: "your-network-name"`
- Check the `MachineDeployment` for the configured network; expected value should be `your-network-name`, but will be an empty string

### Environment

- KKP Version: 2.19.0
- Shared or separate master/seed clusters?: Shared

embik:

#6588 introduced a `network` field for Hetzner on various levels (datacenter and cluster spec). The description suggests that the field on the cluster spec takes precedence over the datacenter field.

Unfortunately, it looks like the implementation back then did not consider the field on the cluster spec level for the generated `MachineDeployments` and only applied the datacenter-level value.

This can result in `MachineDeployments` failing to deploy if the datacenter value is not given or set to an empty string, or a mismatch from the network that is defined in CCM settings (as that does consider the cluster spec value):

https://github.com/kubermatic/kubermatic/blob/e234573f6cd2c52e948ddc681f56cc79046ce036/pkg/resources/cloudcontroller/hetzner.go#L79-L82

Fixes #8868

**Documentation**:

```release-note
If a network is set in the Hetzner cluster spec, it is now correctly applied to generated machines
```

### DIFF
--- a/pkg/resources/cloudcontroller/util.go
+++ b/pkg/resources/cloudcontroller/util.go
@@ -46,7 +46,7 @@ func ExternalCloudControllerFeatureSupported(dc *kubermaticv1.Datacenter, cluste
 		return !isOTC(dc.Spec.Openstack) && OpenStackCloudControllerSupported(cluster.Spec.Version)
 
 	case cluster.Spec.Cloud.Hetzner != nil:
-		return dc.Spec.Hetzner.Network != ""
+		return cluster.Spec.Cloud.Hetzner.Network != "" || dc.Spec.Hetzner.Network != ""
 
 	case cluster.Spec.Cloud.VSphere != nil:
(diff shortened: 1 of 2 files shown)


xrstf (approved):
Oh thank god someone fixed this. I experienced some weirdness myself but never got around to investigating where I fucked up.

/approve

PR #8875, opened 2022-02-04
Fixing linter complains about frontal commas

pkprzekwas:
For some reason linter started to complain about frontal commas. This PR removes such occurrences. 

Both me and @zreigz have experienced this issue. This has to be something new as @zreigz hasn't experienced this in the morning.
Screenshot 2022-02-04 at 14 48 09

```release-note
NONE
```

Signed-off-by: Patryk Przekwas <patryk@kubermatic.com>

### DIFF
--- a/pkg/apis/kubermatic/v1/addon.go
+++ b/pkg/apis/kubermatic/v1/addon.go
@@ -39,7 +39,7 @@ const (
 
 // Addon specifies a add-on.
 type Addon struct {
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta   `json:"inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec   AddonSpec   `json:"spec,omitempty"`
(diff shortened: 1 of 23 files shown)


xrstf:
Not sure what you're doing wrong, but `make lint` works perfectly fine for me with the golangci-lint version we use in CI (1.43) on the current master.

xrstf:
Did you build from source and get tagliatelle https://github.com/ldez/tagliatelle/commit/7b219253661560e60d284e3cddfb912dc49862fc by accident? That seems to have been fixed by https://github.com/ldez/tagliatelle/commit/c13286cdb4c6fd98c732a65bb0505247ab9a8363 right in the next commit.

kron4eg:
seems like this "broken" version of tagliatelle got into the golangci-lint v1.44.0, probably will be fixed in the next release

pkprzekwas:
I was using v1.44. They released it a week ago or so @xrstf. @kron4eg I'm not sure if this was introduced by accident. I'll check issues and release notes.

kron4eg:
@pkprzekwas this doesn't fix anything but actually breaks JSON marshalling, removing comma in front of `inline` simply makes embedded type to be marshalled to literally `inline` JSON field, instead of embedding it in the resulted .

pkprzekwas:
You're right @kron4eg , I wasn't aware of that. In this case this PR can be closed. I'll downgrade to 1.43 for the time being.

pkprzekwas:
Looks like upgrading github.com/ldez/tagliatelle to 0.3.1 is going to solve this issue https://github.com/ldez/tagliatelle/commit/c13286cdb4c6fd98c732a65bb0505247ab9a8363

PR #8877, opened 2022-02-04
Use KKP version for Helm chart versions

xrstf:
Now that our CRDs are fully spec'ed, we update the kubermatic-operator chart much more frequently. This leads -- at least for me :-D -- often to forgotten version updates. Also this leads to conflicts when to people change the CRDs in separate PRs.

<del>I propose that we stop my bad decision from years ago and instead just bump the chart version prior to a release. This will make the effort for a release a bit bigger, but since the dashboard also already has a release-procedure more complex than `git tag`, I guess this should be fine. The upside of not having to constantly bump this stupid version is much higher.</del>

The scripts have been changed to set the chartVersion of each chart to the current Git tag in our repo, i.e. when we release v2.20.7, all charts will also have that version.

Since Helm requires a semver, on dev we use a synthetic "v9.9.9-" version.

```release-note
NONE
```


xrstf:
/hold

Discuss!

embik:
I'm all for this change, but I wonder if we should include a variant of the chart verification script in our release pipeline? If it fails during a release, that release is unusable, but that seems preferable over having bad Helm chart versions. WDYT?

xrstf:
I have incorporated the feedback and now the scripts have been changed to set the chartVersion of each chart to the current Git tag in our repo, i.e. when we release v2.20.7, all charts will also have that version.

Since Helm _requires_ a semver, on dev we use a synthetic "v9.9.9-<githash>" version.

xrstf:
@wurbanski WDYT? Anything that speaks against going this route?

wurbanski:
@xrstf my concern would be that migrations in the installer are based on chart versions. Can we check we're not downgrading or breaking anything by going this route?

xrstf:
> migrations in the installer are based on chart versions

I presume you mean nginx and cert-manager.

* nginx triggers a migration if <1.3.0 is installed in the cluster
* cert-manager triggers a migration if <2.0 and another one if <2.1 is installed

Both charts would now be bumped to v2.20+, so we should be safe. However on dev things will stay tricky, but that was the case since forever.

wurbanski:
> Both charts would now be bumped to v2.20+, so we should be safe. However on dev things will stay tricky, but that was the case since forever.

so we are accidentally fine :D ok, I guess I don't see any more concerns as long as we keep our charts as overlays for the upstream ones (and the current plan is to keep doing that).

very nice life quality improvement IMHO!

/lgtm

xrstf:
@wurbanski had to rebase, care to re-review? :)

wurbanski:
sure, looks as fine as ever

/lgtm

wurbanski:
Do we actually need the placeholder `__KUBERMATIC_TAG__` version in the charts? can we just use the v9.9.9 as a placeholder?

xrstf:
That's an excellent point. Back in the days we had __KUBERMATIC_TAG__ and __DASHBOARD_TAG__ sprinkled around in our old Helm chart and other places, but since we have the operator, those placeholders slowly went away.

Except for the old `kubermatic` chart, which I just won't touch anymore until I finally remove it (sooon... soooooon....), I've set all charts to be v9.9.9-dev and use yq magic to rewrite the versions when needed.

hack/verify-spelling.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
xrstf, line 1: Using bash now because a) this script is part of the new big verify job which uses bash anyway and b) because the hack/lib.sh requires bas hnow due to the while-read loop redirect.

wurbanski:
🚢 it!

/lgtm

PR #8893, opened 2022-02-05
Improve ClusterStatus handling

xrstf:
In KKP <2.20, we had the ClusterStatus as part of our Cluster object and everything was fine. Now with the new API group, we made `status` a real subresource. This lead to race conditions between creating the Cluster object and creating the status subresource.

There are a number of issues that this PR attempts to solve:

1. Cluster conditions as slice. Kubernetes has the "strategic merge patch" to update slices, where you can specify a key element in each item. However this is not supported for CRDs. To prevent patch conflicts with the cluster conditions, this PR turns them into a map. It looks a bit weird at first, but makes things so much easier (upstreamsurely has a reason why they invented strategic patches for lists...).

1. Webhooks. It turns out that having a webhook for the status subresource is possible, but not really useful. It massively increases the latency (a few milliseconds are already "massive") and thereby lead to much, much more race conditions. It was not rare to see like 8 UPDATE requests at the same time being handled by the webhook. The network overhead then effectively adds some randomness to the order in which patches can be processed. Due to the other reasons listed below, this PR abandoed the idea of adding a webhook to default the ClusterStatus.

1. In general, KKP is special. The ClusterStatus is shared between many different controllers. Since each of them only updates a part of the status, none of them send a full patch. This means that if we had mandatory fields in the ClusterStatus, it would be rare to impossible to ever receive a patch that sets those fields, i.e. we will never actually be able to update the status.
   In Kubernetes this doesn't happen, because only a single controller manages the status for a single kind. So many of the problems we experience do not happen upstream.
   Because of this, this PR makes every field in the ClusterStatus optional. They are not really "optional" in the sense of "we never really need them", but more in a sense of "not every controller needs to be able to set them". The namespaceName for a Cluster for example is absolutely NOT optional and totally required to reconcile a cluster, but in the OpenAPI it's marked as optional anyway.

1. There is the special "userEmail" field in our ClusterStatus. This is special becaues right now we cannot determine this just from looking at a Cluster object. It _needs_ to be set by the KKP API. In the future we want fewer places where the KKP API has to have special code to auto-create subresources. But for now this PR keeps it the way it is. In general the PR however tried to move away from this, e.g. the namespaceName is now not set during creation, but only later by the Kubernetes controller in KKP. Always keep "in the future, we want to create clusters with kubectl only" in mind.

1. All of our controllers are now modified to wait for the status fields that they need. For most this means waiting until namespaceName has been set by the Kubernetes controller. Some others wait for the extendedHealth to be healthy, which implies a namespaceName being set already.

The biggest new addition to the codebase is `pkg/apis/kubermatic/v1/helper.UpdateClusterStatus()`, which will attempt to patch the status. It does automatic retries and fetches the current state to minimize  the chances of race conditions and conflicts. It's kind of a big hammer, but damn does it reduce the number of misleading warnings and errors in our logs.

The rest of the seemingly giant PR is just adjusting the code to map vs. slice for the conditions and to use the new `UpdateClusterStatus()` everywhere we can.

With all fields in the ClusterStatus now being optional, there is no point in defaulting them anymore and the original intent of this PR (to create a webhook to do that) is moot.

**NB:** For the cherrypick we need to also update the CRD migration code accordingly.

```release-note
NONE
```


kron4eg (approved):
/approve

xrstf:
@kron4eg It's ready for review. :tada:

pkg/collectors/addon.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
 
+	v1 "k8s.io/api/core/v1"
```
kron4eg: corev1

pkg/controller/seed-controller-manager/mla/mla.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 type cleaner interface {
-	cleanUp(context.Context) error
+	CleanUp(context.Context) error
```
kron4eg, line 40: Why this became a public method?

xrstf: Because it was hard for me to see which functions on the various reconcilers are called by other controllers and which are only called internally. It's not super duper important, but I liked to see `func (r *reconciler) CleanUp()` and immediately know "oh, this is probably called from somewhere outside". "CleanUp()" was similar to "Reconcile()" to me.

kron4eg (approved):
/hold
to prevent accidental merge

xrstf:
@kron4eg Why shouldn't this be merged now?

kron4eg:
@xrstf I wanted to let you do it ;)

xrstf:
Awww :)

/hold cancel

Time to break dev 😁

PR #8908, opened 2022-02-07
Add deletionTimestamp support for core objects to API

floreks:
Some of the core resources did not expose the `deletionTimestamp` as part of the `ObjectMeta` or did not have the `ObjectMeta` embedded at all. We need in the UI to be able to properly show the state of the resource and if it is being deleted.

```release-note
NONE
```


pkg/api/v1/types.go
```diff
@@ -494,6 +494,13 @@ func ConvertInternalUserToExternal(internalUser *kubermaticv1.User, includeSetti
 			ID:                internalUser.Name,
 			Name:              internalUser.Spec.Name,
 			CreationTimestamp: NewTime(internalUser.CreationTimestamp.Time),
+			DeletionTimestamp: func() *Time {
+				if internalUser.DeletionTimestamp != nil {
+					deletionTimestamp := NewTime(internalUser.DeletionTimestamp.Time)
+					return &deletionTimestamp
+				}
+				return nil
+			}(),
```
xrstf, line 503: You could expose this as a global function and then use it in all the handlers to convert timestamps. :)

xrstf (commented):
A comment

pkg/api/v1/types.go
floreks: I am sure someone will eventually refactor this :)

PR #8910, opened 2022-02-07
Dropping presetsFile & dynamicPresets flags from kubermatic-api

### LINKED ISSUE #8909: Drop support for `presetsFile` in kubermatic-api
Seems that we no longer use loading presets from file functionality, hence we should remove this flag and related code from our API.

pkprzekwas:
- Removing support from reading presets from file in kubermatic-api
- Removing dynamicPresets flag from api as this has been used as default for some time

Fixes #8909

**Documentation**:
Using presets file was removed from docs in KKP 2.17

```release-note
NONE
```

Signed-off-by: Patryk Przekwas patryk@kubermatic.com


cmd/kubermatic-api/options.go
```diff
@@ -41,7 +41,6 @@ type serverRunOptions struct {
 	internalAddr   string
 	prometheusURL  string
 	workerName     string
-	presetsFile    string
```
xrstf, line 44: You should also remove `-dynamic-presets`, which would always be true anyway. Removing the field from `serverRunOptions` will then trickle down into the other parts of the code where you can remove more code.

pkg/provider/kubernetes/preset.go
```diff
@@ -17,19 +17,14 @@ limitations under the License.
 package kubernetes
 
 import (
-	"bufio"
 	"context"
 	"fmt"
-	"io/ioutil"
-	"os"
-
```
xrstf, line 25: Please check the verify job and fix the errors it reported regarding the import statements.

charts/kubermatic/values.yaml
```diff
@@ -45,8 +45,6 @@ kubermatic:
   # clusters on one central Service of type LoadBalancer via the NodePort proxy or "LoadBalancer" to create a LoadBalancer service per cluster
   # **Note:** The `seed_dns_overwrite` setting of the `datacenters.yaml` doesn't have any effect if this is set to `LoadBalancer`
   exposeStrategy: "NodePort"
-  # base64 encoded presets.yaml. Predefined presets for all supported providers.
```
xrstf: Please undo all changes to the old `kubermatic` Helm chart. The chart is not used anymore and will just be removed alltogether in the near future. Until then, just forget it even exists. ^^

xrstf (changes requested):
Please also adjust https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/master/resources/kubermatic/api.go#L99

xrstf:
Also, just FYI, we never consider our CLI flags to be a stable API, so changes to them can be made whenever we like. The KKP operator knows how to setup KKP. The interface for admins is the KubermaticConfiguration CRD.

Oh, https://github.com/kubermatic/kubermatic/blob/master/hack/run-api.sh#L34-L36 can go away as well. Maybe grep through the codebase for the 2 CLI flags involved in this PR. :)

cmd/kubermatic-api/options.go
pkprzekwas: Done. Getting rid of dynamicPresets simplified a bit preset's provider logic.

pkg/provider/kubernetes/preset.go
pkprzekwas: > Please also adjust https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/master/resources/kubermatic/api.go#L99

Done 👍

charts/kubermatic/values.yaml
pkprzekwas: Reverted all changes related to helm chart.

pkprzekwas:
> Please also adjust https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/master/resources/kubermatic/api.go#L99

Adjusted.

pkprzekwas:
> Also, just FYI, we never consider our CLI flags to be a stable API, so changes to them can be made whenever we like. The KKP operator knows how to setup KKP. The interface for admins is the KubermaticConfiguration CRD.
> 
> Oh, https://github.com/kubermatic/kubermatic/blob/master/hack/run-api.sh#L34-L36 can go away as well. Maybe grep through the codebase for the 2 CLI flags involved in this PR. :)

In that case should I put ```NONE``` into release note?

I've greped for different combinations of `dynamic` & `presets` in the codebase. Seems like you've captured all of them in your comments. Only helm chart seems to still use them, which should be fine as we want to remove it soon.

xrstf (approved):
/approve

PR #8913, opened 2022-02-07
Change NodePortProxyEnvoy to a pointer in the componentOverride struct.

moadqassem:
In a previous PR, I added a new field to the componentOverride  struct, so users can customize the resources requests and limit for the NodePortProxy deployment in the cluster namespace in the seed. The field was not a pointer, thus it was perceived as a required field. 

Since the field has been added recently, any operations that's gonna be carried out on old clusters will be rejected by the validation webhook, since it doesn't find the corresponding field. This PR converts the field to a pointer to avoid validating the field when it is an old cluster(marked as non required),
Fixes #

**Documentation**:

```release-note
None
```


pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -484,7 +484,7 @@ type ComponentSettings struct {
 	Scheduler          ControllerSettings      `json:"scheduler"`
 	Etcd               EtcdStatefulSetSettings `json:"etcd"`
 	Prometheus         StatefulSetSettings     `json:"prometheus"`
-	NodePortProxyEnvoy NodeportProxyComponent  `json:"nodePortProxyEnvoy"`
+	NodePortProxyEnvoy *NodeportProxyComponent `json:"nodePortProxyEnvoy,omitempty"`
```
xrstf, line 487: just out of interest, what makes the nodeportenvoy settings different from all the others in this struct?

moadqassem: Nothing really, I just didn't want touch something was there before. Shall I reflect the change on all of them ?

xrstf: If I look at dev, a sample cluster has all those fields defaulted:

```yaml
  componentsOverride:
    apiserver:
      nodePortRange: 30000-32767
      replicas: 2
    controllerManager:
      leaderElection: {}
      replicas: 1
    etcd:
      clusterSize: 3
      diskSize: 5Gi
    prometheus: {}
    scheduler:
      leaderElection: {}
      replicas: 1
```

So I wonder why the NodePort settings are not defaulted. Those settings should trickle down from the Seed to the userclusters via the code in pkg/defaulting. And the Seed gets defaulted in pkg/controller/operator/defaults.

moadqassem: Ah I see, completely overlooked this part. Thanks for the hint.

moadqassem:
This needs to be adjusted to have validation on the seed level  thus would close it and open a new PR for the seed defaulting.

PR #8915, opened 2022-02-08
Make Cilium non-exclusive CNI for compatibility with Multus

### LINKED ISSUE #8898: Make Multus compatible with Cilium CNI
The aim of this story is to make Cilium CNI in KKP compatible with Multus.

Background:
Cilium modifies `/etc/cni/net.d` upon each cilium pod (re)start, which effectively disables Multus.

rastislavs:
This PR makes Multus compatible with Cilium by running Cilium as a non-exclusive CNI, which means it will NOT remove `/etc/cni/net.d` cofig files of other CNIs upon each Cilium pod (re)start.

Fixes #8898 

**Documentation**:

```release-note
Make Cilium non-exclusive CNI for compatibility with Multus.
```

### DIFF
--- a/addons/cilium/cilium.yaml
+++ b/addons/cilium/cilium.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Generated from:
-#   helm template cilium cilium/cilium --version 1.11.0 --namespace kube-system --set operator.replicas=1 \
+#   helm template cilium cilium/cilium --version 1.11.0 --namespace kube-system --set cni.exclusive=false --set operator.replicas=1 \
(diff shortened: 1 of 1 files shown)


WeirdMachine:
Can we set this parameter based on if multus is actually used?

rastislavs:
> Can we set this parameter based on if multus is actually used?

Potentially, with some more logic added into the addon controller, yes.

I am just not sure if it is worth introducing this kind of addon dependency, as this parameter is not changing anything in clusters where no other CNI than Cilium was ever installed. 

It only affects Canal -> Cilium migration scenario a bit, but I think even in that case we should be fine, as we recommend doing rolling restart of machine deployments in the cluster after the migration.

PR #8916, opened 2022-02-08
Increase metering image version

simon-wessel:
Update metering image version to include changes to API groups for KKP 2.20 (https://github.com/kubermatic/metering/pull/89).
Also exit metering minio containers if the mirror fails, so that errors don't go unnoticed (https://github.com/kubermatic/metering/issues/85). This has already been changed in the helm chart (https://github.com/kubermatic/metering/pull/88).

```release-note
NONE
```


xrstf (approved):
/approve

PR #8924, opened 2022-02-08
Set constraint parameters to schemaless

lsviben:
Makes Constraint Parameters schemaless as it can in reality be many different types, depending on the Constraint Template.

```release-note
NONE
```


xrstf (approved):
/approve

PR #8929, opened 2022-02-09
Only clean up Nutanix project category value if no resources present

embik:

This improves the logic around deleting the project category value. If a second cluster is created in the same KKP project, this will cause problems as there are leftover resources. If that is the case (resources outside of our user cluster), cleanup should be skipped. If we find resources owned by this cluster in our query result, something is wrong and an error is thrown.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

xrstf:
/override pre-kubermatic-opa-e2e

This one just flakes.

PR #8932, opened 2022-02-09
Fix an issue with the websocket usage when running api locally

floreks:
It was impossible to use the WebSocket when starting the API locally as by default the debug flag is set to `true` and the custom response writer did not implement the required `Hijack` method.

```release-note
NONE
```


xrstf (approved):
/approve

PR #8956, opened 2022-02-10
Do not use dot imports for blackbox testing

xrstf:
This fulfills a very old wish from Lukasz, for another Lukasz: https://github.com/kubermatic/kubermatic/pull/2534/files#r242824252

```release-note
NONE
```


xrstf:
@zreigz PTAL

xrstf:
@zreigz PTAL

PR #8976, opened 2022-02-11
Create eks nodepool

### LINKED ISSUE #8977: Create EKS Node Group
CREATE EKS Node Group and return details for GET endpoint

imharshita:
**What does this PR do / Why do we need it**: Create eks nodepool and return nodegroup details for GET endpoint

Fixes #8977

**Documentation**:

```release-note
CREATE EKS nodepool
```


pkg/handler/v2/external_cluster/eks.go
```diff
@@ -413,3 +435,57 @@ func deleteEKSNodeGroup(cluster *kubermaticv1.ExternalCluster, nodeGroupName str
 	}
 	return nil
 }
+
+func checkCreatePoolReqValid(machineDeployment apiv2.ExternalClusterMachineDeployment) error {
+	eksMD := machineDeployment.Cloud.EKS
+	if eksMD == nil {
+		return fmt.Errorf("EKS MachineDeployment Spec cannot be nil")
+	}
+	fields := reflect.ValueOf(eksMD).Elem()
```
lsviben: can we just check if the required fields are set manually instead of using reflect?  There are only 2 from what I see

imharshita: By manually do you mean using if/else on required fields?

imharshita: I did this for consistency to check all required fields for externalcluster like this, irrespective of the count of required fields. though I agree that its not time efficient to run the loop for just 2 fields.
wdyt, should we use reflect only when there is considerate number of required fields and for the rest just use if/else

lsviben: Yes, i meant for now just using if/else, if we get into a position where we have lots of required fields we can reconsider.

PR #8989, opened 2022-02-14
Use Hetzner instead of Digitalocean for e2e tests

xrstf:
Digitalocean can be slow and flaky. Hetzner is also not the best hoster, but it's fast and super cheap.

* Hetzner's CX21 instance is €0.010 / h or roughtly €6 per month.
* DO's s-2vcpu-2gb droplets are $0.02232 / h or roughly $15 per month.

So we would save roughly 50%. Of course machines in these tests are super duper short lived (usually less than a minute), so price isn't really the driving factor here. Reliability and speed are.

```release-note
NONE
```


hack/ci/run-etcd-launcher-tests.sh
```diff
@@ -34,18 +34,18 @@ export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
...
+  name: e2e-hetzner
   namespace: kubermatic
 spec:
-  digitalocean:
-    token: ${DO_E2E_TESTS_TOKEN}
+  hetzner:
+    token: ${HZ_E2E_TOKEN}
 EOF
-retry 2 kubectl apply -f preset-digitalocean.yaml
+retry 2 kubectl apply -f preset-hetzner.yaml
```
embik, line 48: Why don't we use the `kubeadm` ("BYOC") provider for this (and perhaps other tests that only concern the control plane)? That makes sure we do not rely on a cloud API to be up and running when not needed, as we only get the control plane. But that would suffice for `etcd-launcher` tests.

xrstf: Except for one test, all of the tests use 0 nodes. But for one e2e test, we do want an actual node with an actual cloud provider (or kind, maybe).

embik: Yeah, I get that, I don't mean to suggest that we completely drop using a cloud provider like Hetzner, but how do you think about using `byoc` where it makes sense (like the etcd-launcher test above)?

xrstf: We only have one single function to create a cluster in our tests, and that one uses Hetzner. I did not want to duplicate code needlessly.

embik: Okay, sounds reasonable for now.

embik (approved):

xrstf:
@floreks PTAL

PR #8990, opened 2022-02-14
Update nutanix-client-go to include category query fix

embik:

The Go client for Nutanix that we extracted some time ago from the Terraform provider has a bug for `GetCategoryQuery()`, which we are using in the Nutanix provider implementation. Fixed in https://github.com/embik/nutanix-client-go/commit/260fb79c80367eeefb3968e6b597adb5ed27a175.

In addition, I modified the client code some time ago to expose a `context.Context` argument, because I really do not like that it hardcoded `context.TODO()` for all HTTP requests.  The code changes in this PR don't change to a different context, but at least we can use it now. Might happen in a follow-up PR.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/provider/cloud/nutanix/provider.go
+++ b/pkg/provider/cloud/nutanix/provider.go
@@ -155,7 +155,7 @@ func deleteCategoryValues(client *ClientSet, cluster *kubermaticv1.Cluster) erro
 
 	projectID, ok := cluster.Labels[kubermaticv1.ProjectIDLabelKey]
 	if ok {
-		_, err := client.Prism.V3.GetCategoryValue(context.TODO(), ProjectCategoryName, projectID)
+		_, err := client.Prism.V3.GetCategoryValue(ctx, ProjectCategoryName, projectID)
 		if err != nil {
 			nutanixError, parseErr := ParseNutanixError(err)
 
--- a/pkg/provider/cloud/nutanix/provider.go
+++ b/pkg/provider/cloud/nutanix/provider.go
@@ -17,6 +17,7 @@ limitations under the License.
 package nutanix
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/provider/cloud/nutanix/provider.go
```diff
@@ -150,9 +151,11 @@ func (n *Nutanix) reconcileCluster(cluster *kubermaticv1.Cluster, update provide
 }
 
 func deleteCategoryValues(client *ClientSet, cluster *kubermaticv1.Cluster) error {
+	ctx := context.TODO()
+
 	projectID, ok := cluster.Labels[kubermaticv1.ProjectIDLabelKey]
 	if ok {
-		_, err := client.Prism.V3.GetCategoryValue(ProjectCategoryName, projectID)
+		_, err := client.Prism.V3.GetCategoryValue(context.TODO(), ProjectCategoryName, projectID)
```
kron4eg:
```suggestion
		_, err := client.Prism.V3.GetCategoryValue(ctx, ProjectCategoryName, projectID)
```

kron4eg (changes requested):
I think possibility of blocking the whole system because of no timeout is greatly underestimated.

At least, please setup local timeouts (using `context.WithTimeout()`).

pkg/provider/cloud/nutanix/provider.go
embik: That's literally why I exposed the context. But I only want this PR to be a fix to the category query because it breaks cluster deletion for Nutanix. I'll work on a follow-up PR that uses proper contexts.

The unfortunate reality is that that will put the Nutanix provider implementation ahead of others, e.g. Azure, which also uses `context.TODO()`: https://github.com/kubermatic/kubermatic/blob/47212a3996051a671c99544ea08ce44a6e49b55c/pkg/provider/cloud/azure/provider.go#L73. So maybe we group that under a larger refactor issue.

embik:
/hold

Still same problem as before in `seed-controller-manager` logs: 

```
{"level":"error","time":"2022-02-14T11:46:33.304Z","logger":"controller.kubermatic_cloud_controller","caller":"controller/controller.go:317","msg":"Reconciler error","name":"mttv9z5wrc","namespace":"","error":"failed cloud provider cleanup: error: The method is not allowed for the requested URL.","errorCauses":[{"error":"failed cloud provider cleanup: error: The method is not allowed for the requested URL."}]}
```

No idea why. Working on it.

embik:
/hold cancel

Turns out things work best if you actually pull in the revision of your client library that has the fix.

pkg/provider/cloud/nutanix/provider.go
embik: Fixed the highlighted problem, but as said, I'd much prefer to tackle TODO contexts across our provider implementations in a more coordinated fashion. I don't think this PR is the right place for that (I really just want to fix Nutanix cluster deletion right now).

embik:
@kron4eg since I have no good idea of what acceptable timeouts are yet, I'd really like to keep this as is (before this, the client library was using `context.TODO()` as well) to fix the cluster cleanup for Nutanix and then follow up with a PR that improves the context situation.

kron4eg (approved):

embik:
The Nutanix installation is getting rate limited by Dockerhub.

PR #8923, opened 2022-02-08
Add prometheus-operator proposal

wurbanski:
**What does this PR do / Why do we need it**:
Proposal regarding moving the monitoring stack to prometheus-operator on master/seed.

**Does this PR close any issues?**:
Closes #8849 

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #8992, opened 2022-02-14
update usercluster Prometheus to 2.33.3, allow more CPU time

xrstf:
This simply bumps Prometheus to the latest stable version. There seem to be no breaking changes. I allowed Prometheus a bit more CPU, considering how large clusters can get.

```release-note
Update usercluster Prometheus to 2.33.3
```


wurbanski:
/lgtm

but: 
1. I think the seed cluster Prometheus should be upgraded as well (in the chart)
2. obligatory PTAL #8923

PR #8961, opened 2022-02-10
Remove some deprecated fields from new CRD types

kron4eg:
**What does this PR do / Why do we need it**:
Cleanup our new shiny CRD types from different deprecations

**Does this PR close any issues?**:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Remove deprecated fields from CRD types
```

PR #8993, opened 2022-02-14
Bring back vSphere cluster field and make it required

kron4eg:
Revert SOME changes from #8961.

As it turned out we actually need vSphere cluster field for CSI driver, vsphere cluster field is back and is made **REQUIRED**.

Fixes #

**Documentation**:

```release-note
Bring back vSphere cluster field and make it required
```

### DIFF
--- a/pkg/apis/kubermatic/v1/datacenter.go
+++ b/pkg/apis/kubermatic/v1/datacenter.go
@@ -368,6 +368,8 @@ type DatacenterSpecVSphere struct {
 	DefaultDatastore string `json:"datastore"`
 	// The name of the datacenter to use.
 	Datacenter string `json:"datacenter"`
+	// The name of the vSphere cluster to use. Used for out-of-tree CSI Driver.
+	Cluster string `json:"cluster"`
 	// The name of the storage policy to use for the storage class created in the user cluster.
 	DefaultStoragePolicy string `json:"storagePolicy,omitempty"`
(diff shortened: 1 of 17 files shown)


xmudrii (approved):

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -368,6 +368,8 @@ type DatacenterSpecVSphere struct {
 	DefaultDatastore string `json:"datastore"`
 	// The name of the datacenter to use.
 	Datacenter string `json:"datacenter"`
+	// The name of the vSphere cluster to use.
```
moadqassem: Can we please add that this is required for the OOT csi driver ?

kron4eg: On this level no, it's not possible to differentiate in schema if vSphere is using OOT CSI or in-tree volumes.

moadqassem: Not sure what does that mean. But the point behind my comment is, to make sure that we don't overlook this fact(the fact that this is needed for thjs CSI driver) and remove it in future releases!

kron4eg: oh, you meant the comment...

kron4eg: Done

moadqassem (approved):
/approve
/lgtm

PR #8994, opened 2022-02-14
Use retry for adding/removing finalizers

xrstf:
This introduces new helpers that use `retry` just like `UpdateClusterStatus()` does (since #8893). Since retrying seems very successful and hides a lot of transient, pointless errors from our logs, I think it's worth it.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/backup/backup_controller.go
+++ b/pkg/controller/seed-controller-manager/backup/backup_controller.go
@@ -47,7 +47,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	utilpointer "k8s.io/utils/pointer"
@@ -259,7 +258,7 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 	// Cluster got deleted - regardless if the cluster was ever running, we cleanup
 	if cluster.DeletionTimestamp != nil {
 		// Need to cleanup
-		if sets.NewString(cluster.Finalizers...).Has(cleanupFinalizer) {
+		if kuberneteshelper.HasFinalizer(cluster, cleanupFinalizer) {
 			if !r.disabled {
 				if err := r.Create(ctx, r.cleanupJob(cluster)); err != nil {
 					// Otherwise we end up in a loop when we are able to create the job but not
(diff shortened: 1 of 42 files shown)


pkg/controller/seed-controller-manager/backup/backup_controller.go
```diff
@@ -285,12 +282,8 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
...
 
 	// Always add the finalizer first
-	if !kuberneteshelper.HasFinalizer(cluster, cleanupFinalizer) && !r.disabled {
-		oldCluster := cluster.DeepCopy()
-		kuberneteshelper.AddFinalizer(cluster, cleanupFinalizer)
-		if err := r.Patch(ctx, cluster, ctrlruntimeclient.MergeFrom(oldCluster)); err != nil {
-			return fmt.Errorf("failed to update cluster after adding cleanup finalizer: %w", err)
-		}
+	if err := kuberneteshelper.TryAddFinalizer(ctx, r, cluster, cleanupFinalizer); err != nil {
+		return fmt.Errorf("failed to add finalizer: %w", err)
```
embik: This change lost the `&& !r.disabled` condition. Is that okay?

pkg/kubernetes/helper.go
```diff
@@ -61,13 +65,85 @@ func RemoveFinalizer(obj metav1.Object, toRemove ...string) {
...
+		RemoveFinalizer(obj, finalizers...)
+		current := sets.NewString(obj.GetFinalizers()...)
+
+		// save some work
+		if previous.Equal(current) {
+			return nil
+		}
+
+		// update the object
+		return client.Update(ctx, obj)
```
embik, line 93: A lot of the code this replaced uses as `Patch()` instead of an `Update()`. Any reason we're going with an update here? Wouldn't a patch be more likely to succeed?

embik (commented):
Looks good overall but I had two questions during the review.

pkg/controller/seed-controller-manager/backup/backup_controller.go
xrstf: Oh wow, well spotted!

xrstf: Fixed.

pkg/kubernetes/helper.go
xrstf: To create a patch, I would need to clone the old object. Unfortunately a `ctrlruntimeclient.Object` only has `DeepCopyObject()`, which returns a `runtime.Object`, for which we cannot create patches. So I had to resort to `Update()`.

embik: Gotcha, makes sense. That's unfortunate.

embik (approved):

embik (approved):

PR #9001, opened 2022-02-14
Refactor usercluster-controller-manager's RBAC controller

xrstf:
`klog` should not be used in KKP. This PR removes its usage in the rbac controller and at the same time refactors it to use our reconciling framework instead of reinventing the "get-create-check-update" pattern.

```release-note
NONE
```


pkg/controller/user-cluster-controller-manager/rbac/reconciler.go
```diff
@@ -19,90 +19,85 @@ package rbacusercluster
...
-	err = r.ensureRBACClusterRoleBinding(ctx, resourceName)
+
+	err = r.ensureRBACClusterRoleBinding(ctx, request.Name)
 	if err != nil {
 		return err
 	}
 
+	r.rLock.Lock()
+	defer r.rLock.Unlock()
+	r.reconciledSuccessfullyOnce = true
```
kron4eg, line 74: I don't see the use of this var, what's it's for?

xrstf: Check the `registerReconciledCheck` call

kron4eg (approved):

PR #9003, opened 2022-02-14
Remove CLI flags for backup containers / legacy backup options

xrstf:
This PR is part of the ongoing effort to get rid of redundant CLI flags on our controller-managers and instead of configuration dynamically from the KubermaticConfiguration. 

In general, this PR simply started to remove the CLI flags for the backup containers (store, cleanup and delete containers), but from there it grew in complexity, as more and more hacks were uncovered (stuff like defaulting side effects happening inside ConfigMapCreators) and now this PR not just removes a few flags, but also kills a few deprecated things.

**NB:** "new controllers" means `pkg/controller/seed-controller/etcdbackup` and `etcdrestore`. "old/legacy controller" is `pkg/controller/seed-controller/backup`.

* The deprecated `KubermaticConfiguration.spec.seedController.backupRestore` has been removed alltogether. It was introduced in 2.17 and deprecated in 2.18 already. Also migrating to the new configuration syntax is easy, so the impact of removing this should be minimal.
* The deprecated `Seed.spec.backupRestore` has been removed as well. This was introduced in 2.18 and deprecated one release later in 2.19, being superseded by the multi destination handling. Again, the impact should be minimal.
* The new controllers do not attempt to be a drop-in replacement for the old controller anymore. This means that the new controllers will not run `cleanup` jobs (i.e. the new controllers only use "store" and "delete" containers).
* Both old and new controllers load the containers dynamically from the `KubermaticConfiguration` (the OG reason for this PR).
* Importantly, the backup containers are **not implicitly defaulted anymore**. There was so much `if cfg.container == defaultValue` in our codebase that it became apparent that defaulting these fields is counterproductive (at least while we're still supporting the old controller). While KKP still has its default values for those containers and we still use them, we do not set them at runtime in the `KubermaticConfiguration` object.
* The default location for seed backups is now mandatory. The webhook ensures that the default matches on of the configured destinations. This was done in part to get rid of the usage of the deprecated (now removed) fields. `EtcdBackupConfig.spec.destination` is now also a required field.
* Both old and new controllers are always started and react dynamically to wether or not they are needed.
* It cannot happen anymore that the delete container is empty, so the special handling for that has been removed as well.

```release-note
The etcd-backup-related containers are now loaded dynamically from the KubermaticConfiguration, the relevant CLI flags like `-backup-container=<file>` have been removed.
The deprecated configuration options `KubermaticConfiguration.spec.seedController.backupRestore` and `Seed.spec.backupRestore` have been removed. Please migrate to `Seed.spec.etcdBackupRestore`.
```


xrstf:
@multi-io @furkhat This PR cleans up the backup logic you guys worked on. Any concerns? :)

pkg/controller/operator/seed/reconciler.go
```diff
@@ -429,32 +429,23 @@ func (r *Reconciler) reconcileConfigMaps(ctx context.Context, cfg *kubermaticv1.
...
-	}
-	// Seed backup and restore configuration takes precedence.
-	if seed.Spec.BackupRestore != nil {
-		backupRestore = seed.Spec.BackupRestore
-	}
-	if creator := kubermaticseed.RestoreS3SettingsConfigMapCreator(backupRestore); creator != nil {
-		kubeSystemCreators = append(kubeSystemCreators, creator)
-	}
+		creators = []reconciling.NamedConfigMapCreatorGetter{
+			kubermaticseed.RestoreS3SettingsConfigMapCreator(destination),
```
lsviben: this should not be needed anymore if we support just the newest destination backups. Now the backups will take the destination endpoint/bucketname/credentials from the seed destinations and set the env vars dynamically based on the destination, instead of the hardcoded configmap.
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go#L864-L875
and 
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/etcdbackup/etcd_backup_controller.go#L1018-L1029


Which means we can also remove the env part of the newStoreContainer and newDeleteContainer in `defaults.go`

pkg/handler/v1/admin/seed_test.go
```diff
@@ -187,29 +187,29 @@ func TestUpdateSeedEndpoint(t *testing.T) {
...
 			httpStatus:             http.StatusOK,
 			existingKubermaticObjs: []ctrlruntimeclient.Object{genUser("Bob", "bob@acme.com", true), test.GenTestSeed()},
 			existingAPIUser:        test.GenDefaultAPIUser(),
 		},
 		// scenario 5
 		{
 			name:                   "scenario 5: authorized user updates seed - just backup",
```
lsviben: this whole scenario can be removed, it was testing the old backupRestore

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -162,33 +162,20 @@ type SeedSpec struct {
...
-	S3BucketName string `json:"s3BucketName,omitempty"`
-}
-
-// EtcdBackupRestore holds the configuration of the automatic backup restores.
+// EtcdBackupRestore holds the configuration of the automatic backup and restores.
 type EtcdBackupRestore struct {
 	// Destinations stores all the possible destinations where the backups for the Seed can be stored. If not empty,
 	// it enables automatic backup and restore for the seed.
 	Destinations map[string]*BackupDestination `json:"destinations,omitempty"`
 	// DefaultDestination Optional setting which marks the default destination that will be used for the default etcd backup config which is
```
lsviben:
```suggestion
	// DefaultDestination Marks the default destination that will be used for the default etcd backup config which is
```

lsviben (commented):
Feels good to have this cleaned up, left a few comments. 

We would need to adjust also some part of the API and UI, as well as a migration in 2.20 for the etcdbackupconfigs

xrstf:
> as well as a migration in 2.20 for the etcdbackupconfigs

Yeah I would then include the migration bits and pieces in the backport PR.

lsviben:
Just one more comment that you missed is that we can remove the env vars in:
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/defaults/defaults.go#L794-L814
and 
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/defaults/defaults.go#L848-L867

otherwise LGTM, and when its merged ill do a follow-up with some changes to the API handler

xrstf:
Well spotted, again!

Fixed.

xrstf:
@rastislavs You made the original PR, WDYT about backporting this to 2.18?

Issue #8490, opened 2021-12-17
Initialize Virtual Machine Presets


mfranczy:
For better user experience we should allow people to use virtual machine presets. That requires from us to create a standard [virtual machine instance preset](https://kubevirt.io/user-guide/virtual_machines/presets/) and allow customers to create their own presets on KubeVirt cluster in the default namespace. 

Later we have to clone all presets (defined by customer in the default namespace and standard created by us) on each cluster creation to the cluster namespace on KubeVirt infra cluster. Additionaly we should update the list method to return a standard preset together with those defined by customers ahead.

Specification for standard preset should include:
* 2 vCPUs
* 8Gi of memory

Listing presets:

![Presets - list.drawio.png](https://images.zenhubusercontent.com/5dc5412ea240680001cc35aa/1282d5a0-862a-4a5f-ba29-5c9b8e013e3a)

Cloning standard presets:
![Preset - clone.drawio.png](https://images.zenhubusercontent.com/5dc5412ea240680001cc35aa/3ba87ba7-9a60-4a8d-acbf-50081aea9aec)

Acceptance criteria:
* Create a standard Virtual Machine Preset template spec, it should be placed in the [kubevirt cloud provider implementation](https://github.com/kubermatic/kubermatic/tree/master/pkg/provider/cloud/kubevirt).
* Update the [API list endpoint](https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/routes_v2.go#L3575-L3597) to combine presets that are queried on KubeVirt cluster with a standard preset
* On each cluster creation the all presets must be copied to the cluster namespace on KubeVirt infra cluster

**Note that listing and cloning implementation should consider a use case where we have more than a one presets defined by us.**

PR #9014, opened 2022-02-15
Simplify KubeVirt API types

hdurand0710:

Goal is to simplify the api types for Kubevirt (flavor, storage class). For Flavor, we keep only as properties the name, namespace and a marshalled string of the Flavour Spec. For the Storage Class, only the name.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/api/v2/types.go
```diff
@@ -1062,55 +1059,13 @@ type GKEUpgradeSettings struct {
...
-}
-
 // VirtualMachineInstancePreset represents a KubeVirt Virtual Machine Instance Preset
 // swagger:model VirtualMachineInstancePreset
 type VirtualMachineInstancePreset struct {
-	apiv1.ObjectMeta `json:",inline"`
-	// VirtualMachineInstance Spec contains the VirtualMachineInstance specification.
-	Spec VirtualMachineInstancePresetSpec `json:"spec,omitempty" valid:"required"`
+	Name      string `json:"name,omitempty"`
+	Namespace string `json:"namespace,omitempty"`
```
maciaszczykm, line 1066: Is the namespace required, will it be displayed in the UI?

hdurand0710: That's a good point. When implementing [Initialize Virtual Machine Presets](https://github.com/kubermatic/kubermatic/issues/8490), we will fetch Presets from the `default` namespace, but also add a/several `standard` hard-coded Presets (empty namespace). I think that it's good to display then the namespace (`default` or empty).

maciaszczykm: Okay 👍🏻

PR #8859, opened 2022-02-03
Override resource requirements for nodeport-proxy for user clusters

moadqassem:
**What does this PR do / Why do we need it**:
At the moment there is direct and clean way to update the nodeport-proxy pods for the user clusters. Once the user enable LoadBalancer as an expose strategy for seeds, we deploy nodeport-proxy-envoy pod. This pod sometimes it goes above the resources limit which would lead to OOM error which would kill the pod and dont recover at all. This PR changes this logic by adding support for customizing NodePort-Proxy deployment resources via the componentOverride field in the user cluster CR.
 
**Does this PR close any issues?**:
Fixes #8826

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Support custom pod resources for NodePort-Proxy pod for the user cluster.
```

PR #9016, opened 2022-02-16
Fix resource overrides and use resource values in naming for nodeport-proxy

embik:
The nodeport-proxy resource overrides introduced in #8859 don't work because the resource override map populated by `GetOverrides()` holds the nodeport-proxy values under key `nodeport-proxy-envoy` (the Deployment name), but `SetResourceRequirements` uses the container names (which are `envoy-manager`, `envoy` and `lb-updater` for the `nodeport-proxy-envoy` Deployment). Because of that, the override was not applied correctly.

Since I don't want to rename the container (there is no point in causing a restart), I'm adding "global" `resources` values and use it on both sides for consistency. The map now uses the envoy container name instead of the Deployment name.

Fixes #

**Documentation**:

```release-note
NONE
```

No changelog because this just fixes the feature that was supposed to work with #8859.

### DIFF
--- a/pkg/resources/nodeportproxy/nodeportproxy.go
+++ b/pkg/resources/nodeportproxy/nodeportproxy.go
@@ -38,7 +38,7 @@ import (
 const (
 	name               = "nodeport-proxy"
 	imageName          = "kubermatic/nodeport-proxy"
-	envoyAppLabelValue = name + "-envoy"
+	envoyAppLabelValue = resources.NodePortProxyEnvoyDeploymentName
 
 	// NodePortProxyExposeNamespacedAnnotationKey is the annotation key used to indicate that
 	// a service should be exposed by the namespaced NodeportProxy instance.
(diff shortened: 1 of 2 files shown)


xrstf:
Will this create a 2nd envoy Deployment once it's deployed into an existing KKP setup?

moadqassem:
> Will this create a 2nd envoy Deployment once it's deployed into an existing KKP setup?

Yeah, it will. I would say we just change the name on the map instead. @embik nice catch!

embik:
Sorry, not following you, @xrstf @moadqassem. This should keep the name of the Deployment the same, `nodeport-proxy-envoy`. It just makes more consistent use of the `resources` package. Am I missing something?

moadqassem:
> Sorry, not following you, @xrstf @moadqassem. This should keep the name of the Deployment the same, `nodeport-proxy-envoy`. It just makes more consistent use of the `resources` package. Am I missing something?

Ah I see, you are just using the full name without any concatenation. So eventually `NodePortProxyEnvoyDeploymentName` is `nodeport-proxy-envoy` which means the concatenation is not needed

embik:
The PR is not even changing container names (well, it starts using a constant, but it's still the same value), it just makes sure that the resource override map has an entry under `envoy` (the container name for the `nodeport-proxy-envoy` Deployment) :-)

xrstf:
Yeah i looked at the wrong constant. All good.

/approve
/lgtm

PR #9018, opened 2022-02-16
[release/v2.19] Override resource requirements for nodeport-proxy for user clusters

### LINKED ISSUE #8826: Limits not configurable for customer cluster Nodeport-Proxy-Envoy - Test Release 2.20
### What happened?
I have at least one cluster where the nodeport-proxy-envoy is using ~70mb of memory,
as we have a hardcoded limit if 64, in get constant OOMs.

### What should happen?
In the same way that we can set this for the seeds, we would need that in the cluster overrides too
### Environment

- KKP Version: 2.18.5
- Shared or separate master/seed clusters?: seperate 
- Master/seed cluster provider: Openstack

embik:

Backports #8859 to `release/v2.19`.

Fixes #8826

**Documentation**:

```release-note
Support custom pod resources for NodePort-Proxy pod for the user cluster.
```


xrstf (approved):
/approve

pkg/resources/nodeportproxy/nodeportproxy.go
```diff
@@ -38,7 +38,7 @@ import (
 const (
 	name               = "nodeport-proxy"
 	imageName          = "kubermatic/nodeport-proxy"
-	envoyAppLabelValue = name + "-envoy"
```
qeqar, line 41: Sorry to comment on an old PR, but this breaks all clusters.

as the frontend lodbalancer still uses this name

PR #9019, opened 2022-02-16
bugfix: check data dir state before starting etcd

furkhat:

Fixes broken feature when for any reason data directory for one of the etcd instances was lost then the etcd instance re-adds itself to the cluster and starts as a new member.
There was a **race** in the code because **checking if data directory empty after starting etcd always false**. The data directory must be checked before starting etcd instance.

You can test this by creating a cluster then deleting PVC/PV for one of the etcd members in the cluster. Without this fix the instance will always fail. With this fix the instance will rejoin the cluster as a new member.

**Documentation**:

```release-note
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -119,6 +119,9 @@ func main() {
 		}
 	}
 
+	_, err = os.Stat(filepath.Join(e.dataDir, "member"))
+	dataDirectoryIsEmpty := os.IsNotExist(err)
+
 	// setup and start etcd command
 	etcdCmd, err := startEtcdCmd(e, log)
 	if err != nil {
@@ -138,7 +141,7 @@ func main() {
 	if thisMember != nil {
 		log.Infof("%v is a member", thisMember.GetPeerURLs())
 
-		if _, err := os.Stat(filepath.Join(e.dataDir, "member")); os.IsNotExist(err) {
+		if dataDirectoryIsEmpty {
 			client, err := e.getClusterClient()
 			if err != nil {
 				log.Panicw("can't find cluster client: %v", zap.Error(err))


embik:
@furkhat thanks for catching this! There is a step in the e2e test that is supposed to test this scenario, but it looks like the test is flawed. I'll look into fixing that.

I have some doubts about this change though: When I deploy a fresh cluster with etcd-launcher enabled, I would expect this change to always fail because `getMemberByName()` will always throw an error (since the etcd cluster isn't up yet, so you cannot query for the list of members). Have you tested that?

cmd/etcd-launcher/main.go
```diff
@@ -158,6 +146,23 @@ func main() {
...
+	if err = wait.Poll(1*time.Second, 60*time.Second, func() (bool, error) {
+		return e.isClusterHealthy(log)
+	}); err != nil {
+		log.Panicw("manager thread failed to connect to cluster", zap.Error(err))
+	}
+
+	// handle changes to peerURLs (https -> http)
+	if err := e.updatePeerURLs(log); err != nil {
+		log.Panicw("failed to update peerURL", zap.Error(err))
+	}
```
embik: Why is this being called here? The `https -> http` change is no longer true since `etcd-launcher` now supports peer TLS connections. `updatePeerURLs()` is already being called earlier in the code.

embik:
I raised #9020 to have proper tracking on this issue. Unfortunately this PR doesn't seem to fix it, the `pre-kubermatic-etcd-launcher-e2e` is failing and it looks kind of bad:

```
$ k get pods -A | grep etcd
cluster-j7f59b7xpt         etcd-0                                                  0/1     CrashLoopBackOff   12         39m
cluster-j7f59b7xpt         etcd-1                                                  0/1     CrashLoopBackOff   12         39m
cluster-j7f59b7xpt         etcd-2                                                  0/1     CrashLoopBackOff   12         39m
cluster-lpjkz67zqs         etcd-0                                                  0/1     CrashLoopBackOff   9          24m
cluster-lpjkz67zqs         etcd-1                                                  0/1     CrashLoopBackOff   9          24m
cluster-lpjkz67zqs         etcd-2                                                  0/1     CrashLoopBackOff   9          24m
```

Logs show something that looks related to my concerns above:

```
{"level":"info","time":"2022-02-16T13:58:17.697Z","caller":"etcd-launcher/main.go:105","msg":"initializing etcd.."}
{"level":"info","time":"2022-02-16T13:58:17.698Z","caller":"etcd-launcher/main.go:106","msg":"initial-state: existing"}
{"level":"info","time":"2022-02-16T13:58:17.698Z","caller":"etcd-launcher/main.go:107","msg":"initial-cluster: etcd-0=https://etcd-0.etcd.cluster-lpjkz67zqs.svc.cluster.local:2381,etcd-1=https://etcd-1.etcd.cluster-lpjkz67zqs.svc.cluster.local:2381,etcd-2=https://etcd-2.etcd.cluster-lpjkz67zqs.svc.cluster.local:2381"}
{"level":"info","time":"2022-02-16T13:58:17.698Z","caller":"etcd-launcher/main.go:109","msg":"peer-tls-mode: strict"}
{"level":"warn","ts":"2022-02-16T13:58:22.699Z","logger":"etcd-client","caller":"v3@v3.5.1/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00047c8c0/etcd-0.etcd.cluster-lpjkz67zqs.svc.cluster.local:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: Error while dialing dial tcp 10.244.0.98:2379: connect: connection refused\""}
{"level":"warn","time":"2022-02-16T13:58:22.699Z","caller":"etcd-launcher/main.go:118","msg":"failed to update peerURL, etcd node might fail to start ...","error":"context deadline exceeded"}
{"level":"warn","ts":"2022-02-16T13:58:27.700Z","logger":"etcd-client","caller":"v3@v3.5.1/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00047ce00/etcd-0.etcd.cluster-lpjkz67zqs.svc.cluster.local:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: Error while dialing dial tcp 10.244.0.98:2379: connect: connection refused\""}
{"level":"panic","time":"2022-02-16T13:58:27.700Z","caller":"etcd-launcher/main.go:124","msg":"failed to check cluster membership","error":"context deadline exceeded"}
panic: failed to check cluster membership

goroutine 1 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc000a420c0, {0xc00004ba00, 0x1, 0x1})
	go.uber.org/zap@v1.19.1/zapcore/entry.go:232 +0x446
go.uber.org/zap.(*SugaredLogger).log(0xc0005a23e0, 0x4, {0x1b5866a, 0x3b}, {0x0, 0x0, 0x0}, {0xc000ac7b30, 0x1, 0x1})
	go.uber.org/zap@v1.19.1/sugar.go:227 +0xee
go.uber.org/zap.(*SugaredLogger).Panicw(...)
	go.uber.org/zap@v1.19.1/sugar.go:204
main.main()
	k8c.io/kubermatic/v2/cmd/etcd-launcher/main.go:124 +0x96a
```

cmd/etcd-launcher/main.go
furkhat: make sense I'll remove this.

furkhat:
> Edit: Or, different scenario, what happens when all three etcd pods are rescheduled at the same time (without losing their data directories)? The code will always fail at `getMemberByName()` since it's now called before the etcd process is started, and as such no etcd cluster will be available to query that data.

@embik you're right, my change breaks for new clusters even though it helps against pvc deletion in running clusters. That appears to be chicken-egg problem.

embik:
@furkhat I've been poking around this a bit. It seems to me that if the `PersistentVolume` gets lost/deleted, the `pvwatcher` controller kicks in and restarts the etcd StatefulSet. I haven't been able to reproduce any problem with that. So the feature works as [documented](https://docs.kubermatic.com/kubermatic/v2.19/cheat_sheets/etcd/etcd-launcher/#automated-persistent-volume-recovery).

Obviously, we already have the code in there to also recover from an empty data dir (I guess this mostly covers accidental deletion of the `PersistentVolumeClaim` resource), so we should make that code operational. I propose that we do move the logic starting at `thisMember, err := e.getMemberByName(e.podName, log)` that cleans up the membership if needed to before starting the etcd command, but it should not fail if `e.getMemberByName()` returns an error. Instead, if prints a warning and continues, like the attempt to update peer URLs:

https://github.com/kubermatic/kubermatic/blob/9c2b9dbf1ccffa0570d91911591a893d4893b91c/cmd/etcd-launcher/main.go#L114-L120

Any thoughts on that?

furkhat:
@embik thank you for sharing your findings. Before agreeing on solution there is one thing I need to understand. 
I looked into `pvwatcher` controller and it's very interesting that it **does not delete** etcd member corresponding to lost PV but restarts all the etcd member pods. And the etcd member that lost PV will **starts with empty dir** but it is **not removed** and re-added to the cluster (because of buggy code I was aiming to fix here). And it still looks to work.

Etcd documentation states that if data directory is corrupt the member should be deleted and then added with empty dir. But pvwatcher+etcdlauncher don't do and I wonder if we might introduce corrupt state in the cluster by this. What do you think?

embik:
@furkhat very good question. I'd need to check etcd internals to figure out why it's working. I'd assume it's because the WAL is reset on the healthy etcd nodes when they restart and that allows the empty node to re-sync. But you're right that that is not the recommendation from upstream. I need to do some research on when this was introduced and why.

embik:
@furkhat After some research, I think the PV recovery process was supposed to trigger the empty data dir check as well, but in my tests it did not seem to happen. I'll test again, but it looks like this is unintentionally relying on the WAL log being restarted when the etcd StatefulSet gets reset.

I would like to work on introducing another test case that covers the PVC recovery as well. Potentially, I would also like to look into changing the `pvwatcher` controller to not reset the complete StatefulSet if we can void it.

Would you like me deferring to you on that? Do you want to keep working on this PR to only cover the specific fix? I'd like to coordinate so we don't do duplicate work.

furkhat:
@embik I can polish current approach to allow recovery from pvc deletion. The rest is best if you work on.

PS: Sorry for late response, I was away/vacationing.

embik:
@furkhat this should be superseded by #9600. Can you check if that PR results in the expected behaviour for you?

PR #9023, opened 2022-02-16
Add support for dual-stack pods & services CIDR

### LINKED ISSUE #9005: Allow dual-stack user clusters in k8s control plane
Allow passing dual-stack CIDRs for pod CIDR and services CIDR to enable dual-stack in k8s control plane for KKP user clusters.

rastislavs:
This PR is the initial step towards [dual-stack user clusters](https://kubernetes.io/docs/concepts/services-networking/dual-stack/) on KKP. It allows configuring dual-stack pod & services CIDR (one IPv4 + one IPv6, IPv4 has to be the primary one, as it defines the address family for the apiserver).

It also makes Cilium CNI dual-stack compatible, even though more work on this fill follow as well (mainly making the cluster-pool-mask-size configurable).

Note that in case that Cilium CNI is used, we are not passing `--allocate-node-cidrs` and related flags to controller-manager anymore, as Cilium uses its own node IPAM. On the other hand, `--service-cluster-ip-range` is passed for non-Cilium CNIs (it was missing, could affect node IPAM in case of overlapping pod/service subnets).

Fixes #9005 

**Documentation**:

```release-note
Add support for dual-stack pods & services CIDR.
```


addons/cilium/cilium.yaml
```diff
@@ -82,7 +83,11 @@ data:
 
   # Enable IPv6 addressing. If enabled, all endpoints are allocated an IPv6
   # address.
```
PratikDeoghare, line 85: The check would fail if somebody wanted two different IPv4 ranges for their pods? 
We don't allow setting more than one range from the UI so not sure if this will ever be an issue.

rastislavs: that is not allowed in k8s - and is prevented by KKP validation webhook

PratikDeoghare: Cool. Good to know.

PratikDeoghare (approved):

PR #9025, opened 2022-02-16
ensure the old webhook volume is removed from the master-ctrl-mgr

xrstf:
I noticed thhat we (I) never removed the `volume` statement from the Deployment.

```release-note
NONE
```


embik (approved):
/approve
/lgtm

PR #9026, opened 2022-02-16
Fix missing ClusterRole to give webhook access to clustertemplates - Test Release 2.20

xrstf:
The webhook needs to read ClusterTemplates.

```release-note
NONE
```


embik (approved):
/approve
/lgtm

PR #9027, opened 2022-02-16
WIP - use GOCACHE for non-amd64 architectures

xrstf:
Disclaimer: I have no idea if this will work. I tried the fisrt part (assembling the gocaches), but if the buildah stuff works... noooo idea.

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -18,177 +18,177 @@ presubmits:
   # unit tests
   #########################################################
 
-  - name: pre-kubermatic-test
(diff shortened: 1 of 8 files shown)


xrstf:
Okay, I give up for now.

I tried all possible combinations of build flags, I tried setting `GODEBUG` according to the documentation to make Go output information regarding cache usage (but didn't work), etc.pp.

It _seems_ that _some_ of the caching is working, as I do not see Go's stdlib being compiled anymore. But all 3rd party code cache is ignored.

If someone wants to pick this up and try again, have fun!

PR #9030, opened 2022-02-17
Update helm Chart.yaml apiVersions to v2

### LINKED ISSUE #8962: Helm default values for nginx-ingress-controller are not picked up by Flux
### What happened?

I was tinkering around with Flux and KKP and tried to install the nginx-ingress-controller Helm chart with the following manifest (using only the default values provided by the Helm chart):

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: mgmt-nginx-ingress-controller
  namespace: nginx-ingress-controller
spec:
  chart:
    spec:
      chart: charts/nginx-ingress-controller
      sourceRef:
        kind: GitRepository
        name: kubermatic
        namespace: kubermatic
  install:
    createNamespace: true
  interval: 5m
  releaseName: mgmt-nginx-ingress-controller
```

However the installation failed with the error message "Helm install failed: Service "mgmt-nginx-ingress-controller-ingress-nginx-controller-admission" is invalid: metadata.name: Invalid value: "mgmt-nginx-ingress-controller-ingress-nginx-controller-admission": must be no more than 63 characters".

The reason for this error is that Flux did not pick up the default values (more specifically the value for [nginx.fullnameOverride](https://github.com/kubermatic/kubermatic/blob/v2.19.0/charts/nginx-ingress-controller/values.yaml#L16)) of the Helm chart. Because the `Chart.yaml` uses the apiVersion v1 (I think at least) Flux falls back to strict Helm 2 behavior and therefore is not aware of the alias "nginx" defined in the dependencies of the `Chart.yaml` and ignores all default values.

The Helm 3 CLI is not that strict and will still [load and render](https://v3.helm.sh/docs/faq/changes_since_helm2/#consolidation-of-requirementsyaml-into-chartyaml) the Helm chart correctly - `helm lint` however also emits an error "[ERROR] Chart.yaml: dependencies are not valid in the Chart file with apiVersion 'v1'. They are valid in apiVersion 'v2'".

Possible solutions are:
1. Change the apiVersion to v2 (the upstream Helm chart is also using [apiVersion v2](https://github.com/kubernetes/ingress-nginx/blob/helm-chart-4.0.9/charts/ingress-nginx/Chart.yaml#L1))
2. Add file `requirements.yaml` with the same alias as defined in `Chart.yaml`

Since the nginx-ingress-controller Helm chart is currently the only one defining an alias and the kubermatic-installer as well as manual installations with the Helm CLI are not affected, this is not a big deal. However since we are in the process of switching to upstream Helm charts (where possible) I wanted to point out this behavior as it could affect other Helm charts as well. Would one of the proposed solutions work for you?

### How to reproduce?

```
kubectl create namespace kubermatic
kubectl create namespace nginx-ingress-controller

flux install
flux -n kubermatic create source git kubermatic \
  --url=https://github.com/kubermatic/kubermatic.git \
  --tag=v2.19.0
flux -n nginx-ingress-controller create helmrelease mgmt-nginx-ingress-controller \
  --source=GitRepository/kubermatic.kubermatic \
  --chart=charts/nginx-ingress-controller

kubectl -n nginx-ingress-controller describe helmrelease mgmt-nginx-ingress-controller
```

### Environment

- Kubernetes Version: v1.21.1
- KKP Version: v2.19.0
- Flux Version: v0.26.2
- Helm CLI Version: v3.8.0

wurbanski:
`dependencies` field in Chart.yaml is only supported in apiVersion: v2

Fixes #8962 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/cert-manager/Chart.yaml
+++ b/charts/cert-manager/Chart.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: v1
+apiVersion: v2
 name: cert-manager
 version: v9.9.9-dev
 appVersion: v1.6.1
--- a/charts/cert-manager/test/test-certmanager.sh
+++ b/charts/cert-manager/test/test-certmanager.sh
@@ -33,7 +33,7 @@ helm upgrade \
   --atomic \
   cert-manager charts/cert-manager/
 
-if ! which cmctl; then 
+if ! which cmctl; then
   echodate "Downloading cmctl..."
   OS=$(go env GOOS); ARCH=$(go env GOARCH); curl -sLo cmctl.tar.gz https://github.com/jetstack/cert-manager/releases/latest/download/cmctl-$OS-$ARCH.tar.gz
   tar xzf cmctl.tar.gz
@@ -41,9 +41,8 @@ if ! which cmctl; then
   function cmctl_cleanup {
     echodate "Cleaning up..."
     rm cmctl cmctl.tar.gz
-    exit $exitcode
   }
-  trap cmctl_cleanup EXIT
+  appendTrap cmctl_cleanup EXIT
 fi
 echodate "Testing cert-manager..."
 ./cmctl check api --wait=2m
--- a/charts/logging/loki/Chart.yaml
+++ b/charts/logging/loki/Chart.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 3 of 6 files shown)


xrstf:
@wurbanski This is not failing because of the charts, this is failing because the `socat` process for tunneling the Docker mirror into the kind cluster is still dangling. Either you or me need to adjust this PR to kill all descendant processes (hack/lib.sh should already have a function for that).

PR #9033, opened 2022-02-17
reconcile konnectivity server kubeconfig on apiserver service change

### LINKED ISSUE #9035: Konnectvity e2e test fails ~1/3 times with dial timeout
Konnectivity test fails with dial timeout because there are no agents to forward traffic to.  Agents fail to register with the proxy due to authentication failure. This blocks merging of PRs that are not related to Konnectivity. Example https://github.com/kubermatic/kubermatic/pull/8836. 

This happens only in CI tests. We haven't observed this in the wild.

PratikDeoghare:
Konnectivity e2e test fails in CI in Kind because agents cannot connect to proxy-server. 
Agents cannot connect because kubeconfig used by proxy-server is stale. It is not reconciled in if apiserver-external service hanges. This PR reconciles the kubeconfig if change is detected in the service. It also re-enables the e2e test for Konnectivity. 

Fixes #9035 

```release-note
NONE
```


pkg/resources/konnectivity/secrets.go
```diff
@@ -32,8 +32,22 @@ import (
...
+				err := json.Unmarshal(oldConfData, &oldConf)
+				if err != nil {
+					return nil, fmt.Errorf("failed to unmarshal old config: %w", err)
+				}
+
+				if len(oldConf.Clusters) == 0 {
+					return nil, fmt.Errorf("old config has no clusters")
+				}
+
+				if oldConf.Clusters[0].Cluster.Server == data.Cluster().Address.URL {
```
PratikDeoghare, line 55: Should I do deep equals check down below instead? :thinking:

rastislavs: The question is what else can change in time. For sure you should not generate a new client key pair on each reconcile. But I think you should also check for root CA changes - it may expire and keys can be re-generated. Everything else seems to be constant.

rastislavs (approved):

PR #9037, opened 2022-02-18
use yaml.UnmarshalStrict

PratikDeoghare:
I noticed old `KuberamaticConfiguration` was parsed happily by `loadKubermaticConfiguration`. Strict parsing would have saved me some debugging. This PR changes all the  `use yaml.Unmarshal` to `use yaml.UnmarshalStrict`. 

NO 

```release-note
NONE
```

### DIFF
--- a/cmd/image-loader/operator.go
+++ b/cmd/image-loader/operator.go
@@ -38,7 +38,7 @@ func loadKubermaticConfiguration(log *zap.SugaredLogger, filename string) (*kube
 	}
 
 	config := &kubermaticv1.KubermaticConfiguration{}
-	if err := yaml.Unmarshal(content, &config); err != nil {
+	if err := yaml.UnmarshalStrict(content, &config); err != nil {
 		return nil, fmt.Errorf("failed to parse file as YAML: %w", err)
 	}
 
--- a/cmd/node-resource-documenter/documenter.go
+++ b/cmd/node-resource-documenter/documenter.go
@@ -59,25 +59,25 @@ func (d *documenter) scanAll() error {
 
 func (d *documenter) scanManifest(manifest []byte) error {
 	var u unstructured.Unstructured
-	if err := yaml.Unmarshal(manifest, &u); err != nil {
+	if err := yaml.UnmarshalStrict(manifest, &u); err != nil {
 		return err
 	}
 	switch u.GetKind() {
 	case "Deployment":
 		var ad appsv1.Deployment
-		if err := yaml.Unmarshal(manifest, &ad); err != nil {
+		if err := yaml.UnmarshalStrict(manifest, &ad); err != nil {
 			return err
 		}
 		d.addSpec(ad.Spec.Template.Spec)
 	case "DaemonSet":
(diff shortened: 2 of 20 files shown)


cmd/node-resource-documenter/documenter.go
```diff
@@ -59,25 +59,25 @@ func (d *documenter) scanAll() error {
 
 func (d *documenter) scanManifest(manifest []byte) error {
 	var u unstructured.Unstructured
-	if err := yaml.Unmarshal(manifest, &u); err != nil {
+	if err := yaml.UnmarshalStrict(manifest, &u); err != nil {
```
xrstf, line 62: (How) does strict unmarshalling work on / affect unstructured objects?

pkg/util/yamled/document.go
```diff
@@ -347,7 +347,7 @@ func (d *Document) normalize() interface{} {
 	encoded, _ := yaml.Marshal(d.root)
 
 	var normal interface{}
-	_ = yaml.Unmarshal(encoded, &normal)
+	_ = yaml.UnmarshalStrict(encoded, &normal)
```
xrstf, line 350: How does this change the behaviour?

xrstf:
Have you considered looking at the golangci-lint stuff and checking if we can have a forbid list of functions? Then we could by default forbid the unstrict Unmarshal function.

PratikDeoghare:
@xrstf I will look at it now. Should be possible to do.

pkg/util/yamled/document.go
PratikDeoghare: Not in any way that I can imagine.

xrstf (approved):
/approve

cmd/image-loader/operator.go
```diff
@@ -38,7 +38,7 @@ func loadKubermaticConfiguration(log *zap.SugaredLogger, filename string) (*kube
 	}
 
 	config := &kubermaticv1.KubermaticConfiguration{}
-	if err := yaml.Unmarshal(content, &config); err != nil {
+	if err := yaml.UnmarshalStrict(content, &config); err != nil {
```
lsviben, line 41: I have a question about this, from what I understand, this(and others like this) will fail if there is a field that does not fit the struct. 

So what in the case when somebody wants to add a new field to the CRD like KubermaticConfiguration, and use it in for example in seed-controller-manager. To test this on dev, the change has to be present in all the components, otherwise they will be throwing errors?

PratikDeoghare: Good point. This is not forward compatible.

xrstf: I'm not quite sure I understand the issue here. When you add a new field to the KubermaticConfiguration, your local code has it, so UnmarshalStrict() should just work (since we have openapi CRDs now, there is no chance of _storing_ new fields on dev before something is merged).

lsviben: ahh good point, we have openapi CRDs so its the same

PR #9044, opened 2022-02-21
fix sets resource requirements. Null pointer exception

zreigz:
**What does this PR do / Why do we need it**: This PR prevents null pointer exception. Our customer noticed a problem with release 2.18.6 where this part of the code was backported: https://support.kubermatic.com/helpdesk/tickets/3632#

Fixes #

```release-note
NONE
```


zreigz:
I will apply on master

PR #9045, opened 2022-02-21
Fix applying incomplete resource requirements

zreigz:
This PR prevents null pointer exception. Our customer noticed a problem with release 2.18.6 where this part of the code was backported: https://support.kubermatic.com/helpdesk/tickets/3632#

Fixes #

```release-note
Fix applying resource requirements when using incomplete overrides (e.g. specifying only limits, but no request for a container)
```

### DIFF
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -1243,11 +1243,12 @@ func SetResourceRequirements(containers []corev1.Container, defaultRequirements,
 		}
 	}
 	for k, v := range overrides {
-		if v.Requests == nil {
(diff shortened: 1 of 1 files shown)


kron4eg (approved):
/approve

embik (approved):

xrstf:
Just for reference, this gets triggered when an admin for example uses an override like `{"prometheus":{"limits":....}}` with no `requests`, but our defaults would not contain a `prometheus` container, so `defaultRequirements[k]` is nil.

PR #9047, opened 2022-02-21
Reconcile kubernetes service endpoints in KKP user-cluster-controller

### LINKED ISSUE #9039: Unable to reconcile kubernetes svc endpoints upon apiserver-external svc change
### What happened?
E2e tests uncovered an issue with reconciling the `kubernetes` service endpoints of the user clusters. If the endpoint changes (e.g. if the `apiserver-external` service is re-created and a new nodeport is allocated), the endpoint is not reconciled to the new value. The apiserver logs errors like:

```
E0217 09:06:22.295548       1 controller.go:152] Unable to remove old endpoints from kubernetes service: StorageError: key not found, Code: 1, Key: /registry/masterleases/54.194.246.235, ResourceVersion: 0, AdditionalErrorMsg: 
```

or

```
E0217 17:02:03.887760       1 controller.go:152] Unable to remove old endpoints from kubernetes service: no master IPs were listed in storage, refusing to erase all endpoints for the kubernetes service
```

### What should happen?
Upon apiserver-external svc change (nodeport change), the `kubernetes` service endpoints should be properly reconciled to the new value.

### How to reproduce?
With the nodeport expose strategy, delete the `apiserver-external` service in the seed. That will cause new nodeport allocation and trigger this issue.

rastislavs:
Kubernetes by default uses kubernetes service endpoint reconcilers that are designed to work with multiple different apiserver IP addresses. In case of KKP, all apiserver instances share the same advertise-address, which causes issues in some edge scenarios (e.g. hitting issues like https://github.com/kubernetes/kubernetes/issues/76146 & https://github.com/kubernetes/kubernetes/issues/104252 occasionally in our e2e tests) and breaks the user cluster is the apiserver-external service is removed (https://github.com/kubermatic/kubermatic/issues/9039).

As KKP has special needs for configuring the kubernetes service endpoints given its expose strategies, it makes more sense to reconcile them by KKP itself. With this PR, the KKP user-cluster-controller-manager becomes the owner of the endpoints of the kubernetes service (by setting the apiserver flag `--endpoint-reconciler-type=none`) and reconciles them according to the used KKP expose strategy needs.

This also allows for future enhancements (not part of this PR), such as:
- Having multiple endpoints for kubernetes service if apiserver/nodeport-proxy is exposed on a LB address that resolves to multiple IP addresses (e.g. in case of AWS), thus aching better HA,
- By the Tunneling expose strategy, the tunneling agents can run as standard pods, without the need to bind on arbitrary IP address in the host network namespace. That would mean simplification and better stability.
- Ability to expose the apiserver via dual-stack for in-cluster pods, even if not supported in the given kubernetes version.

Further notes: 
- To be aligned with the upstream endpoint reconcilers, we reconcile EndpointSlices as well. This also makes sure that in existing clusters we do not end up having multiple EndpointSlices for the same service after this switch.
- This change is effective only on clusters with k8s version 1.21 and higher, as EndpointSlices became stable in v1.21. Making this change compatible with older versions is not worth the effort given that v1.20 end of life is coming soon.
- This change also unifies the flags passed to the apiserver across expose strategies. There is no need for passing `kubernetes-service-node-port` and making the kubernetes service a NodePort service in the user cluster - ClusterIP actually makes much more sense and aligns with other similar platforms.

Fixes #9039 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -666,19 +666,20 @@ func (r *reconciler) reconcileServices(ctx context.Context) error {
 }
 
 func (r *reconciler) reconcileEndpoints(ctx context.Context, data reconcileData) error {
-	if data.reconcileK8sSvcEndpoints {
-		epCreators := []reconciling.NamedEndpointsCreatorGetter{
-			kubernetesresources.EndpointsCreator(data.clusterAddress),
-		}
-		if err := reconciling.ReconcileEndpoints(ctx, epCreators, metav1.NamespaceDefault, r.Client); err != nil {
-			return fmt.Errorf("failed to reconcile Endpoints: %w", err)
-		}
-		epSliceCreators := []reconciling.NamedEndpointSliceCreatorGetter{
-			kubernetesresources.EndpointSliceCreator(data.clusterAddress),
-		}
-		if err := reconciling.ReconcileEndpointSlices(ctx, epSliceCreators, metav1.NamespaceDefault, r.Client); err != nil {
-			return fmt.Errorf("failed to reconcile EndpointSlices: %w", err)
-		}
+	if !data.reconcileK8sSvcEndpoints {
+		return nil
+	}
+	epCreators := []reconciling.NamedEndpointsCreatorGetter{
+		kubernetesresources.EndpointsCreator(data.clusterAddress),
+	}
+	if err := reconciling.ReconcileEndpoints(ctx, epCreators, metav1.NamespaceDefault, r.Client); err != nil {
+		return fmt.Errorf("failed to reconcile Endpoints: %w", err)
+	}
+	epSliceCreators := []reconciling.NamedEndpointSliceCreatorGetter{
+		kubernetesresources.EndpointSliceCreator(data.clusterAddress),
+	}
+	if err := reconciling.ReconcileEndpointSlices(ctx, epSliceCreators, metav1.NamespaceDefault, r.Client); err != nil {
+		return fmt.Errorf("failed to reconcile EndpointSlices: %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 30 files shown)


pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -662,6 +665,24 @@ func (r *reconciler) reconcileServices(ctx context.Context) error {
 	return nil
 }
 
+func (r *reconciler) reconcileEndpoints(ctx context.Context, data reconcileData) error {
+	if data.reconcileK8sSvcEndpoints {
```
mate4st: Nit: return immediately if condition is false to reduce indentation.
```
if !data.reconcileK8sSvcEndpoints { 
    return nil
}

File: pkg/controller/user-cluster-controller-manager/resources/resources/kubernetes/endpoints.go [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	// ServiceName is the name of the kubernetes service in the default namespace.
```
mate4st: This is the Endpont(slice) name not the service name, right? I mean maybe it makes sense because they are bound together.

rastislavs: Right, we intentionally keep the name same as the service name to be aligned with the upstream reconcilers. But I will make it more obvious in the code and comments.

moadqassem (approved):
/approve
/lgtm

rastislavs:
/hold
looks like cilium e2e test is broken

PR #9053, opened 2022-02-22
Add seccomp profiles to all pod specs created via reconciling

### LINKED ISSUE #9052: Add SeccompProfile to control plane components in Seed cluster

While user cluster components have all been updated with a seccomp profile to reduce the amount of allowed syscalls and improve isolation, control plane components in the seed cluster haven't been updated to do so as well to secure the Seed itself.

### Acceptance criteria

- All control plane components that support it run with seccompProfile=runtimeDefault

embik:

Adds seccomp profiles to every PodSpec that gets defaulted during a reconcile. Also adds a test to check that all control plane components are using the seccomp profile.

Fixes #9052

**Documentation**:

```release-note
All pods created by KKP are assigned the `RuntimeDefault` seccomp profile
```

### DIFF
--- a/pkg/resources/reconciling/wrapper.go
+++ b/pkg/resources/reconciling/wrapper.go
@@ -146,14 +146,10 @@ func DefaultPodSpec(oldPodSpec, newPodSpec corev1.PodSpec) (corev1.PodSpec, erro
 	// set KKP specific defaults for every Pod created by it
 
 	if newPodSpec.SecurityContext == nil {
-		newPodSpec.SecurityContext = &corev1.PodSecurityContext{
-			SeccompProfile: &corev1.SeccompProfile{
-				Type: corev1.SeccompProfileTypeRuntimeDefault,
-			},
-		}
+		newPodSpec.SecurityContext = &corev1.PodSecurityContext{}
 	}
 
-	if newPodSpec.SecurityContext != nil && newPodSpec.SecurityContext.SeccompProfile == nil {
+	if newPodSpec.SecurityContext.SeccompProfile == nil {
 		newPodSpec.SecurityContext.SeccompProfile = &corev1.SeccompProfile{
 			Type: corev1.SeccompProfileTypeRuntimeDefault,
 		}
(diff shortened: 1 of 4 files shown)


kron4eg:
@embik it looks like you want to use https://github.com/kubermatic/kubermatic/blob/5108893a10277e75563c75dc186520639fd418e7/pkg/resources/reconciling/wrapper.go#L113 but you didn't knew about its existence

embik:
> @embik it looks like you want to use
> 
> https://github.com/kubermatic/kubermatic/blob/5108893a10277e75563c75dc186520639fd418e7/pkg/resources/reconciling/wrapper.go#L113
> 
> but you didn't knew about its existence

I was thinking of using that, but this is not used for a lot of things (for example, image pull secrets are probably injected into everything as well manually) - Do you think we should add the seccomp profile there instead?

kron4eg:
> Do you think we should add the seccomp profile there instead?

Yeah, if we are serious about having seccomp by default always and to not forget it somewhere in the future.

embik:
Good point. Let me rework the PR then.

pkg/resources/reconciling/wrapper.go
```diff
@@ -142,6 +143,22 @@ func DefaultPodSpec(oldPodSpec, newPodSpec corev1.PodSpec) (corev1.PodSpec, erro
...
+
+	if newPodSpec.SecurityContext == nil {
+		newPodSpec.SecurityContext = &corev1.PodSecurityContext{
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
+			},
+		}
+	}
+
+	if newPodSpec.SecurityContext != nil && newPodSpec.SecurityContext.SeccompProfile == nil {
```
kron4eg:
```suggestion
	if newPodSpec.SecurityContext.SeccompProfile == nil {
```

checking here `newPodSpec.SecurityContext` for nilness is useless since you just initialized it above

embik: Good point. I initially misunderstood the function and was checking `oldPodSpec`, but now it does not make sense anymore to check for `nil`.

embik:
/hold

I want CI to run against this, but I have not finished adjusting code comments to reflect on this yet. So the code itself is ready for being reviewed, but I'll work on properly documenting this before merging it.

kron4eg (approved):
/approve

kron4eg (approved):

PR #9054, opened 2022-02-22
Fix missing features and bugs for nutanix csi

mate4st:

- Fixes missing dereference for nutanix port pointer
- Fixes missing default storage class
- Allow storageclass configuration

```release-note
NONE
```


addons/default-storage-class/storage-class.yaml
```diff
@@ -124,3 +124,26 @@ provisioner: csi.hetzner.cloud
...
+provisioner: csi.nutanix.com
+parameters:
+  storageType: NutanixVolumes
+  csi.storage.k8s.io/provisioner-secret-name: ntnx-secret
+  csi.storage.k8s.io/provisioner-secret-namespace: kube-system
+  csi.storage.k8s.io/node-publish-secret-name: ntnx-secret
+  csi.storage.k8s.io/node-publish-secret-namespace: kube-system
+  csi.storage.k8s.io/controller-expand-secret-name: ntnx-secret
+  csi.storage.k8s.io/controller-expand-secret-namespace: kube-system
+  storageContainer: {{ default "Default" .Params.storageContainer | quote }}
```
embik: Can a user even change this?

mate4st: No this is a copy paste leftover from kubeone. I thought i removed all

pkg/addon/template.go
```diff
@@ -216,6 +218,18 @@ type MLASettings struct {
...
+
+	//vsphere
+	// StoragePolicy is the storage policy to use for vsphere csi addon
+	StoragePolicy string
+
+	//nutanix
+	StorageContainer        string
+	Fstype                  string
+	SsSegmentedIscsiNetwork *bool
+}
```
embik: Can we maybe change this to something like:

```go
type CSIOptions struct {
     Vsphere *VsphereCSIOptions
     Nutanix *NutanixCSIOptions
}

type VsphereCSIOptions struct {
    StoragePolicy string
}

type NutanixCSIOptions struct {
    StorageContainer        string
    Fstype                  string
    SsSegmentedIscsiNetwork *bool
}
```

I don't think it's a good idea to have a shared struct holding all provider-specific options.

embik (changes requested):

pkg/addon/template.go
mate4st: As this is only used for the templating and for now we do not have a lot of these features its okay in my opinion.

embik: Not a huge fan of it to be honest, but okay :D

embik (approved):
/approve
/lgtm

xrstf (approved):
/approve

Issue #8489, opened 2021-12-17
Create KubeVirt VMs in dedicated cluster namespace


mfranczy:
We allow users to create a virtual machine in any namespace on the KubeVirt infrastructure cluster. We have plan to provide a better isolation between user clusters created on KubeVirt. The first step is to create a dedicated cluster namespace on each cluster creation 

Those are listed components that require a change:

**KKP**: -it should add the `metadata.namespace` env variable in the `machine-controller` pod [KKP deployment for machine-controller](https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/machinecontroller/deployment.go#L197).  Goal is that in `machine-controller`, [link](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/kubevirt/provider.go#L350)  we are able to get the namespace where the `machine-controller` is running. [Doc](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)

**KKP KubeVirt provider** - it should create a `cluster-<id>` namespace on the infra cluster [link to kubevirt cloud provider](https://github.com/kubermatic/kubermatic/tree/master/pkg/provider/cloud/kubevirt)

**Machine Controller KubeVirt provider**  - should pass appropriate namespace to the VM spec [link](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/kubevirt/provider.go#L350). Note that MC already runs in the user cluster namespace on the seed. Therefore the name of the namespace doesn't have to be passed in the config.

**KubeVirt Cloud Controller Manager** - currently it takes a namespace from kubeconfig. We should add an option to set a namespace where CCM should try to find VMs. If the option is empty the namespace should be taken from the kubeconfig. That could be a good place where we can add namespace option [link](https://github.com/kubermatic/cloud-provider-kubevirt/blob/master/pkg/cloudprovider/kubevirt/cloud.go#L30-L35). The CCM config is being created [here](https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/cloudconfig/configmap.go#L266-L270)

**KubeVirt CSI driver operator** - pass appropriate namespace for csi config, https://github.com/kubermatic/kubermatic/blob/master/addons/csi/kubevirt/csi-driver-operator.yaml#L373 and fix CSI access initialization https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/kubevirt/reconciler.go#L52

Acceptance criteria:
[...]

PR #9056, opened 2022-02-22
Implement Provider Reconciling for KubeVirt

### LINKED ISSUE #8144: Implement Provider Reconciling for KubeVirt

hdurand0710:

Implement the Provider Reconciling for Kubevirt for the CSI Access (Role and Rolebinding). As discussed, the needed ServiceAccount is not reconciled in this process, only created at cluster creation time.
This prepares also for the namespace creation in #8489 that will be reconciled too.

Fixes #8144

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/provider/cloud/kubevirt/provider.go
+++ b/pkg/provider/cloud/kubevirt/provider.go
@@ -118,9 +118,6 @@ func (k *kubevirt) GetClientWithRestConfigForCluster(cluster *kubermaticv1.Clust
 	if err != nil {
 		return nil, nil, err
 	}
-	if err != nil {
-		return nil, nil, err
-	}
 
 	return client, restConfig, nil
 }
--- a/pkg/provider/cloud/kubevirt/provider.go
+++ b/pkg/provider/cloud/kubevirt/provider.go
@@ -27,7 +27,9 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type kubevirt struct {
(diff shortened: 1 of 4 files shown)


pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -84,6 +104,27 @@ func (k *kubevirt) ValidateCloudSpecUpdate(oldSpec kubermaticv1.CloudSpec, newSp
...
+		return nil, nil, err
+	}
+
+	client, restConfig, err := NewClientWithRestConfig(kubeconfig)
+	if err != nil {
+		return nil, nil, err
+	}
+	if err != nil {
+		return nil, nil, err
+	}
```
xrstf: Why is this here? ^^

hdurand0710: Thanks for noticing! Removed

pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -72,8 +74,26 @@ func (k *kubevirt) ValidateCloudSpec(spec kubermaticv1.CloudSpec) error {
 	return nil
 }
 
-func (k *kubevirt) InitializeCloudProvider(c *kubermaticv1.Cluster, p provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
-	return c, nil
+func (k *kubevirt) InitializeCloudProvider(cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
+	return k.reconcileCluster(cluster, update)
```
xrstf, line 78: Normally, the `InitializeCloudProvider` should be a no-op for cluster that have already been initialized. I can see that this is not talking to any cloud API, but I am still a bit worried that it might be too slow to reconcile KubeVirt clusters basically on every single cluster reconciliation.

xrstf: Nevermind, the InitializeCloudProvider isn't even used for providers that have a Reconcile function, so it doesn't matter.

xrstf (approved):
/approve

PR #9061, opened 2022-02-23
Use cluster name as cluster-id in cloud-config for vSphere

### LINKED ISSUE #8705: Wrong ClusterID in vSphere cloud config
**Problem**
Container Volumes created by the CSI are marked with the wrong cluster name in vSphere. This works, but makes it very difficult to tell which of the volumes, that show up in vSphere, belong to which user cluster.

**Cause**
The ClusterID field in the cloud config for vSphere is currently populated with the value from the cluster field in the datacenter configuration

https://github.com/kubermatic/kubermatic/blob/03641d064f518bbf26622d531d2bcca1e3bdbba2/pkg/resources/cloudconfig/configmap.go#L312

For the latter the [documentation](https://docs.kubermatic.com/kubermatic/v2.18/tutorials_howtos/project_and_cluster_management/seed_cluster/) states that it should (a) contain the name of the vSphere compute cluster (which is completely unrelated to the CSI) and (b) is currently unused.

The ClusterID value in the cloud config is used by vSphere CSI to set the metadata of the created container volumes. This causes the above mentioned problem.

**Solution**
The line linked above should be 

```go
ClusterID:        cluster.Name
```

This is the value used in other cloud configs (e.g. AWS).

A quick search of the CCM sources and documentation yielded no usage of the `cluster-id` field in the CCM. The CSI only seems to use the value setting the above mentioned metadata. I think fixing this should cause no problems in existing deployments.

ahmedwaleedmalik:

Use cluster name as cluster-id in cloud-config for vSphere. `cluster-id` is only consumed by CSI driver and it MUST be unique across each Kubernetes cluster.

Fixes #8705 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/kubectl/kubectl.go
+++ b/pkg/util/kubectl/kubectl.go
@@ -26,6 +26,8 @@ import (
 // BinaryForClusterVersion returns the full path to a kubectl binary
 // that shall be used to communicate with a usercluster. An error is
 // returned if no suitable kubectl can be determined.
+// We take advantage of version skew policy for kubectl, v1.1.1 would support v1.2.x and v1.0.x, to ship
+// only mandatory variants for kubectl.
 func BinaryForClusterVersion(version *semver.Version) (string, error) {
 	var binary string
 
@@ -33,7 +35,7 @@ func BinaryForClusterVersion(version *semver.Version) (string, error) {
 	case 20:
 		binary = "kubectl-1.20"
 	case 21:
-		binary = "kubectl-1.21"
+		binary = "kubectl-1.22"
 	case 22:
 		binary = "kubectl-1.22"
 	default:
--- a/pkg/util/kubectl/kubectl.go
+++ b/pkg/util/kubectl/kubectl.go
@@ -33,7 +33,7 @@ func BinaryForClusterVersion(version *semver.Version) (string, error) {
 	case 20:
 		binary = "kubectl-1.20"
 	case 21:
-		binary = "kubectl-1.22"
+		binary = "kubectl-1.21"
 	case 22:
 		binary = "kubectl-1.22"
 	default:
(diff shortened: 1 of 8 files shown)


pkg/util/kubectl/kubectl.go
```diff
@@ -33,7 +33,7 @@ func BinaryForClusterVersion(version *semver.Version) (string, error) {
 	case 20:
 		binary = "kubectl-1.20"
 	case 21:
-		binary = "kubectl-1.22"
```
xrstf: Why this change?

ahmedwaleedmalik: Why'd you want to run the binary for `1.22` if the minor version is `1.21`? I'd like to understand the need behind that.

I was getting errors when running seed-controller-manager locally; unable to apply addon, file(binary) not found.

xrstf: Because we do not ship kubectl binaries for all Kubernetes versions, and instead make use of the version skew policy. https://github.com/kubermatic/kubermatic/blob/master/Dockerfile#L23-L24

> I was getting errors when running seed-controller-manager locally; unable to apply addon, file(binary) not found.

And you have kubectl-1.21 binary on your local machine, but no kubectl-1.22 one? :-D

In any way, this would be totally unrelated to vSphere changes, no?

xrstf: cc @xmudrii

ahmedwaleedmalik: My idea was to not create another PR for such a simple change. But anyway, I get your point now.

xmudrii:
/hold
for discussion

pkg/util/kubectl/kubectl.go
ahmedwaleedmalik: Updated

xrstf (approved):
/approve

PR #9084, opened 2022-02-23
re-enable vsphere integration tests

xrstf:
Now that we have a working vSphere setup, this PR undoes the changes from #8841.

```release-note
NONE
```

### DIFF
--- a/pkg/provider/cloud/vsphere/folder_test.go
+++ b/pkg/provider/cloud/vsphere/folder_test.go
@@ -64,18 +64,14 @@ func TestProvider_GetVMFolders(t *testing.T) {
 		name            string
 		dc              *kubermaticv1.DatacenterSpecVSphere
 		expectedFolders sets.String
-		// If we check for folders on the root we might have other tests creating folders, so we have to skip that check
-		skipAdditionalFoldersCheck bool
 	}{
 		{
-			name:                       "successfully-list-default-folders",
-			skipAdditionalFoldersCheck: true,
-			dc:                         getTestDC(),
+			name: "successfully-list-default-folders",
+			dc:   getTestDC(),
 			expectedFolders: sets.NewString(
 				path.Join("/", vSphereDatacenter, "vm"),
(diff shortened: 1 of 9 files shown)


pkg/provider/cloud/vsphere/folder_test.go
```diff
@@ -64,30 +64,26 @@ func TestProvider_GetVMFolders(t *testing.T) {
 		name            string
 		dc              *kubermaticv1.DatacenterSpecVSphere
 		expectedFolders sets.String
-		// If we check for folders on the root we might have other tests creating folders, so we have to skip that check
-		skipAdditionalFoldersCheck bool
```
xrstf, line 68: There was a folder named after a cluster in our vCenter. Assuming that can always happen, I decided to make all folder tests to only check for a subset, not equality.

pkg/provider/cloud/vsphere/network_test.go
```diff
@@ -33,16 +33,10 @@ func TestGetPossibleVMNetworks(t *testing.T) {
...
-				},
-				{
-					AbsolutePath: "/kubermatic-e2e/network/e2e-networks/subfolder/e2e-distributed-switch-uplinks",
-					RelativePath: "e2e-networks/subfolder/e2e-distributed-switch-uplinks",
-					Type:         "DistributedVirtualPortgroup",
-					Name:         "e2e-distributed-switch-uplinks",
+					AbsolutePath: "/dc-1/network/VM Network",
+					RelativePath: "VM Network",
+					Type:         "Network",
+					Name:         "VM Network",
```
xrstf, line 39: Those values look weird, but it's what we get from the API. No idea if this is correct.

moadqassem: They are correct. Our vCenter atm doesn't use dvSwitch they use svswitch.

moadqassem (approved):
/approve 
/lgtm

PR #9105, opened 2022-02-24
[release/v2.18] Fix LB expose strategy for LBs with external DNS names instead of IPs (#7933)

xrstf:
This just tripped us up when testing 2.18 and I think it's worth it to backport this, just so we support AWS LB's in case we ever do another 2.18.x release.

```release-note
Fix LoadBalancer expose strategy for LBs with external DNS names instead of IPs
```


xrstf:
@rastislavs You made the original PR, WDYT about backporting this to 2.18?

(If you got another notification for this exact same comment, disregard it. Wrong PR. ^^)

PR #9112, opened 2022-02-24
[release/v2.20] Remove CLI flags for backup containers / legacy backup options (#9003)

xrstf:
This is a manual backport of #9003, with additions to the CRD migration.

Since we cannot really choose a default backup destination for the customer and we also do not want to create new destinations based on some naming scheme (because those would basically be permanent), we offload the effort to migrate to the new settings onto the admin. This means the preflight checks will catch if there are still outdated configuration options.

```release-note
The etcd-backup-related containers are now loaded dynamically from the KubermaticConfiguration, the relevant CLI flags like `-backup-container=<file>` have been removed.
The deprecated configuration options `KubermaticConfiguration.spec.seedController.backupRestore` and `Seed.spec.backupRestore` have been removed. Please migrate to `Seed.spec.etcdBackupRestore`.
```


pkg/install/crdmigration/preflight.go
```diff
@@ -382,3 +387,19 @@ func validateCRDsExist(ctx context.Context, logger logrus.FieldLogger, opt *Opti
...
+
+func validateEtcdBackupConfiguration(ctx context.Context, logger logrus.FieldLogger, opt *Options) error {
+	logger.Info("Validating etcd backup configuration…")
+
+	for name, seed := range opt.Seeds {
+		if seed.Spec.BackupRestore != nil {
+			return fmt.Errorf("Seed %s uses the deprecated `backupRestore` configuration; please migrate to the `etcdBackupRestore` configuration before proceeding, as the deprecated options have been removed in KKP 2.20", name)
+		}
+
+		if seed.Spec.EtcdBackupRestore.DefaultDestination == nil {
```
lsviben, line 400: if seed.Spec.EtcdBackupRestore == nil, no need for the check or migration as the backups are not even configured

lsviben:
we should change the migration of  the EtcdBackupConfigs which dont have a destination set, and set the default destination?
https://github.com/kubermatic/kubermatic/blob/release/v2.20/pkg/install/crdmigration/clones.go#L1393

Or we can also block it in preflight check, but this could be much more time-consuming for the users as there could be a lot of them.

xrstf:
Good catch! I will have to think about how we can solve this .. :/

xrstf:
I don't think we have much of an option, other than to offload this to the admin. If we cannot even auto-migrate the backupRestore => etcdBackupRestore stuff because of the destinations, we also cannot choose a destination for each EtcdBackup object :-/

I extended the preflight check accordingly, with a tear in my eye.

PR #9113, opened 2022-02-24
Add Vsphere snapshotter

### LINKED ISSUE #9013: Implement CSI Snapshotter
CSI Snapshotter is part of the CSI Spec and CSI driver may require it. See: https://github.com/kubermatic/kubermatic/issues/8928

https://github.com/kubernetes-csi/external-snapshotter

The snapshotter should be independent of the csi driver, which does not always seem to be the case.

It consist of 3 components: 

- Snapshot CRD
- Snapshot Controller
- Validating Webhook

### Acceptance criteria
* Make sure all mentioned components are deployed together with the csi driver
* Note that some csi driver provide some of the components as part of their csi driver

mate4st:

Adds the snapshot controller and validation webhook for vsphere csi

Fixes #9013

**Documentation**:
https://github.com/kubernetes-csi/external-snapshotter
https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/master/manifests/vanilla/deploy-csi-snapshot-components.sh

```release-note
Add vSphere Snapshotter
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package csi
+package csisnaphotter
 
 import (
 	"fmt"
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package csisnaphotter
+package csisnapshotter
 
 import (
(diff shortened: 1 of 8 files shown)


pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package csi
+package csisnaphotter
```
xrstf: Typo in package name

xrstf (changes requested):
Typonista

pkg/controller/user-cluster-controller-manager/resources/resources/csi-snapshotter/secret.go
mate4st: Its obviously new component, the snap Hotter :fire:

xrstf:
Those errors will be fixed once #9420 is merged. Sorry 'bout that.

PR #9122, opened 2022-02-25
ensure new Cluster objects from templates appear in the cache before proceeding

xrstf:
During the CRD refactoring, we changed all code places in KKP to work based on controllers, so that we never have to do

```
client.Create(ctx, cluster)
// and then immediately
client.Status().Patch(ctx, cluster) // this will fail most of the time
```

This is because of the caches. However, the cluster template controller is the only place that needs to do this sequence of operations. But for this to work, it first needs to wait for each of the new Cluster objects to appear in the cache. That's why this PR re-uses parts of our reconciling framework (lines 202-214), which already contains a "wait for cache" loop. I also added a bit more logging to see what's going on.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cluster-template-controller/controller.go
+++ b/pkg/controller/seed-controller-manager/cluster-template-controller/controller.go
@@ -30,12 +30,14 @@ import (
 	"k8c.io/kubermatic/v2/pkg/provider/kubernetes"
 	kubernetesprovider "k8c.io/kubermatic/v2/pkg/provider/kubernetes"
 	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 	"k8c.io/kubermatic/v2/pkg/util/workerlabel"
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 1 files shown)


xrstf:
@zreigz PTAL :)

pkg/controller/seed-controller-manager/cluster-template-controller/controller.go
```diff
@@ -188,17 +170,57 @@ func (r *reconciler) createClusters(ctx context.Context, instance *kubermaticv1.
 
 				return fmt.Errorf("failed to create desired number of clusters. Created %d of %d", created, totalReplicas)
 			}
```
PratikDeoghare, line 172: Don't we want update the number of replicas when we succeed creating a replica? 
Shouldn't this be outside of err != nil block? 

```
				created := int64(i + 1)
				totalReplicas := instance.Spec.Replicas

				if err := r.patchInstance(ctx, instance, func(i *kubermaticv1.ClusterTemplateInstance) {
					i.Spec.Replicas = totalReplicas - created
				}); err != nil {
					return err
				}
```

xrstf: What's the point in patching the ClusterTemplateInstance n times in a loop? Either the loop succeeds and we delete the ClusterTemplateInstance or it fails and we update the spec.

PratikDeoghare: 1. Say instance.Spec.Replicas = 3 
2. r.createCluster fails on the first call. 

From the current code looks like 

instance.Spec.Replicas will be  2. It should be 3, right?

xrstf: No, it should be 2. Because on the next reconciliation, we only want to create the remaining 2 clusters. Otherwise we end up in an infinite loop.

xrstf: No, it should be 2.

@zreigz

PR #9123, opened 2022-02-25
do not attempt to reconcile Gatekeeper contraints if cluster is not yet healthy

xrstf:
During cluster creation, the `gatekeeper_constraint_template_controller` controllers tries to connect to the usercluster immediately to deploy some stuff. This will not work until the control plane is up and running, so lots and lots of bogus errors are logged. This PR fine tunes the controller a bit and makes it wait for the control plane to come up.

I also removed the explicit check for the CacheSync, because when starting the controller-runtime manager, it automatically waits for the cache before even starting the controllers.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
+++ b/pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
@@ -144,10 +144,8 @@ func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, cons
 					Name: constraintTemplate.Name,
 				},
 			})
-			if kerrors.IsNotFound(err) {
-				return nil
-			}
-			return err
+
+			return ctrlruntimeclient.IgnoreNotFound(err)
 		})
 		if err != nil {
 			return err
@@ -189,7 +187,7 @@ func (r *reconciler) syncAllClusters(
 		}
 
 		// if the control plane is not healthy, we cannot possibly create a functioning usercluster client
-		if userCluster.Status.ExtendedHealth.ControlPlaneHealthy() {
+		if !userCluster.Status.ExtendedHealth.ControlPlaneHealthy() {
 			clusterLog.Debug("Cluster control-plane is not healthy yet, skipping")
 			continue
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/seed-controller-manager/constraint-template-controller/controller.go
```diff
@@ -186,36 +181,40 @@ func (r *reconciler) syncAllClusters(
...
+
+		if userCluster.Spec.Pause {
+			clusterLog.Debug("Cluster is paused, skipping")
 			continue
 		}
 
-		if userCluster.Spec.Pause {
-			log.Debugw("Cluster paused, skipping", "cluster", userCluster.Spec.HumanReadableName)
+		// if the control plane is not healthy, we cannot possibly create a functioning usercluster client
+		if userCluster.Status.ExtendedHealth.ControlPlaneHealthy() {
```
lsviben: will the Cluster getting healthy trigger the watch event? Now when we have proper Cluster status I am not sure? In that case the CT wont get reconciled unless there is some cluster spec or CT change

xrstf: Excellent question. I just checked by [stalking](https://github.com/xrstf/stalk) a Cluster object and editing its status to set a new reconcile time. This then triggered a watch event with this diff:

screenshot-2022-02-25-145816

lsviben: awesome! Btw love the [stalk](https://github.com/xrstf/stalk)

xrstf:
Let me debug this, maybe I broke it.

/hold

lsviben:
> Let me debug this, maybe I broke it.
> 
> /hold

If there is a issue with canal, then its https://github.com/kubermatic/kubermatic/pull/9047

xrstf:
It seems that immediately after enabling OPA, we consider the old usercluster-ctrl-mgr pod to be good enough and the usercluster-ctrl-mgr Deployment to be healthy, so that the SeedResourcesUpToDate considition also immediately goes to `True` again. This means that later when OPA gets ready, no changes to the cluster object happen and no reconciling is triggered.

screenshot-2022-02-25-171233

xrstf:
Let me try and see what happens if I manually trigger some changes to the cluster object..

xrstf:
I wonder if this behaviour is the general root cause of the flaky OPA tests..

xrstf:
As it turns out, I simply forgot a `!` in the condition to check if the cluster is healthy :-D

xrstf:
/hold cancel

@lsviben I think it's now ready to be reviewed :)

PR #9131, opened 2022-02-25
Fix CRD migration oversights

xrstf:
During testing, we found a couple more minor issues. This PR fixes them, one per commit.

* Stale ownerRefs on Deployments nuked the nodeport-proxy, making all userclusters temporarily unavailable.
* Not super help messages about blocked objects, because no namespace was logged.
* Mutating webhooks were not removed, leading to trouble during the `--remove-old-resources` phase.
* BYO datacenters were not correctly migrated.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #9148, opened 2022-02-28
WIP - re-organize master/seed controllers

xrstf:
This PR brings some order into our controllers:

* all controllers are named `thing-controller` or `thing-synchronizer` (in general, `<noun>-controller` should be used)
* The Go files do not have redunant prefixes anymore (no more "backup/backup_controller.go") -- these longer names were useful when using klog, but not anymore with zap.
* All controller names internally have a "kkp-" prefix now and use dashes instead of underscores.

```release-note
NONE
```


xrstf:
This will probably never work as one big PR. The chances of not getting any other PR merged (and thereby conflicts being introduced) is slim to nil.

Time to make multiple smaller PRs.

PR #9149, opened 2022-02-28
Use sigs.k8s.io/yaml for KubermaticConfiguration unmarshalling in cmd/image-loader

embik:

`sigs.k8s.io/yaml` is better than `gokpk.in/yaml/v3` when it comes to parsing Kubernetes YAML, including the `KubermaticConfiguration` file. While this already works on `master`, this is a problem on earlier release branches like `release/v2.19`, where a configured `versions` struct fails with this error:

```
{"error": "failed to parse file as YAML: yaml: unmarshal errors:\n  line 67: cannot unmarshal !!str `v1.21.8` into semver.Version"}
````

That's because `semver.Version` does not implement `UnmarshalYAML`. It's not a problem on `master` or `release/v2.20`, as the new CRDs uses `semver.Semver` (a `string` type) instead, but it's broken for `release/v2.19` and below. To make `image-loader` work in those branches, I'm starting this change top-down to have all release branches aligned, but the goal is to have this backported to older release branches.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/image-loader/operator.go
+++ b/cmd/image-loader/operator.go
@@ -21,12 +21,13 @@ import (
 	"io/ioutil"
 
 	"go.uber.org/zap"
-	"gopkg.in/yaml.v2"
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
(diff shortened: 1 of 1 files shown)


xrstf (approved):
/approve

PR #9151, opened 2022-02-28
Add proposal for single source of truth CR validation

SimonTheLeg:

Proposal for a single source of truth from which CustomResources can be validated for our CustomResourceDefinitions.

Current state: To be discussed in next sig-crd

```release-note
none
```


SimonTheLeg:
closed as decision was made to go ahead without the need of a proposal and topic has been implemented

PR #9158, opened 2022-02-28
Use GOCACHE for building images with buildah

xmudrii:

Use GOCACHE for building images with buildah. This should speed up the image building, hopefully, by a few minutes.

Currently, this doesn't work out of the box. If you check the `pre-kubermatic-test-user-ssh-key-agent-multiarch` job, you'll see that GOCACHE is not used. However, I believe there might be a problem because we're changing the `cmd/user-ssh-keys-agent` package, which is for some reason invalidating the cache. I would propose that we merge this PR and continue debugging in a follow-up PR.

h/t to @xrstf for initially implementing this in #9027

```release-note
NONE
```


xrstf:
ok then, #yolo #vorwärtsimmerrückwärtsnimmer

/approve
/lgtm

PR #9167, opened 2022-03-01
Remove optional=true setting from prow jobs

embik:

This is a PR to start a discussion, not necessarily to be merged as a whole.

IMO, no prow jobs should be configured with `optional: true`. The reason for that is that it skips [checks for auto merging](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#requiring-jobs-for-auto-merge-through-tide). Usually, all of our jobs should succeed. In particular, cloud provider jobs are configured with this, but they only trigger when provider-specific code is changed - If the provider implementation is changed, our e2e tests should be a prerequisite for merging it! If they fail, our e2e setup might not be working, but then we cannot trust the PR anyway. Otherwise we might break a provider and don't realise we did.

There's just very few exceptions where these jobs can be disregarded, in particular if a provider is known to be flaky due to provider-side issues or long-term infra issues. I'm not aware we have that at the moment, but we can introduce it for jobs later on.

WDYT @xmudrii @xrstf @moadqassem?

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -110,7 +110,6 @@ presubmits:
           value: ee
 
   - name: pre-kubermatic-shellcheck
-    optional: true
     decorate: true
     clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
     labels:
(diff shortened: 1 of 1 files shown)


embik:
/hold

For discussing this PR.

WeirdMachine:
/lgtm 

:100:

PR #9169, opened 2022-03-01
cleanup codespell.exclude

xrstf:
Since #9014 we do not need to have these two in our exclusion list anymore :)

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve
/lgtm

PR #9171, opened 2022-03-01
cleanup kubeletdnat controller naming

xrstf:
After #9148 (the attempt to rename all controllers in one go) was doomed to fail, this PR is part 1 of doing it step by step.

```release-note
NONE
```

### DIFF
--- a/cmd/kubeletdnat-controller/main.go
+++ b/cmd/kubeletdnat-controller/main.go
@@ -24,7 +24,7 @@ import (
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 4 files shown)


kron4eg (approved):
/approve

kron4eg:
As the last step probably we should should start utilize the https://golangci-lint.run/usage/linters/#importas linter with config something like
```yaml
      - pkg: k8c.io/kubermatic/v2/pkg/controller/(\w+)-controller
        alias: $1controller
```

PR #9178, opened 2022-03-01
Fix copying GOCACHE in release-docker-images script

xmudrii:

Fix copying GOCACHE in release-docker-images script. This is a regression from #9158.

```release-note
NONE
```

/assign @xrstf


xrstf (approved):
/approve

PR #9198, opened 2022-03-02
allow not having any user-accessible addons

### LINKED ISSUE #9111: Not possible to disable accessibleAddons form KKP operator
Looks like the user can disable accessibleAddons. Always default ones are added

https://support.kubermatic.com/helpdesk/tickets/3634

xrstf:
This allows admins to configure `accessibleAddons: []` without having the operator overwrite this with the default addons.

Fixes #9111

```release-note
It is now possible to disable all user accessible addons in the operator by setting `spec.api.accessibleAddons=[]`.
```

### DIFF
--- a/codegen/example-yaml/main.go
+++ b/codegen/example-yaml/main.go
@@ -206,9 +206,7 @@ func createExampleKubermaticConfiguration() *kubermaticv1.KubermaticConfiguratio
 				Domain: "example.com",
 			},
 			FeatureGates: map[string]bool{},
-			API: kubermaticv1.KubermaticAPIConfiguration{
-				AccessibleAddons: []string{},
-			},
+			API:          kubermaticv1.KubermaticAPIConfiguration{},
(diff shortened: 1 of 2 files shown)


embik:
> It is now possible to disable all user accessible addons in the operator by setting `spec.api.accessibleAddons=[]`.

Is that how it works? I am under the impression that `configCopy.Spec.API.AccessibleAddons == nil` will only be true if `accessibleAddons` is unset (as the comment suggests). So I think the release note might be wrong?

xrstf:
If you don't set the field at all, it will be `nil` in Go. In this case, we default it. To disable the default accessible addons, you have to explicitly configure an empty slice. I don't see where the release message is wrong or misleading?

embik:
> If you don't set the field at all, it will be `nil` in Go. In this case, we default it. To disable the default accessible addons, you have to explicitly configure an empty slice. I don't see where the release message is wrong or misleading?

Ah, sorry, I mixed things up. Never mind :-)

embik (approved):
/approve

PR #9200, opened 2022-03-02
add mutating webhook for UserSSHKeys

xrstf:
This PR adds a new webhook, responsible for the small task of ensuring that the fingerprint in each UserSSHKey always matches the actual public key. Since the public key can be changed, I also adjusted the logic for key names, which previously relied on the _initial_ public key.

```release-note
The KKP webhook now ensures that UserSSHKey fingerprints always match their public key.
```


xrstf:
@floreks can i get an LGTM as well?

PR #9202, opened 2022-03-02
Implement vsphereCSIClusterID feature flag

xmudrii:

The vSphere CSI docs state that the `cluster-id` in the CSI config should be a unique identifier. Originally, we used vSphere Compute Cluster as a value for `cluster-id` for all user clusters, but this was changed recently by #9061.

However, the following [vSphere KB article](https://kb.vmware.com/s/article/84446) recommends a special procedure for changing the `cluster-id`. This means that we can't just change the value, instead, users will have to migrate manually. A document describing how to migrate will be added in a follow-up to the docs website.

To make this possible, we're introducing a feature flag on the cluster object (`vsphereCSIClusterID`). Enabling this feature gate will modify the `cluster-id` to the name of the user cluster (instead of the name of the vSphere Compute Cluster provided via Datacenter config).

**Documentation**:

Will be added in a follow-up.

```release-note
Add `vsphereCSIClusterID` feature flag for the cluster object. This feature flag changes the cluster-id in the vSphere CSI config to the cluster name instead of the vSphere Compute Cluster name provided via Datacenter config. Migrating the cluster-id requires manual steps (docs link to be added).
```

### DIFF
--- a/pkg/apis/kubermatic/v1/cluster.go
+++ b/pkg/apis/kubermatic/v1/cluster.go
@@ -239,6 +239,15 @@ const (
 	// The cluster-name flag is often used for naming cloud resources, such as load balancers.
 	ClusterFeatureCCMClusterName = "ccmClusterName"
 
+	// ClusterFeatureVsphereCSIClusterID sets the cluster-id in the vSphere CSI config to
+	// the name of the user cluster. Originally, we have been setting cluster-id to the
+	// vSphere Compute Cluster name (provided via the Datacenter object), however,
+	// this is supposed to identify the Kubernetes cluster, therefore it must be unique.
+	// This feature flag is enabled by default for new vSphere clusters, while existing
+	// vSphere clusters must be migrated manually (preferably by following advice here:
+	// https://kb.vmware.com/s/article/84446).
+	ClusterFeatureVsphereCSIClusterID = "vsphereCSIClusterID"
+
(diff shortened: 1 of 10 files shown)


ahmedwaleedmalik (approved):
/approve

xrstf (approved):
/approve

dharapvj:
I feel that this flag can be made default on. It will only be causing transient failures. Do we really need this to be looked at as feature flag at all?

toschneck:
I'm totally with @dharapvj who could this be a feature flag? As without it vsphere clusters wouldn't work correctly

xmudrii:
@dharapvj @toschneck It’s a feature flag because it requires manual migration steps. We are still discussing how the migration should look like and we’ll document it as soon as we have a solution. The feature flag is enabled by default for newly created clusters.

toschneck:
@xmudrii  thanks for this info! Understood, I'm fine to migrate one time existing clusters, but new clusters should definitly have this enabled 👍  Regards the migration, I think a option would be to include this into the kubermatic installer CLI

xmudrii:
@toschneck Yes, new clusters are already having this option enabled by default. I don't think we can integrate migration into the KKP installer because the migration procedure is specific and requires downtime of about 1 hour. See the [VMware KB](https://kb.vmware.com/s/article/84446) for some details about how the migration is supposed to look like.

toschneck:
Understood, ok then a good migration doc or a small helper script would be needed. I just think about to minimize the risk something get broken ;-)

PR #9206, opened 2022-03-02
more preflight checks

xrstf:
This adds more checks to ensure that newly added manadatory fields for providers are validated before we attempt a migration.

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #9208, opened 2022-03-02
Proposal for resource quotas

### LINKED ISSUE #9115: Investigate project and user quotas
Investigate how to achieve project and user quotas, write a proposal doc

lsviben:
Proposal for the workings and how to implement resource quotas for Projects and Users

Fixes #9115 

```release-note
NONE
```

### DIFF
--- a/docs/proposals/project-user-resource-quotas.md
+++ b/docs/proposals/project-user-resource-quotas.md
@@ -1,0 +1,122 @@
+# Project/User/Datacenter resource quotas
+
+**Authors**: Lovro Sviben (@lsviben)
+
+**Status**: Draft proposal.
+
+**Issue**: https://github.com/kubermatic/kubermatic/issues/8042
+
+## Goals
+The idea is to allow admins to control how many resources are users using by introducing resource quotas.
+
+The quotas would restrict:
+- resource usage per project
+- resource usage of a datacenter per project
+- user resource usage
+
+Resources that will be under quota:
+- vCPU
+- RAM
+- storage
+
+## Non-Goals
+
+
+## Motivation and Background
+
+Clients would like to have options to fine-grain restrict resource usage per project, of a datacenter per project and individual resource usage.
+The reasons vary from being able to stop the depletion of resources in a shared datacenter, to being able to sell packages (projects) with a set 
+number of resources available. 
+
+This will be an EE feature.
+
+## How will the resource quotas work
+
+Three types of resource quotas:
+- Project quota - amount of resources (vCPU, RAM, storage) that can allocated for nodes in a project
+- Project datacenter quota - amount of resources (vCPU, RAM, storage) that can allocated for nodes in a datacenter for a project. 
+- User quota - amount of resources (vCPU, RAM, storage) that can allocated for nodes when a user is creating them
+
+
+Rules:
+
+- The quotas are for node resources
+- The resource quotas management will be admin-only.
+- By default, projects, project datacenters and users will not have a quota, meaning they can make as many resources they want
+- There will be a way to set a default resource quota for every project (should we do the same for users?)
+- When a user A creates a cluster in a project B, in datacenter C, the quotas for all of those are checked, and get filled up
+- When a node is deleted, the quota usage needs to decrease for:
+  - project
+  - user who created the node (not the one who deleted it)
+  - project datacenter
+
+Questions?
+- how should this work for external clusters?
+  - For example importing a cluster to a project will change its resource quotas or not?
+  - should we block imports that go over the quota
+  - what about node creation for external clusters
+
+## Implementation
+
+### How to ensure that we get proper resource usage and that it stays in sync?
+
+The obvious way seems to get the amount of resources requested when a KKP NodeDeployment is being created, and then check against
+the user, datacenter and project quota and fill them up accordingly. But this would only work through the API, and if there are some issues
+we could have resource quotas which are out of sync.
+
+The safest way would be to check the Node capacity on user clusters, but there are some issues with that as well, as the Nodes get created
+based on the NodeDeployments. So there could be races in between when the Node is created(which can take some time) and when we check if user can create a NodeDeployment.
+
+So the best would be to create a controller which watches NodeDeployments, and fills out the project, datacenter and user quotas accordingly.
+And the check for resource quotas could then be in the webhook, so to cover both kubectl and API creation.
+
+There is still a chance of a race, in which multiple users could create NodeDeployments for a project in the same time, before the
+controller updates the resource usage, thus bypassing the limit. In this case we have 2 options:
+- admins could notice/get informed about it and react according to their company policy
+- add a "pending"/"reserved" resource mechanism to the API. So as soon as a NodeDeployment request is received in the API, it would reserve some of the quota. Later the NodeDeployment controller could move this from reserved to real usage.
+
+
+### CRD Changes
+
+Add a new flexible ResourceQuota CRD which will hold the desired quota and current consumption of the quota. 
+
+The ResourceQuota will need to have an OwnerReference to a Project or User to which the quota applies to. If a specific datacenter is targeted for a project,
+the Datacenter field needs to be set.
+
+```go
+type ResourceQuota struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec   ResourceQuotaSpec   `json:"spec,omitempty"`
+    Status ResourceQuotaStatus `json:"status,omitempty"`
+}
+
+type ResourceQuotaSpec struct {
+	// Datacenter is the name of the datacenter for which a resource quota should be set.
+    Datacenter     string          `json:"datacenter,omitempty"`
+	// ResourceQuotas is a map of maximum resource quotas per resource
+    ResourceQuotas corev1.ResourceList `json:"resourceQuotas"`
+}
+
+type ResourceQuotaStatus struct {
+	// ResourceConsumption is map which holds the current usage of resources per resource 
+    ResourceConsumption corev1.ResourceList `json:"resourceConsumption"`
+}
+```
+
+### Possible Enhancements in the future
+
+1. Try to make the resourcequtas bulletproof by adding a mechanism of "pending"/"reserved" quota which is filled before the NodeDeployment creation
+2. Add a possibility to set user groups and set quotas per group
+
+## Tasks and effort
+
+1. Investigate project and user quotas
+2. Implement resource quotas CRD
+3. Implement default project quotas 
+4. ResourceQuota API endpoints
+5. Implement controller for NodeDeployments which fills the resource quotas
+6. Webhook validation for resource quotas
+7. (Optional) Implement a "pending/reserved" mechanism to decrease chances of races
+
--- a/docs/proposals/project-user-resource-quotas.md
+++ b/docs/proposals/project-user-resource-quotas.md
@@ -37,7 +37,7 @@ Rules:
 - The quotas are enforced(we should block creation of new nodes that would exceed the quota)
 - The resource quotas management will be admin-only.
 - By default, projects will not have a quota, meaning they can use as many resources they want
-- There will be a way to set a default resource quota for every project 
+- There will be a way to set a default resource quota for all projects
 - When a node is deleted, the quota usage needs to decrease for the project
 - If the quota is exceeded(by lowering the quota under current capacity for example), it is not the responsibility of KKP to remove nodes. Admins should fix that in communication with the project/user.
 - The storage quota targets just the disk size of the node. It will only affect node-local PV storage.
@@ -102,6 +102,11 @@ when we check if user can create a MachineDeployment.
 
 So we need to focus on the MachineDeployments and control the quota through them. The nodes get created anyway based 
 on the MachineDeployment spec, and we can get the desired node size based on the MD, although for each cloud provider its a bit different.
+So the best would be to create a controller which watches MachineDeployments, and fills out the project quotas accordingly.
+
+The seed-machine-deployment-controller(SMDC) would watch the seed cluster MachineDeployments(assuming we go with the plan to move MDs to the seed cluster, otherwise we
+need to have it watch MDs on user clusters, and do it a bit differently), calculate the resource usage, and update quotas for that clusters
+project (if such a quota exists). Also it needs to watch for new ResourceQuotas, so that for new ones it calculates the usage.
 
 ### How to get node size
 
@@ -141,7 +146,6 @@ The pro is that it wouldn't be a big change, the con is that it won't be so visi
 We can show the info that there is not enough resource quota left to create the nodes, but users would only see that in MachineDeployment
 Events. So it would be something like Pods not getting scheduled because there is no space.
 
-
 #### 1. Webhook for MachineDeployments
 
 This variant is dependent on migrating the MachineDeployments from being user-cluster resources to being seed-cluster resources.
@@ -151,8 +155,7 @@ project's resource quota if the MD fits. If not, we would just deny the request.
 
 The problem here is if its possible and at what cost to migrate the MD's, while thinking about KubeOne as well.
 We could eventually dynamically set the kubeconfig of the cluster where the machine-controller watches for MachineDeployments, 
-and set it to the seed-cluster in case of KKP. But then, as MachineDeployments are cluster-scoped we will need additional 
-cluster info label or something on them.
+and set it to the seed-cluster in case of KKP. 
 
 Another issue could be that cluster users won't have direct kubectl access to MD's anymore, at least until we start 
 supporting it somewhere in the future. But they can use the KKP API. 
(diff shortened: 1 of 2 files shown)


xrstf:
This is a good start, but I have loads of questions.

1. Is this in any way supposed to also restrict seed cluster resources? In this case, we would either need a `maxClusters` restriction for each project/user or very strict resourceRequests/Limits on all the deployments for the usercluster control planes.

1. What about nodes that join the cluster but are not managed by a MachineDeployment? As far as I understand it, users could simply use kubeadm to build new nodes and make them join the cluster, no?

1. How do we determine the capacities that are in use? By queyring the Kubernetes Nodes and taking the capacity info from them? Or do we keep a map somewhere with "AWS t3.large = 8 CPU, 16GB"?

1. How do we prevent spawning new nodes from MachineDeployments when the quotas are already reached? What prevents a user from doing `kubectl scale machinedeployment/workers --replicas=1000000`? Unless we have a dedicated webhook that enforces changes, how would we deal with this?

1. What happens if there are more resources in use than are allowed (e.g. because the quota got lowered or at the very beginning when this feature is enabled and the new quotas are introduced to existing clusters)? How do we determine what nodes to remove? Can we just nuke nodes?

1. How does storage work? Do we just watch all PVs in the usercluster? What happens if someone installs Prometheus StatefulSet and scales it up to 50 pods, each with 10GB volume, but only 100GB would be allowed in total? Even if we started playing whack-a-mole with PVs, the StatefulSet controller of Kubernetes will just endlessly try to fix the situation. Do we create a new webhook to deny new PVs? How would this impact rollouts of StatefulSets?

1. Owner references are IMHO not the correct method to express semantic relationships between objects. OwnerRefs are a way to manage cleanup and I would not like us to rely on ownerRefs. Querying things in reverse via ownerRefs is super annoying and brittle. Take the 2.20 CRD migration in which we would break all ResourceQuotas because the ownerRefs get shuffled around.

lsviben:
> This is a good start, but I have loads of questions.
> 
> 1. Is this in any way supposed to also restrict seed cluster resources? In this case, we would either need a `maxClusters` restriction for each project/user or very strict resourceRequests/Limits on all the deployments for the usercluster control planes.
> 2. What about nodes that join the cluster but are not managed by a MachineDeployment? As far as I understand it, users could simply use kubeadm to build new nodes and make them join the cluster, no?
> 3. How do we determine the capacities that are in use? By queyring the Kubernetes Nodes and taking the capacity info from them? Or do we keep a map somewhere with "AWS t3.large = 8 CPU, 16GB"?
> 4. How do we prevent spawning new nodes from MachineDeployments when the quotas are already reached? What prevents a user from doing `kubectl scale machinedeployment/workers --replicas=1000000`? Unless we have a dedicated webhook that enforces changes, how would we deal with this? And such a webhook would need to pre-calculate the impact of such a change, so it would need to know the CPU/memory of the machine type of that MachineDeployment (see previous point).
> 5. What happens if there are more resources in use than are allowed (e.g. because the quota got lowered or at the very beginning when this feature is enabled and the new quotas are introduced to existing clusters)? How do we determine what nodes to remove? Can we just nuke nodes?
> 6. How does storage work? Do we just watch all PVs in the usercluster? What happens if someone installs Prometheus StatefulSet and scales it up to 50 pods, each with 10GB volume, but only 100GB would be allowed in total? Even if we started playing whack-a-mole with PVs, the StatefulSet controller of Kubernetes will just endlessly try to fix the situation. Do we create a new webhook to deny new PVs? How would this impact rollouts of StatefulSets? What about node-local volumes and other StorageClas smechanisms -- do we need to filter to only count a certain set of StorageClasses?
> 7. Owner references are IMHO not the correct method to express semantic relationships between objects. OwnerRefs are a way to manage cleanup and I would not like us to rely on ownerRefs. Querying things in reverse via ownerRefs is super annoying and brittle. Take the 2.20 CRD migration in which we would break all ResourceQuotas because the ownerRefs get shuffled around. Also it would be super nice to see some helpful info when doing `kubectl get resourcequotas` and using ownerRefs would make that impossible.

Thanks for the great feedback, this brought some new considerations. TBH with some of those things I could use suggestions, so if you have some please send them. 

1. No, but that would be a good addition. I could see something like `maxClusters` per project being usefull. I expect it would work somewhat like current `projectLimit` works, outside of the quotas. I would not try to restrict the usage of seed cluster resources in any other way, because cluster users dont have visibility or choice in using them. Which also brings the question of resources deployed by KKP in user clusters, should we count them? But for the sake of simplicity I would not treat them differently for now.

2. They could if they have credentials. If admins give users just presets to use I dont think they can get them. And if the use their own credentials to add nodes from somewhere else, it doesnt cost the KKP admins company.

3. Good question. I was also thinking first to take the Node capacity, but until the node gets created the users could overfill the quota. So I would use the mechanism that we already have for filtering node sizes, there we already have the getting node sizes from different providers implemented. Like here: https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/common/provider/azure.go#L330-L354

4. Thats a great question and I have no idea now. Tbh I forgot that MachineDeployments  now are on the user cluster, so users have access to them. Anyway I planned that a webhook would check this using above logic, but if that is all on the user cluster it makes things complicated. 

5. I would not do any actions and leave it to the admins to figure out with their users. It would be impossible to figure out from us what to remove.

6. The storage is just for the node disk size that is chosen when creating a node. So it would only affect the node-local storage, so users would get blocked naturally when its used up. We get the disk size in the same way as the RAM and CPU, from the providers. So no need to deal with PVs from our side. 

7. Great point, tbh I was also not 100% sure about using ownerRef, but I couldnt bring it to words. Ill change it to `type` or something similar in the `spec`

xrstf:
> They could if they have credentials. If admins give users just presets to use I dont think they can get them. And if the use their own credentials to add nodes from somewhere else, it doesnt cost the KKP admins company.

So how do we filter nodes that we count towards the resource quota? I need this proposal to be way more specific and detailed and not just "yeah I'll do something to make that work" ^^

> So I would use the mechanism that we already have for filtering node sizes, there we already have the getting node sizes from different providers implemented

So you do want to maintain instance information for all instance types and all cloud providers in KKP? Can we even do that for systems like Openstack with non-default instance types?

lsviben:
> > They could if they have credentials. If admins give users just presets to use I dont think they can get them. And if the use their own credentials to add nodes from somewhere else, it doesnt cost the KKP admins company.
> 
> So how do we filter nodes that we count towards the resource quota? I need this proposal to be way more specific and detailed and not just "yeah I'll do something to make that work" ^^

Well we cant control everything. I would say that we control the resource quotas through MachineDeployments, KKP is not responsible for other ways of adding nodes to a cluster. 

Ill update the proposal with specifics to make it more clear.
> 
> > So I would use the mechanism that we already have for filtering node sizes, there we already have the getting node sizes from different providers implemented
> 
> So you do want to maintain instance information for all instance types and all cloud providers in KKP? Can we even do that for systems like Openstack with non-default instance types?

No I dont want to mantain instance information. I would like to use provider clients to get info about the size. We already do that for most providers, Openstack included. If for some provider we cant do it, we dont have to support it with resource quotas. Also from what I see, for vSphere, which is tricky as there are no set of instance sizes, the user just sets the requested amout of resources, so we can use that. 

Ill add some table to the proposal about how to get the size from the provider and if its possible

lsviben:
Updated it a bit, and put to WIP status as I feel its not finished. The more i dig the more possible issues I see. 

The biggest one now is that to be able to control user actions on MachineDeployments, and to have a good way to validate them against quotas through a webhook, we would need to move MachineDeployments to the seed-cluster.

More info is in the proposal, if you have some suggestions please come forth and provide them :). I need to check how machine-controller works and if this is even possible.

xrstf:
We all hate meetings, but maybe we need a meeting to discuss this? I also see many possible pitfalls ^^

lsviben:
> We all hate meetings, but maybe we need a meeting to discuss this? I also see many possible pitfalls ^^

Right, added it to tomorrows sig-cluster-mgmt agenda

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,221 @@
...
+Another issue could be that cluster users won't have direct kubectl access to MD's anymore, at least until we start 
+supporting it somewhere in the future. But they can use the KKP API. 
+
+#### 2. Machine-controller checks quota
+
+In this variant, the machine-controller is checking the quota before it creates nodes for a MachineDeployment. 
+If there is no quota left, it should not schedule the node, and emit some event and set status that there is not enough resources
+left in the quota.
+
+### How to ensure that we get proper resource usage and that it stays in sync?
```
embik: There is a section at line 93 with the same title. I assume the two should be merged in some way.

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,221 @@
...
+
+This variant is dependent on migrating the MachineDeployments from being user-cluster resources to being seed-cluster resources.
+
+If the MD's are on the seed cluster, we can control their creation/modification through a validation webhook which would check with the 
+project's resource quota if the MD fits. If not, we would just deny the request.
+
+The problem here is if its possible and at what cost to migrate the MD's, while thinking about KubeOne as well.
+We could eventually dynamically set the kubeconfig of the cluster where the machine-controller watches for MachineDeployments, 
+and set it to the seed-cluster in case of KKP. But then, as MachineDeployments are cluster-scoped we will need additional 
+cluster info label or something on them.
```
embik: > But then, as MachineDeployments are cluster-scoped we will need additional 
cluster info label or something on them.

Not sure if this is what you mean, but `MachineDeployments` are namespaced-scoped (they just all sit in `kube-system` on the user cluster), so they _could_ run in the cluster namespace on the seed and that maps them to the right cluster. Still, quite a big change, but just wanted to add this information.

docs/proposals/project-user-resource-quotas.md
```diff
@@ -26,32 +21,24 @@ Resources that will be under quota:
...
 - The resource quotas management will be admin-only.
-- By default, projects, project datacenters and users will not have a quota, meaning they can make as many resources they want
-- There will be a way to set a default resource quota for every project (should we do the same for users?)
-- When a user A creates a cluster in a project B, in datacenter C, the quotas for all of those are checked, and get filled up
-- For simplicity, the user which is the owner of a cluster, is used when calculating quotas for the cluster nodes. Even though another user can create nodes in that cluster.
-- When a node is deleted, the quota usage needs to decrease for:
-  - project
-  - the cluster owner
-  - project datacenter
+- By default, projects will not have a quota, meaning they can use as many resources they want
```
shaase-ctrl: Or the admin can configure a default quota. It is common practice with cloud providers as well (e.g. 100 vcpu, 400 GB RAM, 5000 GB Storage)

docs/proposals/project-user-resource-quotas.md
lsviben: Right, removed. Somehow I had it programmed in my head that its cluster-scoped

docs/proposals/project-user-resource-quotas.md
lsviben: Ah, I updated the doc a few times and it got mixed up :D. Thanks for the tip, fixed.

docs/proposals/project-user-resource-quotas.md
lsviben: The line just under it says this that admin can set a default quota :).

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+
+Resources that will be under quota:
+- vCPU
+- RAM
+- storage (node disk size) - TBD, we can remove this as it's not a strict requirement
+
+## Non-Goals
+- The quotas won't work for imported external clusters.
+- The quotas won't restrict the usage of resources for the control plane on the seed cluster.
+- Nodes which join the cluster using other means then through KKP are not supported in the quotas.
```
xrstf: _than_

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Motivation and Background
+
+Clients would like to have options to fine-grain restrict resource usage per project.
```
xrstf: Please use "KKP Admins", so we do not always talk in us+customer terms, but us+community. I do not want to give off the vibe that what we do here is only to satisfy customers :)

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+This will be an EE feature.
+
+## How will the resource quotas work
+
+Project quota - amount of resources (vCPU, RAM, storage) that can allocated for nodes in a project.
+
+Rules:
+
+- The quotas are for node resources
+- The quotas are enforced(we should block creation of new nodes that would exceed the quota)
```
xrstf: creation of new nodes != creation of new machines

We cannot really prevent nodes from joining, but we can prevent Machines from appearing, which should be sufficient for 99% of userclusters (I presume that cases where people make their own nodes join a cluster (e.g. with the byo provider) are rare and not relevant for this proposal anyway). But let's make sure we consistently talk about Machines instead of nodes when applicable.

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+- The resource quotas management will be admin-only.
+- By default, projects will not have a quota, meaning they can use as many resources they want
+- There will be a way to set a default resource quota for all projects
+- When a node is deleted, the quota usage needs to decrease for the project
```
xrstf: Do we want to do this once the Node is gone or once the Machine object is gone? In general, the Machine object should disappear after the node has gone, IIRC.

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+
+The ResourceQuota is a master-cluster resource. It will need to be synced to seed clusters to allow webhooks access.
+As we have a distributed platform over multiple seeds, each seed will have to report its usage in the status. Then the master quota controller
+can recalculate the total resource consumption and re-sync the quotas on the seeds.
+
+### How to sync quota info from seed clusters and master cluster and back
+
+We would need a controller for ResourceQuotas in the master cluster: `master-resource-quota-controller` - MRQC
+
+The MRQC would watch master cluster ResourceQuotas. These are the ones that are created through the API, and maybe one day through `kubectl`.
```
xrstf: We already have a bunch of "...-sync" controllers that mirror master resources onto seed clusters. It might be possible/good to also separate this proposal into 2 controllers, once to implement the business logic (possibly running on the seed?), and one that just blindly mirrors things (running on the master cluster).

Just as a thought.

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+
+The obvious way seems to get the amount of resources requested when a KKP MachineDeployment is being created, and then check against
+the project quota and fill it up accordingly. But this would only work through the API, and if there are some issues
+we could have resource quotas which are out of sync.
+
+The safest way would be to check the Node capacity on user clusters, but there are some issues with that as well, as the Nodes get created
+based on the MachineDeployments. So there could be races in between when the Node is created(which can take some time) and 
+when we check if user can create a MachineDeployment.
+
+So we need to focus on the MachineDeployments and control the quota through them. The nodes get created anyway based
```
xrstf: This chapter talks about MachineDeployments, but what about standalone Machine objects? Shouldn't we focus on those?

docs/proposals/project-user-resource-quotas.md
```diff
@@ -0,0 +1,209 @@
...
+| Nutanix      | TBD                       | TBD                                                    |                                  |
+| Equinox      | Y                         |                                                        |                                  |
+| vSphere      | N                         | We set the requested size directly into MachineDeployment |                                  |
+
+
+### Enforcing Quotas - Implementation variants
+
+I see 2 possible ways to implement the enforcing of project quotas through MachineDeployments. 
+
+One is through a MachineDeployment webhook, although it requires moving the MD's to the seed cluster, because the webhook
```
xrstf: This is still outdated after our discussion in the SIG meeting :)

lsviben: updated it, but sill need to finish the webhook POC so it could change with new findings.

docs/proposals/project-user-resource-quotas.md
lsviben: Makes sense to do it when the Machine object is gone

docs/proposals/project-user-resource-quotas.md
lsviben: I see it like this:
MRQC does the mirroring to and back from the seed clusters
A bit down is a paragraph about 2 more controllers:
1. One running as part of the `user-cluster-controller` which calculates the resource usage for the cluster and updates cluster status
2. One running as part of `seed-cluster-controller` which calculates the resource usage for a project on the Seed, and updates the ResourceQuotas `status.localConsumption`

xrstf:
@lsviben With PRs like #9810 getting merged, shouldn't this PR also be merged already?

lsviben:
> @lsviben With PRs like #9810 getting merged, shouldn't this PR also be merged already?

whoops I forgot about it. Ill update it

xrstf:
We already merged parts of the implementation, so let's get this merged as well.

/approve
/lgtm

PR #9172, opened 2022-03-01
suppress repeated warnings regarding deprecated policy/v1beta1

xrstf:
**What does this PR do / Why do we need it**:
This gets rid of the repeated warnings in our logs:

screenshot-2022-03-01-161509

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #9215, opened 2022-03-03
try different method of suppressing deprecation warnings

xrstf:
#9172 didn't have any effect, so let's try it this way. Sadly this is not really testable on CI, we have to deploy it to dev to see if it works.

```release-note
NONE
```


embik (approved):
/approve

PR #9205, opened 2022-03-02
add mutating webhook for Addons

xrstf:
**What does this PR do / Why do we need it**:
This does for Addons what #9200 does for UserSSHKeys.

**Does this PR introduce a user-facing change?**:
```release-note
The KKP webhook now ensures that Addons are only created in cluster namespaces and assigns a proper Cluster reference.
```

PR #9221, opened 2022-03-03
make addon.spec.variables a pointer and truly optional

xrstf:
While working in #9205 it turns out that having a RawExtension being optional and _not_ a pointer is impossible. When the webhook would unmarshal and later re-marshal the response, a missing `"variables": ...` in the JSON will be replaced with `"variables":null`, which is not allowed:

 > {"level":"error","time":"2022-03-03T13:36:58.180Z","logger":"kubermatic_addoninstaller_controller","caller":"addoninstaller/addoninstaller_controller.go:156","msg":"Reconciling failed","request":"/htx58b2qnp","error":"failed to create addon \"canal\": failed to create addon: [Addon.kubermatic.k8c.io](http://addon.kubermatic.k8c.io/) \"canal\" is invalid: spec.variables: Invalid value: \"null\": spec.variables in body must be of type object: \"null\"","errorCauses":[{"error":"failed to create addon \"canal\": failed to create addon: [Addon.kubermatic.k8c.io](http://addon.kubermatic.k8c.io/) \"canal\" is invalid: spec.variables: Invalid value: \"null\": spec.variables in body must be of type object: \"null\""}]}

This PR changes the type of the field to be truly optional.

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubermatic/v1/addon.go
+++ b/pkg/apis/kubermatic/v1/addon.go
@@ -61,7 +61,7 @@ type AddonSpec struct {
 	// Cluster is the reference to the cluster the addon should be installed in
 	Cluster corev1.ObjectReference `json:"cluster"`
 	// Variables is free form data to use for parsing the manifest templates
-	Variables runtime.RawExtension `json:"variables,omitempty"`
+	Variables *runtime.RawExtension `json:"variables,omitempty"`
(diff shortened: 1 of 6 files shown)


kron4eg (approved):
/approve

xrstf:
@lsviben PTAL :)

PR #9223, opened 2022-03-03
add OpenStack credential validation

### LINKED ISSUE #8965: Add validation for OpenStack

zreigz:
**What does this PR do / Why do we need it**: add OpenStack credential validation. Reverte the` CreateOrUpdateCredentialSecretForCluster` method and create the new one with validation because it's used only in one place and need some extra parameters like datacenter and caBundle for the rest of the providers.

Fixes #8965

```release-note
NONE
```


zreigz:
@imharshita @lsviben PTAL

PR #9226, opened 2022-03-03
use the same k8s client for MLA so we don't accidentally reset the global KubeAPIWarningLogger

xrstf:
The custom client in the MLA controller was the reason why the customized KubeAPIWarningLogger was replaced and the efforts from #9172 and #9215 were undone. This PR gets rid of the new client, because I do not why we would use the config from the manager to create a new client instead of just using the existing client.

On the road to implementing this, the MLA controller was refactored in a way where it doesn't create the Grafana client statically on boot, but rather dynamically during each reconciliation. This means the seed-ctrl-mgr will automatically react to changed Grafana credentials (though this should never happen).

```release-note
NONE
```

### DIFF
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -47,6 +47,9 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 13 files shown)


xrstf:
@embik PTAL

rastislavs:
/lgtm

I especially like the dynamic Grafana client!

PR #9230, opened 2022-03-04
Update secret name for S3 credentials

ahmedwaleedmalik:

Currently, there is a conflict between the `s3-credentials` secret created by KKP and KubeOne. Both use the same secret name and deploy it in the same namespace i.e. kube-system. But the keys within the secret are different for both of them.

In a bid to solve this issue completely, we are renaming the secret in KKP first. So that the new secret `kkp-s3-credentials` only has a single consumer. The next step is to rename this in KubeOne and add a deterministic prefix `kubeone-` to the secret name.

Fixes #

I have intentionally not included prefixes in the name like `kubermatic-` or called it `s3-exporter-credentials` etc. because this is created using **minio** chart and then consumed by minio and s3-exporter both.

**Documentation**:

```release-note
[ACTION REQUIRED] Secret name for S3 credentials updated to `kubermatic-s3-credentials`. If the secret `s3-credentials` was manually created instead of using the `minio` helm chart, new secret `kubermatic-s3-credentials` must be created.
```

### DIFF
--- a/charts/minio/test/values.example.ce.yaml.out
+++ b/charts/minio/test/values.example.ce.yaml.out
@@ -18,7 +18,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: "s3-credentials"
+  name: "kubermatic-s3-credentials"
   namespace: "kube-system"
 type: Opaque
 data:
--- a/charts/minio/test/values.example.ee.yaml.out
+++ b/charts/minio/test/values.example.ee.yaml.out
@@ -18,7 +18,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: "s3-credentials"
+  name: "kubermatic-s3-credentials"
   namespace: "kube-system"
 type: Opaque
 data:
--- a/charts/minio/values.yaml
+++ b/charts/minio/values.yaml
@@ -30,13 +30,13 @@ minio:
     secretKey: '' # 64 byte long
 
     # When set to true, a Secret will be created in the given namespace.
-    # Kubermatic requires an "s3-credentials" Secret in the kube-system
+    # Kubermatic requires an "kubermatic-s3-credentials" Secret in the kube-system
     # namespace to perform usercluster etcd snapshots, _if_ the default
     # backup containers are used (see KubermaticConfiguration).
     # Otherwise, this can be disabled for example by setting
     # `--set "minio.credentials.secret=null"` when running Helm.
     secret:
-      name: s3-credentials
+      name: kubermatic-s3-credentials
       namespace: kube-system
 
   flags:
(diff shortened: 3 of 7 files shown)


embik:
Pretty sure some of the etcd backup logic expects this secret at that name, you might need to check that out. Probably also needs an update in our docs and an `Action required` notice if we choose to change the secret name expected by the backup.

xrstf:
I am not sure calling it "aws-s3-..." makes sense, as it's uber specific and it's not even pointing to AWS, it's by defaulting poiting to our Minio. Calling it "aws-something" is super misleading.

Also, why does KKP have to change it KubeOne just started to use the same name? KKP has a much longer history with this Secret. Why not fix KubeOne instead?

ahmedwaleedmalik:
> etcd backup logic expects this secret at that name

That's super wrong, why are we creating a secret using `minio` helm chart that is then being consumed by multiple components?

xrstf (changes requested):
The secret was purposefully named after the _protocol_ (S3) and not the concrete backend. Calling it "aws-s3-..." or "minio-s3-..." is breaking that abstraction layer. That is also why we do not call it "AWS_ACCESS_KEY_ID".

A better name would be "kkp-s3-..." or "kubermatic-s3-..." to make sure it's related to KKP.

ahmedwaleedmalik:
Ah I forgot that minio would just need the creds for ANY S3 compatible storage.

xrstf:
The backup stuff isn't tied to Minio or AWS. You can use any S3-compatible object store.

ahmedwaleedmalik:
Yeah my poor memory didn't comply with me, I remember it now :)

ahmedwaleedmalik:
@xrstf PTAL

ahmedwaleedmalik:
> Pretty sure some of the etcd backup logic expects this secret at that name, you might need to check that out. Probably also needs an update in our docs and an Action required notice if we choose to change the secret name expected by the backup.

I'll update the docs. Not sure if this should break ANYTHING since we are updating all the references for the credentials as well.

embik:
> > Pretty sure some of the etcd backup logic expects this secret at that name, you might need to check that out. Probably also needs an update in our docs and an Action required notice if we choose to change the secret name expected by the backup.
> 
> I'll update the docs. Not sure if this should break ANYTHING since we are updating all the references for the credentials as well. But I can be wrong since I haven't explored the etcd backups part in KKP.

In that case, the release note needs an action required, given that people might create this credential manually when they dont use minio for backup. So they need to create the new Secret manually.

ahmedwaleedmalik:
@embik Updated, please check. Let me know, if the description is not sufficient.

xrstf (approved):
/approve

PR #9232, opened 2022-03-04
rename zone_suffixes to zoneSuffixes in error message

PratikDeoghare:
Rename for better error message. This one was missed during refactor.

```release-note
NONE
```


xrstf:
The error message specifically describes the _old_ setting name, because if there is a problem, the user has to fix it in the old CRDs, not the new ones. And if the user adds `zoneSuffixes` to the old resources, it a) won't work and b) won't give any feedback because the old CRDs had no schema.

So I don't think this PR is doing the right thing.

xrstf (approved):
/approve

xrstf:
/hold

something is broken on CI

PR #9233, opened 2022-03-06
temporarily disable storage tests for AWS

xrstf:
For unknown reasons, storage tests currently do not work on AWS, as once the PVC is gone, AWS is unable to detach the volume, leading to a stale PV, which in turn blocks further cleanup of the KKP cluster object.

> Normal  VolumeDelete  10s (x6 over 72s)  persistentvolume-controller  error deleting EBS volume "vol-069200b78803e177d" since volume is currently attached to "i-0fece311fec5b805f"

This PR is only here to unblock our CI pipeline. We do not intend to stop supporting/testing storage support on AWS.

```release-note
NONE
```


moadqassem (approved):
/approve 
/lgtm

PR #9234, opened 2022-03-06
use reconciling framework in role-cloner controller

xrstf:
This PR does a simple refactoring to get rid of home grown object creation code. Instead let's simply use our reconciling framework.

```release-note
NONE
```


xrstf:
@hdurand0710 PTAL

PR #9235, opened 2022-03-06
remove redundant SSH key detaching during cluster deletion

xrstf:
The usersshkeysynchronizer controller is already detaching SSH keys whenever a Cluster is marked for deletion. While we might argue that this is too early (makes debugging stuck clusters much harder), it at the very least makes the code in the API's DELETE endpoint redundant.

This PR removes the redundant code and also the home grown string set managing code.

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubermatic/v1/sshkeys.go
+++ b/pkg/apis/kubermatic/v1/sshkeys.go
@@ -18,6 +18,7 @@ package v1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/apis/kubermatic/v1/sshkeys.go
```diff
@@ -52,29 +53,15 @@ type SSHKeySpec struct {
...
-	if sk.Spec.Clusters == nil {
-		return false
-	}
-	for _, name := range sk.Spec.Clusters {
-		if name == clustername {
-			return true
-		}
-	}
-	return false
+	return sets.NewString(sk.Spec.Clusters...).Has(clustername)
```
floreks, line 56: :heart:

PR #9248, opened 2022-03-07
validate openstack datacenter authURL and region

PratikDeoghare:
When these are not set you get cryptic error `unsupported protocol scheme`. Validation at `getOpenstackAuthURLAndRegion` makes obvious where the error is.  

```release-note
NONE. 
```


xrstf (approved):
Just for the record, we should not put even more validation logic in our API handlers. We should put it in our webhooks and prevent broken Seeds/datacenters.

/approve

PR #9253, opened 2022-03-07
Validate etcd backup and restore credentials

moadqassem:
During the creation of the a seed object, there is no check for the secret which has the credentials for the etcd-backup-and-restore controller. Which means, if the controller was enabled, with any credentials, it will try to create backups regardless whether the s3 bucket credentials existence. This would lead to creating `etcdbackupconfig` that fails all the time and also prevent the cluster of being deleted, since the controller fails to clean the etcdbackupconfig objects. Thus the cluster will be in a stuck-in-deletion state. 

This PR makes sure that, if the seed enables the EtcdBackupAndRestore controller, it validates that the secret exists and if not, it will reject it. 
Fixes #

**Documentation**:

```release-note
None
```

### DIFF
--- a/pkg/webhook/seed/validation.go
+++ b/pkg/webhook/seed/validation.go
@@ -26,7 +26,9 @@ import (
 	"k8c.io/kubermatic/v2/pkg/features"
 	"k8c.io/kubermatic/v2/pkg/provider"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
@@ -189,6 +191,16 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object, isDelete b
 
 		if _, exists := subject.Spec.EtcdBackupRestore.Destinations[subject.Spec.EtcdBackupRestore.DefaultDestination]; !exists {
 			return fmt.Errorf("invalid etcd backup configuration: default destination %q does not exist", subject.Spec.EtcdBackupRestore.DefaultDestination)
+		}
+
+		for name, dest := range subject.Spec.EtcdBackupRestore.Destinations {
+			if dest.Credentials != nil {
+				etcdBackupSecret := corev1.Secret{}
+				if err := seedClient.Get(ctx, types.NamespacedName{Name: dest.Credentials.Name,
+					Namespace: dest.Credentials.Namespace}, &etcdBackupSecret); err != nil {
+					return fmt.Errorf("invalid etcd backup configuration: invalid destination %q credentials %s: %w", name, dest.Credentials.Name, err)
+				}
+			}
 		}
 	}
 


xrstf:
Validating the credentials is good, but what happens if the credentials simply change out of Kubermatic's control? Like someone rotating the AWS S3 credentials without updating the KKP configuration? Should our cleanup code, which would currently be blocked by this, be more forviging?

moadqassem:
> Validating the credentials is good, but what happens if the credentials simply change out of Kubermatic's control? Like someone rotating the AWS S3 credentials without updating the KKP configuration? Should our cleanup code, which would currently be blocked by this, be more forviging?

You will be blocked anyway by the failing etcd backup

pkg/webhook/seed/validation.go
```diff
@@ -190,6 +193,12 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object, isDelete b
 		if _, exists := subject.Spec.EtcdBackupRestore.Destinations[subject.Spec.EtcdBackupRestore.DefaultDestination]; !exists {
 			return fmt.Errorf("invalid etcd backup configuration: default destination %q does not exist", subject.Spec.EtcdBackupRestore.DefaultDestination)
 		}
+
+		etcdBackupSecret := corev1.Secret{}
+		if err := seedClient.Get(ctx, types.NamespacedName{Name: resources.EtcdRestoreS3CredentialsSecret,
```
lsviben: we need to check if the credentials of each destination exist, not `resources.EtcdRestoreS3CredentialsSecret`.

`resources.EtcdRestoreS3CredentialsSecret` is a leftover of old code, ill remove it in a separate PR.

moadqassem: Done

moadqassem:
> I just realized one thing. We should only check if Credentails exist, IF they are set.
> 
> Because a destination can exist without credentials, although it wont work, and backup/restore jobs wont get started. Which is better then bad credentials, where jobs are started, but stuck because the secret is missing.

Hmm not exactly, so what will happen is, a default etcdbackupconfig is gonna be created, regardless of the secret being there or not. That backup will live in the cluster namespace forever with a status Failed. And since the credentials are wrong, then the controller won't be able to cleanup those which eventually would lead to cluster being stuck in termination forever.

pkg/webhook/seed/validation.go
```diff
@@ -190,6 +192,16 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object, isDelete b
 		if _, exists := subject.Spec.EtcdBackupRestore.Destinations[subject.Spec.EtcdBackupRestore.DefaultDestination]; !exists {
 			return fmt.Errorf("invalid etcd backup configuration: default destination %q does not exist", subject.Spec.EtcdBackupRestore.DefaultDestination)
 		}
+
+		for _, dest := range subject.Spec.EtcdBackupRestore.Destinations {
+			if dest.Credentials != nil {
+				etcdBackupSecret := corev1.Secret{}
+				if err := seedClient.Get(ctx, types.NamespacedName{Name: dest.Credentials.Name,
+					Namespace: dest.Credentials.Namespace}, &etcdBackupSecret); err != nil {
+					return fmt.Errorf("invalid etcd backup configuration: invalid etcd backup configs for destination %v: %w", dest.Credentials.Name, err)
```
xrstf: > invalid etcd backup configuration: invalid etcd backup configs for destination %v

sounds a bit stuttery. WDYT about

> invalid etcd backup configuration: invalid destination %s: %w

and replace %s with the name of the destination, not the same of the credential Secret.

moadqassem: There is no name for a destination

moadqassem: That's why I used the credential ;). Either cases I updated the error message

xrstf (approved):
/approve

Issue #6231, opened 2020-12-01
Container Registry - user acccount management


shaase-ctrl:
Seed Cluster
As a user, I want to store a second set of credentials or dockerhub paid users
- Image pull secret is available.

User Cluster
As a user, I want to configure the dockerhub credential for each user cluster so that I can make use of my paid account.
- No configuration possible at the moment.

Thoughts & Criteria:
- GUI for credential management
- Dockerhub credentials can be stored centralized 
- Centralized account is used if no user cluster account is available
- For all users to spawn user cluster all services that are defined as standard-service by the admin can use the master/seed credentials
- ~~User cluster controller also creates pods via no configurable account (e.g. coredns pod) / dockerhub is hard coded~~ (There should be no hard coded registries anymore after #8055)
- AddOn creation has to align

PR #9260, opened 2022-03-08
Add proposal for container registry credential handling on user clusters

dermorz:
This PR is adding a proposal to add container registry credential management to KKP. This is motivated by #6231.

Yet another iteration of this proposal after constructive discussions in #sig-cluster-mgmt.

```release-note
NONE
```

### DIFF
--- a/docs/proposals/container-registry-credentials-per-usercluster.md
+++ b/docs/proposals/container-registry-credentials-per-usercluster.md
@@ -22,7 +22,7 @@
   * [Reconciliation of all Pods running on user clusters](#reconciliation-of-all-pods-running-on-user-clusters)
   * [Setting imagePullSecrets on ServiceAccounts](#setting-imagepullsecrets-on-serviceaccounts)
   * [Extension of addon templates with imagePullSecrets](#extension-of-addon-templates-with-imagepullsecrets)
-* [Task & effort](#task-&-effort)
+* [Task and effort](#task-and-effort)
 
 ## Goals
 
@@ -248,7 +248,7 @@ Pods or ServiceAccounts in addon templates could be templated to use an imagePul
 does not cover reconciled KKP components nor user apps. So this approach would need to be combined
 with at least one of the 2 above.
 
-## Task & effort
+## Task and effort
 * Create CRD `RegistryCredentialSet`
 * [UI] Extend Admin UI by Registry Credential Management
 * [UI] Extend Project Management UI by Registry Credential Management
--- a/docs/proposals/container-registry-credentials-per-usercluster.md
+++ b/docs/proposals/container-registry-credentials-per-usercluster.md
@@ -164,7 +164,7 @@ set).
 ### Agent
 
 Similar to the SSH key agent this agent watches a named Secret containing registry credentials and
-also the containerd config on the node the agent is running on. It makes sure that the containerd
+also the containerd-config on the node the agent is running on. It makes sure that the containerd-
 config always contain the selected credentials.
 
 #### Refactoring potential
--- a/docs/proposals/container-registry-credentials-per-usercluster.md
+++ b/docs/proposals/container-registry-credentials-per-usercluster.md
@@ -58,7 +58,7 @@ cluster.
 
 For this first iteration the secrets referenced on `Cluster` resources need to be created manually.
 
-There is a also the possibility to define default credentlals in the `KubermaticConfiguration` at
+There is a also the possibility to define default credentials in the `KubermaticConfiguration` at
 `spec.userCluster.imagePullSecret`. If defined, this will be used for any user cluster that has no
 explicit `imagePullSecrets` set.
 
@@ -162,7 +162,7 @@ does not cover reconciled KKP components nor user apps.
 
 ### A look ahead
 
-To make this more accessible to customers we could build ontop of this and extend the KKP dashboard
+To make this more accessible to customers we could build on top of this and extend the KKP dashboard
 with an interface to manage this. Similar to the [SSH key feature][ssh key agent] container registry
 credentials could be managed from there.
 
--- a/docs/proposals/container-registry-credentials-per-usercluster.md
+++ b/docs/proposals/container-registry-credentials-per-usercluster.md
@@ -1,0 +1,195 @@
+# Container Registry Credentials per User Cluster
+
+**Author**: Moritz Bracht (@dermorz)
+
+**Status**: Draft proposal; prototype in progress.
+
+## Goals
+
+Defining separate container registry credentials per user cluster so cluster owners can make use of
+their dockerhub credentials to avoid image pull limits or to access private images.
+
+## Non-Goals
+
+Setting local mirrors or private container registries for KKP components or addons.
+
+## Motivation and Background
+
+Based on the user story [#6231][] users want to be able to use their Docker Hub paid accounts to
+mitigate image pull limits. Setting this up cluster wide is currently not possible.
+
+### Current state
+
+Currently it's possible to set `spec.imagePullSecret` in the `KubermaticConfiguration`. The
+dockerconfig-json defined on this key is used in the Kubermatic Operator where the master- and the
+seed-controller create a `dockerconfigjson`-secret respectively. When creating the deployments for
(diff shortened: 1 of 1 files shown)


csengerszabo:
Walking through this PR, based on my understanding, these questions have been arisen in me:

- As we'd like to manage this centrally and we should have a centralised Docker Hub account for fallback, that means managing these credentials could be done from the admin panel just like Provider Presets. We can also match different credentials to different projects here, but probably don't want to manage cluster level from here. Or do we?

- If we add a feature like the SSH keys onto Project level, we still have to create a centralised Admin level account for a fallback. Or we can have Project-level centralised fallback accounts that should be a mandatory setup when creating a Project then.

What do you think?

dermorz:
@csengerszabo
You are absolutely right and I agree with both points. I will add a section regarding the Admin panel and also fill some gaps.

shaase-ctrl:
I just read the PR and my question is if we need this on 
- user level
- project level
- admin level
The fallback option from admin perspective is quite nice, but what if I am the editor of a cluster (not the owner) and have my own credentialset? This is possible in KKP, but might be to artificial? SSH keys are bound to the user iirc.

dermorz:
@shaase-ctrl 
Do I get you correctly that you propose to have a full hierarchy from top to bottom so the higher levels are able to create credential sets for the lower levels to select from or set them as default fallback?

Let's talk about it later in person. /cc: @csengerszabo

wurbanski:
This is a topic that comes up very often in a lot of environments, thanks for tackling this!
My question would be: if we are patching something and creating secrets anyway - maybe we could patch serviceaccounts instead so that we don't manipulate files on the node? [Doing something like this](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) should be CRI-independent (I think).

docs/proposals/container-registry-credentials-per-usercluster.md
```diff
@@ -0,0 +1,264 @@
...
+apiVersion: kubermatic.k8c.io/v1
+kind: RegistryCredentialSet
+metadata:
+  name: my-registry-credentials
+  uid: ..
+  ownerReferences:
+  - apiVersion: kubermatic.k8c.io/v1
+    kind: Project
+    name: <<PROJECT_ID>>
+    uid: ..
```
xrstf: Please do not use owner refs for semantical relationships. Owner refs are super clunky and make GitOps much, much harder (thanks to the UID), they cannot be displayed as extra columns in kubectl and are annoying to use in Go code (one has to walk through all the refs to find the owner ref that points to a Project).

A field in the spec would be much better IMHO.

docs/proposals/container-registry-credentials-per-usercluster.md
```diff
@@ -0,0 +1,264 @@
...
+
+* A controller within the user-cluster-controller-manager to set up the DaemonSet running the Agent
+on every user cluster node. (similar to usersshkeys)
+* The user-cluster-controller synchronizes selected RegistryCredentialSets and referenced secrets
+from the cluster namespace to the user cluster. If none is selected, it uses the default fallback
+credentials that was either set in the admin panel or in the KubermaticConfiguration.
+
+### User Cluster
+
+* The DaemonSet running the Agent on every user cluster node (see above)
```
xrstf: Wouldn't setting up the docker/containerd daemon be the job of the machine-controller/OSM? So wouldn't this be also solved by adding an `imagePullSecret` field to a Machine object?

dermorz: Good question. Definitely an approach I have to check. This would simplify things quite a bit.

embik:
To mirror my comment from Slack: I'm not sure a container runtime specific implementation is the way to go here. It will make switching out runtimes for whatever reason (maybe to offer better isolation of containers via a different runtime) harder.

The "Alternatives considered" section, to me, misses an approach of a small controller that distributes the secret to all namespaces and then a mutating webhook injecting `imagePullSecrets` into all Pods. Wouldn't that be the cleanest and most generic solution that is easily portable to all runtimes supported by Kubernetes?

dermorz:
@wurbanski @embik 
Thanks for your input. This is the kind of feedback I was looking for. How about we explore and discuss your ideas in the #sig-cluster-management weekly meeting? I will put it on the agenda.

embik:
Sounds good to me!

dermorz:
Just to have it in one place and not scattered around, I will add a few points discussed on slack:

@embik (like @wurbanski) is in favor of solving this container runtime agnostic and I agree with them. Marvin proposed injecting `imagePullSecrets` into pods via a mutating webhook.

Open question here: Can the webhook be initialized early enough to affect KKP components? If not, that case could be handled by KKP's reconciliation.

We will discuss and explore this further in the upcoming #sig-cluster-management weekly and I will adjust the proposal afterwards.

xrstf:
> Open question here: Can the webhook be initialized early enough to affect KKP components? If not, that case could be handled by KKP's reconciliation.

Which components do you mean? The only ones running on the user cluster should be the user-ssh-key-agent, right? And that comes from quay which -- for now -- has no pull limits.

docs/proposals/container-registry-credentials-per-usercluster.md
```diff
@@ -0,0 +1,207 @@
...
+  userCluster:
+    imagePullSecret: |-
+      {
+        "auths": {
+          "quay.io": {
+            "username": "<<DOCKERHUB_USERNAME>>",
+            "password": "<<DOCKERHUB_PASSWORD>>",
+          }
+        }
+      }
```
ahmedwaleedmalik: Let's not repeat the same mistake of having secrets in CR. Please use a secret ref here.

docs/proposals/container-registry-credentials-per-usercluster.md
```diff
@@ -0,0 +1,207 @@
...
+
+```yaml
+apiVersion: kubermatic.k8c.io/v1
+kind: KubermaticConfiguration
+metadata:
+  name: <<mykubermatic>>
+  namespace: kubermatic
+spec:
+  userCluster:
+    imagePullSecret: |-
```
ahmedwaleedmalik: Can we have a field to toggle creation in system namespaces i.e. `kube-system`? I think we should approach this in a more generic way then to just see this as a fix for the `docker hub rate limit` issue.

ahmedwaleedmalik: We should be sure that we filter the domain for the image and only mutate the `imagePullSecret` in the pod spec when there is a match

docs/proposals/container-registry-credentials-per-usercluster.md
dermorz: Alright. I thought consistency with the existing `imagePullSecret` for the master/seed components I use the same format. Will change to a secret ref.

docs/proposals/container-registry-credentials-per-usercluster.md
dermorz: On the pods we don't modify any secret. We merge the ones already set in the pod spec with those defined on the `Cluster` CR (and the one referenced in the `KubermaticConfiguration`).
`imagePullSecrets` is a list of secret references and the kubelet will only use those where the hostname matches. It's also fine to have multiple secrets for the same hostname, the kubelet will try all of them until it succeeds to pull the image.

dermorz: Can you elaborate on the more generic way? The docker hub rate limit issue is the initial motivation, but I don't see how this proposal is limited to only fix this.

ahmedwaleedmalik: If an end-user wants to provide credentials for their own private registry using this. Then most likely they wouldn't need to copy then over to `kube-system`. But you are absolutely right, your solution is generic.

docs/proposals/container-registry-credentials-per-usercluster.md
ahmedwaleedmalik: @xrstf PTAL

ahmedwaleedmalik (approved):
/approve :zap:

PR #9261, opened 2022-03-08
Additional configuration parameteres for metering reporting tool

### LINKED ISSUE #9182: Extend metering configuration with additional parameters
Current Metering config:
```go
// MeteringConfiguration contains all the configuration for the metering tool.
type MeteringConfiguration struct {
	Enabled bool `json:"enabled"`
	// StorageClassName is the name of the storage class that the metering tool uses to save processed files before
	// exporting it to s3 bucket. Default value is kubermatic-fast.
	StorageClassName string `json:"storageClassName"`
	// StorageSize is the size of the storage class. Default value is 100Gi.
	StorageSize string `json:"storageSize"`
}
```
Expected:
```
// MeteringConfiguration contains all the configuration for the metering tool.
type MeteringConfiguration struct {
	Enabled bool `json:"enabled"`
	// StorageClassName is the name of the storage class that the metering tool uses to save processed files before
	// exporting it to s3 bucket. Default value is kubermatic-fast.
	StorageClassName string `json:"storageClassName"`
	// StorageSize is the size of the storage class. Default value is 100Gi.
	StorageSize string `json:"storageSize"`
	// Schedule is expressed as a cron expression. Default value is "0 6 * * 1".
	Schedule string          `json:"schedule,omitempty"`
	Period   *MeteringPeriod `json:"period,omitempty"`
        // Retention specify number of days after which reports are deleted. -1 keeps reports forever
	Retention int `json:"retention,omitempty"`
}

type MeteringPeriod struct {
	Yesterday   bool `json:"yesterday,omitempty"`
	LastWeek    bool `json:"lastWeek,omitempty"`
	LastMonth   bool `json:"lastMonth,omitempty"`
	LastNumDays int  `json:"lastNumDays,omitempty"`
}
```

pkprzekwas:
Added new metering configuration parameters in order to enable customisation of metering reporting.

Fixes #9182 

This is first ticket from a series enhancing metering report configurations.

```release-note
NONE
```


docs/zz_generated.seed.yaml
```diff
@@ -337,6 +337,12 @@ spec:
   # Metering configures the metering tool on user clusters across the seed.
   metering:
     enabled: false
+    # Interval defines period for metering timestamps
+    interval: ""
+    # IntervalInDays defines period for metering timestamps. Overwrites Interval.
+    intervalInDays: 0
+    # Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+    schedule: ""
```
xrstf: Please ensure that this generated YAML actually represents the default values. For now, until we maybe have something more fancy and automated based on the openapi schema, please manually extend https://github.com/kubermatic/kubermatic/blob/master/codegen/example-yaml/main.go#L173

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -541,8 +541,25 @@ type MeteringConfiguration struct {
...
 	// StorageSize is the size of the storage class. Default value is 100Gi.
 	StorageSize string `json:"storageSize"`
+	// Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
+	// +kubebuilder:default:=`0 1 * * 6`
+	Schedule string `json:"schedule,omitempty"`
+	// Interval defines period for metering timestamps
+	// +kubebuilder:default:=week
+	Interval Interval `json:"interval,omitempty"`
+	// IntervalInDays defines period for metering timestamps. Overwrites Interval.
+	IntervalInDays int `json:"intervalInDays,omitempty"`
```
xrstf: From the comments alone I do not understand why we have 2 flags for the interval. Both seem to define the exact same thing ("period for metering timestamps"). Can "interval" not represent multiple days?

lsviben: From what I see its because the metering tool accepts different flags for the time period. 

But maybe this will be too clunky because we could have a config that looks like this:
```
...
interval: week
intervalInDays: 5
...
```
which will be confusing. 

What about just making one Interval string, no enum. And make a comment that valid values are:
- "yesterday"
- "week"
- "month"
- some int which denotes days?
and you can keep the "week" as default, and handle figuring out what is what in code?

xrstf: Do we need a human readable interval or could it be a ISO 8601 duration (for example "P6T2H") or a cron expression?

pkprzekwas: As @lsviben noticed it was done mainly to mirror flags from our metering report tool (https://github.com/kubermatic/metering/blob/master/cmd/kubermatic-metering-report/main.go#L23-L26). I do agree that having those two parameters is very confusing and I shouldn't impose metering tool's API on our CRD.

ISO 8601 as I see can define very precise duration (up to seconds) and our tool at the moment supports days as its base unit, which I guess if fine for metering reports. Cron expression doesn't fit as well as we just need to select moment in time from which we want to start calculation.

I decided to have the distinction between `Interval` and `IntervalInDays` because the metering tool works a bit differently in both cases. Not getting into the details I think this difference is negligible.

What do you think about removing `Interval` and leaving just `IntervalInDays` (possibly renamed to `Interval` `int` and documented to stress that the unit are days)?

pkprzekwas:
```suggestion
	// Interval defines the numer of days consulted in the metering report.
	// +kubebuilder:default:=7
	Interval int `json:"interval,omitempty"`
```

xrstf: A single `Interval` is IMHO certainly much clearer :)

pkprzekwas: It would simplify implementation as well

pkg/test/e2e/utils/apiclient/models/metering_configuration.go
```diff
@@ -20,6 +22,14 @@ type MeteringConfiguration struct {
 	// enabled
 	Enabled bool `json:"enabled,omitempty"`
 
+	// Interval defines the number of days consulted in the metering report.
+	// +kubebuilder:default=7
+	// Minimum: 1
+	Interval int64 `json:"interval,omitempty"`
+
+	// Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.
```
xrstf: I know I'm pedantic/stupid, but without having read any other documentation, I would still be slightly confused by having both an interval and a cron schedule. Can we add a sentence like "Even if your cron schedule would create daily reports, those reports can span N number of days [though typically this is [not?] desirable.]"?

pkprzekwas: You are right. It's not easy to decipher. I can extend description. Now that you brought this I'm also thinking about renaming `Interval` to `ReportScope`. Naming things ain't easy 😅

pkprzekwas: @xrstf I've extended the doc a bit differently than you proposed. Let me know if you find it sufficiently informative.

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -541,6 +541,15 @@ type MeteringConfiguration struct {
 	StorageClassName string `json:"storageClassName"`
 	// StorageSize is the size of the storage class. Default value is 100Gi.
 	StorageSize string `json:"storageSize"`
+	// Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron. Please take a note that Schedule is responsible
+	// only for setting the time when a report generation mechanism kicks off. The Interval MUST be set independently.
+	// configured separately.
```
xrstf: stray line?

pkprzekwas: My bad, removed.

xrstf (approved):
/approve

PR #9266, opened 2022-03-08
Resource optimization performance fixes

dharapvj:
Optimizing Resource requirements for various Core Kubermatic components as well as ancillery components installed via helm charts.

No

Our real world usage indicated number of issues in current resource assignment:
* Over allocations
    * Grafana, kube-state-metrics, prometheus thanos store, were using really less CPU than even `request`ed CPU
    * kubermatic API server, seed controller were having extra `request` CPU than their regular usage.

* Under allocation causing Crashes
    * Loki was crashing with OOM when we use longer timeframe for queries like one month
    * Kubermatic master-controller-manager was OOMing where there were many large user clusters

These changes optimize resource usage and reduce the resource requirements and cost for the worker machines.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/logging/loki/values.yaml
+++ b/charts/logging/loki/values.yaml
@@ -107,12 +107,13 @@ loki:
   replicas: 1
 
   resources:
-   limits:
-     cpu: "1"
-     memory: 700Mi
-   requests:
-     cpu: 300m
-     memory: 256Mi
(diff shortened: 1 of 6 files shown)


xrstf (approved):
/approve

PR #9278, opened 2022-03-09
give alpha releases a nice name on Github as well

xrstf:
This PR fixes the oversight where alpha releases would not get a nice-looking name on Github. Since we are already on 2.20 beta, backporting this to 2.20 is not needed.

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #9288, opened 2022-03-10
prevent addon webhook from stalling addon cleanup when Cluster is gone already

xrstf:
The new addon webhook rejected removing finalizers when the Cluster object is gone already, erroring out with

> {"level":"error","time":"2022-03-10T09:57:54.618Z","logger":"controller.kubermatic_addon_controller","caller":"controller/controller.go:317","msg":"Reconciler error","name":"kubeadm-configmap","namespace":"cluster-rv5kkdpntd","error":"failed to remove addon cleanup finalizer: failed to remove finalizers [cleanup-manifests] from Addon cluster-rv5kkdpntd/kubeadm-configmap: admission webhook \"addons.kubermatic.io\" denied the request: addon mutation request 7225ccb3-95c1-4806-84e8-49125934083d failed: Addons can only be created in cluster namespaces, but no matching Cluster was found"}

This PR makes it so that the ClusterRef is only set during creation and then is considered immutable, so that during cleanups no Cluster object is fetched/checked.

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/addon/mutation/mutation.go
+++ b/pkg/webhook/addon/mutation/mutation.go
@@ -30,6 +30,7 @@ import (
 
 	admissionv1 "k8s.io/api/admission/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

PR #9289, opened 2022-03-10
KCP proposal / research

### LINKED ISSUE #8948: Investigate KCP as a new API / API skeleton
As a user I want to have a 100% upstream compatible k8s API that is also future proof, so that I am ready for the next innovations within k8s. 

### Acceptance criteria
- Is KCP a feasible option to serve as our API?
- What timeframe do we need to consider in order to implement it?
- What constraints on the rest of the development team result from an implementation?
- How does it fit to our design (projects, users etc.)?
- What are the limitations on it?

https://github.com/kcp-dev/kcp/

pkprzekwas:
This document summarises my investigation of [kcp](https://github.com/kcp-dev/kcp).

Fixes #8948

```release-note
NONE
```
Signed-off-by: Patryk Przekwas [patryk@kubermatic.com](mailto:patryk@kubermatic.com)

### DIFF
--- a/docs/proposals/kcp.md
+++ b/docs/proposals/kcp.md
@@ -1,0 +1,71 @@
+# **KCP**
+
+**Author**: Patryk Przekwas (@pkprzekwas)
+
+**Status**: Draft proposal; research.
+
+**Issue**: https://github.com/kubermatic/kubermatic/issues/8948
+
+## Goals
+
+The document focuses on an evaluation of [kcp](https://github.com/kcp-dev/kcp) to expose Kube API to KKP end-users.
+
+## Non-Goals
+
+This text will not cover extensively `kcp`'s approach to transparent multi-clustering. The main reasons are the premature state of implementation and the clash with solutions that are already working fine in the KKP.
+
+## Motivation and Background
+
+### Current state
+
+The KKP API cannot be exposed through tools like `kubectl` to non-admin platform users. The main reasons are:
+* many Kubermatic resources are cluster scoped
+* no mechanism for suppressing fields in CRDs (e.g. exposing Clusters CRD but without `DebugLog`)
+
+### Expected state
+
+As a KKP project owner, I want to be able to interact with the project using native tools, like `kubectl`, `helm`, `flux`, or `argoCD`. At the same time, some options ought not to be exposed, e.g. enabling `DebugLog` in the user clusters (only administrators should be able to do it).
+
+### What is kcp?
+
+`kcp` is a minimal Kubernetes api-server. It doesn't know about `pods`, `deployments`, `services`, or even `nodes`. Some resources like `namespaces` or `serviceAccounts` have been preserved for multi-tenancy purposes. The tool is a good base for creating declarative APIs with the use of CRDs and a reconciliation loop. Putting it shortly, imagine a declarative Kubernetes-style API for anything separate from the Kubernetes container orchestrator. 
+
+Moreover, additional controllers have been embedded inside `kcp`. Those are designed for advanced multi-tenancy and multi-clustering scenarios. This document will cover only the former, as multi-clustering functionalities double some of existing KKP functionalities.
+
+This is still a prototype. APIs are not fixed and are expected to change.
+
+### Terminology
+
+* `kcp` - the primary executable which includes functionalities of minimal api-server. Also, the name of the prototype.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/kcp.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Implementation
+
+In the terminology section of this document, we see that some of the described terms are fitting nicely to the KKP's problem with building Kube-like API for end-users. Concepts like Projects in KKP could benefit combined or replaced with Workspaces.
+
+Unfortunately, `kcp` is still a prototype, meaning that the vast majority of ideas described in the terminology section are either not implemented or don't have a fixed API. Building PoC based on `kcp` is hard or not possible at this stage of the prototype.
+
+## Alternatives considered
+
+While `kcp` and its Workspaces have great multi-tenancy capabilities (at least on paper), there are other ongoing initiatives to solve K8s multi-tenancy issues, e.g. [A Multi Tenant Framework for Cloud Container Services](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/virtualcluster/doc/vc-icdcs.pdf).
```
xrstf, line 63: ... for example Kubermatic, which has unlimited number of userclusters. Kubermatic already has a "bringyourown" provider, where by default the usercluster will have no nodes.

How does KKP solve the problems in this proposal, compared to kcp? What advantages -- for our usecase -- does kcp offer? What advantages would we gain from simply using KKP userclusers to manage the public-facing clusters?

pkprzekwas: > ... for example Kubermatic, which has unlimited number of userclusters. Kubermatic already has a "bringyourown" provider, where by default the usercluster will have no nodes.

I need to check how this "bringyourown" provider thing works. First time I hear about it 😅 . Are you implying that we could use userclusters to expose Kube like APIs for non-admin users? (For instance, creating usercluster control plane per KKP project or a single usercluster instance for all projects with namespaced project specific resources? - just throwing some ideas at you, as you can probably simply say if those make any sense)

> How does KKP solve the problems in this proposal, compared to kcp? What advantages -- for our usecase -- does kcp offer?

At this moment there is no mechanism for exposing Kube like APIs for non admin KKP users. A single instance of `kcp` can manage thousands of Kube-APIs at negligible bootstrap time and cost. Thanks to logical clusters you have additional level of tenancy - dedicated RBAC per logical cluster instance.  Unfortunately, this is only on paper. All of my research is based on the design docs. 

> What advantages would we gain from simply using KKP userclusers to manage the public-facing clusters?

My personal opinion is that we don't have time to wait for kcp. The prototype is still evolving. Using KKP userclusters seems to be a reliable way to go, as we are the ones who maintain codebase. On the other hand, it has a burden of maintaining homemade solution.

xrstf: > (For instance, creating usercluster control plane per KKP project or a single usercluster instance for all projects with namespaced project specific resources? - just throwing some ideas at you, as you can probably simply say if those make any sense)

Both of these are options, yes.

> At this moment there is no mechanism for exposing Kube like APIs for non admin KKP users.

I don't understand this argument. It reads to me like "kcp does it, but KKP not". Yeah because we haven't added this usercluster mode to KKP. But it's trivial to do (after some developer adds it to KKP, just to be clear) just deploy a controlplane with no scheduler/controller-manager as a KKP usercluster. Probably much, much easier and quicker than integrating and relying on a 3rd party tool that does seemingly nothing more than what we would already gain from deploying a kube-apiserver.

> A single instance of kcp can manage thousands of Kube-APIs at negligible bootstrap time and cost.

And a kube-apiserver cannot?

> Thanks to logical clusters you have additional level of tenancy

We don't need them, we can just have an inifite number of KKP userclusters. At the extreme end, if we want to, we can give every user a dedicated cluster just for them.

pkprzekwas: > I don't understand this argument. It reads to me like "kcp does it, but KKP not". Yeah because we haven't added this usercluster mode to KKP. But it's trivial to do (after some developer adds it to KKP, just to be clear) just deploy a controlplane with no scheduler/controller-manager as a KKP usercluster. Probably much, much easier and quicker than integrating and relying on a 3rd party tool that does seemingly nothing more than what we would already gain from deploying a kube-apiserver.

You are right, I was wrongly focusing on the current state of KKP, not what we could achieve with userclusters. I do agree that using third party, especially one that is a prototype, makes no sense.

> And a kube-apiserver cannot?

It definitely can, but without mentioned benefits (which I believe are not deal-breakers). I was thinking about a scenario where we would have more than one API server. It would require additional resources and some time for provisioning api-server instance. Logical clusters/workspaces behave like API resources so technically are more convenient to use. They modify etcd key structure to accomplish that within one etcd instance. 

> We don't need them, we can just have an inifite number of KKP userclusters. At the extreme end, if we want to, we can give every user a dedicated cluster just for them.

Agree. All the things kcp does for multi-tenancy are very interesting concepts. But those concepts are highly opinionated, a bit confusing for regular Kubernetes users and not battle-tested. I believe same could be achieved with KKP's userclusters.

pkprzekwas:
There were some discussion on the topic and it was concluded that we as Kubermatic will not use `kcp`.

PR #9292, opened 2022-03-10
Validate project-id label on Cluster objects

xrstf:
KKP requires that clusters are always part of a project. This relationship is established via the project-id label. This PR makes it so that our webhook validates a correct label and rejects "homeless" clusters. Not having a functioning project relation means that some of our controllers would go crazy (most notably the rbac stuff).

To ensure the project continues to exist while a cluster exists, I also added a new finalizer to projects that is blocking their deletion until all clusters have been deleted.

Additionally, similarily to #9311 this PR moves the responsibility for setting the initial project phase out of the seed/project controller and into the master/rbac controller. This is because projects need to be initialized on the master cluster, not the seeds. And the RBAC controller is kind of the owner of the whole "project phase" concept. In contrast to #9311, which removes the seed/project controller alltogether, this PR repurposes it to handle the project cleanup.

**Extra Background Info**:
At first I considered not just checking whether the project exists, but also that it is *Active*. Projects become active once the RBAC controller on the master-ctrl-mgr has finished reconciling it. It felt like a nice feature to deeply check the project whenever a cluster is created.

However it turned out this is a bad idea. To get a project to become healthy, you also need to setup the owning user properly. This involves setting owner references between user and project. This makes using `kubectl` super inconvenient (waiting for the project to become active is somewhat trivial, but getting the user's UID to fill in the owner ref gets ugly fast).

So when we validate "too much", the declarative nature of Kubernetes is somewhat lost. Kubernetes is built around eventual consistency and us trying to enforce strict ACID rules is a fight we will never able to win. Also, all of our code needs to handle all eventualities and problems anyway, so having super strict webhooks with strong consistencies would not save us any code in the controller loops.

It should be possible to create a project and cluster and have them become healthy _eventually_. That is why this PR only checks the existence of a project and not its status.

```release-note
The cluster validation webhook now validates that Cluster objects have a proper `project-id` label pointing to an existing project.
```


pkg/webhook/cluster/validation/validation.go
```diff
@@ -121,3 +134,53 @@ func (v *validator) buildValidationDependencies(ctx context.Context, c *kubermat
...
+			return field.InternalError(fieldPath, fmt.Errorf("failed to list projects: %w", err))
+		}
+
+		var project *kubermaticv1.Project
+		for i, p := range projects.Items {
+			if projectID == p.Name {
+				project = &projects.Items[i]
+				break
+			}
+		}
```
lsviben: why not just `v.client.Get(..)` the project?

pkg/webhook/cluster/validation/validation.go
```diff
@@ -121,3 +134,53 @@ func (v *validator) buildValidationDependencies(ctx context.Context, c *kubermat
...
+	if oldCluster != nil && cluster.Labels[label] != oldCluster.Labels[label] {
+		return field.Invalid(fieldPath, cluster.Labels[label], fmt.Sprintf("the %s label is immutable", label))
+	}
+
+	// during cluster creation, we enforce the project label;
+	// during updates we are more relaxed and only require that the label isn't changed,
+	// so that if a project gets removed before the cluster (for whatever reason), then
+	// the cluster cleanup can still progress and is not blocked by the webhook rejecting
+	// the stale label
+	if oldCluster == nil {
```
lsviben: what do you think about skipping the update check only if the project doesnt exist?

lsviben (commented):
Generally looks great, added a small suggestion and question :)

pkg/webhook/cluster/validation/validation.go
xrstf: That.... that ... 

YAGpXPdh

xrstf:
PTAL again if you can :)

lsviben:
> PTAL again if you can :)

Oh, didnt notice the update :sweat_smile: 

/approve
/lgtm

PR #9296, opened 2022-03-10
Initialize VirtualMachineInstancePresets

### LINKED ISSUE #8490: Initialize Virtual Machine Presets

hdurand0710:

Initialisation of the VirtualMachineInstancePresets when a cluster a created.
The customer can define some VirtualMachineInstancePresets in the `default` namespace. When a cluster is created, those are cloned into the dedicated <cluter-xxxyy> dedicated namespace, allowing the VirtualMachinesInstances to match those Presets.
We also add a standard `kubermatic-standard` preset in the <cluster-xxxxyyy> namespace, which is also returned in the List VirtualMachineInstancePresets endpoint to allow the selection in the dashboard.

Fixes #8490

**Documentation**:

```release-note
KubeVirt: initialisation of VirtualMachineInstancePresets in a dedicated namespace in the infra KubeVirt Cluster.
```

### DIFF
--- a/codegen/reconcile/main.go
+++ b/codegen/reconcile/main.go
@@ -219,6 +219,12 @@ func main() {
 				ResourceImportPath: "k8c.io/kubermatic/v2/pkg/apis/apps.kubermatic/v1",
 				APIVersionPrefix:   "AppKubermaticV1",
 			},
+			{
+				ResourceName:       "VirtualMachineInstancePreset",
+				ImportAlias:        "kubevirtv1",
+				ResourceImportPath: "kubevirt.io/api/core/v1",
+				APIVersionPrefix:   "KubeVirtV1",
+			},
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


maciaszczykm:
/approve

Looks good from the API perspective.

Issue #601, opened 2017-11-28
Add range & validation for node-port selection


mrIncompetent:
It needs to be possible to define a range from which we pick the nodeports.
As big enterprises wont allow the whole 30k-32k port range.

Also we need to have a validation for the apiserver which checks if there are still ports left from the range

PR #9297, opened 2022-03-11
Added autodownsampling so that Grafana finds data on longer range

dharapvj:
Added autodownsampling so that Grafana finds data on longer range as suggested on [thanos issue tracker #601](https://github.com/thanos-io/thanos/issues/601) 

Currently, our grafana does not read downsampled Thanos data. This happens due to in default mode, Thanos will respond only with raw data - which we keep only for short duration like 7days.

So effectively, we cannot make use of Thanos for long term analysis at all with current setting.

Similar change was also [done](https://github.com/prometheus-operator/prometheus-operator/pull/2441) in prometheus operator

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/monitoring/prometheus/templates/thanos-query-deployment.yaml
+++ b/charts/monitoring/prometheus/templates/thanos-query-deployment.yaml
@@ -44,6 +44,7 @@ spec:
         args:
         - query
         - --query.replica-label=replica
+        - --query.auto-downsampling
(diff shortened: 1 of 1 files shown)


toschneck:
Please backport to 2.18 aas this is needed for our customer.

xrstf (approved):
/approve

> /cherry-pick release-2.18

This is not how the release branch is named.

PR #9159, opened 2022-03-01
Fix endpoint to create project

ahmedwaleedmalik:
Signed-off-by: Waleed Malik <ahmedwaleedmalik@gmail.com>

**What does this PR do / Why do we need it**:
Project creation is resulting in a conflict since the project is reconciled instantaneously. This results in an update in its state; finalizers and org annotation is added. To resolve this we explicitly fetch the resource again before patching the status.

**Does this PR close any issues?**:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #9311, opened 2022-03-12
Fix setting initial project phase

xrstf:
During preparations for 2.20, we added a new controller to the *seed-controller-manager* in #9159. Its job was to set the initial project phase. However this must be done on the **master** cluster, not on the seed. From the master it is then synchronized down into each seed by the `project-synchronizer` controller.

This PR removes the new controller and updates the RBAC controller to set the initial phase. The RBAC controller is kind of the "owner" of the concept of project phases (it's the only controller to ever flip a project from Inactive to Active), so it makes sense that this controller also sets the initial phase.

Note that #9292 is the PR I realized our mistake in. However that PR refactors project cleanup and *does* in fact need a "project" controller on the seed-controller-manager.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #9320, opened 2022-03-14
update cert-manager Go dependency to 1.7.1

xrstf:
We only use this during installation, so the upgrade is pretty safe and easy.

```release-note
NONE
```


xrstf:
@wurbanski PTAL

xrstf:
@wurbanski  PTAL

wurbanski:
Versions 1.6 and 1.7 deprecate the old APIs (v1alpha2, v1alpha3, and v1beta1) but we do not use them explicitly in the installer, so it should be fine, I think. 

/lgtm

PR #9321, opened 2022-03-14
migrate from github.com/robfig/cron to v3

xrstf:
v3 was already released more than 2 years ago, I think it's time for us to upgrade.

I copied `cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor` from v1's definition of the "standard parser": https://github.com/robfig/cron/blob/v1/parser.go#L158-L160

```release-note
NONE
```


xrstf:
@imharshita PTAL

PR #9322, opened 2022-03-14
Adjust ETCD_INITIAL_CLUSTER during peer TLS migration

### LINKED ISSUE #9369: etcd-launcher unable to rejoin etcd ring during peer TLS migration
### What happened?

When `etcd-launcher` is enabled on a cluster, the removal of an etcd member while the peer TLS migration is ongoing will not allow the member to rejoin the cluster automatically with the following error:

```
unmatched member while checking PeerURLs (len(["http://etcd-1.etcd.cluster-xxxxxxxx.svc.cluster.local:2380" "https://etcd-1.etcd.cluster-xxxxxxxx.svc.cluster.local:2381"]) != len(["http://etcd-1.etcd.cluster-xxxxxxxx.svc.cluster.local:2380"]))
```

This comes from the `ETCD_INITIAL_CLUSTER` environment variable that is generated by `etcd-launcher`. It either generates the peer list with the plain http port (if peer TLS is not enabled) or with the TLS port only. That is not a problem for existing members that restart, so the peer TLS migration works. It's also correct once the peer TLS migration has finished, so new nodes added afterwards do not have a problem either. But during the migration, (some) members have both ports advertised as peer ports. That does not match what is in `ETCD_INITIAL_CLUSTER`, and members freshly joining the cluster fail to join because of that.

There is no intention in the migration to remove existing members. The most likeliest candidate are older versions of `etcd-launcher` that attempt to clean up the etcd member list in error, but so far that is not confirmed.

### What should happen?

`etcd-launcher` enabled etcd members should be able to automatically rejoin the etcd ring if they get removed in error while the peer TLS migration happens.

### How to reproduce?

1. create a normal cluster and wait for it to deploy
2. enable the etcd launcher feature flag
3. Observe the upgrade starting. the STS needs to be updated to do the "mixed" etcd-launcher mode (this is running a TLS peer migration)
4. set the cluster to pause while the STS is still updating, wait for it to finish
5. Drop into a shell and run `etcdctl member list` and `etcdctl member remove <ID>` on the member you want to drop
6. That member will restart, try to join the cluster, and start failing with the peer URL issues we've seen before

### Environment

- KKP Version: 2.19 or higher
- Shared or separate master/seed clusters?: -

embik:
See #9369 for more details of the problem, but the gist is: When an etcd member is removed during the peer TLS migration, `etcd-launcher` fails to rejoin the etcd ring automatically. This PR makes `etcd-launcher` consider the peer TLS migration when generating the `ETCD_INITIAL_CLUSTER` env variable for "new" / re-joining members. It does so by querying pod state from Kubernetes and generate the peer URLs for the initial cluster list based on the pod configuration.

Fixes #9369

**Documentation**:

```release-note bugfixes
etcd-launcher is now capable of automatically rejoining the etcd ring when a member is removed during the peer TLS migration
```

### DIFF
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -381,7 +381,6 @@ func initialMemberList(n int, namespace string, useTLSPeer bool, client ctrlrunt
 				)
 			}
 		}
-
 	}
 
 	return members
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -104,7 +104,7 @@ func main() {
 		log.Panicw("failed to set initialState", zap.Error(err))
 	}
 
-	e.initialMembers = initialMemberList(e.clusterSize, e.namespace, e.usePeerTLSOnly, e.clusterClient)
+	e.initialMembers = initialMemberList(e.clusterSize, e.namespace, e.usePeerTLSOnly, e.clusterClient, log)
 
 	log.Info("initializing etcd..")
 	log.Infof("initial-state: %s", e.initialState)
@@ -356,7 +356,7 @@ func (e *etcdCluster) updatePeerURLs(log *zap.SugaredLogger) error {
 	return nil
 }
 
-func initialMemberList(n int, namespace string, useTLSPeer bool, client ctrlruntimeclient.Client) []string {
+func initialMemberList(n int, namespace string, useTLSPeer bool, client ctrlruntimeclient.Client, log *zap.SugaredLogger) []string {
 	members := []string{}
 	for i := 0; i < n; i++ {
 		if useTLSPeer {
@@ -365,8 +365,10 @@ func initialMemberList(n int, namespace string, useTLSPeer bool, client ctrlrunt
 			members = append(members, fmt.Sprintf("etcd-%d=http://etcd-%d.etcd.%s.svc.cluster.local:2380", i, i, namespace))
 			// check if the pod is already annotated as TLS aware
 			var pod *corev1.Pod
-			if err := client.Get(context.Background(), types.NamespacedName{Name: "etcd", Namespace: namespace}, pod); err == nil {
-				if _, ok := pod.Annotations[resources.EtcdTLSEnabledAnnotation]; ok {
+			if err := client.Get(context.Background(), types.NamespacedName{Name: fmt.Sprintf("etcd-%d", i), Namespace: namespace}, pod); err != nil {
+				log.Warnw("failed to get Pod information for etcd", zap.Error(err))
+			} else {
+				if _, ok := pod.ObjectMeta.Annotations[resources.EtcdTLSEnabledAnnotation]; ok {
 					members = append(
 						members,
 						fmt.Sprintf("etcd-%d=https://etcd-%d.etcd.%s.svc.cluster.local:2381", i, i, namespace),
--- a/cmd/etcd-launcher/main.go
+++ b/cmd/etcd-launcher/main.go
@@ -68,6 +68,7 @@ const (
 type etcdCluster struct {
 	namespace             string
 	clusterSize           int
+	clusterClient         ctrlruntimeclient.Client
 	podName               string
 	podIP                 string
 	etcdctlAPIVersion     string
@@ -75,6 +76,7 @@ type etcdCluster struct {
 	token                 string
 	enableCorruptionCheck bool
 	initialState          string
+	initialMembers        []string
 	usePeerTLSOnly        bool
 }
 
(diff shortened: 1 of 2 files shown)


lsviben:
@moelsayed  could you give this a look. For me it makes sense, but you are the expert :)

cmd/etcd-launcher/main.go
```diff
@@ -89,22 +91,24 @@ func main() {
...
+	if err := e.setClusterSize(); err != nil {
 		log.Panicw("failed to set cluster size", zap.Error(err))
 	}
 
-	if err := e.setInitialState(clusterClient, log); err != nil {
+	if err := e.setInitialState(log); err != nil {
 		log.Panicw("failed to set initialState", zap.Error(err))
 	}
 
+	e.initialMembers = initialMemberList(e.clusterSize, e.namespace, e.usePeerTLSOnly, e.clusterClient, log)
```
kron4eg, line 107: Looks like initialMemberList() should became a etcdCluster method, since it uses so many fields.

embik: It felt redundant given that `initialMembers` is already a field of `etcdCluster`. Hard to gauge for me when it should switch to become a method, but this way the function signature makes clear what is used to determine the member list. I think I prefer that to be honest.

cmd/etcd-launcher/main.go
```diff
@@ -352,20 +356,46 @@ func (e *etcdCluster) updatePeerURLs(log *zap.SugaredLogger) error {
...
-		format = "etcd-%d=https://etcd-%d.etcd.%s.svc.cluster.local:2381"
-	}
-
+func initialMemberList(n int, namespace string, useTLSPeer bool, client ctrlruntimeclient.Client, log *zap.SugaredLogger) []string {
 	members := []string{}
 	for i := 0; i < n; i++ {
-		members = append(members, fmt.Sprintf(format, i, i, namespace))
+		var pod corev1.Pod
+		if err := client.Get(context.Background(), types.NamespacedName{Name: fmt.Sprintf("etcd-%d", i), Namespace: namespace}, &pod); err != nil {
+			log.Warnw("failed to get Pod information for etcd, guessing peer URLs", zap.Error(err))
```
kron4eg, line 364: Why this error is not important?

embik: 99.99% of the time the information extracted from the Pods isn't needed. We only need it when the etcd ring is mid-migration, in all other situations the "guess" based on local information is enough. It seemed excessive to fail and possibly take down etcd rings when the seed cluster Kubernetes API has a hiccup or the client cache isn't aware of a new Pod yet.

cmd/etcd-launcher/main.go
```diff
@@ -352,20 +356,46 @@ func (e *etcdCluster) updatePeerURLs(log *zap.SugaredLogger) error {
...
+			if !hasStrictTLS(&pod) {
+				members = append(members, fmt.Sprintf("etcd-%d=http://etcd-%d.etcd.%s.svc.cluster.local:2380", i, i, namespace))
+			}
+
+			if _, ok := pod.ObjectMeta.Annotations[resources.EtcdTLSEnabledAnnotation]; ok {
+				members = append(
+					members,
+					fmt.Sprintf("etcd-%d=https://etcd-%d.etcd.%s.svc.cluster.local:2381", i, i, namespace),
+				)
+			}
```
kron4eg, line 382: I think the those two conditions should be mutually exclusive

embik: they shouldn't, or at least they were deliberately not mutually exclusive when I implemented them. The way you handle multiple peer URLs for a single etcd member is having several entries in the `ETCD_INITIAL_CLUSTER` list (I found this behaviour to be somewhat poorly documented and cannot find a reference to this in the docs right now). So, if `etcd-0` has two peer URLs, it should look like this:

```bash
ETCD_INITIAL_CLUSTER=etcd-0=http://etcd-0:2380,etcd-0=https://etcd-0:2381,etcd-1=http://etcd-1:2380,etcd-2=http://etcd-2:2380
```

While the migration is happening, both peer URLs need to be in the cluster list, and as such both conditions will be (and need to be) true.

embik: The answer to "why did this even work before" (which you haven't asked but I assume is a question one might have after the explanation for `ETCD_INITIAL_CLUSTER`) is that `ETCD_INITIAL_CLUSTER` is only used when an etcd node tries to join a cluster, which usually is not the case during the peer TLS migration. So the incorrect list was ignored, unless a node happened to be removed mid-migration.

kron4eg: ETCD_INITIAL_CLUSTER is used only once, at cluster bootstrap, when cluster is assembled for the first time. After cluster is started ETCD_INITIAL_CLUSTER has no meaning, it doesn't reflect the current member list.

Moreover: https://etcd.io/docs/v3.5/op-guide/clustering/ states that
> Note that the URLs specified in `initial-cluster` are the advertised peer URLs, i.e. they should match the value of `initial-advertise-peer-urls` on the respective nodes.

embik: No, `ETCD_INITIAL_CLUSTER` is not used at cluster bootstrap, but at _node_ bootstrap (which a node performs when it's not a member yet; if the cluster is new, that's also the cluster bootstrap of course). Otherwise, `ETCD_INITIAL_CLUSTER_STATE=existing` would not exist. Also see here: https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#add-a-new-member:

> Start the new member with the new cluster configuration, including a list of the updated members (existing members + the new member).

As for the link, yes, that's what I mean by poor documentation. It does not well reflect the format needed for `initial-cluster`. I will try to find a source for my claims.

embik: @kron4eg I'm still struggling to find a reference for this, which is crazy to be honest, but here's a few bits from the etcd source code:

- The initial cluster is treated as an `URLsMap`: https://github.com/etcd-io/etcd/blob/bfb9aa42059d84ea60a4cddd75d7f3e94b27935c/server/embed/config.go#L803
- The code doc for `NewURLsMap` even shows an example where two URLs are added for the same name: https://github.com/etcd-io/etcd/blob/bfb9aa42059d84ea60a4cddd75d7f3e94b27935c/client/pkg/types/urlsmap.go#L26-L41 (`mach0=http://1.1.1.1:2380,mach0=http://2.2.2.2::2380,mach1=http://3.3.3.3:2380,mach2=http://4.4.4.4:2380`)

There's actually no other way to pass multiple peer URLs for one etcd member in the `ETCD_INITIAL_CLUSTER` env variable.

kron4eg (approved):

PR #9323, opened 2022-03-14
remove -node-hyperkube-image flag from machine-controller

xrstf:
The `-node-hyperkube-image` flag was removed from the machine-controller via kubermatic/machine-controller#1060, which was released in MC v1.36+ a long time ago.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #9325, opened 2022-03-14
use seedsGetter in master-constraint-template-synchronizer controller

xrstf:
One TODO comment less and some less code duplication.

```release-note
NONE
```


embik (approved):
/approve

PR #8960, opened 2022-02-10
Use new version of the Anexia Cloud Controller Manager

kstiehl:
**What does this PR do / Why do we need it**:

* This PR updates the Anexia cloud controller manager to use the version 1.1.3
** Now supports service of type loadbalancer
** contains bug fixes

**Does this PR close any issues?**:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Anexia - update the cloud controller version manager (v1.1.3) to support services of type load balancer
```

PR #9335, opened 2022-03-15
Update Anexia CCM

LittleFox94:

This PR updates Anexia CCM version, replacing PR #8960 because @kstiehl is not working for us anymore (thanks for all your work!).

Fixes #8960 

Sorry for the additional noise, the new repo we originate these PRs from is now in the organization, so stuff like this will not be required again.

I also saw on the original PR you are trying to figure out how to test it - is there any way we can help with this to get this merged?

**Documentation**:

```release-note
Anexia: update cloud controller manager to v1.4.1, now supporting LoadBalancer services
```

### DIFF
--- a/pkg/resources/cloudcontroller/anexia.go
+++ b/pkg/resources/cloudcontroller/anexia.go
@@ -63,10 +63,11 @@ func anexiaDeploymentCreator(data *resources.TemplateData) reconciling.NamedDepl
 			deployment.Spec.Template.Spec.Containers = []corev1.Container{
 				{
 					Name:  ccmContainerName,
-					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:0.1.0",
+					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:1.1.3-kubermatic",
 					Command: []string{
 						"/app/ccm",
 						"--cloud-provider=anexia",
+						fmt.Sprintf("--cluster-name=%s", data.Cluster().Spec.HumanReadableName),
 						"--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig",
 					},
 					Env: []corev1.EnvVar{
--- a/pkg/resources/cloudcontroller/anexia.go
+++ b/pkg/resources/cloudcontroller/anexia.go
@@ -73,7 +73,7 @@ func anexiaDeploymentCreator(data *resources.TemplateData) reconciling.NamedDepl
 			deployment.Spec.Template.Spec.Containers = []corev1.Container{
 				{
 					Name:  ccmContainerName,
-					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:1.1.3-kubermatic",
+					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:1.4.1",
 					Command: []string{
 						"/app/ccm",
 						"--cloud-provider=anexia",
@@ -85,6 +85,10 @@ func anexiaDeploymentCreator(data *resources.TemplateData) reconciling.NamedDepl
 						{
 							Name:  "ANEXIA_TOKEN",
 							Value: credentials.Anexia.Token,
+						},
+						{
+							Name:  "ANEXIA_AUTO_DISCOVER_LOAD_BALANCER",
+							Value: "true",
 						},
 					},
 					Ports: []corev1.ContainerPort{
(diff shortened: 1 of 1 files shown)


pkg/resources/cloudcontroller/anexia.go
```diff
@@ -58,22 +58,38 @@ func anexiaDeploymentCreator(data *resources.TemplateData) reconciling.NamedDepl
...
 			deployment.Spec.Template.Spec.Containers = []corev1.Container{
 				{
 					Name:  ccmContainerName,
-					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:0.1.0",
+					Image: data.ImageRegistry(resources.RegistryAnexia) + "/anexia/anx-cloud-controller-manager:1.3.0",
 					Command: []string{
 						"/app/ccm",
 						"--cloud-provider=anexia",
+						"--cloud-config=/etc/kubernetes/cloud/config",
+						fmt.Sprintf("--cluster-name=%s", data.Cluster().Spec.HumanReadableName),
```
LittleFox94, line 81: This should probably set to Kubernetes controller-manager instead, as it is done here for azure: https://github.com/kubermatic/kubermatic/blob/c5daecbb129173098df5ddf82eb3c94482855d21/pkg/resources/controllermanager/deployment.go#L259

is there a reason this isn't just set for every cluster? should I extract this question into an issue?

ahmedwaleedmalik (approved):
/approve
/hold for @moadqassem to confirm if this should be tested via CI.

LittleFox94:
@ahmedwaleedmalik:
> @LittleFox94 Thanks for your contribution. We rely on testing changes at our end before rolling them out to the general public. How can we test this? Given that we don't have an active subscription/account on Anexia, at least I am not aware of one.

Hi, sorry for getting back this late

afaik you have an account in our Engine for integration tests for both KKP and machine-controller, though I never saw anything running from you there (we also use it for development of related things - like our CCM). Just checked and see a lot of deactivated users, but also a user specifically for integration testing who is activated. Can you ask around internally, I'll do the same here? If you email me (address in commits), I can give you some names I can see - don't want to leak them via public GitHub comments.

We are also currently working on getting our CCM OpenSource'd. While we do not have integration tests at the moment, that's something we'll build in the short- to midterm. The part in KKP is only about volumes and arguments given to CCM, so maybe that will be enough already?

On another note, another release of our CCM is probably coming very soon and it is still in active development. Are there better ways to have it updated in KKP than via pull requests? Don't want to spam you ^^'

LittleFox94:
I just changed the CCM version to use and rebased the branch on master.

My last comment on this PR had a yet-unanswered question, just making sure it's seen :)

@ahmedwaleedmalik quick status update: will probably get back to your mail later today or tomorrow - hopefully with a working test setup :)

ahmedwaleedmalik:
> On another note, another release of our CCM is probably coming very soon and it is still in active development. Are there better ways to have it updated in KKP than via pull requests? Don't want to spam you

The best and only way would be via PRs. Once we have those e2e tests set up then it would be really easy to move forward with these PRs.

pkg/resources/cloudcontroller/anexia.go
ahmedwaleedmalik: This depends on the CCM implementation for the cloud provider, like azure has support for this but others don't necessarily support this flag.

LittleFox94: That option is included in [`k8s.io/cloud-provider/options.KubeCloudSharedOptions`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/options/kubecloudshared.go#L59), but CCMs aren't required to use `k8s.io/cloud-provider` so yeah, correct, depends on CCM.

ahmedwaleedmalik (approved):
/approve ⚡ 

Without this, the CCM is non-functional at the moment. I've tested this PR by creating a cluster manually and that works fine. This is required to proceed with https://github.com/kubermatic/kubermatic/pull/9561

xrstf:
@LittleFox94 I just randomly stumbled upon this PR and the line

```
fmt.Sprintf("--cluster-name=%s", data.Cluster().Spec.HumanReadableName)
```

Are you sure this is the intended solution, rather than `data.Cluster().Name`? The human readable name is completely freeform and can contain emojis.

LittleFox94:
@xrstf whoops, forgot about this comment

It definitely makes sense in our current setup, but maybe it is problematic with clusters not created via our automations (which filter the allowed names)..

I'll take this as a TODO in our team, but iirc the biggest problem we had was not really having a way to pass some custom configuration into our CCM, this being the most generic way.

PR #9337, opened 2022-03-15
Add Control Plane versions to the ClusterStatus

xrstf:
This PR is part of #5949. To handle gradual deployments of the control plane, a new update controller will smartly slowly roll out first the apiserver, then the rest of the control plane etc.

To keep the review simple, I split the branch into multiple PRs and this is the first one. It only adds the new fields to the ClusterStatus and has a dummy implemenntation in the old update controller that simply mirrors `spec.version` into the status.

In general, `spec.version` should now be seen as "the eventual goal we want to reach", never as "the current control plane version". Instead, `status.versions.controlPlane` is the currently active version. So most of the reconciling code that depends on the cluster version now uses this field in the status, except for etcd and CoreDNS, which are bound the `status.versions.apiserver`.

```release-note
Add new ClusterVersionsStatus to the ClusterStatus which represents the currently active control plane versions. `cluster.spec.version` should now always be treated as the intended, eventual version, not the current version.
```


pkg/handler/common/cluster.go
```diff
@@ -1004,7 +1004,7 @@ func ConvertInternalClusterToExternal(internalCluster *kubermaticv1.Cluster, dat
 			CNIPlugin:                            internalCluster.Spec.CNIPlugin,
 		},
 		Status: apiv1.ClusterStatus{
-			Version:              internalCluster.Spec.Version,
+			Version:              internalCluster.Spec.Version, // TODO: Should this use .Status.Versions instead of the spec?
```
embik: IMO yes, this should use the Controlplane version status.

pkg/resources/controllermanager/deployment.go
```diff
@@ -243,7 +245,7 @@ func getFlags(data *resources.TemplateData) ([]string, error) {
 	featureGates := []string{"RotateKubeletServerCertificate=true"}
 
 	// starting with k8s 1.21, this is always true and cannot be toggled anymore
-	if cluster.Spec.Version.Semver().Minor() < 21 {
+	if version.Minor() < 21 {
```
embik: This is kind of out of scope, but it's gonna break if we ever get k8s 2.0. Can we make it more robust and include the major version in this check?

pkg/resources/etcd/statefulset.go
```diff
@@ -363,7 +363,12 @@ func GetBasePodLabels(cluster *kubermaticv1.Cluster) map[string]string {
 // ImageTag returns the correct etcd image tag for a given Cluster
 // TODO: Other functions use this function, switch them to getLauncherImage.
 func ImageTag(c *kubermaticv1.Cluster) string {
-	if c.Spec.Version.Semver().Minor() < 22 {
+	// most other control plane parts refer to the controller-manager's version, which
+	// during updates lacks behind the apiserver by one minor version; this is so that
+	// also external components like the kubernetes dashboard or external ccms wait for
+	// the new apiserver to be ready; etcd however is different and gets updated together
+	// with the apiserver
+	if c.Status.Versions.Apiserver.Semver().Minor() < 22 {
```
embik: same here IMHO. If we touch it, can we maybe match it against `< 1.22`?

embik (changes requested):
few minor comments that we could improve while we're working on the version constraints.

embik (approved):
/approve

xrstf:
@lsviben After Marvin's feedback, our API now possibly returns an empty string for the cluster version (see the "adjust unit tests" commit). I'm not sure if something in the dashboard would explode if there is no version set in the ClusterStatus.

embik:
That's actually a good point I haven't considered when giving that feedback. I guess the dashboard should show a spinning icon when the field is empty, but yeah it might be broken with that right now?

lsviben:
> @lsviben After Marvin's feedback, our API now possibly returns an empty string for the cluster version (see the "adjust unit tests" commit). I'm not sure if something in the dashboard would explode if there is no version set in the ClusterStatus.

Ill check, if I wont figure it out ill ask somebody from UI team :)

lsviben:
> > @lsviben After Marvin's feedback, our API now possibly returns an empty string for the cluster version (see the "adjust unit tests" commit). I'm not sure if something in the dashboard would explode if there is no version set in the ClusterStatus.
> 
> Ill check, if I wont figure it out ill ask somebody from UI team :)

Ok I ran this locally, and nothing happens. Looks like the UI is not using this field for the version. It is using the version in the cloud provider spec.

Ill do a quick review of this PR now.

embik:
@xrstf btw, I think that this PR should come with a release note since it changes the behaviour of the KKP API status field as you said.

xrstf:
@embik good point, I need to remind myself that our types are now an actual public API. :)

embik (approved):

PR #9339, opened 2022-03-16
Fix StorageContainer default from Default to SelfServiceContainer

mate4st:

Change the StorageContainer default to an existing Storage Container

```release-note
NONE
```


embik (approved):
/approve

PR #9343, opened 2022-03-16
machine controller health check moved to user-cluster-ctrl-manager

### LINKED ISSUE #8837: Initial MachineDeployment bypass the machine-controller-webhook
**User Story** 
It seems that the initial machine deployment created in the very early stage of the user cluster creation bypasses the validation process by the machine controller webhook. The MD object is being create by a dedicated controller which watches for cluster object, and apply the MD once the user cluster api server is there. This would lead to create the MD without validating it, since the machine controller webhook and the configurations are not rolled out yet. 

**Acceptance Criteria**
- Initial MD is validated via the machine controller webhook

mlavacca:
The health check for the Machine controller has been moved to the user-cluster-controller-manager to ensure that:
* the `machineController` deployment is healthy
* the `machineControllerWebhook` deployment is healthy
* the `machineController` `mutatingWebhookConfiguration` has been correctly created.

The last condition is ensured by the fact that when the `healthCheck` function is called, all the needed resources have been created (among which the `machineController` `mutatingWebhookConfiguration`)

Fixes #8837 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/health.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/health.go
@@ -43,7 +43,6 @@ func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cl
 		resources.ApiserverDeploymentName:             {healthStatus: &extendedHealth.Apiserver, minReady: 1},
 		resources.ControllerManagerDeploymentName:     {healthStatus: &extendedHealth.Controller, minReady: 1},
 		resources.SchedulerDeploymentName:             {healthStatus: &extendedHealth.Scheduler, minReady: 1},
-		resources.MachineControllerDeploymentName:     {healthStatus: &extendedHealth.MachineController, minReady: 1},
 		resources.OpenVPNServerDeploymentName:         {healthStatus: &extendedHealth.OpenVPN, minReady: 1},
 		resources.UserClusterControllerDeploymentName: {healthStatus: &extendedHealth.UserClusterControllerManager, minReady: 1},
 	}
--- a/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go
@@ -227,10 +227,8 @@ func (r *reconciler) reconcile(ctx context.Context) error {
 		}
 	}
 
-	if r.opaIntegration || r.userClusterMLA.Logging || r.userClusterMLA.Monitoring {
-		if err := r.healthCheck(ctx); err != nil {
-			return err
-		}
+	if err := r.healthCheck(ctx); err != nil {
+		return err
 	}
 
 	if !r.userClusterMLA.Logging && !r.userClusterMLA.Monitoring {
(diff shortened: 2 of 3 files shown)


pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -43,7 +43,6 @@ func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cl
 		resources.ApiserverDeploymentName:             {healthStatus: &extendedHealth.Apiserver, minReady: 1},
 		resources.ControllerManagerDeploymentName:     {healthStatus: &extendedHealth.Controller, minReady: 1},
 		resources.SchedulerDeploymentName:             {healthStatus: &extendedHealth.Scheduler, minReady: 1},
-		resources.MachineControllerDeploymentName:     {healthStatus: &extendedHealth.MachineController, minReady: 1},
```
moadqassem, line 46: Did you test this? I mean it could cause some issues to the UI but I am not quite sure tbh. As this being checked in the seed and moved to the user cluster controller manager, the status of the machine controller in the UI might be fiddled with.

moadqassem: Also I am not quite sure that this should really be moved to the user cluster controller manager. Why not simply check in initialMDController in the seed, the health check of the machien-controller-webhook, and then apply the initial MD, wouldn't this be more straightforward?

moadqassem: Here, https://github.com/kubermatic/kubermatic/blob/master/pkg/apis/kubermatic/v1/cluster.go#L962

mlavacca: It is not enough to check only the machineController and machineControllerWebhook deployment healths, we should check the existence of the mutatingWebhookConfiguration as well. It is deployed in the user cluster and not easily retrievable from that part of the seed-controller-manager. I moved the check to the user-cluster-controller-manager because here we have the certainty the configuration has been correctly created (the machineController health update means that all the resources have been correctly created). Otherwise, we could fall in having both the deployments created and healthy, but the configuration does not exist yet (and the validation won't be performed).

mlavacca: And we are following the same approach that we use for MLA and OPA: https://github.com/kubermatic/kubermatic/blob/3c750f28c78acbc80c16493536e86033a7038c6b/pkg/controller/user-cluster-controller-manager/resources/reconciler.go#L1083-L1117
if the controller deploys a resource, it's up to it to update the cluster health for what concerns that specific resource (and the MC is deployed by the user-cluster-controller-manager).

moadqassem: > It is not enough to check only the machineController and machineControllerWebhook deployment healths, we should check the existence of the mutatingWebhookConfiguration as well.

As I agree with you with this, I didn't only mean checking the health checks, what I meant is, you could actually achieve whatever you are trying to achieve in that controller as well. 

> It is deployed in the user cluster and not easily retrievable from that part of the seed-controller-manager. 

However, I don't agree with this point here, you can get whatever you like from that user cluster, using that controller as you can see over here https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/initialmachinedeployment/controller.go#L149. 

Here you have the user cluster client which you can use to get whatever resource in that user cluster. The difference between this, OPA and MLA is, those don't have a prior and dedicated controller so people simply throw it in there, but if you have a dedicated controller for what you are validating, just extend the health check over there instead of. moving a functionality from its right place.    

> if the controller deploys a resource, it's up to it to update the cluster health for what concerns that specific resource (and the MC is deployed by the user-cluster-controller-manager).

That's actually not true, MC is deployed via the seed controller manager not the user cluster controller manager: https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/kubernetes/resources.go#L298-L299

mlavacca: You are right, it is deployed by the seed controller manager. I'll move this check there

PR #9344, opened 2022-03-16
Add nodeCidrMaskSizeIPv4 & nodeCidrMaskSizeIPv6 to the networkConfig of the Cluster API

### LINKED ISSUE #9006: Allow configurable node CIDR size
At this point, KKP always uses the default `node-cidr-mask-size` (`/24` for IPv4 and `/64` for IPv6) and it is not possible to override it.

This can be an issue for IPv6 CIDRs, as `node-cidr-mask-size` must be equal or larger than the pods CIDR.

The aim of this task is to make the mask size used for node addressing configurable.

rastislavs:
Before this change, KKP always used the default `node-cidr-mask-size` (`24` for IPv4 and `64` for IPv6) and it was not possible to override it.

This was a limitation especially for IPv6 CIDRs, as the `node-cidr-mask-size` must be larger than the pods CIDR.

This PR allows optional configuration of `nodeCidrMaskSizeIPv4` / `nodeCidrMaskSizeIPv6` at the cluster level and propagates these settings to the kube-controller-manager and Cilium CNI.

Fixes #9006 

**Documentation**:

```release-note
Add nodeCidrMaskSizeIPv4 & nodeCidrMaskSizeIPv6 to the networkConfig of the Cluster API.
```

### DIFF
--- a/pkg/defaulting/cluster.go
+++ b/pkg/defaulting/cluster.go
@@ -124,54 +124,8 @@ func DefaultClusterSpec(ctx context.Context, spec *kubermaticv1.ClusterSpec, tem
 		spec.CNIPlugin.Version = cni.GetDefaultCNIPluginVersion(spec.CNIPlugin.Type)
 	}
 
-	if len(spec.ClusterNetwork.Services.CIDRBlocks) == 0 {
-		if spec.Cloud.Kubevirt != nil {
-			// KubeVirt cluster can be provisioned on top of k8s cluster created by KKP
-			// thus we have to avoid network collision
-			spec.ClusterNetwork.Services.CIDRBlocks = []string{resources.DefaultClusterServicesCIDRKubeVirt}
-		} else {
-			spec.ClusterNetwork.Services.CIDRBlocks = []string{resources.DefaultClusterServicesCIDR}
-		}
-	}
-
-	if len(spec.ClusterNetwork.Pods.CIDRBlocks) == 0 {
-		if spec.Cloud.Kubevirt != nil {
-			spec.ClusterNetwork.Pods.CIDRBlocks = []string{resources.DefaultClusterPodsCIDRKubeVirt}
-		} else {
-			spec.ClusterNetwork.Pods.CIDRBlocks = []string{resources.DefaultClusterPodsCIDR}
-		}
-	}
-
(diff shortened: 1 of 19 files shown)


pkg/defaulting/cluster.go
```diff
@@ -211,3 +173,53 @@ func DatacenterForClusterSpec(spec *kubermaticv1.ClusterSpec, seed *kubermaticv1
 
 	return nil, field.Invalid(field.NewPath("spec", "cloud", "dc"), datacenterName, "invalid datacenter name")
 }
+
+func defaultClusterNetwork(spec *kubermaticv1.ClusterSpec) {
```
PratikDeoghare, line 177: A: If both cidrs are specified, cluster would become dual stack by default. 

Should we set  `.DualStack = true` in here? Or A is false?

pkg/validation/cluster.go
```diff
@@ -296,6 +278,44 @@ func ValidateClusterNetworkConfig(n *kubermaticv1.ClusterNetworkingConfig, cni *
...
+			return field.Invalid(fldPath.Index(i), cidr, fmt.Sprintf("couldn't parse CIDR %q: %v", cidr, err))
+		}
+		// At this point, KKP only supports IPv4 as the primary CIDR and IPv6 as the secondary CIDR.
+		// The first provided CIDR has to be IPv4
+		if i == 0 && addr.To4() == nil {
+			return field.Invalid(fldPath.Child("pods", "cidrBlocks").Index(i), cidr,
+				fmt.Sprintf("invalid address family for primary CIDR %q: has to be IPv4", cidr))
+		}
+		// The second provided CIDR has to be IPv6
+		if i == 1 && addr.To4() != nil {
```
PratikDeoghare, line 294: did you mean == nil here?

PratikDeoghare (approved):

pkg/defaulting/cluster.go
rastislavs: We don't have a `DualStack` field in the cluster API. Are you suggesting that we should add it?

rastislavs: BTW right, if both cidrs are specified, the cluster becomes dual-stack implicitly.

PratikDeoghare: > We don't have a DualStack field in the cluster API. 

I thought it was there and we didn't set it. 

> Are you suggesting that we should add it? 

We are better without one more field to forget I guess.

rastislavs: ack :)

PR #9372, opened 2022-03-17
remove notion of cluster types from update manager

xrstf:
The concept of different cluster types has been removed at least in 2.20, if not much earlier (openshift support has been long, looooong abandoned for KKP). This PR removes the remnants from the UpdateManager and renames the "GetVersionsV2" to a more permanent name, seeing how we will probably not soon get rid of it.

I marked the "Type" field in our API as deprecated and removed all related code, so that even bogus values will not a validation error anymore.

```release-note
NONE
```


xrstf:
@zreigz PTAL

PR #9373, opened 2022-03-17
Add version label to control plane components

xrstf:
This PR is part of #5949. To handle gradual deployments of the control plane, a new update controller will smartly slowly roll out first the apiserver, then the rest of the control plane etc.

To keep the review simple, I split the branch into multiple PRs and this is the second one (after #9337). This PR adds a version label to the control plane components, so that the new update controller will be able to easily and reliably tell which version is currently active, without having to rely on parding Docker images.

```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #9374, opened 2022-03-17
run lint job when Go dependencies change

xrstf:
When updating dependencies, functions can become deprecated. This is something the lint job can catch, and so we should run it for go.mod/sum changes as well.

```release-note
NONE
```


kron4eg (approved):
/approve

PR #9375, opened 2022-03-17
Update clusters according to the version skew policy

### LINKED ISSUE #5949: Cluster version skew rule violation

xrstf:
### What does this PR do / Why do we need it:
This PR implements a new update controller that implements version-skew-policy-aware updating for userclusters. It can handle the `cluster.spec.version` jumping from 1.20 to 1.23 properly.

### The old update controller

KKP already had an `update` controller, which was responsible for watching Clusters and applying the auto-updates rules to bump the cluster's version accordingly. Likewise it would update MachineDeployments when an auto-node-update rule matches.

This controller still exists and does the exact same thing, but was renamed to `auto-update-controller`. Note that this controller does not care or know how to gradually roll out a version change. It could very well jump from 1.20 to 1.69 (right now it doesn't, but it could).

### The new update controller

The new `update-controller` implements the gradual cluster update. It uses the new version information in the ClusterStatus, the version from the cluster spec and gathers version info from the ReplicaSets in the cluster namespace. Based on these 3 pieces of information it makes decisions whether or not to update and which component to update. These decisions are then persisted in the ClusterStatus, which in turn will lead to reconciliations by other controllers.

In contrast the the auto-update controller, the new update controller does not care _why_ `.Spec.Version` was updated, it just focussed on making it happen.

Also importantly, all our reconciling coding (the creators that create the apiserver Deployment, the etcd StatefulSet, etc.) all simply rely on the new ClusterStatus. None of them have anything to do with update logic, they do not care why apiserver vX.Y is going to be deployed and they also do not need to dynamically query the control plane status. That makes the creators faster, easier to understand and easier to test. Creators should always be stateless.

### `Cluster.Spec.Version` now considered harmful

Since we now keep track of the effective current versions of the control plane components in the ClusterStatus, pretty much no code should ever use `.Spec.Version` if it needs to know the _current_ cluster version. `.Spec.Version` could be far, far in the future and might not even be reachable (because old old nodes or no update path being configured).

Instead, use `Cluster.Status.Versions.ControlPlane` to know what is actually running right now.

### The node-version controller

A new controller in the user-cluster-controller-manager is responsible for effective "caching" node information. It watches all nodes in the user cluster and stores the oldest version in `Cluster.Status.Versions.OldestNodeVersion`. This is both useful for speeding up the update controller (as it doesn't need to query the nodes every time a Cluster object changes) and is a nice trigger signal for when old nodes disappear from the cluster.

### Update Procedure

As per version skew policy, the new controller will always first ensure that the cluster namespace is completely healthy before doing anything. This is required to ensure that when for example etcd is updated (because the apiserver went from 1.21 to 1.22), we wait for etcd to roll out completely before progressing.

The controller then queries the cluster namespace and collects version information from the ReplicaSets and the `OldestNodeVersion` field. It now knows what is going on in the cluster and what is currently intended (`ClusterStatus.Versions`). It also knows the ultimate goal (`.Spec.Version`).

The current apiserver version is always mirrored in the `Cluster.Status.Versions.ControlPlane`, so once a new apiserver comes alive, this field is updated and in turn triggers a whooooole lot of other reconciliations by other controllers. So first `Versions.Apiserver` is updated, then we wait, once the apiservers are all rotated we bump `Versions.ControlPlane`.

Once the apiserver is updated, controller-manager and scheduler are updated. Then we wait again for them to roll out and for the cluster namespace to become healthy again.

Once all 3 control plane components are updated, the controller checks the `OldestNodeVersion`. If this version is more than 1 minor releases away from the current cluster version, no further updates will happen, as the nodes need to be rotated/replaced first. **This is outside of the responsibility of this controller.**

### Debuggability

Putting the version numbers in the ClusterStatus makes debugging much, much easier. It also allows for emergency interventions, where an KKP admin might choose to manually edit the ClusterStatus. It also made writing and testing the controller pretty straightforward, as no reconciling is involved directly.

To further improve the process, the new controller emits an event every time it changes something. It also makes use of a new ClusterCondition to inform about the current state (e.g. whether or not it's waiting right now and what it is waiting for). `kubectl describe`ing a cluster would look this this (in this example, I updated a 1.20 cluster to 1.22):

```
Status:
  Conditions:
    Update Progress:
      Kubermatic Version:   750cdd363e205ef0f655cb222cf58e49d3f7fbda
      Last Heartbeat Time:  2022-03-18T20:37:12Z
      Message:              No update in progress, cluster has reached its desired version.
      Reason:               UpToDate
      Status:               True
  Versions:
    Apiserver:            1.22.7
    Control Plane:        1.22.7
    Controller Manager:   1.22.7
    Oldest Node Version:  1.20.14
    Scheduler:            1.22.7
Events:
  Type     Reason                      Age                  From                        Message
  ----     ------                      ----                 ----                        -------
  Normal   ApiserverUpdated            2m6s                 kkp-update-controller       Kubernetes apiserver was updated to version 1.21.10.
  Normal   ControlPlaneVersionChanged  115s                 kkp-update-controller       Cluster control plane has reached version 1.21.10.
  Normal   ControllerManagerUpdated    115s                 kkp-update-controller       Kubernetes controller-manager was updated to version 1.21.10.
  Normal   SchedulerUpdated            115s                 kkp-update-controller       Kubernetes scheduler was updated to version 1.21.10.
  Normal   ApiserverUpdated            85s                  kkp-update-controller       Kubernetes apiserver was updated to version 1.22.7.
  Normal   ControlPlaneVersionChanged  64s                  kkp-update-controller       Cluster control plane has reached version 1.22.7.
  Normal   ControllerManagerUpdated    21s                  kkp-update-controller       Kubernetes controller-manager was updated to version 1.22.7.
  Normal   SchedulerUpdated            21s                  kkp-update-controller       Kubernetes scheduler was updated to version 1.22.7.
```

(Note that in the example above, the controller would refuse to update to 1.23.x, as there are 1.20 nodes in the cluster).

If one follows the individual steps of this cluster upgrade, it would look like this:

```diff
--- (none)
+++ Cluster cxl7d75cwb v2289 (2022-03-18T21:30:54+01:00) (gen. 1)
@@ -0 +1,2 @@
+spec:
+  version: 1.20.14        <- cluster was created

--- Cluster cxl7d75cwb v2289 (2022-03-18T21:30:54+01:00) (gen. 1)
+++ Cluster cxl7d75cwb v2292 (2022-03-18T21:30:54+01:00) (gen. 1)
@@ -1,2 +1,8 @@
 spec:
   version: 1.20.14
+status:                   <- initial version gets set
+  versions:
+    apiserver: 1.20.14
+    controlPlane: 1.20.14
+    controllerManager: 1.20.14
+    scheduler: 1.20.14

--- Cluster cxl7d75cwb v3676 (2022-03-18T21:32:27+01:00) (gen. 9)
+++ Cluster cxl7d75cwb v4277 (2022-03-18T21:34:53+01:00) (gen. 9)
@@ -5,4 +5,5 @@
     apiserver: 1.20.14
     controlPlane: 1.20.14
     controllerManager: 1.20.14
+    oldestNodeVersion: 1.20.14           <- the first nodes appear a few minutes later
     scheduler: 1.20.14

--- Cluster cxl7d75cwb v4277 (2022-03-18T21:34:53+01:00) (gen. 9)
+++ Cluster cxl7d75cwb v4386 (2022-03-18T21:35:20+01:00) (gen. 10)
@@ -1,5 +1,5 @@
 spec:
-  version: 1.20.14
+  version: 1.22.7                   <- I triggered the update
 status:
   versions:
     apiserver: 1.20.14

--- Cluster cxl7d75cwb v4386 (2022-03-18T21:35:20+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v4387 (2022-03-18T21:35:20+01:00) (gen. 10)
@@ -2,7 +2,7 @@
   version: 1.22.7
 status:
   versions:
-    apiserver: 1.20.14
+    apiserver: 1.21.10               <- time to go to the next apiserver
     controlPlane: 1.20.14
     controllerManager: 1.20.14
     oldestNodeVersion: 1.20.14

--- Cluster cxl7d75cwb v4589 (2022-03-18T21:35:31+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v4590 (2022-03-18T21:35:31+01:00) (gen. 10)
@@ -3,7 +3,7 @@
 status:
   versions:
     apiserver: 1.21.10
-    controlPlane: 1.20.14
+    controlPlane: 1.21.10               <- new apiserver is alive, no 1.20 apiservers remain
     controllerManager: 1.20.14
     oldestNodeVersion: 1.20.14
     scheduler: 1.20.14

--- Cluster cxl7d75cwb v4592 (2022-03-18T21:35:31+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v4593 (2022-03-18T21:35:31+01:00) (gen. 10)
@@ -4,6 +4,6 @@
   versions:
     apiserver: 1.21.10
     controlPlane: 1.21.10
-    controllerManager: 1.20.14
+    controllerManager: 1.21.10           <- make ctlr-mgr follow
     oldestNodeVersion: 1.20.14
     scheduler: 1.20.14

--- Cluster cxl7d75cwb v4595 (2022-03-18T21:35:31+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v4596 (2022-03-18T21:35:31+01:00) (gen. 10)
@@ -6,4 +6,4 @@
     controlPlane: 1.21.10
     controllerManager: 1.21.10
     oldestNodeVersion: 1.20.14
-    scheduler: 1.20.14
+    scheduler: 1.21.10                   <- make scheduler follow

--- Cluster cxl7d75cwb v4872 (2022-03-18T21:36:01+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v4873 (2022-03-18T21:36:01+01:00) (gen. 10)
@@ -2,7 +2,7 @@
   version: 1.22.7
 status:
   versions:
-    apiserver: 1.21.10
+    apiserver: 1.22.7              <- final update step
     controlPlane: 1.21.10
     controllerManager: 1.21.10
     oldestNodeVersion: 1.20.14

--- Cluster cxl7d75cwb v4920 (2022-03-18T21:36:02+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v5147 (2022-03-18T21:36:22+01:00) (gen. 10)
@@ -3,7 +3,7 @@
 status:
   versions:
     apiserver: 1.22.7
-    controlPlane: 1.21.10
+    controlPlane: 1.22.7              <- new apiserves came online
     controllerManager: 1.21.10
     oldestNodeVersion: 1.20.14
     scheduler: 1.21.10

--- Cluster cxl7d75cwb v5527 (2022-03-18T21:37:05+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v5528 (2022-03-18T21:37:05+01:00) (gen. 10)
@@ -4,6 +4,6 @@
   versions:
     apiserver: 1.22.7
     controlPlane: 1.22.7
-    controllerManager: 1.21.10
+    controllerManager: 1.22.7               <- make ctrl-mgr follow (a minute later, because since the last update, etcd was updated as well!)
     oldestNodeVersion: 1.20.14
     scheduler: 1.21.10

--- Cluster cxl7d75cwb v5528 (2022-03-18T21:37:05+01:00) (gen. 10)
+++ Cluster cxl7d75cwb v5530 (2022-03-18T21:37:05+01:00) (gen. 10)
@@ -6,4 +6,4 @@
     controlPlane: 1.22.7
     controllerManager: 1.22.7
     oldestNodeVersion: 1.20.14
-    scheduler: 1.21.10
+    scheduler: 1.22.7                  <- finally, scheduler is updated
```

Fixes #5949

```release-note
Kubermatic now updates clusters according to the Kubernetes version skew policy.
```


pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -292,6 +292,8 @@ const (
 
 	ClusterConditionEtcdClusterInitialized ClusterConditionType = "EtcdClusterInitialized"
 
+	ClusterConditionUpdateProgress ClusterConditionType = "UpdateProgress"
```
xmudrii, line 295: Why don't we call this `UpdateInProgress` (if that's correct at all)?

pkg/controller/user-cluster-controller-manager/node-version-controller/doc.go
```diff
@@ -0,0 +1,21 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+Package nodeversioncontroller a controller that watches Nodes and
```
xmudrii: Nit:

```suggestion
Package nodeversioncontroller contains a controller that watches Nodes and
```

pkg/apis/kubermatic/v1/cluster.go
xrstf: Because the status can be "no update currently in progress", so the condition should not be called "UpdateInProgress". Maybe "UpdateStatus" is clearer?

xmudrii: In that case, UpgradeProgress is fine.

xmudrii (approved):
/lgtm
/approve

xrstf:
/hold

prioritize other PRs to get merged first

PR #9384, opened 2022-03-21
Remove support for Kubernetes 1.20

xrstf:
Kubernetes 1.20 has gone EOL on February 28th, 2022. Time to remove it. This PR removes support for it, but configures an update rule to allow auto-updating all clusters to 1.21. To not run into errors when reconciling old clusters, most of the reconciling code can still handle 1.20 (e.g. the kubernetes-dashboard, etcd, ...). -- Those code places can be removed in a follow up KKP release when we're sure no 1.20 cluster exists anymore.

```release-note
Remove support for Kubernetes 1.20.
```


addons/cluster-autoscaler/cluster-autoscaler.yaml
```diff
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 {{ $version := "UNSUPPORTED" }}
-{{ if eq .Cluster.MajorMinorVersion "1.20" }}
-{{ $version = "v1.20.0"}}
-{{ end }}
```
xmudrii: I think we discussed it before, but it might be worth bringing it up again. Isn't this going to break existing 1.20 clusters until they're not upgraded to v1.21? Upgrading to v1.21 might take some time and you can also limit how many clusters will be upgraded at the same time (if I'm not mistaken). You can also disable auto-upgrades by overriding the default config.

Shouldn't we leave this for one more version just to be sure we're not going to break something?

xrstf: Yeah good point. I had the same thought for all the Go code, but treated addons differently.

xrstf: Fixed. Restored addons to their previous state.

xmudrii:
@xrstf I think we should also consider changing the default Kubernetes version to v1.22.

xrstf:
@xmudrii I was thinking "always use the most recent supported version, minus 1", that's why i kep the default. No strong feelings either way. Convince me! 😁

xmudrii:
> I was thinking "always use the most recent supported version, minus 1", that's why i kep the default. No strong feelings either way. Convince me! 😁

I've always understood that rule as the latest Kubernetes version, minus 1, not the latest supported in KKP (as long as n-1 is supported by KKP). Considering that 1.21 is entering the maintenance mode at the end of April, and reaching EOL at the end of June, I think we should start promoting/defaulting to 1.22.

xrstf:
While rebasing I already changed the default in this PR :grin:

pkg/util/kubectl/kubectl.go
```diff
@@ -33,11 +33,11 @@ func BinaryForClusterVersion(version *semver.Version) (string, error) {
 
 	switch version.Minor() {
 	case 20:
-		binary = "kubectl-1.20"
+		binary = "kubectl-1.21"
 	case 21:
-		binary = "kubectl-1.22"
+		binary = "kubectl-1.23"
```
xmudrii: The 1.21 clusters should use `kubectl-1.21`.

pkg/resources/apiserver/deployment.go
```diff
@@ -316,6 +316,7 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		"--requestheader-extra-headers-prefix", "X-Remote-Extra-",
 		"--requestheader-group-headers", "X-Remote-Group",
 		"--requestheader-username-headers", "X-Remote-User",
+		"--endpoint-reconciler-type", "none",
```
xmudrii: The problem with this change is the same as we already discussed for addons. Let's say users disable auto-upgrades, so they can better plan upgrades. The existing 1.20 clusters will be reconciled with this flag, which might break those clusters. If we want to give some more time just to be sure, we should probably revert this change.

pkg/resources/controllermanager/deployment.go
```diff
@@ -242,14 +242,7 @@ func getFlags(data *resources.TemplateData, version *semver.Version) ([]string,
...
 
-	featureGates := []string{"RotateKubeletServerCertificate=true"}
-
-	// starting with k8s 1.21, this is always true and cannot be toggled anymore
-	if version.LessThan(semver.MustParse("1.21.0")) {
-		featureGates = append(featureGates, "RotateKubeletClientCertificate=true")
-	}
-
-	featureGates = append(featureGates, data.GetCSIMigrationFeatureGates()...)
+	featureGates := append([]string{"RotateKubeletServerCertificate=true"}, data.GetCSIMigrationFeatureGates()...)
```
xmudrii: Same issue as for the API server flag.

pkg/resources/apiserver/deployment.go
xrstf: > users disable auto-upgrades

With users you mean KKP admins, right? Or can endusers simply reject our auto update rules?

pkg/util/kubectl/kubectl.go
xrstf: How do you spot these typos? Is your mother an eagle? Mine is a mole.

xmudrii (approved):
/lgtm
/approve

PR #9387, opened 2022-03-21
Fix buildah not using gocache

xmudrii:

It turns out that GOCACHE is not working if we build the user-ssh-keys-agent binary from the `cmd/user-ssh-keys-agent directory` instead of the root directory. I've tried investigating this, but I couldn't find the root cause. I've been able to reproduce the issue with buildah, Docker, and locally.

This PR modifies the multiarch Dockerfile for `user-ssh-keys-agent` to invoke the build from the root directory. With this change, the time needed to build the multiarch Docker images is 2-3 minutes instead of 15 minutes.

```release-note
NONE
```


xrstf (approved):
/lgtm

PR #9388, opened 2022-03-21
fix user cluster owner

### LINKED ISSUE #9380: Fix user cluster owner when you create cluster from template
Right now the user cluster template controller fetches templates and creates instances. The template contains the user name who created the template but not the user cluster owner.

Extend cluster template instance for the cluster owner. Adapt this change to the cluster template controller.

zreigz:
**What does this PR do / Why do we need it**: fix user cluster owner. The template contains the user name who created the template but not the user cluster owner. This PR extends the cluster template instance for the cluster owner.

Fixes #9380

```release-note
Fix user cluster owner when you create a cluster from the template
```


xrstf (approved):
/approve

Issue #8995, opened 2022-02-14
Migrate installer away from urfave/cli


xrstf:
### What would you like to be added? / User Story
urfave/cli's parsing rules are so incredibly strict and got even worse in v2, that any application built on it is so hard to use, it becomes pointless. Users would need to specifically know in what order to put global flags, commands, arguments and command flags. Tell me, should this work?

```
./installer deploy kubermatic-master --config FOO
```

It should. But urfave **silently** ignores `--config` because an argument ("kubermatic-master") was given. And then the v2 docs say "every flag is now a global flag!" and you try it with `--verbose` and nope, `./installer deploy --verbose` **does not work**.

See #8991 for more on this.

### Acceptance criteria
- [ ] We do not use urfave/cli anymore.

PR #9394, opened 2022-03-21
rewrite storeuploader to use cobra

xrstf:
We want to move away from urfave/cli due to its super weird un unpredicatble and error prone parsing of CLI flags. This PR is a step towards that goal, refactoring the S3 store uploader to use Cobra instead.

I chose not to use viper to get the env variables, because viper and cobra together are weird and do not work the way one would expect when using `..Var` functions (see https://github.com/spf13/viper/issues/461). It's much clearer to just use `os.Getenv` and be done with it.

This is part of #8995.

```release-note
s3-storeuploader uses Cobra instead of urfave/cli, but the CLI flags are identical, so no changes to backup containers should be necessary.
```

### DIFF
--- a/cmd/s3-storeuploader/main.go
+++ b/cmd/s3-storeuploader/main.go
@@ -17,12 +17,11 @@ limitations under the License.
 package main
 
 import (
-	"context"
 	"crypto/x509"
 	"fmt"
 	"os"
 
-	"github.com/urfave/cli"
+	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 
 	"k8c.io/kubermatic/v2/pkg/log"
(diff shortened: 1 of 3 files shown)


cmd/s3-storeuploader/main.go
```diff
@@ -17,175 +17,117 @@ limitations under the License.
...
-		return nil
-	}
-
-	defer func() {
-		if logger != nil {
-			if err := logger.Sync(); err != nil {
-				fmt.Println(err)
-			}
-		}
-	}()
```
xmudrii, line 173: I don't see that we do this any longer. Just to confirm, this is not required?

xmudrii (approved):

PR #9395, opened 2022-03-21
adding configurable metering reports to seed

### LINKED ISSUE #9181: Introduce MeteringReportConfiguration list to the Seed CRD
```go
type SeedSpec struct {
  Country string `json:"country,omitempty"`
	...
  // Deprecated: Metering configures the metering tool on user clusters across the seed.  
  Metering  *MeteringConfiguration  `json:"metering,omitempty"`
  // New field
// +kubebuilder:default:={{name: kubermatic-metering-report-weekly}}
	ReportConfigurations map[string]*MeteringReportConfiguration `json:"reports,omitempty"`
}

type MeteringReportConfiguration struct {
	// Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron. Please take a note that Schedule is responsible
	// only for setting the time when a report generation mechanism kicks off. The Interval MUST be set independently.
	// +kubebuilder:default:=`0 1 * * 6`
	Schedule string `json:"schedule,omitempty"`
	// Interval defines the number of days consulted in the metering report.
	// +kubebuilder:default=7
	// +kubebuilder:validation:Minimum:=1
}
```

pkprzekwas:
Adding option for defining more then one reporting cron job per seed. 

Fixes #9181 

This commit will break existing seed controller. `Interval` and `Schedule` fields of seed's `MeteringConfiguration` fields have been nested in additional `ReportingConfiguration` struct breaking compatibility. After installing CRD existing seeds have to be manually cleaned from `interval` and `schedule`. I don't know if there is a better way to handle this update (any ideas @xrstf ?)

```release-note
NONE
```

Signed-off-by: Patryk Przekwas [patryk@kubermatic.com](mailto:patryk@kubermatic.com)


xrstf:
Please give your PRs proper titles, as they will be permanent commits in our history. Or did you create 9891 metering reports? Why?

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -538,13 +538,24 @@ type MeteringConfiguration struct {
...
 	// Schedule in Cron format, see https://en.wikipedia.org/wiki/Cron. Please take a note that Schedule is responsible
 	// only for setting the time when a report generation mechanism kicks off. The Interval MUST be set independently.
-	// +kubebuilder:default:=`0 1 * * 6`
 	Schedule string `json:"schedule,omitempty"`
-	// Interval defines the number of days consulted in the metering report.
+
 	// +kubebuilder:default=7
 	// +kubebuilder:validation:Minimum:=1
+
+	// Interval defines the number of days consulted in the metering report.
```
xrstf, line 558: Can't do this. We have committed to supporting the old structure, now we have to support it. You cannot just remove fields from our API. You have to deal with the old config.

xrstf: cc @kron4eg

pkprzekwas: I thought this is fine as long as the code was not officially released.

pkprzekwas: (Those fields landed on master about a week ago)

pkprzekwas:
> Please give your PRs proper titles, as they will be permanent commits in our history. Or did you create 9891 metering reports? Why?

Sorry, I've improved it a bit.

pkg/apis/kubermatic/v1/datacenter.go
xrstf: Ah that's a good point, I thought we added them in 2.20 already. If they are unreleased yet, it's fine :)

pkg/controller/operator/seed/reconciler_ee_test.go
```diff
@@ -0,0 +1,167 @@
+//go:build ee
```
xrstf, line 1: The code for EE _has_ to live in pkg/ee. In here, you can only have a thin wrapper, though it's an intersting case where you're actually kind of just testing the reconciler, but for the EE creators.. Hm...

pkprzekwas: Added wrapper although I have mixed feeling if that should be the way to handle this particular case. Let me know what you think.

xrstf (approved):
/approve

pkg/ee/metering/reconcile.go
```diff
@@ -91,17 +98,67 @@ func ReconcileMeteringResources(ctx context.Context, client ctrlruntimeclient.Cl
...
+func cleanupAllMeteringResources(ctx context.Context, client ctrlruntimeclient.Client, meteringConfig *kubermaticv1.MeteringConfiguration) error {
 	key := types.NamespacedName{Namespace: resources.KubermaticNamespace, Name: meteringToolName}
 	if err := cleanupResource(ctx, client, key, &appsv1.Deployment{}); err != nil {
 		return fmt.Errorf("failed to cleanup metering Deployment: %w", err)
 	}
 
-	key.Name = meteringCronJobWeeklyName
-	if err := cleanupResource(ctx, client, key, &batchv1beta1.CronJob{}); err != nil {
-		return fmt.Errorf("failed to cleanup metering CronJob: %w", err)
+	if meteringConfig != nil {
```
lsviben: this wont remove all cronjobs if the metering is disabled by setting `meteringConfig == nil` , which is one of the possible conditions for a cleanup. We could get orphaned cronjobs, its better to list all metering cronjobs and delete them

pkg/ee/metering/reconcile.go
```diff
@@ -75,11 +81,12 @@ func ReconcileMeteringResources(ctx context.Context, client ctrlruntimeclient.Cl
 
 	modifiers := []reconciling.ObjectModifier{
 		common.VolumeRevisionLabelsModifierFactory(ctx, client),
+		common.OwnershipModifierFactory(seed, scheme),
+		common.LabelModifierFactory(map[string]string{RelatedToLabel: seed.Name}, false),
```
lsviben: I dont think you need to do this. 

In a given seed cluster, there will be only one Seed for which the operator reconcile will create the metering cronjobs, so there is no reason to label them based on seed. You can just slap a `kubermatic-metering: <metering-report-name>` label on all of them in the cronJobCreator and list them through that.

pkg/controller/operator/seed/reconciler_ee_test.go
lsviben: I mean you could make a test instead in `/ee/metering/reconcile_test.go` and test there without the whole reconcile, just using `ReconcileMeteringResources(...)` , just that it will take a bit more work.

I dont have strong feelings about this, your solution works as well.

pkprzekwas: I see your point, but as you said it would take some additional effort to implement the test. If you don't have a strong feelings about it I would stick to the way it it implemented to safe some time.

pkg/ee/metering/reconcile.go
pkprzekwas: changed to iterate over the list of existing cron jobs as suggested.

pkg/ee/metering/reconcile.go
pkprzekwas: Added `kubermatic-metering: <metering-report-name>` label through label modifier. I did that as the function was already created and `cronJobCreator` function doesn't mutate object meta at the moment (not sure if it brings any value). If your feel like adding label in the `cronJobCreator` function would simplify the code, I can quickly fix it.

lsviben: For me it seems like just putting it into cronJobCreator is much simpler then having a modifier func

pkprzekwas: Ok, I will add it there

lsviben (commented):
/approve
/lgtm

PR #9398, opened 2022-03-21
rewrite installer to use cobra

### LINKED ISSUE #8995: Migrate installer away from urfave/cli

xrstf:
This convert the installer to use cobra. I kept it simple and did not add things like auto-completion, dedicated validation stuff etc.

Fixes #8995.

```release-note
The kubermatic-installer uses Cobra instead of urfave/cli, but the CLI flags are identical, so no changes to scripts or automation should be necessary.
```

### DIFF
--- a/cmd/kubermatic-installer/main.go
+++ b/cmd/kubermatic-installer/main.go
@@ -21,25 +21,24 @@ import (
 	"os"
 	"time"
 
-	"github.com/urfave/cli"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 
 	"k8c.io/kubermatic/v2/pkg/log"
 	kubermaticversion "k8c.io/kubermatic/v2/pkg/version/kubermatic"
 )
 
-var (
-	verboseFlag = cli.BoolFlag{
-		Name:  "verbose, v",
-		Usage: "enable more verbose output",
-	}
+type Options struct {
+	Verbose         bool
+	ChartsDirectory string
+}
 
-	chartsDirectoryFlag = cli.StringFlag{
-		Name:   "charts-directory",
-		Value:  "charts",
-		Usage:  "filesystem path to the Kubermatic Helm charts",
-		EnvVar: "KUBERMATIC_CHARTS_DIRECTORY",
-	}
-)
+func (o *Options) CopyInto(other *Options) {
+	other.ChartsDirectory = o.ChartsDirectory
+	other.Verbose = o.Verbose
+}
+
+var options = &Options{}
 
 func main() {
 	rand.Seed(time.Now().UnixNano())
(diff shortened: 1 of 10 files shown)


cmd/kubermatic-installer/main.go
```diff
@@ -21,39 +21,70 @@ import (
...
-	app.Commands = commands(logger, versions)
+	rootCmd := &cobra.Command{
+		Use:           "kubermatic-installer",
+		Short:         "Installs and updates Kubermatic Kubernetes Platform",
+		Version:       versions.Kubermatic,
+		SilenceErrors: true,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			if env := os.Getenv("KUBERMATIC_CHARTS_DIRECTORY"); env != "" {
+				options.ChartsDirectory = env
+			}
```
xmudrii, line 57: Isn't this going to override the `--charts-directory` flag? I think the flag should have precedence over the environment variable. Unless it worked the same way before, in that case, we should leave it as it is.

cmd/kubermatic-installer/main.go
```diff
@@ -21,39 +21,70 @@ import (
...
+	// so would always show the Usage, no matter what error occurred. Tow ork around this, we
+	// set SilenceUsages on all commands and manually print the error using the FlagErrorFunc.
+	rootCmd.SetFlagErrorFunc(func(c *cobra.Command, err error) error {
+		if err := c.Usage(); err != nil {
+			return err
+		}
+
+		// ensure we exit with code 1 later on
+		return err
+	})
```
xmudrii, line 80: Have you tried using `SilenceUsage`? I think it might accomplish something similar, but slightly more elegant.

cmd/kubermatic-installer/main.go
```diff
@@ -21,39 +21,70 @@ import (
...
-		Usage:  "filesystem path to the Kubermatic Helm charts",
-		EnvVar: "KUBERMATIC_CHARTS_DIRECTORY",
-	}
-)
+func (o *Options) CopyInto(other *Options) {
+	other.ChartsDirectory = o.ChartsDirectory
+	other.Verbose = o.Verbose
+}
+
+var options = &Options{}
```
xmudrii, line 41: Why don't we put default values here out of the box?

cmd/kubermatic-installer/main.go
xrstf: SilenceUsage is set everywhere for all commands already, but in that case, cobra _never_ shows an usage, so we need this helper to bring back the usage help on actual FlagErrors.

cmd/kubermatic-installer/main.go
xrstf: verbose's default is false, so no need to mention it.

The default for the charts-directory flag is "charts", but if we put this in the flag, we cannot distinguish between "flag not set" and "flag set to 'charts'", so I kept the default out and only set it dynamically later in the PreRun function.

cmd/kubermatic-installer/main.go
xrstf: I would not know how to write the code differently. We want to have the flag default to "charts", but the env must take precedence, as we use it in our Docker image to override the default location (https://github.com/kubermatic/kubermatic/blob/master/Dockerfile#L18).

xmudrii (approved):
/lgtm
/approve

PR #9411, opened 2022-03-23
WIP added validation webhook for project owner

### LINKED ISSUE #9367: Add project ownerRef validation to webhooks
To achieve a kubernetes native API, we need to make sure that when creating projects not through the KKP API, the same rules apply. 

The KKP API automatically sets the User OwnerReference to the Project based on which user(or KKP serviceAccount) made the request. This is essential for creating the RBAC rules for Users further along the way. So to make sure users create proper Projects, we need to validate them and check if they have the proper OwnerReference set.

lsviben:
Adds a validation webhook for Projects, where it checks that Projects have a proper User OwnerReference. Without it, the UserProjectBindings are not created, and the whole rbac magic that KKP does gets messed up.

Fixes #9367 

```release-note
Added a validation webhook for Projects which makes sure that Projects have a valid User OwnerReference.
```

### DIFF
--- a/cmd/kubermatic-webhook/main.go
+++ b/cmd/kubermatic-webhook/main.go
@@ -37,6 +37,7 @@ import (
 	mlaadminsettingmutation "k8c.io/kubermatic/v2/pkg/webhook/mlaadminsetting/mutation"
 	oscvalidation "k8c.io/kubermatic/v2/pkg/webhook/operatingsystemmanager/operatingsystemconfig/validation"
 	ospvalidation "k8c.io/kubermatic/v2/pkg/webhook/operatingsystemmanager/operatingsystemprofile/validation"
+	projectvalidation "k8c.io/kubermatic/v2/pkg/webhook/project/validation"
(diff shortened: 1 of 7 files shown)


xrstf:
Before checking for owner refs, can we instead start refactoring the Project resource? Owner refs are impossible to set via kubectl (at least not nicely) and we should probably move to a different model, like our UserSSHKeys, which simply have a list of cluster IDs in their spec.

A controller/webhook should then later fill in the owner refs / keep them in-sync with the `spec.owners` list of user IDs.

lsviben:
> Before checking for owner refs, can we instead start refactoring the Project resource? Owner refs are impossible to set via kubectl (at least not nicely) and we should probably move to a different model, like our UserSSHKeys, which simply have a list of cluster IDs in their spec.
> 
> A controller/webhook should then later fill in the owner refs / keep them in-sync with the `spec.owners` list of user IDs.

Eh you are right. I thought to do this as some next step, but maybe we can do it first, to not have this in-the-middle solution. Ill close this for now and dig it out when we need the webhook for projects

PR #9414, opened 2022-03-23
Add phases to Cluster objects

xrstf:
Inspired by the new update controller, this PR introduces cluster phases, similar to how projects already have phases. This PR adds a new field to the ClusterStatus (`Phase`) and defines a couple of possible phases:

- `Terminating` if the deletion timestamp is not nil.
- `Creating` if the `ClusterInitialized` condition wasn't set yet.
- `Updating` if an update status indicates that the control plane is being updated.
- `Running` else.

I chose not to take the cluster health into account, as I fear we would be constantly flip-flopping between Running and Reconciling.

```release-note
Clusters now have a phase (creating, updating, running, terminating) to allow getting a quick overview over the health on a seed cluster.
```


kron4eg (approved):
/approve

PR #9416, opened 2022-03-23
Do not reference GOPATH or go get in dev environment instructions

### LINKED ISSUE #9415: Unable to create dev env.
### What happened?
The command `go get k8c.io/kubermatic` is throwing an error: `go: downloading k8c.io/kubermatic v8.88.1+incompatible`

### What should happen?
The command shouldn't throw any error and create a dev env.

### How to reproduce?
Run the command `go get k8c.io/kubermatic` in your terminal.

### Environment

- OS: Windows + WSL(ubuntu)
- KKP Version:
- Master/seed cluster provider:

embik:
It's 2022, let's not recommend downloading kubermatic source code into `GOPATH` or via `go get` anymore. It's not needed and `go get` doesn't even seem to work that way anymore.

Fixes #9415

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Thanks for taking the time to join our community and start contributing!
 ### Development environment
 
 ```bash
-git clone git@github.com:kubermatic/kubermatic
+git clone git@github.com:kubermatic/kubermatic.git
 cd kubermatic
 ```
 
--- a/README.md
+++ b/README.md
@@ -73,16 +73,8 @@ Thanks for taking the time to join our community and start contributing!
 ### Development environment
 
 ```bash
-mkdir -p $(go env GOPATH)/src/k8c.io
-cd $(go env GOPATH)/src/k8c.io
 git clone git@github.com:kubermatic/kubermatic
 cd kubermatic
-```
-
-Or alternatively:
-
-```bash
-go get k8c.io/kubermatic
 ```
 
 There are a couple of scripts in the `hacks` directory to aid in running the components locally


jayesh-srivastava:
@embik I guess the command `git clone git@github.com:kubermatic/kubermatic` isn't working too. That is showing this error.

```
Cloning into 'kubermatic'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

embik:
Hey @jayesh-srivastava, the development instructions assume familiarity with git and GitHub to a certain extend. That looks like you've never provided GitHub with an SSH public key. You can use any method to clone the repo since it's public, the setup steps are more guard rails and less step-by-step instructions.

README.md
```diff
@@ -73,18 +73,10 @@ Thanks for taking the time to join our community and start contributing!
 ### Development environment
 
 ```bash
-mkdir -p $(go env GOPATH)/src/k8c.io
-cd $(go env GOPATH)/src/k8c.io
 git clone git@github.com:kubermatic/kubermatic
```
xrstf: General question: Git suggests git@github.com:kubermatic/kubermatic.git as the URL (note the .git extension). I know it's not needed, but I always specify it when using SSH URLs. Without them, they look .... naked to me. :grin:

jayesh-srivastava:
Hey @embik I have a certain level of familiarity with Git and GitHub and I have provided GitHub with an SSH public key. Still, this issue was coming. I pinged you just to confirm if this was only on my system or not?

xrstf:
I tried to reproduce your issue, but cannot do so. Without a SSH key loaded in my agent, the clone fails. The moment a key is present, it works. Have not tried adding a key that is unknown to Github, though. Only my regular SSH key.

README.md
embik: Yeah I'm fine with adding that, it did look weird to me as well and I had to double-check if it was actually a valid clone URL.

embik:
> Hey @embik I have a certain level of familiarity with Git and GitHub and I have provided GitHub with an SSH public key. Still, this issue was coming. I pinged you just to confirm if this was only on my system or not?

That's strange then, but as @xrstf said I cannot reproduce it either. Can you try cloning any other repository via the SSH url (not the HTTPS one), e.g. Kubernetes' `kubectl` repository via `git clone git@github.com:kubernetes/kubectl.git`? I assume the issue is not specific to kubermatic repositories.

embik:
> > It's 2022
> 
> I think this is more a nice side effect from migrating to our new CRDs, because now the code generation stuff that we're using finally works outside of GOPATH (I hope, never tried it myself).

It kind of worked (or could have worked) before the new CRDs too, the script just did not properly consider some file locations changing when moving things around, but now it seems to work just fine. I never had `kubermatic/kubermatic` in my GOPATH since the beginning.

xrstf (approved):
/approve
/hold

unhold if and when the cloning issue is resolved

ahmedwaleedmalik:
Maybe we can just use the HTTPS URL instead? @embik @xrstf 

`git clone https://github.com/kubermatic/kubermatic.git`

embik:
> Maybe we can just use the HTTPS URL instead in the docs? @embik @xrstf
> 
> `git clone https://github.com/kubermatic/kubermatic.git`

I personally prefer cloning via SSH, and the GitHub UI does offer alternatives to clone a repo, including the HTTPS URL.

@jayesh-srivastava any updates from your side? Have you been able to make it work?

xrstf:
> I personally prefer cloning via SSH

Same here. Many years of badly setup Windows machines with borked Github Auth Helpers and things just not working because of credentials made me always just use SSH URLs, even though it is a bit slower. ^^

ahmedwaleedmalik:
> I personally prefer cloning via SSH

My personal preference would be the same but from my perspective, it's more about ease of use. Especially since it won't have any pre-requisites like setting up SSH keys etc.

embik:
> > I personally prefer cloning via SSH
> 
> My personal preference would be the same but from my perspective, it's more about ease of use. Especially since it won't have any pre-requisites like setting up SSH keys etc.

Counter argument: If you clone a repository, you likely end up pushing to a fork at some point. HTTPS pushing is cumbersome because you either need to put in your credentials every time or you need to correctly configure a helper (and as @xrstf, those tend to be buggy). An SSH key "just works" as long as you pass it correctly.

ahmedwaleedmalik:
> IIRC, if you have 2-factor auth enabled, you cannot push with HTTPS. So my preference is SSH

You have to use (PAT)[https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token] in that case and it works.

Again, my point was that it's not really about "personal preference" but rather ease of use. Since HTTPS is more accessible than SSH in general, SSH also requires a pre-requisite to "clone" the repo. If someone knows GIT and is cloning via HTTPS I'd assume that he knows how to push changes back in, at least should know his password :D 

Anyways, I don't really think we should be spending our energies on this. 

```bash
# To clone using HTTPS instead of SSH, run `git clone https://github.com/kubermatic/kubermatic.git`
git clone git@github.com:kubermatic/kubermatic.git
cd kubermatic
```

embik:
> Anyways, I don't really think we should be spending our energies on this.

I have to agree with this one, while this is my favourite bikeshedding thread right now, I think that there's not _that_ big of a difference between clone methods. Both are somewhat ergonomic for developers.

Unholding this PR so it gets merged as we haven't heard back for a while now. If there are any further problems around cloning the repo, feel free to reach out.

/hold cancel

PR #9420, opened 2022-03-23
fix creating vsphere clusters

xrstf:
Since introducing the ClusterVersionsStatus, we switched to using it everywhere. However this function that is fixed in this PR is also called during cluster _creation_, where under no circumstances the ClusterVersionsStatus exist. So in those cases, we have to rely on the spec.

```release-note
NONE
```


moadqassem (approved):
/approve 
/lgtm

PR #9421, opened 2022-03-23
Add usersshkey.Spec.Project to allow easier GitOps management

xrstf:
To make it easier to apply UserSSHKeys with `kubectl`, using OwnerReferences is not the best option. This PR adds a new field to UserSSHKeys and uses that for all the logic. The existing mutation webhook ensures the owner ref so that deletion works just as before.

The kubermatic installer was extended to auto-migrate all UserSSHKeys on the master cluster. From there the sync controller will try to mirror that onto seeds, but fail because the seeds might not have the new CRDs (but also not the new webhook, which would reject keys without project). Once the seeds are updated, the mirroring should continue as normal.

I tested the procedure in kind:

* create empty kind cluster
* install old UserSSHKey CRD
* create UserSSHKey (without project at this point)
* update UserSSHKey CRD
* edit the existing key to fill in the project

So I think it's fine to "break the API" by adding a new required field.

This is part of #9439.

```release-note
UserSSHKey.Spec.Project was added, making it easier to manage SSH keys declaratively. Existing UserSSHKey objects must be migrated, the kubermatic-installer takes care of that during the upgrade.
```

### DIFF
--- a/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_usersshkeys.yaml
+++ b/charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_usersshkeys.yaml
@@ -22,6 +22,9 @@ spec:
     - jsonPath: .spec.owner
       name: Owner
       type: string
+    - jsonPath: .spec.project
+      name: Project
+      type: string
     - jsonPath: .spec.fingerprint
       name: Fingerprint
       type: string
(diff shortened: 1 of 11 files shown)


xrstf:
cc @lsviben, just FYI this is how I think we should move forward. WDYT?

lsviben:
> cc @lsviben, just FYI this is how I think we should move forward. WDYT?

I like the approach, makes it much easier for users.

lsviben:
Looks good to me, but before approving I have one question. What happens if the user wants to revert the upgrade to KKP 2.21 and downgrade back to 2.20 (for whatever reason). Will all the UserSSHKey handling start failing because the unexpected project field will be there, and they would need to remove this manually?

Not sure if that is a problem, just a though.

xrstf:
To my knowledge, we never supported downgrading KKP and all hell would probably break loose, if I think about the massive changes in CSI/CNI/CCM defaulting logic.

lsviben:
> To my knowledge, we never supported downgrading KKP and all hell would probably break loose, if I think about the massive changes in CSI/CNI/CCM defaulting logic.

OK, makes sense.

/approve
/lgtm

PR #9434, opened 2022-03-24
Cleanup RuleGroup Cluster ref

xrstf:
Kubernetes documentation says regarding ObjectReferences:

> New uses of this type are discouraged because of difficulty describing its usage when embedded in APIs. Ignored fields.  It includes many fields which are not generally honored.  For instance, ResourceVersion and FieldPath are both very rarely valid in actual usage.

We are now more or less stuck with an ObjectReference here, but at least we can document that only the name is ever relevant.

```release-note
NONE
```


xrstf:
@floreks PTAL

xrstf:
@floreks PTAL

PR #9436, opened 2022-03-24
Re-enable AWS Storage Tests

xmudrii:
Since #9233 we had disabled the storage tests on AWS because for unknown reasons, they simply stopped working. We tested _everything_. Clusters, credentials, APIs, we tried awscli, different Kubernetes versions, etc.pp. -- nothing worked.

Now we have updated the KKP seed that hosts our CI cluster, and then also updated the Kubernetes version in our CI cluster. And suddenly, for no apparent reason, the storage tests work again.

Neither @xmudrii nor @xrstf really know why, but we're both happy that it works again. Let's not ask any further questions.

**Release Notes:**
```release-note
NONE
```


xrstf (approved):
/approve

Issue #9439, opened 2022-03-25
Seed GitOps


xrstf:
### What would you like to be added? / User Story
This ticket aims towards allowing cluster admins (Kubernetes cluster admins) to work declaratively on the master/seed clusters in KKP. This means that instead of using the KKP dashboard, admins should be able to manage KKP CRDs with tools like `kubectl` only.

**NB:** This ticket does not include the following aspects:

* User Authentication on the seed (i.e. the question how to authenticate and identify users and whether or not we even want regular KKP users having access to the seed cluster).
* Discussions regarding KCP or other means to have standalone management clusters.

The main task here is to go through our codebase and find all the little places where our API currently does more than simply creating new resources. Those places need to me moved to either a webhook or controllers to make them independent.

To follow the 80/20 rule, we should first aim to fix the resources strictly necessary for userclusters (users, projects, clusters, userprojectbinding) and work on the others (like AllowedRegistry) later.

### TODO

* [ ] Users
* [ ] Projects
* [ ] UserProjectBinding
* [ ] Clusters

### Acceptance criteria
* [ ] Someone with admin access to the seed/master clusters can create a fully functioning usercluster just by doing `kubectl apply`.

PR #9441, opened 2022-03-25
Add new webhooks to improve User/Service Account GitOps handling

xrstf:
This does for Users what #9421 does for UserSSHKeys. This PR adds new webhooks

* a new validating webhook ensures that users are properly defined; it will ensure service account users are configured properly.
* a new mutating webhook will ensure the owner reference from the SA to the project is setup properly

A new field, `spec.Project`, was added to users. It must be set (only) for Service Account Users (the webhooks ensure this from now on).

As there is no user webhook in previous versions, it's not required to shutdown the operator and remove webhooks prior to the migration. :crossed_fingers: 

This is part of #9439.

```release-note
`User.Spec.Project` was added to signal the Service Account <-> Project relationship.
```


pkg/controller/master-controller-manager/serviceaccount-projectbinding-controller/controller.go
```diff
@@ -14,89 +14,99 @@ See the License for the specific language governing permissions and
...
-	for _, owner := range sa.GetOwnerReferences() {
-		if owner.APIVersion == kubermaticv1.SchemeGroupVersion.String() && owner.Kind == kubermaticv1.ProjectKindName &&
-			len(owner.Name) > 0 && len(owner.UID) > 0 {
-			projectName = owner.Name
-			break
-		}
-	}
+	projectName := sa.Spec.Project
 	if len(projectName) == 0 {
+		// TODO: Is this fallback still needed?
```
lsviben, line 146: i dont think so. The API is setting the ownerRef (now after this PR the project field) and if somebody wants to use gitops they can use the new project field.

pkg/controller/master-controller-manager/serviceaccount-projectbinding-controller/controller.go
```diff
@@ -14,89 +14,99 @@ See the License for the specific language governing permissions and
...
 		projectName = sa.GetLabels()[kubermaticv1.ProjectIDLabelKey]
 	}
 
 	if len(projectName) == 0 {
-		return fmt.Errorf("unable to find owing project for the service account %q", sa.Name)
+		return errors.New("unable to find owing project")
```
lsviben: while changing this, you can also change `owing->owner`, not sure if owing is the right word for this

pkg/webhook/user/mutation/mutation.go
```diff
@@ -0,0 +1,176 @@
...
+
+		ownerRefs = append(ownerRefs, metav1.OwnerReference{
+			APIVersion: apiVersion,
+			Kind:       kind,
+			UID:        project.UID,
+			Name:       project.Name,
+		})
+
+		kuberneteshelper.SortOwnerReferences(ownerRefs)
+		user.SetOwnerReferences(ownerRefs)
```
lsviben: We did this in one or 2 more places, and I expect we could use it again somewhere. Maybe it would be useful to extract this to a function?

lsviben (commented):
Added a minor comments, otherwise LGTM.

As the serviceaccount users need to have a label `initialGroup` I think we should also validate that during CREATE so that we make sure its set. Otherwise the service-account-binding-controller will fail as it wont know what group to put the user in. The label is removed after that.

Also as this mechanism of "label->controller generates smt->label gets removed" is a bit wonky, maybe we should put the group as a spec field. From one side, I dont like it because it makes the User CRD even more messy, but on the other hand i feel it should be a field so if something happens to the generated UserProjectBinding it can get reconciled. 

Maybe refactoring this more can be a part of a different PR.

pkg/controller/master-controller-manager/serviceaccount-projectbinding-controller/controller.go
xrstf: The RBAC controller has some code to read the project from the label. Are you sure we can remove this? I would be too scared :D

xrstf:
> As the serviceaccount users need to have a label initialGroup I think we should also validate that during CREATE so that we make sure its set. Otherwise the service-account-binding-controller will fail as it wont know what group to put the user in. The label is removed after that.

Implemented.

In general I refactored the PR to _not_ use the webhooks to set the owner refs, but instead rely on controllers. One new controller takes over the responsibility for SSH keys, the existing controller for SA users takes care of the owner ref now.

I was thinking: If a user has a single YAML file with user, project  & SSH key and `kubectl apply`'s that file, the webhook will most likely not yet have the project in its cache, so it will reject the SSH key. That is super annoying. So I figured we should be less strict in the webhook and only ever validate that objects are internally consistent, not that referenced objects exist (but if they exist, we can check them, which the new `OptimisticallyCheckIfProjectIsValid` does).

pkg/controller/master-controller-manager/serviceaccount-projectbinding-controller/controller.go
lsviben: You mean here?
https://github.com/kubermatic/kubermatic/blob/2ffbd70a7531cd9056fd5a0eb619f6a977899d29/pkg/controller/master-controller-manager/rbac/sync_resource.go#L223-L240

It checks for ownerRef first as well, which should be set by the new controller. 

I mean yeah, somebody put this check here in serviceAccount controller for some reason I guess. But it could be some legacy. I dont see anywhere in code that the `projectID` label is put on serviceaccount users(or users at all), and on dev no user has it (but thats no proof ofc).

Maybe lets remove it as a separate PR so we can revert it easily if needed

lsviben:
> > As the serviceaccount users need to have a label initialGroup I think we should also validate that during CREATE so that we make sure its set. Otherwise the service-account-binding-controller will fail as it wont know what group to put the user in. The label is removed after that.
> 
> Implemented.
> 
> In general I refactored the PR to _not_ use the webhooks to set the owner refs, but instead rely on controllers. One new controller takes over the responsibility for SSH keys, the existing controller for SA users takes care of the owner ref now.
> 
> I was thinking: If a user has a single YAML file with user, project & SSH key and `kubectl apply`'s that file, the webhook will most likely not yet have the project in its cache, so it will reject the SSH key. That is super annoying. So I figured we should be less strict in the webhook and only ever validate that objects are internally consistent, not that referenced objects exist (but if they exist, we can check them, which the new `OptimisticallyCheckIfProjectIsValid` does).

Ok, that makes sense. K8s native objects do that to when referencing other objects so we should follow this logic as well. Probably moving we should also move from using the ownerRefs for anything else other than for deletion purposes so we don't depend on it anymore, and just use the `spec` for referencing relations.

xrstf:
> Probably moving we should also move from using the ownerRefs for anything else other than for deletion purposes so we don't depend on it anymore, and just use the spec for referencing relations.

I'm on it, it just takes forever ^^

pkg/webhook/user/mutation/mutation.go
```diff
@@ -0,0 +1,122 @@
...
+
+	mutatedKey, err := json.Marshal(newUser)
+	if err != nil {
+		return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("marshaling user object failed: %w", err))
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, mutatedKey)
+}
+
+func (h *AdmissionHandler) ensureProjectRelation(ctx context.Context, user *kubermaticv1.User, oldUser *kubermaticv1.User) error {
```
lsviben: ensureProjectRelation (for both user and sshkeys) is now not mutating anything, but validating. Maybe for the serviceaccounts we dont need the mutating webhook at all and we can move this to the validating webhook ( it already validates that project is immutable, so jus the OptimisticallyCheckIfProjectIsValid

xrstf: Excellent point.

xrstf: In fact, since we do not set the owner ref anymore, there is no need for a mutating webhook for users. So I moved the validation logic into the validating webhook and removed the mutating one.

PR #9446, opened 2022-03-25
refactor master/user-project-binding controller

xrstf:
This PR's goal was to not let the KKP API add the CleanupFinalizer to new UserProjectBindings (so `pkg/provider/kubernetes/member.go` was the starting point for this PR). It was weird that the API was always insisting on adding this finalizer, even though it itself doesn't handle the deletion. However, the user-project-binding controller on the master does control the deletion and "owns" the finalizer. So this PR rewrote the controller to add the finalizer when appropriate and removed its handling from the API.

To make the controller easier to understand, I added a few helper functions for dealing with owner references. Since owner refs and finalizers are added later, dynamically, a whole bunch of tests had to be adjusted accordingly.

Part of #9439

```release-note
NONE
```


xrstf:
@zreigz PTAL

PR #9448, opened 2022-03-28
KubeOne container-runtime migrate endpoint

### LINKED ISSUE #9452: KubeOne Migrate Endpoint
Extend PATCH endpoint for KubeOne container-runtime migration

imharshita:

**What does this PR do / Why do we need it**: Extend PATCH endpoint for KubeOne container-runtime migration

Fixes #9452

**Documentation**:

```release-note
none
```


pkg/api/v2/types.go
```diff
@@ -451,9 +451,10 @@ type ExternalClusterCloudSpec struct {
 
 type KubeOneSpec struct {
 	// Manifest Base64 encoded manifest
-	Manifest  string            `json:"manifest"`
-	SSHKey    KubeOneSSHKey     `json:"sshKey"`
-	CloudSpec *KubeOneCloudSpec `json:"cloudSpec"`
+	Manifest               string            `json:"manifest"`
+	SSHKey                 KubeOneSSHKey     `json:"sshKey"`
+	ContainerRuntimeDocker bool              `json:"containerRuntimeDocker"`
```
zreigz: it should be a string,  like for the KKP Cluster object. It will support other container runtimes in the future

pkg/handler/v2/external_cluster/kubeone.go
```diff
@@ -74,55 +80,163 @@ func importKubeOneCluster(ctx context.Context, name string, userInfoGetter func(
...
 }
+
+func isContainerRuntimeDocker(ctx context.Context,
+	externalCluster *kubermaticv1.ExternalCluster,
+	externalClusterProvider provider.ExternalClusterProvider,
+) (bool, error) {
+	var containerRuntimeDocker bool
+	nodes, err := externalClusterProvider.ListNodes(ctx, externalCluster)
+	if err != nil {
+		return containerRuntimeDocker, err
```
zreigz: this one is dangerous. You don't know what kind of error happened. You should retrieve the container runtime type not assume it's docker. If ListNodes throws an error because of controller runtime then you should have some predefined error for this. You can have a runtime error, connection error, and even more

zreigz (changes requested):

pkg/handler/v2/external_cluster/kubeone.go
imharshita: good point! thanks:)

pkg/handler/v2/external_cluster/kubeone.go
```diff
@@ -74,55 +79,165 @@ func importKubeOneCluster(ctx context.Context, name string, userInfoGetter func(
...
+	nodes, err := externalClusterProvider.ListNodes(ctx, externalCluster)
+	if err != nil {
+		return "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+	for _, node := range nodes.Items {
+		if _, ok := node.Labels[NodeControlPlaneLabel]; ok {
+			containerRuntimeVersion := node.Status.NodeInfo.ContainerRuntimeVersion
+			strSlice := strings.Split(containerRuntimeVersion, ":")
+			for _, v := range strSlice {
+				containerRuntime = v
```
zreigz: you can return here

zreigz: maybe return strSlice[0]

imharshita: i just did this to avoid the case when the slice is empty, indexnotfound err.
but yes, we can return here.

zreigz: otherwise, you will return an empty string. Is it OK?

imharshita: we can return an err, in case of empty string.
wdyt?

imharshita: err could be something like no container runtime found...

zreigz: err or some default, I am not sure

zreigz: the error will be better

PR #9456, opened 2022-03-28
mark usersshkey.project as optional

xrstf:
The plan in #9421 of "create a new mandatory field and then fill it during the installation" doesn't work, as Kubernetes rejects updates to invalid objects, so even if you tried to add the `spec.project` field, it won't work. So sadly we have to mark the field as optional, even though it is not. :-(

```release-note
NONE
```


xrstf:
This error was not caused by Kubernetes, but by us having an old webhook in the cluster, which was confused when we tried to edit the UserSSHKeys and add a field that it doesn't know.

Removing the webhook temporarily is the correct solution. So this PR can be closed.

PR #9457, opened 2022-03-28
disallow snapshot restore if etcd-launcher is not enabled

### LINKED ISSUE #9453: Snapshot restore shouldn't be allowed if etcd launcher is not enabled

### Acceptance criteria

PratikDeoghare:

Etcd launcher is required for proper working of snapshot restore. This PR checks if it enabled before restore is created. 

Fixes #9453 

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/etcdrestore/validation.go
+++ b/pkg/webhook/etcdrestore/validation.go
@@ -83,7 +83,7 @@ func (v validator) getCluster(ctx context.Context, clusterRef v1.ObjectReference
 		return nil, fmt.Errorf("failed to get current Seed: %w", err)
 	}
 	if seed == nil {
-		return nil, fmt.Errorf("webhook not configured for a Seed cluster, cannot validate Addon resources")
+		return nil, fmt.Errorf("webhook not configured for a Seed cluster, cannot validate EtcdRestore resources")
 	}
 
 	client, err := v.seedClientGetter(seed)
--- a/pkg/webhook/etcdrestore/validation.go
+++ b/pkg/webhook/etcdrestore/validation.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package etcdrestore
+
+import (
+	"context"
+	"fmt"
+
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/provider"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	client2 "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// validator for mutating Kubermatic EtcdRestore CRD.
+type validator struct {
+	seedGetter       provider.SeedGetter
+	seedClientGetter provider.SeedClientGetter
+}
+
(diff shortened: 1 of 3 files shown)


pkg/handler/v2/etcdrestore/etcdrestore.go
```diff
@@ -51,6 +51,10 @@ func CreateEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider prov
 			return nil, err
 		}
 
+		if !c.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] {
```
lsviben: Its ok to do it here. But we are moving everything to webhooks to support GitOps at one point, so it would be better to do this in a webhook

PratikDeoghare: Thanks! I will take check how I can do it there.

PratikDeoghare: @lsviben added webhooks. Ptal. :pray:

PratikDeoghare: Spoke too soon. Testing this locally now.

pkg/webhook/etcdrestore/validation.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, fmt.Errorf("failed to get current Seed: %w", err)
+	}
+	if seed == nil {
+		return nil, fmt.Errorf("webhook not configured for a Seed cluster, cannot validate EtcdRestore resources")
+	}
+
+	client, err := v.seedClientGetter(seed)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Seed client: %w", err)
+	}
```
lsviben, line 92: I think that there is no need to get the seed and then the client, you can just get the local client from the `mgr` and get the Cluster. Anyway the EtcdRestores will be created on Seeds.

lsviben, line 92: I think that there is no need to get the seed and then the client, you can just get the local client from the `mgr` and get the Cluster. Anyway the EtcdRestores will be created on Seeds.

pkg/webhook/etcdrestore/validation_test.go
```diff
@@ -0,0 +1,118 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
lsviben, line 2: NIT: 2022

lsviben, line 2: NIT: 2022

pkg/webhook/etcdrestore/validation.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"fmt"
+
+	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/provider"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	client2 "sigs.k8s.io/controller-runtime/pkg/client"
```
lsviben, line 28: please change `client2` to `ctrlruntimeclient`

lsviben, line 28: please change `client2` to `ctrlruntimeclient`

lsviben (commented):
For this you will also need to deploy a `validatingwebhook`, check https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/seed/resources/kubermatic/webhooks.go.

But TBH, i am checking this whole feature again and I fear I may have sent you in the wrong direction in my previous comment. Im not sure anymore that this should be a a validating webhook for EtcdRestore, and rather that this should be checked in the etcdrestore controller(https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go), and just blocked there. My thinking here is that the EtcdRestore by itself is valid, there is nothing wrong with its spec, the problem is that etcd-launcher is not enabled, but this is done somewhere else, so no point to block EtcdRestore creation. 

Users wont be able to see this right away, but then we can return this check in the KKP API create you made earlier, this should cover most users.


@moadqassem what do you think?

PratikDeoghare:
@lsviben Thanks for the review.

> But TBH, i am checking this whole feature again and I fear I may have sent you in th....

No worries. Happy to rework it.

PratikDeoghare:
@moadqassem and I came up with this https://github.com/kubermatic/kubermatic/pull/9570 after discussion. 
Closing this PR in favor of that. Ptal.

PR #9458, opened 2022-03-28
shutdown usersshkey webhook before attempting a migration

xrstf:
In #9421 I added some migration, which failed on dev. As it turns out, our webhook would reject the new UserSSHKeys, as it does not know the new `spec.project` field. There are 2 options we have:

1. Shut down the webhook before migrating the data. This means:
   * Install new CRDs
   * Shutdown KKP Operator
   * Remove webhook
   * Migrate data
   * Install new Helm chart, which will
   * ... scale up the KKP Operator, which will
   * ... re-create the webhook
   The downside here is that we need slightly more code to deal with the shutdown, but this could be (and was in this PR) copied  from the 2.20 migration code, which did the exact same thing.
2. Install first, migrate later:
   * Install new CRDs
   * Install new KKP Operator, which installs the new webhook
   * Migrate data
   The downside to this is that we risk that new controllers will start up, see invalid/incomplete data and go crazy.

Because of the possibility that controllers might choke on old, incomplete data, I chose to implement option 1 in this PR.

```release-note
NONE
```

### DIFF
--- a/pkg/install/stack/kubermatic-master/migrations.go
+++ b/pkg/install/stack/kubermatic-master/migrations.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/install/stack/kubermatic-master/migrations.go
+++ b/pkg/install/stack/kubermatic-master/migrations.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


pkg/install/stack/kubermatic-master/migrations.go
```diff
@@ -0,0 +1,110 @@
...
+			return fmt.Errorf("key %s no project owner reference, cannot determine project association", key.Name)
+		}
+
+		oldKey := key.DeepCopy()
+		key.Spec.Project = projectID
+
+		if err := client.Patch(ctx, &key, ctrlruntimeclient.MergeFrom(oldKey)); err != nil {
+			return fmt.Errorf("failed to update key %s: %w", key.Name, err)
+		}
+	}
```
embik: Not that it matters too much, but IMO this should happen within the `if needMigration {}` block. No key should be migrated if no migration is needed, but it seems unnecessary to loop over the keys a second time if that isn't needed.

embik (changes requested):

embik (approved):

PR #9459, opened 2022-03-28
fix missing permissions for webhook on pure seed clusters

xrstf:
We forgot to create the necessary permissions on seed clusters. The code was already meant to be used like that, we simply forgot to call it.

```release-note
NONE
```


pkprzekwas (approved):
lgtm

embik:
Does this need a cherry-pick to `release/v2.20`?

PR #9460, opened 2022-03-28
Create nodeport-proxy resources (for LoadBalancer strategy) while reconciling other cluster resources

### LINKED ISSUE #9083: Refactor nodeportproxy resource creators to include them in fixtures

This is a backlog item / reminder for myself that it might be a good idea to look into refactoring the `nodeportproxy` package under `pkg/resources`. The reason for that is that _except_ the `Service` object, all resources for `nodeport-proxy-envoy` are created here: https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/nodeportproxy/nodeportproxy.go#L156-L194.

The `Service` is added in `GetServiceCreators()` with in the seed-controller-manager's `kubernetes` controller here: https://github.com/kubermatic/kubermatic/blob/a426e68e5706308a564e8cd5628bbdeffa4b0cc7/pkg/controller/seed-controller-manager/kubernetes/resources.go#L276-L278

We should investigate adding the other resources to the respective `Get<X>Creators()` funcs of that controller as well instead of calling `nodeportproxy.EnsureResources`: https://github.com/kubermatic/kubermatic/blob/a426e68e5706308a564e8cd5628bbdeffa4b0cc7/pkg/controller/seed-controller-manager/kubernetes/resources.go#L151

That would add the `nodeportproxy` resources to our fixtures, not only the `Service`. #9059 only happened because we did not realise that labels where changing when there was no intention to change them.

This should be done carefully because it might be used elsewhere, I haven't checked.

### Acceptance criteria

- `nodeportproxy.EnsureResources` is eliminated and nodeport-proxy resources are added based on conditions in `seed-controller-manager/kubernetes/resources.go`.
- `nodeportproxy` resources are added to fixtures.

embik:
The resources created for the `LoadBalancer` expose strategy lived in their own little world that no other part of the code was aware of beyond calling `nodeportproxy.EnsureResources`. I broke `nodeport-proxy` a while back because of weird variable shadowing and no fixtures that warned of the unintended chage.

This PR cleans up the last traces of shadowing the `name` variable in the `nodeportproxy` package and reworks the whole package to be included in the `seed-controller-manager`'s resource reconcile logic. That adds fixtures for them as well. 

Fixes #9083

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/resources.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/resources.go
@@ -151,12 +151,6 @@ func (r *Reconciler) ensureResourcesAreDeployed(ctx context.Context, cluster *ku
 		return nil, err
 	}
 
-	if cluster.Spec.ExposeStrategy == kubermaticv1.ExposeStrategyLoadBalancer {
-		if err := nodeportproxy.EnsureResources(ctx, r.Client, data); err != nil {
-			return nil, fmt.Errorf("failed to ensure NodePortProxy resources: %w", err)
-		}
-	}
-
 	// Remove possible leftovers of older version of Gatekeeper, remove this in 1.19
 	if err := r.ensureOldOPAIntegrationIsRemoved(ctx, data); err != nil {
 		return nil, err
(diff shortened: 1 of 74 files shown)


rastislavs:
/lgtm

nice :)

PR #9463, opened 2022-03-29
hack: configure seed-name and external-url via env variables

embik:
Minor nit I had with the `hack/run-seed-controller-manager.sh` script: Most of it was parameterised via environment variables, but `seed-name` and `external-url` were hardcoded to values only sensible for the dev environment.

This PR allows setting `KUBERMATIC_SEED` and `KUBERMATIC_EXTERNAL_URL` to pass different values to the local `seed-controller-manager` instance.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #9469, opened 2022-03-29
Machine controller health check extended

### LINKED ISSUE #8837: Initial MachineDeployment bypass the machine-controller-webhook

mlavacca:
The health check for the machine-controller has been extended so that:
* the `machineController` deployment must be healthy
* the `machineControllerWebhook` deployment must be healthy
* the `mutatingWebhookConfiguration` in the user cluster must exist

When the above conditions are met, the machineController health is set to up, otherwise, it is set to down.

Fixes #8837 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/health.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/health.go
@@ -24,10 +24,12 @@ import (
 	kubermaticv1helper "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1/helper"
 	"k8c.io/kubermatic/v2/pkg/resources"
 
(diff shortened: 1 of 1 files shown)


moadqassem (approved):
/approve
/lgtm

PR #9471, opened 2022-03-30
Allow toggling user-ssh-keys-agent after cluster creation

### LINKED ISSUE #9470: Allow enabling/disabling user-ssh-keys-agent after cluster creation

As a user, I'd like the ability to toggle `user-ssh-keys-agent` after cluster creation. Right now it's treated as an immutable field and cannot be modified once the cluster is created. This was intentional due to the risks involved, mainly:

- When **enabled** after the cluster is created, any previously added ssh keys in the worker
nodes(except the keys that have been added during the cluster creation) will be
deleted.
- When **disabled** after the cluster is created, any pre-existing keys won't be cleaned up.

### Acceptance criteria

It is possible to toggle `user-ssh-keys-agent` after cluster creation.

ahmedwaleedmalik:

Make the `enableUserSSHKeyAgent` field mutable in cluster spec. So that users can modify it after the cluster has been created.

Fixes #9470

**Documentation**:

```release-note
`spec.enableUserSSHKeyAgent` was made mutable in Cluster resource; this adds ability to enable/disable user-ssh-keys-agent after cluster creation
```


ahmedwaleedmalik:
This is not required anymore.

PR #9473, opened 2022-03-30
Fix context handling

xrstf:
This PR fixes one case where we stored a context in a struct (which is a code smell) and improves the operator to not use so many background contexts.

```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

xmudrii (approved):

PR #9474, opened 2022-03-30
remove legacy owner-remover tool

xrstf:
The owner remover was added in #1264 and since then was only "well, it compiles"-style kept alive. Chances are that it doesn't properly work anymore on all the resources and with all the changes in KKP since 2018. Also it was never documented and probably never used by anyone.

This PR simply gets rid of it entirely.

```release-note
The legacy `owner-remover` tool was removed from the codebase.
```


xmudrii (approved):
/lgtm
/approve

PR #9475, opened 2022-03-30
fix make cover

xrstf:
This fixes `make cover`.

```release-note
NONE
```

### DIFF
--- a/hack/coverage.sh
+++ b/hack/coverage.sh
@@ -16,10 +16,7 @@
 
 ### Generate test coverage statistics for Go packages.
 ###
-### Works around the fact that `go test -coverprofile` currently does not work
-### with multiple packages, see https://code.google.com/p/go/issues/detail?id=6909
-###
-###     Usage: cover.sh [--html]
+###     Usage: coverage.sh [--html]
 ###
 ###     --html      Additionally create HTML report and open it in browser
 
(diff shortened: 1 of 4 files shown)


vgramer:
@xrstf github doesn't let me comment on code that hasn't change but when I execute script I have an error when it try to show the report. otherwise the works good :)

in the functionn `show_cover_report` surronding `$2` with quote generate error

```bash
show_cover_report() {
  echo "Generating $1 report"
  go tool cover -"$1"="$profile" "$2"
}

```

call `show_cover_report func`: 
```bash
Generating func report
+ go tool cover -func=.cover/cover.out ''
too many arguments
```

call `show_cover_report html "-o $workdir/cover.html"`:
```
Generating html report
+ go tool cover -html=.cover/cover.out '-o .cover/cover.html'
flag provided but not defined: -o .cover/cover.html
Usage of 'go tool cover':
Given a coverage profile produced by 'go test':
	go test -coverprofile=c.out

Open a web browser displaying annotated source code:
	go tool cover -html=c.out

Write out an HTML file instead of launching a web browser:
	go tool cover -html=c.out -o coverage.html

Display coverage percentages to stdout for each function:
	go tool cover -func=c.out

Finally, to generate modified source code with coverage annotations
(what go test -cover does):
	go tool cover -mode=set -var=CoverageVariableName program.go

Flags:
  -V	print version and exit
  -func string
    	output coverage profile information for each function
  -html string
    	generate HTML representation of coverage profile
  -mode string
    	coverage mode: set, count, atomic
  -o string
    	file for output; default: stdout
  -var string
    	name of coverage variable to generate (default "GoCover")

  Only one of -html, -func, or -mode may be set.
make: *** [cover] Error 1
```

possibles fixes:
1. remove the quote around `$2` `show_cover_report` in but I guess shellcheck will not like it
2. use a `if` in   `show_cover_report` 
```bash
show_cover_report() {
  echo "Generating $1 report"
  if [ -z "$2" ]; then
    go tool cover -"$1"="$profile"
  else
     go tool cover -"$1"="$profile" -o "$2"
  fi
}
```
and change  call to HTML report from `show_cover_report html "-o $workdir/cover.html" ` to `show_cover_report html "$workdir/cover.html"`

xrstf:
@vgramer please try again, I also removed the workaround for the cover thing that got solved in Go 1.10 already :grin:

xrstf:
(shellcheck was just for informational purpose, it's doesn't have to pass)

hack/coverage.sh
```diff
@@ -16,44 +16,37 @@
...
-}
+rm -rf "$workdir"
+mkdir "$workdir"
+
+echo "Generating coverage data…"
+go test -covermode="$covermode" -coverprofile="$profile" ./cmd/... ./pkg/...
 
-generate_cover_data "$(go list ./cmd/... ./pkg/...)"
-show_cover_report func
+generate_cover_data
```
vgramer:
```suggestion
```

xrstf: Ooops :D Fixed

vgramer (commented):
/lgtm

PR #9476, opened 2022-03-30
Allow configuration for front-loadbalancer service of nodeportproxy

### LINKED ISSUE #7048: Restrict nodeport_proxy service to specific IP range
**User Story**
Restrict nodeport-proxy service to specific ip ranges, by exposing `loadBalancerSourceRanges` spec for nodeport-proxy service,  which is currently allowed to `0.0.0.0/0` by default.

**Configuration**
Following is the Seed CustomResourceDefinition we are using

```
kind: Seed
metadata:
  name: kubermatic
  namespace: kubermatic
spec:
  country: EU
  location: Ireland
  kubeconfig:
    name: kubeconfig-europe-west1
    namespace: kubermatic
  expose_strategy: "NodePort"  
  datacenters:
    aws-eu-west-1a:
      location: EU West (Ireland)
      country: Ireland
      spec:
        aws:
          images:
            centos: ""
            coreos: ""
            flatcar: ""
            rhel: ""
            sles: ""
            ubuntu: ""
          region: eu-west-1
  nodeport_proxy:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-type: nlb
      external-dns.alpha.kubernetes.io/hostname: "*.kubermatic.kubermatic.example.com" 
      external-dns.alpha.kubernetes.io/ttl: "10"
```

**Acceptance criteria**
- Provide a way to restrict nodeport-proxy to given IP ranges from seed CustomResourceDefinition by exposing `loadBalancerSourceRanges` spec for nodeport-proxy service

### LINKED ISSUE #9301: Allow to configure annotations on Front-LoadBalancers
For exposeStrategy=NodePortProxy we already support setting `seed.spec.nodePortProxy.annotations` to allow admins to tweak the LB service. For exposeStrategy=LoadBalancer, there is currently no option to customize the Front-LoadBalancer (pkg/resources/nodeportproxy/nodeportproxy.go)

One approach woul be to simply re-use the annotations from the NodePortProxy, since no seed can have both strategies at the same time. On the other hand, this could be also be a surprising.

### Acceptance criteria
* There is a way to configure annotations on the front LB.

ahmedwaleedmalik:

Fixes #9301 #7048

**Documentation**:

```release-note
- Add support for configuration annotations and loadBalancerSourceRanges for front-loadbalancer service of node port proxy
- For seed CR, spec.NodeportProxy.Annotations is deprecated and spec.NodeportProxy.Envoy.LoadBalancerService.Annotations should be used instead.
```


docs/zz_generated.seed.yaml
```diff
@@ -367,6 +367,12 @@ spec:
     envoy:
       # DockerRepository is the repository containing the component's image.
       dockerRepository: docker.io/envoyproxy/envoy-alpine
+      loadBalancerService:
+        # Annotations are used to further tweak the LoadBalancer integration with the
+        # cloud provider.
+        annotations: null
+        # LoadBalancerSourceRanges will restrict front-loadbalancer service to specific ip ranges
+        sourceRanges: null
```
xrstf: Just from looking at this, I would not know how to configure this. Is Source ranges an object? A map? Does it look like AWS EC2 ingress rules? Are those CIDRs? Can you add some sentence to the docs like "This should be a list of ............." and ideally extend the example YAML generator so it doesn't generate `null`, but `[]` (which is a bit more helpful also).

Same goes for the annotations field, which would look nicer as `[]` in the example YAML.

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -197,14 +197,26 @@ type NodeportProxyConfig struct {
...
+type EnvoyLoadBalancerService struct {
+	// Annotations are used to further tweak the LoadBalancer integration with the
+	// cloud provider.
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// LoadBalancerSourceRanges will restrict front-loadbalancer service to specific ip ranges
+	SourceRanges []string `json:"sourceRanges,omitempty"`
+}
+type NodePortProxyComponentEnvoy struct {
+	LoadBalancerService    EnvoyLoadBalancerService `json:"loadBalancerService,omitempty"`
+	NodeportProxyComponent `json:",inline"`
```
xrstf, line 222: Please list embedded fields before regular fields.

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -197,14 +197,26 @@ type NodeportProxyConfig struct {
...
 	// service.
 	Updater NodeportProxyComponent `json:"updater,omitempty"`
 }
 
+type EnvoyLoadBalancerService struct {
+	// Annotations are used to further tweak the LoadBalancer integration with the
+	// cloud provider.
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// LoadBalancerSourceRanges will restrict front-loadbalancer service to specific ip ranges
+	SourceRanges []string `json:"sourceRanges,omitempty"`
```
xrstf: Can you add some validation for this?

rastislavs:
This PR seems to configure `loadBalancerSourceRanges` for the nodeport-proxy that is deployed in user-cluster namespace in case of the LoadBalancer expose strategy.

What about the seed-level nodeport-proxy used for NodePort expose strategy? (see [pkg/controller/operator/seed/resources/nodeportproxy](https://github.com/kubermatic/kubermatic/blob/2ffbd70a7531cd9056fd5a0eb619f6a977899d29/pkg/controller/operator/seed/resources/nodeportproxy/service.go)). I think the https://github.com/kubermatic/kubermatic/issues/7048 is more about that one, as it explicitly mentions the ` nodeport-proxy` service, not the `front-loadbalancer` service.

But in any case, it makes sense to have this option applied for both of them :)

ahmedwaleedmalik:
> This PR seems to configure `loadBalancerSourceRanges` for the nodeport-proxy that is deployed in user-cluster namespace in case of the LoadBalancer expose strategy.
> 
> What about the seed-level nodeport-proxy used for NodePort expose strategy? (see [pkg/controller/operator/seed/resources/nodeportproxy](https://github.com/kubermatic/kubermatic/blob/2ffbd70a7531cd9056fd5a0eb619f6a977899d29/pkg/controller/operator/seed/resources/nodeportproxy/service.go)). I think the #7048 is more about that one, as it explicitly mentions the ` nodeport-proxy` service, not the `front-loadbalancer` service.
> 
> But in any case, it makes sense to have this option applied for both of them :)

You are actually right but `.nodeportProxy.envoy.loadBalancerService.sourceRanges` can't be directly applied to that service since it's annotations come from ``.nodeportProxy.annotations` and not ``.nodeportProxy.envoy.loadBalancerService.annotations`. This would be very confusing for the end-users and doesn't seem like the right approach.

PR #2460, opened 2018-11-29
Enable etcd experimental corruption checks

mrIncompetent:
**What this PR does / why we need it**:
Enable etcd experimental corruption checks.

**Release note**:
```release-note
Enable etcd corruption checks
```

PR #9477, opened 2022-03-30
Enable etcd corruption checks for etcd 3.5 and increase to 4h cycle

embik:
Yesterday, the etcd team announced [that they do not recommend usage of etcd 3.5 in production anymore](https://groups.google.com/a/kubernetes.io/g/dev/c/B7gJs88XtQc/m/rSgNOzV2BwAJ) due to data consistency issues that can occur when etcd is terminated under high load. They expect that an upcoming etcd 3.5.3 release will fix this problem, but no ETA has been communicated so far.

This PR, as an initial mitigation, enables corruption checks for etcd 3.5 StatefulSets. This was initially added in #2460 as a Seed level feature flag and is available both in plain `etcd` and `etcd-launcher`. We're setting it to always true for the affected etcd version.

Corruption checks will not repair corrupted nodes, but will prevent them from starting, and therefore hopefully prevent data inconsistency in the live cluster. Corrupted members need to be [manually removed and replaced](https://docs.kubermatic.com/kubermatic/v2.20/cheat_sheets/etcd/replace_a_member/) if running plain `etcd`. `etcd-launcher` can be forced to rebuild the member by deleting the matching PV and [letting the automatic PV recovery kick in](https://docs.kubermatic.com/kubermatic/v2.20/cheat_sheets/etcd/etcd-launcher/#automated-persistent-volume-recovery).

In addition, this PR bumps the periodic corruption checks up to 4 hours from 10 minutes. The reason behind that is [this comment on the etcd data consistency issue](https://github.com/etcd-io/etcd/issues/13766#issuecomment-1076277339). Since the checks produce some overhead, running them every 10 minutes seems rather excessive.

Fixes #

**Documentation**:

```release-note
For user clusters that use etcd 3.5 (Kubernetes 1.22 clusters), etcd corruption checks are turned on to detect [etcd data consistency issues](https://github.com/etcd-io/etcd/issues/13766). Checks run at etcd startup and every 4 hours
```

### DIFF
--- a/pkg/resources/etcd/statefulset.go
+++ b/pkg/resources/etcd/statefulset.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"text/template"
 
+	semverlib "github.com/Masterminds/semver/v3"
 	"github.com/Masterminds/sprig/v3"
 
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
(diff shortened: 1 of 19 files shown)


pkg/resources/etcd/statefulset.go
```diff
@@ -73,6 +74,25 @@ func StatefulSetCreator(data etcdStatefulSetCreatorData, enableDataCorruptionChe
...
+
+			etcdConstraint, err := semverlib.NewConstraint(">= 3.5.0, < 3.6.0")
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse etcd constraint: %w", err)
+			}
+
+			// enable initial and periodic etcd data corruption checks by default if running etcd 3.5.
+			// The etcd team has recommended to enable this feature for etcd 3.5 due to data consistency issues.
+			// Reference: https://groups.google.com/a/kubernetes.io/g/dev/c/B7gJs88XtQc/m/rSgNOzV2BwAJ
+			if matches, _ := etcdConstraint.Validate(imageTagVersion); matches {
```
xmudrii: Nit: you can use `Check` instead of `Validate`.

xmudrii (approved):
/lgtm
/approve

pkg/resources/etcd/statefulset.go
embik: Yeah good point, not sure why I went with `Validate`. Fixed.

xmudrii (approved):

Issue #8384, opened 2021-12-06
Create ApplicationInstallationController


SimonTheLeg:
**Acceptance Criteria**
* ensure that each user cluster has its separate ApplicationInstallationController
* ApplicationInstallationController can receive on ApplicationInstallations events (simple printf("ApplicationInstallation xyz has been created") is sufficient)

**Implementation Hints**
* during the app proposal we came to the conclusion that the ApplicationInstallationController should live in the seed cluster in the namespace that belongs to the corresponding user-cluster

**tasks**:
*These tasks were implemented in a bigger PR(#9313)but to facilitate the review, it split in smaller pr*
- [x] update the ApplicationInstallation status to store that applicationVersion used to install /uninstall the application. (#9417)
- [x] install applicationInstallationCR into user-cluster (#9418)
- [x] simple applicationInstallation Controller that enqueue ApplicationInstallation For ApplicationDefition (#9500)
- [x] implement logic to call ApplicationInstaller (#9559)

dermorz:
I added the open questions to the issue description:
>**Open questions**
>* Where does the ApplicationInstallationController live? Will it be part of the user-cluster-controller-manager or should it be a stand-alone controller like the machine-deployment-controller? (-> add reference to research story here)

We need to answer these questions before we can continue estimating.

PR #9500, opened 2022-03-31
Application controller skeleton

vgramer:

This pr adds the skeleton for the application installation controller. domain logic will be implemented in further.

part of #8384

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
+++ b/pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
@@ -1,0 +1,149 @@
+/*
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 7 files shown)


pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
```diff
@@ -0,0 +1,149 @@
...
+		var res []reconcile.Request
+		for _, appInstallation := range appList.Items {
+			if appInstallation.Spec.ApplicationRef.Name == applicationDefinition.GetName() {
+				res = append(res, reconcile.Request{NamespacedName: types.NamespacedName{Name: appInstallation.Name}})
+			}
+		}
+
+		if res != nil {
+			log.Debugf("ApplicationDefinition '%s' has changed. Enqueue ApplicationInstallations %v", applicationDefinition.GetName(), res)
+		}
```
xrstf: I think this is redundant, given the logs in the Reconcile() function. This would be the only controller in KKP that logs which elements it puts on the queue.

xrstf (commented):
One minor nitty picky

xrstf (approved):
/approve

PR #9508, opened 2022-03-31
add validation for ApplicationDefinitions

### LINKED ISSUE #8368: ApplicationDefinitions are validated before create/update
**Acceptance Criteria**

- [x] must be implemented according to [Refactor Cluster Defaulting & Validation](https://github.com/kubermatic/kubermatic/pull/8346) criteria
- [x] validate those constraints have valid values (i.e. k8s and kkp versions range are correct)
- [x] only one source must be defined and be valid (i.e. git or helm)
- [x] only one template method must be defined and be valid (currently only helm)
- [x] the version field inside versions[] must be unique. There cannot be the same value twice inside the array
- [x] since the ApplicationDefinitions live inside the master-cluster they should be validated from there as well

**Implementation Hints**
* we think an AdmissionWebhook will be required to do all the validation
* it could be a good idea to place it in the master-controller-manager. We also think it should be a separate controller to do the validation

SimonTheLeg:

Fixes #8368 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/validation/application_definition.go
+++ b/pkg/validation/application_definition.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


vgramer (commented):
/lgtm

ahmedwaleedmalik (approved):
/approve

PR #9511, opened 2022-04-01
Support for disabling kubernetes-dashboard

### LINKED ISSUE #9510: Support for disabling kubernetes-dashboard

As a user, I'd like the ability to disable `kubernetes-dashboard` for user clusters.

### Acceptance criteria

- `kubernetes-dashboard` should be enabled by default
- There should be an option to disable it, per cluster basis
- UI support

ahmedwaleedmalik:

Introduces a new field in cluster spec `enableKubernetesDashboard` which can be used to toggle `kubernetes-dashboard`. Since `kubernetes-dashboard` should be enabled by default this field has a default value of `true`.

Fixes #9510

**Documentation**:

```release-note
Support for disabling kubernetes-dashboard
```


pkg/controller/user-cluster-controller-manager/resources/controller.go
```diff
@@ -305,6 +307,7 @@ type reconciler struct {
 	ccmMigration                 bool
 	ccmMigrationCompleted        bool
 	enableOperatingSystemManager bool
+	kubernetesDashboardEnabled   bool
```
xrstf: Do you actually need this? Your code is already correctly using cluster.Spec everywhere in the reconciler. Ideally the usercluster-ctrl-mgr should react dynamically to changes in the cluster without needing to restart.

xrstf (commented):
C

pkg/controller/user-cluster-controller-manager/resources/controller.go
ahmedwaleedmalik: `usercluster-ctrl-mgr` depends on the flags and not on the cluster.Spec. Maybe I'm misunderstanding your comment?

xrstf: I've pushed a commit into your branch to show what I am talking about. I see where you came from and what you took as a template, but I think we can do better than the old code that relied on CLI flags. :)

Let me know what you think

ahmedwaleedmalik: Oh, so you meant that, I was also curious why we don't do this already. I love the idea and it sounds super good to me 👍

charts/kubermatic-operator/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
```diff
@@ -1472,6 +1472,11 @@ spec:
               debugLog:
                 description: DebugLog enables more verbose logging by KKP's usercluster-controller-manager.
                 type: boolean
+              enableKubernetesDashboard:
+                default: true
+                description: EnableKubernetesDashboard controls whether the kubernetes-dashboard
+                  should be deployed for the user cluster or not.
+                type: boolean
```
xrstf: I know that we have the exact same flag for OSM, but during the 2.20 restructing we kind of decided to not have these boolean flags anymore, because they are impossible to extend. If another dashboard-related option is added in the future, we end up with multiple flags next to each other, like here: https://github.com/kubermatic/kubermatic/blob/master/pkg/apis/kubermatic/v1/settings.go#L55-L58

I would propose that we use a more extensible structure like

```yaml
spec:
  kubernetesDashboard:
    enabled: true|false
```

xrstf (commented):
Comment

pkg/api/v1/types.go
```diff
@@ -855,6 +855,9 @@ type ClusterSpec struct {
 	// EnableOperatingSystemManager enables OSM which in-turn is responsible for creating and managing worker node configuration
 	EnableOperatingSystemManager bool `json:"enableOperatingSystemManager,omitempty"`
 
+	// KubernetesDashboard holds the configuration for kubernetes-dashboard component
+	KubernetesDashboard *kubermaticv1.KubernetesDashboard `json:"kubernetesDashboard,omitempty"`
```
floreks, line 859: I'd use the same naming as the 2 above fields use if possible, `EnableKubernetesDashboard bool`.

ahmedwaleedmalik: Initially, I went for that but then restructured it. IMO we should try to be as close as possible to the Cluster type(CR), to be consistent. In general, even `EnableOperatingSystemManager` will be converted into a struct in a similar fashion.

But I don't really have any strong opinions on this. For me, it would just make it more consistent and intuitive.

floreks: If other fields are planned to be refactored in the same way then I am fine with it.

/lgtm

ahmedwaleedmalik:
@floreks missing approval 🤔

PR #9333, opened 2022-03-15
add generic CRD validation

SimonTheLeg:
Signed-off-by: Simon Bein <simontheleg@gmail.com>

**What does this PR do / Why do we need it**:

This PR adds two funcs for validating CRs against CRDs
1. Creating a validator from any CRD in yaml/json format
2. Creating a validator for kubermatics CRDs (e.g. cluster, user, ...) 

Both work fully client-side and do not require a kubernetes api-server. The funcs then can be used by individual ValidationFuncs for their respective types.

**Does this PR close any issues?**:

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
none
```

PR #9530, opened 2022-04-04
copy CRDs into the operator chart during release

xrstf:
Since #9333, we intended to create the CRDs in both `pkg/validation/` and `charts/kubermatic-operator/`, but the changes from 9333 made it so that _only_ `pkg/validation/` was updated (because apparently there can be only one `output:crd:dir` flag).

This lead to inconsistencies and the verify-codegen script not working as intended anymore. To solve this, this PR declares `pkg/validation/` to be the now canonical source of CRDs. During releases and for e2e tests, we simply copy the CRDs to their appropriate places. So for the end user / KKP admin, nothing will change.

In the future we might want to change/extend the installer to use the embedded CRDs instead of relying on the YAML files, but that could also have downsides, so we'll see.

```release-note
NONE
```


hack/update-codegen.sh
```diff
@@ -29,5 +29,4 @@ go run sigs.k8s.io/controller-tools/cmd/controller-gen \
   crd \
   object:headerFile=./hack/boilerplate/ce/boilerplate.go.txt \
   paths=./pkg/apis/... \
-  output:crd:dir=./charts/kubermatic-operator/crd/k8c.io \
   output:crd:dir=./pkg/validation/openapi/crd/k8c.io
```
kron4eg, line 32: so this line was a lie? lol, good to know

kron4eg (approved):

PR #9534, opened 2022-04-04
Refactor conformance-tests binary

xrstf:
The goal of this PR is to bring the conformance-tester program in shape to be extended for GitOps approaches. This requires extracting all operations that happen via the API and restructuring the tests. Basically a 90% rewrite. The new binary works exactly like the old one, but it has been split into a couple smaller packages.

```release-note
NONE
```


cmd/conformance-tester/pkg/clients/client_api.go
```diff
@@ -0,0 +1,401 @@
...
+	deleteTimeout := 15 * time.Minute
+	if cluster.Spec.Cloud.Azure != nil {
+		// 15 Minutes are not enough for Azure
+		deleteTimeout = 30 * time.Minute
+	}
+
+	if err := util.JUnitWrapper(
+		"[Kubermatic] Delete cluster",
+		report,
+		func() error {
```
embik: is there any reason we're doing the junit wrapper within the `DeleteCluster` implementation? Wouldn't outside and wrapping around the various implementations be better?

embik (commented):
Quick Q.

cmd/conformance-tester/pkg/clients/client_api.go
xrstf: Moved it.

embik (approved):

PR #9535, opened 2022-04-05
Resolve race condition in metering report configuration tests

pkprzekwas:
We've detected a race condition in our tests and this code is fixing that issue.

```release-note
NONE
```


xrstf (approved):
/approve

PR #9541, opened 2022-04-05
Add GitOps-based e2e test

xrstf:
Now that the conformance-tester can be extended more easily, this PR adds a new KKP client typo to it. This new client will use a regular ctrlruntime client instead of the KKP REST API. This tests the basic GitOps workflow.

```release-note
NONE
```


cmd/conformance-tester/pkg/clients/client_gitops.go
```diff
@@ -0,0 +1,265 @@
...
+	opts *ctypes.Options
+}
+
+var _ Client = &gitopsClient{}
+
+func NewGitopsClient(opts *ctypes.Options) Client {
+	return &gitopsClient{
+		opts: opts,
+	}
+}
```
embik: Nit: Should this client/runner/mode really be named "gitops"? It seems more accurate to say this is the "CRD" client or the "Kubernetes" client. You don't really need to set up a GitOps pipeline to profit from interacting with the Kubermatic CRDs directly instead of using the Kubermatic API (I'm not opposed to highlight this as "enabling GitOps" in marketing fluff).

hack/ci/testdata/kubermatic_gitops.yaml
```diff
@@ -0,0 +1,35 @@
...
+    apiserverReplicas: 1
+  api:
+    replicas: 0
+  ui:
+    replicas: 0
+  # Dex integration
+  auth:
+    tokenIssuer: "http://dex.oauth:5556/dex"
+    issuerRedirectURL: "http://localhost:8000"
+    serviceAccountKey: "__SERVICE_ACCOUNT_KEY__"
```
embik: Is this needed?

cmd/conformance-tester/pkg/clients/client_gitops.go
embik: What I mean is: There's not git involved at all (not more than with the API mode) in this client/mode.

hack/ci/testdata/kubermatic_gitops.yaml
xrstf: It is. I tried to remove it. Next PR will make it optional.

cmd/conformance-tester/pkg/clients/client_gitops.go
xrstf: Tbh, all other names are equally bad in my opinion :D

xrstf: client_kube.go ?

embik: I would prefer it to be `client_kube.go` and `NewKubeClient` or `NewKubeCRClient` or something, yeah. But it's not a hard requirement, I can approve this PR even if it's not renamed.

xrstf: I renamed it.

embik (approved):

PR #9544, opened 2022-04-06
Add HeadlessInstallation feature gate to disable API/UI/Ingress

xrstf:
Now that most of the bare minimum resources work just fine with GitOps and we have a dedicated gitops e2e test, it's now time to strip down the gitops environment even more and allow to disable nginx/dex/cert-manager entirely.

To achieve this, I chose to add a feature gate to KKP. Determining the headless state just based on "are ui.replicas=0 and api.replicas=0?" seems brittle and could lead to unintended consequences. With a dedicated feature gate, the admins can make their intentions absolutely clear and the operator can take the appropriate actions.

Note that this leaves behind a KKP system that has **no user contexts**. Meaning that only cluster-admins with access to master/seed clusters can make use of KKP. User authentication is still a giant TBD epic.

```release-note
A new `HeadlessInstallation` (preview, not yet ready for production) feature flag can be used to disable the KKP UI, API and Ingress, which will also skip installing nginx/Dex/cert-manager. Use this if you intend to only access the master/seed clusters directly and need no user separation.
```

### DIFF
--- a/docs/zz_generated.kubermaticConfiguration.yaml
+++ b/docs/zz_generated.kubermaticConfiguration.yaml
@@ -36,7 +36,8 @@ spec:
       requests:
         cpu: 100m
         memory: 150Mi
-  # Auth defines keys and URLs for Dex.
+  # Auth defines keys and URLs for Dex. These must be defined unless the HeadlessInstallation
+  # feature gate is set, which will disable the UI/API and its need for an OIDC provider entirely.
   auth:
     clientID: kubermatic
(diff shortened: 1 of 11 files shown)


embik:
This looks good to me, do you want me to lgtm it?

ahmedwaleedmalik (approved):
Good stuff 😁 

/approve 🚀 
/hold if other reviews are required

xrstf:
/hold cancel

#yolo

PR #9547, opened 2022-04-06
gracefully handle missing owner in cluster status

xrstf:
The owner email being in the ClusterStatus is one of the last remaining ugly spots in our Cluster CRD. The email is used in exactly 2 spots:

* as a addon template variable (unused in our own templates)
* to bind the email to cluster-admin inside the usercluster

I personally thinnk we should move the owner to the ClusterSpec, but until we can find a solution for this, this PR at least handles the missing owner email gracefully by simply not binding the owner to cluster-admin.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/retest
/approve 🚀

ahmedwaleedmalik:
> I personally thinnk we should move the owner to the ClusterSpec

Agreed, something for the future 👍

PR #9553, opened 2022-04-07
Fix potential race in cleanup of Azure resources

rastislavs:
Fixes occasional issue by deletion of Azure resources that would manifest with the following error message and cluster stuck in deletion:

```
failed cloud provider cleanup: failed to delete resource group "kubernetes-9hv4krpggv": resources.GroupsClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group 'kubernetes-9hv4krpggv' could not be found."
```

Apart from that it also unifies the deletion logic across all Azure resources.

Fixes #

**Documentation**:

```release-note
Fixed potential race in cleanup of Azure resources.
```

### DIFF
--- a/pkg/provider/cloud/azure/resource_group.go
+++ b/pkg/provider/cloud/azure/resource_group.go
@@ -87,7 +87,7 @@ func ensureResourceGroup(ctx context.Context, groupsClient resourcesapi.GroupsCl
 }
 
 func deleteResourceGroup(ctx context.Context, clients *ClientSet, cloud kubermaticv1.CloudSpec) error {
-	// We're doing a Get to see if its already gone or not.
+	// We first check existence of the resource group to see if its already gone or not.
 	// We could also directly call delete but the error response would need to be unpacked twice to get the correct error message.
 	// Doing a get is simpler.
 	resp, err := clients.Groups.CheckExistence(ctx, cloud.Azure.ResourceGroup)
--- a/pkg/provider/cloud/azure/resource_group.go
+++ b/pkg/provider/cloud/azure/resource_group.go
@@ -89,7 +89,6 @@ func ensureResourceGroup(ctx context.Context, groupsClient resourcesapi.GroupsCl
 func deleteResourceGroup(ctx context.Context, clients *ClientSet, cloud kubermaticv1.CloudSpec) error {
 	// We first check existence of the resource group to see if its already gone or not.
 	// We could also directly call delete but the error response would need to be unpacked twice to get the correct error message.
-	// Doing a get is simpler.
 	resp, err := clients.Groups.CheckExistence(ctx, cloud.Azure.ResourceGroup)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


pkg/provider/cloud/azure/resource_group.go
```diff
@@ -90,9 +90,13 @@ func deleteResourceGroup(ctx context.Context, clients *ClientSet, cloud kubermat
 	// We're doing a Get to see if its already gone or not.
 	// We could also directly call delete but the error response would need to be unpacked twice to get the correct error message.
 	// Doing a get is simpler.
```
embik: Can you also update the comment here, since `Get` is not called anymore?

embik (changes requested):

embik:
Seems we have this bad cleanup code in place for all Azure resources (git blame says it was me, but I just copied the code into dedicated files 😛 ). Is it too much to ask for adjustment of the other resources as well?

rastislavs:
> Seems we have this bad cleanup code in place for all Azure resources (git blame says it was me, but I just copied the code into dedicated files 😛 ). Is it too much to ask for adjustment of the other resources as well?

Sure, I can fix that ;)

embik:
Thank you!

pkg/provider/cloud/azure/vnet.go
```diff
@@ -130,6 +130,17 @@ func deleteVNet(ctx context.Context, clients *ClientSet, cloud kubermaticv1.Clou
...
+
+	// We first do Get to check existence of the VNet to see if its already gone or not.
+	// We could also directly call delete but the error response would need to be unpacked twice to get the correct error message.
+	res, err := clients.Networks.Get(ctx, resourceGroup, cloud.Azure.VNetName, "")
+	if err != nil {
+		return err
+	}
+	if isNotFound(res.Response) {
+		return nil
+	}
```
embik: I think this is not going to work because the `Get` method will return an error for everything that isn't HTTP status code `200`: https://github.com/Azure/azure-sdk-for-go/blob/v63.1.0/services/network/mgmt/2021-05-01/network/securitygroups.go#L264-L272

```suggestion
	res, err := clients.Networks.Get(ctx, resourceGroup, cloud.Azure.VNetName, "")
	if err != nil {
		if isNotFound(res.Response) {
			return nil
		}
		return err
	}
	
```

WDYT?

embik (changes requested):

pkg/provider/cloud/azure/vnet.go
rastislavs: yeah, you are right ;)

embik (approved):
/approve

Thanks for fixing this!

embik:
Minor note: Maybe you can update the release note block, but it's not urgent (since it can be updated after the PR merges).

PR #9558, opened 2022-04-07
Upgrade to klog/v2

xrstf:
This PR makes the rbac controller use our default logging stuf (i.e. zap) and upgrades klog to v2 for all the upstream components (our codebase itself doesn't use klog anymore (:partying_face:), but we still want to have `-v` available to toggle the verbosity of kubernetes upstream libraries).

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #9562, opened 2022-04-07
Add missing CRDs for OSM

ahmedwaleedmalik:

These CRDs got mistakenly removed with https://github.com/kubermatic/kubermatic/pull/9530

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/charts/kubermatic-operator/crd/operatingsystemmanager.k8c.io/operatingsystemmanager.k8c.io_operatingsystemconfig.yaml
+++ b/charts/kubermatic-operator/crd/operatingsystemmanager.k8c.io/operatingsystemmanager.k8c.io_operatingsystemconfig.yaml
@@ -1,0 +1,180 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: operatingsystemconfigs.operatingsystemmanager.k8c.io
+spec:
+  group: operatingsystemmanager.k8c.io
+  names:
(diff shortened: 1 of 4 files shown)


xrstf:
Well this throws a giant wrench into the "pkg/validation/ is the canonical source for k8c.io CRDs", if we put files there that are _not_ generated. Hmmm.... I think I would feel better if those OSM CRD's are only in charts/kubermatic-operator/, just like the VPA CRDs already are.

ahmedwaleedmalik:
> I think I would feel better if those OSM CRD's are only in charts/kubermatic-operator/, just like the VPA CRDs already are.

Was hesitant while doing this but then having CRDs for `k8c.io` at two distinct places seemed kinda off. But I think it's fine if we keep external CRDs separate.

ahmedwaleedmalik:
I think it makes sense to name the directory `operatingsystemmanager.k8c.io`, instead of just osm or operatingsystemmanager. @xrstf WDYT?

xrstf:
The installer will not find the CRDs in this directory, the installer only applies what is in k8c.io/ and k8s.io/ -- you would have to adjust the installer.

ahmedwaleedmalik:
> The installer will not find the CRDs in this directory,

Umm, I thought we were relying on Helm for the installation and it's the default `crd` directory for Helm. I should have tested this locally. Let me check on this again.

xrstf (approved):
/approve

PR #9566, opened 2022-04-08
Improve CRDs with better type/field descriptions

embik:

Since we're moving towards GitOps, our CRDs become more important as something we expose to end users. That includes the descriptions associated with fields and types, as they show up in our [CRD reference](https://docs.kubermatic.com/kubermatic/master/references/crds/) and when exploring resources via `kubectl explain`.

This PR updates and adds comments in the type definitions where appropriate and tries to shift focus from development-centric comments to end user explanations.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -396,22 +410,22 @@ type ClusterStatus struct {
 	// +optional
 	Versions ClusterVersionsStatus `json:"versions,omitempty"`
 
-	// UserName contains the name of the owner of this cluster
+	// UserName contains the name of the owner of this cluster.
 	// +optional
 	UserName string `json:"userName,omitempty"`
```
xrstf: Can you please mark this as deprecated as well? It is IIRC not used anywhere and never set by anyone anymore. It got added in The Before Times (#523) without any context or usecase.

The UserEmail is still used for one single thing (binding the email to the cluster-admin ClusterRole inside a usercluster), but that is also of vanishingly importance with GitOps (if you use GitOps, you will most likely setup RBAC in your cluster in your own way, instead of relying on the ClusterStatus magic for one very specific usecase). So that's why I would say to _not_ mark the UserEmail as deprecated.

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -396,22 +410,22 @@ type ClusterStatus struct {
...
-	// UserEmail contains the email of the owner of this cluster
+	// UserEmail contains the email of the owner of this cluster.
```
xrstf, line 423: You could add

> This email address is bound to the cluster-admin ClusterRole inside the usercluster.

But this can be misleading, as this binding happens only once under special circumstances and changing the field in the ClusterStatus *will not* update the binding. So I'm not sure how much we want to document this field (a field I am thinking about moving to the ClusterSpec anyway, maybe, potentially not, not sure).

xrstf: The email is also made available, just like the name, as an addon templating variable, but KKP itself doesn't use it for templates and I doubt(?) that external addons use the fields either.

pkg/apis/kubermatic/v1/cluster.go
embik: Done.

pkg/apis/kubermatic/v1/cluster.go
embik: I added this as second line:

```
// During cluster creation only, this field will be used to bind the `cluster-admin` `ClusterRole` to a cluster owner.
```

WDYT?

xrstf:
Regarding the lint error, I think a nolint comment would be fine in this case.

pkg/apis/kubermatic/v1/common.go
```diff
@@ -18,7 +18,8 @@ package v1
 
 // +kubebuilder:validation:Enum=NodePort;LoadBalancer;Tunneling
 
-// ExposeStrategy is the strategy to expose the cluster with.
+// ExposeStrategy is the strategy used to expose a cluster control plane.
+// Possible values are `NodePort`, `LoadBalancer` or `Tunneling` (requires a feature gate).
```
xrstf, line 22: I am slightly worried that comments like these will quickly get out-of-sync with the surrounding code. Do you have a solution to keep them in-sync?

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -117,125 +119,134 @@ type Cluster struct {
...
+	// Optional: Configures the EventRateLimit admission plugin (if enabled via `useEventRateLimitAdmissionPlugin`)
+	// to create limits on Kubernetes event generation. The EventRateLimit plugin is capable of comparing and rate limiting incoming
+	// `Events` based on several configured buckets.
 	EventRateLimitConfig *EventRateLimitConfig `json:"eventRateLimitConfig,omitempty"`
 
-	// AdmissionPlugins provides the ability to pass arbitrary names of admission plugins to kube-apiserver
-	AdmissionPlugins []string `json:"admissionPlugins,omitempty"`
+	// Optional: Deploys the UserSSHKeyAgent to the user cluster.
+	// If enabled, the agent will be deployed and used to sync user ssh keys attached by users to the cluster.
+	// No SSH keys will be synced after node creation if this is disabled. Once the agent is enabled/disabled
```
xrstf: > Once the agent is enabled/disabled it cannot be changed after cluster creation.

This reads a bit weird. What about a simple "This field is immutable"?

pkg/apis/kubermatic/v1/common.go
embik: I agree this is a problem, I'm not sure how to fix it though. IMHO, our CRDs need to expose this information, as it will otherwise leave users skimming through documentation until they find the valid values (if that is documented in another place at all). We could maybe enforce through validation a particular format on comments for enum fields so that they have to list the possible values, but I'm not entirely sure that is the way to go (or how to do it reliably).

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -117,125 +119,134 @@ type Cluster struct {
...
-	// MLA contains monitoring, logging and alerting related settings for the user cluster.
+	// Optional: MLA contains monitoring, logging and alerting related settings for the user cluster.
 	MLA *MLASettings `json:"mla,omitempty"`
 
-	// +kubebuilder:validation:Enum=docker;containerd
-	// +kubebuilder:default=containerd
-
-	// ContainerRuntime to use, i.e. Docker or containerd. By default containerd will be used.
-	ContainerRuntime string `json:"containerRuntime,omitempty"`
+	// Pause tells that this cluster is currently not managed by the controller.
```
xrstf: _The_ controller? Which controller? This flag affects many controllers.

What about "... is currently not reconciled by KKP." ?

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -117,125 +119,134 @@ type Cluster struct {
...
+	// It indicates that the user needs to do some action to resolve the pause.
+	// +kubebuilder:default=false
+	Pause bool `json:"pause"`
+	// PauseReason is the reason why the cluster is no being managed.
```
xrstf: _not_ being managed.

xrstf: Also please add "This is purely informational and to be set by the user who pauses the cluster" or something along those lines.

pkg/apis/kubermatic/v1/cluster.go
embik: Yeah I just copied this around and didn't update it, but I should. Good catch.

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -496,11 +520,16 @@ const (
...
 type AuditLoggingSettings struct {
-	Enabled      bool              `json:"enabled,omitempty"`
+	// Enabled will enable or disable audit logging.
+	Enabled bool `json:"enabled,omitempty"`
+	// Optional: PolicyPreset can be set to utilize a pre-defined set of audit policy rules.
 	PolicyPreset AuditPolicyPreset `json:"policyPreset,omitempty"`
 }
 
+// EventRateLimitConfig configures the `EventRateLimit` admission plugin.
+// Also see: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#eventratelimit
```
xrstf: Should this be "See also" instead of "Also see"?

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -515,19 +544,19 @@ type EventRateLimitConfigItem struct {
 }
 
 type OPAIntegrationSettings struct {
```
xrstf, line 547: Can we add a comment that at least once in our codebase we explain what "OPA" actually means? Like

> OPAIntegrationSettings configure the usage of the Open Policy Agent (OPA) inside the user cluster.

pkg/apis/kubermatic/v1/cluster.go
embik: Yeah good point. I'll switch to "More info:", which is what the Kubernetes resources seem to use for links like that.

pkg/apis/kubermatic/v1/common.go
xrstf: Well, at least we share the pain. So it's 4 eyes in the future to keep those comments up-to-date.

embik: Keeping documentation in sync is a general problem. We need to remind each other during PR reviews if we catch something that needs docs updates, I guess.

pkg/apis/kubermatic/v1/cluster.go
embik: Sure, will add a comment explaining it as ~~Outer Planets Alliance~~ Open Policy Agent.

pkg/apis/kubermatic/v1/common.go
embik: @xrstf if it's of any comfort to you, adding new values to the CRD needs updating the `+kubebuilder:validation` annotation anyway, I'm banking on people seeing that it needs updating two lines down as well.

PR #9570, opened 2022-04-08
ensure that etcd launcher is enabled if etcd backup/restore is enabled

### LINKED ISSUE #9453: Snapshot restore shouldn't be allowed if etcd launcher is not enabled

PratikDeoghare:

Etcd launcher is required for proper working of snapshot restore. While validating seed we check if it enabled before restore is enabled. 

Fixes #9453 

```release-note
NONE
```

### DIFF
--- a/hack/ci/testdata/kubermatic_backup.yaml
+++ b/hack/ci/testdata/kubermatic_backup.yaml
@@ -33,6 +33,8 @@ spec:
     replicas: 1
   ui:
     replicas: 0
+  featureGates:
+    EtcdLauncher: true
   # Dex integration
   auth:
     tokenIssuer: "http://dex.oauth:5556/dex"
--- a/pkg/webhook/seed/validation.go
+++ b/pkg/webhook/seed/validation.go
@@ -182,6 +182,10 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object, isDelete b
 	}
 
 	if subject.Spec.EtcdBackupRestore != nil {
+		if !v.features.Enabled(features.EtcdLauncher) {
+			return errors.New("EtcdLauncher feature must be enabled for etcd backup and restore")
+		}
+
 		if len(subject.Spec.EtcdBackupRestore.Destinations) == 0 {
 			return errors.New("invalid etcd backup configuration: must define at least one backup destination")
 		}


hack/ci/testdata/kubermatic_backup.yaml
```diff
@@ -33,6 +33,8 @@ spec:
     replicas: 1
   ui:
     replicas: 0
+  featureGates:
+    EtcdLauncher: true
```
embik, line 37: This is going to be a problem because `pre-kubermatic-etcd-launcher-e2e` is specifically built to test the "plain" `etcd` to `etcd-launcher` migration. Applying this change, all clusters created in the e2e test will be created with `etcd-launcher` by default and we won't notice if a PR breaks the migration path. See how the test immediately sees the etcd cluster as `etcd-launcher` enabled in the latest e2e run:

```
etcd_test.go:294: enabling etcd-launcher...
etcd_test.go:538: etcd cluster became healthy after 13.249266ms.
etcd_test.go:561: etcd cluster is running in strict TLS mode after 13.18454ms.
```

Perhaps backups need to have their separate e2e tests.

embik (changes requested):

hack/ci/testdata/kubermatic_backup.yaml
embik: Or maybe the e2e test needs two `Seeds` (not sure that is possible though). Maybe the test can flip the feature gate on and off.

pkg/webhook/seed/validation.go
```diff
@@ -182,6 +182,10 @@ func (v *validator) validate(ctx context.Context, obj runtime.Object, isDelete b
 	}
 
 	if subject.Spec.EtcdBackupRestore != nil {
+		if !v.features.Enabled(features.EtcdLauncher) {
+			return errors.New("EtcdLauncher feature must be enabled for etcd backup and restore")
+		}
```
lsviben, line 187: I think Clusters can have `etcd-launcher` enabled even without it being enabled for the seed. So in that case this wont work properly.

Why not just check somewhere here, and skip the restore if the cluster doesnt have `etcd-launcher` enabled?
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go#L120-L145

PratikDeoghare: > Why not just check somewhere here, and skip the restore .... 

This (etcd backup/restore) is a feature that depends upon other feature(etcd-launcher) being enabled. So we decided to validate at the features level.

PratikDeoghare: > I think Clusters can have etcd-launcher enabled even without it being enabled for the seed. So in that case this wont work properly.

I see. Hmm... could we make it not possible? or it is by design?

hack/ci/testdata/kubermatic_backup.yaml
PratikDeoghare: I see. Let me check the test. Maybe  it can be made immune to such changes.

pkg/webhook/seed/validation.go
embik: Individual clusters having the `etcdLauncher` feature flag enabled while the Seed or KubermaticConfiguration do not is by design. You can, for example, [disable etcd-launcher on a global level](https://docs.kubermatic.com/kubermatic/v2.20/cheat_sheets/etcd/etcd-launcher/#disabling-etcd-launcher), but it cannot be rolled back on existing clusters, so that would be a scenario (`Seed` does not have the feature flag anymore, but `Clusters` associated with that `Seed` do from before) where this way of validating it might become problematic.

PratikDeoghare: Gotcha. Thanks!

PratikDeoghare: @lsviben @embik Ptal. https://github.com/kubermatic/kubermatic/pull/9633

PratikDeoghare:
Closing. We decided in favor of https://github.com/kubermatic/kubermatic/pull/9633

PR #9571, opened 2022-04-08
Add options to specify multiple IP ranges from which access to NodePort services is allowed

### LINKED ISSUE #9442: Support multiple NodePort allowed IP ranges
For AWS, Azure, GCP and OpenStack cloud providers, the CIDR range from which NodePort communication is allowed can be configured in the provider cloud spec (`NodePortsAllowedIPRange`).

However it is possible to configure only single CIDR range. This is a problem especially for dual-stack clusters, where normally at least one CIDR for each address family would be needed.

The aim of this story is to allow configuring any number of allowed CIDR ranges for NodePorts.

rastislavs:
This PR adds options to specify multiple IP ranges from which access to NodePort services is allowed in AWS, Azure, GCP and OpenStack. Although single allowed IP range is already configurable, multiple ranges are needed especially for dual-stack clusters, where we need at least one IP range for each address family.

To maintain backward compatibility, the existing `nodePortsAllowedIPRange` options are preserved and a new option `nodePortsAllowedIPRanges` is added for each affected cloud provider. Both options can be used at the same time.

Handling of these ranges is also refactored and unified across cloud providers.

Fixes #9442 

**Documentation**:

```release-note
Added nodePortsAllowedIPRanges option to specify multiple IP ranges from which access to NodePort services is allowed in AWS, Azure, GCP and OpenStack.
```


PratikDeoghare (approved):

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -893,11 +898,18 @@ type AWSCloudSpec struct {
...
+	InstanceProfileName string `json:"instanceProfileName"`
+	SecurityGroupID     string `json:"securityGroupID"`
+	// A CIDR range that will be used to allow access to the node port range in the security group to. Only applies if
+	// the security group is generated by KKP and not preexisting.
+	// If NodePortsAllowedIPRange nor NodePortsAllowedIPRanges is set, the node port range can be accessed from anywhere.
 	NodePortsAllowedIPRange string `json:"nodePortsAllowedIPRange,omitempty"`
+	// Optional: CIDR ranges that will be used to allow access to the node port range in the security group to. Only applies if
+	// the security group is generated by KKP and not preexisting.
+	// If NodePortsAllowedIPRange nor NodePortsAllowedIPRanges is set,  the node port range can be accessed from anywhere.
+	NodePortsAllowedIPRanges *NetworkRanges `json:"nodePortsAllowedIPRanges,omitempty"`
```
ahmedwaleedmalik, line 912: @rastislavs is it possible to have some validation for this? Some `+kubebuilder:validation` marker, like https://github.com/ahmedwaleedmalik/kubermatic/blob/master/pkg/apis/kubermatic/v1/datacenter.go#L30

rastislavs: You mean for the `NetworkRanges` type that I am reusing here?
I guess that would have to be something complex as:
```
// +kubebuilder:validation:Pattern="((^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))/([0-9]|[1-2][0-9]|3[0-2])$)|(^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))/([0-9]|[0-9][0-9]|1[0-1][0-9]|12[0-8])$))"
```

But I am not sure if that is doable without modifying the `NetworkRanges` type - which I guess could break backward compatibility?

rastislavs: @ahmedwaleedmalik it would look like this. Do you think this would be an allowed API change?

```diff
--- a/pkg/apis/kubermatic/v1/cluster.go
+++ b/pkg/apis/kubermatic/v1/cluster.go
@@ -692,9 +692,12 @@ type MachineNetworkingConfig struct {
 	DNSServers []string `json:"dnsServers"`
 }

+// +kubebuilder:validation:Pattern="((^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))/([0-9]|[1-2][0-9]|3[0-2])$)|(^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))/([0-9]|[0-9][0-9]|1[0-1][0-9]|12[0-8])$))"
+type CIDRBlock string
+
 // NetworkRanges represents ranges of network addresses.
 type NetworkRanges struct {
-	CIDRBlocks []string `json:"cidrBlocks"`
+	CIDRBlocks []CIDRBlock `json:"cidrBlocks"`
 }
```

ahmedwaleedmalik: > But I am not sure if that is done without modifying the NetworkRanges type - which I guess would break backward compatibility

Yes, that is correct. Maybe a different type for this case to not break backward compatibility? I know it's code duplication but at least won't break anything. But let's consider this as a "nit" and I'll leave it up to you if you want to add this validation.

rastislavs: If it was just the data type, I would have no problem with duplicating it. But as you can see, `NetworkRanges` have some util methods, that would need to be duplicated as well. I would like to avoid duplicating those.

In any case, you reminded me that I forgot to add validation in the validation webhook. I'll do that. Then, if the change in the `NetworkRanges` type is not allowed, I would rely only on webhook validation.

ahmedwaleedmalik: > But as you can see, NetworkRanges have some util methods, that would need to be duplicated as well. I would like to avoid duplicating those.

Yes, that's the main reason why I'm divided on this one because it would overcomplicate a lot of things. 

Let's just settle with the webhook validation please.

pkg/validation/cluster.go
```diff
@@ -607,6 +628,13 @@ func validateAzureCloudSpec(spec *kubermaticv1.AzureCloudSpec) error {
 			return err
 		}
 	}
+	if spec.NodePortsAllowedIPRanges != nil {
```
PratikDeoghare: nit: this block could be a function 
(but it is also nice to be able to see the whole code right here)

PratikDeoghare (approved):

pkg/validation/cluster.go
rastislavs: good point

PR #9572, opened 2022-04-08
make node-resource-documenter work again, move it to codegen/

xrstf:
The *node-resource-documenter* did, as its name implied, create an overview over the resource requests/limits of the KKP addons...

It didn't work anymore for ages (it was added in #4264, by the way) and its output also, as far as I can see, was never part of any documentation anywhere.

This PR cleans up the tool, renames it to `codegen/addon-resources` and makes it spit out a JSON file, which is then synced into our docs repo, just like the addondata.go or seed.yaml files. A template in the documentation can then render this content.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve 🚀

PR #9585, opened 2022-04-12
fix expired JWT token

zreigz:
**What does this PR do / Why do we need it**: replace the test JWT token. Generated here https://github.com/kubermatic/kubermatic/blob/master/pkg/serviceaccount/jwt_test.go#L49

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/handler/test/helper.go
+++ b/pkg/handler/test/helper.go
@@ -129,7 +129,7 @@ const (
 	TestSeedDatacenter = "us-central1"
 	// TestServiceAccountHashKey authenticates the service account's token value using HMAC.
 	TestServiceAccountHashKey = "eyJhbGciOiJIUzI1NeyJhbGciOiJIUzI1N"
-	// TestFakeToken signed JWT token with fake data. It will expire after 3 years from 12-04-2022. To generate new token use kubermatic/pkg/serviceaccount/jwt_test.go
+	// TestFakeToken signed JWT token with fake data. It will expire after 3 years from 12-04-2022. To generate new token use kubermatic/pkg/serviceaccount/jwt_test.go.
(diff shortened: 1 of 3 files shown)


pkg/handler/test/helper.go
```diff
@@ -130,7 +130,7 @@ const (
 	// TestServiceAccountHashKey authenticates the service account's token value using HMAC.
 	TestServiceAccountHashKey = "eyJhbGciOiJIUzI1NeyJhbGciOiJIUzI1N"
 	// TestFakeToken signed JWT token with fake data.
```
xrstf: Please extend this comment with the token's expiry

zreigz: done

xrstf (approved):
Thank you!

/approve

pkg/handler/test/helper.go
```diff
@@ -129,8 +129,8 @@ const (
 	TestSeedDatacenter = "us-central1"
 	// TestServiceAccountHashKey authenticates the service account's token value using HMAC.
 	TestServiceAccountHashKey = "eyJhbGciOiJIUzI1NeyJhbGciOiJIUzI1N"
-	// TestFakeToken signed JWT token with fake data.
-	TestFakeToken = "eyJhbGciOiJIUzI1NiJ9.eyJlbWFpbCI6IjEiLCJleHAiOjE2NDk3NDg4NTYsImlhdCI6MTU1NTA1NDQ1NiwibmJmIjoxNTU1MDU0NDU2LCJwcm9qZWN0X2lkIjoiMSIsInRva2VuX2lkIjoiMSJ9.Q4qxzOaCvUnWfXneY654YiQjUTd_Lsmw56rE17W2ouo"
+	// TestFakeToken signed JWT token with fake data. It will expire after 3 years from 12-04-2022. To generate new token use kubermatic/pkg/serviceaccount/jwt_test.go
```
embik: Linter says this needs a period at the end to pass the linting job.

embik:
```suggestion
	// TestFakeToken signed JWT token with fake data. It will expire after 3 years from 12-04-2022. To generate new token use kubermatic/pkg/serviceaccount/jwt_test.go.
```

PR #9586, opened 2022-04-12
allowing arbitrary name for the cluster + adding length validation

### LINKED ISSUE #7004: Investigate possible changes in the API

hbernardo:
- Changes the machine deployment name generation to use the cluster generated id instead of the name chosen by the user, so it won't break the machine deployment when receiving arbitrary name for the cluster.
- Limits the accepted cluster name length (max 100 characters), so it won't break any other part of the system.

Fixes #7004

**Documentation**:

```release-note
Auto generated names for the machine deployments now contain the cluster id as prefix, instead of the cluster human readable name.
```

### DIFF
--- a/pkg/handler/common/cluster.go
+++ b/pkg/handler/common/cluster.go
@@ -1048,19 +1048,19 @@ func ConvertInternalClusterToExternal(internalCluster *kubermaticv1.Cluster, dat
 
 func ValidateClusterSpec(updateManager common.UpdateManager, body apiv1.CreateClusterSpec) error {
 	if body.Cluster.Spec.Cloud.DatacenterName == "" {
-		return fmt.Errorf("cluster datacenter name is empty")
+		return errors.New("cluster datacenter name is empty")
 	}
 	if body.Cluster.ID != "" {
-		return fmt.Errorf("cluster.ID is read-only")
+		return errors.New("cluster.ID is read-only")
 	}
 	if !ClusterTypes.Has(body.Cluster.Type) {
 		return fmt.Errorf("invalid cluster type %s", body.Cluster.Type)
 	}
 	if body.Cluster.Spec.Version.Semver() == nil {
-		return fmt.Errorf("invalid cluster: invalid cloud spec \"Version\" is required but was not specified")
+		return errors.New("invalid cluster: invalid cloud spec \"Version\" is required but was not specified")
 	}
 	if len(body.Cluster.Name) > 100 {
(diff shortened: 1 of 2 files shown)


zreigz:
this test needs some adjustment `k8c.io/kubermatic/v2/pkg/controller/user-cluster-controller-manager/application-installation-controller/controller_integration_test.go:46`

pkg/resources/machine/machinedeployment.go
```diff
@@ -49,7 +49,7 @@ func Deployment(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc *kubermati
 	} else {
 		// GenerateName can be set only if Name is empty to avoid confusing error:
 		// https://github.com/kubernetes/kubernetes/issues/32220
-		md.GenerateName = fmt.Sprintf("%s-worker-", c.Spec.HumanReadableName)
+		md.GenerateName = fmt.Sprintf("%s-worker-", c.Name)
```
embik, line 52: Isn't this going to update existing `MachineDeployments`? What will happen to them?

hbernardo: Hi @embik . This is only for new Machine Deployments with empty name. Before this PR, it was using the cluster name to generate the MD name and, because of that, when the cluster name had not accepted characters (e.g. space), the auto generated name was invalid and it wasn't possible to create the new machine deployment.

zreigz: it's only autogenerated name with the KKP cluster name as a prefix The KKP human-readable name  should support an arbitrary name and shouldn't be used here

pkg/handler/common/cluster.go
```diff
@@ -1059,6 +1059,9 @@ func ValidateClusterSpec(updateManager common.UpdateManager, body apiv1.CreateCl
 	if body.Cluster.Spec.Version.Semver() == nil {
 		return fmt.Errorf("invalid cluster: invalid cloud spec \"Version\" is required but was not specified")
 	}
+	if len(body.Cluster.Name) > 100 {
+		return fmt.Errorf("invalid cluster name: too long (greater than 100 characters)")
```
xrstf: This can simply be `errors.New` (same for the error above)

pkg/resources/machine/machinedeployment.go
embik: Gotcha, thanks for explaining! it seems this is only called when converting an incoming API call to a `MachineDeployment` spec.

PR #9588, opened 2022-04-13
applications installation controller: reconcile target namespace

### LINKED ISSUE #8376: Reconcile target namespace
**Acceptance Criteria**
* namespace creation should be independent of the selected method so it is the same for all methods (=> e.g. we do not plan to use helms automatic create-namespace feature)

vgramer:
This pr reconciles namespace if `ApplicationInstallation.Spec.Namespace.Create = true`. Meaning:

- create the namespace if it does not exist
- ensure the namespace has desired labels and annotations

Fixes #8376 

/assign @SimonTheLeg 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/applications/installer.go
+++ b/pkg/applications/installer.go
@@ -18,11 +18,16 @@ package applications
 
 import (
 	"context"
+	"fmt"
 
 	"go.uber.org/zap"
 
 	appkubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/apps.kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/applications/installer_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	fakectrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestApplicationManager_applyNamespace_with_create_ns(t *testing.T) {
```
SimonTheLeg: I have to say I am not a fan of the multi-mixed Camel and Snake case here. I see that we do it sometimes in other tests as well. However in these cases (e.g. https://github.com/kubermatic/kubermatic/blob/master/pkg/api/v1/types_test.go#L29) it's usually only a single underscore. This would be an okay compromise for me :) 
So TestApplicationManager_ApplyNamespaceWithCreateNs and so on

pkg/applications/installer.go
```diff
@@ -34,3 +39,79 @@ type ApplicationInstaller interface {
...
+}
+
+// Delete uninstalls the application and deletes the namespace where the application was installed if necessary.
+func (a *ApplicationManager) Delete(ctx context.Context, log *zap.SugaredLogger, seedClient ctrlruntimeclient.Client, userClient ctrlruntimeclient.Client, applicationInstallation *appkubermaticv1.ApplicationInstallation) error {
+	// todo logic to uninstall application
+	return a.deleteNamespace(ctx, log, applicationInstallation, userClient)
+}
+
+// applyNamespace ensures namespace is created and has desired labels and annotations if applicationInstallation.Spec.Namespace.Create flag is set.
+func (a *ApplicationManager) applyNamespace(ctx context.Context, log *zap.SugaredLogger, applicationInstallation *appkubermaticv1.ApplicationInstallation, userClient ctrlruntimeclient.Client) error {
```
SimonTheLeg: I would rename this func to something like reconcileNamespace, because it contains the checking logic if the namespace should be created in the first place. And if I see it correctly, if "Create" is set to false, this func will not apply anything.

Alternatively the checking logic could also be moved into the `(*ApplicationManager) Apply`, but I do see how this makes the tests unnecessarily clunky.

SimonTheLeg (changes requested):
only minor nitpicks

pkg/applications/installer.go
vgramer: For the naming, I was hesitating, `reconcileNamespace`is good to me :)

> Alternatively the checking logic could also be moved into the (*ApplicationManager) Apply, but I do see how this makes the tests unnecessarily clunky.
 yes, moreover this function will also contain the logic to download the source and install the application with helm. So I think it's better to split logic into dedicated functions for more clarity and easy testing.

SimonTheLeg (approved):

Issue #9584, opened 2022-04-12
Owner not set correctly for clusters created from clusterTemplate


wurbanski:
### What happened?
After migrating from kubermatic v2.19 to v2.20, creating a cluster from clusterTemplate no longer works. 
usercluster-controller crashloops, owner is not set correctly on the Cluster object.

logs excerpt:
```
{"level":"error","time":"2022-04-12T08:27:53.731Z","logger":"cluster_template_controller","caller":"cluster-template-controller/controller.go:111","msg":"ReconcilingError","worker-name":"","request":"/bfm7gffsb6-mtncfsnsm6-2h8hpb6jpt","error":"failed to set cluster status: Cluster.kubermatic.k8c.io \"qthhl64nsp\" not found"}
{"level":"error","time":"2022-04-12T08:27:53.731Z","logger":"controller.cluster_template_controller","caller":"controller/controller.go:317","msg":"Reconciler error","name":"bfm7gffsb6-mtncfsnsm6-2h8hpb6jpt","namespace":"","error":"failed to set cluster status: Cluster.kubermatic.k8c.io \"qthhl64nsp\" not found"}
```

### What should happen?
User cluster should be created from ClusterTemplate without any issues.

### How to reproduce?
-  create a ClusterTemplate from the UI. 
-  create a cluster from clusterTemplate.

### Environment

- KKP Version: v2.20
- Shared or separate master/seed clusters?: n/a
- Master/seed cluster provider: openstack

PR #9593, opened 2022-04-13
wait for new cluster objects to appear in cache before attempting to …

xrstf:
This PR was triggered by #9584, before I realized that in master we already fixed the issue via #9122 and should just backport that to 2.20.

However, this PR applies the same fix to more places in our codebase, which had a very slight chance of running into a race condition as well.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #9601, opened 2022-04-13
Add validating webhooks for ApplicationDefinition and ApplicationInstallation

### LINKED ISSUE #8368: ApplicationDefinitions are validated before create/update

### LINKED ISSUE #8385: ApplicationInstallations are validated on create/update
**Acceptance Criteria**
* `Namespace.Create`field is a **readonly** boolean
* `Namespace.Name` field is **readonly** valid name (ie according to k8s resource naming)
* `ApplicationRef` field is **readonly**
* `ApplicationRef.name ` field is valid name (ie according to k8s resource naming)

**Implementation Hints**
* we think it makes sense to build a controller which is the receiver of a webhook. The controller then would be part of the user-controller-manager

ahmedwaleedmalik:

Add validating webhooks for ApplicationDefinition and ApplicationInstallation.

Fixes #8385 #8368

**Documentation**:

```release-note
NONE
```


cmd/user-cluster-webhook/README.md
```diff
@@ -0,0 +1,5 @@
+# User Cluster Webhook
```
SimonTheLeg, line 1: This was so helpful!

pkg/controller/user-cluster-controller-manager/application-installation-controller/controller_test.go
```diff
@@ -56,8 +56,8 @@ func TestEnqueueApplicationInstallation(t *testing.T) {
 					genApplicationInstallation("appInstallation-3", "app-def-1", "1.0.0")).
 				Build(),
 			expectedReconcileRequests: []reconcile.Request{
-				{NamespacedName: types.NamespacedName{Name: "appInstallation-1"}},
-				{NamespacedName: types.NamespacedName{Name: "appInstallation-3"}},
+				{NamespacedName: types.NamespacedName{Name: "appInstallation-1", Namespace: "apps"}},
```
SimonTheLeg: nitpick: I think I would prefer to use a variable here like we do in the integration tests.

```
const (
	ApplicationNamespace = "apps"
)
```

And then use a `// nolint:` flag so it passes linting

pkg/controller/operator/common/resources.go
```diff
@@ -55,6 +55,9 @@ const (
 	// UserAdmissionWebhookName is the name of the validating webhook for Users.
 	UserAdmissionWebhookName = "kubermatic-users"
 
+	// ApplicationDefinitionAdmissionWebhookName is the name of the validating webhook for ApplicationDefnition.
+	ApplicationDefinitionAdmissionWebhookName = "kubermatic-application-definitions"
```
SimonTheLeg, line 59: q: Is there a specific reason why the webhook can be found under two different names? 

1. "kubermatic-application-definitions"
2. "applicationdefinitions.apps.kubermatic.k8c.io" (from https://github.com/kubermatic/kubermatic/blob/a28e749d6489b9a61b0accf1fb236d04e287c77a/pkg/controller/operator/master/resources/kubermatic/webhooks.go#L214)

Or could we consolidate the two?

pkg/controller/operator/seed/resources/kubermatic/webhooks.go
```diff
@@ -367,3 +368,56 @@ func OperatingSystemProfileValidatingWebhookConfigurationCreator(ctx context.Con
 		}
 	}
 }
+
+func ApplicationDefinitionValidatingWebhookConfigurationCreator(ctx context.Context, cfg *kubermaticv1.KubermaticConfiguration, client ctrlruntimeclient.Client) reconciling.NamedValidatingWebhookConfigurationCreatorGetter {
```
SimonTheLeg: For me it is fine that this code is duplicated for **master**/.../webhooks.go and **seed**/.../webhooks.go and I think extracting the few lines into a separate package might be a bit overkill. Plus I guess it ties in with with Gos "duplication is far cheaper than the wrong abstraction" philosophy. But I would recommend that in this case we write a short comment on both funcs that their functionality is mirrored, so future devs are aware of it.

pkg/validation/applicationinstallation.go
```diff
@@ -0,0 +1,102 @@
...
+	specPath := field.NewPath("spec")
+	allErrs := field.ErrorList{}
+
+	// Validation for new ApplicationInstallation Spec
+	if errs := ValidateApplicationInstallationSpec(ctx, client, newAI.Spec); len(errs) > 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
+	// Validate .Spec.Namespace.Create for immutability
+	if oldAI.Spec.Namespace.Create != newAI.Spec.Namespace.Create {
```
SimonTheLeg: If I read the code correctly, ValidateImmutableField automatically does all of the validation for you. Therefore I think the the first if statement can be skipped.

Same would apply for the subsequent occurences of this pattern.

pkg/validation/applicationinstallation.go
```diff
@@ -0,0 +1,102 @@
...
+	if oldAI.Spec.Namespace.Name != newAI.Spec.Namespace.Name {
+		allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(
+			newAI.Spec.Namespace.Name,
+			oldAI.Spec.Namespace.Name,
+			specPath.Child("namespace", "name"),
+		)...)
+	}
+
+	// Validate .Spec.ApplicationRef for immutability
+	if oldAI.Spec.ApplicationRef != newAI.Spec.ApplicationRef {
```
SimonTheLeg: Additionally I think this if statement will only be the same as `apiequality.Semantic.DeepEqual` (which is being used by ValidateImmutableField below) as long as ApplicationRef only contains primitive types.

pkg/validation/applicationinstallation_test.go
```diff
@@ -0,0 +1,205 @@
...
+				Spec: func() appkubermaticv1.ApplicationInstallationSpec {
+					spec := ai.Spec.DeepCopy()
+					spec.Namespace.Name = invalidResource
+					return *spec
+				}(),
+			},
+			expectedError: `[spec.namespace.name: Invalid value: "default": field is immutable]`,
+		},
+		{
+			name: "Update ApplicationInstallation Failure - .ApplicationRef.Name is immutable",
```
SimonTheLeg, line 147: cc: @vgramer If my memory serves me correctly, when a user wants to upgrade their Application, we would delete the old ApplicationInstallation and create a new one, instead of patching the `.ApplicationRef.Version` field, right?

If this is still true, then I think we should add a testcase for `.ApplicationRef.Version is immutable`

ahmedwaleedmalik: `ApplicationRef` on a whole is immutable. Let me add another test for that use case 👍

pkg/validation/applicationinstallation.go
ahmedwaleedmalik: You are absolutely right, I missed out on that part in haste.

pkg/controller/operator/common/resources.go
ahmedwaleedmalik: `kubermatic-application-definitions` is the name of the `Validatingwebhookconfiguration` resource that will be created in the cluster. Whereas `applicationdefinitions.apps.kubermatic.k8c.io` is the name of the webhook in that resource i.e. `webhooks[].name`. The name of the webhook within the resource should be an FQDN pointing to the API resource against which the admission configuration is created, for example, applicationdefinitions.apps.kubermatic.k8c.io.

pkg/controller/operator/seed/resources/kubermatic/webhooks.go
ahmedwaleedmalik: I wanted them to be isolable resources but now that I think of it, it's better to have them in a separate shared package. We already do this for certain shared resources so it's completely fine to follow the same here.

pkg/controller/user-cluster-controller-manager/resources/resources/applications/webhook.go
```diff
@@ -0,0 +1,79 @@
...
+				{
+					Name:                    "applicationinstallations.apps.kubermatic.k8c.io", // this should be a FQDN
+					AdmissionReviewVersions: []string{admissionregistrationv1.SchemeGroupVersion.Version, admissionregistrationv1beta1.SchemeGroupVersion.Version},
+					MatchPolicy:             &matchPolicy,
+					FailurePolicy:           &failurePolicy,
+					SideEffects:             &sideEffects,
+					TimeoutSeconds:          pointer.Int32Ptr(30),
+					ClientConfig: admissionregistrationv1.WebhookClientConfig{
+						CABundle: triple.EncodeCertPEM(caCert),
+						URL:      &url,
```
vgramer, line 56: Any reason for not using an `admissionregistrationv1.ServiceReference` ?

pkg/controller/user-cluster-controller-manager/application-installation-controller/controller_test.go
```diff
@@ -147,7 +151,8 @@ func genApplicationDefinition(name string) *appkubermaticv1.ApplicationDefinitio
 func genApplicationInstallation(name string, applicationDefName string, appVersion string) *appkubermaticv1.ApplicationInstallation {
 	return &appkubermaticv1.ApplicationInstallation{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			Name:      name,
+			Namespace: applicationNamespace,
```
vgramer, line 155: `genApplicationInstallation` is also used in integration tests where applications are expected to be created under 
`pkg/controller/user-cluster-controller-manager/application-installation-controller/suite_test.go::ApplicationNamespace` namespace. Even if it's the same value as `applicationNamespace` i think it's better to have only one constant.

pkg/webhook/application/applicationinstallation/validation/validation.go
```diff
@@ -0,0 +1,99 @@
...
+		if err := h.decoder.Decode(req, ad); err != nil {
+			return webhook.Errored(http.StatusBadRequest, err)
+		}
+		if err := h.decoder.DecodeRaw(req.OldObject, oldAD); err != nil {
+			return webhook.Errored(http.StatusBadRequest, err)
+		}
+		allErrs = append(allErrs, validation.ValidateApplicationInstallationUpdate(ctx, h.client, *ad, *oldAD)...)
+
+	case admissionv1.Delete:
+		// NOP we always allow delete operations
```
vgramer, line 88: As we only validate ApplicationInstallation for creating and updating operations, maybe we should set up the webhook configuration to match only these operations? it will reduce a little bit the chattiness between the api-server and the webhook. WDYT?

pkg/validation/applicationinstallation_test.go
vgramer: I think to be production-ready, we have to allow users to update and application version because deleting applicationInstallation means uninstalling the application and losing data if the workload was stateful which is not acceptable for upgrading a version of an application.

However, IIRC, when we write a proposal we said in the first version we make it immutable the time to dig a little bit about how to make difference between an admin deleting an application version from applicationDefinition and the user referencing a version that does not exist, and maybe some other corner cases around the upgrade...

vgramer: I think to be production-ready, we have to allow users to update and application version because deleting applicationInstallation means uninstalling the application and losing data if the workload was stateful which is not acceptable for upgrading a version of an application.

However, IIRC, when we write a proposal we said in the first version we make it immutable the time to dig a little bit about how to make difference between an admin deleting an application version from applicationDefinition and the user referencing a version that does not exist, and maybe some other corner cases around the upgrade...

pkg/controller/user-cluster-controller-manager/resources/resources/applications/webhook.go
ahmedwaleedmalik: Our pod that encapsulates the webhook is running in the cluster namespace in seed. Whereas this resource i.e. `ValidatingWebhookConfiguration` will be created in the user cluster. The only way they can communicate is via the URL.

pkg/webhook/application/applicationinstallation/validation/validation.go
ahmedwaleedmalik: Good point 👍

pkg/validation/applicationinstallation_test.go
ahmedwaleedmalik: I think upgrading the version would be a bit complicated. My suggestion would be to work on that in a separate PR. We can update these webhooks later based on how we want to approach upgrades.

vgramer: I am ok with that :)

pkg/controller/user-cluster-controller-manager/resources/resources/applications/webhook.go
vgramer: oh right! my bad

vgramer (approved):
/lgtm

@SimonTheLeg if it's goog for you, I let you approve the PR

pkg/controller/user-cluster-controller-manager/resources/resources/applications/webhook.go
ahmedwaleedmalik: No problem 👍

pkg/validation/applicationinstallation_test.go
SimonTheLeg: sounds good to me

SimonTheLeg (approved):

Issue #9349, opened 2022-03-16
Add dual-stack support to KKP UI Network Configuration


rastislavs:
The main aim of this task is to add dual-stack support to KKP UI Network Configuration. Apart from that, it also aims to enhance the Network Configuration UI experience in general.

Basic KKP users may not care about all networking details, so these can be hidden in the UI by default. The only options visible by default should be the CNI + CNI version and (the new) "IPv4" vs. "Dual Stack" option:

![image.png](https://images.zenhubusercontent.com/5f9feacdd87c6f0418bee1e8/b3069a59-3e05-4dc8-ad88-bc0deaade1e8)

When user clicks on "Additional Network Configuration", more configuration options should be available. As an enhancement to the current state, they should all be pre-populated with the default values, so that users would know what exactly will be applied in their cluster:

![Screenshot 2022-04-13 at 16.41.32.png](https://images.zenhubusercontent.com/5f9feacdd87c6f0418bee1e8/e267b5e8-1744-4e6c-a5ec-bf2286c5cd80)

(“Node CIDR Mask Size” and “Node Local DNS Cache” are new config options not yet available in the UI, these should be added as well)

When "Dual Stack" option is selected, some of the configuration options needs to be duplicated for each address family (IPv4 + IPv6), as shown below:

![image.png](https://images.zenhubusercontent.com/5f9feacdd87c6f0418bee1e8/ccfc9406-b137-462b-bc28-2591b06c7925)

Multiple PODs CIDR & Services CIDR, IPv4 & IPv6 Node CIDR Mask Size and Multiple Allowed IP Ranges for NodePorts can be already configured via cluster API of KKP.

API endpoints for loading default values for given cloud provider and CNI will be added as part of https://github.com/kubermatic/kubermatic/issues/9348

cschieder:
Good draft, really like the simplification and prefilling, minor tweaks:
- I think we can say `Dual Stack (IPv4 and IPv6)` or `IPv4 and IPv6 (Dual Stack)` in the radio button to make it clearer
- in the dual stack additional settings let's add a little headline, `IPv4` and `IPv6` respectively so it's immediately apparent that these are the settings for each and you can change them separately
- instead of "additional" we should probably say "advanced" as the settings are not optional add ons but an elaboration/advanced setting of the radio button options

rastislavs:
[...]

PR #9612, opened 2022-04-14
Add API endpoint for retrieving default networking configuration

### LINKED ISSUE #9348: API endpoints for default pod/service CIDRs
Dual-stack can be enabled by providing both IPv4 and IPv6 pods and services CIDRs in KKP API when creating new clusters.

In order to be able to integrate dual-stack with the KKP UI, we need to add API endpoints that can be used to retrieve default values for individual CIDRs (and other networking parameters) for individual address families.

rastislavs:
Adds API endpoint `/api/v2/providers/{provider_name}/cni/{cni_plugin_type}/networkdefaults` to allow pre-filling of network configuration in KKP UI as described in https://github.com/kubermatic/kubermatic/issues/9349

Fixes #9348

**Documentation**:

```release-note
NONE
```


pkg/handler/v2/networkdefaults/networkdefaults.go
```diff
@@ -0,0 +1,101 @@
...
+// Validate validates getNetworkDefaultsReq request.
+func (r getNetworkDefaultsReq) Validate() error {
+	if r.ProviderName == "" {
+		return fmt.Errorf("the provider name cannot be empty")
+	}
+	if !kubermaticv1.IsProviderSupported(r.ProviderName) {
+		return fmt.Errorf("invalid provider name %s", r.ProviderName)
+	}
+	if r.CNIPluginType == "" {
+		return fmt.Errorf("the CNI plugin type cannot be empty")
```
PratikDeoghare:
```suggestion
		return fmt.Errorf("CNI plugin type cannot be empty")
```

pkg/handler/v2/networkdefaults/networkdefaults.go
```diff
@@ -0,0 +1,101 @@
...
+	CNIPluginType string `json:"cni_plugin_type"`
+}
+
+// Validate validates getNetworkDefaultsReq request.
+func (r getNetworkDefaultsReq) Validate() error {
+	if r.ProviderName == "" {
+		return fmt.Errorf("the provider name cannot be empty")
+	}
+	if !kubermaticv1.IsProviderSupported(r.ProviderName) {
+		return fmt.Errorf("invalid provider name %s", r.ProviderName)
```
PratikDeoghare:
```suggestion
		return fmt.Errorf("unsupported provider: %q", r.ProviderName)
```

pkg/api/v2/types.go
```diff
@@ -1353,3 +1353,29 @@ type CNIVersions struct {
 	// Versions represents the list of the CNI Plugin versions that are supported
 	Versions []string `json:"versions"`
 }
+
+// NetworkDefaults contains cluster network default settings.
+// swagger:model NetworkDefaults
+type NetworkDefaults struct {
```
PratikDeoghare, line 1359: I wonder if we could use this one instead of new structs https://github.com/kubermatic/kubermatic/blob/189186b185bd40712370e79f18412b9a7887c9a9/pkg/apis/kubermatic/v1/cluster.go#L714

PratikDeoghare (approved):

pkg/api/v2/types.go
rastislavs: Not a bad idea. I just wanted an API that is more convenient to use for the UI - e.g. explicit IPv4 / IPv6 defaults instead of parsing them form pod / services CIDR slices (where they can be mixed in any order).

Issue #9412, opened 2022-03-23
Resource quota webhook


lsviben:
Add a validation webhook for Machines which enforces resource quotas.

PR #9613, opened 2022-04-14
Deploy user cluster webhook

lsviben:
Part of #9412 .

Adds a user-cluster webhook deployment, on which the webhook handlers for user cluster can be added.
For now there is no webhook implementation, but that will be added in subsequent PRs.

```release-note
NONE
```


pkg/resources/resources.go
```diff
@@ -431,6 +434,10 @@ const (
 	// configuration.
 	MachineControllerMutatingWebhookConfigurationName = "machine-controller.kubermatic.io"
 
+	// MachineValidatingWebhookConfigurationName is the name of the machine validating webhook
+	// configuration.
+	MachineValidatingWebhookConfigurationName = "machine.kubermatic.io"
+
```
ahmedwaleedmalik: We don't need these additions

pkg/resources/webhook/service.go
```diff
@@ -0,0 +1,113 @@
...
+
+type tlsServingCertCreatorData interface {
+	GetRootCA() (*triple.KeyPair, error)
+	Cluster() *kubermaticv1.Cluster
+}
+
+// TLSServingCertificateCreator returns a function to create/update the secret with the machine-webhook tls certificate.
+func TLSServingCertificateCreator(data tlsServingCertCreatorData) reconciling.NamedSecretCreatorGetter {
+	return func() (string, reconciling.SecretCreator) {
+		return resources.MachineWebhookServingCertSecretName, func(se *corev1.Secret) (*corev1.Secret, error) {
```
ahmedwaleedmalik: `resources.MachineWebhookServingCertSecretName` We should have a dedicated secret for this.

pkg/resources/webhook/deployment.go
```diff
@@ -0,0 +1,148 @@
...
+			} else {
+				args = append(args, "-v=2")
+			}
+
+			volumes := []corev1.Volume{
+				{
+					Name: "webhook-serving-cert",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: resources.MachineWebhookServingCertSecretName,
```
ahmedwaleedmalik: We should have a dedicated secret for these webhooks.

ahmedwaleedmalik (changes requested):

pkg/resources/resources.go
lsviben: right, i split this PR from my code with a new Machine webhook and this was leftover, will remove

ahmedwaleedmalik (approved):
/approve 🚀

PR #9621, opened 2022-04-14
Allow for longer-running conformance tests

xrstf:
Our periodic tests test multiple scenarios at once, compared to the presubmit jobs, which only test a single combination of provider+k8sversion+OS at a time. These periodic jobs can then easily run longer than the fixed 90 minute limit, leading to failures.

This PR makes the timeout a bit more dynamic.

```release-note
NONE
```


ahmedwaleedmalik (approved):
🚀

PR #9623, opened 2022-04-15
Extend KubeOne controller to migrate kubeone cluster to containerd

### LINKED ISSUE #9491: Extend KubeOne Controller for ContainerRuntime migration to Containerd
Extend KubeOne cluster to migrate kubeone cluster to contained by creating a kubeone pod running `kubeone migrate to-containerd`.
This is to avoid this error while upgrading kubeone clusters:
"Support for docker will be removed with Kubernetes 1.24 release. It is recommended to switch to containerd as container runtime using `kubeone migrate to-containerd`"

### Acceptance criteria
The container runtime of kubeone cluster is migrated to containerd

imharshita:

Extend KubeOne controller to migrate kubeone cluster to containerd by creating a kubeone pod running `kubeone migrate to-containerd`  when action = "migrate" and some minor refactoring to make the code more structured.

Fixes #9491

```release-note
none
```


pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -66,13 +66,28 @@ const (
 	ImportAction = "import"
 
 	// UpgradeControlPlaneAction is the action to upgrade kubeone cluster.
-	UpgradeControlPlaneAction = "upgradeCP"
+	UpgradeControlPlaneAction = "upgrade"
+
+	// MigrateControlPlaneAction is the action to upgrade kubeone cluster.
```
xrstf: This should probably be "migrate" and not "upgrade"

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -66,13 +66,28 @@ const (
...
 
 	// KubeOneUpgradePod is the name of kubeone pod performing upgrade.
 	KubeOneUpgradePod = "kubeone-upgrade"
 
-	// KubeOneKubeconfigPod is the name of kubeone pod fetching kubeconfig.
-	KubeOneKubeconfigPod = "kubeone-kubeconfig"
+	// KubeOneMigratePod is the name of kubeone pod performing container runtime migration.
+	KubeOneMigratePod = "kubeone-migrate"
+
+	// KubeOneImportConfigMap is the name of kubeone configmap performing import.
```
xrstf: A ConfigMap that performs an import? How does it do that?

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	if err != nil {
-		return reconcile.Result{}, err
+		return nil, err
 	}
 	currentVersion := version.String()
+	wantedVersion, err := getWantedVersion(currentManifest)
+	if err != nil {
+		return nil, err
+	}
+	if clusterStatus != kubermaticv1.StatusReconciling && clusterStatus != kubermaticv1.StatusError {
```
xrstf: Nit: If you flip the condition around, you don't have to indent as much code:

```go
if clusterStatus == ... || clusterStatus == ... {
   return nil, nil 
}

// more stuff...
```

Also, even more nitty, it's easier to me to read and comprehend "if thing is A or thing is B" than "if thing is NOT A and thing is NOT B".

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
-	currentManifest := manifestSecret.Data[resources.KubeOneManifest]
+	wantedContainerRuntime, err := returnWantedContainerRuntime(currentManifest)
+	if err != nil {
+		return nil, err
+	}
+	pod := &corev1.Pod{}
+	if clusterStatus != kubermaticv1.StatusReconciling && clusterStatus != kubermaticv1.StatusError {
+		if currentContainerRuntime != wantedContainerRuntime {
+			if wantedContainerRuntime != "" && wantedContainerRuntime == resources.ContainerRuntimeContainerd {
```
xrstf: Flipping these conditions around can also make this function more readable and keep the code on the left side of the screen.

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	}
+	return nil
+}
+
+func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, externalCluster *kubermaticv1.ExternalCluster) (reconcile.Result, error) {
+	if err := r.initiateImportAction(ctx, log, externalCluster); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	clusterStatus := externalCluster.Spec.CloudSpec.KubeOne.ClusterStatus.Status
```
xrstf: This PR doesn't introduce it, but why do we keep status innformation in the spec? This line looks confusing to me... can you explain why we have status in the spec?

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+		return reconcile.Result{}, fmt.Errorf("can not retrieve kubeone manifest secret: %w", err)
+	}
+	currentManifest := manifestSecret.Data[resources.KubeOneManifest]
+
+	_, err = r.upgradeAction(ctx, log, externalCluster, externalClusterProvider, currentManifest, clusterStatus)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err = r.checkPodStatusIfExists(ctx, log, externalCluster, UpgradeControlPlaneAction); err != nil {
```
xrstf: How about giving this function the pod name instead of the action name? Then it would not have to have some if-else-elseif-else-if-if-else chain :)

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+	manifestSecret := &corev1.Secret{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: externalCluster.Spec.CloudSpec.KubeOne.ManifestReference.Namespace, Name: externalCluster.Spec.CloudSpec.KubeOne.ManifestReference.Name}, manifestSecret); err != nil {
+		return reconcile.Result{}, fmt.Errorf("can not retrieve kubeone manifest secret: %w", err)
+	}
+	currentManifest := manifestSecret.Data[resources.KubeOneManifest]
+
+	_, err = r.upgradeAction(ctx, log, externalCluster, externalClusterProvider, currentManifest, clusterStatus)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
xrstf: This could also be written as `if _, err = ....; err != nil { ... }`, no?

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
-			return fmt.Errorf("failed to add kubeconfig reference to %s: %w", externalCluster.Name, err)
+			return fmt.Errorf("failed to update cluster status %s: %w", externalCluster.Name, err)
 		}
-		log.Debug("KubeOne Cluster Upgraded!")
-		if err := r.Delete(ctx, pod); err != nil {
-			return err
+		if action == UpgradeControlPlaneAction {
+			log.Debug("KubeOne Cluster Upgraded!")
+		} else if action == MigrateControlPlaneAction {
+			log.Debug("KubeOne Cluster Migrated!")
```
xrstf: If this happens only once, I would say we even use Info() and not Debug(). Those messages seem important enough for the info level. You even put an exclamation point in them ^^

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 		}
-	}
-	if pod.Status.Phase == corev1.PodFailed {
-		upgradeErr := fmt.Sprintf("Failed to upgrade kubeone cluster %s", externalCluster.Name)
+	} else if pod.Status.Phase == corev1.PodFailed {
+		upgradeErr := fmt.Sprintf("Failed to %s kubeone cluster %s", action, externalCluster.Name)
```
xrstf: I don't think we need to include the cluster name in this error message. It's being used in the ExternalCluster object itself (so the context is clear) and it's being logged (and I think the logger already contains the cluster name as well).

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -231,91 +229,220 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
 	log.Debugw("Checking kubeone pod status", "Pod", pod.Name)
-	for pod.Status.Phase == corev1.PodSucceeded {
+	if pod.Status.Phase == corev1.PodSucceeded {
 		oldexternalCluster := externalCluster.DeepCopy()
 		// update kubeone externalcluster status.
 		externalCluster.Spec.CloudSpec.KubeOne.ClusterStatus.Status = kubermaticv1.StatusRunning
 
 		if err := r.Patch(ctx, externalCluster, ctrlruntimeclient.MergeFrom(oldexternalCluster)); err != nil {
-			return fmt.Errorf("failed to add kubeconfig reference to %s: %w", externalCluster.Name, err)
+			return fmt.Errorf("failed to update cluster status %s: %w", externalCluster.Name, err)
```
xrstf: All of the errors returned by this function should not contain the cluster name. Instead, if you need, either attach it to the logger when you log the error (further up the stacktrace) or add the cluster name at a higher level in the stacktrace.

Meaning don't do

```go
func foobar(name string) error {
   if bla {
      return fmt.Errorf("%s is bad, because: foo", name)
   }
 
   if foobar {
      return fmt.Errorf("%s is bad, because schwupsdiwups", name)
   }

   return nil
}

if err := foobar("xrstf"); err != nil {
    return err
}
```

but instead do

```go
func foobar(name string) error {
   if bla {
      return errors.new("foo")
   }
 
   if foobar {
      return errors.New("schwupsdiwups")
   }

   return nil
}

if err := foobar("xrstf"); err != nil {
    return fmt.Errorf("xrstf is invalid: %w", err)
}
```

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -337,21 +461,21 @@ func (r *reconciler) importCluster(ctx context.Context, log *zap.SugaredLogger,
...
+	generatedPod, err := r.generateKubeOneActionPod(ctx, log, externalCluster, ImportAction)
 	if err != nil {
 		return fmt.Errorf("Could not generate kubeone pod for kubeone cluster %s: %w", externalCluster.Name, err)
 	}
 
 	log.Debug("Create kubeone pod to fetch kubeconfig")
 	if err := r.Create(ctx, generatedPod); err != nil {
 		if !kerrors.IsAlreadyExists(err) {
-			return fmt.Errorf("Could not create kubeone pod %s/%s for kubeone cluster %s: %w", KubeOneKubeconfigPod, generatedPod.Namespace, externalCluster.Name, err)
+			return fmt.Errorf("Could not create kubeone pod %s/%s for kubeone cluster %s: %w", KubeOneImportPod, generatedPod.Namespace, externalCluster.Name, err)
```
xrstf: Same here, I think the `for kubeone cluster %s` can be removed from all error messages created in this function.

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -580,137 +694,20 @@ func (r *reconciler) getKubeOneSecret(ctx context.Context, ref providerconfig.Gl
...
+	switch {
+	case action == ImportAction:
+		name = KubeOneImportConfigMap
+		scriptToRun += "kubeone kubeconfig -m kubeonemanifest/manifest"
+	case action == UpgradeControlPlaneAction:
+		name = KubeOneUpgradeConfigMap
 		scriptToRun += "kubeone apply -m kubeonemanifest/manifest -y"
+	case action == MigrateControlPlaneAction:
+		name = KubeOneMigrateConfigMap
+		scriptToRun += "kubeone migrate to-containerd -m kubeonemanifest/manifest"
```
xrstf: Ultra nitpick, but for written scripts, I prefer the long form of CLI flags (`--manifest` instead of `-m`)

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -724,17 +721,17 @@ func generateConfigMap(name, namespace, action string) *corev1.ConfigMap {
...
 }
 
-func (r *reconciler) upgradeCluster(ctx context.Context, log *zap.SugaredLogger, externalCluster *kubermaticv1.ExternalCluster) error {
+func (r *reconciler) upgradeCluster(ctx context.Context, log *zap.SugaredLogger, externalCluster *kubermaticv1.ExternalCluster) (*corev1.Pod, error) {
 	log.Debug("Generate kubeone pod to upgrade kubeone")
-	generatedPod, err := r.generateKubeOneUpgradePod(ctx, log, externalCluster)
+	generatedPod, err := r.generateKubeOneActionPod(ctx, log, externalCluster, UpgradeControlPlaneAction)
 	if err != nil {
-		return fmt.Errorf("Could not generate kubeone pod for kubeone cluster upgrade %s: %w", externalCluster.Name, err)
+		return nil, fmt.Errorf("Could not generate kubeone pod for kubeone cluster upgrade %s: %w", externalCluster.Name, err)
```
xrstf: `for kubeone cluster upgrade %s` can go away as well.

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -743,13 +740,38 @@ func (r *reconciler) upgradeCluster(ctx context.Context, log *zap.SugaredLogger,
...
-	return nil
+	return generatedPod, nil
 }
 
-func (r *reconciler) generateKubeOneUpgradePod(ctx context.Context, log *zap.SugaredLogger, externalCluster *kubermaticv1.ExternalCluster) (*corev1.Pod, error) {
+func (r *reconciler) migrateCluster(ctx context.Context, log *zap.SugaredLogger, externalCluster *kubermaticv1.ExternalCluster) (*corev1.Pod, error) {
+	log.Debug("Generate kubeone pod to migrate kubeone")
+	generatedPod, err := r.generateKubeOneActionPod(ctx, log, externalCluster, MigrateControlPlaneAction)
+	if err != nil {
+		return nil, fmt.Errorf("Could not generate kubeone pod to migrate container runtime %s: %w", externalCluster.Name, err)
```
xrstf: Error messages should start with a lowercase letter

pkg/handler/v2/external_cluster/external_cluster.go
```diff
@@ -624,6 +625,26 @@ func PatchEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider provi
 	}
 }
 
+func CheckContainerRuntime(ctx context.Context,
```
xrstf: So I think this is bad architecture. The controllers are now calling code from the API handlers, which they should have no knowledge of whatsoever. Now the controllers cannot work without our API and since we might move the API out of this repo in the mid future, this will break.

Can you restructure your code to put commonly used code outside of the API handlers (i.e. outside of `pkg/handler/`)?

pkg/handler/v2/external_cluster/external_cluster.go
```diff
@@ -624,6 +625,26 @@ func PatchEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider provi
...
+	externalCluster *kubermaticv1.ExternalCluster,
+	externalClusterProvider provider.ExternalClusterProvider,
+) (string, error) {
+	nodes, err := externalClusterProvider.ListNodes(ctx, externalCluster)
+	if err != nil {
+		return "", fmt.Errorf("Failed to fetch container runtime: not able to list nodes %w", err)
```
xrstf: Errors should begin with a lowercase letter.

xrstf (changes requested):
Took a bit [for me to reply], but here are my comments.

pkg/controller/master-controller-manager/kubeone/controller.go
imharshita: For any operation to perform for kubeone cluster, we are verifying if the cluster status is `Running`.
this will help in 2 things:
- avoiding 2 parallel reconciling(upgrading/migrating) operations being performed.
- avoiding operation on a cluster with an error status.
So we will trigger a new operation only when the Cluster is in `Running` (!Reconciling/!Error) status

pkg/handler/v2/external_cluster/external_cluster.go
imharshita: Good Point!
I am moving this method to pkg/kubernetes

pkg/controller/master-controller-manager/kubeone/controller.go
xrstf: That doesn't explain why status information is in the spec.

imharshita: So what do you suggest ?

imharshita: Where should we place the status?

xrstf: In the status subresource: https://book-v1.book.kubebuilder.io/basics/status_subresource.html -- the same thing we use for Project status, Cluster status, User status and (soon) Seed status :)

Issue #9345, opened 2022-03-16
Deletion gets stuck for clusters restored from snapshots


PratikDeoghare:
### What happened?

- Create AWS cluster. 
- Take a snapshot. 
- Restore the cluster from the snapshot.
- Try deleting. 
- Fail with` failed cloud provider cleanup: failed to clean up security group: DependencyViolation: resource sg-00febb5ba33c69f5d has a dependent object status code: 400, request id: 6cd96f18-fe4f-4367-a993-c892f1b79937 `

Also affects Azure clusters. They fail with this error: 

`failed cloud provider cleanup: failed to delete security group "kubernetes-kcgnt88w8x": network.SecurityGroupsClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="InUseNetworkSecurityGroupCannotBeDeleted" Message="Network security group /subscriptions/528474a9-8cc6-46c8-9b74-1dbc31c4bb7f/resourceGroups/kubernetes-kcgnt88w8x/providers/Microsoft.Network/networkSecurityGroups/kubernetes-kcgnt88w8x cannot be deleted because it is in use by the following resources: /subscriptions/528474a9-8cc6-46c8-9b74-1dbc31c4bb7f/resourceGroups/kubernetes-kcgnt88w8x/providers/Microsoft.Network/networkInterfaces/gracious-volhard-worker-ds7hsj-57ffffd656-xbgg5-netiface. In order to delete the Network security group, remove the association with the resource(s). To learn how to do this, see aka.ms/deletensg." Details=[]`

### What should happen?
Cluster should have been successfully deleted. 

### How to reproduce?

See instructions in _What happened_ section. 

### Environment
kubermatic dev environment. 
- KKP Version: 2.20 
- Shared or separate master/seed clusters?: Shared
- Master/seed cluster provider: GCP

PratikDeoghare:
Postmortem:
- Clusters got stuck during deletions because security group could not be deleted. 
- Security group could not be deleted because it was attached to a network interface which 
  was attached to an orphaned VM. 
- How did the VM get orphaned?
  Snapshot restore doesn't restore the cluster from the correct snapshot if 
  etcd-launcher is not enabled. Etcd-launcher was disabled by mistake. 
  So restore looked like it worked but it really didn't. Restoring like this 
  orphaned the VM because there was no corresponding machine deployment in 
  the restored cluster.

Etcd-launcher should be enabled for proper functioning of the snapshot restore functionality.

[...]

PR #9633, opened 2022-04-19
fail etcd restore if etcdLauncher feature not enabled on cluster

### LINKED ISSUE #9453: Snapshot restore shouldn't be allowed if etcd launcher is not enabled

PratikDeoghare:

Etcd launcher is required for proper working of snapshot restore. This PR checks if it enabled on the cluster while reconciling. If it is not enabled error is returned. 

Fixes #9453 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
+++ b/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
@@ -144,6 +144,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	if !cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] {
+		return reconcile.Result{}, fmt.Errorf("etcdLauncher not enabled on cluster: %q", cluster.Name)
+	}
+
 	if cluster.Status.NamespaceName == "" {
 		log.Debug("Cluster has no namespace name yet, skipping")
 		return reconcile.Result{}, nil
--- a/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
+++ b/pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
@@ -144,10 +144,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
-	if !cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] {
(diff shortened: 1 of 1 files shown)


pkg/controller/seed-controller-manager/etcdrestore/etcd_restore_controller.go
```diff
@@ -144,6 +144,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	if !cluster.Spec.Features[kubermaticv1.ClusterFeatureEtcdLauncher] {
+		return reconcile.Result{}, fmt.Errorf("etcdLauncher not enabled on cluster: %q", cluster.Name)
+	}
```
embik: Will this surface to users in any way? Are they gonna see it as failed in the UI? Maybe I'm wrong, but this is probably going to look like a ~backup~ restore forever in "pending", no?

embik: I was wondering if a valIdating webhook made sense here to prevent creating an `EtcdRestore` on the API, but I think we don't have a webhook for this in place yet?

lsviben: We already had a PR with a webhook :D. But then decided that a validating webhook would be a wrong approach as there is nothing unvalid with the EtcdRestore itself, rather something else(etcd-launcher is not enabled)

lsviben: But I agree that it would be good to inform the user in some way that the etcdrestore is not working. Not sure where to put this info? Maybe add new phase in EtcdRestore phases, `EtcdLauncherNotEnabled` ?

embik: Oh lol I was not aware of the webhook PR ... I would have approved it 😂 

But yeah, a phase is the next best thing IMHO. At least this should be moved into the private `reconcile` function so an event gets emitted.

PratikDeoghare: Good idea. Moved to private reconcile.

PratikDeoghare: Could I implement EtcdLauncherNotEnabled in a different story? This anyway is part of series of stories that came from https://github.com/kubermatic/kubermatic/issues/9345 issue. Probably should be put under an epic.

embik (approved):
/approve

PR #9636, opened 2022-04-19
Update Alertmanager to 0.24.0

xrstf:
The changelogs do not mention any BC breaks between 0.22.2 and 0.24.0, so I hope this is a safe upgrade. Same goes for the configmap-reloader.

```release-note
Update Alertmanager to 0.24.0
```


ahmedwaleedmalik (approved):
🏄‍♂️

PR #9582, opened 2022-04-12
Add encryption-at-rest fields and validation to kubermaticv1.Cluster CRD

embik:
**What does this PR do / Why do we need it**:

This PR tries to split #8834 up by moving the CRD changes to a dedicated PR. The underlying `encryption_controller` that will provide the main functionality will be implemented in a later PR.

I'm adjusting the encryption-at-rest proposal based on results from my prototyping of the last few weeks, improving the API to both reduce complexity and surface the Kubernetes API feature more accurately. A first set of validation rules is also part of the PR.

**Does this PR close any issues?**:
xref #8400

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

(the PR implementing the actual functionality will have the release note entry announcing this feature)

PR #9637, opened 2022-04-19
Do not validate secretbox keys if secretbox is not configured

embik:

Noticed #9582 introduced a nil pointer exception in the validation webhook when an invalid encryptionConfiguration was passed. Woops. Good news is that the configuration is invalid anyway and shouldn't be accepted, but obviously a server error was not the intended response. This PR fixes this by not trying to iterate over secretbox keys if secretbox isn't configured (duh).

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #9642, opened 2022-04-19
Update Helm-Exporter to 1.2.2

xrstf:
This updates our Helm exporter and does quite a jump forward. I tested the new config on our dev system, so it should work pretty well everywhere else.

```release-note
ACTION REQUIRED: Update Helm-Exporter to 1.2.2; Helm 2 is not supported anymore, on clusters with many Helm releases, performance tweaks might be necessary.
```


ahmedwaleedmalik (approved):
🚀

PR #9647, opened 2022-04-20
update kube-state-metrics to 2.4.2

xrstf:
According to the release notes, no existing metrics were changed, so this should be a safe upgrade.

```release-note
Update kube-state-metrics to 2.4.2
```


ahmedwaleedmalik (approved):
/approve 🚀 
/retest

PR #9650, opened 2022-04-20
Machine resource quota webhook 

### LINKED ISSUE #9625: Add machine validating webhook
Add a validating webhook in the user cluster for machines. It will be used for resource quota enforcement.

lsviben:
Part of #9412. This PR adds a machine validating webhook for resource quotas.

As we dont have yet the ResourceQuotas for projects implemented, or its calculation, so for now getting the quotas are hardcoded with some fake value. Also getting the resource requirements for the machine, per provider, will be implemented in the next PRs, so for now no providers are supported - meaning the webhook wont do anything yet and will just pass the validation.

What this PR brings is the scaffolding of the resource quota validation, with the checks for requested resources vs. quota and current quota usage.

Fixes #9625 

```release-note
Added a Machine validating webhook which checks the Machine resource request(CPU, Memory, Storage) against its projects resource quota(if set). This is an EE feature only. 
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/machine/webhook.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/machine/webhook.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/controller/user-cluster-controller-manager/resources/resources/machine/webhook.go
+++ b/pkg/controller/user-cluster-controller-manager/resources/resources/machine/webhook.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package machine
+
+import (
+	"crypto/x509"
+	"fmt"
+
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/certificates/triple"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	clusterAPIGroup   = "cluster.k8s.io"
+	clusterAPIVersion = "v1alpha1"
+)
+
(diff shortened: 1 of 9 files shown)


pkg/controller/user-cluster-controller-manager/resources/resources/machine/webhook.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
xrstf: 2022 please

pkg/ee/validation/machine/resourcequota.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// add requested resources to current usage and compare
+	combinedUsage := NewResourceQuota(currentUsage.cpu, currentUsage.mem, currentUsage.storage)
+	combinedUsage.Cpu().Add(*quotaReq.Cpu())
+	combinedUsage.Memory().Add(*quotaReq.Memory())
+	combinedUsage.Storage().Add(*quotaReq.Storage())
+
+	if quota.Cpu().Cmp(*combinedUsage.Cpu()) < 0 {
+		log.Debugf("requested CPU %q would exceed current quota (quota/used %q/%q)",
```
xrstf: I love readable log messages, but should we use structured logging here? Honest question. "No." is a valid answer.

pkg/ee/validation/machine/resourcequota.go
```diff
@@ -0,0 +1,155 @@
...
+	storageUsed, err := resource.ParseQuantity("60G")
+	if err != nil {
+		return nil, nil, fmt.Errorf("error parsing quantity: %w", err)
+	}
+
+	return NewResourceQuota(cpu, mem, storage),
+		NewResourceQuota(cpuUsed, memUsed, storageUsed), nil
+}
+
+type ResourceQuota struct {
```
xrstf: This struct seems to be used not just for quotas, but also to express the current consumption. So the name for this struct might be a bit unfortunate. What about `ResourceDetails` maybe?

xrstf: What about using `corev1.ResourceList`? That one looks super similar, maybe similar enough?

pkg/resources/resources.go
```diff
@@ -431,6 +431,9 @@ const (
 	// configuration.
 	MachineControllerMutatingWebhookConfigurationName = "machine-controller.kubermatic.io"
 
+	// MachineValidatingWebhookConfigurationName is the name for the machine validating webhook.
+	MachineValidatingWebhookConfigurationName = "machine.kubermatic.io"
```
xrstf: Should we use our new domain (`kubermatic.k8c.io`) here?

pkg/webhook/machine/validation/validation.go
```diff
@@ -0,0 +1,69 @@
...
+
+// validator for validating Kubermatic Machine CRD.
+type validator struct {
+	log        *zap.SugaredLogger
+	seedClient ctrlruntimeclient.Client
+}
+
+// NewValidator returns a new Machine validator.
+func NewValidator(seedClient ctrlruntimeclient.Client, log *zap.SugaredLogger) *validator {
+	log.Named("machine-validator")
```
xrstf: This line looks out of place.

xrstf (changes requested):
A few nits here and there.

pkg/ee/validation/machine/resourcequota.go
lsviben: My first thought and implementation was with `corev1.ResourceList`, but it has some additional things like `ResourceEphemeralStorage` and Pods() which I thought could make it confusing. So I decided on something simpler

pkg/ee/validation/machine/resourcequota.go
lsviben: Changed to structured, looks better

xrstf (approved):
/approve

Issue #9560, opened 2022-04-07
Invalid minio scraping target when backup and TLS are enabled


heilerich:
### What happened?
When the minio chart is deployed with both TLS and backups enabled the minio deployment will have a second container named `backup`. This container is discovered as an additional scraping target by the minio scraping job introduced in #8467.
This will create false positive `PromScrapeFailed` alerts because the backup sidecar is not exposing any metrics.

### What should happen?
The above mentioned scraping job needs an additional selector on `__meta_kubernetes_pod_container_name` to prevent the discovery of the additional scraping target.

### How to reproduce?
Deploy minio helm chart with TLS and backups enabled.

### Environment

- KKP Version: v2.19.0
- Shared or separate master/seed clusters?: shared
- Master/seed cluster provider: vsphere

PR #9662, opened 2022-04-21
Add support for prometheus.io/scheme annotation in Prometheus chart

embik:

I was looking at #9560 and #8467 and wondered why that later PR introduced a completely new scraping config instead of just properly annotating the minio Pods ... Until I figured out that we don't support the `prometheus.io/scheme` annotation in the `prometheus` chart under `charts/monitoring`. _However_, we support that standard annotation in user clusters via the MLA Prometheus created there:

https://github.com/kubermatic/kubermatic/blob/d61215605dbd093add135cac67f5e06b9279e4dc/pkg/controller/user-cluster-controller-manager/resources/resources/mla/prometheus/configmap.go#L128-L132

This PR adds support for `prometheus.io/scheme=https` to the Master/Seed MLA Prometheus as well, allowing us to potentially swap out the minio scraping configuration (or others) with something much simpler and giving Master/Seed administrators the expected set of `prometheus.io` annotations they know from other Prometheus installations.

Fixes #

**Documentation**:

```release-note
The Master/Seed MLA Prometheus from `charts/monitoring/prometheus` supports annotating Pods with `prometheus.io/scheme=https` to use HTTPS
```

### DIFF
--- a/charts/monitoring/prometheus/config/scraping/pods.yaml
+++ b/charts/monitoring/prometheus/config/scraping/pods.yaml
@@ -80,6 +80,12 @@ relabel_configs:
   replacement: $1:$2
   target_label: __address__
 
+# allow to override the scrape scheme with https (defaults to plain http) via prometheus.io/scheme
+- source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scheme]
+  action: replace
+  regex: (https?)
+  target_label: __scheme__
+
 # refine labels further
 - action: labelmap
   regex: __meta_kubernetes_pod_label_(.+)


embik:
/test pre-kubermatic-mla-e2e

for good measure, although I'm not even sure this chart is deployed as part of that test.

wurbanski:
This is a good first step to clearing the monitoring chart of per-service scrape configs. Will make it easier before we get to migrate to prometheus operator and move configs out of the main chart completely.

/lgtm

embik:
> This is a good first step to clearing the monitoring chart of per-service scrape configs. Will make it easier before we get to migrate to prometheus operator and move configs out of the main chart completely.
> 
> /lgtm

Ah, if we're planning to move to prometheus-operator I'm not sure how much sense it makes to do work after this (having support for this is still neat), since `ServiceMonitors` and `PodMonitors` lean more towards the "per-service scrape configs" anyway.

PR #9668, opened 2022-04-22
Move ClusterAddress to ClusterStatus

xrstf:
This PR moves the `ClusterAddress` into the `ClusterStatus`.

We are thinking hard about how KKP will evolve in the future and one of the options is something like kcp. Solutions like that are geared towards a world where CRDs have a spec and a status, nothing more. Tools like kcp's syncer will copy the spec of a resource in one directory and update the status in the opposite direction.

While I anticipate that this behaviour can be tweaked to support other fields, it's probably safer and easier if we conform the the upstream expectations.

I checked why the address was never part of the status: It was introduced in #6 (yes, that long ago) with no further information attached. I therefore think that we can move it without breaking any intentions Stefan had at the time. From what I heard, this was done in a time where CRDs not even existed on paper, so the special meaning of a "status" "subresource" didn't exist either.

**GetAddress(), really?**
Just to be on the safe side, I introduced `cluster.GetAddress()` which will dynamically assemble the status and take care of smoothing over the data, so that even if for some reason a field hasn't made it into the status or a component on the master needs to process data from an older seed, the controllers will still work properly. In KKP 2.22 we should remove the helper function again.

**Optional Fields, why?**
Just like with other ClusterStatus fields, different fields are set at different times. The AdminToken is set independently of the remaining fields.

Previously, the address was set when a Cluster was created (even though all fields are initially empty). Now the address fields are set for the first time later during the reconciliation and they are importantly *patched* into the ClusterStatus. And patches will ignore empty fields because they do not produce any changes, and so you might think you can do

```
cluster.Status.Address = ClusterAddress{}
client.Status().Patch(ctx, cluster)
```

but this will *not* send the empty fields. And this then causes validation errors because the CRD marked the fields as required. 

Just because the address fields and the AdminToken are set at different times, this PR marks all fields in the ClusterAddress to be opional. At least we don't end up with tons of pointers and additional nil checks. But to be clear, the `+optional` is solely because when we Patch() the status, empty fields are not sent.

```release-note
The ClusterAddress for user clusters was moved to the ClusterStatus; the old `address` field remains only for the migration and should not be relied upon anymore.
```


xrstf:
Note to self: I am unsure when and how to run the migration and to solve this, we should first ensure that the KKP operator installs the new CRDs onto seed clusters.

So this PR might be on hold for quite some time.

hack/ci/deploy-dev.sh
```diff
@@ -45,4 +45,5 @@ echodate "Successfully got secrets for dev from Vault"
 
 echodate "Deploying ${DEPLOY_STACK} stack to dev"
 ./hack/ci/deploy.sh master ${VALUES_FILE}
+./hack/ci/deploy.sh seed ${VALUES_FILE}
```
embik: perhaps a stupid question, but why add this here? Didn't we essentially eliminate the need to run the installer against a seed with #9748? I feel like I'm missing something here.

pkg/controller/operator/seed/reconciler.go
```diff
@@ -338,19 +338,19 @@ func (r *Reconciler) reconcileResources(ctx context.Context, cfg *kubermaticv1.K
...
 
-	if err := r.reconcileServices(ctx, cfg, seed, client, log); err != nil {
+	if err := r.reconcilePodDisruptionBudgets(ctx, cfg, seed, client, log); err != nil {
 		return err
 	}
 
-	if err := r.reconcileAdmissionWebhooks(ctx, cfg, seed, client, log); err != nil {
+	if err := r.reconcileServices(ctx, cfg, seed, client, log); err != nil {
 		return err
 	}
```
embik, line 355: Why move around the order of resource reconciling?

embik (commented):
Looks pretty good, just two minor questions.

hack/ci/deploy-dev.sh
xrstf: True, this is a leftover from when the cluster migration code was living in the kkp installer. It's now in the seed-ctrl-mgr though.

pkg/controller/operator/seed/reconciler.go
xrstf: The old order was

* Deployments
* PDBs
* Services
* Webhooks

The new order is

* first migrate the data
* Webhooks (first because it's more important to have this one working and up-to-date then the seed-ctrl-mgr
* now the remaining parts in their original order: Deployments, PDBs, Services

So all I did is move the webhooks further up.

embik (approved):
/approve

pkg/controller/seed-controller-manager/kubernetes/address.go
```diff
@@ -30,15 +31,15 @@ import (
 func (r *Reconciler) syncAddress(ctx context.Context, log *zap.SugaredLogger, cluster *kubermaticv1.Cluster, seed *kubermaticv1.Seed) error {
 	var err error
 	// TODO(mrIncompetent): The token should be moved out of Address. But maybe we rather implement another auth-handling? Like openid-connect?
```
lsviben, line 33: what about this TODO?

Do we still want to do it? Maybe it would be a good time now then. Or no point so the TODO can be removed?

lsviben (commented):
just left a question, otherwise LGTM

pkg/controller/seed-controller-manager/kubernetes/address.go
xrstf: I am not touching this part. This TODO is ancient and you know the old saying: Don't fix what ain't broken.

Issue #8091, opened 2021-10-26
add serverTLSBootstrap to kubelet-configmap addon


kron4eg:

csengerszabo:
Internal reference: 8106, 3752
/label sig/cluster-management

csengerszabo:
Nodes which are managed by MC serverTLSBootstrap is already enabled, but in case of own machines brought into the cluster, it is not enabled, so the question is if this is possible to implement.

kron4eg:
/close

this is so outdated

PR #9674, opened 2022-04-25
Don't set '--kubelet-certificate-authority' for 'bring-your-own' user clusters

### LINKED ISSUE #9673: kubeadm/byoc provider nodes not available to kubectl commands after certificates expired
### What happened?

In a "kubeadm" / "bring your own cluster" setup, kubelet server TLS certificates expired. Due to that, commands like `kubectl logs` or `kubectl exec` can fail if the TLS certificate for the node hosting the target pod expired. Error message is `x509: certificate has expired or is not yet valid`.

This happens because #8014 introduced `--kubelet-certificate-authority` to `kube-apiserver`, which enables TLS cert validation for the `kube-apiserver` to `kubelet` communication. While the cert presented by `kubelet` is still signed by the correct CA, enabling this flag also starts validation of the cert expiry.

It only affects the "kubeadm" provider because for other providers, nodes are deployed with `machine-controller`, which enables the `serverTLSBootstrap ` kubelet feature flag and as such, automatically rotates server certs. A full fix might be #8091.

This was reported as Kubermatic support ticket 3752.

### What should happen?

Nodes joined to a "kubeadm" user cluster should not stop functioning for `kubectl {logs,exec}` commands.

### How to reproduce?

Set up a "kubeadm" user cluster, join node, wait for TLS cert to expire.

### Environment

- KKP Version: 2.19.0 and higher
- Shared or separate master/seed clusters?: N/A
- Master/seed cluster provider: N/A

embik:
As per #9673, the "kubeadm" provider is special because it does not automatically rotate kubelet server TLS certificates, and the `--kubelet-certificate-authority` flag is problematic in that case. Eventually, we should make sure that TLS rotation properly happens on those clusters as well, but this PR (temporarily?) disables the flag for "kubeadm" clusters and this PR will be cherry-picked into 2.20 and 2.19. #8091 will likely end up being the complete fix, and after that is implemented this PR can be reverted.

Fixes #9673

**Documentation**:

```release-note
The flag `--kubelet-certificate-authority` (introduced in KKP 2.19) is not set for "kubeadm" / "bringyourown" user clusters anymore
```

### DIFF
--- a/pkg/resources/apiserver/deployment.go
+++ b/pkg/resources/apiserver/deployment.go
@@ -310,7 +310,15 @@ func getApiserverFlags(data *resources.TemplateData, etcdEndpoints []string, ena
 		"--client-ca-file", "/etc/kubernetes/pki/ca/ca.crt",
 		"--kubelet-client-certificate", "/etc/kubernetes/kubelet/kubelet-client.crt",
 		"--kubelet-client-key", "/etc/kubernetes/kubelet/kubelet-client.key",
-		"--kubelet-certificate-authority", "/etc/kubernetes/pki/ca/ca.crt",
+	}
+
+	// the "bring-your-own" provider does not support automatic TLS rotation in kubelets yet,
+	// and because of that certs might expire and kube-apiserver cannot validate the connection anymore.
+	if cluster.Spec.Cloud.BringYourOwn == nil {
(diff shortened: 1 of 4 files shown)


xrstf (approved):
/approve

PR #9682, opened 2022-04-25
update cloud provider dependencies

xrstf:
This is just a random round of updates for our cloud provider libraries. The Anexia stuff has a new import path, but other than that, nothing to report here.

```release-note
NONE
```


embik (approved):
/approve
/lgtm

PR #9648, opened 2022-04-20
Update Loki & promtail to 2.5

xrstf:
**What does this PR do / Why do we need it**:
The upgrade notes (https://github.com/grafana/loki/blob/v2.5.0/docs/sources/upgrading/_index.md) don't seem to affect our default configuration.

**Does this PR introduce a user-facing change?**:
(NB: #9693 reverts the promtail part again, so do not mention promtail in the release notes)
```release-note
Update Loki to 2.5
```

PR #9683, opened 2022-04-25
fix promtail client config

xrstf:
In #9648 we updated promtail to v2.5, but I missed that the client configuration was changed since previous versions (via grafana/helm-charts#1119). Since then the client config is not an inline YAML string anymore, but a regular data structure that is YAML-i-fied by the chart itself.

This PR updates our default config to match this new behaviour. Yes it looks weird that we supposedly create a client that only has `url` specified, but that's how the upstream chat works: https://github.com/grafana/helm-charts/blob/5a990ab9883bf2c001204cf9bc867b6a286f4363/charts/promtail/values.yaml#L347-L351

```release-note
NONE
```

### DIFF
--- a/charts/logging/promtail/test/default.yaml.out
+++ b/charts/logging/promtail/test/default.yaml.out
@@ -32,20 +32,13 @@ stringData:
     
     clients:
       - url: http://loki:3100/loki/api/v1/push
-      |
-        # Maximum wait period before sending batch
-        batchwait: 1s
-        # Maximum batch size to accrue before sending, unit is byte
-        batchsize: 102400
-        # Maximum time to wait for server to respond to a request
-        timeout: 10s
-        backoff_config:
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
/approve 🚀 

/hold for @wurbanski to review this as well

ahmedwaleedmalik:
FYI @xrstf 
https://github.com/grafana/helm-charts/issues/1214

xrstf:
Yeah we're kind of aware that this creates a broken config, but it should at least create a config that lets promtail run. I'm expecting the chart to be fixed soon, but don't want to let our job be broken for so long. ^^

xrstf:
You know what, let's simply revert the promtail update. I fear that we would forget this time bomb in our codebase before releasing KKP 2.20.

PR #9685, opened 2022-04-25
update nginx-ingress to 1.2.0

xrstf:
nginx 1.2.0 contains security fixes and we should update quickly, even though endusers (in a KKP stack) are not able to fiddle with Ingresses on the seed/master clusters.

```release-note
Update nginx-ingress to 1.2.0
```


ahmedwaleedmalik (approved):
/approve

PR #9686, opened 2022-04-25
Update Prometheus to 2.35

xrstf:
What it says on the tin.

Note that this disables support for TLS 1.0 and 1.1 in Prometheus, plus we will now reject SHA-1 certificates.

```release-note
Update Prometheus to 2.35 (disables support for SHA-1 certificates and TLS 1.0/1.1)
```


ahmedwaleedmalik (approved):
/approve 🚀

PR #9693, opened 2022-04-25
revert promtail update

xrstf:
#9648 introduced a broken promtail (see #9683), so let's revert that bad update and wait for a fix.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #9694, opened 2022-04-26
Fix IPVS strict ARP defaulting

rastislavs:
In https://github.com/kubermatic/kubermatic/pull/9211 IPVS Strict ARP defaulting has been moved. As a result of the move, is does not work as expected ATM - Strict ARP should be defaulted to true any time when IPVS proxy mode is enabled as it was done in the original PR (https://github.com/kubermatic/kubermatic/pull/7410).

This PR also refactors the mutation webhook tests, where proxy mode handling was a bit convoluted and sometimes even incorrect.

Fixes #

**Documentation**:

```release-note
NONE
```


WeirdMachine:
/lgtm

Thanks for taking care!

PR #9701, opened 2022-04-27
update golang images used for various scripts

xrstf:
There were still a few spots where we used Go 1.17. None of them "mattered", but it's still nice to be consistent.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡

Issue #9617, opened 2022-04-14
Integrate Azure dual-stack machine-controller API in KKP


rastislavs:
Integrate https://github.com/kubermatic/machine-controller/issues/1251 into KKP

Also test dual-stack end-to end on all supported OS platforms.

PratikDeoghare:
Moving this to stalled. For some operating systems we still don't have working dualstack cluster. 

CentOS -- Cilium agent crashes. 
Flatcar -- DHCPv6 is broken. 

TODO: Link issues here.

PratikDeoghare:
Integration is complete. It is not yet supported for some combinations of cloud provider, cni and operating systems. 

Supported/un-supported combinations are tracked in [this test](https://github.com/kubermatic/kubermatic/blob/master/pkg/test/dualstack/cloud_providers_dualstack_test.go#L94) where unsupported combinations are commented out. 

This epic tracks progress of investigations of unsupported combinations. #9994

PR #9703, opened 2022-04-27
add test for validating node, pod, svc for ip family

PratikDeoghare:
This PR adds test that can be run against some cluster to validate if nodes, pods, services of the cluster have addresses in the given IP family. It is based on https://kubernetes.io/docs/tasks/network/validate-dual-stack/

Next PRs will setup the  clusters to be targeted by this test. 

Part of https://github.com/kubermatic/kubermatic/issues/9617

```release-note
NONE
```


PratikDeoghare:
This is now part of https://github.com/kubermatic/kubermatic/pull/9751/files

PR #9706, opened 2022-04-27
Introduce SeedStatus

xrstf:
The setup procedure for a seed cluster is not very convenient because users must run the kubermatic-installer against the seed cluster _after_ the master has been updated. The installer's only task (w.r.t. to KKP itself) is to setup CRDs.

However the KKP operator at this point will most likely already have reconciled the seed cluster and deployed a new version of the seed-ctrl-mgr (when updating KKP from 2.19 to 2.20, for example). This means the seed-ctrl-mgr can be newer than the CRDs it's supposed to work with.

This situation is problematic in many ways. In order to solve it, we want to make the operator install the CRDs into seeds, thereby making it fully automatic. This is however a process that can fail (for many reasons) and we need a way to let the KKP admin know about this (better than just logging it).

Also, some KKP setups make use of the `kubermatic.k8c.io/skip-reconciling` annotation to slowly roll out new KKP versions across multiple seeds. So now we effectively have a version skew and need to deal with that.

To properly coordinate seed rollouts, this PR extends `Seed` resources by their `status` subresource. The SeedStatus will, for now, keep track of:

* The Kubermatic version on the seed.
* The seed cluster's Kubernetes version.
* The number of userclusters on the seed (updated regularly, so no watch is setup for this).
* A number of conditions (at the moment one to signalize whether the kubeconfig is valid and one to signal that reconciling worked).
* A phase to summarize the various conditions and other fields.

A new `seed-status-controller` on the master will take care of all of the fields, except for the ReconcilingSuccessful condition, which is controlled by the seed-operator.

screenshot-2022-04-27-212243

```release-note
Seed resources now make use of the status subresource to keep track of Seed version, health and other conditions.
```


xrstf:
/hold

awaiting feedback from the team

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -90,6 +90,66 @@ func IsProviderSupported(name string) bool {
...
+	SeedUnhealthyPhase SeedPhase = "Unhealthy"
+
+	// SeedInvalidPhase means the seed kubeconfig is defunct.
+	SeedInvalidPhase SeedPhase = "Invalid"
+
+	// SeedTerminatingPhase means the seed is currently being deleted.
+	SeedTerminatingPhase SeedPhase = "Terminating"
+
+	// SeedPausedPhase means the seed is not being reconciled because if SkipReconciling
+	// annotation is set.
```
ahmedwaleedmalik:
```suggestion
	// SeedPausedPhase means the seed is not being reconciled because SkipReconciling
	// annotation is set.
```

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -90,6 +90,66 @@ func IsProviderSupported(name string) bool {
...
+	SeedPausedPhase SeedPhase = "Paused"
+)
+
+// +kubebuilder:validation:Enum="";SeedConditionResourcesReconciled;SeedConditionValidKubeconfig;
```
ahmedwaleedmalik:
```suggestion
// +kubebuilder:validation:Enum="";SeedConditionResourcesReconciled;SeedConditionValidKubeconfig
```

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -90,6 +90,66 @@ func IsProviderSupported(name string) bool {
...
+
+// SeedConditionType is used to indicate the type of a seed condition. For all condition
+// types, the `true` value must indicate success. All condition types must be registered
+// within the `AllSeedConditionTypes` variable.
```
ahmedwaleedmalik, line 121:
```suggestion
// SeedConditionType is used to indicate the type of a seed condition. For all condition
// types, `true` value will indicate success. All condition types must be registered
// within the `AllSeedConditionTypes` variable.
```

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -90,6 +90,66 @@ func IsProviderSupported(name string) bool {
...
+type SeedConditionType string
+
+const (
+	// SeedConditionValidKubeconfig indicates that the configured kubeconfig for the seed is valid.
+	// The seed-sync controller manages this condition.
+	SeedConditionValidKubeconfig SeedConditionType = "ValidKubeconfig"
```
ahmedwaleedmalik: I'd suggest renaming this to `KubeconfigValid`. Such modifiers are better suited as suffixes in my opinion, like Failed, NotFound, Valid, Invalid, etc.

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -90,6 +90,66 @@ func IsProviderSupported(name string) bool {
...
+
+var AllSeedConditionTypes = []SeedConditionType{
+	SeedConditionResourcesReconciled,
+}
+
+type SeedCondition struct {
+	// Status of the condition, one of True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status"`
+	// Last time we got an update on a given condition.
+	LastHeartbeatTime metav1.Time `json:"lastHeartbeatTime"`
```
ahmedwaleedmalik, line 141: Isn't this usually called `LastProbeTime`?

pkg/controller/master-controller-manager/seed-status-controller/controller.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// ControllerName is the name of this very controller.
+	ControllerName = "seed-status-controller"
```
ahmedwaleedmalik: Alt Text

xrstf: :D My bad. I didn't think.

pkg/apis/kubermatic/v1/datacenter.go
xrstf: Removing "the" would turn this into broken English, no?

pkg/apis/kubermatic/v1/datacenter.go
xrstf: I copied this from the ClusterConditions, which also use LastHeartbeatTime. I think it makes sense to be consistent?

embik: `LastProbeTime` is in Pods for when the last probe ran, but it's a Pod specific thing I believe. Nodes also have `LastHeartbeatTime`.

pkg/apis/kubermatic/v1/datacenter.go
embik: I guess you could drop the "value" after "`true`" and that will make it valid English again, but yeah the suggestion does sound a bit broken.

embik (approved):

PR #9714, opened 2022-04-28
Add ApplicationSecret synchronization between master and seed

### LINKED ISSUE #8378: Secrets for Applications get replicated to all seed clusters
**Acceptance Criteria**
* when a new Secret gets created in the master, it automatically gets created in all seed clusters
* when an Secret gets deleted, it automatically gets removed from all seed clusters
* when an Secret gets updated, it automatically gets updated in all the seed clusters
* secrets are stored in the kubermatic namespace in the seed cluster
* secrets should have an annotation, that makes it possible to distinguish helm auths from git auths

**What this story does not include**
* allowing for custom secrets an end-user can set

SimonTheLeg:

Add ApplicationSecret synchronization between master and seed 

Fixes #8378

**Documentation**:

```release-note
None
```

### DIFF
--- a/pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
+++ b/pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
@@ -41,7 +41,6 @@ import (
 const (
 	ControllerName       = "kkp-application-secret-synchronizer"
 	secretTypeAnnotation = "apps.kubermatic.k8c.io/secret-type"
-	seedNS               = "kubermatic"
 )
 
 type reconciler struct {
@@ -103,7 +102,7 @@ func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, requ
 			return err
 		} else {
 			// handling deletion
-			delSecret := &k8scorev1.Secret{ObjectMeta: v1.ObjectMeta{Name: request.Name, Namespace: seedNS}}
+			delSecret := &k8scorev1.Secret{ObjectMeta: v1.ObjectMeta{Name: request.Name, Namespace: r.namespace}}
 			if err := r.handleDeletion(ctx, log, delSecret); err != nil {
 				return fmt.Errorf("failed to delete secret: %w", err)
 			}
@@ -112,14 +111,13 @@ func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, requ
 	}
 
 	seedsecret := secret.DeepCopy()
-	seedsecret.Namespace = seedNS
 	seedsecret.SetResourceVersion("")
 
 	namedSecretCreatorGetter := []reconciling.NamedSecretCreatorGetter{
 		secretCreator(seedsecret),
 	}
 	err = r.reconcileAllSeeds(ctx, log, seedsecret, func(ctx context.Context, log *zap.SugaredLogger, c ctrlruntimeclient.Client, o ctrlruntimeclient.Object) error {
-		return reconciling.ReconcileSecrets(ctx, namedSecretCreatorGetter, seedNS, c)
+		return reconciling.ReconcileSecrets(ctx, namedSecretCreatorGetter, r.namespace, c)
 	})
 	if err != nil {
 		r.recorder.Eventf(secret, k8scorev1.EventTypeWarning, "ReconcilingError", err.Error())
--- a/pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
+++ b/pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
@@ -25,9 +25,9 @@ import (
 	"k8c.io/kubermatic/v2/pkg/controller/util/predicate"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 
-	k8scorev1 "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -74,7 +74,7 @@ func Add(
 		return fmt.Errorf("failed to construct controller: %w", err)
 	}
 
-	if err := c.Watch(&source.Kind{Type: &k8scorev1.Secret{}}, &handler.EnqueueRequestForObject{}, predicate.ByAnnotation(secretTypeAnnotation, "", false), predicate.ByNamespace(r.namespace)); err != nil {
+	if err := c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}, predicate.ByAnnotation(secretTypeAnnotation, "", false), predicate.ByNamespace(r.namespace)); err != nil {
 		return fmt.Errorf("failed to create watch for secrets: %w", err)
 	}
 
@@ -94,15 +94,15 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 }
 
 func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, request reconcile.Request) error {
-	secret := &k8scorev1.Secret{}
+	secret := &corev1.Secret{}
 
 	var err error
 	if err = r.masterClient.Get(ctx, request.NamespacedName, secret); err != nil {
-		if !kerrors.IsNotFound(err) {
+		if !apierrors.IsNotFound(err) {
 			return err
 		} else {
 			// handling deletion
-			delSecret := &k8scorev1.Secret{ObjectMeta: v1.ObjectMeta{Name: request.Name, Namespace: r.namespace}}
+			delSecret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Name: request.Name, Namespace: r.namespace}}
 			if err := r.handleDeletion(ctx, log, delSecret); err != nil {
 				return fmt.Errorf("failed to delete secret: %w", err)
 			}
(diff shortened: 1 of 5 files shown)


SimonTheLeg:
/hold

I think the code is ready to be reviewed, but I would like to do a test-run on a master-controller, before merging this.
@vgramer when you created the ApplicationDefinition Synchronizer, did you do something similar?

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName       = "application_secret_synchronization_controller"
```
xrstf: Please call controllers `kkp-<controllername>`, e.g. `kkp-application-secret-synchronizer`.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	secretTypeAnnotation = "apps.kubermatic.k8c.io/secret-type"
+	seedNS               = "kubermatic"
```
xrstf: You can use a seedGetter or a configGetter to determine the namespace dynamically, please do not hardcode this namespace into controllers.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+}
+
+func (r *reconciler) handleDeletion(ctx context.Context, log *zap.SugaredLogger, secret *k8scorev1.Secret) error {
+	delfunc := func(ctx context.Context, c ctrlruntimeclient.Client, o ctrlruntimeclient.Object) error {
+		err := c.Get(ctx, types.NamespacedName{Name: secret.Name, Namespace: secret.Namespace}, &k8scorev1.Secret{})
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				log.Infof("Secret %q already deleted", secret.Name)
```
xrstf: Please use structured logging.

SimonTheLeg: I am not sure I full understand. So I mean this already produces a structured message:

```
{"level":"info","time":"2022-04-28T12:21:08.719+0200","caller":"application-secret-synchronizer/controller.go:143","msg":"Secret \"secret-1\" already deleted","resource":"secret-1”}
```

Or do you want here that just the logger is extended to have the secret name as a structured field and not templated into the message? Something like this:

```Go
log := log.With("secret", secret.Name)
log.Info("secret was already deleted")
```

```
{"level":"info","time":"2022-04-28T18:22:43.084+0200","caller":"application-secret-synchronizer/controller.go:144","msg":"secret was already deleted","resource":"secret-1","secret":"secret-1"}
```

If so: do we then generally prefer not to log template data in the `Errorf` section, but rather extract them all into separate fields? (asking because then I will make a pull-request to https://github.com/kubermatic/kubermatic/blob/master/docs/logging.md to clarify)

vgramer:
> /hold
> 
> I think the code is ready to be reviewed, but I would like to do a test-run on a master-controller, before merging this. @vgramer when you created the ApplicationDefinition Synchronizer, did you do something similar?

@SimonTheLeg  to test the controller locally I have used Kind.
1. run KKP in Kind using the hack script.
2. create another Kind cluster with a custom SAN for the api-server, add `host.docker.internal`( as you're also on MacOS)
3. add the second Kind cluster as a  seed( following official doc). for the kubeconfig i just change the hostname of the cluster to `host.docker.internal`

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	for seedName, seedManager := range seedManagers {
+		r.seedClients[seedName] = seedManager.GetClient()
+	}
+
+	c, err := controller.New(ControllerName, masterManager, controller.Options{Reconciler: r, MaxConcurrentReconciles: numWorkers})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	if err := c.Watch(&source.Kind{Type: &k8scorev1.Secret{}}, &handler.EnqueueRequestForObject{}, predicate.ByAnnotation(secretTypeAnnotation, "", false)); err != nil {
```
vgramer: i think you have to also filter on namespace `kubermatic`. otherwise, you may have collisions if 2 secrets on the master cluster have the same name but belong to different namespaces. They will be replicated in the seed in the same namespace.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+			return err
+		}
+
+		return c.Delete(ctx, secret)
+	}
+
+	return r.reconcileAllSeeds(ctx, log, secret, delfunc)
+}
+
+// TODO @vgramer I think this generic version should also work for other controllers like the ApplicationDefinition one. Do you think it makes sense to extract this logic into a central place and also use it for the ADs?
```
vgramer: if there are more than 2 controllers using the exact same pattern, it could make sense.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
```diff
@@ -0,0 +1,172 @@
...
+func secretCreator(s *k8scorev1.Secret) reconciling.NamedSecretCreatorGetter {
+	return func() (name string, create reconciling.SecretCreator) {
+		return s.Name, func(existing *k8scorev1.Secret) (*k8scorev1.Secret, error) {
+			return s, nil
+		}
+	}
+}
+
+func (r *reconciler) handleDeletion(ctx context.Context, log *zap.SugaredLogger, secret *k8scorev1.Secret) error {
+	delfunc := func(ctx context.Context, c ctrlruntimeclient.Client, o ctrlruntimeclient.Object) error {
```
vgramer: if you want to log into this function, it's better to pass the logger as an argument, so the name of the seed is included in the log.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
SimonTheLeg: I am struggling a bit to get this done. So I know how to get the seedGetters and configGetter into my reconciler. However I do not understand how I can get the namespace the seed-controller-manager is running in out of a seedGetter or a configGetter.

While looking through other controllers that need to do the same, ~~I realized they circumvent this by just passing in and empty namespace when creating objects (for example https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/master-controller-manager/cluster-template-synchronizer/controller.go#L121). I assume that the empty namespace then resolves to the seed-managers namespace, because its the default in the passed client.~~ 
Edit @ahmedwaleedmalik and I had another look: Most of them are cluster-resources, so namespace does not matter. However we were not able to find an example for a master-controller that is able to get the namespace of the seed-controller-manager(s). The only ones we saw were using the master-manager namespace (e.g. https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/master-controller-manager/master-constraint-controller/controller.go#L150)

How can you get the seed-manager namespace out of a seedGetter of configGetter?

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
xrstf:
```go
log := log.With("secret", secret.Name)
log.Info("secret was already deleted")
```

This is the way, correct. In general we try to avoid formatting messages (though sometimes, just for readability, it makes sense). However controllers usually put the object they are reconciling as a context field in the logger, because reconcilers usually produce multiple log messages. Having them all be tagged properly is nice.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
SimonTheLeg: oh yes that is a good point. I changed it now to only search in the namespace the controller-manager is installed in.

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
SimonTheLeg: okay. Yeah so far we only have two, so let's leave it at that. I also realized that it can be easy to add too many arguments to this func. Like possibly some controllers need more information about the seed controller and then you would have to add additional arguments, which are not required by others

⟹ I have removed the comment so far

pkg/controller/master-controller-manager/application-secret-synchronizer/controller.go
ahmedwaleedmalik: > You can use a seedGetter or a configGetter to determine the namespace dynamically, please do not hardcode this namespace into controllers.

@xrstf I don't think seedGetter is required here. Adding a flag to make that namespace configurable should be enough

ahmedwaleedmalik: > Adding a flag 

NOT a flag but passing `namespace` as a parameter to the `Add` method of the controller

xrstf: Maybe it was too simple?

### For the seed-ctrl-mgr

```
seed, _ := seedGetter()
fmt.Println("This controller is running in %s.\"n, seed.Namespace)

// or...

config, _ := configGetter(ctx)
fmt.Println("This controller is running in %s.\"n, config.Namespace)
```

### For the master-ctrl-mgr

```
seeds, _ := seedsGetter()
if len(seeds) > 0 {
  fmt.Println("This controller is running in %s.\"n, seeds[0].Namespace)
}

// or...

config, _ := configGetter(ctx)
fmt.Println("This controller is running in %s.\"n, config.Namespace)
```

Effectively every KKP setup runs in a single namespace. If you install it into "foo" on the master, then "foo" will also be the namespace on every seed cluster. In 99,999% of cases, KKP is installed into the "kubermatic" namespace, but I (personally :D) would like to try to not hardcode this into KKP.

The namespace info flows from the KKP operator (which gets explicitly told about the namespace via https://github.com/kubermatic/kubermatic/blob/master/cmd/kubermatic-operator/main.go#L63) down the master-ctrl-mgr and seed-ctrl-mgrs, which are being reconciled based on the namespace that the operator is running in, i.e. https://github.com/kubermatic/kubermatic/blob/master/cmd/seed-controller-manager/options.go#L133 is being set.

You can either hand this flag into your controller or take the namespace from the seed object (effectively they will always be the same, because the seedGetter we're constructing is based on the `-namespace` flag).

SimonTheLeg: aaaaah now I finally get it. I did not know you could directly use the Namespace from the ObjectMeta of a seed object. And with the additional info that everything always is in a single namespace this makes sense now.

-----

> If you install it into "foo" on the master, then "foo" will also be the namespace on every seed cluster

And with this piece of information, I understand now why it was done the way it is in the the master-constraint-controller.


Okay now I understand all the options. Now onto picking one! Thanks for all the help :)

xrstf: As to why we have a wild mix of namespaced and cluster-wide resources: History & RBAC.

Some of our resources are owners to others and Kubernetes doesn't allow cross-namespace owner refs. So this already necessitates some global resources.

I personally like namespaced resources and so I made the KubermaticConfiguration namespaced, even though we never fully 100% achieved the lofty goal of running multiple KKP installations in the same Kubernetes cluster (something we kind of somewhat a bit tried and did in the old CI cluster, before we used kind).

So even though it might look a bit chaotic, it's only slightly chaotic and if you tried to refactor it, you would notice the inter-dependencies between objects and that there's really no way to change any of that now :D

ahmedwaleedmalik (approved):
/approve ⚡ 
/hold for @xrstf to take a look as well since he has already reviewed this 🔨

ahmedwaleedmalik:
@xrstf can you please take another look at this PR? For me it's good to be merged :D

xrstf:

vCenter isssssues.

PR #9715, opened 2022-04-28
Resource quota webhook providers - AWS, GCP, Azure, KubeVirt, vSphere, Openstack

### LINKED ISSUE #9626: Add resource quota calculation per provider in machine webhook
For each provider implement getting requested node size from a machine

lsviben:
adds logic for getting providers(AWS, GCP, Azure, KubeVirt, vSphere, Openstack) machine resource details to the resource quota machine webhook.

For now getting the ResourceQuota for the project is still hardcoded (because ResourceQuota is not implemented yet), so the resource calculation is skipped until the full implementation.

Fixes #9626

```release-note
NONE
```

### DIFF
--- a/pkg/ee/validation/machine/resourcequota.go
+++ b/pkg/ee/validation/machine/resourcequota.go
@@ -61,7 +61,12 @@ func ValidateQuota(ctx context.Context, log *zap.SugaredLogger, seedClient, user
 	case types.CloudProviderGoogle:
 		quotaReq, err = getGCPResourceRequirements(ctx, userClient, config)
 		if err != nil {
-			return fmt.Errorf("error getting aws quota request: %w", err)
+			return fmt.Errorf("error getting gcp quota request: %w", err)
+		}
+	case types.CloudProviderAzure:
+		quotaReq, err = getAzureResourceRequirements(ctx, userClient, config)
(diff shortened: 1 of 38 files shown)


pkg/ee/validation/machine/resourcequota.go
```diff
@@ -51,7 +53,37 @@ func ValidateQuota(ctx context.Context, log *zap.SugaredLogger, seedClient ctrlr
...
+		if err != nil {
+			return fmt.Errorf("error getting azure resource requirements: %w", err)
+		}
+	case types.CloudProviderKubeVirt:
+		quotaReq, err = getKubeVirtResourceRequirements(ctx, userClient, config)
+		if err != nil {
+			return fmt.Errorf("error getting kubevirt resource requirements: %w", err)
+		}
+	case types.CloudProviderVsphere:
+		quotaReq, err = getVsphereResourceRequirements(config)
```
kron4eg, line 79: I'd expect that vSphere might required customer CABundle too.

lsviben: we probably would, but for vSphere we dont need to contact vSphere at all, we can just take the resource details from the machine spec

PR #7073, opened 2021-05-19
add aws-node-termination-handler as an addon

moadqassem:
**What this PR does / why we need it**:
Adds aws-node-termination-handler as an addon for aws spot instances.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #6723

**Special notes for your reviewer**:
Please have a look on this PR as well: 
**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
AWS Spot instances support
```

PR #9716, opened 2022-04-28
update aws-node-termination-handler addon

xrstf:
This updates the addon based on the current Helm chart release (https://github.com/aws/aws-node-termination-handler/releases/tag/v1.16.2). I tried to figure out what changes we made originally in #7073 and except for the customized `k8c.io` label (see https://github.com/kubermatic/kubermatic/pull/7073#discussion_r635564043) it seems to have worked out fine.

The labels on the DaemonSet change, but not the labelSelector, so I guess this is fine?

I also removed some aaaaaaancient entries in our gitignore.

```release-note
Update aws-node-termination-handler addon to v1.16.2
```


addons/values-aws-node-termination-handler.yaml
```diff
@@ -0,0 +1,13 @@
+fullnameOverride: "aws-node-termination-handler"
```
ahmedwaleedmalik: Boilerplate missing

ahmedwaleedmalik (approved):
/approve ⚡

PR #9717, opened 2022-04-28
Add validations for backup destination names in seed

### LINKED ISSUE #9707: etcd Backup Settings: Lowercase validation for etcd Backup Credentials

As a end user
I would like to enforce lower case letters validation during Credentials creation time as part of etcd Backup Settings from the UI
in order to avoid the end user from entering the credentials with UPPER CASE and in turn avoid getting below error
```"Invalid value "XXX-XXX" : a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')```

### Solution details

- Requirement 1
- Requirement 2
- Requirement 3

### Use cases

It would be nice to have validation from UI for the inputs provided, since the end user will not be aware until the error occurs.

### PR DESCRIPTION
**Pull request by ahmedwaleedmalik**

This PR intends to add validation for the destination name for etcd backups. Default destination or some specific backup destination name from the list of destinations can be used to create the corresponding secret for storing backup credentials.

Fixes #9707

**Documentation**:

```release-note
NONE
```


cmd/kubermatic-api/swagger.json
```diff
@@ -23423,6 +23423,7 @@
         "defaultDestination": {
           "description": "DefaultDestination marks the default destination that will be used for the default etcd backup config which is\ncreated for every user cluster. Has to correspond to a destination in Destinations.\nIf removed, it removes the related default etcd backup configs.",
           "type": "string",
+          "pattern": "=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`",
```
ahmedwaleedmalik: How can we avoid this? 🤷‍♂️

ahmedwaleedmalik: @xrstf @maciaszczykm Thoughts? 

I also recently came across this issue that if we have the following marker on some resource then the swagger spec generation fails.

`// +kubebuilder:default:=true`

PR #9719, opened 2022-04-28
refactor: ensure controller names comply with our convention

ahmedwaleedmalik:

The controller name should be prefixed with 'kkp-'. Having the suffix 'controller' should only be used when the name doesn't sound meaningful without it.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/master-controller-manager/kubeone/controller.go
+++ b/pkg/controller/master-controller-manager/kubeone/controller.go
@@ -59,8 +59,8 @@ import (
 )
 
 const (
-	// ControllerName is the name of this very controller.
-	ControllerName = "kubeone_controller"
+	// This controller is responsible for managing the lifecycle of KubeOne clusters within KKP.
+	ControllerName = "kkp-kubeone-manager"
 
 	// ImportAction is the action to import kubeone cluster.
 	ImportAction = "import"
--- a/pkg/controller/master-controller-manager/kubeone/controller.go
+++ b/pkg/controller/master-controller-manager/kubeone/controller.go
@@ -60,7 +60,7 @@ import (
 
 const (
 	// This controller is responsible for managing the lifecycle of KubeOne clusters within KKP.
-	ControllerName = "kkp-kubeone-manager"
+	ControllerName = "kkp-kubeone-controller"
 
 	// ImportAction is the action to import kubeone cluster.
 	ImportAction = "import"
--- a/pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go
+++ b/pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	ControllerName = "kkp-seed-controller-lifecycle-manager"
+	ControllerName = "kp-seed-lifecycle-controller"
 
 	// We must only enqueue this one key.
 	queueKey = ControllerName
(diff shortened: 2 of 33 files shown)


pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -59,8 +59,8 @@ import (
 )
 
 const (
-	// ControllerName is the name of this very controller.
-	ControllerName = "kubeone_controller"
+	// This controller is responsible for managing the lifecycle of KubeOne clusters within KKP.
+	ControllerName = "kkp-kubeone-manager"
```
embik: Why not `kkp-kubeone-controller`? It's not a `manager` afaik. That's a different thing in controller-runtime land.

ahmedwaleedmalik: Does that make a difference? What's a `manager` according to you? :)

embik: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager this :D

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
```diff
@@ -39,7 +39,9 @@ import (
 )
 
 const (
-	ControllerName = "kubermatic_flatcar_controller"
+	// This controller is responsible for ensuring that the flatcar-linux-update-operator is installed when we have a healthy(running) flatcar
+	// node in our cluster.
+	ControllerName = "kkp-flatcar-update-operator-manager"
```
embik: Should IMHO drop the `manager` suffix and be something like `kkp-flatcar-operator-controller`.

embik (changes requested):

pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go
```diff
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	ControllerName = "seedcontroller_lifecycle_manager"
+	ControllerName = "kkp-seed-controller-lifecycle-manager"
```
embik: Also this one, maybe `kkp-seed-lifecycle-controller`.

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
ahmedwaleedmalik: I TBH think `-manager` makes so much sense here. Because this controller will be responsible for managing the lifecycle of the `flatcar-update-operator`. I know it might convey the notion that manager = controller-manager. But it's not the case here.

pkg/controller/shared/seed-controller-lifecycle/seed_controller_lifecycle.go
ahmedwaleedmalik: Same as https://github.com/kubermatic/kubermatic/pull/9719#discussion_r861508930

pkg/controller/user-cluster-controller-manager/flatcar/flatcar_controller.go
embik: @xrstf any thoughts? Maybe I'm pushing back too much against the `-manager` suffix, but I really dislike it.

xrstf: To me a `-manager` is something that contains multiple controllers, so I would also argue that we should not put `-manager` as the suffix for a single controller.

ahmedwaleedmalik: Then we have a decision ✌️

embik (approved):
/approve

xrstf (approved):
/approve

PR #9721, opened 2022-04-29
applications: Helm client

### LINKED ISSUE #9549: Application: implement helm client

vgramer:
This Pr adds helm client to interacting with Helm API for download, install, upgrade and uninstall Helm Chart. This client will be used to reconcile applications in user-clusters.

Fixes #9549

To avoid using locking, each time we reconcile an application we will create a new client with HelmSetting pointing to a temporary directory. Chart and helm repository caches will be downloaded in this directory. When reconciliation is done, we remove this directory.

leftover for another PR:
- integration test for `Install` and `Upgrade` and `InstallOrUpgrade` functions

**Documentation**:

```release-note
NONE
```


pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// Namespace where chart will be installed or updated.
+	targetNamespace string
+
+	logger *zap.SugaredLogger
+}
+
+func NewClient(restClientGetter genericclioptions.RESTClientGetter, settings HelmSettings, targetNamespace string, logger *zap.SugaredLogger) (*HelmClient, error) {
+	// Even if namespace is set in the actionConfig.init() function, upgrade action take the namespace from RESTClientGetter.
+	// If the namespace are different, the release will be installed in the namespace set in the RESTClientGetter but the
```
xrstf: _namespaces_ are different

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	// release information will be stored in the targetNamespace which lead to a release we can not uninstall with helm.
```
xrstf: which _leads_

which cannot _be uninstalled_ with _Helm_

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	if kcNamspace, _, err := restClientGetter.ToRawKubeConfigLoader().Namespace(); err != nil || kcNamspace != targetNamespace {
```
xrstf: Can you split this into 2 conditions? The error message that is generated here is confusing as it contains 2 possible, distinct error cases.

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	var out strings.Builder
+	chartDownloader := downloader.ChartDownloader{
+		Out:              &out,
+		Verify:           downloader.VerifyNever,
+		RepositoryConfig: h.settings.RepositoryConfig,
+		RepositoryCache:  h.settings.RepositoryCache,
+		Getters:          h.getterProviders,
+		// todo credentials
+	}
+	// todo note: we may want to check the verificaton return by chartDownloader.DownloadTo. for the moment its set to downloader.VerifyNever in struct init
```
xrstf: _it's_

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	chartRef := repoName + "/" + chartName
+	chartLoc, _, err := chartDownloader.DownloadTo(chartRef, version, dest)
+	if err != nil {
+		h.logger.Errorf("failed to download chart '%s' in version '%s'. log='%s'", chartRef, version, out.String())
```
xrstf: You have a Zap logger, you can use structured logging everywhere :)

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+// Otherwise it updates the chart.
+// charLoc is the path to the chart archive (eg /tmp/foo/apache-1.0.0.tgz) or folder containning the chart (eg /tmp/mychart/apache).
+func (h HelmClient) InstallOrUpgrade(chartLoc string, releaseName string, values map[string]interface{}) (*release.Release, error) {
+	if _, err := h.actionConfig.Releases.Last(releaseName); err != nil {
+		return h.Install(chartLoc, releaseName, values)
+	}
+	return h.Upgrade(chartLoc, releaseName, values)
+}
+
+// Install the chart located at chartLoc into targetNamespace if the chart is already installed, an error is returned.
```
xrstf: Please rephrase this a bit, it reads as "installs the chart if it's already installed" and "an error is returned", which is confusing :D A semicolon could help, like

> Install the chart located at chartLoc; if the chart is installed, an error is returned

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+
+	rel, err := installClient.Run(chartToInstall, values)
+	if err != nil {
+		return nil, err
+	}
+	return rel, nil
+}
+
+// Upgrade the chart located at chartLoc into targetNamespace. if chart is not already installed an error is returned.
+// charLoc is the path to the chart archive (eg /tmp/foo/apache-1.0.0.tgz) or folder containning the chart (eg /tmp/mychart/apache).
```
xrstf: _containing_

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	return rel, nil
+}
+
+// Uninstall the release in targetNamespace.
+func (h HelmClient) Uninstall(releaseName string) (*release.UninstallReleaseResponse, error) {
+	uninstallClient := action.NewUninstall(h.actionConfig)
+	return uninstallClient.Run(releaseName)
+}
+
+// buildDependencies add missing repositories and then do a Helm dependency build (ie download the chart dependencies
```
xrstf: _adds_ and _does_

also, it's _i.e._ and _e.g._, not "ie" and "eg" ;)

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+		RegistryClient:   regClient,
+		Debug:            false,
+		Verify:           downloader.VerifyNever,
+		SkipUpdate:       true,
+		// todo credentials
+	}
+
+	chartToInstall, err := loader.Load(chartLoc)
+	if err != nil {
+		return nil, fmt.Errorf("can not load char: %w", err)
```
xrstf: _chart_

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	uninstallClient := action.NewUninstall(h.actionConfig)
+	return uninstallClient.Run(releaseName)
+}
+
+// buildDependencies add missing repositories and then do a Helm dependency build (ie download the chart dependencies
+// from repositories into "charts" folder).
+func (h HelmClient) buildDependencies(chartLoc string) (*chart.Chart, error) {
+	regClient, err := registry.NewClient() // todo credentials
+	if err != nil {
+		return nil, fmt.Errorf("can not initialize registry client %w", err)
```
xrstf: Colon missing before the error placeholder (i.e. `client: %w`)

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("can not load char: %w", err)
+	}
+
+	fi, err := os.Stat(chartLoc)
+	if err != nil {
+		return nil, fmt.Errorf("can not find chart at `%s': %w", chartLoc, err)
+	}
+
+	// If we got the chart from the filesystem (ie cloned from a git repository). We have to do build dependencies because
```
xrstf: This should be one sentence. No full stop after the "(i.e. cloned from a git repository)"

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+
+		// Equivalent of helm dependency build.
+		err = man.Build()
+		if err != nil {
+			h.logger.Errorf("can not build dependencies for chart '%s'. log='%s'", chartLoc, out.String())
+			return nil, fmt.Errorf("can not build dependencies: %w", err)
+		}
+	}
+	h.logger.Debugf("succesully build dependencies for chart '%s'. log='%s'", chartLoc, out.String())
```
xrstf: _successfully built_

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+	return chartToInstall, nil
+}
+
+// ensureRepository add the repository url if it's not already exist and download the last index file.
```
xrstf: _adds_

_if it doesn't exist_ or _if it's not already existing_

_downloads_

"latest" might be better than "last"

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,311 @@
...
+		// todo credential  ???
+	}
+
+	// Ensure we have the last version of the index file.
+	chartRepo, err := repo.NewChartRepository(desiredEntry, h.getterProviders)
+	if err != nil {
+		return "", err
+	}
+	// Constructor of the ChartRepository uses the default Helm cache. To avoid polluting the local repository when executing
+	// TU, we override it here.
```
xrstf: What is "TU"?

pkg/applications/helmclient/doc.go
```diff
@@ -0,0 +1,21 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+Package helmclient contains tools to interact with Helm API for download, install, upgrade and uninstall Helm Chart.
```
xrstf: _downloading, installing, upgrading and uninstalling_

pkg/applications/helmclient/client.go
vgramer: oops, my bad it's the french abbreviation for Unit Test. But since we don't use locking, we have to override the cache repository location anyway... 

I have updated the comment.

pkg/applications/helmclient/client.go
vgramer: indeed much better with  structured logging :)

xrstf:
Congratulations! :tada: 

Your PR has been chosen among hundreds of candidates to serve as a test for our CI cluster. Do not be alarmed if the following tests fail, it's most likely not your fault:

/test pre-kubermatic-e2e-aws-ubuntu-1.21

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	h.logger.Debugw("successfully built dependencies", "chart", chartLoc, "log", out.String())
+	return chartToInstall, nil
+}
+
+// ensureRepository adds the repository url if it doesn't exist and downloads the latest index file.
+// The repository is added with the name helm-manager-$(sha256 url).
+func (h HelmClient) ensureRepository(url string) (string, error) {
+	repoFile, err := repo.LoadFile(h.settings.RepositoryConfig)
+	if err != nil && !os.IsNotExist(errors.Cause(err)) {
```
xrstf: Do we really need `pkg/errors` as a dependency for this one single function call?`Shouldn't IsNotExist() automatically take care of unwrapping things?

vgramer: good catch :) i can is `errors.Is()` from go standard library instead of  `pkg/errors`

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,332 @@
...
+// charLoc is the path to the chart archive (e.g. /tmp/foo/apache-1.0.0.tgz) or folder containing the chart (e.g. /tmp/mychart/apache).
+func (h HelmClient) InstallOrUpgrade(chartLoc string, releaseName string, values map[string]interface{}) (*release.Release, error) {
+	if _, err := h.actionConfig.Releases.Last(releaseName); err != nil {
+		return h.Install(chartLoc, releaseName, values)
+	}
+	return h.Upgrade(chartLoc, releaseName, values)
+}
+
+// Install the chart located at chartLoc into targetNamespace. If the chart was already installed, an error is returned.
+// charLoc is the path to the chart archive (eg /tmp/foo/apache-1.0.0.tgz) or folder containing the chart (eg /tmp/mychart/apache).
```
SimonTheLeg:
```suggestion
// charLoc is the path to the chart archive (eg /tmp/foo/apache-1.0.0.tgz) or folder containing the chart (e.g. /tmp/mychart/apache).
```

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	h.logger.Debugw("successfully downloaded chart", "chart", chartRef, "version", version, "log", out.String())
+	return chartLoc, nil
+}
+
+// InstallOrUpgrade installs the chart located at chartLoc into targetNamespace if it's not already installed.
+// Otherwise it upgrades the chart.
+// charLoc is the path to the chart archive (e.g. /tmp/foo/apache-1.0.0.tgz) or folder containing the chart (e.g. /tmp/mychart/apache).
+func (h HelmClient) InstallOrUpgrade(chartLoc string, releaseName string, values map[string]interface{}) (*release.Release, error) {
```
SimonTheLeg, line 165: personally I think these funcs should take in a context as well, so they can be cancelled. Because at the moment `installClient.Run` will create a background context that is unavailable to us.

Luckily they all also expose a RunWithContext func, so it should be possible to just pass in a context into our funcs and then use that func.

pkg/applications/helmclient/client.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	repoName, err := computeRepoName(url)
+	if err != nil {
+		return "", fmt.Errorf("can not compute repository's name for '%s': %w", url, err)
+	}
+	desiredEntry := &repo.Entry{
+		Name: repoName,
+		URL:  url,
+		// todo credential  ???
```
SimonTheLeg: If I read the helm code correctly, we could handle the complete auth in the getters themselves:
Example for HTTP:
* HTTP Getter can be created with getter Options set https://github.com/helm/helm/blob/main/pkg/getter/httpgetter.go#L101
* and then any form of credentials can be set like BasicAuth, TLSClientConfig, ... https://github.com/helm/helm/blob/657850e44b880cca43d0606ebf5a54eb75362c3f/pkg/getter/getter.go#L63 And then that should be passed along every request with 

And since funcs like `NewChartRepository` also always pass the getters along, that should work. 
I think the big advantage would be that we could support both HTTP as well as OCI Getters

pkg/applications/helmclient/client_test.go
```diff
@@ -0,0 +1,411 @@
...
+	tf := cmdtesting.NewTestFactory().WithNamespace(defaultNs)
+	defer tf.Cleanup()
+
+	_, err := NewClient(tf, settings, "another-ns", log)
+	if err == nil {
+		t.Fatalf("helmclient.NewClient() should fail when RESTClientGetter namespace is different than targetNamespace : %s", err)
+	}
+	expectedErrMsg := "namespace set in RESTClientGetter should be the same as targetNamespace"
+	if !strings.Contains(err.Error(), expectedErrMsg) {
+		t.Fatalf("helmclient.NewClient() fails for the wrong reason. expected error message: '%s' to contains: '%s'", err, expectedErrMsg)
```
SimonTheLeg:
```suggestion
		t.Fatalf("helmclient.NewClient() fails for the wrong reason. expected error message: '%s' to contain: '%s'", err, expectedErrMsg)
```

pkg/applications/helmclient/client_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+				helmClient, err := NewClient(tf, settings, defaultNs, log)
+				if err != nil {
+					t.Fatalf("can not init helm Client: %s", err)
+				}
+
+				chartLoc, err := helmClient.DownloadChart(tc.repoUrl, tc.chartName, tc.chartVersion, downloadDest)
+
+				if (err != nil) != tc.wantErr {
+					t.Fatalf("DownloadChart() error = %v, wantErr %v", err, tc.wantErr)
```
SimonTheLeg, line 144: I am not entirely sure, but why are fataling here? Wouldn't this just be a normal Error? I mean yes we would finish this test case, but we would continue with the others, right?

pkg/applications/helmclient/client_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
+
+	cmdtesting "k8s.io/kubectl/pkg/cmd/testing"
+	"sigs.k8s.io/yaml"
+)
+
+const defaultNs = "default"
+
+func TestHelmClient_NewShouldFailWhenRESTClientGetterNamespaceIsDifferentThanTargetNamespace(t *testing.T) {
```
SimonTheLeg: I am not sure about the func names throughout this file: I saw that we rarely use underscore prefixes in our code for test funcs.

Given that it is a small package, I think It is fine to drop the `HelmClient` part of the prefix

SimonTheLeg (changes requested):

SimonTheLeg:
> leftover for another PR:
> * remove examplechart-0.1.0.tgz which is used for testing. I think we can easily build it in tests.

Question is, if we would not want to do that right away. Because I think it is more of a hassle to remove the .tgz file from git history again, than it is to add a small func in the test that zips up the examplechart dir. Or am I missing something? But of course that's just my 2 cents on it. No blocker for getting this merged :)

pkg/applications/helmclient/client_test.go
vgramer: if the download failed, the next assertions will fail too. To avoid confusion message we stop this test case. Other test cases will still be launched even if one fails with fatal.

pkg/applications/helmclient/client.go
vgramer: done :)

pkg/applications/helmclient/client_test.go
vgramer: done

pkg/applications/helmclient/client_test.go
vgramer: done :)

pkg/applications/helmclient/client_test.go
SimonTheLeg: understood

vgramer:
/hold 

fix go dependency problem need to be tested

pkg/applications/helmclient/client.go
vgramer: i think so too... I removed the it :)

SimonTheLeg (approved):

PR #9722, opened 2022-04-29
Enable leader election for kubermatic-operator chart by default

embik:
For your consideration. We discussed this recently and wondered if the operator does leader elections - Turns out, it does, just not by default. In my opinion, we should enable this, to prevent situations where two operators from different versions are working against each other.

Fixes #

**Documentation**:

```release-note
kubermatic-operator is deployed with leader election by default, can be disabled from Chart values
```

### DIFF
--- a/charts/kubermatic-operator/templates/deployment.yaml
+++ b/charts/kubermatic-operator/templates/deployment.yaml
@@ -56,6 +56,9 @@ spec:
         - -log-debug=true
         - -v=8
         {{- end }}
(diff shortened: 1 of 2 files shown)


xrstf:
/test pre-kubermatic-etcd-launcher-e2e

For demo purposes only.

xrstf:
This is technically already prevented by the Recreate strategy and us only using a single replica, but sure, we can enable leader election. Even though it will make the operator startup slower :-(

/approve
/lgtm

PR #9734, opened 2022-05-03
auto-detect default namespaces/webhooks when cleaning up in-between conformance tests

xrstf:
Now that we have a webhook for applications inside userclusters, we would have to adjust our cleanup code to make sure to not remove this webhook. Removing it will just make KKP reconcile and bring it back, so we will never truly succeed in getting rid of it (and that's a good thing, that means KKP is working as intended).

This PR makes it so that we do not have to explicitly maintain a list of default namespaces and webhooks, but instead we just check the cluster state after it was created and consider whatever we find to be the "default contents".

```release-note
NONE
```


embik (approved):
/approve

PR #9736, opened 2022-05-04
Support Kubernetes 1.24

### LINKED ISSUE #9735: Support deploying Kubernetes 1.24 clusters

Kubernetes 1.24 [is out](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md) and we should support it in KKP for user cluster creation.

### Solution details

- Make sure support for docker is removed (dockershim is removed in 1.24)
- Investigate `LegacyServiceAccountTokenNoAutoGeneration` feature and what it means for KKP secrets (possibly manual `Secret` creation for all components is required), see https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets
- Check [deprecations](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#deprecation): `Service.Spec.LoadBalancerIP`, couple of `kube-apiserver` flags, etc.

### Use cases

- Kubernetes releases lose upstream support fairly quickly and we want to support an up-to-date set of Kubernetes versions.

embik:

Support for [Kubernetes 1.24](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md) in user clusters.

Currently missing are:
- cluster-autoscaler image
- cinder-csi-driver image
- probably a bunch of other images as well; 1.23-verified versions might very well work with 1.24 as well though

Fixes #9735

**Documentation**:

```release-note
Add support for Kubernetes 1.24
```


embik:
Needs a new machine-controller release.

addons/kubelet-configmap/kubelet-configmap.yaml
```diff
@@ -375,3 +375,94 @@ data:
     - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
     - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
     volumeStatsAggPeriod: 1m0s
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kubelet-config-1.24
```
xmudrii: Please note that this has been changed. `kubelet-config` ConfigMaps are no longer versioned (there's no `x.yy` suffix).

Changelog has some more details about this: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#deprecation

xmudrii (changes requested):

addons/kubelet-configmap/kubelet-configmap.yaml
embik: I missed that, thank you. Not sure how to gracefully handle this but I will take a look.

addons/csi/openstack/controllerplugin.yaml
```diff
@@ -28,6 +28,9 @@
 {{ if eq .Cluster.MajorMinorVersion "1.23" }}
 {{ $version = "v1.23.0"}}
 {{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.24" }}
+{{ $version = "v1.23.0"}}
```
xmudrii, line 32: Can we somehow keep track of things we need to update? For example, we have [this issue](https://github.com/kubermatic/kubeone/issues/2011) for KubeOne.

addons/csi/openstack/nodeplugin.yaml
```diff
@@ -28,7 +28,11 @@
 {{ if eq .Cluster.MajorMinorVersion "1.23" }}
 {{ $version = "v1.23.0"}}
 {{ end }}
+{{ if eq .Cluster.MajorMinorVersion "1.24" }}
+{{ $version = "v1.23.0"}}
+{{ end }}
 
+{{ if not (eq $version "UNSUPPORTED") }}
```
xmudrii, line 35: Should we have something similar for other files (e.g. `addons/csi/openstack/controllerplugin.yaml`)?

pkg/controller/operator/defaults/defaults.go
```diff
@@ -221,6 +221,8 @@ var (
 			newSemver("v1.22.7"),
 			// Kubernetes 1.23
 			newSemver("v1.23.5"),
+			// Kubernetes 1.24
+			newSemver("v1.24.0"),
```
xmudrii, line 225: I guess we should change default to v1.23 since we always recommend using n-1 minor release.

pkg/controller/operator/defaults/defaults.go
```diff
@@ -294,23 +302,12 @@ var (
...
-				Condition: kubermaticv1.ExternalCloudProviderCondition,
+				Condition: kubermaticv1.NonAMD64WithCanalAndIPVSClusterCondition,
 				Operation: kubermaticv1.UpdateOperation,
 			},
-			{
-				Provider:  kubermaticv1.VSphereCloudProvider,
-				Version:   "1.24.*",
-				Condition: kubermaticv1.ExternalCloudProviderCondition,
-				Operation: kubermaticv1.SupportOperation,
-			},
```
xmudrii: Maybe we should reject CCM/CSI migration until there's no CCM for that minor release. We can allow upgrades, but generally, CCM migrations are more "risky", so we should proper support.

pkg/controller/user-cluster-controller-manager/resources/resources/mla/promtail/daemonset.go
```diff
@@ -185,6 +185,11 @@ func DaemonSetCreator(overrides *corev1.ResourceRequirements, registryWithOverwr
 					Key:      "node-role.kubernetes.io/master",
 					Operator: corev1.TolerationOpExists,
 				},
+				{
+					Effect:   corev1.TaintEffectNoSchedule,
+					Key:      "node-role.kubernetes.io/master",
+					Operator: corev1.TolerationOpExists,
+				},
```
xmudrii: Isn't this a duplicate of the toleration above? Also, for what do we use this master toleration? Note that Kubernetes/kubeadm has switched to `node-role.kubernetes.io/control-plane` starting with v1.24 (I'm not sure are we affected by this at all).

pkg/resources/kubernetes-dashboard/deployment.go
```diff
@@ -188,9 +188,12 @@ func getDashboardVersion(clusterVersion semver.Semver) (string, error) {
 	case "1.21":
 		return "v2.4.0", nil
 	case "1.22":
+		// nolint:goconst
```
xmudrii: Nit: can we make a const instead of ignoring linter?

pkg/util/kubectl/kubectl.go
```diff
@@ -37,9 +37,12 @@ func BinaryForClusterVersion(version *semverlib.Version) (string, error) {
 	case 21:
 		binary = "kubectl-1.21"
 	case 22:
+		// nolint:goconst
```
xmudrii: Nit: can we use a const here?

xmudrii (changes requested):
In addition to the comments I left, what about the [cluster-autoscaler addon](https://github.com/kubermatic/kubermatic/blob/04d31bfad264c45db2a49674201df5abc205736f/addons/cluster-autoscaler/cluster-autoscaler.yaml#L25-L27)?

pkg/resources/kubernetes-dashboard/deployment.go
embik: I did not want to make it a const because it is one of the TODOs that you mentioned in a comment above. So it’s going to go away anyway and then the constant doesn’t make much sense anymore.

pkg/controller/user-cluster-controller-manager/resources/resources/mla/promtail/daemonset.go
embik: I think I wanted to add that new toleration but missed Updating it.

addons/csi/openstack/controllerplugin.yaml
embik: Good point, I will raise a follow-up tracker issue.

embik: #9818

pkg/controller/operator/defaults/defaults.go
embik: Agreed, I added a blank CCM migration block.

addons/rbac/allow-kubeadm-join-configmap.yaml
```diff
@@ -21,6 +21,7 @@ rules:
 - apiGroups:
   - ""
   resourceNames:
+  - kubelet-config
```
xrstf, line 24: Any reason why the 1.24 configmap doesn't have a version suffix? Is it generic for all 1.24+ releases?

embik: https://github.com/kubermatic/kubermatic/pull/9736#discussion_r867901759

embik: they changed it for 1.24 to drop the suffix.

.prow/provider-vsphere.yaml
```diff
@@ -103,7 +103,7 @@ presubmits:
             limits:
               memory: 4Gi
 
-  - name: pre-kubermatic-e2e-vsphere-ubuntu-1.23-datastore-cluster
+  - name: pre-kubermatic-e2e-vsphere-ubuntu-1.24-datastore-cluster
```
xmudrii, line 106: Please change the Kubernetes version for this job.

xmudrii (changes requested):

.prow/provider-vsphere.yaml
embik: Good catch!

xmudrii (approved):
/lgtm

PR #9746, opened 2022-05-06
show konnectivity health in extended cluster health

PratikDeoghare:
Currently, OpenVPN health status is shown even when the cluster is using Konnectivity. This can be confusing while debugging cluster issues. 
This PR shows the status depending on which one is actually deployed. 

```release-note
NONE
```


pkg/controller/seed-controller-manager/kubernetes/health.go
```diff
@@ -45,16 +45,25 @@ func (r *Reconciler) clusterHealth(ctx context.Context, cluster *kubermaticv1.Cl
 		resources.ApiserverDeploymentName:             {healthStatus: &extendedHealth.Apiserver, minReady: 1},
 		resources.ControllerManagerDeploymentName:     {healthStatus: &extendedHealth.Controller, minReady: 1},
 		resources.SchedulerDeploymentName:             {healthStatus: &extendedHealth.Scheduler, minReady: 1},
-		resources.OpenVPNServerDeploymentName:         {healthStatus: &extendedHealth.OpenVPN, minReady: 1},
 		resources.UserClusterControllerDeploymentName: {healthStatus: &extendedHealth.UserClusterControllerManager, minReady: 1},
 	}
 
+	if r.features.Konnectivity && cluster.Spec.ClusterNetwork.KonnectivityEnabled != nil && *cluster.Spec.ClusterNetwork.KonnectivityEnabled {
+		// because konnectivity server is in apiserver pod
+		healthMapping[resources.KonnectivityDeploymentName] = &depInfo{healthStatus: &extendedHealth.Apiserver, minReady: 1}
```
embik, line 53: If it's the same health status as the apiserver because it's part of the deployment, couldn't we just omit this health status?

PratikDeoghare: If we omit it then we get the status for openvpn but not for konnectivity.

I prefer to see it there instead of having to deduce it from apiserver status. 

Shouldn't the decision to show status be independent of where/how it is deployed?

embik: Yeah that's fair. It's something mainly exposed for cluster admins.

embik (approved):
/approve

PR #9747, opened 2022-05-06
Make CRDs more accessible

xrstf:
Previously we moved the CRDs to `pkg/validation/openapi/crd/`, mostly because that's the only place in our codebase that used the CRDs: the validation logic.

However soon the KKP operator will install CRDs on seed clusters and for that it also needs access to the CRDs. In this case IMHO it doesn't feel right to have the CRDs in a  "validation" package. So this PR moves them to `pkg/crd/` (yes, the directory we used in the ancient times before we had `pkg/apis/`), adds some utilities to make it easy to retrieve a single _or all_ CRDs and adjusts the validation package to use the new helpers.

This PR also removes support for v1beta1 CRDs, as we do not generate that API version anymore and I don't see why our code that only handles our CRDs should handle an unused, outdated API version :-)

```release-note
NONE
```


pkg/crd/fs.go
```diff
@@ -0,0 +1,129 @@
...
+	return groups.List(), nil
+}
+
+// CRDForType returns the CRD for a given object or returns an error if the
+// object is not using one of the known types (*.k8c.io).
+func CRDForObject(obj runtime.Object) (*apiextensionsv1.CustomResourceDefinition, error) {
+	return CRDForGVK(obj.GetObjectKind().GroupVersionKind())
+}
+
+func CRDForGVK(gvk schema.GroupVersionKind) (*apiextensionsv1.CustomResourceDefinition, error) {
```
SimonTheLeg, line 66: nitpick: We should add a description for this func, since it is public

pkg/validation/openapi/openapi_test.go
```diff
@@ -131,47 +137,6 @@ spec:
       storage: true
 `
```
SimonTheLeg, line 139: While viewing this, I realized that this was the only file that checked for the empty-line yaml separator issue (see other comment). So if we are touching this part of the code, it makes sense to change v1crdSingleversion and v1crdMultiversion variables to start with new line and yaml separator. Like this

```go
const v1crdSingleversion = `
---
apiVersion....
`
```

pkg/crd/fs.go
```diff
@@ -0,0 +1,129 @@
...
+
+func loadCRD(filename string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	f, err := embeddedFS.Open(rootDir + "/" + filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	u := &unstructured.Unstructured{}
+	dec := yaml.NewYAMLOrJSONDecoder(f, 1024)
```
SimonTheLeg: I think, this currently still requires the emptyness check and subsequent double decode, like so:

```go
        // Currently there is an issue in the k8s yaml package when working with documents that start with a blank line followed by the yaml document separator
	// (https://github.com/kubernetes/apimachinery/issues/133). So for now we work around the empty buffer by re-reading from it
	// Note: alternative yaml packages (e.g. gopkg.in/yaml) cannot be used, as they do not handle kubernetes-specifc struct tags
	if u.GetAPIVersion() == "" {
		err := dec.Decode(u)
		if err != nil {
			return nil, err
		}
	}
```

This is because all our CRDs start with an empty line followed by the document separator, which is standard for controller-gen.

The detailed issue and description can be found here https://github.com/kubernetes/apimachinery/issues/133.

Additionally, I think it makes sense to write a test for that separately, as we now extracted the logic into its own package.

SimonTheLeg (changes requested):
Overall change makes sense to me. There is one smaller issue issue with the yaml decoding which got swallowed, because one test-string got removed as well.

I have linked the detailed issue in the review comment.

pkg/crd/fs.go
xrstf: Hm but the CRD setup works (the applicationinstallation CRD is setup by reading it from the embeddedFS), so I'm not sure we need more logic here? I just checked, the ApplicationInstallation CRD is complete inside the usercluster, so it doesn't seem to be that the entire CRD or parts of it were not loaded correctly.

SimonTheLeg: Hmm then I guess when you read it from file, the first blank line is being stripped. Because there is definitely still the issue that if the first line is blank followed by the `---` document separator, it is going to fail. You can double-check this, by making the following change: 


```diff
--- a/pkg/validation/openapi/openapi_test.go
+++ b/pkg/validation/openapi/openapi_test.go
@@ -34,6 +34,7 @@ import (

 func TestValidatorFromCRD(t *testing.T) {
        const v1crdMultiversion = `
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
```

But then I guess it's fine for our use case, since now you basically guarantee that it is being read from file or alternatively the user of the lib already needs to make sure they correctly parse into the CustomResourceDefinition Type.

And I guess in the future, when/if they merge my PR in upstream, it will work anyways

pkg/crd/fs.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := dec.Decode(u); err != nil {
+		return nil, err
+	}
+
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), crd); err != nil {
+		return nil, err
+	}
```
SimonTheLeg: since we now drop support for different versions of CustomResourceDefinitions, I think this part can be simplified to:

```go
	crd := &apiextensionsv1.CustomResourceDefinition{}

	dec := yaml.NewYAMLOrJSONDecoder(f, 1024)
	if err := dec.Decode(crd); err != nil {
		return nil, err
	}
```

SimonTheLeg (approved):

SimonTheLeg (approved):

PR #9748, opened 2022-05-06
Automatically manage KKP CRDs on seed clusters

xrstf:
This PR moves the responsibility for installing/updating the CRDs on seed clusters from the KKP installer to the KKP operator. It will now, as part of its regular reconciling loop, ensure the CRDs are up-to-date on seed clusters.

We do this because

* It makes setting up seeds easier.
* It makes updating KKP easier (because now the installer doesn't need to run on seeds if you do not use Minio (the S3-exporter was moved already via #9765)).
* Previously, #9668 planned to perform migrations on Cluster objects in the KKP installer. However Cluster objects must only be modified/migrated *after* the CRDs were updated, which would always happen after updating the master cluster (i.e. the KKP admin needed to run the installer against all seeds). So this means the KKP installer cannot do migrations.
* Compounding the problem above is the fact that seed reconciliation can be paused (temporarily) via annotation on the Seed object, which means the migrations do not just need to wait until an update is done, but also until a Seed is unpaused.

To handle the situation described above, it's much saner to let the KKP operator update the CRDs and then also (in a future PR) perform any necessary migrations. #9706 introduced the SeedStatus already to allow the KKP master to keep track of seeds and their versions. We can later tighten this handling to for example "ignore" old seeds (e.g. the KKP dashboard could reject showing datacenters from paused/old seeds when creating a new cluster).

So this PR simply uses the CRDs that are already embedded in the KKP binaries and updates them on seeds. The PR also moved some of the unit tests around, because having a general Go file (no CE/EE tags) import an EE-only package (metering) caused lots of grief.

There is only one special scenario that needs special attention paid to it: Updating KKP on a shared master/seed system. This is how it would work out:

1. The KKP installer installs the new CRDs onto the master cluster (as it has done since forever) / The admin manually applies the new CRDs with kubectl.
1. Now we have a problem: The KKP operator will always try to reconcile the CRDs and now it will notice that the CRDs "on the seed" (which is the master) do not match anymore. So the operator will begin to override (i.e. downgrade) the new CRDs down to its version.
1. Now the KKP installer will update the KKP operator.
1. A minute or so later when the new operator booted up, it will update the old CRDs on the seed (which is the master).

So in effect we're doing update=>downgrade=>update and I very much suspect that this will explode. There are 3 possible solutions for this:

1. The installer scales down the operator to 0 before updating the CRDs. This would work, but is a PITA for all users who do not use the installer. And I presume that those who don't will also never bother to implement this update dance procedure.
1. The operator could install the **master CRDs** instead of **its own CRDs**. So it would just replicate whatever is on master. However this can be problematic if there is an inconsistent set of CRDs on the master (suppose someone updates them and the process stops for whatever reason after 3 CRDs. Now the operator will replicate the 3 new CRDs onto seeds and thereby spread the problem even further. Even if the operator could determine "OK, the set of CRDs on the master is now consistent.", there could be problems while replicating them to the seeds and in the meantime the master might already get inconsistent again.). So this seems nice, but could lead to absolute chaos and become unmanageable quickly.
1. The operator can determine whether a CRD change would be an upgrade or downgrade. With this knowledge, it can (once the installer has updated new CRDs on the master) determine that those CRDs are newer and not supposed to be overridden.

I chose option (3), which seems to be the most reliable. For this reason this PR now adds the `app.kubernetes.io/version` annotation to all CRDs. Notably, we use a version built by `git describe` instead of just the git HEAD hash. This is because we need to know whether it's a new version, not just that the version changed, so we need some value that continually increases. This is only important for our dev system, all tagged releases have a proper version already.

```release-note
The KKP Operator now updates CRDs on seed clusters. If the KKP Minio chart is not used and the legacy etcd backup configuration is also not used anymore, the KKP Installer does not need to be used for seed clusters anymore.
```


xrstf:
/override pre-kubermatic-e2e-aws-ubuntu-1.23-headless

first timed out, then orphaned because we moved to 1.24, so the job doesn't exist anymore

embik (approved):
/approve

🥳

PR #9749, opened 2022-05-07
make cluster name a valid rfc1035 label

PratikDeoghare:
GCP  refuses to delete instance if its name starts with a number with log message like this [1]

```
> E0501 12:05:31.901609       1 machine_controller.go:381] Failed to reconcile machine "2kcj8bgchx-worker-zl7vb6-68cd96cd4b-ndzlp": An error of type = InvalidConfiguration, with message = Failed to delete instance: googleapi: Error 400: Invalid value for field 'instance': '2kcj8bgchx-worker-zl7vb6-68cd96cd4b-ndzlp'. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?|[1-9][0-9]{0,19}', invalid occurred
```

These are  the requirements for name of an instance.

```
	// Name: The name of the resource, provided by the client when initially
	// creating the resource. The resource name must be 1-63 characters
	// long, and comply with RFC1035. Specifically, the name must be 1-63
	// characters long and match the regular expression
	// `[a-z]([-a-z0-9]*[a-z0-9])?` which means the first character must be
	// a lowercase letter, and all following characters must be a dash,
	// lowercase letter, or digit, except the last character, which cannot
	// be a dash.
	Name string `json:"name,omitempty"`
```

We generate clustername using [rand.String](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String) function which can return  strings that start with a number. 

This PR makes sure that the first char of the string is alphabet. 

Earlier we had `27^10` possible clusters name due to choosing randomly from `bcdfghjklmnpqrstvwxz2456789`.
Now we  have `27^9 x  26` possible cluster names because the first char is now from `a-z`. Still plenty!

[1] https://kubermatic.slack.com/archives/CM5CXH8A3/p1651406901097239

```release-note
NONE
```

### DIFF
--- a/pkg/handler/common/cluster.go
+++ b/pkg/handler/common/cluster.go
@@ -156,6 +156,12 @@ func CreateEndpoint(
 	return ConvertInternalClusterToExternal(newCluster, dc, true, supportManager.GetIncompatibilities()...), nil
 }
 
+func makeClusterName() string {
+	alpha := "abcdefghijklmnopqrstuvwxyz"
+	r := rand.Intn(len(alpha))
+	return fmt.Sprintf("%c%s", alpha[r], rand.String(9))
+}
+
 func GenerateCluster(
 	ctx context.Context,
 	projectID string,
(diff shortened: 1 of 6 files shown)


xrstf (changes requested):
This fix won't work when clusters are created using `kubectl`. We might need to add some name restriction to the cluster admission webhook, though we must be careful to only validate on CREATE, or else existing clusters suddenly become invalid.

xrstf (approved):
/approve

moadqassem:
@lsviben PTAL

PR #9750, opened 2022-05-08
Update vsphere csi/ccm to support k8s 1.23

### LINKED ISSUE #9731: Update vsphere CSI driver
The new major Version (v1.5) of the vsphere csi driver is released and needed for support of k8s 1.23.

https://github.com/kubernetes-sigs/vsphere-csi-driver

Note: v1.5 supports k8s 1.21 to 1.23 for vsphere

Acceptance Criteria:
* vsphere csi driver is on version 1.5

mate4st:

Updates vSphere csi driver to version v2.5.1 that enables support for k8s 1.21 to 1.23 for vsphere.

Snapshot controller is updated to v5.0.1 as a dependency.

Updates vSphere CCM version

Fixes #9731 

```release-note
Support latest vSphere Cloud Controller Manager and  CSI driver for Kubernetes 1.23
```


addons/csi/vsphere/snapshot-controller.yaml
```diff
@@ -58,13 +58,20 @@ rules:
...
+    verbs: ["patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
   - apiGroups: ["snapshot.storage.k8s.io"]
     resources: ["volumesnapshots/status"]
-    verbs: ["update"]
+    verbs: ["update", "patch"]
+  # Enable this RBAC rule only when using distributed snapshotting, i.e. when the enable-distributed-snapshotting flag is set to true
```
xrstf, line 71: (How) is a user supposed to enable this?

xrstf (commented):
QQ

addons/csi/vsphere/snapshot-controller.yaml
mate4st: We do not provide csi configuration for vSphere yet so its not changeable, however I let it in as it is part of the source manifest file.

embik:
Shouldn't this PR drop https://github.com/kubermatic/kubermatic/blob/5609acc5df7a3feb68c507b74bda4b8801537c56/pkg/controller/operator/defaults/defaults.go#L297-L313 too as it enables 1.23 support in vSphere? Maybe even bump the e2e jobs for vSphere to 1.23?

WeirdMachine:
> Shouldn't this PR drop
> 
> https://github.com/kubermatic/kubermatic/blob/5609acc5df7a3feb68c507b74bda4b8801537c56/pkg/controller/operator/defaults/defaults.go#L296-L313
> too as it enables 1.23 support in vSphere? Maybe even bump the e2e jobs for vSphere to 1.23?

Is the csi driver the only reason we have these checks in place?

embik:
Ah, good point. As far as I understand the previous PR https://github.com/kubermatic/kubermatic/pull/8505, it would also include the CCM. Thankfully, 1.23.0 for the vSphere CCM is out too. But it's fine by me to follow that up with another PR if you dont want to scope creep this PR.

WeirdMachine:
Yes, but you are right in the sense that it makes a lot of sense to upgrade all components together. So actually I would like to do it in one PR

WeirdMachine:
/hold

Adding CSI and general vsphere k8s update to the PR

WeirdMachine:
@embik assigned you because this has a little impact on https://github.com/kubermatic/kubermatic/pull/9736

embik (approved):
/approve

Impact on my 1.24 PR is fine, I'll resolve them after this is merged :-)

PR #9752, opened 2022-05-08
Move kubeone cluster status from spec to status

### LINKED ISSUE #9753: Move kubeone cluster status from spec to status

I would like to move modify kubeone external cluster CRD to move status from spec to status
in order to achieve better design

### Solution details

- Requirement 1
- Requirement 2
- Requirement 3

### Use cases
Moving kubeone cluster status from spec to status for consistency with kkp design and to be prepared for future automation tools like kcp

imharshita:

**What does this PR do / Why do we need it**: Modify kubeone externalcluster CRD to move clusterstatus from spec to status

Fixes #9753

**Documentation**:

```release-note
none
```


pkg/apis/kubermatic/v1/external_cluster.go
```diff
@@ -43,7 +43,30 @@ type ExternalCluster struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec ExternalClusterSpec `json:"spec,omitempty"`
+	Spec   ExternalClusterSpec   `json:"spec,omitempty"`
```
xrstf: The spec can (hopefully? :D) never be empty, so it doesn't need `omitempty`.

pkg/apis/kubermatic/v1/external_cluster.go
```diff
@@ -74,47 +97,30 @@ type ExternalClusterCloudSpec struct {
 	KubeOne *ExternalClusterKubeOneCloudSpec `json:"kubeone,omitempty"`
 }
 
-type Status string
+type ConditionStatus string
```
xrstf: Can we call this `ExternalClusterCondition`? `ConditionStatus` is both unspecific (yet the constants only make sense for external clusters) and overly specific (like "dead corpse").

Maybe even "phase" instead of "status", as that's what we're using for clusters and seeds already. Plus it wouldn't collide with the pre-existing concept of Conditions and their statusses. So maybe just `ExternalClusterPhase`?

xrstf (commented):
Just two questions regarding wording

pkg/apis/kubermatic/v1/external_cluster.go
```diff
@@ -43,7 +43,30 @@ type ExternalCluster struct {
...
+}
+
+// ExternalClusterStatus denotes status information about an ExternalCluster.
+type ExternalClusterStatus struct {
+	// Conditions contains conditions an externalcluster is in, its primary use case is status signaling for controller
+	Condition ExternalClusterCondition `json:"conditions,omitempty"`
+}
+
+type ExternalClusterCondition struct {
+	Phase ExternalClusterPhase `json:"status"`
```
xrstf: Please ensure the JSON tag matches the field name.

imharshita: oops, i will correct it

PR #1482, opened 2018-07-05
Add S3 exporter

alvaroaleman:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1407 

**Special notes for your reviewer**:

**Release note**:
```release-note monitoring
Added an S3 exporter for metrics
```

PR #9765, opened 2022-05-09
Duplicate S3 collector and make it run inside the seed-ctrl-mgr

xrstf:
The S3-Exporter is only here to produce Prometheus metrics tailored towards KKP's etcd backups (which is the reason we're not just using a generic S3 exporter). It was added in #1482.

The exporter has a couple of issues:

* It is packaged as a Helm chart that strongly depends on our `minio` Helm chart, because the Minio chart is creating the `kube-system/s3-credentials` secret, which is consumed by the S3-Exporter. So even though the exporter code itself is fairly generic, its chart is only useful if you also a) installed Minio and b) installed Minio using **our** Helm chart.
* It cannot deal with the modern etcd backup/restore controllers, which allow the admin to define multiple backup destinations.
* It must be installed "manually" using the KKP installer, which will also setup Minio on the seed. If an admin doesn't want to use Minio at all (but still use a single backup location), they must install the exporter even more manuallalier.

As in the future we want to reconcile more things automatically on seeds (like #9748), so it would be nice if the tiny S3 exporter is also managed automatically and learns how to deal with the multiple backup destinations. This is what this PR achieves.

This PR *duplicates* the old S3 collector because the new implementation is built around having access to the S3 bucket via the credentials configured in KKP (as apposed to relying on the s3-credentials Secret). The old backup configuration doesn't allow to specify credentials (again, it relies on this Secret created by our Minio chart) and so the new exporter cannot handle the old configuration and the old exporter cannot handle the new configuration.

Until we deprecate the old backup configuration, it's simply easier to just have this new collector run side-by-side and deal with the new configuration (etcd backup/restore) only.

The new exporter also runs as part of the seed-ctrl-mgr. There is little to gain by splitting it apart into a dedicated binary (or at least I cannot see the advantage) and since the seed-ctrl-mgr already manages the backups, it makes sense IMHO to let it also handle the metrics.

Once we remove the old backup configuration stuff, we can `rm -rf charts/s3-exporter cmd/s3-exporter`. At that point we can then also finally get rid of the `s3-credentials` secret (which I have hated for a long time).

```release-note
The seed-controller-manager is now providing Prometheus metrics regarding etcd backups (only for the new etcd backup/restore controllers).
```

### DIFF
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -218,6 +218,8 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 	// and return errors otherwise.
 	// Ideally, the cache wouldn't require the leader lease:
 	// https://github.com/kubernetes-sigs/controller-runtime/issues/677
+	log.Debug("Starting cluster backup collector")
+	collectors.MustRegisterClusterBackupCollector(prometheus.DefaultRegisterer, ctrlCtx.mgr.GetAPIReader(), log, options.caBundle, seedGetter)
 	log.Debug("Starting clusters collector")
 	collectors.MustRegisterClusterCollector(prometheus.DefaultRegisterer, ctrlCtx.mgr.GetAPIReader())
 	log.Debug("Starting addons collector")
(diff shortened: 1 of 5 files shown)


embik (approved):

embik:
Any thoughts on mentioning the s3-exporter by name in the release note? The relationship isn't clear to an outsider IMHO.

xrstf:
/hold

temporarily

PR #9769, opened 2022-05-09
fix S3 exporter

xrstf:
In #8783 I made a typo which broke the S3 exporter. This PR fixes that typo and releases a new version (0.5 is not compatible with KKP 2.20+).

(release note is on the 2.20 branch, no nee to mention this again in 2.21)
```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #9781, opened 2022-05-10
update kubermatic/util image, include yq4

xrstf:
This updates the included binaries in our util Docker image. Since KKP supports k8s 1.21+, I also updated the kubectl release to allow us to be a bit closer to 1.23/1.24 (soon).

yq4 is now included by default, allowing us to slooooowly migrate our stuff over and then at some point make the switch and remove v3 entirely. The `kubermatic/build` image does the same already, so with this PR we should have `yq4` available everywhere.

```release-note
NONE
```

### DIFF
--- a/charts/minio/test/values.example.ce.yaml.out
+++ b/charts/minio/test/values.example.ce.yaml.out
@@ -178,7 +178,7 @@ spec:
             cpu: 100m
             memory: 32Mi
       - name: backup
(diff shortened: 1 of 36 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

xrstf:
/hold

temporarily

PR #9787, opened 2022-05-10
Extend metering report configuration with retention parameter

### LINKED ISSUE #9713: Introduce "Retention" parameter to metering schedule configuration
As a KKP administrator, I would like to define for how long metering reports from a specific schedule configuration should be stored. Parameter doesn't need to be required. If not provided we should not delete any reports from an object store.

pkprzekwas:
Add `retention` parameter to the metering report schedule configuration. If `retention` provided, we create object lifecycle rule on S3 defining the time after which objects with a specific prefix are scheduled for removal.

Fixes #9713 

```release-note
Add optional parameter allowing to set retention for metering configurations.
```


pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -693,6 +693,11 @@ type MeteringReportConfiguration struct {
 
 	// Interval defines the number of days consulted in the metering report.
 	Interval int `json:"interval,omitempty"`
+
+	// +optional
+
+	// Retention defines the number of days after which reports are queued for removal. The default value of "0" keeps reports forever.
+	Retention int `json:"retention,omitempty"`
```
xrstf: Can we add a validation rule to ensure this is never negative? Maybe there's some kubebuilder validation annotation thingy?

pkg/ee/metering/report_config_handler.go
```diff
@@ -112,6 +118,14 @@ func (m updateReportConfigurationReq) Validate() error {
 		}
 	}
 
+	if m.Body.Interval < 0 { // not 1 because 0 is a defaulted value
+		return k8cerrors.NewBadRequest("interval value cannot be smaller than 1.")
```
xrstf: The error message doesn't match the condition.

pkg/ee/metering/report_config_handler.go
```diff
@@ -357,6 +374,10 @@ func updateMeteringReportConfiguration(ctx context.Context, reportCfgReq updateR
 		reportConfiguration.Interval = reportCfgReq.Body.Interval
 	}
 
+	if reportCfgReq.Body.Retention > 0 {
+		reportConfiguration.Retention = reportCfgReq.Body.Retention
+	}
```
xrstf: Does this make it impossible to reset the retention to infinite (i.e. 0)?

docs/zz_generated.seed.yaml
```diff
@@ -343,6 +343,8 @@ spec:
       weekly:
         # Interval defines the number of days consulted in the metering report.
         interval: 7
+        # Retention defines the number of days after which reports are queued for removal. The default value of "0" keeps reports forever.
```
xrstf: Not sure if this makes sense, but can we add a hint that the retention uses the storage backend's lifecycle mechanism and so the backend needs to support that? I'm not sure there are S3 implementations out there that don't, but .. maybe?

pkg/ee/metering/report_config_handler.go
pkprzekwas: You are right. It should be `>=`.

pkg/ee/metering/report_config_handler.go
pkprzekwas: I had a problem with this one. Technically `interval` shouldn't be smaller than `1`, but as this is an optional parameter it defaults to `0` in case it's not provided in the request, hence I left the comment.

I can match the error message with condition but then we would have different validation rules for `POST` and `PUT` of the same model field.

Alternatively, I could put it differently like this:
```suggestion
	if m.Body.Interval != 0 { // checking if interval value was defaulted
		if m.Body.Interval < 1 {
			return k8cerrors.NewBadRequest("interval value cannot be smaller than 1.")
```

docs/zz_generated.seed.yaml
pkprzekwas: Will add some more description to make it more explicit about lifecycle management mechanism.

pkg/ee/metering/report_config_handler.go
xrstf: Alternatively, use a `*int` (or even `*uint32`), then you can easily distinguish between "not set" and 0. :) Wouldn't be the first time int pointers are used in Kubernetes.

PR #9795, opened 2022-05-11
Update e2e tests to kind 0.13 (with k8s 1.24)

xrstf:
Pretty much what it says on the tin.

```release-note
NONE
```


vgramer (commented):
/lgtm

xrstf:
/lgtm cancel

For some bizarre reason, the verify job worked once and since then fails because the new promtool rejects our rules. No idea why it worked on the first iteration, but I clearly need to fix this PR before it can be merged.

PR #9797, opened 2022-05-11
improve cluster deletion

zreigz:
**What does this PR do / Why do we need it**: The cluster deletion throws the error "object has been changed" from time to time. This PR changes the DELETE method and uses the modified cluster instead crating a new cluster object.

Fixes #

```release-note
NONE
```


xrstf (changes requested):
Please describe what problem this is trying to solve and how.

xrstf (approved):
/approve

PR #9799, opened 2022-05-11
update resource handling documentation

SimonTheLeg:

Update documentation on resource handling

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/docs/resource-handling.md
+++ b/docs/resource-handling.md
@@ -132,6 +132,8 @@ func MyWonderfulSecretCreator() reconciling.NamedSecretCreatorGetter {
 }
 ```
 
+The namespace of your object should not be set in the Creator, but rather passed in as an argument to the typed Reconcile func (in our example `ReconcileSecrets`)
+
 ### Template data
 
 Some resources require some dynamic data during reconciling (Such as other Services, Secrets, Configmaps, etc.).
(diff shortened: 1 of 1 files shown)


docs/resource-handling.md
```diff
@@ -121,14 +113,14 @@ This avoids the need to call the creator function twice (1st time to get the obj
...
 ```go
-func MyWonderfulSecretCreator() resources.NamedSecretCreatorGetter {
-	return func() (string, resources.SecretCreator) {
-		return "my-name", func(existing *corev1.Secret) (*corev1.Secret, error) {
-			existing.Name = "wonderful-secret"
+func MyWonderfulSecretCreator() reconciling.NamedSecretCreatorGetter {
+		return "my-wonderful-secret", func(existing *corev1.Secret) (*corev1.Secret, error) {
+
+			// We don't need to set the object's name, as that's being done by the typed Reconcile function.
```
xrstf, line 123: name _or namespace_

xrstf (changes requested):
One minor improvement left

docs/resource-handling.md
SimonTheLeg: As far as I understand it, "name". Afaik the namespace should not be set in the CreatorGetter, but rather passed directly to the corresponding typed Reconcile function.

That being said, I have now added a sentence to this paragraph to make this more clear. ptal

xrstf (approved):
/approve

PR #9807, opened 2022-05-12
Add audit logging to presets for Cluster API & Gatekeeper resources

### LINKED ISSUE #8784: Improve audit policy presets with KKP custom resources

The [audit logging presets](https://github.com/kubermatic/kubermatic/blob/d43dc9cb7c490f84c22472293e6e8343095ced48/pkg/resources/apiserver/audit.go#L27) provide a nice set of Kubernetes built-in resources that are being audited. We can improve that even further by adding custom resources we are always deploying to KKP user clusters, like `MachineDeployments` and `Machines`. That will improve audit coverage even more.

Also look at other resources we deploy, maybe even only with addons, and see if they should be in the audit policies by default. If a resource is on the audit policy but not deployed to the cluster (e.g. because Addons are optional), it's not an issue for the policy, the resource is just skipped.

### Acceptance criteria

Audit policy presets include `MachineDeployments`, `MachineSets` and `Machines` and potentially other audit-relevant resources.

embik:

The audit policy presets we offer cover Kubernetes core/default resources, but do not cover custom resources that might be of interest. Obviously we cannot cover what users install onto clusters, but this PR adds logging to the `minimal` and the `recommended` audit logging presets for Cluster API resources and the Gatekeeper template API group. 

It also adds logging for the Kubernetes secret that is used to control SSH access to nodes via `user-ssh-keys-agent`. Quite an important `Secret` that you want to tightly audit, as writing to that secret allows you to access all nodes.

Fixes #8784

**Documentation**:

```release-note
Audit logging presets `recommended` and `minimal` now include ResponseRequest level logging for `Machine`, `MachineSets` and `MachineDeployments`, any Gatekeeper template resources, and the user-ssh-keys-agent secret for SSH keys
```

### DIFF
--- a/pkg/resources/apiserver/audit.go
+++ b/pkg/resources/apiserver/audit.go
@@ -37,14 +37,24 @@ kind:  Policy
 omitStages:
   - "RequestReceived"
 rules:
-  # log all changes to workloads (Pods, Deployments, etc) in full, including request and response
   - level: RequestResponse
     verbs: ["create", "delete", "update", "patch"]
     resources:
+      # log all changes to workloads (Pods, Deployments, etc)
       - group: ""
         resources: ["pods"]
       - group: "apps"
         resources: ["deployments", "statefulsets", "daemonsets", "replicasets"]
+      # log all changes to machines and higher level machine objects
+      - group: "cluster.k8s.io"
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

PR #9808, opened 2022-05-12
add quota filter for Kubevirt VMIPresets

### LINKED ISSUE #9778: Add resource quota filtering for KubeVirt VMs
KKP admins can set a resource quota for all cloud providers. KubeVirt implementation must support that too.
We have to add API filtering to see if quota is matched. Similar like AWS provider has that implemented.
https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/common/provider/aws.go#L285

sankalp-r:
This PR introduces quota-filtering over API which lists VMIPresets for KubeVirt cloud provider.
VMIPresets that are within the quota limit will be returned as a response. 

Fixes #9778 

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/handler/common/provider/kubevirt.go
+++ b/pkg/handler/common/provider/kubevirt.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 
 	kubevirtv1 "kubevirt.io/api/core/v1"
@@ -31,6 +33,7 @@ import (
 	handlercommon "k8c.io/kubermatic/v2/pkg/handler/common"
 	"k8c.io/kubermatic/v2/pkg/handler/middleware"
 	"k8c.io/kubermatic/v2/pkg/handler/v1/common"
+	"k8c.io/kubermatic/v2/pkg/log"
 	"k8c.io/kubermatic/v2/pkg/provider"
 	"k8c.io/kubermatic/v2/pkg/provider/cloud/kubevirt"
 	kubernetesprovider "k8c.io/kubermatic/v2/pkg/provider/kubernetes"
@@ -38,6 +41,7 @@ import (
 	utilerrors "k8c.io/kubermatic/v2/pkg/util/errors"
 
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +228,28 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+	// If the record passes all the filters, add it to the final slice.
+	for _, vmiPreset := range vmiPresets {
+		presetSpec := kubevirtv1.VirtualMachineInstancePresetSpec{}
+		if err := json.Unmarshal([]byte(vmiPreset.Spec), &presetSpec); err != nil {
+			return filteredVMIPresets
+		}
+
+		var cpu, memory int
+		if cpu = int(presetSpec.Domain.Resources.Requests.Cpu().Value()); cpu == 0 {
+			// defaults to 1 (reference: https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/virt-config.go#L44)
```
hdurand0710: 1 or 100m ?
Can we directly use the `DefaultCPURequest` rather than hard code it here ?

sankalp-r: `FilterCPU(record, min, max int)` works with integer, so that's why used `1`.

hdurand0710: But if the quota is 1 to 32 for example, 100m should be rejected. Here it would be accepted.

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +228,22 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+	filteredVMIPresets := apiv2.VirtualMachineInstancePresetList{}
+
+	// Range over the records and apply all the filters to each record.
+	// If the record passes all the filters, add it to the final slice.
+	for _, vmiPreset := range vmiPresets {
+		presetSpec := kubevirtv1.VirtualMachineInstancePresetSpec{}
+		if err := json.Unmarshal([]byte(vmiPreset.Spec), &presetSpec); err != nil {
+			return filteredVMIPresets
+		}
+		cpu := int(presetSpec.Domain.Resources.Requests.Cpu().AsApproximateFloat64())
```
lsviben: From what I understand from talking to @mfranczy , there are multiple ways to set the resource request for a VMIPreset, described here:
https://github.com/kubermatic/kubermatic/blob/master/pkg/ee/validation/machine/providers.go#L264-L273

So I think first you need to find where exactly its set, `presetSpec.Domain.Resources.Requests.` is not the only place

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +228,40 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+		}
+		if !presetSpec.Domain.Resources.Requests.Cpu().IsZero() {
+			cpu = int(presetSpec.Domain.Resources.Requests.Cpu().AsApproximateFloat64())
+		}
+
+		memory := int(presetSpec.Domain.Resources.Requests.Memory().Value() / (1 << 30))
+		if handlercommon.FilterCPU(cpu, quota.MinCPU, quota.MaxCPU) && handlercommon.FilterMemory(memory, quota.MinRAM, quota.MaxRAM) {
+			filteredVMIPresets = append(filteredVMIPresets, vmiPreset)
+		}
+	}
```
lsviben: I think this is still not fully correct. 
The CPU can also be set just in `presetSpec.Domain.Resources.Limits`, and also you should have a default value set (from what I understand its 100mi). 
And for the memory, it can also be just set in `presetSpec.Domain.Resources.Limits`, and the default is 1Gi

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +231,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+	// Range over the records and apply all the filters to each record.
+	// If the record passes all the filters, add it to the final slice.
+	for _, vmiPreset := range vmiPresets {
+		presetSpec := kubevirtv1.VirtualMachineInstancePresetSpec{}
+		if err := json.Unmarshal([]byte(vmiPreset.Spec), &presetSpec); err != nil {
+			continue
+		}
+
+		cpu, memory, err := GetKubeVirtPresetResourceDetails(presetSpec)
+		if err != nil {
```
mfranczy, line 249: It would be nice to add some log here that the preset was skipped in case of error

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +231,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+}
+
+// GetKubeVirtPresetResourceDetails extracts cpu and mem resource requests from the kubevirt preset
+// for CPU, take the value by priority:
+// - if resource limit is set, use that
+// - check if resource request is set, use that
+// - check if spec.cpu is set, if socket and threads are set then do the calculation, use that
+// for memory, take the value by priority:
+// - if resource limit is set, use that
+// - if resource request is set, use that.
```
mfranczy: please update the comment

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +231,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+		return resource.Quantity{}, resource.Quantity{}, errors.New("should not specify both spec.domain.cpu and spec.domain.resources.[requests/limits].cpu in VMIPreset")
+	}
+	if !presetSpec.Domain.Resources.Requests.Cpu().IsZero() && !presetSpec.Domain.Resources.Limits.Cpu().IsZero() && presetSpec.Domain.Resources.Requests.Cpu().Value() != presetSpec.Domain.Resources.Limits.Cpu().Value() {
+		return resource.Quantity{}, resource.Quantity{}, errors.New("spec.domain.resources.requests.cpu must be equal to spec.domain.resources.limits.cpu in VMIPreset")
+	}
+	if presetSpec.Domain.CPU == nil && presetSpec.Domain.Resources.Requests.Cpu().IsZero() && !presetSpec.Domain.Resources.Limits.Cpu().IsZero() {
+		return resource.Quantity{}, resource.Quantity{}, errors.New("either one of spec.domain.cpu or spec.domain.resources.requests.cpu must be set in VMIPreset")
+	}
+
+	if presetSpec.Domain.CPU != nil && presetSpec.Domain.CPU.Cores != 0 {
```
mfranczy: I know the function has been copied,  but you have to check only if `.CPU` is not `nil`. It can be that somebody just set sockets instead of cores.
I would validate this like:
```
if domain.CPU != nil {
     if CPU.cores == 0 {
          cores = 1
      }
      ... // rest should be as it is already
}
```

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +231,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+// - check if resource request is set, use that
+// - check if spec.cpu is set, if socket and threads are set then do the calculation, use that
+// for memory, take the value by priority:
+// - if resource limit is set, use that
+// - if resource request is set, use that.
+func GetKubeVirtPresetResourceDetails(presetSpec kubevirtv1.VirtualMachineInstancePresetSpec) (resource.Quantity, resource.Quantity, error) {
+	var err error
+	// Get CPU
+	cpuReq := resource.Quantity{}
+	if presetSpec.Domain.CPU != nil && (!presetSpec.Domain.Resources.Requests.Cpu().IsZero() || !presetSpec.Domain.Resources.Limits.Cpu().IsZero()) {
```
mfranczy: the conditions we spoke about could have a more simple form since the function checks already some of them.
It could look like this

```
if presetSpec.Domain.CPU != nil {

		if !presetSpec.Domain.Resources.Requests.Cpu().IsZero() || !presetSpec.Domain.Resources.Limits.Cpu().IsZero() {
			// raise an error
		}

		cores := presetSpec.Domain.CPU.Cores
		if cores == 0 {
			cores = 1
		}
		threads := presetSpec.Domain.CPU.Threads
		if threads == 0 {
			threads = 1
		}
		sockets := presetSpec.Domain.CPU.Sockets
		if sockets == 0 {
			sockets = 1
		}

		cpuReq, err = resource.ParseQuantity(strconv.Itoa(int(cores * threads * sockets)))
		if err != nil {
			return resource.Quantity{}, resource.Quantity{}, fmt.Errorf("error parsing calculated KubeVirt VCPU: %w", err)
		}
	} else {
		if !presetSpec.Domain.Resources.Requests.Cpu().IsZero() {
			cpuReq = *presetSpec.Domain.Resources.Requests.Cpu()
		}

		if !presetSpec.Domain.Resources.Limits.Cpu().IsZero() {
			cpuReq = *presetSpec.Domain.Resources.Limits.Cpu()
		}
	}
```

mfranczy (changes requested):
Looks good, just a few things to change

mfranczy (approved):
/approve
/lgtm

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +232,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+
+	// get MEM
+	memReq := resource.Quantity{}
+	if presetSpec.Domain.Resources.Requests.Memory().IsZero() && presetSpec.Domain.Resources.Limits.Memory().IsZero() {
+		return resource.Quantity{}, resource.Quantity{}, errors.New("spec.domain.resources.[requests/limits].memory must be set in VMIPreset")
+	}
+	if !presetSpec.Domain.Resources.Requests.Memory().IsZero() {
+		memReq = *presetSpec.Domain.Resources.Requests.Memory()
+	}
+	if !presetSpec.Domain.Resources.Limits.Memory().IsZero() {
```
lsviben, line 313: so the memory can be set through limits in the end? There wont be an error?

pkg/handler/common/provider/kubevirt.go
```diff
@@ -224,3 +232,87 @@ func KubeVirtStorageClassesWithClusterCredentialsEndpoint(ctx context.Context, u
...
+
+// GetKubeVirtPresetResourceDetails extracts cpu and mem resource requests from the kubevirt preset
+// for CPU, take the value by priority:
+// - check if spec.cpu is set, if socket and threads are set then do the calculation, use that
+// - if resource request is set, use that
+// - if resource limit is set, use that
+// for memory, take the value by priority:
+// - if resource request is set, use that
+// - if resource limit is set, use that.
+func GetKubeVirtPresetResourceDetails(presetSpec kubevirtv1.VirtualMachineInstancePresetSpec) (resource.Quantity, resource.Quantity, error) {
```
lsviben, line 269: what about the default `100mi` CPU and `1Gi` memory? Why dont we have this?

mfranczy: After all we decided to skip presets that they don't have it. We require from presets that those are explicitly set.

pkg/handler/common/provider/kubevirt.go
mfranczy: It depends on the right QoS class. We plan to iterate over this implementation. The same for dedicated cpu placement. For now it's ok. Also it is aligned with this what we support from the KKP side regarding KubeVirt so far, so all good.

pkg/handler/common/provider/kubevirt.go
lsviben: Somehow my eyes just went over the check which throws the error when nothing is set. Looks good

PR #9817, opened 2022-05-13
fix health konnectivity/apiserver status infinite toggle

PratikDeoghare:
Fixes crashloop due to shared pointer and looping over map in the health status. 
Minimum repro https://go.dev/play/p/iLuIekyfOKt
 
```release-note
NONE
```


xrstf (approved):
/approve

PR #9823, opened 2022-05-16
Use k8c.io/kubermatic/v2/pkg/semver for some version comparisons

embik:
This is a personal pet peeve of me, we have been using `k8c.io/kubermatic/v2/pkg/semver.Semver` in a couple of places, but `semverlib` is also used in some locations. In particular, some of our Kubernetes version checks only ran against the minor version and omitted the major version. While that should never create a problem since we're only supporting a very limited set of versions (and we'll never support Kubernetes `1.20` and a theoretical `2.20` at the same time), I don't feel comfortable discarding that part of the version information.

This moves a couple of places to `semver.Semver` and checks against `MajorMinor()`. Maybe we don't want to merge this, I don't know. WDYT?

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/addon/addon_controller_test.go
+++ b/pkg/controller/seed-controller-manager/addon/addon_controller_test.go
@@ -418,14 +418,18 @@ func TestController_ensureAddonLabelOnManifests(t *testing.T) {
 
 func TestController_getApplyCommand(t *testing.T) {
 	controller := &Reconciler{}
-	clusterVersion := defaults.DefaultKubernetesVersioning.Default.Semver()
+
+	clusterVersion := defaults.DefaultKubernetesVersioning.Default
+	if clusterVersion == nil {
+		t.Fatalf("Should be able to determine default Kubernetes version, but got nil")
+	}
 
 	binary, err := kubectl.BinaryForClusterVersion(clusterVersion)
 	if err != nil {
 		t.Fatalf("Should be able to determine a kubectl binary for %q, but got %v", clusterVersion, err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/seed-controller-manager/addon/addon_controller_test.go
```diff
@@ -418,14 +418,18 @@ func TestController_ensureAddonLabelOnManifests(t *testing.T) {
 
 func TestController_getApplyCommand(t *testing.T) {
 	controller := &Reconciler{}
-	clusterVersion := defaults.DefaultKubernetesVersioning.Default.Semver()
+
+	clusterVersion := defaults.DefaultKubernetesVersioning.Default
+	if clusterVersion == nil {
+		t.Fatalf("Should be able to determine default Kubernetes version, but got nil")
```
xrstf: `Fatal()` is enough

xrstf:
I would love if we somehow (never checked or even thought hard about it) contribute our patches back upstream and completely get rid of our own version package. Just as my three fiddy.

embik:
> I would love if we somehow (never checked or even thought hard about it) contribute our patches back upstream and completely get rid of our own version package. Just as my three fiddy.

I don't disagree. If Go supported easy tuples, you could also do this, but alas this isn't valid Go code:

```go
switch (version.Major(), version.Minor()) {
   case (1, 22):
       doSomething()
}
```

In any case, I wanted to make things more consistent for now.

xrstf (approved):
/approve
/retest

PR #9826, opened 2022-05-16
Update to controller-runtime 0.12.1 / Kubernetes 1.24

xrstf:
This bumps our Kubernetes dependencies.

I chose to "fork" the aws-iam-authenticator for 2 reasons:

* It blocks the 1.24 update because of https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/458
* It prevents us from updating to AWS SDK v2.

Since it's only 2 files and not a lot of code, it seems easier to just copy-paste them and get rid of the dependency alltogether.

This PR also includes a bugfix for the addon logging, which used the wrong variable, plus it makes use of the new `BaseContext` option, which will hopefully make shutdowns a bit faster.

**Cilium**
Cilium has some code in it that prevents us from using 1.11.x. Since there is no 1.12 release yet, I had to bump to the master branch unfortunately.

Cilium then required another toml dependency, which is why I had to update and pin that one as well. Otherwise Go complained that Go 1.16 chooses different versions that 1.17 or something. The same goes for the envoy control plane, which would have been bumped too far and broke things.

TL;DR: It's a bit of a mess, also because some machine-controller deps are suuuper outdated (toml 0.3 from 2018).

```release-note
Update to controller-runtime 0.12.1 / Kubernetes 1.24
```


ahmedwaleedmalik (approved):
/approve 🚀

PR #9828, opened 2022-05-16
removed buggy log sync

lsviben:
Removes the deferred `log.Sync()` which causes:
```
"sync /dev/stderr: invalid argument"
```
on some OS. Its caused by: https://github.com/uber-go/zap/issues/370.%C2%A0, and there are no plans to fix it.
As its not essential, and just causes confusion because it fails on exit, so people think the whole binary failed, lets remove it.

```release-note
Fixed `image-loader` and other binaries failing with "sync /dev/stderr: invalid argument", by removing the log.Sync() on command exit.
```

### DIFF
--- a/cmd/conformance-tester/main.go
+++ b/cmd/conformance-tester/main.go
@@ -68,11 +68,6 @@ func main() {
 	}
 
 	log := rawLog.Sugar()
-	defer func() {
-		if err := log.Sync(); err != nil {
-			fmt.Println(err)
-		}
-	}()
 
 	// parse our CLI flags
 	if err := opts.ParseFlags(); err != nil {
--- a/cmd/image-loader/main.go
+++ b/cmd/image-loader/main.go
@@ -103,11 +103,6 @@ func main() {
 
 	rawLog := kubermaticlog.New(logOpts.Debug, logOpts.Format)
 	log := rawLog.Sugar()
-	defer func() {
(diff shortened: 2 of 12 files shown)


xrstf (changes requested):
We do the same in nearly all binaries, so maybe we should ajdust all of them?

xrstf:
IIRC @embik had some strong-ish opinions on leaving out the sync last time I touched it, so let's allow them a chance to interject.

I would vote to remove the Sync() everywhere, because if we don't, someone(tm) (like me) will forget in half a year why the image-loader is "special" and bring back the seemingly missing Sync() call.

lsviben:
> IIRC @embik had some strong-ish opinions on leaving out the sync last time I touched it, so let's allow them a chance to interject.
> 
> I would vote to remove the Sync() everywhere, because if we don't, someone(tm) (like me) will forget in half a year why the image-loader is "special" and bring back the seemingly missing Sync() call.

Ok, will wait for @embik to give his 2 cents, if nothing against ill just remove it everywhere

embik:
> IIRC @embik had some strong-ish opinions on leaving out the sync last time I touched it, so let's allow them a chance to interject.

I don't remember having an opinion on this to be honest, so it cannot be that strong 🤷🏻‍♂️

xrstf (approved):
/approve

PR #9830, opened 2022-05-17
applicationInstallation crd: change .spec.values type from byte to object

vgramer:
Currently the `.spec.Values` has format `string` and type `byte` which is painful to use.

Thi pr switch it to type `object` which allows the user to set values like this:
```yaml
spec:
  values:
    a: b
    c:
      d: e
```

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apis/apps.kubermatic/v1/application_installation.go
```diff
@@ -55,6 +55,9 @@ type ApplicationInstallationSpec struct {
 	ApplicationRef ApplicationRef `json:"applicationRef"`
 
 	// Values describe overrides for manifest-rendering
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
 	Values json.RawMessage `json:"values,omitempty"`
```
ahmedwaleedmalik: A better way to solve this would be to use either of these instead of `json.RawMessage`:
- "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1".JSON
- "k8s.io/apimachinery/pkg/runtime".RawExtension

They are perfect for our use case and built exactly for this scenario. With RawExtension you'll have to use `// +kubebuilder:pruning:PreserveUnknownFields` marker.

pkg/controller/seed-controller-manager/initial-application-installation-controller/controller_test.go
```diff
@@ -275,11 +275,15 @@ func TestReconcile(t *testing.T) {
 
 func generateApplication(name string) apiv1.Application {
 	var values json.RawMessage
-	_ = json.Unmarshal([]byte(`{
+	err := json.Unmarshal([]byte(`{
 		"key": "value",
-		"key2": "value2",
```
vgramer, line 280: note: json does not accept trailing comma at the end of the object

ahmedwaleedmalik (approved):
/approve ⚡

Issue #7069, opened 2021-05-19
Remove the cluster-autoscaler deployment from resources/seed-controller-manager


xmudrii:
There is a cluster-autoscaler deployment defined in the resources and deployed by seed-controller-manager, controlled by this annotation: https://github.com/kubermatic/kubermatic/blob/v2.17.0/pkg/crd/kubermatic/v1/cluster.go#L43

This deployment uses a Kubermatic fork of cluster-autoscaler that's not maintained at all. Instead, we have created an addon that uses the upstream cluster-autoscaler, which supports Cluster-API since v1.18.2.

**Acceptance criteria**

* The old deployment is removed in favor of the cluster-autoscaler addon

Issue #9818, opened 2022-05-16
Component images for Kubernetes 1.24


embik:
#9736 is adding support for deploying Kubernetes 1.24 clusters, but upstream components like CCMs, CSI drivers, etc, do not (officially) support it. This issue tracks components that might need updating or a documentation note saying it is compatible with 1.24.

- [x] [cluster-autoscaler](https://github.com/kubernetes/autoscaler) (#9831)
- [x] kubernetes-dashboard (#9948)
- [x] vertical-pod-autoscaler (VPA) - (#9837 was probably enough)
- [x] Anexia CCM (no never version available)
- [x] [Openstack Cinder CSI](https://github.com/kubernetes/cloud-provider-openstack) (#9935)
- [x] Openstack CCM (#9935)
- [x] [Hetzner CSI](https://github.com/hetznercloud/csi-driver) (no new version available)
- [x] [Hetzner CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager) (no new version available)
- [x] Kubevirt CSI (no new version necessary as per sig-virtualization)
- [x] Kubevirt CCM (no new version necessary as per sig-virtualization)
- [x] Nutanix CSI (does not follow Kubernetes releases; version skew policy should support 1.24)
- [x] vSphere CSI (no newer version available)
- [x] vSphere CCM (#10203)

Please add to the list if something is missing.

Also check for `TODO: 1.24` notes in KKP code base when working on this.

dharapvj:
FYI... https://github.com/kubermatic/kubermatic/issues/10461
Vsphere CSI new version is now available which supports k8s 1.24

PR #9831, opened 2022-05-17
Add cluster-autoscaler version for Kubernetes 1.24

embik:
A couple of components do not have Kubernetes 1.24 versions. cluster-autoscaler 1.24.0 just recently released. This PR adds it.

In addition, I found a seemingly ancient `clusterautoscaler` package that apparently last worked with Kubernetes **1.14**. I took the liberty to drop it from our code base as part of this PR. 

towards #9818

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -325,9 +324,6 @@ func GetDeploymentCreators(data *resources.TemplateData, enableAPIserverOIDCAuth
 		)
 	}
 
-	if data.Cluster().Annotations[kubermaticv1.AnnotationNameClusterAutoscalerEnabled] != "" {
```
xrstf, line 328: Does the constant then still need to exist?

embik: I don't think so, good point. Let me remove it.

xrstf (approved):
/approve

xrstf:
This accidentally also fixed #7069.

PR #8775, opened 2022-01-24
Bump kube-state-metrics and VPA to new image location

embik:
**What does this PR do / Why do we need it**:

**Does this PR close any issues?**:
Fixes #5773

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
Update Vertical Pod Autoscaler (VPA) to 0.9.2
```

Signed-off-by: Marvin Beckers <marvin@kubermatic.com>

PR #9837, opened 2022-05-18
Update VPA to 0.10.0

xrstf:
This just updates the VPA version we deploy. At least in our e2e tests the VPA 0.10 is a happy camper and seems to have no trouble working just fine. Not sure why we previously deployed v0.9.2, but kept the Go dependency at 0.4.0.

```release-note
Update Vertical Pod Autoscaler to 0.10.0
```


embik:
> Not sure why we previously deployed v0.9.2, but kept the Go dependency at 0.4.0.

Probably because I forgot about it in #8775.

embik (approved):
/lgtm

PR #9839, opened 2022-05-18
rename usercluster-webhook to be in line with the usercluster controller

xrstf:
screenshot-2022-05-18-115342

This inconsistency triggered me. I would love if binary name == deployment name, but that ship has sailed long ago with the `user-cluster-controller-manager`, which lives in `usercluster-controller-manager`, and renaming that Deployment is more work than is worth it. But the new webhook deployment, that one we can still easily harmonize with the other.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡ 

Agreed with your point here.

ahmedwaleedmalik (approved):
/approve

PR #9843, opened 2022-05-18
run tagliatelle against all APIs

xrstf:
This PR ensures that we check the tags for apps.kubermatic.k8c.io as well. Also this makes use of a dedicated config file to actually disable all linters except for `tagliatelle`.

Previously:

```
level=info msg="Memory: 778 samples, avg is 373.2MB, max is 792.5MB"
level=info msg="Execution took 1m17.652418212s"
```

Now:

```
level=info msg="Memory: 32 samples, avg is 45.2MB, max is 49.0MB"
level=info msg="Execution took 3.006100778s"
```

```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #9854, opened 2022-05-19
set the security context for the WEB terminal Pod

### LINKED ISSUE #9819: Configure a Security Context for a terminal Pod

zreigz:
**What does this PR do / Why do we need it**: set the security context for the WEB terminal Pod. It's required to exec pod as a regular user. Add control on whether a process can gain more privileges than its parent process. 

Fixes #9819

```release-note
NONE
```


xrstf (approved):
/approve

PR #8864, opened 2022-02-04
import and export cluster templates

zreigz:
**What does this PR do / Why do we need it**: Two methods to export and import cluster templates to/from file

**Does this PR close any issues?**:
Fixes #7435

```release-note
Add possibilities to import and export cluster templates from/to file
```

PR #9865, opened 2022-05-19
Fix credential reference for cluster template

ahmedwaleedmalik:

`credentials` field is used during clean up for cluster templates. https://github.com/kubermatic/kubermatic/pull/8864/files#diff-d4de1dd6a513e65aa7b67b778a7fcb4d607b13d85a9e2fbdf81a8b7d3bd83092R354 mistakenly was setting it to an empty string. 

Due to this missing reference, cluster template deletion was broken. It couldn't delete the secret and then remove the corresponding finalizer `kubermatic.k8c.io/cleanup-credentials-secrets`. Ref https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/master-controller-manager/cluster-template-synchronizer/controller.go#L149

Fixes #

**Documentation**:

```release-note
Add missing credentials reference for cluster templates.
```

### DIFF
--- a/pkg/handler/v2/cluster_template/cluster_template.go
+++ b/pkg/handler/v2/cluster_template/cluster_template.go
@@ -355,7 +355,7 @@ func createClusterTemplate(ctx context.Context, userInfoGetter provider.UserInfo
 			Labels:      map[string]string{},
 			Annotations: map[string]string{},
 		},
-		Credential:             "",
+		Credential:             partialCluster.GetSecretName(),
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik:
> The KKP installer could run some migration logic to fix stuck clusters.

My initial idea was to implement that. But this change was never released so it doesn't affect any existing clusters.
https://github.com/kubermatic/kubermatic/blob/release/v2.20/pkg/handler/v2/cluster_template/cluster_template.go#L103

PR #9868, opened 2022-05-20
Fail update validation for unsupported cloud spec changes

### LINKED ISSUE #9855: Prevent patching of cloud provider spec where necessary
When parts of the cloud spec are left empty, provider implementations usually generate resources dynamically and add a finalizer to track that they own the resource and are responsible for cleaning it up.

Unfortunately, it is possible to patch the cloud provider spec afterwards (although we don't really support this, as it will split cluster resources/nodes across several configurations). This can happen accidentally. If the finalizer is not removed at the same time, KKP believes that the given resource is managed by KKP and needs to be removed (some provider implementations might be using a provider-side tag system to track ownership, but not all of them have that available). This can lead to unintended deletion of pre-created or shared resources.

We should ensure that those fields cannot be changed if we don't support changing the cloud resources in-use (I think that's the case for most resources, actually; imagine you randomly change the subscription your cluster is deployed into).

Consider that the cloud provider implementations usually _do_ patch the provider spec with resources they generate. As such, patching empty fields must remain supported for the time being.

## Acceptance Criteria

- Changes to cloud provider spec fields that are not meant to be changed are rejected

embik:

KKP does not really assume that the cloud provider spec changes. That is especially true for resources generated by KKP if no name is given. This PR tightens the update validation for cloud providers so users can no longer patch the cloud spec in a way that breaks KKP (e.g. if you patch the name of a generated resource with a pre-created resource later on, KKP will try to delete that resource because the finalizer is still there).

Fixes #9855

**Documentation**:

```release-note
Cloud provider spec changes fail validation for fields that are not supported by in-place updates (mostly cloud resources that can be auto-generated by KKP)
```

### DIFF
--- a/pkg/provider/cloud/aws/provider.go
+++ b/pkg/provider/cloud/aws/provider.go
@@ -121,7 +121,31 @@ func (a *AmazonEC2) ValidateCloudSpec(ctx context.Context, spec kubermaticv1.Clo
 }
 
 // ValidateCloudSpecUpdate verifies whether an update of cloud spec is valid and permitted.
-func (a *AmazonEC2) ValidateCloudSpecUpdate(ctx context.Context, oldSpec kubermaticv1.CloudSpec, newSpec kubermaticv1.CloudSpec) error {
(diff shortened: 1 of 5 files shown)


embik:
/hold

Needs #9917.

PR #9869, opened 2022-05-20
Use newer 'small' instance in Equinix Metal for e2e

embik:

As per [Equinix Metal documentation](https://metal.equinix.com/developers/docs/hardware/legacy-servers/), `c3.small.x86` is still considered a legacy instance type, but `t1.small.x86` seems to be effectively gone now, so this is the next best thing (if it stays available).

Fixes #

**Documentation**:

```release-note
NONE
```


embik:
This unfortunately does not work either.

PR #9877, opened 2022-05-23
Use protokol for CI logs

xrstf:
Up until now, we printed the logs of the cluster control plane pods and all pods in the usercluster directly to stdout. We did this to let Prow highlight errors/failures, with the hope that "this would allow us to spot problems that do not lead to a failed test".

However, there are a number of reasons why I think this (ironically) failed:

* Calico produces roughly 5MB of logs *per node*, so a good 15MB worth of logs.
* Many pods contain transient errors (like the metris-server, which will always crash twice before getting healthy).
* Either an e2e test fails or it doesn't. If it doesn't fail, I would wager that _nobody_ is looking at the build log so spot potential issues.
* Prow has a max_highlight limit and so if the errors occur after the first 15MB of Calico logs, we won't see them.
* Printing a lot of logs makes it hard to follow the actual test procedure.

For these reasons I decided to use https://github.com/xrstf/protokol to gather the logs. Protokol is a simply standalone tool that downloads/streams the logs for all pods that match a certain set of criteria. We then dump those logs into plain ol' text files and let Prow upload them to GCS for later viewing. Protokol will simply sit in the background and steam the logs, finding new pods automatically.

This means that you need more clicks to see all the logs, but

* You can now actually see _all_ logs, including KKP (which is arguably much more important than Calico logs).
* It won't crash your browser trying to view the logs.

If we decide that this PR is good, I would bake protokol right into our build image.

```release-note
NONE
```


hack/ci/run-e2e-tests.sh
```diff
@@ -37,6 +43,9 @@ echodate "Creating kind cluster"
 export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
 source hack/ci/setup-kind-cluster.sh
 
+protokol --kubeconfig "$KUBECONFIG" --flat --output "$ARTIFACTS/logs/cluster-control-plane" --namespace 'cluster-*' > /dev/null 2>&1 &
```
xmudrii, line 41: Why don't we run this command in other scripts (e.g. dualstack and Cilium)?

xrstf: You probably missed https://github.com/kubermatic/kubermatic/pull/9877/commits/a534002e48e2f46ad7750e218cb857473135b83d ?

xrstf:
/hold

new build image available with protokol baked in, need to update this PR

hack/ci/run-e2e-tests.sh
xmudrii: Looks like so. :D

xmudrii (approved):
/lgtm
/approve
:zap:

Issue #8379, opened 2021-12-06
Implement git source


SimonTheLeg:
**Acceptance Criteria**
* can access both public and private repositories using secret injection mechanism
* uses the Application Secret(s) for private access
* mechanism must allow for multiple sources (e.g. git, but more can follow in the future)
* mechanism must make it easy for other packages to consume the fetched files

PR #9891, opened 2022-05-25
ApplicationDefinition: add missing validation check and field for git

vgramer:

This PR changes the type of some fields to be compliante with implementation and adds missing validation.

filed changes:
-  credentials type  from `GlobalSecretKeySelector` to `corev1.SecretKeySelector`. With PR #9714, the secrets are synced from master to seed only if they have annotation and live in the namespace where Kubermatic is installed. Consequently, the namespace can be inferred, so we no longer need `GlobalSecretKeySelector`.
- GitSource.Ref from `string` to `object`. To clone the project with the  go-git library, we must know the kind of "reference" we checkout (branch, tag, or commit). Moreover with an object, user can set both commit and branch to do a swallow clone on the branch.

validation updates:

- One and only one Source must be defined
-  Git reference: at least a branch, tag or commit must be defined. tag can not be used in conjunction with branch or commit
-  Validate that grit credentials correspond to the git method (eg if gitCredentials.Method is password then gitCredentials.username and gitCredentials.password must be defined.)

ref #8379 

**Documentation**:

```release-note
NONE
```


pkg/apis/apps.kubermatic/v1/application_definition.go
```diff
@@ -19,25 +19,32 @@ package v1
...
+	// Password holds the ref and key in the secret for the Password credential. Secret must exist in the namespace where
+	// KKP is installed.
+	Password corev1.SecretKeySelector `json:"password"`
 }
 
 type HelmSource struct {
-	// URl of the helm repository
+	// URl of the helm repository.
+	// It can be an HTTP(s) repository (e.g. https://localhost/myrepo) or on OCI repository (e.g. oci://localhost:5000/myrepo).
+	// +kubebuilder:validation:MinLength=1
```
ahmedwaleedmalik: We should add a regex to validate this as well. Maybe something like 

```
+kubebuilder:validation:Pattern="(http|oci|https)://([a-zA-Z0-9~!@#$%^&*()_=+/?.:;',-]*)?"
```

pkg/apis/apps.kubermatic/v1/application_definition.go
```diff
@@ -57,25 +64,50 @@ type GitCredentials struct {
...
-	Username *GlobalSecretKeySelector `json:"username,omitempty"`
+	// Username holds the ref and key in the secret for the username credential. Secret must exist in the namespace where
+	// KKP is installed.
+	Username *corev1.SecretKeySelector `json:"username,omitempty"`
 
-	// Password holds the ref and key in the secret for the Password credential
-	Password *GlobalSecretKeySelector `json:"password,omitempty"`
+	// Password holds the ref and key in the secret for the Password credential. Secret must exist in the namespace where
+	// KKP is installed.
+	Password *corev1.SecretKeySelector `json:"password,omitempty"`
```
ahmedwaleedmalik, line 73: Can we reuse `HelmCredentials` here and move Token, SSHKey to that struct as well?

ahmedwaleedmalik (commented):
minor nits

pkg/apis/apps.kubermatic/v1/application_definition.go
vgramer: as discussed, the requirements are different. For  Helm, credentials are required, for git they are optional.

pkg/apis/apps.kubermatic/v1/application_definition.go
vgramer: good point however I have used the more open regex `^(http|https|oci)://.+` because I am not totally sure that it's an URL. i doubt that but it could be an URI. Helm documentation only said  `OCI registry references are now always prefixed with oci://`

ahmedwaleedmalik (approved):
/approve ⚡

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
/retest

`pre-kubermatic-dualstack-e2e` is super flaky so this should eventually just pass and the PR will get merged, after automated retries.

PR #9893, opened 2022-05-25
Add Openstack csi external snapshotter and default VolumeSnapshotClasses

### LINKED ISSUE #9013: Implement CSI Snapshotter

mate4st:

Adds missing external snapshotter components for openstack.

Adds VolumeSnapshotClass for Openstack, Vsphere and Nutanix (At the time of writing we deploy the snapshot controller for them)

Fixes #9013

```release-note
Add external snapshotter for Openstack CSI
Add default VolumeSnapshotClass for supported provider
```


addons/csi/openstack/snapshot-webhook.yaml
```diff
@@ -0,0 +1,77 @@
...
+    metadata:
+      labels:
+        app: snapshot-validation
+    spec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: snapshot-validation
+          image: k8s.gcr.io/sig-storage/snapshot-validation-webhook:v4.2.1 # change the image if you wish to use your own custom validation server image
```
xrstf: This should probably make use of `{{ Registry "k8s.gcr.io" }}`, like all addon manifests.

xrstf (changes requested):
Registry overwrites?

xrstf (approved):
/approve

PR #9894, opened 2022-05-25
Ignore conflicting operation on bucket lifecycle update

pkprzekwas:
- ignore conflicting operation on the bucket lifecycle update. The problem is that in the seed controller we set watchers for events that we create (e.g. cronjobs). Combined with `MaxConcurrentReconciles` parameter above 1, it leads to the situation where multiple seed reconciliations happen at nearly the same time. In such case trying to update S3 lifecycle fails due to the lock on configuration update operation. 
- fix metering configuration update (allow updating the config with empty value to remove retention)

This could be alternatively resolved by fetching existing lifecycle configuration and comparing it with the metering configuration. Initial testing showed that this approach works but I assume it still may fail occasionally as S3 configuration updates are eventually consistent. It was one of reasons why I chose to ignore the error instead. Another benefit of ignoring the error is reduced complexity of solution.

```release-note
NONE
```

### DIFF
--- a/pkg/ee/metering/reconcile.go
+++ b/pkg/ee/metering/reconcile.go
@@ -27,6 +27,7 @@ package metering
 import (
 	"context"
 	"fmt"
+
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/lifecycle"
 
--- a/pkg/ee/metering/reconcile.go
+++ b/pkg/ee/metering/reconcile.go
@@ -27,6 +27,7 @@ package metering
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/minio/minio-go/v7/pkg/lifecycle"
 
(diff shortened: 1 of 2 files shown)


pkg/ee/metering/reconcile.go
```diff
@@ -136,6 +137,10 @@ func reconcileMeteringReportConfigurations(ctx context.Context, client ctrlrunti
 		return err
 	}
 	if err := mc.SetBucketLifecycle(ctx, bucket, config); err != nil {
+		// Ignore conflict error in case lock after previous reconciliation process still exists.
```
xrstf: Please see if you can cast the error to some concrete type and check its properties. A string comparision on errors is too brittle and will break soon. It looks like the Minio client has some HttpError stuff: https://github.com/minio/minio-go/search?q=conflict

xrstf (approved):
Bi-u-ti-full

/approve

Issue #9618, opened 2022-04-14
Integrate AWS dual-stack machine-controller API in KKP


rastislavs:
Integrate https://github.com/kubermatic/kubermatic/issues/9294 into KKP

Also test dual-stack end-to end on all supported OS platforms.

PratikDeoghare:
Integration is complete. It is not yet supported for some combinations of cloud provider, cni and operating systems. 

Supported/un-supported combinations are tracked in [this test](https://github.com/kubermatic/kubermatic/blob/master/pkg/test/dualstack/cloud_providers_dualstack_test.go#L94) where unsupported combinations are commented out. 

This epic tracks progress of investigations of unsupported combinations. #9994

PR #9895, opened 2022-05-25
enable dualstack test for aws+rhel+canal

PratikDeoghare:
Enables dualstack test for aws+rhel+canal.

Partial fulfillment of #9618 

```release-note
NONE
```


PratikDeoghare:
Supported/un-supported combinations are tracked in [this test](https://github.com/kubermatic/kubermatic/blob/master/pkg/test/dualstack/cloud_providers_dualstack_test.go#L94) where unsupported combinations are commented out. 

This epic tracks progress of investigations of unsupported combinations. #9994

PR #9898, opened 2022-05-25
updates in KubeVirtCloudSpec to inject infraStorageClass info

### LINKED ISSUE #8546: Initialize a standard storage class for user cluster
We have to support storage class initialization on the user cluster that customers can use hot-pluggable disks.

sankalp-r:
This PR introduces a new field `InfraStorageClasses` in `KubevirtCloudSpec` which holds the names of all the Kubevirt Infra-StorageClasses. Later this info is passed to `csi-driver-opeartor.yaml` for `kubevirt-csi-driver-operartor` ADDON creation.

Fixes #8546 

**Documentation**:

```release-note
Added support for storage classes initialization on KubeVirt user clusters that customers can use hot-pluggable disks.
```


pkg/addon/template.go
```diff
@@ -201,6 +207,8 @@ type ClusterData struct {
 	MLA MLASettings
 	// CSIMigration indicates if the cluster needed the CSIMigration
 	CSIMigration bool
+	// KubeVirtInfraStorageClasses is a list of KubeVirt Infra-cluster storageClass
```
mfranczy: please be more descriptive about this field, you could change the comment to something like this:
```
KubeVirtInfraStorageClasses is a list of storage classes from KubeVirt infra cluster that are used to 
initialization of user cluster storage classes by the CSI driver kubevirt (hot pluggable disks)
```

pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -177,3 +184,22 @@ func GetCredentialsForCluster(cloud kubermaticv1.CloudSpec, secretKeySelector pr
 
 	return kubeconfig, nil
 }
+
+func (k *kubevirt) updateInfraStorageClassesInfo(ctx context.Context, spec *kubermaticv1.CloudSpec) error {
+	kubeConfig, err := GetCredentialsForCluster(*spec, k.secretKeySelector)
+	if err != nil {
+		return err
+	}
+	client, _, err := NewClientWithRestConfig(kubeConfig)
```
mfranczy: There is no need to create your own client instance here, we have a function to list storage class already
https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/common/provider/kubevirt.go#L207

pkg/provider/cloud/kubevirt/provider.go
```diff
@@ -177,3 +184,22 @@ func GetCredentialsForCluster(cloud kubermaticv1.CloudSpec, secretKeySelector pr
 
 	return kubeconfig, nil
 }
+
+func (k *kubevirt) updateInfraStorageClassesInfo(ctx context.Context, spec *kubermaticv1.CloudSpec) error {
```
mfranczy: There is no need to bind function `updateInfraStorageClassesInfo` to the `kubevirt` type. Please move the function to the `storage.go` file.

pkg/addon/template.go
```diff
@@ -102,6 +102,11 @@ func NewTemplateData(
 		ipvs = *cluster.Spec.ClusterNetwork.IPVS
 	}
 
+	var kubeVirtStorageClasses []string
+	if cluster.Spec.Cloud.Kubevirt != nil {
+		kubeVirtStorageClasses = append(kubeVirtStorageClasses, cluster.Spec.Cloud.Kubevirt.InfraStorageClasses...)
```
mfranczy: you don't have to use `append` to copy, you can just pass-by-value
```
kubeVirtStorageClasses = cluster.Spec.Cloud.Kubevirt.InfraStorageClasses
```

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -1100,6 +1100,7 @@ type KubevirtCloudSpec struct {
 	CSIKubeconfig string `json:"csiKubeconfig,omitempty"`
 
 	PreAllocatedDataVolumes []PreAllocatedDataVolume `json:"preAllocatedDataVolumes,omitempty"`
+	InfraStorageClasses     []string                 `json:"infraStorageClasses,omitempty"`
```
mfranczy: Please add a comment for this field and Pre-allocated data volume field. Eventually we will add comments to each field.

pkg/addon/template.go
```diff
@@ -145,7 +150,8 @@ func NewTemplateData(
 				MonitoringEnabled: cluster.Spec.MLA != nil && cluster.Spec.MLA.MonitoringEnabled,
 				LoggingEnabled:    cluster.Spec.MLA != nil && cluster.Spec.MLA.LoggingEnabled,
 			},
-			CSIMigration: csiMigration,
+			CSIMigration:                csiMigration,
+			KubeVirtInfraStorageClasses: kubeVirtStorageClasses,
```
mfranczy, line 154: I am not sure if we should not add a new struct for cloud provider specifc data.
We already have a field `cloudProviderName`, so we could create 
```
type CloudProviderData struct {
    KubeVirtData // struct ...
}
```

I am not sure because KubeVirt is an exception and maybe other cloud providers do not need this.
Don't change that yet, we need another opinion.

mfranczy (changes requested):
Generally speaking, from sig-virt perspective looks good to me, but there is a few things to fix.

mfranczy:
and please add a release note for this that we finally support storage classes initialization.

pkg/provider/cloud/kubevirt/provider.go
mfranczy: as discussed with @sankalp-r, that was a bad idea, @sankalp-r will do small refactoring and move the functions to the provider that we pull provider pkg -> handler common.

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -167,3 +170,42 @@ func createPreAllocatedDataVolume(dv kubermaticv1.PreAllocatedDataVolume, namesp
 		},
 	}, nil
 }
+
+func KubeVirtStorageClasses(ctx context.Context, client ctrlruntimeclient.Client) (apiv2.StorageClassList, error) {
```
mfranczy: please rename that to `ListStorageClasses`

pkg/provider/cloud/kubevirt/storage.go
```diff
@@ -167,3 +170,42 @@ func createPreAllocatedDataVolume(dv kubermaticv1.PreAllocatedDataVolume, namesp
...
+	storageClassList := storagev1.StorageClassList{}
+	if err := client.List(ctx, &storageClassList); err != nil {
+		return nil, err
+	}
+
+	res := apiv2.StorageClassList{}
+	for _, sc := range storageClassList.Items {
+		res = append(res, *newAPIStorageClass(&sc))
```
mfranczy: could you just append the object instead of having a function that returns the object?

mfranczy (changes requested):
Two last things

ahmedwaleedmalik (commented):
/approve

PR #9909, opened 2022-05-26
Add delete propagation policy while removing metering cron jobs

### LINKED ISSUE #9822: Cleanup orphaned metering jobs
After deleting kubernetes cron job its child jobs are left orphaned in the cluster. Extend controller to remove orphaned metering kuberentes jobs.

pkprzekwas:
At the moment controller leaves orphaned jobs and pods after deleting cronjob during reconciliation. My initial though was that DeletePropagationBackground is a default option and we do not need to set it explicitly but I guess I was wrong.

Additionally, I included fix for updating metering configuration. It no longer overwrites the whole metering object. 

Fixes #9822 

```release-note
NONE
```


xrstf (approved):
/approve

PR #9911, opened 2022-05-27
adjust resources and probes for konnectivity proxy sidecar

PratikDeoghare:
- Increases memory limits for konnectivity-proxy sidecar. 
- Allows overriding the limits via components override in the cluster-resource. 
- Adjusts readiness  probe timings.

```release-note
Fix probes, resources and allow overriding resource requests/limits for Konnectivity proxy via components override in the cluster resource
```


PratikDeoghare:
@ahmedwaleedmalik The dualstack e2e test was made optional on Friday. Can we unhold this now?

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -675,6 +676,10 @@ type APIServerSettings struct {
 	NodePortRange               string `json:"nodePortRange,omitempty"`
 }
 
+type KonnectvityProxySettings struct {
+	DeploymentSettings `json:",inline"`
+}
```
embik: Neither replicas nor tolerations from `DeploymentSettings` are respected here because konnectivity is not a `Deployment` but a sidecar. Shouldn't this be this instead?

```suggestion
type KonnectvityProxySettings struct {
	Resources   *corev1.ResourceRequirements `json:"resources,omitempty"`
}
```

embik (changes requested):

pkg/apis/kubermatic/v1/cluster.go
PratikDeoghare: Definitely. Thanks!

embik (approved):
/approve

embik (changes requested):
~~From `pkg/resources/konnectivity/sidecar.go:126`, it seems that resources for the sidecar are set to `defResourceRequirements`, so I think this PR is missing loading the resource overrides (unless I missed something).~~

embik:
Never mind my previous comments, the resource overrides are happening at a higher level. I guess this works, just needs fixing of the verify test.

PratikDeoghare:
@embik This https://github.com/kubermatic/kubermatic/blob/3ef51e9c5e3a9a24b73354bf6d040631786f4227/pkg/resources/apiserver/deployment.go#L213 overrides the resources. The PR changes the GetOverrides function https://github.com/kubermatic/kubermatic/blob/7a120d482b21a6657474c467c703a715f0355a96/pkg/resources/resources.go#L1375.  

Should work if I didn't miss something.

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -674,6 +674,8 @@ type ComponentSettings struct {
 	// the `LoadBalancer` expose strategy is used. This is not effective if a different expose
 	// strategy is configured.
 	NodePortProxyEnvoy NodeportProxyComponent `json:"nodePortProxyEnvoy"`
+	// KonnectivityProxy configures resources limits/requests for konnectivity-server sidecar
```
embik: super super uber nitpick, but there's a period missing at the end of the sentence.

embik (approved):
/approve

PR #9912, opened 2022-05-27
Validate cluster version against supported versions in webhook

embik:
Since we're moving towards GitOps, it is possible that users create clusters via CRs directly. At the moment, we do not validate `spec.version` for our `Cluster` resources in the validating webhook. This PR adds that additional validation. The KKP API does this already, so we should strive for "feature parity" here, IMHO.

Fixes #

**Documentation**:

```release-note
`Cluster` custom resources are validated against supported Kubernetes versions as defined in `KubermaticConfiguration` or as defaulted by the respective KKP release
```


embik:
@zreigz can you please take a look? The sig-api part is really small :-)

PR #9917, opened 2022-05-27
Make Nutanix e2e optional

embik:

Nutanix setup is still not available, so this makes Nutanix e2e tests optional for the time being.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow/provider-nutanix.yaml
+++ b/.prow/provider-nutanix.yaml
@@ -15,6 +15,7 @@
 presubmits:
   - name: pre-kubermatic-e2e-nutanix-ubuntu-1.24
     decorate: true
+    optional: true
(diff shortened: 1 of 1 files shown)


.prow/provider-nutanix.yaml
```diff
@@ -15,6 +15,7 @@
 presubmits:
   - name: pre-kubermatic-e2e-nutanix-ubuntu-1.24
     decorate: true
+    optional: true
     run_if_changed: "(addons/csi/nutanix|pkg/provider/cloud/nutanix)"
```
xmudrii: If there's no Nutanix setup, we probably shouldn't run those tests at all, so I'd drop this `run_if_changed`.

embik: Good point. Done.

xmudrii (approved):

PR #9922, opened 2022-05-27
 run dualstack e2e for Azure in East US region 

PratikDeoghare:
Dualstack e2e tests were flaky due to hitting quota limits on Azure West Europe region. 
This PR makes the tests run in East US where we don't hit the quota limits (at least right now). 

```release-note
NONE
```


xmudrii:
We fixed the cleanup yesterday, so this is not needed any longer.

PR #9926, opened 2022-05-30
extend the user settings for new flag

### LINKED ISSUE #9915: Add user setting for using clusters as default view inside project

Backend update for https://github.com/kubermatic/dashboard/issues/4510.

### Additional information

Add simple flag to the user settings object that will allow users saving their preference on which view inside project should be the default one (overview or clusters).

zreigz:
**What does this PR do / Why do we need it**: flag to the user settings object that will allow users saving their view preferences

Fixes #9915 

```release-note
New flag for the user settings to allow users saving their view preferences
```

### DIFF
--- a/pkg/apis/kubermatic/v1/user.go
+++ b/pkg/apis/kubermatic/v1/user.go
@@ -88,6 +88,7 @@ type UserSettings struct {
 	CollapseSidenav            bool   `json:"collapseSidenav,omitempty"`
 	DisplayAllProjectsForAdmin bool   `json:"displayAllProjectsForAdmin,omitempty"`
 	LastSeenChangelogVersion   string `json:"lastSeenChangelogVersion,omitempty"`
+	UseClustersView            bool   `json:"useClustersView"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/apis/kubermatic/v1/user.go
```diff
@@ -88,6 +88,7 @@ type UserSettings struct {
 	CollapseSidenav            bool   `json:"collapseSidenav,omitempty"`
 	DisplayAllProjectsForAdmin bool   `json:"displayAllProjectsForAdmin,omitempty"`
 	LastSeenChangelogVersion   string `json:"lastSeenChangelogVersion,omitempty"`
+	UseClustersView            bool   `json:"useClustersView"`
```
kron4eg: To maintain backward compatibility all new fields has to be:
* omitempty (i.e. optional)
* their lack should be handled in code (i.e. code should be OK with the fact that field wasn't set)

kron4eg:
to stop useless retests

PR #9935, opened 2022-05-31
Update Openstack CCM and CSI to 1.24.0

embik:
What it says on the title. Openstack CCM and CSI support 1.24.0 now, so this adds that version to KKP for Kubernetes 1.24+. This PR is mostly based on https://github.com/kubermatic/kubeone/pull/2061 to make sure we offer the same configuration across Kubermatic products.

xref #9818

**Documentation**:

```release-note
Add support for OpenStack CCM v1.24.0 (for Kubernetes 1.24 clusters)
Add support for OpenStack Cinder CSI driver v1.24.0 (for Kubernetes 1.24 clusters)
Update CSI components in OpenStack Cinder CSI driver
```

### DIFF
--- a/addons/csi/openstack/snapshot-crds.yaml
+++ b/addons/csi/openstack/snapshot-crds.yaml
@@ -111,7 +111,7 @@ spec:
                 format: date-time
                 type: string
               error:
-                description: error is the last observed error during snapshot creation, if any. This field could be helpful to upper level controllers(i.e., application controller) to decide whether they should continue on waiting for the snapshot to be created based on the type of error reported. The snapshot controller will keep retrying when an error occurrs during the snapshot creation. Upon success, this error field will be cleared.
+                description: error is the last observed error during snapshot creation, if any. This field could be helpful to upper level controllers(i.e., application controller) to decide whether they should continue on waiting for the snapshot to be created based on the type of error reported. The snapshot controller will keep retrying when an error occurs during the snapshot creation. Upon success, this error field will be cleared.
                 properties:
                   message:
                     description: 'message is a string detailing the encountered error during snapshot creation if specified. NOTE: message may be logged, and it should not contain sensitive information.'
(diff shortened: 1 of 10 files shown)


addons/csi/openstack/snapshot-crds.yaml
```diff
@@ -13,11 +13,7 @@
 # limitations under the License.
 
 # Sourced from
-# https://github.com/kubernetes-csi/external-snapshotter/tree/release-4.2/client/config/crd
-
-{{ if .Cluster.Features.Has "externalCloudProvider" }}
-{{ if eq .Cluster.CloudProviderName "openstack" }}
```
mate4st, line 19: I think we need this

mate4st (changes requested):

addons/csi/openstack/snapshot-crds.yaml
embik: ah damn it, that was unintentional. Thanks!

WeirdMachine:
/hold 

waiting for https://github.com/kubermatic/kubermatic/pull/9893

pkg/resources/controllermanager/deployment.go
```diff
@@ -273,7 +273,7 @@ func getFlags(data *resources.TemplateData, version *semverlib.Version) ([]strin
 	if cloudProviderName != "" && cloudProviderName != "external" {
 		flags = append(flags, "--cloud-provider", cloudProviderName)
 		flags = append(flags, "--cloud-config", "/etc/kubernetes/cloud/config")
-		if cloudProviderName == "azure" && version.Minor() >= 15 {
+		if cloudProviderName == "azure" && version.GreaterThan(semverlib.MustParse("1.15.0")) {
```
dermorz: Good thing we are checking the whole version now. Is the change from GreaterThanEqual to GreaterThen intended?

embik: It is actually not, good catch. I guess it doesn't make a difference because we're long past 1.15 as a supported version, but let me fix this real quick.

embik: fixed with a constraint.

embik:
/hold cancel

#9893 is merged, let's go 🚀

embik:
/hold cancel

#9893 is merged, let's go 🚀

PR #9941, opened 2022-05-31
wwhrd.yml: add github.com/acomagu/bufpipe to the exeption list

vgramer:
For #8379  we use [go-git](https://github.com/go-git/go-git) to interact with git repositroy. [go-git](https://github.com/go-git/go-git) has a dependency on [acomagu/bufpipe](https://github.com/acomagu/bufpipe) which is under MIT license.
the license is not recognized because it's slightly different from the one provided by SDPX

```
	diff  'SDPX-MIT.txt' 'bufpipe-LICENSE.txt'
	3c3
	< Copyright (c) <year> <copyright holders>
	---
	> Copyright (c) 2019 acomagu
	7c7
	< The above copyright notice and this permission notice (including the next paragraph) shall be included in all copies or substantial portions of the Software.
	---
	> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
```

full license: [https://github.com/acomagu/bufpipe/blob/master/LICENSE](https://github.com/acomagu/bufpipe/blob/master/LICENSE )

/hold
/assign @scheeles 

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.wwhrd.yml
+++ b/.wwhrd.yml
@@ -28,6 +28,7 @@ exceptions:
   - github.com/ajeddeloh/go-json # Since it's a fork, https://github.com/golang/go/blob/master/LICENSE
   - github.com/cristim/ec2-instances-info # Public domain: https://github.com/cristim/ec2-instances-info/blob/master/LICENSE.
   - github.com/cristim/ec2-instances-info/data # MIT: https://github.com/powdahound/ec2instances.info/blob/master/LICENSE.
+  - github.com/acomagu/bufpipe # MIT: https://github.com/acomagu/bufpipe/blob/master/LICENSE (used by go-git)
   - github.com/embik/nutanix-client-go/internal/utils # MPL-2.0
   - github.com/embik/nutanix-client-go/pkg/client # MPL-2.0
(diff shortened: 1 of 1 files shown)


vgramer:
@xrstf  IIRC it's a leftover from our first draft when we had only one type of Source: git. It was an implementation hint to have a generic interface to download the application's source.  the last 2 points of the issue should be in the parent epic...

Currently, 2 source implementations are planned:
* Git Source
* Helm Source (download from Helm repository)

I'm not aware of the need for another SCM Source...

xrstf:
So this would lead to the seed-ctrl-mgr doing Git operations in its own memory? That makes me extremely uncomfortable from a security standpoint. Why is shelling out to git not possible, like we do for Helm?

vgramer:
as discussed in the sig-app-management meeting:
* implementation of the controller will be part of user-cluster-controller-manager 
* dealing with errors is easier with go library
* using go library keeps credentials in memory
* when cloning the repository a swallow clone will be performed to reduce the amount of downloaded data

/unhold

PR #9952, opened 2022-06-01
update integration-test image to use Go 1.18

xrstf:
This will prevent the integration tests from failing to find a gocache. There are no 1.24.x binaries yet available, see https://github.com/kubernetes-sigs/kubebuilder/issues/2559 and related issues.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #9953, opened 2022-06-01
Bump gopkg.in/yaml.v3 to v3.0.1

xrstf:
This fixes https://github.com/kubermatic/kubermatic/security/dependabot/6. I chose to update as many places as I could

* in a few places we were unmarshalling into a map[string]interface{} and only used UnmarshalStrict() so that maybe in the future we could have a linter that forbids the normal Unmarshal() -- now with the new v3 API, we could not easily forbid a function, so there is no point in going the extra mile to configure a decoder if the target is just a map
* our yamled package needed a complete rewrite, as yaml.v2 and v3 use totally different data structures internally. Thankfully we had a bunch of unit tests and the v3 API is actually easier to use.

```release-note
NONE
```


pkg/controller/seed-controller-manager/mla/alertmanager_controller.go
```diff
@@ -348,10 +348,13 @@ func (r *alertmanagerController) ensureAlertmanagerConfiguration(ctx context.Con
 	if err != nil {
 		return err
 	}
+
 	expectedConfig := map[string]interface{}{}
-	if err := yaml.UnmarshalStrict(config, &expectedConfig); err != nil {
+	decoder := yaml.NewDecoder(bytes.NewReader(config))
```
PratikDeoghare, line 353: missing `decoder.KnownFields(true)`?

PratikDeoghare: Maybe a func like this can protect against this and also be easy to grep. 

```
func NewStrictDecoder(r io.Reader) *yaml.Decoder {
	decoder := yaml.NewDecoder(r)
	decoder.KnownFields(true)
	return decoder
}

```

PratikDeoghare, line 353: missing `decoder.KnownFields(true)`?

pkg/controller/seed-controller-manager/mla/org_grafana_controller_test.go
```diff
@@ -394,11 +394,16 @@ func jsonEqual(expectedBody, body []byte) bool {
 func yamlEqual(expectedBody, body []byte) bool {
 	expectedBodyMap := map[string]interface{}{}
 	bodyMap := map[string]interface{}{}
-	if err := yaml.UnmarshalStrict(expectedBody, &expectedBodyMap); err != nil {
+
```
PratikDeoghare, line 397: missing `decoder.KnownFields(true)`?

pkg/controller/seed-controller-manager/mla/rulegroup_controller.go
```diff
@@ -319,7 +319,8 @@ func (r *ruleGroupController) ensureRuleGroup(ctx context.Context, ruleGroup *ku
 		return err
 	}
 	expectedRuleGroup := map[string]interface{}{}
-	if err := yaml.UnmarshalStrict(ruleGroup.Spec.Data, &expectedRuleGroup); err != nil {
+	decoder := yaml.NewDecoder(bytes.NewReader(ruleGroup.Spec.Data))
```
PratikDeoghare, line 322: missing `decoder.KnownFields(true)`?

pkg/handler/v2/alertmanager/alertmanager.go
```diff
@@ -110,9 +111,12 @@ type updateAlertmanagerReq struct {
 
 func (req *updateAlertmanagerReq) validateUpdateAlertmanagerReq() error {
 	bodyMap := map[string]interface{}{}
-	if err := yaml.UnmarshalStrict(req.Body.Spec.Config, &bodyMap); err != nil {
+
+	decoder := yaml.NewDecoder(bytes.NewReader(req.Body.Spec.Config))
```
PratikDeoghare, line 115: missing `decoder.KnownFields(true)`?

pkg/handler/v2/rulegroup/request.go
```diff
@@ -193,7 +192,7 @@ func DecodeRuleGroupID(r *http.Request) (string, error) {
 
 func GetRuleGroupNameInData(data []byte) (string, error) {
 	bodyMap := map[string]interface{}{}
-	if err := yaml.UnmarshalStrict(data, &bodyMap); err != nil {
+	if err := yaml.Unmarshal(data, &bodyMap); err != nil {
```
PratikDeoghare, line 195: `v3.Unmarshal` is not equivalent to `v2.UnmarshalStrict`. Is this intentional? 

https://play.golang.org/p/tb4QNt5jAAk

PratikDeoghare:
> so there is no point in going the extra mile to configure a decoder if the target is just a map

Read this after leaving bunch of comments. Apologies. 

We can wrap the code to configure the decoder and ban that with linter.

PR #9954, opened 2022-06-02
add e2e test for dualstack clusters on openstack

PratikDeoghare:
Adds e2e test for dualstack clusters on openstack.

Fixes part of #9929 

**Documentation**:

```release-note
add e2e test for dualstack clusters on openstack
```

### DIFF
--- a/hack/ci/testdata/seed.yaml
+++ b/hack/ci/testdata/seed.yaml
@@ -159,6 +159,24 @@ spec:
             minimumMemory: 0
             minimumVCPUs: 0
           region: dbl
+    syseleven-fes1:
+      country: DE
+      location: Syseleven - fes1
+      spec:
+        openstack:
+          authURL: https://api.cbk.cloud.syseleven.net:5000/v3
+          availabilityZone: fes1
+          dnsServers:
+            - 37.123.105.116
+            - 37.123.105.117
+          enforceFloatingIP: true
+          images:
+            centos: "CentOS 8 (2021-07-05)"
+            ubuntu: "Ubuntu Focal 20.04 (2021-07-01)"
+          nodeSizeRequirements:
+            minimumMemory: 0
+            minimumVCPUs: 0
+          region: fes
     nutanix-ger:
       country: DE
       location: Hamburg
--- a/hack/ci/testdata/seed.yaml
+++ b/hack/ci/testdata/seed.yaml
@@ -176,7 +176,7 @@ spec:
           nodeSizeRequirements:
             minimumMemory: 0
             minimumVCPUs: 0
-          region: fes
+          region: fes # ipv6 is only available in this zone
     nutanix-ger:
       country: DE
       location: Hamburg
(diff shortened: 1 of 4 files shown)


hack/ci/testdata/seed.yaml
```diff
@@ -159,6 +159,24 @@ spec:
...
+        openstack:
+          authURL: https://api.cbk.cloud.syseleven.net:5000/v3
+          availabilityZone: fes1
+          dnsServers:
+            - 37.123.105.116
+            - 37.123.105.117
+          enforceFloatingIP: true
+          images:
+            centos: "CentOS 8 (2021-07-05)"
+            ubuntu: "Ubuntu Focal 20.04 (2021-07-01)"
```
ahmedwaleedmalik, line 175: Why did we change the availability zone and images?

pkg/test/dualstack/cloud_providers_dualstack_test.go
```diff
@@ -176,6 +177,34 @@ func TestCloudClusterIPFamily(t *testing.T) {
 			skipNodes:           true,
 			skipHostNetworkPods: true,
 		},
+		{
+			cloudName: "openstack",
+			osNames: []string{
+				// "centos",
+				// "flatcar",
+				// "rhel",
+				// "sles",
```
ahmedwaleedmalik: Please remove these comments, you should add them back once they are implemented. Not sure if there is any value in having commented out values here.

pkg/test/dualstack/cloud_providers_dualstack_test.go
```diff
@@ -176,6 +177,34 @@ func TestCloudClusterIPFamily(t *testing.T) {
...
+			skipNodes:           true,
+			skipHostNetworkPods: true,
+		},
+		{
+			cloudName: "openstack",
+			osNames: []string{
+				"centos",
+				// "flatcar",
+				// "rhel",
+				// "sles",
```
ahmedwaleedmalik: Please remove these comments, you should add them back once they are implemented. Not sure if there is any value in having commented out values here.

hack/ci/testdata/seed.yaml
rastislavs: IPv6 is available only in this zone, and images are from the ones that are available in that zone. 
@PratikDeoghare maybe it would make sense to leave a comment for that.

pkg/test/dualstack/cloud_providers_dualstack_test.go
rastislavs: Agree, just listing them does not bring that much benefit. I would propose to add reason, e.g. `centos - not working because of ...`, then it makes sense as documentation

PratikDeoghare: Will remove because reasons are not always available.

ahmedwaleedmalik (approved):
/approve

PR #9960, opened 2022-06-02
WIP Support for groups

### LINKED ISSUE #9897: Groups support for Project roles
Add OIDC groups support to KKP, so that Groups can be assigned project roles, instead of individual users.

The proposal is here: https://docs.google.com/document/d/12hwWdReEyNlpPA16ms960B-YRdDqyn7PIH84XVmGCwA/edit#heading=h.gqpzn4xl04pb

lsviben:
Adds support for groups In KKP by:
- adding a new CRD GroupProjectBinding, which binds groups to projects and to the KKP defined roles (owners, viewers, editors, projectmanagers) 
- adding group support rbac controller
- makes the KKP API use the groups received from authorization in the impersonation client

As the rbac-controller is a complicated and also essential component, this PR tries to not refactor it too much. Rather gently sliding in the changes for the group support.
Working on this, I noticed that one of the biggest issues with the component is that it considers groups and roles the same, because in KKP this was the case. Now as we want to add custom groups, but bound to a KKP role, the goal becomes to unravel the roles and groups in the rbac controller.

There is a need for a deeper refactor of the rbac-controller, to make it more friendly :). But I would leave that for another PR.

This is a WIP, it still misses additional unit tests, cleanup logic, and general cleaning. 

Fixes #9897 

```release-note
Adds support for assigning groups to projects in a similar way the users are assigned now. For an user to have access through a group, they need to authenticate through a platform which supports groups.
```


lsviben:
closing this as its replaced by smaller PRs

Issue #9392, opened 2022-03-21
Implement support for Azure CSI drivers


xmudrii:
Add support for Azure disk and file CSI drivers.

## Acceptance Criteria

- Azure Disk CSI driver is implemented.
- Azure File CSI driver is implemented.

PR #9963, opened 2022-06-02
Azure external CCM support

### LINKED ISSUE #8290: Azure route table not assigned to subnet
**User Story**

In https://github.com/kubermatic/kubermatic/blob/b4a028de8ca96b2b18cce4e395015ffa7e2536b9/pkg/provider/cloud/azure/provider.go#L533-L544 a route table is created in Azure. It is passed a subnet reference, presumably to connect the route table to the subnet.

Unfortunately, this is ineffective and doesn't happen. Azure just silently ignores the `RouteTablePropertiesFormat.Subnets` argument, it seems. If you check the [API documentation](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/route-tables/create-or-update#request-body) for the API endpoint that is called here, you'll see that it only accepts a list of routes as a property, not a subnet.

Assigning a route table to a subnet happens in the subnet API calls. According to git blame, this is in there since the very first PR that introduced Azure support 4 years ago. It's not clear to me if the API changed since then (needs to be checked) and when it changed.

We should be hesitant to reconcile this situation (e.g. via the better provider features introduced for Azure in #8213), although I believe adding the route table to the subnet would not cause any problems, since it's empty. But we should be very careful not to break any existing setups with that.

**Acceptance criteria**

- Remove or reconcile Azure route table in a way that it's actually in use

### LINKED ISSUE #9958: Support for external Azure CCM
We want to support the external Azure CCM (https://github.com/kubernetes-sigs/cloud-provider-azure) as in-tree implementations are deprecated.

## Acceptance Criteria

- Azure CCM is deployed if `externalCloudProvider: true` feature gate is set.
- Existing clusters can be migrated by enabling the feature gate.

embik:
This adds the external Azure CCM and supports migrating existing clusters to it. The Azure CCM actually comes in two components, the main CCM that will run in the control plane and a DaemonSet running on the user cluster. The later is implemented as an addon, mirroring what we do for the aws-node-termination-handler or for CSI drivers.

In addition, this fixes a kinda long standing bug with the route table not being properly attached to a subnet; that required some light modifications to the order of creating/deleting Azure resources. I figured out it's a good opportunity for fixing this since the CCM syncs routes to the route table.

Without CSI drivers, this isn't _fully_ functional, so I implemented a hack that doesn't automatically create Azure clusters with the CCM - That is going to be changed once someone/I work(s) on #9392. But I did not want to throw CSI drivers into what is already a large PR.

Fixes #9958
Fixes #8290

**Documentation**:

```release-note
- Attach previously unattached Azure route table to generated subnet
- Add support for Azure's external CCM and allow migrating existing Azure clusters to it
```


pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -173,9 +173,9 @@ func (h *AdmissionHandler) mutateCreate(newCluster *kubermaticv1.Cluster) error
 
 func (h *AdmissionHandler) mutateUpdate(oldCluster, newCluster *kubermaticv1.Cluster) error {
 	// This part of the code handles the CCM/CSI migration. It currently works
-	// only for OpenStack clusters, in the following way:
+	// for OpenStack, vSphere and Azure clusters, in the following way:
 	//   * Add the CCM/CSI migration annotations
-	//   * Enable the UseOctaiva flag
+	//   * Enable the UseOctaiva flag (for OpenStack only)
```
xrstf: UseOctavia (typo)

pkg/resources/cloudcontroller/azure.go
```diff
@@ -0,0 +1,214 @@
...
+					Image:   data.ImageRegistry(resources.RegistryMCR) + "/oss/kubernetes/azure-cloud-controller-manager:v" + version,
+					Command: []string{"cloud-controller-manager"},
+					Args:    getAzureFlags(data),
+					VolumeMounts: append(getVolumeMounts(),
+						corev1.VolumeMount{
+							Name:      resources.CloudConfigConfigMapName,
+							MountPath: "/etc/kubernetes/cloud",
+							ReadOnly:  true,
+						},
+						/*
```
xrstf: leftover or left on purpose?

pkg/provider/cloud/azure/subnet.go
```diff
@@ -53,6 +53,11 @@ func reconcileSubnet(ctx context.Context, clients *ClientSet, location string, c
 		return cluster, err
 	}
 
+	routeTable, err := clients.RouteTables.Get(ctx, cluster.Spec.Cloud.Azure.ResourceGroup, cluster.Spec.Cloud.Azure.RouteTableName, "")
+	if err != nil && !isNotFound(vnet.Response) {
```
lsviben:
```suggestion
	if err != nil && !isNotFound(routeTable.Response) {
```

embik: Good catch! thanks.

pkg/resources/cloudcontroller/azure.go
embik: I wasn't actually sure what to do with it. When I tried adding it, it did not recognize the Azure API's TLS cert anymore, so I guess I need to fix it before we merge this. Thanks for reminding me.

embik: I removed this because I don't see a use case for setting a custom TLS CA bundle when talking to Azure.

xrstf (approved):
/approve

PR #9972, opened 2022-06-03
Gracefully handle new Seeds that have no CRDs yet in our webhook

### LINKED ISSUE #9961: Adding new Seeds is denied by webhook
### What happened?

It can happen that the seed-webhook on the master rejects a seed because it cannot list Clusters on it (as the operator has not yet installed the CRDs into the Seed). This will lead to

> "message": "admission webhook \"[seeds.kubermatic.io](http://seeds.kubermatic.io/)\" denied the request: failed to list clusters: no matches for kind \"Cluster\" in version \"[kubermatic.k8c.io/v1](http://kubermatic.k8c.io/v1)\""

### Expected behavior

`kubectl apply -f newseed.yaml` (creating a new Seed) should work.

The KKP operator should either check if the Seed already has a status (i.e. was reconciled once) or gracefully handle non-existing CRDs on it (if no Cluster CRD exists, there can be no clusters, so from the webhook's point of view, CRDs missing should be perfectly fine)

* Create a Seed as YAML
* Try to apply it

- KKP version: master branch
- Shared or separate master/seed clusters?: N/A

xrstf:
Since we now manage CRDs ourselves (via the KKP operator), it's now normal for admins to add new, empty clusters as seeds and expect KKP to set it up. But the KKP webhook would reject new seeds, as it could not query them for clusters, which caused a chicken-egg-problem.

This PR solves that by first probing for the CRD. This should be more stable and reliable than checking the SeedStatus for some flags.

**NOTE:** There is another PR coming to merge the seed-sync controller into the operator and further improve the seed setup experience. I kept this smaller PR standalone to make reviews easier.

Fixes #9961

```release-note
NONE
```

### DIFF
--- a/cmd/kubermatic-webhook/main.go
+++ b/cmd/kubermatic-webhook/main.go
@@ -45,6 +45,7 @@ import (
 	usersshkeyvalidation "k8c.io/kubermatic/v2/pkg/webhook/usersshkey/validation"
 	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik (approved):
/approve

PR #8480, opened 2021-12-16
Add ApplicationInstallation CRD

SimonTheLeg:
Signed-off-by: Simon Bein <simontheleg@gmail.com>

**What this PR does / why we need it**:

Introduces the ApplicationInstallation CRD which we are going to use to track Installations of Applications in user-clusters.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #8383 

**Does this PR introduce a user-facing change?**:

```release-note
none
```

PR #9975, opened 2022-06-03
Migration for ApplicationInstallations

ahmedwaleedmalik:

In our [initial implementation](https://github.com/kubermatic/kubermatic/pull/8480), `ApplicationInstallations` were `cluster` scoped but we later [decided to switch](https://github.com/kubermatic/kubermatic/pull/9601/files#diff-60650466b80ab7fc9d7a10d979f7c5958da3348fe965d51a1bd149d2ca7d129bL28) to `namespaced` scope resources.

This CRD was shipped with `KKP 2.19.0` release. And since `scope` is an immutable field, we need a way to migrate this CRD so that we can install the new one. This CRD is a component of an unreleased feature i.e. applications, so it's completely safe to just delete the old CRD and ship the new one.

We don't need this CRD on the seed cluster. In general, it's fine to install all the CRDs on seed but I think even in the future we'll have exceptions like these. So I've handled that with this PR as well.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/apps.kubermatic/v1/application_installation.go
+++ b/pkg/apis/apps.kubermatic/v1/application_installation.go
@@ -24,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const ApplicationInstallationsFQDNName = "applicationinstallations." + GroupName
+
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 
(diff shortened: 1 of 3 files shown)


vgramer (commented):
/lgtm

xrstf (approved):
/approve

last one for this month

PR #9977, opened 2022-06-03
Application: git source

### LINKED ISSUE #8379: Implement git source

vgramer:

Add code to download the sources from a git repository

Fixes #8379 

**Documentation**:

```release-note
NONE
```


pkg/applications/providers/types.go
```diff
@@ -0,0 +1,48 @@
...
+	// DownloadSource into the destination and return the full path to the source.
+	// destination must exist.
+	DownloadSource(destination string) (string, error)
+}
+
+// NewSourceProvider returns the concrete implementation of SourceProvider according to source defined in appSource.
+func NewSourceProvider(ctx context.Context, client ctrlruntimeclient.Client, appSource *appskubermaticv1.ApplicationSource, secretNamespace string) (SourceProvider, error) {
+	switch {
+	case appSource.Helm != nil:
+		return nil, fmt.Errorf("helmSource is not inmpleted yet")
```
ahmedwaleedmalik: `s/inmpleted/implemented`

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+		Progress:      nil,
+		ReferenceName: plumbing.NewBranchReferenceName(gitSource.Ref.Branch),
+		SingleBranch:  true,
+		Depth:         1,
+		Tags:          git.NoTags,
+	})
+	return err
+}
+
+// checkoutFromTag clone the repository and checkout the desired Tag. A swallow clone is performed.
```
ahmedwaleedmalik: `s/swallow/shallow`

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("failed to get worktree's repository: %w", err)
+	}
+	commit, err := repo.CommitObject(plumbing.NewHash(gitSource.Ref.Commit))
+	if err != nil {
+		return fmt.Errorf("can not find commit: %w", err)
+	}
+	return workTree.Checkout(&git.CheckoutOptions{Hash: commit.Hash})
+}
+
+// checkoutFromBranch clone the repository and checkout the desired branch. A swallow clone is performed.
```
ahmedwaleedmalik: `s/swallow/shallow`

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+
+// DownloadSource clone the repository into destination and return the full path to the application's sources.
+func (g GitSource) DownloadSource(destination string) (string, error) {
+	auth, err := g.authFromCredentials()
+	if err != nil {
+		return "", err
+	}
+
+	checkoutStrategy := g.getCheckoutStrategy()
+	if err := checkoutStrategy(g.Ctx, destination, g.Source, auth); err != nil {
```
ahmedwaleedmalik: nit

`checkoutStrategy(g.Ctx, destination, g.Source, auth)` looks a bit misleading. Since we are actually performing a `git checkout` here. Maybe we can do something like `checkout := g.getCheckoutStrategy()`

P.s. My concern here is just code coherence, nothing logical or syntactical.

pkg/applications/providers/source/git.go
vgramer: i also updated the one in tests too :)

cmd/user-cluster-controller-manager/main.go
```diff
@@ -382,7 +382,7 @@ func main() {
 		log.Info("Registered constraintsyncer controller")
 	}
 
-	if err := applicationinstallationcontroller.Add(rootCtx, log, seedMgr, mgr, isPausedChecker, &applications.ApplicationManager{}); err != nil {
+	if err := applicationinstallationcontroller.Add(rootCtx, log, seedMgr, mgr, isPausedChecker, &applications.ApplicationManager{SecretNamespace: runOp.namespace}); err != nil {
```
SimonTheLeg: q: So it was right to use the runOp.namespace, because it just gets propagated from master through seed to user, right?

go.mod
```diff
@@ -51,10 +52,12 @@ require (
 	github.com/open-policy-agent/frameworks/constraint v0.0.0-20210802220920-c000ec35322e
 	github.com/open-policy-agent/gatekeeper v0.0.0-20201111000257-4450f08fa95e
 	github.com/packethost/packngo v0.24.0
+	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/prometheus/client_golang v1.12.1
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/sirupsen/logrus v1.8.1
+	github.com/sosedoff/gitkit v0.3.0
```
SimonTheLeg, line 66: is there a reason we are adding gitkit? I don't see it being imported anywhere.

SimonTheLeg: Same goes for freeport

pkg/applications/installer.go
```diff
@@ -42,14 +44,35 @@ type ApplicationInstaller interface {
...
+	sourceProvider, err := providers.NewSourceProvider(ctx, seedClient, &applicationInstallation.Status.ApplicationVersion.Template.Source, a.SecretNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to initialize source provider: %w", err)
+	}
+
+	downloadDest, err := os.MkdirTemp("", applicationInstallation.Namespace+"-"+applicationInstallation.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed to create temp directory where application source will be downloaded: %w", err)
+	}
+	defer os.RemoveAll(downloadDest)
```
SimonTheLeg: since this can throw an error, we should at least log the error. So we can see if there might be a bug filling up the volume of the pod

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+			return fmt.Errorf("could not determine which reference to checkout")
+		}
+	}
+}
+
+// checkoutFunc define a function to clone and checkout code from repository defined in gitSource into destination using auth as credentials.
+type checkoutFunc func(ctx context.Context, destination string, gitSource *appskubermaticv1.GitSource, auth transport.AuthMethod) error
+
+// checkoutFromCommit clone the repository and checkout the desired commit.
+// If a branch is defined then a shallow clone of the branch is performed. The commit must belongs to this branch.
```
SimonTheLeg, line 149:
```suggestion
// If a branch is defined then a shallow clone of the branch is performed. The commit must belong to this branch.
```

pkg/applications/installer.go
SimonTheLeg: Since this can throw an error, I think we should at least log the error. This could help us in production if something unexpected happens

pkg/applications/installer.go
```diff
@@ -42,14 +44,35 @@ type ApplicationInstaller interface {
...
 		return err
 	}
-	// todo logic to download source and install app
+
+	appSourcePath, err := sourceProvider.DownloadSource(downloadDest)
+	if err != nil {
+		return fmt.Errorf("failed to download application source: %w", err)
+	}
+
+	log.Debugw("application download", "location", appSourcePath)
```
SimonTheLeg:
```suggestion
	log.Debugw("application successfully downloaded", "location", appSourcePath)
```

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+import (
+	"context"
+	"fmt"
+	"path"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
```
SimonTheLeg: as these have quite generic names, I think it might make sense to import them prefaced with "gogit":

```
gogittransport "github.com/go-git/go-git/v5/plumbing/transport"
```

Especially I think for funcs like authFromCredentials to make it easier for the consumer

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+	checkout := g.getCheckoutStrategy()
+	if err := checkout(g.Ctx, destination, g.Source, auth); err != nil {
+		return "", fmt.Errorf("failed to clone repository: %w", err)
+	}
+
+	return path.Join(destination, g.Source.Path), nil
+}
+
+// authFromCredentials returns a new transport.AuthMethod according to credentials defined in the GitSource.
+// If no crednetials are defined in the GitSource then nil is returned.
```
SimonTheLeg:
```suggestion
// If no credentials are defined in the GitSource then nil is returned.
```

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+			privateKey, err := g.getCredentialFromSecret(credentials.SSHKey.Name, credentials.SSHKey.Key)
+			if err != nil {
+				return nil, err
+			}
+
+			authssh, err := ssh.NewPublicKeys("git", []byte(privateKey), "")
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse private ssh key: %w", err)
+			}
+			authssh.HostKeyCallback = gossh.InsecureIgnoreHostKey()
```
SimonTheLeg: I guess for now it is okay to allow any host. Maybe we should add a small section to the explanation of the func, that for now all hosts are accepted by default

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+	var err error
+	if len(gitSource.Ref.Branch) > 0 {
+		repo, err = git.PlainCloneContext(ctx, destination, false, &git.CloneOptions{
+			URL:           gitSource.Remote,
+			Auth:          auth,
+			Progress:      nil,
+			SingleBranch:  true,
+			ReferenceName: plumbing.NewBranchReferenceName(gitSource.Ref.Branch),
+			Tags:          git.NoTags,
+			NoCheckout:    true,
```
SimonTheLeg: Now that I see this again, I am wondering if this is not too complicated for the user. Essentially this creates a lot of possible states:
* branch only → super fast, depth 1 shallow clone
* branch+commit → medium fast, depth max, but only for one branch shallow clone
* commit only → full clone, either medium fast (if not too many other branches) or slow
* tag only → super fast, depth 1 shallow clone

In the end, this is due to git's design, which does not allow you to shallow clone a commit. And we know this, but the end-user doesn't. Therefore I think, we should at least re-word the user facing section here https://github.com/vgramer/kubermatic/blob/5c835b66f7315b1cbbcd6257a5e32f91a15a5c96/pkg/apis/apps.kubermatic/v1/application_definition.go#L84
to include the information.

Maybe we could do something like:
"For large repositories we recommend to either use Tag, Branch+Commit, or Branch. This allows a shallow clone, which dramatically speeds up performance"

pkg/applications/providers/source/git_test.go
```diff
@@ -0,0 +1,701 @@
...
+// the content of the repository for feature-1 branch is:
+// .
+// ├── file1
+// ├── subdir
+// │   └── file2
+// └── subdir2
+//     └── file5
+func createGitRepository(t *testing.T, temp string) repoInfo {
+	repoDir := temp + "/repo1"
+	err := os.Mkdir(repoDir, 0700)
```
SimonTheLeg, line 610: mega nitpick, but I think we can completely create this in memory. afaik both go-git as well as go stdlib now support io/fs, which has an implementation for in-memory systems.

But again, just a nitpick :)

SimonTheLeg (changes requested):
only some smaller changes. Overall this is a great PR. Everything is documented and very good to understand. And testing is 👌

pkg/applications/providers/source/git.go
vgramer: done :)

pkg/applications/providers/source/git.go
vgramer: done :)

cmd/user-cluster-controller-manager/main.go
vgramer: As you know, application credentials are replicated from master to seed clusters into "kubermatic" namespace by the `application-secret-synchronizer`. 

The user cluster does not have the permission to read secrets into the "kubermatic" namespace.  I could add RBAC to provide access to user-cluster-controller-manager to the kubermatic namespace. Nevertheless, it feels wrong as this namespace has many sensible secrets (especially on master/seed setup). 

After discussing with @sig-cluster-management, we agree the better solution is to have a new controller in the seed-controller-manager that replicates application secrets into all cluster namespaces (e.g cluster-znx6n7jbdz). (it will be implemented in another PR...)

`runOp.Namespace`is the namespace where the user-cluster-controller is running (ie cluster-<cluster_id>).  Currently for testing, I manually copy application credentials into the cluster namespace....

pkg/applications/installer.go
```diff
@@ -42,14 +44,39 @@ type ApplicationInstaller interface {
...
 
 // Apply creates the namespace where the application will be installed (if necessary) and installs the application.
 func (a *ApplicationManager) Apply(ctx context.Context, log *zap.SugaredLogger, seedClient ctrlruntimeclient.Client, userClient ctrlruntimeclient.Client, applicationInstallation *appskubermaticv1.ApplicationInstallation) error {
+	// initialize tools
+	sourceProvider, err := providers.NewSourceProvider(ctx, seedClient, &applicationInstallation.Status.ApplicationVersion.Template.Source, a.SecretNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to initialize source provider: %w", err)
+	}
+
+	downloadDest, err := os.MkdirTemp("", applicationInstallation.Namespace+"-"+applicationInstallation.Namespace)
```
embik: This can potentially create HUGE tmp directories depending on the git source. We need to put in safeguards, like using an empty dir in the user-cluster-controller-manager Deployment, and limiting the amount of disk space this can eat. Otherwise you can fully consume the underlying node's disk.

embik: (I know there's a shallow clone happening but IMHO we should still make sure this gets its own temporary volume with storage limits on it)

pkg/applications/providers/source/git_test.go
vgramer: In fact, it's a bit more complicated. I use go-git to create a local repository and easily keep track of references needed for tests (tag, master branch ...) then I use `gitkit` to start a local git server that serves this repository over HTTP (with and without auth) and SSH.  So this repository can not be in memory.

notes:
* freeport is used to allocate a freeport when git server serves the repository over ssh.
* gitkit only need a bare git repository. I guess I could use the go-git plumbing command to create a bare repository but creating a plane one was easier....

go.mod
vgramer: it used for testing cf https://github.com/kubermatic/kubermatic/pull/9977#discussion_r892469553

pkg/applications/installer.go
vgramer: you're totally right, i forgot we don't have default limits :s  working on it.

go.mod
SimonTheLeg: ah now I get it. The test file is collapsed by default. Okay makes sense

cmd/user-cluster-controller-manager/main.go
SimonTheLeg: ah okay understood.

pkg/applications/installer.go
vgramer: @embik 
I have added an empty dir with a default size is 300Mi.

To choose this default value, I take a look at the size of 3 repositories:
* https://github.com/bitnami/charts: set of ~100 charts provided by bitnami
* https://github.com/prometheus-community/helm-charts: contains the different charts for deploying prometheus stack
* https://github.com/kubermatic/kubermatic

I choose these repositories because the first 2 serve only charts and the last one serves charts and code (with history).

|   repository         | clone all repo   | clone a tag  |
|-----------------|----------------|-------------|
| bitanmi               | 199M                 | 38M             |
| prometheus       | 87M                   | 6,7M             |
| kubermatic.       | 191M                  | 50M             |

So a 300Mi cache should be more than enough to download sources from a repo and handle dependencies.

_note: the max reconciliation of the application controller is 1._

embik: Great, thanks for adding that!

hack/run-user-cluster-controller-manager.sh
```diff
@@ -108,4 +108,5 @@ set -x
   -seed-kubeconfig=${SEED_KUBECONFIG} \
   -owner-email=${OWNER_EMAIL} \
   -dns-cluster-ip=10.240.16.10 \
+  -application-cache="${TMPDIR}" \
```
ahmedwaleedmalik: Wouldn't a dedicated temp directory for this make more sense for this? For reference https://github.com/kubermatic/kubermatic/blob/master/hack/lib.sh#L198

pkg/apis/apps.kubermatic/v1/application_definition.go
```diff
@@ -106,7 +106,9 @@ type GitSource struct {
 	// +kubebuilder:validation:MinLength=1
 	Remote string `json:"remote"`
 
-	// Git reference to checkout
+	// Git reference to checkout.
+	//
+	// For large repositories, we recommend to either use Tag, Branch or Branch+Commit. This allows a shallow clone, which dramatically speeds up performance"
```
ahmedwaleedmalik:
```suggestion
	// For large repositories, we recommend to either use Tag, Branch or Branch+Commit. This allows a shallow clone, which dramatically speeds up performance
```

pkg/applications/installer.go
```diff
@@ -42,14 +44,42 @@ type ApplicationInstaller interface {
 
 // ApplicationManager handles the installation / uninstallation of an Application on the user-cluster.
 type ApplicationManager struct {
+	// ApplicationCache is the path to the directory used for chaching applications. (i.e. locaction where application's source will be downloaded, Helm repository cache ...)
```
ahmedwaleedmalik: `s/chaching/caching`
`s/locaction/location`

pkg/applications/installer.go
```diff
@@ -42,14 +44,42 @@ type ApplicationInstaller interface {
...
 
 // Apply creates the namespace where the application will be installed (if necessary) and installs the application.
 func (a *ApplicationManager) Apply(ctx context.Context, log *zap.SugaredLogger, seedClient ctrlruntimeclient.Client, userClient ctrlruntimeclient.Client, applicationInstallation *appskubermaticv1.ApplicationInstallation) error {
+	// initialize tools
+	sourceProvider, err := providers.NewSourceProvider(ctx, seedClient, &applicationInstallation.Status.ApplicationVersion.Template.Source, a.SecretNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to initialize source provider: %w", err)
+	}
+
+	downloadDest, err := os.MkdirTemp(a.ApplicationCache, applicationInstallation.Namespace+"-"+applicationInstallation.Namespace)
```
ahmedwaleedmalik:
```suggestion
	downloadDest, err := os.MkdirTemp(a.ApplicationCache, applicationInstallation.Namespace+"-"+applicationInstallation.Name)
```

pkg/applications/providers/source/git.go
```diff
@@ -0,0 +1,217 @@
...
+			Auth:         auth,
+			Progress:     nil,
+			SingleBranch: false,
+			Tags:         git.NoTags,
+			NoCheckout:   true,
+		})
+		if err != nil {
+			return err
+		}
+	}
```
ahmedwaleedmalik: Just a nit; you can skip this if you want to.

I think for code coherence, we can resort to something like this:

```go
	cloneOpts := git.CloneOptions{
		URL:          gitSource.Remote,
		Auth:         auth,
		Progress:     nil,
		Tags:         git.NoTags,
		NoCheckout:   true,
	}

	if len(gitSource.Ref.Branch) > 0 {
		// We need to checkout a commit form specific branch
		cloneOpts.SingleBranch = true
		cloneOpts.ReferenceName = plumbing.NewBranchReferenceName(gitSource.Ref.Branch),
	} else {
		// We need to checkout a commit from head
		cloneOpts.SingleBranch = false
	}

	repo, err = git.PlainCloneContext(ctx, destination, false, cloncloneOpts)
	if err != nil {
		return err
	}
```

ahmedwaleedmalik: Also, can't we use `SingleBranch` if branch is not specified? Because in that case `HEAD` will be used.

pkg/applications/installer.go
vgramer: oh shit :s  nice catch!

pkg/applications/providers/source/git.go
vgramer: > I think for code coherence, we can resort to something like this:

agree with you :)

> Also, can't we use SingleBranch if branch is not specified? Because in that case HEAD will be used.

No, because if the user specifies a commit we should look for it in all history. As mentioned earlier it's not recommended to clone only with commit but it can be useful in some cases.

ahmedwaleedmalik: > we should look for it in all history. 

By that, you mean all branches or just the default branch?

vgramer: all branches

ahmedwaleedmalik: That's not right then. If user only specifies the `commit hash` then we should assume that it's from the master/main branch.

vgramer: why should we assume that?

ahmedwaleedmalik (approved):
/approve 🚀 

Let's get back and resolve(if required) https://github.com/kubermatic/kubermatic/pull/9977#discussion_r915762004 based on the consensus in `sig-cluster-management`, in a follow-up PR. It's better to at least get these parts merged.

pkg/applications/providers/source/git.go
ahmedwaleedmalik: As discussed, we'll create a follow up ticket to discuss this.

PR #9993, opened 2022-06-07
update cluster mutation webhook

sankalp-r:
This PR introduces skipping applying default to cluster object in case of cluster deletion.

Current issue:
For the KubeVirt Cloud provider, we want to apply`DefaultCloudSpec` to the cluster object and for this, we first fetch the secret `credential-kubevirt-xxx` and then update the spec.
We noticed during cluster deletion when finalizers get removed, `DefaultCloudSpec` gets called from the cluster mutation webhook which again tries to fetch secret `credential-kubevirt-xxx`. As the secret is already deleted, we end up in error state, which then blocks cluster deletion.
To avoid this issue we want to skip applying defaults to cluster during cluster deletion.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/cluster/mutation/mutation.go
+++ b/pkg/webhook/cluster/mutation/mutation.go
@@ -110,16 +110,18 @@ func (h *AdmissionHandler) Handle(ctx context.Context, req webhook.AdmissionRequ
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 
-		// apply defaults to the existing clusters
-		err := h.applyDefaults(ctx, cluster)
-		if err != nil {
-			h.log.Info("cluster mutation failed", "error", err)
-			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("cluster mutation request %s failed: %w", req.UID, err))
-		}
+		if cluster.DeletionTimestamp == nil {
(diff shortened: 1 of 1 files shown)


pkg/webhook/cluster/mutation/mutation.go
```diff
@@ -110,16 +110,19 @@ func (h *AdmissionHandler) Handle(ctx context.Context, req webhook.AdmissionRequ
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 
-		// apply defaults to the existing clusters
-		err := h.applyDefaults(ctx, cluster)
-		if err != nil {
-			h.log.Info("cluster mutation failed", "error", err)
-			return webhook.Errored(http.StatusInternalServerError, fmt.Errorf("cluster mutation request %s failed: %w", req.UID, err))
-		}
+		// skip applying defaults in case of cluster deletion
```
mfranczy: i think the comment is not needed, it's self explanatory

mfranczy:
@sankalp-r please add more descriptive commit message about the change.

xrstf (approved):
F me sideways, a PR with an actual description and background info. I love it. @mfranczy thanks for asking for it.

/approve

PR #9996, opened 2022-06-07
Fix: DNAT controller iptables-restore call blocking for big user clusters

### LINKED ISSUE #9995: DNAT fails to install iptables rules in big clusters
### What happened?

The DNAT controller creates a set of iptables rules and then applies them calling `iptables-restore` with the rules being fed into stdin.
In big clusters, the command is never executed, thus the stdin pipe is never read from and the function call blocks indefinitely.

### Expected behavior

iptables rules are applied.

1. Create a user cluster with ~100 nodes (using a "long" node CIDR probably helps, too)
2. Observe dnat controller apiserver or metrics-server pod

- KKP version: 2.18.5

not relevant.

OpenStack, but should be irrelevant.

not relevant, as the problem is in seed cluster

### Additional information

If the rules are applied can be checked like so:

```
for pod in $(kubectl -n cluster-7zfjxr86nm get po -l app=metrics-server -o name); do
        kubectl -n cluster-7zfjxr86nm exec -it $pod -c dnat-controller -- iptables-save -t nat
done
```

squ94wk:
This makes the write to stdin asynchronous, so that the `iptables-restore` process always runs (and reads from it).

If the size of the input to `iptables-restore` doesn't exceed the linux stdin buffer size, then everything is fine, as the write call can succeed.
If it's bigger, then `io.WriteString` blocks indefinitely, which means, `cmd` is never run and the function blocks.

This leads to the apiserver -> Kubelet communication to be broken, and thus `kubectl exec|port-forward|logs`  and metrics-server not working.

Fixes #9995

**Documentation**:

```release-notes
Fix: DNAT controller doesn't install NAT rules for big clusters
```

### DIFF
--- a/pkg/controller/kubeletdnat-controller/controller.go
+++ b/pkg/controller/kubeletdnat-controller/controller.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/go-test/deep"
 	"go.uber.org/zap"
@@ -263,7 +264,10 @@ func (r *Reconciler) applyDNATRules(rules []string, haveJump, haveMasquerade boo
 	restore = append(restore, rules...)
 	restore = append(restore, "COMMIT")
 
-	return execRestore(restore)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	return execRestore(ctx, restore)
 }
 
 func execSave() ([]string, error) {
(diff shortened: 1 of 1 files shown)


embik:
Hey @squ94wk. thanks for the contribution and the DCO sign-off changes. This PR includes the unrelated commit for etcd PDB changes, can you perhaps make sure it only includes the commit for the DNAT controller changes? Thanks!

pkg/controller/kubeletdnat-controller/controller.go
```diff
@@ -275,28 +279,30 @@ func execSave() ([]string, error) {
...
 
 	out, err := cmd.CombinedOutput()
-	if err == nil {
-		return nil
-	}
-	if len(out) > 0 {
-		return fmt.Errorf("iptables-restore failed: %w (output: %s)", err, string(out))
+	if err != nil {
+		if len(out) > 0 {
+			return fmt.Errorf("iptables-restore failed: %v (output: %s)", err, string(out))
```
embik, line 300: Please use `%w` instead of `%v` to wrap errors (this is why [`pre-kubermatic-lint`](https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_kubermatic/9996/pre-kubermatic-lint/1537054311412928512) failed).

embik (changes requested):

pkg/controller/kubeletdnat-controller/controller.go
embik: Ping @squ94wk, any update on this PR?

xrstf:
Your work has been copied into a new PR and I tried to keep you as the rightful author of the changes.

Thanks! :)

PR #10007, opened 2022-06-07
Add validation webhook for ClusterTemplates

### LINKED ISSUE #9914: ClusterTemplate validation webhook

We should validate `ClusterTemplate` resources via an admission webhook. Currently, the only validation is happening in the KKP API handler, which can be circumvented (and is supposed to be in GitOps) by applying resources directly to the Kubernetes API.

### Solution details

- A `ValidatingWebhookConfiguration` targets `ClusterTemplates`
- Invalid `ClusterTemplates` (for example with versions not supported by the current configuration) are rejected at the Kubernetes API level

### Use cases

GitOps: Creating valid `ClusterTemplates` directly via `kubectl create` (or GitOps tools) instead of using the KKP API.

### Additional information

Validation logic happening in KKP API: https://github.com/kubermatic/kubermatic/blob/246e6851c05819c335e6c1e0634c78a82a631947/pkg/handler/v2/cluster_template/cluster_template.go#L92

embik:
Adds a validation webhook for `ClusterTemplates`. Another step towards operating KKP from the Kubernetes API directly. Previously, the Kubermatic API was doing the only validation and as such, invalid `ClusterTemplates` could easily be created via kubectl.

Fixes #9914

**Documentation**:

```release-note
Add a validation webhook for `ClusterTemplate` CRs
```

### DIFF
--- a/cmd/kubermatic-webhook/main.go
+++ b/cmd/kubermatic-webhook/main.go
@@ -35,6 +35,7 @@ import (
 	applicationdefinitionvalidation "k8c.io/kubermatic/v2/pkg/webhook/application/applicationdefinition/validation"
 	clustermutation "k8c.io/kubermatic/v2/pkg/webhook/cluster/mutation"
 	clustervalidation "k8c.io/kubermatic/v2/pkg/webhook/cluster/validation"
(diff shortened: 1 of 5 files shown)


xrstf (approved):
/approve

PR #10008, opened 2022-06-07
fix deprecated nodePortProxy annotations being ignored

xrstf:
Since #9476 we keep track of the annotations in a different field (`spec.nodePortProxy.envoy.loadBalancerService.annotations`) and were supposed to use the old settings _and_ the new settings.

The code for the "actual" nodeportproxy (`pkg/controller/operator/seed/resources/nodeportproxy`) was already doing this correctly, but the code for the LoadBalancer exposeStrategy (which confusingly lives in `pkg/resources/nodeportproxy`) forgot to take the deprecated annotations into account. In effect, whatever is configured in `seed.spec.nodePortProxy.annotations` was ignored.

This PR fixes that oversight.

```release-note
fix deprecated nodePortProxy annotations (in `spec.nodePortProxy.annotations` in a Seed object) being ignored
```

### DIFF
--- a/pkg/resources/nodeportproxy/nodeportproxy.go
+++ b/pkg/resources/nodeportproxy/nodeportproxy.go
@@ -395,15 +395,26 @@ func FrontLoadBalancerServiceCreator(data *resources.TemplateData) reconciling.N
 				}
 			}
 
+			seed := data.Seed()
+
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

PR #10019, opened 2022-06-08
Generate valid cluster names where a random name is generated

embik:
#9749 introduced more strict checking of cluster names (only rfc1035 is allowed now, no names starting with numbers are forbidden). Unfortunately, the PR missed a couple of places that generate random cluster names, so this is a follow-up to make sure that for example clusters generated from cluster templates are always valid.

Fixes #

**Documentation**:

```release-note
NONE
```


PratikDeoghare (approved):

PratikDeoghare (approved):

embik:
@PratikDeoghare needs another lgtm, sorry.

PR #10031, opened 2022-06-09
Add ImagePullSecret to Cluster

### LINKED ISSUE #9925: Add imagePullSecrets to Cluster Resource
As described in the proposal [Container Registry Credentials per User Cluster](https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/container-registry-credentials-per-usercluster.md) we want to add a field `imagePullSecret` to the `Cluster` resource.

This secret is then passed to the machine-controller, which sets global registry credentials on node level.

dermorz:
This adds a new field `ImagePullSecrets` to the `Cluster` CRD. The secret referenced by this  `corev1.SecretReference` is passed to a user cluster's machine-controller as a flag (`-node-registry-credentials-secret`).

Fixes #9925

For context see [Proposal: Container Registry Credentials per User Cluster](https://github.com/kubermatic/kubermatic/blob/c8f0695f4c2b2c443281b4b22a83c7f550d3da9f/docs/proposals/container-registry-credentials-per-usercluster.md)

**Documentation**:
Will follow with the extension of KubermaticConfiguration

```release-note
It's now possible to reference a secret with container registry credentials on Cluster resources. These credentials are implicitly available on every node of the cluster.
```

### DIFF
--- a/pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
+++ b/pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
@@ -1945,6 +1945,19 @@ spec:
                 description: HumanReadableName is the cluster name provided by the
                   user.
                 type: string
+              imagePullSecret:
+                description: 'Optional: ImagePullSecret references a secret containing
+                  container registry credentials that are set on node level.'
+                properties:
+                  name:
+                    description: Name is unique within a namespace to reference a
+                      secret resource.
+                    type: string
+                  namespace:
+                    description: Namespace defines the space within which the secret
+                      name must be unique.
+                    type: string
+                type: object
               kubernetesDashboard:
                 description: KubernetesDashboard holds the configuration for the kubernetes-dashboard
                   component.
--- a/pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
+++ b/pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
@@ -1946,8 +1946,9 @@ spec:
                   user.
                 type: string
               imagePullSecret:
-                description: 'Optional: ImagePullSecret references a secret containing
-                  container registry credentials that are set on node level.'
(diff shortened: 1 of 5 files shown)


pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
```diff
@@ -1945,6 +1945,19 @@ spec:
...
+                description: 'Optional: ImagePullSecret references a secret containing
+                  container registry credentials that are set on node level.'
+                properties:
+                  name:
+                    description: Name is unique within a namespace to reference a
+                      secret resource.
+                    type: string
+                  namespace:
+                    description: Namespace defines the space within which the secret
+                      name must be unique.
```
SimonTheLeg, line 1959: I find this sentence a bit confusing. I mean the k8s api enforces that the name of the secret is unique. What I mean is that you cannot just create another secret with the same name in the same namespace. 
In addition I would want to stress in the description that this is the secret the controller looks for registry credentials

SimonTheLeg (commented):
only minor nitpick from my side. Not adding approved right now, so it does not auto-merge

pkg/crd/k8c.io/kubermatic.k8c.io_clusters.yaml
dermorz: I will extend the description about how the referenced secret is used by the machine-controller.

Regarding the verbose explanation of "Namespace": I did not write this, it was generated because I used [`corev1.SecretReference`](https://github.com/kubernetes/api/blob/832b1f4dc5da5a95b836f56045ac2ee7590e950e/core/v1/types.go#L925-L935). I think I would need to copy/shadow this type to change these descriptions. But is it worth it?

SimonTheLeg: okay no. Then it's fine :)

SimonTheLeg (approved):

PR #10052, opened 2022-06-10
Update CA Bundle to 2022-04-26

xrstf:
This PR updates our CA bundle and fixes a few remaining places where we still used an older Go version.

```release-note
The default CA bundle (provided by Mozilla) was updated from 2021-04-13 to 2022-04-26.
```


ahmedwaleedmalik (approved):
/approve

PR #10053, opened 2022-06-10
Update to Alpine 3.16

xrstf:
Alpine 3.16 was released just a few weeks ago and I think it marks a nice time to update our images.

```release-note
All Alpine-based Docker images have been updated from 3.13 to 3.16.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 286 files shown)


xrstf:
This would probably require at least Docker 20.10 ( https://alpinelinux.org/posts/Alpine-3.14.0-released.html ), which is not the default version that the machine-controller installs.

PR #10066, opened 2022-06-13
[release/v2.20] vSphere/Openstack External Snapshotter

mate4st:

Backport of external Snapshotter for vSphere and Openstack.

```release-note
Add External Snapshotter for Openstack and vSphere CSI
```

### DIFF
--- a/addons/csi/vsphere/snapshot-controller.yaml
+++ b/addons/csi/vsphere/snapshot-controller.yaml
@@ -24,10 +24,9 @@
 
 {{ if .Cluster.Features.Has "externalCloudProvider" }}
 {{ if eq .Cluster.CloudProviderName "vsphere" }}
-
-{{ $controllerVersion = "v4.0.0" }}
+{{ $controllerVersion = "v4.0.1" }}
 {{ if eq .Cluster.MajorMinorVersion "1.23" }}
-{{ $controllerVersion = "v5.0.0" }}
+{{ $controllerVersion = "v5.0.1" }}
 {{ end }}
 
 
--- a/addons/csi/vsphere/snapshot-controller.yaml
+++ b/addons/csi/vsphere/snapshot-controller.yaml
@@ -1,0 +1,154 @@
+# Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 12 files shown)


addons/csi/vsphere/snapshot-controller.yaml
```diff
@@ -0,0 +1,154 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sourced from:
+# - https://github.com/kubernetes-csi/external-snapshotter/blob/release-4.1/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+# - https://github.com/kubernetes-csi/external-snapshotter/blob/release-4.1/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+# Modifications:
+#   - Set image tag to 5.0.1 to match the release version
```
embik: This doesn't seem to be accurate, the image tag is `v4.0.0` in this file?

addons/csi/vsphere/vsphere-csi-driver.yaml
```diff
@@ -22,6 +22,9 @@
 #   - remove node selector for operator
 #   - set "improved-volume-topology": "false"
 #   - set image versions and features based on selected k8s version
+#   - set "block-volume-snapshot": "true"
```
embik: Is this really set? I don't see it, but maybe it's a default and block-volume-snapshot=false was removed.

embik (changes requested):
A few nitpicks for modification comments.

Plus, can we please link the original PRs that this one cherry-picks in the description?

addons/csi/vsphere/snapshot-controller.yaml
mate4st: Yes, I templated the version according to the csi driver version

addons/csi/vsphere/snapshot-controller.yaml
```diff
@@ -0,0 +1,168 @@
...
+#   - change replica amount to 1
+#   - template controller version
+
+
+{{ if .Cluster.Features.Has "externalCloudProvider" }}
+{{ if eq .Cluster.CloudProviderName "vsphere" }}
+
+{{ $controllerVersion = "v4.0.0" }}
+{{ if eq .Cluster.MajorMinorVersion "1.23" }}
+{{ $controllerVersion = "v5.0.0" }}
```
embik: Should remove the empty line. Can we also use the latest patch releases `v4.0.1` and `v5.0.1` instead of the base releases (for the webhook, the v5 version is even `v5.0.1`, IMHO those should match)?

embik (changes requested):

addons/csi/vsphere/snapshot-controller.yaml
mate4st: The base version is used here by the vmware csi upsream repo thats why I left it

mate4st: But yes it makes sense to align the version

embik (approved):
/approve

PR #10069, opened 2022-06-13
Add junit output for tests

### LINKED ISSUE #10065: Produce junit test results from go test runs in e2e
Currently, our e2e jobs product junit results, but only for setup steps. We should also include results of our actual test runs in junit artifacts. Let's start with converting `go test` results into junit files.

## Acceptance Criteria

- e2e job steps that run `go test` write junit artifacts that include test results in a detailed way

embik:
This PR attempts to capture as many test results in junit files as possible. This is mostly done through `go-junit-report`, which can process `go test -v` output and produces junit files we write to artifacts. conformance tests already produced junit files but needed to be copied to the proper location, apparently.

Prow jobs are decorated with junit results, so this makes it easier to find the failed test results and makes Prow jobs sortable by failed tests, I think.

Fixes #10065

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow/api.yaml
+++ b/.prow/api.yaml
@@ -32,7 +32,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-        - image: quay.io/kubermatic/build:go-1.18-node-16-kind-0.14-7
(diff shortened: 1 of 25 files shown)


xrstf (approved):
/approve

PR #10073, opened 2022-06-13
Fix azure e2e size test

lsviben:
Tries to fix the azure e2e test:
TestAzureSizesWithCredentials
which got really flaky recently. It seems that the problem is that getting the Azure sizes sometimes lasts longer then 10s which is the default timeout. 

```release-note
NONE
```

### DIFF
--- a/pkg/test/e2e/utils/http.go
+++ b/pkg/test/e2e/utils/http.go
@@ -32,7 +32,7 @@ import (
 type Backoff wait.Backoff
 
 const (
-	apiRequestTimeout = 10 * time.Second
+	apiRequestTimeout = 30 * time.Second
 )
 
 type requestParameterHolder interface {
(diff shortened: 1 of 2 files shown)


pkg/test/e2e/utils/http.go
```diff
@@ -32,7 +32,7 @@ import (
 type Backoff wait.Backoff
 
 const (
-	apiRequestTimeout = 10 * time.Second
+	apiRequestTimeout = 30 * time.Second
```
xrstf, line 36: Please leave a comment on this, mentioning Azure.

lsviben: sure,

I can also revert to my first commit where I treat the azure test differently:
https://github.com/kubermatic/kubermatic/pull/10073/commits/180a6bae863dc3ca9a20aa063da942afdc7155f5

but I thought this is simpler. WDYT?

xrstf (approved):
/approve

PR #10074, opened 2022-06-13
Automatically create Kubermatic namespace on seed clusters

xrstf:
Up until now, we have always insisted that admins create the `kubermatic` namespace on seed clusters manually. All controllers in KKP would refuse to reconcile if that namespace doesn't exist yet.

Now with the improved CRD handling for seeds, we want to automate their setup even more and so we have decided to drop this requirement (which was, IIRC, only meant as a stop if someone accidentally adds a cluster that should not be a seed) and create the namespace if we need to.

This PR implements this change. The setup flow for new seeds is a bit complex, but on purpose:

* The _Operator_ is always responsible for setting up KKP components (like controller-managers, secrets etc.). But it must be possible to run KKP without the operator being present (i.e. scaling it to 0 replicas should have no effect if an admin doesn't want to change anything in KKP).
* Synchronizing Seed resources into their seed clusters is something that always needs to happen, so my idea to merge the seed-sync controller into the operator died, as the operator is "optional".

So when you add a new Seed,

* the seed-sync controller will create the namespace on the seed,
* the seed-sync controller will copy the Seed resources into the seed cluster
* at this point the operator can find it there and begin to reconcile the seed-ctrl-mgr, webhooks etc.

And since the operator cannot do anything on the seed before the seed-sync controller has done its job, there is no point in having namespace-reconciling-code in the operator. This PR therefore removes the effectively dead code.

```release-note
KKP now creates the `kubermatic` namespace on seed clusters automatically, so admins need to do one less step when setting up new seed clusters.
```

### DIFF
--- a/pkg/controller/master-controller-manager/seed-sync/reconciler.go
+++ b/pkg/controller/master-controller-manager/seed-sync/reconciler.go
@@ -126,12 +126,6 @@ func (r *Reconciler) reconcile(ctx context.Context, config *kubermaticv1.Kuberma
 		return fmt.Errorf("failed to add finalizer: %w", err)
 	}
 
-	if seed.Spec.ExposeStrategy != "" {
-		if !kubermaticv1.AllExposeStrategies.Has(seed.Spec.ExposeStrategy) {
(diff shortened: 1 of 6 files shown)


moadqassem (approved):
/approve
/lgtm

xrstf:

This is endlessly flaking now for some reason, across all PRs.

ahmedwaleedmalik (approved):
👍

PR #10079, opened 2022-06-14
Resource Quota API endpoints

### LINKED ISSUE #9118: ResourceQuotas API endpoints
Implement API endpoints for project resource quotas:

- Add CRUD endpoints
- admin only

pkprzekwas:
- introduce API endpoints to interact with KKP's resource quotas

Fixes #9118 

I was not sure if resource quota provider should be under EE license and if so where should it be located. I went for keeping it EE in `/pkg/ee/resource-quotas`. An alternative would be `/pkg/ee/provider` where we currently store only `SeedGetterFactory` and `SeedKubeconfigGetter`.

```release-note
Introduce new API endpoints for CRUD operations on KKP Resource Quotas.
```


pkg/apis/kubermatic/v1/register.go
```diff
@@ -91,6 +91,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&ClusterTemplateInstanceList{},
 		&RuleGroup{},
 		&RuleGroupList{},
+		&ResourceQuota{},
```
lsviben: i guess this is a leftover

pkprzekwas: good catch, thanks

pkg/ee/resource-quotas/handler.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, common.KubernetesErrorToHTTPError(err)
+	}
+
+	return &apiv1.ResourceQuota{
+		Name:        projectResourceQuota.Name,
+		SubjectKind: projectResourceQuota.Spec.Subject.Kind,
+		SubjectName: projectResourceQuota.Spec.Subject.Name,
+		Quota:       projectResourceQuota.Spec.Quota,
+		Status: kubermaticv1.ResourceQuotaStatus{
+			LocalUsage: projectResourceQuota.Status.LocalUsage,
```
lsviben: Global usage would make more sense to send back, as its the global resource usage of the project. The local usage is just the usage on a particular seed, which could be nothing if the master is not a master/seed setup

pkg/api/v1/types.go
```diff
@@ -2673,6 +2673,15 @@ type MeteringReportConfiguration struct {
 // swagger:model MeteringReportURL
 type ReportURL string
 
+// swagger:model ResourceQuota
+type ResourceQuota struct {
+	Name        string                           `json:"name"`
+	SubjectName string                           `json:"subject_name"`
+	SubjectKind string                           `json:"subject_kind"`
```
lsviben: why not use the `kubermaticv1.Subject`?

pkg/ee/resource-quotas/provider.go
```diff
@@ -0,0 +1,154 @@
...
+		Groups:   []string{userInfo.Group},
+	}
+	masterImpersonatedClient, err := p.createMasterImpersonatedClient(impersonationCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	resourceQuota := &kubermaticv1.ResourceQuota{}
+	if err := masterImpersonatedClient.Get(ctx, types.NamespacedName{
+		Name:      fmt.Sprintf("%s-%s", kind, name),
```
lsviben: i suggest making a helper func for generating the name

pkg/ee/resource-quotas/provider.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (p *ResourceQuotaProvider) CreateUnsecured(ctx context.Context, subject kubermaticv1.Subject, quota kubermaticv1.ResourceDetails) error {
+	rq := &kubermaticv1.ResourceQuota{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{},
+			// Add labels for easier filtering
+			Labels: map[string]string{
+				kubermaticv1.ResourceQuotaSubjectNameLabelKey: subject.Name,
+				kubermaticv1.ResourceQuotaSubjectKindLabelKey: subject.Kind,
+			},
```
lsviben: I prefer if we set this in a mutating webhook so it will be constant if created from API or kubectl. ( we can remove this when we add it in a mut webhook)

pkg/handler/v1/admin/resource_quotas.go
```diff
@@ -0,0 +1,121 @@
...
+
+		err = createResourceQuota(ctx, req, provider)
+		if err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+}
+
+func UpdateResourceQuotaEndpoint(userInfoGetter provider.UserInfoGetter, provider provider.ResourceQuotaProvider) endpoint.Endpoint {
```
lsviben: I think it would be better if we make a PATCH instead of an UPDATE endpoint. As we have a lot of controller and things that touch KKP resources, it makes sense to make the gentlest changes as possible.

pkg/apis/kubermatic/v1/resource_quota.go
```diff
@@ -21,6 +21,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ResourceQuotaNamespace           = "kubermatic"
```
lsviben: you can just use `resources.KubermaticNamespace`

pkg/handler/v1/admin/resource_quotas.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/go-kit/kit/endpoint"
+
+	"k8c.io/kubermatic/v2/pkg/provider"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func GetResourceQuotaEndpoint(userInfoGetter provider.UserInfoGetter, provider provider.ResourceQuotaProvider) endpoint.Endpoint {
```
lsviben: I suggest you put all the endpoints into v2, no point to put new stuff to v1. 

In v2 we dont have a division into admin/other, so its simpler as well

pkg/api/v1/types.go
pkprzekwas: Unfortunately, it conflicts with https://github.com/kubermatic/kubermatic/blob/master/pkg/api/v1/types.go#L2513 and swagger is generating corrupted client because of that.

pkg/ee/resource-quotas/handler.go
pkprzekwas: My mistake, I thought that maybe GlobalUsage is something that we don't want to present to non-admin users. But your comment clarifies this case.

pkg/ee/resource-quotas/provider.go
pkprzekwas: Great idea.

PR #10082, opened 2022-06-14
update promtail to 2.5.0

xrstf:
After #9648 broke our promtail chart and #9693 brought it back to life, this PR now properly updates the chart after upstream has fixed its issues.

Note that a single "client" configuration has been deprecated in promtail since forever (https://github.com/grafana/loki/pull/536), so this PR now moves to the new "clients" configuration, as does the upstream chart. This is a breaking change.

```release-note
ACTION REQUIRED: Update Promtail to v2.5.0; the `config.client` configuration has been replaced by `config.clients` instead. If you overwrote the Promtail client (=Loki), please adjust your `values.yaml` accordingly.
```

### DIFF
--- a/charts/logging/promtail/Chart.yaml
+++ b/charts/logging/promtail/Chart.yaml
@@ -15,7 +15,7 @@
 apiVersion: v2
 name: promtail
 version: v9.9.9-dev
-appVersion: v2.4.2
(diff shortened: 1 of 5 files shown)


ahmedwaleedmalik (approved):
/approve 🚀

PR #10088, opened 2022-06-14
WIP - Replace ObjectReference with a dedicated SeedKubeconfigReference

xrstf:
`ObjectReference` objects are deprecated and offer waaay more fields than we actually need, plus some that we do not even support. Instead of documenting lots of corner cases ("we don't support this and that and that, and this is defaulted to that"), this PR introduces a dedicated SeedKubeconfigReference.

```release-note
Seed's now load their kubeconfig always from the same namespace.
```


xrstf:
For this to work, we would need to first check the existing Seeds and _then_ update the CRDs. I don't think that is worth it and how to communicate this to admins, so this PR might be a bit too ambitious.

PR #10098, opened 2022-06-15
Fix IPAM/dual-stack with Azure external CCM

rastislavs:
Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>

In https://github.com/kubermatic/kubermatic/pull/9963, Azure external CCM support has been added. This is a follow-up PR to fix some networking aspects:

 - Since IPAM in Azure CCM [is not recommended for the production](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/site/content/en/topics/ipam.md), and we would not need its Azure-specific benefits ATM anyway, we should rather keep using the IPAM in kube-controller-manager (which also takes proper care of dual-stack IPAM).
 - NOTE: If we wanted wanted to use IPAM in Azure CCM in the future, we should disable it in kube-controller-manager (`--allocate-node-cidrs=false`), otherwise both are enabled and competing.
 - We don't need CCM to configure cloud routes for pod CIDRs, as ATM we use VXLAN overlay for pod network in all KKP-managed user clusters. That may change in the future if we do some cloud-provider-specific networking optimisations, but it would require some more changes.

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/resources/cloudcontroller/azure.go
+++ b/pkg/resources/cloudcontroller/azure.go
@@ -164,12 +164,11 @@ func getAzureVersion(version semver.Semver) (string, error) {
 }
 
 func getAzureFlags(data *resources.TemplateData) []string {
-	clusterCIDR := data.Cluster().Spec.ClusterNetwork.Pods.GetIPv4CIDR()
 	flags := []string{
-		// "false" for Azure CNI and "true" for other network plugins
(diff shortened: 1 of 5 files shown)


embik (approved):
/approve

looks good to me, assuming CI succeeds.

embik (approved):

PR #10102, opened 2022-06-15
Add endpoint for OIDC kubeconfig secret for WEB terminal

zreigz:
**What does this PR do / Why do we need it**: add an endpoint for OIDC kubeconfig secret for the WEB terminal.

Fixes #

In order to test this locally:
1. Alter our dex configuration (`kubermaticIssuer` client) to whitelist the `redirectURI`: `http://localhost:8080/api/v2/kubeconfig/secret` and `https://kubermatic.io/api/v2/kubeconfig/secret`
2.
  - `kubectl -n oauth edit configmap dex`
  - `kubectl -n oauth rollout restart deployment dex`
image
3. Create and wait for any cluster to start (using our dev env).
4. Run the API locally, i.e. `make clean && FEATURE_GATES=OIDCKubeCfgEndpoint=true KUBERMATIC_EDITION=ee KUBERMATIC_DEBUG=false ./hack/run-api.sh`
5. 
  Open WEB browser and enter: `http://localhost:8080/api/v2/kubeconfig/secret?cluster_id=<clusterID>&project_id=<projectID>&user_id=<userID>`. Make sure to replace `clusterID`,  `userID` and `projectID` with the values of the created cluster. The userID is the ID for the user object in the KKP.

6. Get user cluster kubeconfig and check if the secret was created in the `kube-system` namespace. The secret name is your email encoded to md5 hash. You can check the name here https://www.md5hashgenerator.com/. For example for user `lukasz.zajaczkowski@kubermatic.com` we will get `40ce6d00769edaab4eb1d8cacfea5cb1`

```release-note
Add an endpoint for OIDC kubeconfig secret for the WEB terminal
```


pkg/handler/common/kubeconfig.go
```diff
@@ -286,6 +294,190 @@ func CreateOIDCKubeconfigEndpoint(ctx context.Context, projectProvider provider.
...
+		Data: secretData,
+	}
+	return client.Create(ctx, secret)
+}
+
+func getRedirectURL(r *http.Request) string {
+	scheme := "http"
+	if r.TLS != nil {
+		scheme = "https"
+	}
```
ahmedwaleedmalik: This doesn't seem to work, at least it doesn't here https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/kubernetes-dashboard/login.go#L152

And the failure seems right because there is no mTLS and the API inside the pod is served over HTTP; `0.0.0.0:8080`.

zreigz: good point. I've changed it. We can rely on the current redirect URI and just change the path of the endpoints

zreigz:
@ahmedwaleedmalik @imharshita PTAL

ahmedwaleedmalik (approved):
/approve ⚡

PR #10110, opened 2022-06-16
try more than 2 commits when finding a gocache

xrstf:
Going back 2 commits was already pretty clever, but in cases where Prow does a batch merge, this might simply not be enough. And if we go back 2 commits, there is no reason not to go back even more. Even a cache that is 10 commits old would be helpful (I guess).

So this PR turns the if-if struct into a loop that tries 5 times to find a gocache. And it also fixes an oversight where the Azure CCM tests where still on Go 1.18.2, so they could never find a cache.

```release-note
NONE
```

### DIFF
--- a/.prow/ccm-migrations.yaml
+++ b/.prow/ccm-migrations.yaml
@@ -94,7 +94,7 @@ presubmits:
               cpu: 3.5
             limits:
               memory: 4Gi
-  
+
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

xrstf:

flaky flaky

PR #10120, opened 2022-06-20
use the new web-terminal Docker image

xrstf:
This makes sure that we actually use the new image for our web terminal. It also removes some tools which I find questionable, like socat or rsync, for a simple web terminal.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡

PR #10123, opened 2022-06-20
Override operating system manager deployment  image repo and tag 

moadqassem:
At some cases, we would like to control which image and tag we would like to deploy operating system manager in a user cluster. While this is not a great practice, due to the conflicts and compatibility problems that might show up. However, sometimes we have some urgent fix that we need to introduce or for PoC purposes that we find it a bit complex to wait until a new release for KKP, where the new OSM version is included. Thus, I added the possibility to override the default version of OSM(the version that is shipped within KKP) to achieve the goals that are mentioned above. 
 
Fixes #

**Documentation**:

```release-note
OSM deployment image repo and tag override. 
```


ahmedwaleedmalik (approved):
/approve ⚡

cmd/seed-controller-manager/options.go
```diff
@@ -145,6 +149,9 @@ func newControllerRunOptions() (controllerRunOptions, error) {
 	flag.StringVar(&c.lokiRulerURL, "loki-ruler-url", "http://loki-distributed-ruler.mla.svc.cluster.local:3100", "The URL of loki ruler which is running for MLA stack.")
 	flag.StringVar(&c.machineControllerImageTag, "machine-controller-image-tag", "", "The Machine Controller image tag.")
 	flag.StringVar(&c.machineControllerImageRepository, "machine-controller-image-repository", "", "The Machine Controller image repository.")
+	flag.StringVar(&c.operatingSystemManagerImageTag, "operating-system-manager-image-tag", "", "The Operating System Manager image tag.")
+	flag.StringVar(&c.operatingSystemManagerImageRepository, "operating-system-manager-image-repository", "", "The Operating System Manager image repository.")
+
```
xrstf: Why do we need these as new CLI flags? All controllers should have access to the KubermaticConfiguration and can react dynamically to it changing. No need to redeploy the seed-ctrl-mgr when something is changed. Or am I overlooking something?

In general, we want to _reduce_ the number of CLI flags over time, not grow them. ^^

pkg/apis/kubermatic/v1/configuration.go
```diff
@@ -270,6 +272,18 @@ type MachineControllerConfiguration struct {
 	ImageTag string `json:"imageTag,omitempty"`
 }
 
+// OperatingSystemManager configures the image repo and the tag version for osm deployment.
+type OperatingSystemManager struct {
+	// ImageRepository is used to override the OperatingSystemManager image repository.
+	// It is recommended to use this field only for development, tests and PoC purposes.For production environments.
```
xrstf: > "purposes.For production environments.it is not recommended, to use"

should probably be

> "purposes. For production environments it is not recommended to use"

pkg/apis/kubermatic/v1/configuration.go
```diff
@@ -270,6 +272,18 @@ type MachineControllerConfiguration struct {
...
+	// it is not recommended, to use this field due to compatibility with the overall KKP stack.
+	ImageRepository string `json:"imageRepository,omitempty"`
```
xrstf, line 280: Everywhere else we call these fields "DockerRepository". Can we do the same here, same for "DockerTag" instead of "ImageTag"?

cmd/seed-controller-manager/options.go
moadqassem: I completely agree with you, let me adjust the code.

pkg/apis/kubermatic/v1/configuration.go
moadqassem: Honestly image repo/tag is the correct term here, not `DockerRepository`

pkg/apis/kubermatic/v1/configuration.go
moadqassem: Done

cmd/seed-controller-manager/options.go
moadqassem: Done

pkg/apis/kubermatic/v1/configuration.go
lsviben: I would also rather start using ImageTag and ImageRepository over Docker tag/repo. It makes more sense even if it costs us some consistency.

xrstf: And are there any reasons why "Image repo is the correct term"? Otherwise I would like to just say "Docker repo is the correct term" :-p

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -230,6 +230,8 @@ func (r *Reconciler) getClusterTemplateData(ctx context.Context, cluster *kuberm
 		WithDnatControllerImage(r.dnatControllerImage).
 		WithMachineControllerImageTag(r.machineControllerImageTag).
 		WithMachineControllerImageRepository(r.machineControllerImageRepository).
+		WithOperatingSystemManagerImageTag(config).
+		WithOperatingSystemManagerImageRepository(config).
```
xrstf: Do we even need those 2 functions? Is there ever a case where we want to override the images so they do not match what is in the KubermaticConfiguration?

moadqassem: Yeah, those are the setter functions for the TemplateData field, which is used to change the values of the osm specs. Otherwise I need to pass the KubermaticConfigurations variable everywhere, so instead I am injecting those overrides via the template data, which is already present in the function call.

xrstf: But we're already setting the config with `WithKubermaticConfiguration()`. If the values we set never diverge from the KubermaticConfiguration, then you can simply implement

```
func (d *TemplateData) OperatingSystemManagerImageTag() string {
	return d.config.Spec.OperatingSystemManager.ImageTag
}

func (d *TemplateData) OperatingSystemManagerImageRepository() string {
	return d.config.Spec.OperatingSystemManager.ImageRepository		
}
```

and don't need 2 extra setters.

moadqassem: Ah ok, I overlooked that, yeah in that case you are right, those are redundant.

pkg/resources/data.go
```diff
@@ -60,26 +60,28 @@ type CABundle interface {
...
+	etcdDiskSize                          resource.Quantity
+	oidcIssuerURL                         string
+	oidcIssuerClientID                    string
+	kubermaticImage                       string
+	etcdLauncherImage                     string
+	dnatControllerImage                   string
+	machineControllerImageTag             string
+	machineControllerImageRepository      string
+	operatingSystemManagerImageTag        string
+	operatingSystemManagerImageRepository string
```
xrstf: Now these 2 fields can be removed as well. :)

moadqassem: Wth, I removed them!  ok will try again

xrstf (approved):
/approve

PR #10124, opened 2022-06-20
Resource quota syncer

### LINKED ISSUE #9120: resource quota syncer
Add a ResourceQuota controller in the master cluster which:
- syncs Resource quotas from the master to seed clusters

lsviben:
Syncs Resource Quotas from the master to seed clusters. It syncs only the Global resource usage(all seeds wide) from its status, so that the local user cluster webhooks can know what is the global resource usage for a given project

Fixes #9120 

```release-note
NONE
```


pkg/ee/resource-quota/resource-quota-syncer/syncer.go
```diff
@@ -0,0 +1,183 @@
...
+
+	c, err := controller.New(ControllerName, masterMgr, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	for seedName, seedManager := range seedManagers {
+		r.seedClients[seedName] = seedManager.GetClient()
```
kron4eg, line 83: what is new seed will be added in runtime?

lsviben: This controller is a part of the seed-controller-lifecycle, so if I understand correctly, every time a new seed is added or removed, all the controllers are restarted

kron4eg (approved):
/approve
/lgtm

PR #10127, opened 2022-06-20
Cleanup expose-strategy tests

xrstf:
Some code maintenance to remove unneeded cruft from the tests.

```release-note
NONE
```


pkg/test/e2e/expose-strategy/tunneling_test.go
```diff
@@ -20,73 +20,102 @@ package exposestrategy
...
+
+	logger.Debug("Testing SNI when Kubeconfig is used e.g. Kubelet")
+	if result := client.QueryApiserverVersion("", false, options.kubernetesVersion, 5, 4); result != true {
+		t.Error("Apiserver should be reachable passing from the SNI entrypoint in nodeport proxy")
+	}
+
+	logger.Debug("Tunneling requests using HTTP/2 CONNECT when no SNI is present e.g. pods relying on kubernetes service in default namespace")
+	if result := client.QueryApiserverVersion(agentConfig.GetKASHostPort(), true, options.kubernetesVersion, 5, 4); result != true {
+		t.Error("Apiserver should be reachable passing from the SNI entrypoint in nodeport proxy")
+	}
```
embik: This might be really nitpicky, but any chance we wrap those to into two `t.Run` calls (or make a simple table-driven test out of this)? That way, we will get separate test results and two entries in junit files (ref #10069), so we immediately see which test failed.

embik (changes requested):

pkg/test/e2e/expose-strategy/tunneling_test.go
xrstf: Good idea!

embik (approved):

PR #10137, opened 2022-06-21
Add a flag -o/--output to kubermatic-installer to allow json output

embik:
Not sure if we want to merge this, but I thought it would be neat to have the option for json log formatting in `kubermatic-installer`. The default of course stays with the nice formatter that logrus uses for human readability (filled under "console" in this new flag), but an `--output` (or shorthand `-o`) flag will allow to get json output if that is wanted (I can see this being useful with the upcoming `mirror-images` subcommand, for example).

This is what it would look like, as an example for an early failure on deploying:

```sh
$ ./_build/kubermatic-installer deploy
INFO[0000] 🚀 Initializing installer…                     edition="Enterprise Edition" version=2a9ece9131272f12e01b6331769fdd301239230d
ERRO[0000] ❌ Operation failed: no kubeconfig (--kubeconfig or $KUBECONFIG) given.

$  ./_build/kubermatic-installer deploy -o json
{"edition":1,"level":"info","msg":"🚀 Initializing installer…","time":"2022-06-21T17:00:30+02:00","version":"2a9ece9131272f12e01b6331769fdd301239230d"}
{"level":"error","msg":"❌ Operation failed: no kubeconfig (--kubeconfig or $KUBECONFIG) given.","time":"2022-06-21T17:00:30+02:00"}
```

WDYT?

Fixes #

**Documentation**:

```release-note
`kubermatic-installer` output can be formatted as json via a new `--output` flag
```

### DIFF
--- a/cmd/kubermatic-installer/main.go
+++ b/cmd/kubermatic-installer/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"math/rand"
 	"os"
 	"time"
@@ -30,6 +31,7 @@ import (
 
 type Options struct {
 	Verbose         bool
+	LogFormat       log.LogrusFormat
 	ChartsDirectory string
 }
 
@@ -63,6 +65,13 @@ func main() {
 			if options.Verbose {
 				logger.SetLevel(logrus.DebugLevel)
 			}
+
+			switch options.LogFormat {
+			case log.LogrusFormatJSON:
+				logger.SetFormatter(&logrus.JSONFormatter{})
+			case "", log.LogrusFormatConsole:
+				logger.SetFormatter(&logrus.TextFormatter{})
+			}
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


embik:
> In all other apps, we use `-log-format`, should we use the same here instead of `-o`? I don't mind `-o`, just wondering.

I was thinking about that, but since we're already doing different logging in the installer, I thought offering something that looks more like other Kubernetes client applications was better. I don't have strong opinions about that though, I can change it if you prefer it to be consistent.

> The log messages do have indentations, so we would end up with something like "msg:" foo". Can you live with that? (I tried sooo hard to somehow automate the indentation somehow, as if it was structured logging, but manually indenting all lines was much easier, hence the many empty spaces).

I don't think that's an issue 🤷🏻

xrstf:
/approve
/lgtm

I wish we had telemetry to tell us how many users discovered and then used the JSON output in the future :D

PR #10143, opened 2022-06-22
Improve etcd backup filenames

### LINKED ISSUE #7337: User Cluster Backup Naming not S3 compliant
**User Story**
Based on the AWS docu, the current used filenames are not compliant with the S3 standard `Safe characters`: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

This leads to errors at e.g. Netapp S3 storage (potential also some others). Issue needs to get fixed:
- legacy backup controller: https://github.com/kubermatic/kubermatic/blob/master/docs/zz_generated.kubermaticConfiguration.yaml#L193
- new etcd Backup and Restore Controller: https://docs.kubermatic.com/kubermatic/v2.17/cheat_sheets/etcd/backup-and-restore/

**Acceptance criteria**
- Alle created Files of KKP Backup Solution are compliant to S3

xrstf:
Not all S3 backends can handle files with colons (:) in them, apparently. Also it's hard to use the files because no file extension is present and you're not even sure if it's a tar.gz of the etcd or not.

This PR addresses both points. Existing backups should be untouched, but new ones should be more compatible.

Fixes #7337

```release-note
etcd backup files are named differently (`foo-YYYY-MM-DDThh:mm:ss` to `foo-YYYY-MM-DDThhmmss.db`) to improve compatibility with different storage solutions
```

### DIFF
--- a/pkg/storeuploader/storeuploader.go
+++ b/pkg/storeuploader/storeuploader.go
@@ -93,14 +93,14 @@ func (u *StoreUploader) Store(ctx context.Context, file, bucket, prefix string, 
 			return err
 		}
 		if !exists {
-			logger.Infow("Creating bucket")
+			logger.Info("Creating bucket")
 			if err := u.client.MakeBucket(ctx, bucket, minio.MakeBucketOptions{}); err != nil {
 				return err
 			}
 		}
 	}
 
-	objectName := fmt.Sprintf("%s-%s-%s-%s", prefix, prefixSeparator, time.Now().Format("2006-01-02T15:04:05"), path.Base(file))
+	objectName := fmt.Sprintf("%s-%s-%s-%s", prefix, prefixSeparator, time.Now().Format("2006-01-02T150405"), path.Base(file))
 	logger.Infow("Uploading file", "src", file, "dst", objectName)
 
 	_, err := u.client.FPutObject(ctx, bucket, objectName, file, minio.PutObjectOptions{})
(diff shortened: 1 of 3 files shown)


pkg/storeuploader/storeuploader.go
```diff
@@ -93,14 +93,14 @@ func (u *StoreUploader) Store(ctx context.Context, file, bucket, prefix string,
...
-			logger.Infow("Creating bucket")
+			logger.Info("Creating bucket")
 			if err := u.client.MakeBucket(ctx, bucket, minio.MakeBucketOptions{}); err != nil {
 				return err
 			}
 		}
 	}
 
-	objectName := fmt.Sprintf("%s-%s-%s-%s", prefix, prefixSeparator, time.Now().Format("2006-01-02T15:04:05"), path.Base(file))
+	objectName := fmt.Sprintf("%s-%s-%s-%s", prefix, prefixSeparator, time.Now().Format("2006-01-02T150405"), path.Base(file))
```
embik, line 103: If we change the object name here, don't we need to change it in the restore produce in `etcd-launcher` as well? I was looking at that but I couldn't find a reference to the generated object name, so I'm a bit confused how the restore works.

embik (commented):
Q

pkg/storeuploader/storeuploader.go
xrstf: The only place I found where the etcdrestore is taking its name from is https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/etcdrestore/etcdrestore.go#L325, so the dashboard is sending us the name of the backup.

screenshot-2022-06-24-141456

The list of backups seems to be created based on the EtcdBackupConfig status, so as long as there are names with .db in the status, they should show up in the dashboard and be transmitted 1:1 to the backend. So ...... we're fine?

embik: After some investigation: The s3-storeuploader is used only in the legacy backup feature, that's why there's no "restore" side to this, it did not exist, restoring was a manual process. It does not affect the current backup/restore controllers.

embik (approved):

PR #10150, opened 2022-06-22
Remove last remnants of the redundant User.ID field

xrstf:
In KKP 2.20 we already determined that the User.ID is not useful and optional. However our webhooks still validated it and so the documentation on the field did not match the behaviour enforced by our webhooks.

This PR cleans this up by removing all mentions of the ID from our codebase, but leaves the ID in the CRD (not sure if one can simply remove a field from a CRD if there are resources out there that have it set?).

While cleaning up the User CRD, I also cleaned up the names of the testing accounts:

* roxy stays roxy
* roxy2 is now roxy-admin
* the @loodse.com variants are now finally gone

**Background**:
Roughly 5 years ago, in #621, we added the `User` CRD to KKP. This very first iteration already included an `ID` field. Sadly there was no further comment on what exactly the ID was meant to represent and why we need it if a Kubernetes object already has a globally unique name.

From what I gathered from the history is in KKP 2.5 we introduced the field and then over the next few releases we started to actually put data in it (i.e. do `user.ID = ....` somewhere), but there was never really any code that actually used the ID for something. It was always somehow updated and maaaany unit tests set it, but it seems to have been always pointless.

Do note that `pkg/api/v1.User`, the type we use in our REST API, also has an `ID` field (because Users embed the ObjectMeta). This ID however is _not_ the same as the ID inside a User resource in Kubernetes. This means `apiUser.ID = k8sUser.Name`.

This has the dramatic upside that we can just get rid of the ID entirely and do not accidentally break our API. All API usages are already based on the name (even though the API calls it the "ID") and so nothing will change for API clients once the actual internal ID is removed.

```release-note
NONE
```


pkg/validation/user.go
```diff
@@ -34,10 +34,6 @@ func ValidateUser(u *kubermaticv1.User) field.ErrorList {
 		allErrs = append(allErrs, field.Required(specPath.Child("email"), "this field is required"))
 	}
 
-	if u.Spec.ID == "" {
```
lsviben, line 37: check also down below in ValidateUserUpdate, ID is used, could you remove that as well?

PR #10158, opened 2022-06-23
Add `GroupProjectBinding` CRD

### LINKED ISSUE #9999: GroupProjectBinding CRD
- Make sure its unique
- Groups cant have the same name as the default groups (viewers, editors…)
- GroupsProjectBindings will need to be also defended by RBACs (similar as Clusters), and the RBACs rules added to the rbac-controller

embik:
This extracts the `GroupProjectBinding` CRD from #9960 and adds it as a standalone PR. That way, we can work on various aspects of the overarching epic (#9897) without relying on a single PR.

I've changed the `.spec.project` field to `.spec.projectID` from the old PR, because that is what the field is named in `UserProjectBindings` and the dashboard also shows this as the ID, not the name:

Screenshot 2022-06-23 at 11 53 37

That's a bit confusing because the Kubernetes API resource `Project` uses this as `.metadata.name`, but I want to keep it consistent with existing fields as much as possible.

Fixes #9999

**Documentation**:

```release-note
Add `GroupProjectBinding` CRD
```


xrstf:
How do we prevent stale group information in User objects?

embik:
> How do we prevent stale group information in User objects?

That's a very good question I don't have an answer to. Let me remove that part from the PR, we should add it when we also add the KKP API functionality.

lsviben:
> > How do we prevent stale group information in User objects?
> 
> That's a very good question I don't have an answer to. Let me remove that part from the PR, we should add it when we also add the KKP API functionality.

The KKP is the only one which will use the Group info in User, and it will be the one which sets it on every login. 

It will make more sense with the whole KKP API changes :)

PR #10161, opened 2022-06-23
Webhook for IPAMPool validation + cleanup logic for IPAMPool deletion

### LINKED ISSUE #9906: Add webhook for IPAMPools
Add a webhook ensuring consistency of IPAMPools:

- IPAMPool cannot be deleted if there are existing IPAMAllocations
- poolCIDR cannot be changed for an existing datacenter
- allocation prefix/range can be changed  for an existing datacenter (that would not affect the existing allocations)
- potentially additional consistency checks for IPAMAllocations

hbernardo:
This PR adds a webhook for IPAMPool validation and implements the cleanup logic for IPAMPool deletion.

Fixes #9906

**Documentation**:

```release-note
NONE
```


pkg/controller/seed-controller-manager/ipam/controller.go
```diff
@@ -89,10 +92,22 @@ func Add(
 	}
 
 	enqueueClustersForIPAMPool := handler.EnqueueRequestsFromMapFunc(func(a ctrlruntimeclient.Object) []reconcile.Request {
+		ctx := context.Background()
+
 		ipamPool := a.(*kubermaticv1.IPAMPool)
 
+		wasIPAMPoolDeleted, err := reconciler.checkIPAMPoolCleanup(ctx, ipamPool)
```
lsviben: I dont think that the IPAMPool cleanup should be happening in the EnqueueRequestsFromMapFunc, even though I understand why its convenient when the reconcile is working with Clusterrs, not IPAMPool.

The cleanup would be simpler if the IPAMPool controller would reconcile the IPAMPool, not sure why its doing Clusters and then listing IPAMPools?

pkg/webhook/ipampool/validation/validation.go
```diff
@@ -0,0 +1,160 @@
...
+		switch dcOldConfig.Type {
+		case kubermaticv1.IPAMPoolAllocationTypeRange:
+			if dcOldConfig.AllocationRange != dcNewConfig.AllocationRange {
+				return errors.New("it's not allowed to update the allocation range for a datacenter")
+			}
+		case kubermaticv1.IPAMPoolAllocationTypePrefix:
+			if dcOldConfig.AllocationPrefix != dcNewConfig.AllocationPrefix {
+				return errors.New("it's not allowed to update the allocation prefix for a datacenter")
+			}
+		}
```
lsviben, line 88: not very familiar with the subject, but I am checking the ticket requirements:
one is:
- allocation prefix/range can be changed for an existing datacenter (that would not affect the existing allocations)

this blocks the change, but I guess that is ok if it affects existing allocations? Is there a case where it doesnt affect them, is it even importnant?

pkg/controller/seed-controller-manager/ipam/controller.go
hbernardo: It's reconciling clusters because an IPAMPool can be applied before a cluster is created. So, in that case, when the cluster is created, the controller will allocate the subnet from the IPAMPool for it automatically.

pkg/webhook/ipampool/validation/validation.go
hbernardo: Discussing with @rastislavs , we decided to change this requirement because it wouldn't be consistent to have allocations that are not compatible with the updated pool. Also, the controller algorithm is checking all allocations for building a map of already allocated "subnets" from the pool and, specially for the prefix type, it would stop working. For the range type, a small change would make it work, but for now we decided to keep it simple and forbid this change (anyway, the user can still delete the pool and apply it again).

lsviben: ok, makes sense

lsviben:
Looks good, you are missing just the depoyment of the Validation webhook itself. Check here if you want to deploy it on the master cluster for example
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/master/resources/kubermatic/webhooks.go

PR #10170, opened 2022-06-24
Use Kube API for Cilium e2e tests

xrstf:
This PR rewrites the Cilium e2e tests to use a kube client instead of our KKP API to setup the project/cluster.

As we move more and more towards working declaratively, our API is becoming "optional". But we cannot (re)move the API if other core e2e tests are still depending on it.

The new code to setup the cluster is a bit longer because our API is oh-so convenient and fills in soooo many fields for the MachineDeployment automatically. I can see why it's tempting to use the API to setup things :grin: I plan to adopt this code together with the existing ClusterJigs from the expose-strategy/nodeport/ccm tests to build a simpler test utility to setup stuff.

**NOTE:** This PR currently contains the changes from #10172 to make the tests more stable. Once the other PR is merged, I will rebase this one.

```release-note
NONE
```


hack/ci/run-cilium-e2e-test.sh
```diff
@@ -30,7 +30,7 @@ function cleanup() {
 trap cleanup EXIT SIGINT SIGTERM
 
 export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
-export KUBERMATIC_YAML=hack/ci/testdata/kubermatic_cilium.yaml
+export KUBERMATIC_YAML=hack/ci/testdata/kubermatic_headless.yaml
```
xrstf, line 33: `hubble` addon is now installed explicitly, so no need to configure KKP to have it as a default addon. This also makes it easier to test on dev, where hubble might not be setup as a default addon.

PratikDeoghare (approved):

PR #10172, opened 2022-06-24
create cluster namespace first, then update ClusterStatus

xrstf:
This prevents a race condition that can sometimes lead to controllers not being able to reconcile.

Previously, we first updated the ClusterStatus with the namespaceName and _then_ created the namespace. Between the actual creation and the namespace appearing in our caches, other controllers might already be triggered by the ClusterStatus change and begin their work. But they would then potentially fail.

This PR flips the order to first create the namespace, wait for it to exist in our cache and then update the ClusterStatus.

```release-note
NONE
```


embik (approved):
/approve

PR #10182, opened 2022-06-24
Introduce pkg/util/wait to provide better timeout feedback

xrstf:
We use apimachinery's `wait` ("k8swait") in many places, but I often found it lacking.

Conditions have the signature of `func() (bool, error)`.

* If the error is not nil, the polling is stopped immediately and the err is returned.
* If the bool is true, the condition is met and `nil` is ultimately returned from polling.
* If the bool is false, the condition is tried again, possibly until a timeout happens. If a timeout happens, we only know _that_ a timeout happens, but not necessarily _why_.

We have a lot of polling conditions that consist of multiple criteria. If multiple code paths inside a condition `return false, nil`, we lose the disctinction, as from k8swait's perspective, it's just a plain timeout:

```go
err := wait.Poll("1s", "10s", func() (bool, error) {
   value, err := doStuff()
   if err != nil {
      return false, err
   }

   if value > 7 {
      return false, nil
   }

   computed := magic(value)
   if !computed.IsValid() {
      return false, nil
   }

   return true, nil
})
```

We can (and often did) resort to logging, but to me it was never clear on what log level to log:

```go
err := wait.Poll("1s", "10s", func() (bool, error) {
   value, err := doStuff()
   if err != nil {
      return false, err
   }

   if value > 7 {
      log.Debug("value is still larger than seven") // debug or info?
      return false, nil
   }

   computed := magic(value)
   if !computed.IsValid() {
      log.Debug("computed result is invalid") // debug or info?
      return false, nil
   }

   return true, nil
})
```

If you do it on INFO, it might be noisy and spammy. If you do it on DEBUG and a test fails, you will wonder "awww how do I now rerun these tests with debug logging enabled? What's the prow command for that?". Both options suck.

Another option is to write even more code and to split the 2 parts of the condition into 2 Poll calls:

```go
var value int

log.Info("Waiting for a valid number...")
err := wait.Poll("1s", "10s", func() (bool, err error) {
   value, err = doStuff()
   if err != nil {
      return false, err
   }

   return value <= 7, nil
})
if err != nil {
   log.Error("Failed to wait for valid number: %v", err)
}

log.Info("Waiting for magic...")
err := wait.Poll("1s", "10s", func() (bool, err error) {
   computed := magic(value)
   return computed.IsValid(), nil
})
if err != nil {
   log.Error("Failed to compute things: %v", err)
}
```

But now you need to manage multiple timeouts and if the value can change from the first Poll, you're s** out of luck.

I therefore propose a tiny extra package which wraps k8swait. The conditions that it uses follow `func() (transient error, terminal error)`. It works pretty much exactly like k8swait: In cases where you'd returned `false` with k8swait, you now return a transient error. If a timeout happens, the transient error is included in the final error, so you do not just get "timed out waiting for the condition", but "timed out waiting for the condition, because _transient error_":

```go
err := wait.Poll("1s", "10s", func() (error, error) {
   value, err := doStuff()
   if err != nil {
      return nil, err
   }

   if value > 7 {
      return errors.New("value is still larger than seven"), nil
   }

   computed := magic(value)
   if !computed.IsValid() {
      return errors.New("computed result is invalid") , nil
   }

   return nil, nil
})

fmt.Println(err) // => e.g. "timed out waiting for the condition: computed result is invalid"
```

The new package is not applicable everywhere, cause sometimes we know, from experience, that a progressing INFO/DEBUG logging during a condition is more convenient. By not logging in a condition, we lose a bit of "tracking" ability. So the new package is not a one-size-fits-all solution.

_However_ I extended the new wait package with `PollLog` and `PollImmediateLog`, which will log the current transient error on every iteration:

```go
// This PollLog statement...
err := wait.PollLog(log, "1s", "10s", func() (error, error) {
   conditions, err := getClusterConditions()
   if err != nil {
      return nil, err
   }

   if conditions.Valid() {
      return nil, nil
   }

   return fmt.Errorf("missing conditions: %v", conditions), nil
})

// would lead to

//{"level":"info","time":"2022-06-24T21:55:15Z","msg":"Waiting","status":"nmissing conditions: [a, b, c]"}
//{"level":"info","time":"2022-06-24T21:55:16Z","msg":"Waiting","status":"nmissing conditions: [a, b, c]"}
//{"level":"info","time":"2022-06-24T21:55:17Z","msg":"Waiting","status":"nmissing conditions: [a, c]"}
//{"level":"info","time":"2022-06-24T21:55:18Z","msg":"Waiting","status":"nmissing conditions: [c]"}
```

While this solves the "staring at the screen and not getting feedback is annoying"-problem. The only downside I can see from this is that we lose some of the structured logging benefits. Compare the log lines above to what you could do if you manually logged:

```go
err := wait.Poll("1s", "10s", func() (error, error) {
   conditions, err := getClusterConditions()
   if err != nil {
      return nil, err
   }

   if conditions.Valid() {
      return nil, nil
   }

   log.Info("Missing conditions", "conditions", conditions)

   return fmt.Errorf("missing conditions: %v", conditions), nil
})
```

But I would consider this a minor loss and if structured logging is really important, you can still do it manually and simply use Poll() instead of PollLog().

(During all the rewrites I noticed that Zap is perfectly capable to print nice looking time.Duration values, so there is no need to get the seconds and log `.Infow("...", "duration-in-seconds", elapsed.Seconds())`. So I changed a few places and made the logging a tiny bit more consistent.)

```release-note
NONE
```

### DIFF
--- a/cmd/conformance-tester/pkg/runner/waiters.go
+++ b/cmd/conformance-tester/pkg/runner/waiters.go
@@ -18,6 +18,7 @@ package runner
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -27,11 +28,12 @@ import (
 	ctypes "k8c.io/kubermatic/v2/cmd/conformance-tester/pkg/types"
 	"k8c.io/kubermatic/v2/cmd/conformance-tester/pkg/util"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/util/wait"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/util/sets"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -40,19 +42,19 @@ func waitForControlPlane(ctx context.Context, log *zap.SugaredLogger, opts *ctyp
 	started := time.Now()
 	namespacedClusterName := types.NamespacedName{Name: clusterName}
 
-	err := wait.Poll(3*time.Second, opts.ControlPlaneReadyWaitTimeout, func() (done bool, err error) {
+	err := wait.Poll(3*time.Second, opts.ControlPlaneReadyWaitTimeout, func() (transient error, terminal error) {
 		newCluster := &kubermaticv1.Cluster{}
 
 		if err := opts.SeedClusterClient.Get(ctx, namespacedClusterName, newCluster); err != nil {
 			if apierrors.IsNotFound(err) {
-				return false, nil
+				return err, nil
 			}
 		}
 
 		// Check for this first, because otherwise we instantly return as the cluster-controller did not
 		// create any pods yet
 		if !newCluster.Status.ExtendedHealth.AllHealthy() {
-			return false, nil
+			return errors.New("cluster is not all healthy"), nil
 		}
 
 		controlPlanePods := &corev1.PodList{}
@@ -61,16 +63,21 @@ func waitForControlPlane(ctx context.Context, log *zap.SugaredLogger, opts *ctyp
 			controlPlanePods,
 			&ctrlruntimeclient.ListOptions{Namespace: newCluster.Status.NamespaceName},
 		); err != nil {
-			return false, fmt.Errorf("failed to list controlplane pods: %w", err)
+			return nil, fmt.Errorf("failed to list controlplane pods: %w", err)
 		}
 
+		unready := sets.NewString()
 		for _, pod := range controlPlanePods.Items {
 			if !util.PodIsReady(&pod) {
-				return false, nil
+				unready.Insert(pod.Name)
 			}
 		}
 
-		return true, nil
+		if unready.Len() == 0 {
+			return nil, nil
+		}
+
+		return fmt.Errorf("not all Pods are ready: %v", unready.List()), nil
 	})
 	// Timeout or other error
 	if err != nil {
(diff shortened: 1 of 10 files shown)


xrstf:
/hold

feedback wanted

embik:
I like this a lot :+1: I'll wait with lgtm'ing this since others might want to comment on this as well, but I think this is a neat situational improvement over the standard k8s wait utilities.

embik (approved):

cmd/conformance-tester/pkg/runner/waiters.go
```diff
@@ -27,32 +28,34 @@ import (
...
-				return false, nil
+				return err, nil
 			}
 		}
 
 		// Check for this first, because otherwise we instantly return as the cluster-controller did not
 		// create any pods yet
 		if !newCluster.Status.ExtendedHealth.AllHealthy() {
-			return false, nil
+			return errors.New("cluster is not all healthy"), nil
```
xmudrii, line 58: Why do we use `errors` instead of `fmt.Errorf`? I thought we wanted to get rid of `errors`.

cmd/conformance-tester/pkg/tests/metrics.go
```diff
@@ -157,37 +158,36 @@ func TestUserClusterPodAndNodeMetrics(ctx context.Context, log *zap.SugaredLogge
...
 		metricPod := &corev1.Pod{}
 		if err := userClusterClient.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: pod.Name}, metricPod); err != nil {
-			log.Warnw("Failed to get test metric pod", zap.Error(err))
-			return false, nil
+			return fmt.Errorf("failed to get test metric pod: %w", err), nil
 		}
 
 		if !util.PodIsReady(metricPod) {
-			return false, nil
+			return errors.New("Pod is not ready"), nil
```
xmudrii, line 168: Same here -- why not `fmt.Errorf`?

pkg/util/wait/wait_test.go
```diff
@@ -0,0 +1,83 @@
...
+	err := Poll(1*time.Millisecond, 100*time.Millisecond, func() (error, error) {
+		executions++
+		return nil, nil
+	})
+
+	if err != nil {
+		t.Fatalf("Poll should have returned nil, but returned %v", err)
+	}
+
+	if executions > 1 {
```
xmudrii: Nit: I'd check that `executions != 1`, i.e. to cover the case when it never runs.

pkg/util/wait/wait_test.go
```diff
@@ -0,0 +1,83 @@
...
+		t.Fatalf("err should be a wrapped ErrWaitTimeout, but is %+v", err)
+	}
+
+	if !strings.Contains(err.Error(), "transient") {
+		t.Fatalf("err should have returned the transient error message, but was: %v", err)
+	}
+}
+
+func TestPollTerminalError(t *testing.T) {
+	executions := 0
```
xmudrii, line 64: Nit: we have this variable, but I don't see that we use it anywhere.

xmudrii (commented):
A couple of nits, but otherwise looks great. :zap:

pkg/util/wait/wait_test.go
xrstf: Fixed

xmudrii (approved):

PR #10185, opened 2022-06-27
Update member information to poll for client URL in etcd-launcher

embik:

A `wait` was wrapped around checking potential unwanted members in `etcd-launcher`, but within the wrapped function we were checking for the client URLs array length. That could never change though. This PR adds a small update to the `member` variable in case the member gets updated within the 15 seconds we're waiting for it to be considered unhealthy.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #10190, opened 2022-06-27
consolidate the various ways we construct an S3 client

xrstf:
This PR introduces `pkg/util/s3` with a single function that takes care of building an S3 client both for HTTP and HTTPS endpoints. The remaining 4 or 5 spots in KKP were rewritten to use the new helper. There is only one place, the etcdrestore stuff, that now _learned_ how to deal with HTTP, all others were already capable of handling it.

```release-note
NONE
```

### DIFF
--- a/cmd/s3-storeuploader/main.go
+++ b/cmd/s3-storeuploader/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"crypto/x509"
 	"fmt"
+	"net/url"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


cmd/s3-storeuploader/main.go
```diff
@@ -149,3 +156,25 @@ func getUploaderFromCtx(log *zap.SugaredLogger, opt options) (*storeuploader.Sto
...
 	return uploader, nil
 }
+
+func prependScheme(u string, secure bool) (string, error) {
+	parsed, err := url.Parse(u)
+	if err != nil {
+		return "", err
+	}
+
+	if parsed.Scheme != "" && parsed.Host == "" {
```
embik, line 166: Shouldn't this be the other way around, no scheme and a non-empty host?

embik (changes requested):

embik (approved):

PR #10211, opened 2022-06-29
Use Go Generics for Reconciling

xrstf:
### What does this PR do

This PR is an experiment to see if and how Go Generics can improve our codebase. I chose the obvious target: our reconciling framework, the one piece in KKP where we rely heavily on code generation.

### Goals

* For a developer of a controller, the experience should stay the same. They should still simply call a function and be done, without too much type juggling.
* I want to get rid of the repeated (generated) code.

### The Results

... are mixed.

As it turns out, our reconciling is based around interfaces, generics and closures. Mixing and matching them is not easy and can lead to code that is quickly unreadable.

While this PR achieved the main goal of removing most of the generated code, there are still some leftovers. The `NamedXXXXXCreatorGetter` types are still auto-generated because it's much easier to write `[]reconciling.NamedConfigMapCreatorGetter` than `[]func(string, func (T) (T, error)) func (T) (T, error)`. So those generated types are really only for readability.

For the most part I was able to build the new reconciling based on generics. However the object modifiers (which should be called "object creator modifiers", as they do not modify an object but instead wrap around an ObjectCreator) I left as "interfaced", i.e. `func(ctrlruntime.Object) ctrlruntime.Object`. This is because they are meant to be generic (not _go generic_) and it just makes sense to use the interface that exists already. However this seemingly small piece of the puzzle makes things infinitely more complicated: All of the reconciling is generic, but the modifiers are interface-based. So now we need to convert between T and ctrlruntime.Object in a few places. You can see the type juggling in `makeGenericObjectCreator[T]` and `wrapGenericObjectModifier[T]`.

You might wonder "Why not simply make the modifiers generic?". Easy: If we made them generic (with Go generics), then whoever instantiates them needs to know the type of the object they are reconciling (so that they can do `modifier := reconciling.ObjectOwner[*corev1.ConfigMap]()`). This is trivial to know, but tedious to write. It also means that you cannot reuse a modifier for a different set of resources, so if your controller reconciles ConfigMaps, Secrets and Deployments, you need 3 distinct modifiers, one for each type. I found this to be too much hassle and rather opted to "fix it" in the reconciling framework (so the burden of dealing with all that nonsense is not put on the developer).

Similarily, unstructured objects were already special and continue to be special. We still need to inject the APIVersion/kind at some point and the super generic functions in the package simply do not allow for one special snowflake type. This means that for reconciling unstructured objects, one must use `EnsureNamedUnstructuredObjects`. Using the wrong function will lead to kube API errors and so it's very obvious that you've done something wrong (I thought about adding explicit panics and doing type checks, but doing type checks in a generic function feels just wrong.)

### Conclusion

To be honest, I am a bit underwhelmed. While I was able to save roughly 1k lines of code, the replacement code is super hard to read and even harder to understand. While developing I was blindly hacking on my keyboard until the red squiggly lines disappeared in my editor.

If we want to merge this and use it, cool. If not, also cool. Just because generics are new doesn't mean they are the right solution for us here.

```release-note
NONE
```

### DIFF
--- a/cmd/kubermatic-installer/cmd_convert_kubeconfig.go
+++ b/cmd/kubermatic-installer/cmd_convert_kubeconfig.go
@@ -242,16 +242,16 @@ func reconcileCluster(ctx context.Context, config *rest.Config, namespace string
 	client := mgr.GetClient()
 
 	log.Info("Reconciling ServiceAccount...")
-	if err := reconciling.ReconcileServiceAccounts(ctx, []reconciling.NamedServiceAccountCreatorGetter{
+	if err := reconciling.EnsureNamedObjects(ctx, client, namespace, []reconciling.NamedServiceAccountCreatorGetter{
 		serviceAccountCreatorGetter,
-	}, namespace, client); err != nil {
+	}); err != nil {
 		return "", fmt.Errorf("failed to create ServiceAccount: %w", err)
 	}
 
 	log.Info("Reconciling ClusterRoleBinding...")
-	if err := reconciling.ReconcileClusterRoleBindings(ctx, []reconciling.NamedClusterRoleBindingCreatorGetter{
+	if err := reconciling.EnsureNamedObjects(ctx, client, "", []reconciling.NamedClusterRoleBindingCreatorGetter{
 		clusterRoleCreatorGetterFactory(namespace),
-	}, "", client); err != nil {
+	}); err != nil {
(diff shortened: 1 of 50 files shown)


xrstf:
/hold

Waiting for feedback. This should not be merged unless there is consensus.

xrstf:
There was very little to no feedback regarding this proposal at all. Since I myself am unsure whether this is a good idea, I think the safest approach is to close this and wait for someone to come along and build something better :grin:

PR #10212, opened 2022-06-29
Bump machine controller and osm version

moadqassem:
Upgrading OSM to v0.5.0 and Machine Controller to v1.52.0. 
Fixes #

**Documentation**:

```release-note
Upgrading OSM to v0.5.0 and Machine Controller to v1.52.0 
```


pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/role.go
```diff
@@ -64,6 +64,54 @@ func KubeSystemRoleCreator() reconciling.NamedRoleCreatorGetter {
 	}
 }
 
+// KubePublicRoleCreator returns the func to create/update the Role for the OSM
+// to facilitate leaderelection.
```
ahmedwaleedmalik, line 68:
```suggestion
// to retrieve kube-apiserver address from the cluster-info configmap.
```

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/role.go
```diff
@@ -64,6 +64,54 @@ func KubeSystemRoleCreator() reconciling.NamedRoleCreatorGetter {
...
+					},
+				},
+			}
+			return r, nil
+		}
+	}
+}
+
+// DefaultRoleCreator returns the func to create/update the Role for the OSM
+// to facilitate leaderelection.
```
ahmedwaleedmalik:
```suggestion
// to retrieve kube-apiserver address from the Kubernetes endpoint.
```

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
```diff
@@ -51,6 +52,52 @@ func RoleBindingCreator() reconciling.NamedRoleBindingCreatorGetter {
...
+				Name:     resources.OperatingSystemManagerRoleName,
+				Kind:     "Role",
+				APIGroup: rbacv1.GroupName,
+			}
+			rb.Subjects = []rbacv1.Subject{
+				{
+					Kind:      rbacv1.ServiceAccountKind,
+					Name:      resources.OperatingSystemManagerRoleBindingName,
+					Namespace: metav1.NamespaceSystem,
+				},
```
ahmedwaleedmalik:
```suggestion
					Kind:     rbacv1.UserKind,
					Name:     resources.OperatingSystemManagerCertUsername,
					APIGroup: rbacv1.GroupName,
				},
```

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
```diff
@@ -51,6 +52,52 @@ func RoleBindingCreator() reconciling.NamedRoleBindingCreatorGetter {
...
+			rb.RoleRef = rbacv1.RoleRef{
+				Name:     resources.OperatingSystemManagerRoleName,
+				Kind:     "Role",
+				APIGroup: rbacv1.GroupName,
+			}
+			rb.Subjects = []rbacv1.Subject{
+				{
+					Kind:      rbacv1.ServiceAccountKind,
+					Name:      resources.OperatingSystemManagerRoleBindingName,
+					Namespace: metav1.NamespaceSystem,
```
ahmedwaleedmalik:
```suggestion
					Kind:     rbacv1.UserKind,
					Name:     resources.OperatingSystemManagerCertUsername,
					APIGroup: rbacv1.GroupName,
				},
```

ahmedwaleedmalik (changes requested):

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
moadqassem: Are you sure this is a user kind not a service account subject? https://github.com/kubermatic/operating-system-manager/blob/master/deploy/role.yaml#L114-L118

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
moadqassem: Same goes here https://github.com/kubermatic/operating-system-manager/blob/master/deploy/role.yaml#L143-L146

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
ahmedwaleedmalik: Yes, for the worker cluster we use a dedicated user "operating-system-manager", similar to how it's done for machine-controller.

The [sample deployment manifests](https://github.com/kubermatic/operating-system-manager/blob/master/deploy) will differ since it assumes that everything is happening in a single cluster and that there is no seed/worker relation.

- https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go#L44
- https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/user-cluster-controller-manager/resources/resources/machine-controller/rolebinding.go#L56

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/secret.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
ahmedwaleedmalik: minor nit; `s/2020/2022`

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
moadqassem: Ah ok, yeah that makes sense

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/secret.go
```diff
@@ -0,0 +1,42 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	cloudInitGetterToken      = "cloud-init-getter-token"
+	cloudInitGetterAnnotation = "cloud-init-getter-token"
+)
+
+// SecretCreator returns a function to create a secret in the usercluster containing the user ssh keys.
```
ahmedwaleedmalik: SecretCreator returns a function to create a secret in the usercluster, for generating an API server token against the cloud-init-getter service account.

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/rolebinding.go
ahmedwaleedmalik: I know it's a bit confusing but KKP as a consumer is a special case for OSM. So the sample manifests shouldn't cover this.

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/secret.go
moadqassem: I just love that you are doing the documentation :D

pkg/controller/user-cluster-controller-manager/resources/resources/operating-system-manager/secret.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"k8c.io/kubermatic/v2/pkg/resources"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	cloudInitGetterToken      = "cloud-init-getter-token"
+	cloudInitGetterAnnotation = "cloud-init-getter-token"
```
ahmedwaleedmalik:
```suggestion
	cloudInitGetterAnnotation = "cloud-init-getter"
```

The annotation should point to the service account name i.e. `cloud-init-getter`

Also, can we move this to https://github.com/kubermatic/kubermatic/tree/master/pkg/controller/user-cluster-controller-manager/resources/resources/cloudinitsettings and create this unconditionally? Since this is also consumed for the bare-metal nodes in machine-controller:

- https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/util/cloud_init_settings.go#L41
- https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/baremetal/provider.go#L226

ahmedwaleedmalik (changes requested):

pkg/controller/user-cluster-controller-manager/resources/reconciler.go
```diff
@@ -904,6 +904,17 @@ func (r *reconciler) reconcileSecrets(ctx context.Context, data reconcileData) e
 		}
 	}
 
+	// Operating System Manager
+	if r.enableOperatingSystemManager {
```
ahmedwaleedmalik, line 908: I think we can remove this condition, we should create this unconditionally.

moadqassem: This is only needed for machine controller and the bare metal node because back then when I wrote that module, there was no integration with OSM(since osm didn't exist back then, thus this is only used only with the OSM scenario as the bare metal provider isn't something for production nor tech review, it was just a quick PoC ;)

ahmedwaleedmalik: Then it makes sense, wasn't aware of it's background 👍

ahmedwaleedmalik (approved):
/approve ⚡ 

Although you need to fix the linter errors :D

PR #10227, opened 2022-07-01
fix NotFound detection in user-project-binding controller

xrstf:
IsNotFound only works on actual NotFound errors, not on custom errors. This PR makes sure we create a NotFound when applicable.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡ 
/retest

PR #10228, opened 2022-07-01
prevent sending empty patches

xrstf:
I was under the assumption that ctrlruntime would already prevent empty patches from being sent, but looking at the code ( https://github.com/kubernetes-sigs/controller-runtime/blob/365ae09c4c6c466edaa91c919f7654944057e0b6/pkg/client/typed_client.go#L186-L208 ), I think I was wrong.

This PR adds some equality checks to the status because I have seen lots of 

> I0701 12:27:33.287284       1 request.go:601] Waited for 2.860534999s due to client-side throttling, not priority and fairness, request: PATCH:https://35.200.248.40/apis/kubermatic.k8c.io/v1/projects/tvs8rk48xq/status

messages in our master-ctrl-mgr logs.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/master-controller-manager/project-synchronizer/controller.go
+++ b/pkg/controller/master-controller-manager/project-synchronizer/controller.go
@@ -23,6 +23,7 @@ import (
 	"go.uber.org/zap"
 
 	apiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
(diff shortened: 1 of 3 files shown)


vgramer:
very good PR description and thanks for the patch :)

/lgtm

PR #10233, opened 2022-07-04
vcd: use lowercase for provider name

ahmedwaleedmalik:

In the initial implementation of VMware Cloud Director, the provider name in the preset, datacenter, cluster spec, etc. followed came case i.e. `vmwareCloudDirector`. This is the right approach but it leads to many issues when working with the front-end or API directly. 

Although I hate doing this but to simplify everything, I've just reworded it to `vmwareclouddirector` everywhere instead of having `vmwareCloudDirector` in Cluster Spec, `vmware-cloud-director` as provider name, etc. 

It also makes sense because we already do this for kubeadm provider https://github.com/kubermatic/kubermatic/blob/master/pkg/apis/kubermatic/v1/cluster.go#L907

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow/provider-vmware-cloud-director.yaml
+++ b/.prow/provider-vmware-cloud-director.yaml
@@ -37,7 +37,7 @@ presubmits:
             - name: VERSIONS_TO_TEST
               value: "v1.24.2"
             - name: PROVIDER
-              value: "vmware-cloud-director"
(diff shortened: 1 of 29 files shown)


xrstf (approved):
/approve

PR #10240, opened 2022-07-05
Fix fetching credentials in machine validation webhook

imharshita:

**What does this PR do / Why do we need it**: fix fetching credentials in machine validation webhook

Fixes #

tests not modified yet.

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/ee/validation/machine/providers.go
+++ b/pkg/ee/validation/machine/providers.go
@@ -46,8 +46,30 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func GetMachineResourceUsage(ctx context.Context, userClient ctrlruntimeclient.Client, machine *clusterv1alpha1.Machine,
-	caBundle *certificates.CABundle) (*ResourceDetails, error) {
+const (
+	// GCP credential env.
+	envServiceAccount = "GOOGLE_SERVICE_ACCOUNT"
+	// Azure credential env.
+	envClientID       = "AZURE_CLIENT_ID"
+	envClientSecret   = "AZURE_CLIENT_SECRET"
+	envTenantID       = "AZURE_TENANT_ID"
+	envSubscriptionID = "AZURE_SUBSCRIPTION_ID"
+	// Openstack credential env.
+	envOSUsername                    = "OS_USER_NAME"
+	envOSPassword                    = "OS_PASSWORD"
+	envOSDomain                      = "OS_DOMAIN_NAME"
+	envOSApplicationCredentialID     = "OS_APPLICATION_CREDENTIAL_ID"
+	envOSApplicationCredentialSecret = "OS_APPLICATION_CREDENTIAL_SECRET"
+	envOSToken                       = "OS_TOKEN"
+	envOSProjectName                 = "OS_PROJECT_NAME"
+	envOSProjectID                   = "OS_PROJECT_ID"
+)
+
+func GetMachineResourceUsage(ctx context.Context,
+	userClient ctrlruntimeclient.Client,
+	machine *clusterv1alpha1.Machine,
+	caBundle *certificates.CABundle,
+) (*ResourceDetails, error) {
 	config, err := types.GetConfig(machine.Spec.ProviderSpec)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read machine.spec.providerSpec: %w", err)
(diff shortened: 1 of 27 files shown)


pkg/ee/validation/machine/providers.go
```diff
@@ -42,17 +42,27 @@ import (
...
+	userClient ctrlruntimeclient.Client,
+	machine *clusterv1alpha1.Machine,
+	caBundle *certificates.CABundle,
+) (*ResourceDetails, error) {
 	config, err := types.GetConfig(machine.Spec.ProviderSpec)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read machine.spec.providerSpec: %w", err)
 	}
 
+	clusterID := machine.Spec.Labels[ClusterLabel]
```
pkprzekwas: Is it safe to assume that the label exists without checking? I suggest checking and returning error in case label does not exist.

imharshita: Sure

lsviben: you could also get the cluster name as a argument for starting usercluster-webhook, as anyway each usercluster-webhook is tied to a specific cluster

lsviben:
WDYT about using the same mechanism that machine-controller is using? Set the credentials as env vars on the usercluster-webhook and then you have them ready to go?

imharshita:
> WDYT about using the same mechanism that machine-controller is using? Set the credentials as env vars on the usercluster-webhook and then you have them ready to go?

Ya will do

pkg/resources/usercluster-webhook/deployment.go
```diff
@@ -182,3 +192,96 @@ func DeploymentCreator(data webhookData) reconciling.NamedDeploymentCreatorGette
...
 }
+
+func getEnvVars(data webhookData) ([]corev1.EnvVar, error) {
+	credentials, err := resources.GetCredentials(data)
+	if err != nil {
+		return nil, err
+	}
+
+	var vars []corev1.EnvVar
+	if data.Cluster().Spec.Cloud.AWS != nil {
```
pkprzekwas, line 203: It seems that for AWS we do not read values from envs (or am I wrong here). Should we add them either way?

imharshita: ya we dont, but i thought, if we are including env credentials, lets include all.

pkg/ee/validation/machine/providers.go
```diff
@@ -46,8 +46,30 @@ import (
...
+	envSubscriptionID = "AZURE_SUBSCRIPTION_ID"
+	// Openstack credential env.
+	envOSUsername                    = "OS_USER_NAME"
+	envOSPassword                    = "OS_PASSWORD"
+	envOSDomain                      = "OS_DOMAIN_NAME"
+	envOSApplicationCredentialID     = "OS_APPLICATION_CREDENTIAL_ID"
+	envOSApplicationCredentialSecret = "OS_APPLICATION_CREDENTIAL_SECRET"
+	envOSToken                       = "OS_TOKEN"
+	envOSProjectName                 = "OS_PROJECT_NAME"
+	envOSProjectID                   = "OS_PROJECT_ID"
```
lsviben: NIT: maybe add `envGoogleServiceAccount` , `envAzureClientID` and such so its clear

pkg/resources/usercluster-webhook/deployment.go
```diff
@@ -182,3 +192,96 @@ func DeploymentCreator(data webhookData) reconciling.NamedDeploymentCreatorGette
 		}
 	}
 }
+
+func getEnvVars(data webhookData) ([]corev1.EnvVar, error) {
```
lsviben, line 196: this is the same as the machine controller one. I think we also need to add it to the usercluster-controller as the resource-quota calculation controller also uses the machine getMachineResourceUsage function

pkg/resources/usercluster-webhook/deployment.go
pkprzekwas: I don't have strong opinion about that. One the one hand, it's good to reduce places from which someone may read secrets. On the other hand, I like to keep things consistent.

pkg/ee/validation/machine/providers.go
pkprzekwas: Yeah, I think adding more context may ease code analysis.

pkg/resources/usercluster-webhook/deployment.go
imharshita: can we create a separate pr for this?

lsviben: sure

pkg/ee/validation/machine/providers.go
imharshita: Done

PR #10248, opened 2022-07-05
get rid of the term 'customer'

xrstf:
Pretty much everywhere we use the term "usercluster" (or "user cluster"), but in a few places we also had the term "customer cluster". However, in _maaany_ places we used the term "customer" for "the person(s) who install/manage KKP". But a "customer cluster" is not managed by the customer, but _their_ customers (i.e., users). Confused yet?

This PR cleans up the naming.

* "KKP admins" are now what we call the people who setup/manage KKP itself (i.e. the master and seed clusters).
* "KKP users" are the people that login to KKP and manage userclusters.

Using the term "customer" has always irked me since we moved to open source. It feels gatekeeper-y to me, as in "we only care about people who pay us" and not in "we're an open source product". So getting rid of "customer", which is loaded and misleading, is important to me :-)

```release-note
NONE
```

### DIFF
--- a/addons/README.md
+++ b/addons/README.md
@@ -2,7 +2,7 @@
 
 This directory contains all possible default addons.
 All addons will be build into a docker container which the kubermatic-addon-controller uses to install addons.
-The docker image should be freely accessible to let customers extend & modify this image for their own purpose.
+The docker image should be freely accessible to let admins extend & modify this image for their own purpose.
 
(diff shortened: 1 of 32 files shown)


ahmedwaleedmalik (approved):
/approve 🚀 

Beautiful, love the intention behind this 👍

PR #10250, opened 2022-07-05
vcd: vapp should not be required for node spec

ahmedwaleedmalik:

Currently, in the API request to create clusters, `vApp` is marked as a required field for VMware Cloud Director. This is wrong since it's an optional field that either will be provided by the user or the `seed-controller-manager` would create a dedicated vApp per user cluster for you.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/api/v1/types.go
```diff
@@ -2154,17 +2155,17 @@ func (spec *NutanixNodeSpec) MarshalJSON() ([]byte, error) {
...
-	Network          string `json:"network"`
+	CPUs             int                  `json:"cpus"`
+	CPUCores         int                  `json:"cpuCores"`
+	MemoryMB         int                  `json:"memoryMB"`
+	DiskSizeGB       *int64               `json:"diskSizeGB,omitempty"`
+	DiskIOPS         *int64               `json:"diskIOPS,omitempty"`
+	Template         string               `json:"template"`
+	Catalog          string               `json:"catalog"`
+	StorageProfile   string               `json:"storageProfile"`
+	IPAllocationMode vcd.IPAllocationMode `json:"ipAllocationMode,omitempty"`
```
xrstf, line 2166: I am fine with the intent behind the PR, but I just randomly wonder if it's a good idea to expose types defined by a third party library. It could also be a great thing. WDYT?

ahmedwaleedmalik: From my perspective, we don't have to copy and maintain `IPAllocationMode` type in the KKP code base and we avoid a few type conversions. And even from the documentation(swagger) pov this seems fine:
```
-      "x-go-package": "github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/vmwareclouddirector/types"
+      "x-go-package": "k8c.io/kubermatic/v2/pkg/api/v1"
```

So in general, I like this. But I don't have a strong opinion on this one. I'm pretty much open to doing it by defining the type in KKP as well.

xrstf (approved):
/approve

PR #10263, opened 2022-07-06
[release/v2.20] Bump machine controller and osm version (#10212)

moadqassem:
Bump machine controller v1.52.0 and osm v0.5.0 version

Fixes #

**Documentation**:

```release-note
Upgrading OSM to v0.5.0 and Machine Controller to v1.52.0 
```


xrstf (changes requested):
The kubevirt types have completely changed upstream. This would break all existing Machines. We cannot do that in a KKP patch release, not even in a minor release.

PR #10266, opened 2022-07-06
Make writing e2e tests less repetitive

xrstf:
## What does this PR do / Why do we need it

This PR refactors 6 of our e2e tests suites to not use the KKP API anymore, but use the KubeAPI directly.

## But whhhhhyyyyy?

Starting with KKP 2.21, we want to officially support "GitOps", i.e. we want to allow admins to manage KKP resources directly (using kubectl, for example).

This means that everything in KKP needs to work regardless of our KKP API. This also means that I want our CRDs to be "easy to use". No weird fields that you must set, no weird labels or UIDs and stuff.

So now effectively we have 2 APIs to deal with KKP: The KubeAPI we get for free when creating a CRD, and the KKP API that we lovingly handcrafted.

I suspect that over time we do not want to maintain 2 APIs. So I plan that our API might become a "dashboard-only" API, i.e. our dashboard is the only intended consumer for our API. This would mean fewer stability guarantees are needed and we could break it between releases as much as we want.

And if this holds true, we might even want to move the API into the dashboard repository. But even if we don't, I consider our API to be "optional" in the future and so I want to ensure that all KKP features work without our API being present (and by "work" I mean "are usuable to a non-expert admin").

And to move the API out of the KKP repo, no e2e test must depend on it. And this is why I am on a kind of crusade to rewrite our tests.

## So what happened?

I introduced a new package, `pkg/test/e2e/jig`. I chose to keep the term "jig" that we already used in other e2e tests.

These new jigs are meant to

* make it easy to create projects, clusters and machines
* be flexible so that for example multiple clusters can be created in a single test (so the coherence between the jigs should not be so hard that you cannot use multiple cluster jigs)

I then refactored a couple of e2e tests just to see if my new abstractions would a) work and b) really make the code better (remember #10211?). So ultimately I refactored

* Cilium
* Konnectivity
* MLA
* OPA
* etcd-launcher
* expose startegy

While the new jigs do add to our codebase (I added more code than I removed), the jigs are also simply better and more thought out than the old code.

I also tried to introduce some form of "standardization" for our e2e tests. All of these tests now take a `-datacenter` flag to witch what datacenter to use (note that this does not magically turn an AWS-based test into a Hetzner-based test), but once we know the datacenter, we do not have to hardcode node settings (AWS AZ, for example) into the tests anymore. We can instead read them from the Seed/datacenter.

Likewise, `-project` can be used to define a pre-existing project. If not given, a new project will be created.

All refactored tests are now headless, so no more Dex, Service Account Keys, nginx, cert-manager, roxy or roxy-admin.

## What else?

I left some comments on GitHub in a few places to explain some tricky places. Other than that, good luck reviewing. I could have split this into 6 PRs, but is it worth it?

```release-note
NONE
```


hack/ci/run-etcd-launcher-tests.sh
```diff
@@ -32,36 +32,15 @@ pushElapsed gocache_download_duration_milliseconds $beforeGocache
 export KIND_CLUSTER_NAME="${SEED_NAME:-kubermatic}"
 
 source hack/ci/setup-kind-cluster.sh
-source hack/ci/setup-kubermatic-backups-in-kind.sh
 
-echodate "Creating Hetzner preset..."
```
xrstf, line 37: No need for Hetzner stuff anymore, as the tests are now based on BYO datacenters (i.e. completely cloudless).

pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -1381,44 +1383,55 @@ func NewBytes(b64 string) Bytes {
 }
 
 func (cluster *Cluster) GetSecretName() string {
+	// new clusters might not have a name yet (if the user used GenerateName),
```
xrstf, line 1386: So far this function was only ever called on cluster objects that were already created (and hence have a name), but with the new cluster jig and us using GenerateName in a few places, this function needed to be able to handle that.

pkg/resources/health.go
```diff
@@ -37,6 +38,10 @@ func HealthyDeployment(ctx context.Context, client ctrlruntimeclient.Client, nn
 		return kubermaticv1.HealthStatusDown, err
 	}
 
+	if minReady < 0 {
+		minReady = *deployment.Spec.Replicas
+	}
+
```
xrstf, line 44: We have like 3 or 4 different functions to check if a deployment is "healthy". I chose to use this and extend it a bit to let it auto-discover the intended number of replicas. Lots of Go stdlib functions also use "-1" for "all", so I didn't feel too bad to introduce this special case.

ahmedwaleedmalik (approved):
/approve

Awesome stuff 👏

PR #10279, opened 2022-07-07
fix panic in cluster validation

xrstf:
During my work on #10266 I accdentically stumbled upon a panic in our cluster validation: `spec.Version.Semver()` can return `nil` if the version is invalid or empty.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve 🐞

PR #10292, opened 2022-07-08
fix OSM RBAC reconciling

xrstf:
Creator functions cannot override the name or namespace of the resource that they are reconciling. In the old code, no matter how hard the creators tried to set a namespace, the ReconcileRoles() function would always override it. This made all 3 creator functions effectively reconcile the exact same Role, leading to a reconcile loop.

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #10293, opened 2022-07-10
fix issue with RQ synchronizing where status was not synced properly

lsviben:

Fixes issue where Status was attempted to be synchronized to the seed clusters through the ReconcilingResourceQuotas method which ignores Status as it does a regular .Update(), but still checks equality which then fails as it expect the Status change as well. This causes an endless loop of reconciling.

Now added a separate Status update.

```release-note
NONE
```


xrstf (approved):
on the weekend!

/approve

PR #10295, opened 2022-07-11
bump kubectl binaries in Docker image

xrstf:
We stopped supporting k8s 1.21 (we still have code to reconcile it if an upgrade is delayed, but KKP 2.21 will not support creating new 1.21 clusters), so it's time to bump our kubectl binaries to be ready for k8s 1.25.

I also added some more tests to ensure that we do not only have a matching kubectl, but that this kubectl binary is also actually compatible to the cluster version. And to be super sure, I made it so that we test the actual Docker image we're publishing.

```release-note
NONE
```


xmudrii (approved):
/approve

PR #10300, opened 2022-07-11
Packet resources calculation for quota

imharshita:

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/ee/validation/machine/providers.go
+++ b/pkg/ee/validation/machine/providers.go
@@ -777,15 +777,18 @@ func getPacketResourceRequirements(ctx context.Context,
 
 	var storageReq resource.Quantity
 	for _, drive := range plan.Specs.Drives {
+		// trimming "B" as quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'.
 		storage, err := resource.ParseQuantity(strings.TrimSuffix(drive.Size, "B"))
 		if err != nil {
 			fmt.Println("error parsing machine storage request to quantity: %w", err)
 		}
+		// total storage for each types = drive count *drive Size.
 		strDrive := strconv.FormatInt(storage.Value()*int64(drive.Count), 10)
 		totalStorage, err := resource.ParseQuantity(strDrive)
 		if err != nil {
 			return nil, fmt.Errorf("error parsing machine storage request to quantity: %w", err)
 		}
+		// Adding storage value for all storage types like "SSD", "NVME".
 		storageReq.Add(totalStorage)
 	}
 
--- a/pkg/ee/validation/machine/providers.go
+++ b/pkg/ee/validation/machine/providers.go
@@ -95,7 +95,6 @@ func GetMachineResourceUsage(ctx context.Context,
 	}
 
 	var quotaUsage *ResourceDetails
-	// TODO add all providers
 	switch config.CloudProvider {
 	case types.CloudProviderFake:
 		quotaUsage, err = getFakeQuotaRequest(config)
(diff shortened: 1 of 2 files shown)


pkg/handler/common/provider/packet.go
```diff
@@ -100,6 +100,38 @@ func PacketSizes(apiKey, projectID string, quota kubermaticv1.MachineDeploymentV
...
+
+	_, err = packetclient.Do(req, root)
+	if err != nil {
+		return plan, err
+	}
+
+	plans := root.Plans
+	for _, plan := range plans {
+		if plan.Slug == instanceType {
+			return plan, nil
```
pkprzekwas: nit: name shadowing
```suggestion
	for _, currentPlan := range plans {
		if currentPlan.Slug == instanceType {
			return currentPlan, nil
```

pkg/ee/validation/machine/providers.go
```diff
@@ -725,3 +731,68 @@ func getAnexiaResourceRequirements(ctx context.Context,
...
+		totalCPUs += cpu.Count
+	}
+	cpuReq, err := resource.ParseQuantity(fmt.Sprintf("%d", totalCPUs))
+	if err != nil {
+		return nil, fmt.Errorf("error parsing machine cpu request to quantity: %w", err)
+	}
+
+	var storageReq resource.Quantity
+	for _, drive := range plan.Specs.Drives {
+		storage, err := resource.ParseQuantity(strings.TrimSuffix(drive.Size, "B"))
```
pkprzekwas: I suggest adding some comment explaining why we need to trim `B` here.

pkg/ee/validation/machine/providers.go
```diff
@@ -76,6 +78,10 @@ const (
 	// Alibaba credential env.
 	envAlibabaAccessKeyID     = "ALIBABA_ACCESS_KEY_ID"
 	envAlibabaAccessKeySecret = "ALIBABA_ACCESS_KEY_SECRET"
+	// Packet credential env.
+	envPacketToken           = "PACKET_API_KEY"
+	envPacketProjectID       = "PACKET_PROJECT_ID"
+	packetCustomInstanctType = "storage.custom"
```
lsviben:
```suggestion
	packetCustomInstanceType = "storage.custom"
```

Btw what is this? In this case we cant get the resource quantities?

imharshita: for this type, i got these details:
`packngo.Plan{ID:"###", Slug:"storage.custom", Name:"Custom Storage Appliance", Description:"Custom storage appliance", Line:"baremetal", Specs:packngo.Specs{Cpus:[packngo.Cpus{Count:2, Type:"PureStorage Control"}], Memory:packngo.Memory{Total:""}, Drives:[packngo.Drives{Count:0, Size:"", Type:""}], Nics:[packngo.Nics{Count:0, Type:""}]}, Pricing:packngo.Pricing{Hour:0, Month:0}, DeploymentTypes:[], Class:"storage.custom"}`
as you can see data is nil for this type:
`Drives:[packngo.Drives{Count:0, Size:"", Type:""}
packngo.Memory{Total:""}`

lsviben: well for drives, but you can still get CPU. Do we need to treat it in a special way? It will just go through and set Memory and Storage to 0?

imharshita: yes, well for  drive.Size: "" we get this err:
`error parsing machine storage request to quantity and drive.Size quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$' `

imharshita: we can take the value of the CPU and then put checks in place for drive and memory.
but as it was a "custom" type, I thought to better skip it.

lsviben: Ah ok, the empty string is no good.

Then add a skip if its empty? Better to defend against that, they could add some new type that also has this issue. WDYT?

imharshita: yeah, I mean currently it doesn't have any other type with an empty string.
but  I agree, we can be better prepared for future added types.
so for this type, we will only get CPU values, and I will put checks for drive and memory

PR #10303, opened 2022-07-11
API support for GroupProjectBinding resources [part 2]

### LINKED ISSUE #10003: GroupProjectBinding API endpoints

pkprzekwas:
Add POST, DELETE and PATCH endpoints to interact with GroupProjectBinding resources.

Fixes #10003 

```release-note
Add REST API for interacting with GroupProjectBinding resources.
```


pkg/ee/group-project-binding/handler/handler.go
```diff
@@ -136,3 +170,211 @@ func GetGroupProjectBinding(ctx context.Context, request interface{},
...
+// swagger:parameters patchGroupProjectBinding
+type patchGroupProjectBindingReq struct {
+	groupProjectBindingReq
+
+	// in: body
+	// required: true
+	Body struct {
+		Role string `json:"role"`
+	}
+}
```
lsviben: do we want to allow changing the Group?

pkg/handler/v2/group-project-binding/groupprojectbinding.go
```diff
@@ -59,3 +59,66 @@ func GetGroupProjectBindingEndpoint(
...
+			req,
+			userInfoGetter,
+			projectProvider,
+			privilegedProjectProvider,
+			bindingProvider,
+		); err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
```
lsviben: NIT:

maybe simpler would be:
```suggestion
		return nil, createGroupProjectBinding(
			ctx,
			req,
			userInfoGetter,
			projectProvider,
			privilegedProjectProvider,
			bindingProvider)
	}
```

pkg/ee/group-project-binding/handler/handler.go
```diff
@@ -136,3 +170,211 @@ func GetGroupProjectBinding(ctx context.Context, request interface{},
...
+		},
+		Spec: kubermaticv1.GroupProjectBindingSpec{
+			ProjectID: req.ProjectID,
+			Group:     req.Body.Group,
+			Role:      req.Body.Role,
+		},
+	}); err != nil {
+		return common.KubernetesErrorToHTTPError(err)
+	}
+	return nil
```
lsviben: Usually in KKP we return the created/patched resources, could you do that as well to keep consistency?

pkg/ee/group-project-binding/handler/handler.go
pkprzekwas: That was my initial thought. But I don't have strong opinion.

lsviben: I would allow, I dont see a reason why not

PR #10304, opened 2022-07-11
Adding datacenter-level "IPv6 enabled" switch for OpenStack

### LINKED ISSUE #10135: Add datacenter-level "IPv6 enabled" switch for OpenStack
Since IPv6 support in OpenStack is highly datacenter-setup specific, we should add an datacenter-level "IPv6 enabled" switch to [OpenStack-specific datacenter API](https://github.com/kubermatic/kubermatic/blob/88ab8cf84b916e6df147fbf4fcaafa8e97ec5a93/pkg/apis/kubermatic/v1/datacenter.go#L378) of KKP. Based on the flag, the [datacenter API endpoint](https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v1/dc/datacenter.go#L623-L629) should return `IPv6Enabled` set appropriately and the cluster webook should prevent creating dual-stack clusters in non-IPv6-enabled datacenters.

hbernardo:
This PR adds a new config for enabling IPv6 for OpenStack datacenter.

Fixes #10135

```release-note
NONE
```


pkg/validation/cluster.go
```diff
@@ -335,6 +335,20 @@ func ValidateClusterNetworkConfig(n *kubermaticv1.ClusterNetworkingConfig, cni *
 			fmt.Sprintf("%s proxy mode can be used only when Konnectivity is enabled", resources.EBPFProxyMode)))
 	}
 
+	if n.IPFamily == kubermaticv1.IPFamilyDualStack && dc != nil {
```
rastislavs, line 338: cc @PratikDeoghare now I realised that with this we won't be able to easily test dual-stack on not yet supported providers (as it won't be allowed by webhook). What do you think about this? I think it is good to have this check here for production, but would be good to have some escape path for testing...

PratikDeoghare: If we could read these providers from a configmap then we could patch it for tests?

rastislavs: yeah, or should we do it in `KubermaticConfiguration`?

PratikDeoghare: That should work even better.

rastislavs: Still not 100% sure if we should complicate this with KC.
Maybe the webhook should check only the providers with datacenter-specific IPv6 support (for now OpenStack, next would be vsphere) and allow dual-stack for all other providers (for now).

PratikDeoghare: Right. Too soon to be changing KC.

I don't see anything too bad happening if we allow all for now.

rastislavs: agreed :)

ahmedwaleedmalik (approved):
/approve ⚡

PR #10305, opened 2022-07-12
Use quay.io registry for Canal images, bump Canal v3.20 version to v3.20.5

dharapvj:
Updated default registry for canal to quay.io to beat image pull limits at docker.io

Fixes #

We face imagePullBackoff regularly at customer location and hence this is an attempt to reduce dependency on Dockerhub bit by bit. Similar change is already [in effect in kubeone addons](https://github.com/kubermatic/kubeone/blob/3efce75c4dccc77bf2c1b10f05ba406f5c5af7a2/pkg/templates/images/images.go#L203)

**Documentation**:

```release-note
Use quay.io as the default registry for Canal CNI images, bump Canal v3.20 version to v3.20.5.
```

### DIFF
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -411,7 +411,7 @@ spec:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: '{{ Registry "quay.io" }}/calico/cni:v3.8.0'
+          image: '{{ Registry "docker.io" }}/calico/cni:v3.8.0'
           command: ["/install-cni.sh"]
           env:
             # Name of the CNI config file to create.
@@ -442,7 +442,7 @@ spec:
         # host.
         - name: calico-node
           # FIXME: Remove FELIX_IGNORELOOSERPF from env when this is v3.12.0+
-          image: '{{ Registry "quay.io" }}/calico/node:v3.8.0'
+          image: '{{ Registry "docker.io" }}/calico/node:v3.8.0'
           env:
             # Use Kubernetes API as the backing datastore.
             - name: DATASTORE_TYPE
@@ -529,7 +529,7 @@ spec:
         # This container runs flannel using the kube-subnet-mgr backend
         # for allocating subnets.
         - name: kube-flannel
-          image: '{{ Registry "quay.io" }}/coreos/flannel:v0.11.0'
+          image: '{{ Registry "docker.io" }}/coreos/flannel:v0.11.0'
           command: [ "/opt/bin/flanneld", "--ip-masq", "--kube-subnet-mgr" ]
           securityContext:
             privileged: true
--- a/addons/canal/canal.yaml
+++ b/addons/canal/canal.yaml
@@ -529,7 +529,7 @@ spec:
         # This container runs flannel using the kube-subnet-mgr backend
         # for allocating subnets.
         - name: kube-flannel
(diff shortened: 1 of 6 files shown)


addons/canal/canal_v3.20.yaml
```diff
@@ -3791,7 +3791,7 @@ spec:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: '{{ Registry "docker.io" }}/calico/cni:v3.20.2'
+          image: '{{ Registry "quay.io" }}/calico/cni:v3.20.2'
```
rastislavs: The `3.20.2` version is missing multi-arch manifest (so would not work on ARM nodes). We can:
 - a) keep `3.20.2` on dockerhub
 - b) bump the version to `3.20.5` 
 
image

addons/canal/canal_v3.19.yaml
```diff
@@ -3591,7 +3591,7 @@ spec:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: '{{ Registry "docker.io" }}/calico/cni:v3.19.1'
+          image: '{{ Registry "quay.io" }}/calico/cni:v3.19.1'
```
rastislavs: `v3.19.1` (as well as other 3.19 versions) is missing multi-arch manifest (so would not work on ARM nodes).
Let's keep it on dokerhub.

addons/canal/canal.yaml
```diff
@@ -411,7 +411,7 @@ spec:
         # This container installs the CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: '{{ Registry "docker.io" }}/calico/cni:v3.8.0'
+          image: '{{ Registry "quay.io" }}/calico/cni:v3.8.0'
```
rastislavs: v3.8.0 (as well as other 3.8 versions) is missing multi-arch manifest (so would not work on ARM nodes).
Let's keep it on dokerhub.

addons/canal/canal.yaml
```diff
@@ -529,7 +529,7 @@ spec:
         # This container runs flannel using the kube-subnet-mgr backend
         # for allocating subnets.
         - name: kube-flannel
-          image: '{{ Registry "quay.io" }}/coreos/flannel:v0.11.0'
+          image: '{{ Registry "docker.io" }}/coreos/flannel:v0.11.0'
```
rastislavs: please do not touch coreos/flannel image here

addons/canal/canal_v3.19.yaml
```diff
@@ -3735,7 +3735,7 @@ spec:
         # This container runs flannel using the kube-subnet-mgr backend
         # for allocating subnets.
         - name: kube-flannel
-          image: '{{ Registry "quay.io" }}/coreos/flannel:v0.15.1'
+          image: '{{ Registry "docker.io" }}/coreos/flannel:v0.15.1'
```
rastislavs: please do not touch coreos/flannel image here

addons/canal/canal_v3.20.yaml
dharapvj: bumped to `3.20.5`

addons/canal/canal.yaml
dharapvj: reverted

PR #10311, opened 2022-07-12
Resourcequotas to cluster scope

lsviben:

Initially ResourceQuotas were Namespace-scoped, and "forced" into the `kubermatic` namespace. As its related to the cluster scoped resources, like Project, it makes sense to have it cluster scoped as well.

Also it made rbacs difficult to manage properly as various components in different namespaces needed access to resource quotas.

As the ResourceQuotas were not released yet, better to do this now before the release.

```release-note
NONE
```


xrstf (approved):
/approve

PR #10315, opened 2022-07-12
e2e test for OIDC group support

### LINKED ISSUE #10310: Add e2e test cases for OIDC groups in KKP
**User Story**
Giving access to resources based on group association as well as protecting other user's resources from unwanted access  are crucial features of KKP. To be sure those work correctly we should add comprehensive test to cover this feature.

Test should check if:
- user is not able to access protected resources before assigning group
- user can access protected resources while added to group
- user cannot access protected resources once removed from group

**Acceptance Criteria**
A comprehensive e2e test case set is added and tests the scenarios mentioned above.

pkprzekwas:
This PR adds a simple e2e test that proves that we do support OIDC groups in KKP. Unfortunately, dex doesn't allow groups for static users (https://github.com/dexidp/dex/issues/1080), hence I had to deploy a small OpenLDAP instance to get a user with associated groups. 

Fixes #10310 

```release-note
NONE
```


xrstf (approved):
Hot damn, this is a terrible limitation in Dex, but an awesome PR.

Though I now hope even more that we move the API into the dashboard repo, so the dependency on OpenLDAP can also move over there :grin:

ahmedwaleedmalik (approved):
/approve ⚡ 

Good job 👍

PR #10324, opened 2022-07-13
add providerName to externalclusters, add webhook to default it

xrstf:
Just like we do it for regular Cluster objects, this PR adds a `providerName` field to ExternalClusters, so we can pretty print the provider name in the output of `kubectl get externalclusters`.

This PR contains the necessary new webhook, configuration, defaulting etc.

```release-note
NONE
```

### DIFF
--- a/pkg/apis/kubermatic/v1/external_cluster.go
+++ b/pkg/apis/kubermatic/v1/external_cluster.go
@@ -32,10 +32,22 @@ const (
 	ExternalClusterKind = "ExternalCluster"
 )
 
+type ExternalClusterProvider string
+
+const (
+	ExternalClusterGKEProvider     ExternalClusterProvider = "gke"
+	ExternalClusterEKSProvider     ExternalClusterProvider = "eks"
+	ExternalClusterAKSProvider     ExternalClusterProvider = "aks"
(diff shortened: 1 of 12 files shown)


pkg/apis/kubermatic/v1/external_cluster.go
```diff
@@ -32,10 +32,22 @@ const (
 	ExternalClusterKind = "ExternalCluster"
 )
 
+type ExternalClusterProvider string
```
ahmedwaleedmalik, line 37:
```suggestion
// +kubebuilder:validation:Enum=gke;aks;eks;kubeone

type ExternalClusterProvider string
```

ahmedwaleedmalik (approved):
/approve
/hold for @imharshita to review this as well

pkg/apis/kubermatic/v1/external_cluster.go
ahmedwaleedmalik: Just a nit

xrstf: love it

ahmedwaleedmalik (approved):
/approve 🦈

PR #10325, opened 2022-07-13
Fix log levels

xrstf:
We should use the INFO log level only for actionable messages, like when we actually created/deleted a resource. There is little use (IMHO) to log "Skipping reconcile for StatefulSets, is not healthy yet" over and over again. So this PR changes all log statements that used INFO to DEBUG if they only log non-essential progress information.

I also, randomly, replaced Eventf() with Event() where we don't actually do string formatting.

```release-note
Improve log verbosity
```


ahmedwaleedmalik (approved):
/approve 🏎️

PR #10329, opened 2022-07-13
set env credentials in userclustercontroller

### LINKED ISSUE #10297: Add credential envVars in usercluster-controller

As a developer
I would like to add credential envVars in usercluster-controller
in order to utilise them in getMachineResourceUsage for resourceQuotas

### Solution details

- Requirement 1
- Requirement 2
- Requirement 3

imharshita:

Fixes https://github.com/kubermatic/kubermatic/issues/10297

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/resources/data.go
+++ b/pkg/resources/data.go
@@ -696,7 +696,7 @@ func (d *TemplateData) KubermaticConfiguration() *kubermaticv1.KubermaticConfigu
 	return d.config
 }
 
-func GetEnvVars(data *TemplateData) ([]corev1.EnvVar, error) {
+func (data *TemplateData) GetEnvVars() ([]corev1.EnvVar, error) {
 	credentials, err := GetCredentials(data)
 	if err != nil {
 		return nil, err
--- a/pkg/resources/data.go
+++ b/pkg/resources/data.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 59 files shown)


pkg/resources/usercluster/deployment.go
```diff
@@ -374,3 +381,96 @@ func getLabelsArgValue(cluster *kubermaticv1.Cluster) (string, error) {
 	}
 	return string(bytes), nil
 }
+
+func getEnvVars(data userclusterControllerData) ([]corev1.EnvVar, error) {
```
lsviben: as this is duplicated in multiple places, could you extract it so that we need to maintain it in only one place?

imharshita: can you plz suggest a place where we can move this method https://github.com/kubermatic/kubermatic/blob/master/pkg/resources/machinecontroller/deployment.go#L197
to so that other webhooks and controllers can use it.
is it safe to make this method public.

lsviben: as its used by the:
pkg/resources/machinecontroller/deployment.go
pkg/resources/usercluster-webhook/deployment.go
and now by
pkg/resources/usercluster/deployment.go

maybe lets put it to the TemplateData in pkg/resources/data.go

pkg/resources/data.go
```diff
@@ -694,3 +695,97 @@ func (d *TemplateData) Seed() *kubermaticv1.Seed {
 func (d *TemplateData) KubermaticConfiguration() *kubermaticv1.KubermaticConfiguration {
 	return d.config
 }
+
+func GetEnvVars(data *TemplateData) ([]corev1.EnvVar, error) {
```
lsviben:
```suggestion
func (d *TemplateData) GetEnvVars() ([]corev1.EnvVar, error) {
```

then you can:
- add `GetEnvVars() ([]corev1.EnvVar, error)` to the machinecontrolerdata and usercontrollerdata intefaces
- use `data.GetEnvVars()` instead of  `resources.GetEnvVars(data.(*resources.TemplateData))`

PR #10332, opened 2022-07-14
ResourceQuota API improvements

### LINKED ISSUE #10316: Return project human-readable-name in ResourceQuotas API endpoint
To display nicely the Project ResourceQuotas in the UI, return the project human readable name alongside the generated one in the ResourceQuota API endpoints

lsviben:
- Adds a SubjectHumanReadableName to the API ResourceQuota model so that the UI can display pretty names instead of a generated string.
- Adds SubjectHumanReadableName to mutation
- Fixes gettting resourcequota for project which assumed that the resource quota names are always generated by API.
- Adds tests for above

Fixes #10316 

```release-note
NONE
```


pkg/ee/resource-quota/provider.go
```diff
@@ -77,14 +79,33 @@ func (p *ResourceQuotaProvider) Get(ctx context.Context, userInfo *provider.User
...
+		return nil, fmt.Errorf("error creating resource quota subject name requirement: %w", err)
+	}
+	subjectKindReq, err := labels.NewRequirement(kubermaticv1.ResourceQuotaSubjectKindLabelKey, selection.Equals, []string{kind})
+	if err != nil {
+		return nil, fmt.Errorf("error creating resource quota subject kind requirement: %w", err)
+	}
+	subjectSelector := labels.NewSelector().Add(*subjectNameReq, *subjectKindReq)
+
+	resourceQuotaList := &kubermaticv1.ResourceQuotaList{}
+	if err := p.privilegedClient.List(ctx,
```
pkprzekwas, line 93: Can you elaborate why we need to list before we do get? Now we need 2 kube api request instead of one.

pkprzekwas: Nvmd, I think I get it. It's to support more resources than just projects, right?

pkg/ee/mutation/resourcequota/mutation.go
```diff
@@ -117,6 +118,8 @@ func ensureProjectOwnershipRef(ctx context.Context, client ctrlruntimeclient.Cli
 	ownRefs = append(ownRefs, projectRef)
 	resourceQuota.SetOwnerReferences(ownRefs)
 
+	kubernetes.EnsureLabels(resourceQuota, map[string]string{kubermaticv1.ResourceQuotaSubjectHumanReadableNameLabelKey: project.Spec.Name})
```
pkprzekwas, line 124: Can you handle labels from `ensureResourceQuotaLabels` in the same way and delete that function as part of refactoring in this PR? Or you prefer to do it separately?

pkg/ee/resource-quota/handler_test.go
```diff
@@ -181,16 +182,16 @@ func TestHandlerResourceQuotas(t *testing.T) {
...
+		      "subjectName": "testproject",
 				"quota": {
 					"cpu": 10,
 					"memory": "64Gi",
 					"storage": "256Gi"
 				}
 			}`,
-			existingAPIUser: test.GenDefaultAdminAPIUser(),
-			existingObjects: existingResourceQuotas,
+			existingAPIUser: test.GenAPIUser("John", "john@acme.com"),
```
pkprzekwas, line 193: I wonder why this works for non-admin users. Shouldn't it return 403? (same for other non GET endpoints)

pkg/ee/resource-quota/provider.go
lsviben: First we list to find the resource quota for this project. As we dont know the name, we need to list with the filter on subjectName and subjectKind.
For this we need to use a privileged client, as the users wont be able to list (they only have `get` permissions on resource quotas from their projects)

Second request is the GET using the impersonation client so that we make sure that the user can actually access this ResourceQuota.

pkg/ee/mutation/resourcequota/mutation.go
lsviben: good point, ill change it

pkg/ee/resource-quota/handler_test.go
lsviben: It is an admin user, i added it as an admin above i the test setup

pkg/ee/mutation/resourcequota/mutation.go
lsviben: changed and also fixed a but with the EnsureLabels :)

pkg/ee/resource-quota/provider.go
pkprzekwas: Is there any particular reason why you don't want to relay on resource quota names which are build with a convention of `<kind>-<name>`?

lsviben: Because that convention only works when resource quotas are created through the KKP API.

PR #10333, opened 2022-07-14
Fix osm roles and rolebindings

moadqassem:
Based on the user reconciliations functions, resources that belong to the same namespace are grouped and reconciled together. In the previous implementation, this was not respected for OSM resources such as Roles and RoleBinding, which resulted to create those resources in the wrong namespaces. This PR fixed this issue by creating the needed permissions to the correct namespace. 

Fixes #

**Documentation**:

```release-note
None
```


xrstf (approved):
/approve

PR #10339, opened 2022-07-14
Refactor conformance/CCM tests to use the KKP datacenters

xrstf:
At some point we stopped actually making use of the infrastructure that we already have. New fields for providers were just randomly hardcoded into our scripts (like `-vsphere-datastore=HS-FreeNAS`) or even the codebase (like was the case for the CCM tests). This is a) redundant and b) makes it super duper annoying to update any of these settings.

To improve the situation, this PR refactors the conformance and CCM tests to make use of the KKP datacenters (i.e. what we configure in `hack/ci/testdata/seed.yaml`). This is the same logic that #10266 introduced for Cilium/MLA/OPA etc. tests: Specifiy the KKP datacenter from the command line and then use this information to auto-determine all the cloud provider settings.

In a future PR we then might make the Seeds (in hack/ci/testdata/) more dynamic with more data straight from Vault.

```release-note
NONE
```

### DIFF
--- a/.prow/ccm-migrations.yaml
+++ b/.prow/ccm-migrations.yaml
@@ -55,7 +55,6 @@ presubmits:
               memory: 4Gi
 
   - name: pre-kubermatic-ccm-migration-vsphere-e2e
-    run_if_changed: "(pkg/|.prow/)"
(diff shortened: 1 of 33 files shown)


moadqassem (approved):
/approve 
/lgtm

moadqassem:
@xrstf You need to rebase though ;)

moadqassem (approved):
/approve
/lgtm

PR #10355, opened 2022-07-15
WIP - include resource usage in cluster output

xrstf:

```release-note
Include resource usage in `kubectl get clusters` output
```


xrstf:
Won't work as CRDs may only use _simple_ JSON path expressions:

> Invalid value: "{.status.resourceUsage.cpu} vCPU / {.status.resourceUsage.memory}": must be a simple json path starting with ..

See code here: https://github.com/kubernetes/kubernetes/blob/874d4bff431b601f18777c1968ae7c0ef00bf9c5/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L1372

PR #10356, opened 2022-07-15
fix vmware cloud director storage calculation

lsviben:
fixes vmware cloud director storage calculation

```release-note
NONE
```


xrstf (approved):
/approve

PR #10359, opened 2022-07-15
Block Cluster object from being deleted before the namespace is gone

xrstf:
This PR changes the order in which we tear down a usercluster.

## Previously

Before this PR, it was possible (and normal) for a Cluster object to be gone _before_ the cluster namespace is gone. This has caused some controllers to have special conditions where during cleanup the "Cluster not found" situation has to be anticipated.

The major downside to this however is that we cannot use the Cluster to collect nice, helpful events during cluster deletion. And I plan on making use of more events to give users more insights as to why their clusters might be stuck in deletion.

### Cluster Deletion

Deleting a cluster in KKP involves a number of controllers, but 2 of them are by far the most important:

* The `kubernetes` controller, responsible for nearly all resources on the seed (when it comes to cleanup at least).
* The `cloud` controller, managing the cloudprovider.

There are of course a whole number of finalizers (like cleanup-etcdbackup-configs), so quite a few more controllers are involved in the cleanup, but they only play a minor role for understanding the big picture.

So this is how it worked in the old implementation:

1. `kubectl delete cluster xy123` marks the Cluster object for deletion.
1. A whole number of controllers start to perform their cleanup.
   * The `kubernetes` controller begins, using `pkg/clusterdeletion`, to delete in-cluster LoadBalancers/PVCs (so we don't leave expensive stuff around). This is a waiting loop until all LBs/PVCs are completely gone.
   * The `cloud` controller is at this point still *waiting* and doing nothing. It waits until the `delete-nodes` finalizer is gone.
1. Still waiting for the LBs/PVCs to be gone...
1. Still waiting for the LBs/PVCs to be gone...
1. Still waiting for the LBs/PVCs to be gone...
1. Hooray, LBs/PVCs are gone!
   * Now the `kubernetes` controller will begin to destroy nodes/Machines. This is once again a wait loop until all nodes/Machines are gone.
   * The `cloud` controller is still waiting.
   * The `addon` controller does nothing, as the `Addon` objects in the cluster namespace are not yet marked for deletion (because the namespace is not yet terminating).
1. Still waiting for nodes to be gone...
1. Still waiting for nodes to be gone...
1. Nodes are gone!
   * Now the `kubernetes` controller stops and waits for all remaining finalizers except `cleanup-credential-secret` to be gone (i.e. all other cleanups should be finished).
   * The `cloud` controller finally begins to work and tells the cloudprovider to please cleanup any infrastructure. Now we're finally calling `provider.CleanUpProvider()`. This is a wait loop until the cloudprovider is done.
1. Still waiting for the cloudprovider cleanup...
1. Still waiting for the cloudprovider cleanup...
1. Cloudprovider is done!
   * This is usually the last finalizer left, so now the `kubernetes` controller sees that only one (`cleanup-credential-secret`) is left and so it beings to delete the credential Secret.
   * The `cloud` controller is not involved anymore.
1. Now the credential Secret is gone. This was the last finalizer on the Cluster object, so now that finalizer is removed and the Cluster object is finally actually removed from etcd.
1. Due to the ownerRef, the deleted Cluster will now trigger the namespace deletion. The namespace will now in turn delete all resources in it.
1. Now all Addons have been marked for deletion (because the cluster namespace is terminating). Now the `addon` controller begins its cleanup. And it encounters the "OMG, cluster is gone already?!" situation and handles it by simply removing all finalizers from an Addon and letting it be cleared up. **No `kubectl delete -f addonmanifests.yaml` is happening in this case!**
1. Once all objects in the cluster namespace are gone, the namespace itself is gone.
1. Finished.

### Addons?

Addons are special. At first I wanted to make it so that we cleanly remove addons when a cluster is deleted. Just because it "felt right" to tear down everything properly. However, addons are KKP's mechanism to install CNI/CSI and potentially other critical cluster components. Deleting those addons will destroy the cluster.

And this would lead to the question, when to remove addons? Before the cloud provider cleanup is done (maybe the addons need some cloud provider infrastructure to be able to be cleaned up?) or after (because without the addons, the cluster would be defunct)?

Neither of them were the right answer though. Even when we add a dedicated finalizer on Cluster objects (like `kubermatic.k8c.io/cleanup-addons`), just attempting to cleanly delete an Addon is futile and pointless:

* The cluster is gone in a few seconds/minutes anyway.
* All expensive resources (LBs, PVCs) are deleted already in the clusterdeletion (so the clusterdeletion kind of "breaks" the Addons).

Instead of risking dadlocks when removing the CNI/CSI addons, we simply ignore addons during cluster deletion. The old implementation (described above) would use "Cluster does not exist anymore" as the signal for "do not do a kubectl-delete, just remove the Addon's finalizer and forget it.". In the new implementation (described below), we use the fact that the Cluster has a deletion timestamp for this signal (i.e. once a Cluster is being deleted, the addon controller will never ever use kubectl to reconcile Addons).

TL;DR: Nothing really changed here, Addons are discarded in the same way as we did before.

## New Approach

This PR adds a new `cleanup-namespace` finalizer on Cluster objects. This will ensure that the object is not deleted before its namespace is gone (**gone**, not just terminating).

The cluster deletion procedure stays mostly the same. Just imagine the old procedure is identical up until the point when cloud provider cleanup is happening. Then:

1. Still waiting for the cloudprovider cleanup...
1. Still waiting for the cloudprovider cleanup...
1. Cloudprovider is done!
   * **New:** Now instead of only the `cleanup-credential-secret` finalizer being left, we also have the `cleanup-namespace` finalizer on the Cluster object.
   * The `cloud` controller is not involved anymore.
1. The `kubernetes` controller now, if only the 2 finalizers mentioned above are remaining, deletes the cluster namespace.
1. Now all Addons have been marked for deletion (because the cluster namespace is terminating). Now the `addon` controller begins its cleanup. Just like in the old implementation, no actual kubectl-delete is happening.
1. Once all Addon objects are gone, Kubernetes will delete the cluster namespace.
1. The `kubernetes` controller notices this and removes the `cleanup-namespace` finalizer.
1. `cleanup-credential-secret` is the last remaining finalizer and is now cleanup up the same way as in the old implementation.
1. The last finalizer is removed and the Cluster object is GC'ed.
1. Finished!

So there aren't that many differences. It's mostly that the namespace is explicitly deleted by us at a certain point in time, rather than letting the owner ref take care of that at a later point in time.

## Events! So many Events!

Now that we have the Cluster object around all the time, we can nicely (mis)use it and the event mechanism to emit nice looking events, like so:

screenshot-2022-07-17-170444

<del>(Notice the large number of namespace events? This is because the kubernetes controller is triggered by many changes that happen inside the cluster namespace, and so instead of checking this status every 10 seconds, we do it much more often; we might want to change the controller to discard changes in the cluster namespace if the cluster is in deletion.)</del> The large number of events was fixed already since the screenshot was taken.

Note that there are quite a few events being emitted and Kubernetes by default does not like that. After looking around for hours and hours, I found the reason why: [spam filtering](https://github.com/kubernetes/kubernetes/issues/97367#issuecomment-766783470). However since we rarely emit events in KKP and I think it's legitimate to emit a few more than usual during the cluster deletion, I have added this custom broadcaster.

## Misc

There are also a few improvements overall to the cluster deletion package. Like we do not attempt to re-Delete LBs every 10 seconds anymore, logging has been improved and the ancient `Use DeleteCollection` TODO has finally been resolved.

## Future Work

I patched the AWS cloud provider to not run the cleanup everytime, because the cloud controller will run this function quite often during cluster deletions. In a perfect world, we would not have to do this per provider, but instead use a generic `cleanup-cloud-provider` finalizer (instead of `cleanup-aws`, `cleanup-hetzner`, `cleanup-...`).

Having a dedicated, generic finalizer would allow the cloud controller more control over the cleanup, frees providers from dealing with the finalizer and makes it easier to prevent calling CleanUpProvider() when there is no finalizer.

Maybe in a future PR we want to change our approach here.

## Release Note

```release-note
Cluster object is kept around until the cluster namespace has been entirely removed from etcd.
Improve cluster deletion by emitting events to make it easier to disagnose stuck clusters.
```


pkg/clusterdeletion/deletion.go
```diff
@@ -82,72 +87,91 @@ func (d *Deletion) CleanupCluster(ctx context.Context, log *zap.SugaredLogger, c
...
+		if err != nil {
+			return fmt.Errorf("failed to check if all Loadbalancers are gone: %w", err)
+		}
+		// Return so we check again later
+		if !lbsAreGone {
+			return nil
+		}
 	}
 
+	d.recorder.Event(cluster, corev1.EventTypeNormal, "ClusterCleanup", "Cleanup of volumes/LoadBalancers has been completed.")
```
embik: Uber nit: The other event uses `PersistentVolumeClaims`, so the API resource type. Can we also adjust this to something like

```suggestion
	d.recorder.Event(cluster, corev1.EventTypeNormal, "ClusterCleanup", "Cleanup of LoadBalancers / PersistentVolumes has been completed.")
```

So the two events are more consistent?

pkg/clusterdeletion/node.go
```diff
@@ -63,51 +67,53 @@ func (d *Deletion) cleanupNodes(ctx context.Context, cluster *kubermaticv1.Clust
...
 	}
+
 	if len(machineDeploymentList.Items) > 0 {
-		// TODO: Use DeleteCollection once https://github.com/kubernetes-sigs/controller-runtime/issues/344 is resolved
-		for _, machineDeployment := range machineDeploymentList.Items {
-			if err := userClusterClient.Delete(ctx, &machineDeployment); err != nil {
-				return fmt.Errorf("failed to delete MachineDeployment %q: %w", machineDeployment.Name, err)
-			}
+		if err = userClusterClient.DeleteAllOf(ctx, &clusterv1alpha1.MachineDeployment{}, listOpts); err != nil {
+			return fmt.Errorf("failed to list MachineDeployments: %w", err)
```
embik: Should be

```suggestion
			return fmt.Errorf("failed to delete MachineDeployments: %w", err)
```

pkg/clusterdeletion/node.go
```diff
@@ -63,51 +67,53 @@ func (d *Deletion) cleanupNodes(ctx context.Context, cluster *kubermaticv1.Clust
...
 	}
+
 	if len(machineSetList.Items) > 0 {
-		// TODO: Use DeleteCollection once https://github.com/kubernetes-sigs/controller-runtime/issues/344 is resolved
-		for _, machineSet := range machineSetList.Items {
-			if err := userClusterClient.Delete(ctx, &machineSet); err != nil {
-				return fmt.Errorf("failed to delete MachineSet %q: %w", machineSet.Name, err)
-			}
+		if err = userClusterClient.DeleteAllOf(ctx, &clusterv1alpha1.MachineSet{}, listOpts); err != nil {
+			return fmt.Errorf("failed to list MachineSets: %w", err)
```
embik:
```suggestion
			return fmt.Errorf("failed to delete MachineSets: %w", err)
```

pkg/clusterdeletion/node.go
```diff
@@ -63,51 +67,53 @@ func (d *Deletion) cleanupNodes(ctx context.Context, cluster *kubermaticv1.Clust
...
 	}
+
 	if len(machineList.Items) > 0 {
-		// TODO: Use DeleteCollection once https://github.com/kubernetes-sigs/controller-runtime/issues/344 is resolved
-		for _, machine := range machineList.Items {
-			if err := userClusterClient.Delete(ctx, &machine); err != nil {
-				return fmt.Errorf("failed to delete Machine %q: %w", machine.Name, err)
-			}
+		if err = userClusterClient.DeleteAllOf(ctx, &clusterv1alpha1.Machine{}, listOpts); err != nil {
+			return fmt.Errorf("failed to list Machines: %w", err)
```
embik:
```suggestion
			return fmt.Errorf("failed to delete Machines: %w", err)
```

pkg/controller/seed-controller-manager/auto-update-controller/controller.go
```diff
@@ -103,6 +106,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		return reconcile.Result{}, err
 	}
 
+	// auto-updates should be be applied during cluster deletion
```
embik: Does this mean "should not be applied during cluster deletion"?

embik (changes requested):
Minor nits, looks good overall.

pkg/controller/seed-controller-manager/auto-update-controller/controller.go
xrstf: certainly not be be, be not be the right answer, but be be be wrong entirely.

xrstf: According all known laws of aviation, a bee should not....

embik (approved):
/approve

ahmedwaleedmalik (approved):
/approve 🙈

PR #10362, opened 2022-07-17
Make the KubeAwareEncoder less verbose

xrstf:
This PR does two things (that's a great start, isn't it?):

* It adds a custom Zap encoder that uses `{"obj":"namespace/name"}` instead of `{"obj":{"namespace":"namespace","name":"name"}}`, just better readability. This goes a bit against controller-runtime, but follows Kubernetes upstream.
* While going through our controllers to harmonize the logging code a bit, thanks to the new encoder, I noticed that the `master-constraint-template-synchonizer` is not using the seed-lifecycle-controller (and thereby re-invents getting seed clients). I rewrote it to make it work like its little sister, the `master-constraint-synchronizer`.
* And since I saw a lot of syncAllSeeds() functions (nearly all controllers in the master-ctrl-mgr had such a function), I moved it to the Kubernetes helpers as `SeedClientMap.Each()`. This just gets rid of a lot of redundant code.

```release-note
NONE
```

### DIFF
--- a/cmd/master-controller-manager/controllers.go
+++ b/cmd/master-controller-manager/controllers.go
@@ -65,7 +65,8 @@ func createAllControllers(ctrlCtx *controllerContext) error {
 	)
 	projectLabelSynchronizerFactory := projectLabelSynchronizerFactoryCreator(ctrlCtx)
 	userSSHKeySynchronizerFactory := userSSHKeySynchronizerFactoryCreator(ctrlCtx)
-	masterconstraintSynchronizerFactory := masterconstraintSynchronizerFactoryCreator(ctrlCtx)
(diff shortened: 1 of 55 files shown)


xrstf:
Impossible because Zap will not encoder context values with the Encoder. Which to me doesn't make a lick of sense. I.e.

```
log.With("foo", f).Info("bla")
log.Infow("bla", "foo", f)
```

will NOT produce the same output, as in the first line, the KubeAwareEncoder is not used by Zap. See https://github.com/kubernetes-sigs/controller-runtime/pull/1883

Controller-runtime's solution to this is to basically add layers of wrapping around Zap and handle the encoding before anything is even passed to Zap. Unfortunately this means rewriting EVERYTHING to logr.Logger, which is a terrible interface and certainly not worth the effort.

Issue #8375, opened 2021-12-06
Remove Applications using helm method


SimonTheLeg:
**Acceptance Criteria**
* helm chart is removed
* target namespace is deleted

Issue #8374, opened 2021-12-06
Install Applications using helm method


SimonTheLeg:
**Acceptance Criteria**
* helm chart gets correctly installed / updated -> `helm upgrade --install`
* status of helm releases is tracked in ApplicationInstallationCR status

PR #10363, opened 2022-07-18
Application: helm source and template

### LINKED ISSUE #9775: Fetch Helm source without credentials

vgramer:
This pr implements:
* the download of the application's source from a helm repository (without authentication)
* the installation of an application using Helm

The `ApplicationInstallationStatus` has been extended with a new field containing the information about the helm Release.

The methods `apply` and `delete` from the Installer (`pkg/applications/installer.go`) now return a `statusUpdater`. It a function that updates the applicationInstallation 's status. Using this function allow to postpone the update of the status in the controller and keep updates to the CR in the same place.

Fixes  #9775 #8374 #8375 

**Documentation**:

```release-note
NONE
```


pkg/apis/apps.kubermatic/v1/application_installation.go
```diff
@@ -129,6 +130,42 @@ type ApplicationInstallationStatus struct {
 
 	// ApplicationVersion contains information installing / removing application
 	ApplicationVersion *ApplicationVersion `json:"applicationVersion,omitempty"`
+
+	// HelmRelease holds the information about the helm release installed by this application. This field is only fill if template method is 'helm'.
```
SimonTheLeg:
```suggestion
// HelmRelease holds the information about the helm release installed by this application. This field is only fill if template method is 'helm'.
// HelmRelease holds the information about the helm release installed by this application. This field is only filled if template method is 'helm'.
```

pkg/applications/providers/util/util.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+
+	"go.uber.org/zap"
+
+	appskubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/apps.kubermatic/v1"
+)
+
+// CreateHelmTempDir creates a temporary directory inside cacheDir where helm caches will be download.
+func CreateHelmTempDir(cacheDir string, applicationInstallation *appskubermaticv1.ApplicationInstallation) (string, error) {
+	downloadDest, err := os.MkdirTemp(cacheDir, "helm-"+applicationInstallation.Namespace+"-"+applicationInstallation.Name+"-")
```
SimonTheLeg: is there a reason for the last "-"?

pkg/applications/providers/source/helm.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// HelmSource downloads Helm chart from Helm HTTP or OCI registry.
+type HelmSource struct {
+	Ctx context.Context
+	// Kubeconfig of the user-cluster.
+	Kubeconfig              string
+	CacheDir                string
+	Log                     *zap.SugaredLogger
+	ApplicationInstallation *appskubermaticv1.ApplicationInstallation
```
SimonTheLeg: Like I mentioned in the call: Let's check if the HelmSource can be decoupled from the ApplicationInstallation. But like already mentioned, this is really only a nitpick

pkg/applications/installer.go
```diff
@@ -36,57 +37,73 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize template provider: %w", err)
+	}
+
+	statuUpdater, err := templateProvider.Uninstall(applicationInstallation)
+	if err != nil {
+		return nil, err
+	}
+
+	return statuUpdater, a.deleteNamespace(ctx, log, applicationInstallation, userClient)
```
SimonTheLeg: I am not sure if I am a fan of this. Since we named the corresponding flag `create`, I fear that users might accidentally delete their namespaces. 

Since we control the `kubermatic.k8c.io/cleanup-application-installation` finalizer: Could we maybe build this in a way that it reads an annotation `delete-namespace` to decide what to do?

And then in the UI we would display to the user a small checkbox `delete namespace <namespace> with my application` when they press the delete application button.
And if this is checked the api's kubernetesclient simply sets the annotation prior to running the deletion?

pkg/applications/providers/util/util.go
```diff
@@ -0,0 +1,45 @@
+/*
```
SimonTheLeg, line 1: Overall I thinking if this needs to be its own util package.

I am saying this because:
a) 2 of 3 util-funcs are directly related to helm and I think should be part of the the helm providers (I would also be fine with code-duplication if they are needed in both Template and Sourceprovider
b) StatusUpdate func in my opinion could also become part of providers directly

But this is subjective I think, since we don't have a unique coding-style within the project. My comment is more meant as food-for-thought :)

SimonTheLeg (commented):
some smaller topics from my side and the deleteNamespace question

SimonTheLeg (commented):
some smaller topics from my side and the deleteNamespace question

pkg/applications/providers/util/util.go
vgramer: it's to generate a name like `helm-<namespace>-<app_name>-<rand_number>`

example `default-app2-2818157720`

but i can change it if want, I don't have a strong opinion about it

pkg/applications/installer.go
vgramer: I prefer to rename the field to `managed` and keep the entire desired state in the CRD.  The finalizer a meant to wait for the deletion of an external resource before deleting the crd, allowing retries if the deletion of this external resource failed.

pkg/applications/providers/util/util.go
vgramer: i initially put it in providers package but it create a cycle dependency between provider and template packages

pkg/applications/installer.go
SimonTheLeg: I think renaming the field is not enough to prevent something like this, which I find very realistic for a beginner user
1. create an app in a namespace
2. install other things in their namespace
3. uninstall the app
4. accidentally destroy the other things in their namespace.

vgramer: fair enough

pkg/applications/providers/source/helm.go
vgramer: done

SimonTheLeg (approved):

PR #10370, opened 2022-07-18
refactor master-constraint-template-synchronizer to use seed-lifecycle-controller

xrstf:
This PR refactors one of the last few controllers in the master-ctrl-mgr to use the seed-lifecycle.

I also added `kuberneteshelper.SeedClientMap` with an `Each` function in order to get rid of the maaaaany `syncAllSeeds()` functions throughout the controllers.

```release-note
NONE
```

### DIFF
--- a/pkg/kubernetes/helper.go
+++ b/pkg/kubernetes/helper.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	providerconfig "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
+	"go.uber.org/zap"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
 	kubermaticlog "k8c.io/kubermatic/v2/pkg/log"
 	"k8c.io/kubermatic/v2/pkg/provider"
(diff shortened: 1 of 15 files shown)


pkg/kubernetes/helper.go
```diff
@@ -395,3 +397,18 @@ func EnsureLabels(o metav1.Object, toEnsure map[string]string) {
 	}
 	o.SetLabels(labels)
 }
+
+type SeedClientMap map[string]ctrlruntimeclient.Client
+
+type SeedVisitorFunc func(seedName string, seedClient ctrlruntimeclient.Client, log *zap.SugaredLogger) error
```
lsviben, line 408: is the seedName used anywhere? I see SeedVisitorFunc used like
`r.seedClients.Each(ctx, log, func(_ string, seedClient ctrlruntimeclient.Client, log *zap.SugaredLogger)`  everywhere. But maybe i am missing it?

xrstf: Not yet, but I can imagine the name being useful and I then do not want to refactor the entire codebase again. 😁

lsviben: ok makes sense

lsviben:
Nice cleanup. Feels good :relieved: 

/approve

Ill add the lgtm when the conflicts are solved

PR #10371, opened 2022-07-18
Dual-stack e2e: test host network IPs for DigitalOcean, OpenStack

PratikDeoghare:
Tests host network IPs for DigitalOcean and Openstack. 

Fixes #

**Documentation**:

```release-note
NONE
```


hack/ci/testdata/kubermatic_dualstack.yaml
```diff
@@ -24,6 +24,8 @@ spec:
   imagePullSecret: '__IMAGE_PULL_SECRET__'
   userCluster:
     apiserverReplicas: 2
+    machineController:
+      imageTag: "32c8ec0998445a664938d2f7783a9d12168adf7b"
```
rastislavs, line 28: Maybe mark this as TODO? It can be removed later after MC version is updated.

PratikDeoghare: Yes. It might just get forgotten tho. 

I will create reminder in the sig-networking for next week. If we have new release then update otherwise snooze one more week. What do you say?

rastislavs: yeah week or two :)

embik (approved):
/approve

assuming the openstack-e2e tests succeed. I do have to say that I'm not a huge fan of implementing e2e tests that won't succeed on a "vanilla" KKP `master` installation, but hopefully we can remove this hack soon. Since it's limited to the dualstack tests and sig-networking is okay with it, it should be fine.

PR #10373, opened 2022-07-18
Resource quota adjust API units

### LINKED ISSUE #10372: Adjust ResourceQuota API units
Currently the API uses `resource.Quantity` as the values of CPU, Memory(RAM) and Storage(Disk).

This forces the UI to use text fields for quota values, and is cumbersome to manage. 

Instead, set a unit scale for the values and make the API simpler to use.

CPU: no unit
RAM: GB
Disk: GB

lsviben:

Changes the ResourceQuota API model to have simpler and set units for CPU, RAM and Disk size.

CPU: no unit
RAM: GB
Disk: GB

This will help the UI with working with quantities and make the API simpler. The values get converted to and from the CRD values which are still in `resource.Quantity`

Fixes #10372 

```release-note
NONE
```


pkg/ee/resource-quota/handler.go
```diff
@@ -282,3 +295,50 @@ func DeleteResourceQuota(ctx context.Context, request interface{}, provider prov
 	}
 	return nil
 }
+
+func convertToAPIQuota(resourceDetails kubermaticv1.ResourceDetails) apiv2.Quota {
```
pkprzekwas, line 299: I suggest adding some unit tests to cover cases like `cpu=500m` and rounding logic.

ahmedwaleedmalik: @pkprzekwas +1

lsviben: Good idea, ill add some test cases.

Also about the `cpu=500m` we actually dont have anywhere support for cpu values under 1 in resource calculation. So that needs to be changed. I do that in a different PR.

ahmedwaleedmalik:
```
CPU: no unit
RAM: GB
Disk: GB
```

Should RAM aka MEMORY, have the units GB or MB? In general, we use MB for memory everywhere. I just want us to sort out these things related to units once and for all. So that we have a decision and we move forward.

pkg/ee/resource-quota/handler.go
lsviben: added some coversion to the current test case. Its not ideal, but to test better i would need to refactor the test to not have static resource quotas. Also I would like to avoid exporting this function just for tests. But feel free to change my mind, im not 100% about this.

Talha-Jamil-TJ (approved):
/lgtm

PR #10377, opened 2022-07-18
Group support for external user

pkprzekwas:
Dashboard uses `/me` websocket in order to gather information about current user. We forgot to modify this endpoint after introducing groups and this PR tries to address this issue.

```release-note
NONE
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -558,7 +558,9 @@ type User struct {
 	LastSeen *Time `json:"lastSeen,omitempty"`
 }
 
-func ConvertInternalUserToExternal(internalUser *kubermaticv1.User, includeSettings bool, bindings ...*kubermaticv1.UserProjectBinding) *User {
+func ConvertInternalUserToExternal(internalUser *kubermaticv1.User, includeSettings bool,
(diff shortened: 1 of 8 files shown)


pkg/api/v1/types.go
```diff
@@ -599,6 +602,22 @@ func ConvertInternalUserToExternal(internalUser *kubermaticv1.User, includeSetti
 		}
 	}
 
+	for _, binding := range groupBindings {
+		bindingAlreadyExists := false
+		for _, p := range apiUser.Projects {
+			if p.ID == binding.Spec.ProjectID {
+				bindingAlreadyExists = true
```
lsviben, line 609: we dont cover here the case where the user has a different role as a member of a group then from the user-binding.

IMO we should send back the most permissible role. WDYT?

pkprzekwas: You are right. My idea was to always take role from user binding as it's more specific but more permissive role makes more seanse.

pkprzekwas: Any idea how to handle `projectmanagers` case? Is `owners` role more permissive?

pkprzekwas: I guess we must stick to userBinding priority for the time being. Later on we should switch to returning a set of roles assigned for particular project. This change needs to be introduced in paralel with Dashboard.

embik (approved):

PR #10381, opened 2022-07-18
Move finalizers out of API-only package

xrstf:
`pkg/api/` represents the types used in the KKP REST API, not to be confused with `pkg/apis/`, which is for the Kube API (i.e. our CRDs).

It therefore doesn't make sense to put finalizers that are "owned" by controllers into an API package.

Going further, it's questionable whether finalizers need to be even public. In most cases the finalizers are set and removed by a single controller. This turns them into an implementation detail and we should needlessly pretend like the finalizers are a public interface and that it would be by default okay for a controller to use another controller's finalizer.

So I moved all 1-time-use finalizers into the controller packages where they are handled. The remaining few finalizers, mostly those used in the `clusterdeletion` package, have been moved to `pkg/apis/kubermatic/v1/common.go`.

```release-note
NONE
```


embik (approved):
/approve

ahmedwaleedmalik (approved):
/approve

PR #10383, opened 2022-07-18
[dev] add worker-name controller

xrstf:
This is adding a new controller to make working on a shared development system a bit easier.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
+++ b/pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
@@ -99,10 +99,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 			reason = "no reason given"
 		}
 
-		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ClusterPaused", "Cluster is paused: %s", reason)
-
-		// renew the event to keep it visible
-		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ClusterPaused", "Cluster cannot be cleaned up because it is paused: %s", reason)
 	}
 
 	// no worker name
@@ -110,13 +107,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		// cluster seems stuck (we wait for a bit because we cannot easily
 		// tell if a seed-ctrl-mgr with the given worker-name is actually
 		// up and running right now)
-		if time.Since(cluster.DeletionTimestamp.Time) > 10*time.Minute {
+		if time.Since(cluster.DeletionTimestamp.Time) > 5*time.Minute {
 			r.recorder.Eventf(cluster, corev1.EventTypeWarning, "WorkerName", "A %s label is set, preventing the regular seed-controller-manager from cleaning up.", kubermaticv1.WorkerNameLabelKey)
 		}
-
-		// renew the event to keep it visible
-		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
-	return reconcile.Result{}, nil
+	// renew the event to keep it visible
+	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 }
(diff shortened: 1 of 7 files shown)


embik (approved):

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
```diff
@@ -0,0 +1,122 @@
...
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName = "kkp-worker-name-controller"
```
xmudrii: This should probably be changed to the new controller name (`kkp-cluster-stuck-controller`).

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
```diff
@@ -0,0 +1,122 @@
...
+
+type Reconciler struct {
+	ctrlruntimeclient.Client
+
+	workerName string
+	recorder   record.EventRecorder
+	log        *zap.SugaredLogger
+}
+
+// Add creates a new worker-name controller.
```
xmudrii: Nit: same here.

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
```diff
@@ -0,0 +1,122 @@
...
+		// renew the event to keep it visible
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	}
+
+	// no worker name
+	if cluster.Labels[kubermaticv1.WorkerNameLabelKey] != "" {
+		// cluster seems stuck (we wait for a bit because we cannot easily
+		// tell if a seed-ctrl-mgr with the given worker-name is actually
+		// up and running right now)
+		if time.Since(cluster.DeletionTimestamp.Time) > 10*time.Minute {
```
xmudrii: I would reduce this, even after like 5 minutes it's strange that the cluster is still here.

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
```diff
@@ -0,0 +1,122 @@
...
+	if cluster.Spec.Pause {
+		reason := cluster.Spec.PauseReason
+		if reason == "" {
+			reason = "no reason given"
+		}
+
+		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ClusterPaused", "Cluster is paused: %s", reason)
+
+		// renew the event to keep it visible
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
xmudrii: Should we return here? Maybe the second case is also true, so we could emit both events.

xmudrii (commented):
A few minor nits, otherwise LGTM.

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
xrstf: ouh nice thinking!

pkg/controller/seed-controller-manager/cluster-stuck-controller/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// paused clusters will not be removed
+	if cluster.Spec.Pause {
+		reason := cluster.Spec.PauseReason
+		if reason == "" {
+			reason = "no reason given"
+		}
+
+		r.recorder.Eventf(cluster, corev1.EventTypeWarning, "ClusterPaused", "Cluster is paused: %s", reason)
```
xmudrii: Nit: maybe we could also mention that the cluster cannot be deleted because of this (like as we do for `worker-name`).

xmudrii (commented):
Just one more tiny nit.

xmudrii (approved):
/lgtm
/approve

PR #10385, opened 2022-07-19
ipam-controller: Skip clusters in deletion process

embik:
The `kkp-ipam-controller` already skips `Cluster` objects with unset namespace, but I observed it trying to reconcile IPAM allocations during cluster deletion, presumably when the `IPAMAllocation` resource in the cluster namespace was already deleted by the cluster namespace cleanup process. Cleanup behaviour was recently changed in #10359, so this situation (cluster still exists but namespace is cleaned up) did not happen before. This adjusts the ipam-controller a tiny bit to account for that.

My understanding of the controller logic is that if the cluster is being deleted, we don't want to continue reconciling the cluster. There doesn't seem to be any cleanup logic in the controller as far as I can say.

Fixes #

**Documentation**:

```release-note
NONE
```


xrstf (approved):
/approve

PR #10392, opened 2022-07-19
improve cluster cleanup resource handling

xrstf:
In #10359 we changed the deletion behaviour. To reduce the number of times we needlessly reconcile a Cluster during its deletion, we made it so that changes to resources in its namespacec are ignored. This makes the clusterdeletion wait for the 10s Requeue interval and so significantly reduces the number of events that are being emitted during shutdown.

However, for (to me) unknown reasons, this special handling (with the namespaceMarker) lead to cache inconsistencies. Suddenly, during cluster creation we would get tons of events, mostly regarding Secrets:

screenshot-2022-07-19-134237

At first we suspected GKE issues, but the problem went away the moment we deployed the KKP commit right before the PR above. This also never broke in any e2e test. During more local testing it turned out that our namespace handling optimization was somehow causing these errors. It's still unclear to me why exactly, but the moment we change the handling to be a little bit more clever (and maybe workaround some uniqueness constraint in the worker queue?), the events went away and everything worked smoothly again.

So this PR is just much better. Instead of injecting a weird marker, we simply skip the reconciling request entirely if the cluster is in deletion. Fewer items in the queue, less work to be done. And it fixes the issue. Huzza!

```release-note
NONE
```

### DIFF
--- a/pkg/controller/seed-controller-manager/kubernetes/cluster_controller.go
+++ b/pkg/controller/seed-controller-manager/kubernetes/cluster_controller.go
@@ -58,11 +58,6 @@ import (
 
 const (
 	ControllerName = "kkp-kubernetes-controller"
-
-	// namespaceMarker is injected in reconcile.Request namespaces to signal
-	// that something in the cluster namespace changed (compared to the Cluster
-	// object itself).
-	namespaceMarker = "dummy-marker"
 )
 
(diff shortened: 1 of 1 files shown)


embik (approved):

xmudrii (approved):
/lgtm
/approve

PR #10402, opened 2022-07-19
update externalcluster response

### LINKED ISSUE #10391: Differentiate Imported and Created External Clusters

As a developer
I would like to Differentiate Imported and Created External Clusters using a label
in order to Differentiate between the two

### Solution details

- Requirement 1
- Requirement 2
- Requirement 3

imharshita:

- update "get external cluster md" response to return details to support KKP UI
- Adds a label `IsImported` to distinguish between imported and created external clusters

Fixes #10391

**Documentation**:

```release-note
none
```


pkg/api/v2/types.go
```diff
@@ -629,14 +669,14 @@ type AKSCloudSpec struct {
...
+	// The identity that created the resource.
+	CreatedBy *string `json:"createdBy,omitempty"`
 	// KubernetesVersion - When you upgrade a supported AKS cluster, Kubernetes minor versions cannot be skipped. All upgrades must be performed sequentially by major version number. For example, upgrades between 1.14.x -> 1.15.x or 1.15.x -> 1.16.x are allowed, however 1.14.x -> 1.16.x is not allowed. See [upgrading an AKS cluster](https://docs.microsoft.com/azure/aks/upgrade-cluster) for more details.
 	KubernetesVersion string `json:"kubernetesVersion"`
-	// NodeResourceGroup - The name of the resource group containing agent pool nodes.
-	NodeResourceGroup string `json:"nodeResourceGroup,omitempty"`
 	// EnableRBAC - Whether Kubernetes Role-Based Access Control Enabled.
 	EnableRBAC bool `json:"enableRBAC,omitempty"`
-	// ManagedAAD - Whether The Azure Active Directory configuration Enabled.
-	ManagedAAD bool `json:"managedAAD,omitempty"`
```
lsviben, line 639: why are we removing the ManagedAAD and NodeResourceGroup?

imharshita: i just removed it as it seemed to me like an non critical info and it can crowd the cluster view

lsviben: i see. This was not released in the previous release right? so we are not changing the API? If not, then ok we can  remove it :)

pkg/api/v2/types.go
```diff
@@ -603,17 +605,55 @@ type EKSClusterSpec struct {
...
+	Tags map[string]*string `json:"tags,omitempty"`
 
 	// The Amazon Resource Name (ARN) of the IAM role that provides permissions
 	// for the Kubernetes control plane to make calls to AWS API operations on your
 	// behalf. For more information, see Amazon EKS Service IAM Role (https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html)
 	// in the Amazon EKS User Guide .
 	//
 	// RoleArn is a required field
-	RoleArn string `json:"roleArn" required:"true"`
+	RoleArn string `json:"roleArn,omitempty" required:"true"`
```
pkprzekwas, line 630: Is it fine to have `omitempty` and required set to `true` at the same time?

imharshita:
> 

Yes, we can think about it. But for now there is no solid way to check if cluster is imported or created after the creation.
I can create a issue to discuss this

pkg/api/v2/types.go
imharshita: yeah, i know its confusing.
the reason is "required" for creation/POST request but `omitempty` for get response

pkg/api/v2/types.go
```diff
@@ -603,17 +605,55 @@ type EKSClusterSpec struct {
...
+}
+
+// The Kubernetes network configuration for the cluster. The response contains
+// a value for serviceIpv6Cidr or serviceIpv4Cidr, but not both.
+type KubernetesNetworkConfigResponse struct {
```
pkprzekwas: You mention Amazon CNI  several times in the doc and looks like only EKS Cluster Spec uses it. Is this type EKS specific? If so maybe we should relate that in the name.

pkg/api/v2/types.go
imharshita: wdyt? i can add a comment here to explain, or just remove the `omitempty`

pkg/handler/v2/external_cluster/eks.go
```diff
@@ -859,14 +871,22 @@ func getEKSClusterDetails(ctx context.Context, apiCluster *apiv2.ExternalCluster
...
+			IpFamily:        eksCluster.KubernetesNetworkConfig.IpFamily,
+			ServiceIpv4Cidr: eksCluster.KubernetesNetworkConfig.ServiceIpv4Cidr,
+			ServiceIpv6Cidr: eksCluster.KubernetesNetworkConfig.ServiceIpv6Cidr,
+		}
 	}
 
 	if eksCluster.ResourcesVpcConfig != nil {
 		clusterSpec.ResourcesVpcConfig = apiv2.VpcConfigRequest{
-			SecurityGroupIds: eksCluster.ResourcesVpcConfig.SecurityGroupIds,
-			SubnetIds:        eksCluster.ResourcesVpcConfig.SubnetIds,
```
pkprzekwas, line 869: We can add `SecurityGroupIds ` , it's not a secret.

PR #10411, opened 2022-07-20
Initialize the cloud spec for external clusters

moadqassem:
KKP admission webhooks started validating and defaulting external clusters, which led into denying requests for external cluster of type Any Cluster. The reason behind that is, the webhook simply rejects any cluster creation without having a provider name, however for external clusters of type Any, we don't a provider name thus we should leave it empty. This PR initialize the external cluster spec in order to avoid webhook rejection because of the cloud spec being declared as null.
 
Fixes #

**Documentation**:

```release-note
None
```

### DIFF
--- a/pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
+++ b/pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
@@ -114,7 +114,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.ExternalCluster) (reconcile.Result, error) {
 	cloud := cluster.Spec.CloudSpec
 	secretKeySelector := provider.SecretKeySelectorValueFuncFactory(ctx, r.Client)
-	if cloud == nil {
+	if cloud.ProviderName == "" {
 		return reconcile.Result{}, nil
 	}
 	if cloud.GKE != nil {
--- a/pkg/controller/master-controller-manager/kubeone/controller.go
+++ b/pkg/controller/master-controller-manager/kubeone/controller.go
@@ -186,7 +186,7 @@ func withEventFilter() predicate.Predicate {
 			if !ok {
 				return false
 			}
-			if externalCluster.Spec.CloudSpec == nil {
+			if externalCluster.Spec.CloudSpec.ProviderName == "" {
 				return false
 			}
 			return externalCluster.Spec.CloudSpec.KubeOne != nil
--- a/pkg/defaulting/externalcluster.go
+++ b/pkg/defaulting/externalcluster.go
@@ -28,6 +28,15 @@ import (
 // This function assumes that the KubermaticConfiguration has already been defaulted
 // (as the KubermaticConfigurationGetter does that automatically).
 func DefaultExternalClusterSpec(ctx context.Context, spec *kubermaticv1.ExternalClusterSpec) error {
+	// The import cluster feature doesn't expect any cloud spec as this imported cluster could be any type of
+	// Kubernetes cluster, thus there are no specs in regard of cloud nor infrastructure. Setting the cluster spec
+	// to nil would lead rejecting the cluster so we need to initialize it here.
+	if spec == nil {
+		spec = &kubermaticv1.ExternalClusterSpec{
+			CloudSpec: &kubermaticv1.ExternalClusterCloudSpec{},
+		}
+	}
+
 	// Ensure provider name matches the given spec
 	providerName, err := provider.ExternalClusterCloudProviderName(spec.CloudSpec)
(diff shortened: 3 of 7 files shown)


pkg/handler/v2/external_cluster/external_cluster.go
```diff
@@ -1038,7 +1039,7 @@ func convertClusterToAPIWithStatus(ctx context.Context, clusterProvider provider
 	apiCluster.Status = status
 	cloud := internalCluster.Spec.CloudSpec
 	kubeOneCondtion := internalCluster.Status.Condition
-	if cloud == nil {
+	if cloud == nil || cloud.ProviderName == "" {
```
imharshita, line 1042: plz change this to 
`if cloud != nil && cloud.ProviderName == ""`

imharshita (changes requested):

pkg/handler/v2/external_cluster/external_cluster.go
moadqassem: Why ? `cloud` will never actually equals to nil meaning that the second condition will always hit

pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
```diff
@@ -114,7 +114,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 func (r *Reconciler) reconcile(ctx context.Context, cluster *kubermaticv1.ExternalCluster) (reconcile.Result, error) {
 	cloud := cluster.Spec.CloudSpec
 	secretKeySelector := provider.SecretKeySelectorValueFuncFactory(ctx, r.Client)
-	if cloud == nil {
+	if cloud.ProviderName == "" {
```
lsviben, line 117:
```suggestion
	if cloud == nil || cloud.ProviderName == "" {
```

pkg/controller/master-controller-manager/kubeone/controller.go
```diff
@@ -186,7 +186,7 @@ func withEventFilter() predicate.Predicate {
 			if !ok {
 				return false
 			}
-			if externalCluster.Spec.CloudSpec == nil {
+			if externalCluster.Spec.CloudSpec.ProviderName == "" {
```
lsviben, line 189:
```suggestion
			if externalCluster.Spec.CloudSpec == nil || externalCluster.Spec.CloudSpec.ProviderName == "" {
```

pkg/handler/v2/external_cluster/external_cluster.go
imharshita: i know not possible to have cloud == nil if created through API
, but in cases user creates a external cluster crd manually in the cluster.
this will just eliminate the case where cloud == nil,
as now for `any provider` we will have cloud != nil and providerName == ""

moadqassem: Yeah, I forgot to push the changes for the defaulting webhook 😂

moadqassem: The one that I added during the huddle

pkg/controller/master-controller-manager/external-cluster/cluster_controller.go
imharshita: this should fix controller test failures

pkg/handler/v2/external_cluster/external_cluster.go
xrstf: Does CloudSpec need to be a pointer? If it's never supposed to be nil.

moadqassem: Until now it is not clear how to treat this, thus I would like to keep a pointer. Once there are more of a concrete idea of how to handle external cluster we can revisit this again.

pkg/defaulting/externalcluster.go
```diff
@@ -28,6 +28,15 @@ import (
 // This function assumes that the KubermaticConfiguration has already been defaulted
 // (as the KubermaticConfigurationGetter does that automatically).
 func DefaultExternalClusterSpec(ctx context.Context, spec *kubermaticv1.ExternalClusterSpec) error {
+	// The import cluster feature doesn't expect any cloud spec as this imported cluster could be any type of
+	// Kubernetes cluster, thus there are no specs in regard of cloud nor infrastructure. Setting the cluster spec
```
imharshita, line 32:
```suggestion
	// Kubernetes cluster, thus there are no specs in regard of cloud nor infrastructure. Setting the cloud spec
```

moadqassem: I am setting the cluster spec here not only the cloud spec. And the cloud spec is part if the cluster spec so this is quite inclusive. Also the check was, `if spec == nil`

PR #10414, opened 2022-07-20
Make IPAM e2e tests headless

xrstf:
This rewrites the IPAM tests to be headless. They now also use BYO instead of Hetzner, which is cheaper and faster.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡

PR #10423, opened 2022-07-20
Openstack: Fix inconsistent casing for floatingIPPool in cloud spec

### LINKED ISSUE #4581: Allow allocating floating IPs in exiting OpenStack user clusters
### What happened
For OpenStack user clusters it is not possible to request Floating IP by selecting "Allocate Floating IP" when editing an existing Machine Deployment or adding a new Machine deployment to en existing user cluster:

Editing an existing Machine Deployment:

---
Screenshot 2022-06-10 at 15 24 25
---

Adding a new Machine Deployment:

---
Screenshot 2022-06-10 at 15 28 42
---

In both cases, the checkbox is disabled and the hover tooltip says "Floating IP usage is enforced by selected datacenter" even if the datacenter has the following spec:

```
      spec:
        openstack:
          enforceFloatingIP: false
```

### Expected behavior
"Allocate Floating IP" checkbox should be disabled only when `enforceFloatingIP` is `true`.

### How to reproduce
Create a user cluster in OpenStack datacenter with `enforceFloatingIP: false`. Do not enable "Allocate Floating IP" during cluster creation. After the cluster is up, try editing the existing Machine Deployment or adding a new Machine Deployment - "Allocate Floating IP" will be disabled.

### Environment
Reported in KKP 2.20.3, reproduced in the latest master as well.

ahmedwaleedmalik:

There is a case mismatch between what the [API returns as cloud spec](https://github.com/kubermatic/kubermatic/blob/master/pkg/api/v1/types.go#L1132) and what is persisted in ETCD using the cluster CR https://github.com/kubermatic/kubermatic/blob/master/pkg/apis/kubermatic/v1/cluster.go#L1154.

This PR removes this inconsistency and fixes this bug https://github.com/kubermatic/dashboard/issues/4581

Fixes https://github.com/kubermatic/dashboard/issues/4581

Please refer to https://github.com/kubermatic/dashboard/issues/4581#issuecomment-1181689928

**Documentation**:

```release-note
BREAKING CHANGE: Fix inconsistent casing for floatingIPPool in cloud spec. This affects the endpoints used to list and get cluster; for OpenStack, the field `floatingIpPool` was replaced with `floatingIPPool`. Endpoints for creation and update are not affected.
```


Waseem826 (approved):

ahmedwaleedmalik:
/hold 

For @lsviben to review this as well ;)

ahmedwaleedmalik:
/retest

FLAKEYYYYYY

lsviben:
This change, although its a bug fix, breaks the current API that has been around for a while. 

We should at least inform in the release notes about it. 

Also we could avoid changing the API and just adjust the usage in the UI, but I am not convinced of that. WDYT?

ahmedwaleedmalik:
> This change, although its a bug fix, breaks the current API that has been around for a while.
> 
> We should at least inform in the release notes about it.
> 
> Also we could avoid changing the API and just adjust the usage in the UI, but I am not convinced of that. WDYT?

I think we usually don't add release notes for such changes in API. But I can do that.

> Also we could avoid changing the API and just adjust the usage in the UI, but I am not convinced of that. WDYT?

That is an option but I think it will add a bit more complexity in dealing with this API directly and on our front-end as well. I just don't like the fact that it'd add a "special case" for us.

ahmedwaleedmalik:
@lsviben release note added.

lsviben:
> @lsviben release note added.

thanks. Dont know if we need ACTION REQUIRED on that. But we should at least inform :)

PR #10431, opened 2022-07-21
Introduce role priorities

pkprzekwas:
As discussed on sig api-ui, replaced prioritisation of user project binding with most permissive role.

```release-note
NONE
```


pkg/api/v1/types.go
```diff
@@ -589,28 +597,23 @@ func ConvertInternalUserToExternal(internalUser *kubermaticv1.User, includeSetti
...
+			if pg.ID == binding.Spec.ProjectID {
+				// Check if role from group binding is more permissive.
+				if RolePriority[binding.Spec.Role] > RolePriority[pg.GroupPrefix] {
+					overwriteRole = true
+				}
 				break
 			}
 		}
-		if !bindingAlreadyExists {
+		if overwriteRole {
```
lsviben: I think now you are adding projects only if the role priority is higher. If a userprojectbinding does not exist, you are not adding a project through GPBs

pkprzekwas: Oh man, u r right. Let me fix this.

PR #10433, opened 2022-07-21
Remove human-readable name label from ResourceQuotas

pkprzekwas:
Unfortunately label keys are strict and cannot contain some of characters that we allow in human-readable project names.  

```release-note
NONE
```


pkg/ee/resource-quota/handler.go
```diff
@@ -207,9 +217,26 @@ func ListResourceQuotas(ctx context.Context, request interface{}, provider provi
 		return nil, err
 	}
 
+	// Fetching projects to get their human-readable names.
+	projectMap := make(map[string]*kubermaticv1.Project)
```
lsviben, line 221: to make it a bit lighter, could you make a map[string]string where the value is the human readable name?

PR #10442, opened 2022-07-22
migrate scripts to yq4

xrstf:
It's time to migrate to yq4. See https://mikefarah.gitbook.io/yq/upgrading-from-v3 for more information.

I chose to make it explicit what version we're using for now, in order to slowly get rid of yq3 over the next few PRs. It is a bit annoying that devs would need a  `yq4` binary on their machines, but I suspect many have v3 and v4 around, like I do.

```release-note
NONE
```

### DIFF
--- a/hack/ci/fetch-chart-dependencies.sh
+++ b/hack/ci/fetch-chart-dependencies.sh
@@ -32,7 +32,7 @@ charts=$(find charts/ -name Chart.yaml | sort)
   echodate "Fetching dependencies for ${chartname}..."
 
   i=0
-  for url in $(yq4 '.dependencies.[].repository' Chart.yaml); do
+  for url in $(yq4 '.dependencies.[].repository' "$chartYAML"); do
     i=$((i + 1))
     helm repo add ${chartname}-dep-${i} ${url}
   done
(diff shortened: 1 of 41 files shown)


embik (approved):

hack/ci/fetch-chart-dependencies.sh
```diff
@@ -28,11 +28,11 @@ charts=$(find charts/ -name Chart.yaml | sort)
 
 [ -n "$charts" ] && while read -r chartYAML; do
   dirname="$(dirname $(echo "$chartYAML"))"
-  chartname=$(yq read "$chartYAML" name)
+  chartname=$(yq4 '.name' "$chartYAML")
   echodate "Fetching dependencies for ${chartname}..."
 
   i=0
-  for url in $(yq r "$chartYAML" dependencies --tojson | jq -r .[].repository); do
+  for url in $(yq4 '.dependencies.[].repository' Chart.yaml); do
```
xmudrii: Shouldn't this be `$chartYAML`?

xmudrii (commented):
Just a tiny nit.

hack/ci/fetch-chart-dependencies.sh
xrstf: It should.

xmudrii (approved):
/lgtm
/approve

PR #10445, opened 2022-07-22
improve finalizer handling in cloud providers

xrstf:
This performs two slight improvements that I recently noticed in the AWS provider:

* The kubevirt provider now adds its finalizer before it creates the namespace, so that if adding the finalizer fails for some reason, we do not create an orphaned namespace that we might never clean up.
* The nutanix provider has been optimized slightly to make it more readable IMHO.

```release-note
NONE
```

### DIFF
--- a/pkg/provider/cloud/nutanix/provider.go
+++ b/pkg/provider/cloud/nutanix/provider.go
@@ -73,30 +73,25 @@ func (n *Nutanix) ReconcileCluster(ctx context.Context, cluster *kubermaticv1.Cl
 }
 
 func (n *Nutanix) CleanUpCloudProvider(ctx context.Context, cluster *kubermaticv1.Cluster, update provider.ClusterUpdater) (*kubermaticv1.Cluster, error) {
-	logger := n.log.With("cluster", cluster.Name)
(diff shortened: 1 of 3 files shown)


pkg/provider/cloud/nutanix/provider.go
```diff
@@ -73,30 +73,25 @@ func (n *Nutanix) ReconcileCluster(ctx context.Context, cluster *kubermaticv1.Cl
...
-		cluster, err = update(ctx, cluster.Name, func(cluster *kubermaticv1.Cluster) {
-			kuberneteshelper.RemoveFinalizer(cluster, categoryCleanupFinalizer)
-		})
-
-		if err != nil {
-			return nil, err
-		}
+	if err = deleteCategoryValues(ctx, client, cluster); err != nil {
+		return nil, err
 	}
```
embik, line 90: Why remove the check on the finalizer here? I usually prefer checking for that because we can save some calls to the cloud provider API.

embik (changes requested):

pkg/provider/cloud/nutanix/provider.go
embik: Ahh. I just noticed it was moved early. Nevermind.

embik (approved):

PR #10459, opened 2022-07-23
fix a lot of typos

xrstf:
https://github.com/crate-ci/typos found a lot of false positives, but also lots of actual typos. :-)

```release-note
NONE
```


pkg/apis/kubermatic/v1/cluster.go
```diff
@@ -421,7 +421,7 @@ const (
 	ClusterConditionAddonControllerReconcilingSuccess                   ClusterConditionType = "AddonControllerReconciledSuccessfully"
 	ClusterConditionAddonInstallerControllerReconcilingSuccess          ClusterConditionType = "AddonInstallerControllerReconciledSuccessfully"
 	ClusterConditionBackupControllerReconcilingSuccess                  ClusterConditionType = "BackupControllerReconciledSuccessfully"
-	ClusterConditionCloudControllerReconcilingSuccess                   ClusterConditionType = "CloudControllerReconcilledSuccessfully"
+	ClusterConditionCloudControllerReconcilingSuccess                   ClusterConditionType = "CloudControllerReconciledSuccessfully"
```
xmudrii, line 427: Aren't we going to need some sort of migration for this? What's going to happen with existing clusters when you upgrade?

xmudrii (approved):
/lgtm
/approve

pkg/apis/kubermatic/v1/cluster.go
xrstf: I added the misspelling back to the ENUM so that the CRD stays compatible. The cloud controller will now during its reconciliation simply set the new and remove the old condition, so that in KKP 2.22 we can remove the misspelling from the ENUM.

This will lead to a short moment in which Cluster objects seemingly do not have all necessary conditions to be considered "healthy", but this will sort itself out quickly.

ahmedwaleedmalik (approved):
/approve ⚡

PR #10465, opened 2022-07-25
application: move template method to spec level

### LINKED ISSUE #10463: Allow upgrade of application

As a user
I would like to update the application from one version to another
in order to use new version my application

### Solution details

In `ApplicationDefinition` each `ApplicationVersions` define a template method (`ApplicationDefinition.spec.versions[].template.method`) currently we only support helm as template method.
But in future releases, we may implement other template methods (e.g. `kustomize`, `kubectl` ...)
The question is how to support the upgrade of an `applicationInstallation`. (even if we say we’ll implement the upgrade in another release, it may have an impact on crd structure which is difficult to update)

If both versions use the same template method, there is no problem.
But if the versions use a different template method. In the best case, the upgrade fails, and in the worst case, it crew up the application workload.

Here is an example of such applicationDefinition (some fields have been stripped for clarity)
```
apiVersion: apps.kubermatic.k8c.io/v1
kind: ApplicationDefinition
metadata:
  name: apache
spec:
  description: apache server
  versions:
  - template:
    method: helm
    source:
     git:
      path: apache
      ref:
       tag: v9.0.13
      remote: https://gitlab.com/vgramer1/helm-charts
   version: 1
  - template:
    method: helm
    source:
     git:
      path: apache
      ref:
       tag: v10.0.0
      remote: https://gitlab.com/vgramer1/helm-charts
   version: 2
```

To avoid this problem we have to move the `template method` field to the spec level of `ApplicationDefinition` (i.e. `ApplicatinDefinition.spec.method`) and make it immutable. So every version of an `ApplicationDefinition`will share the same template method, and upgrades will be possible.

vgramer:
This pr  moves the field `ApplicationDefinition.spec.versions[].template.method` to `ApplicationDefinition.spec.method` and make it immutable.  It also makes the field `ApplicationDefinition.spec.versions[].version` mutable in order to allow update of an applicationInstallation.

Fixes #10463

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/cmd/kubermatic-api/swagger.json
+++ b/cmd/kubermatic-api/swagger.json
@@ -22810,6 +22810,9 @@
           "type": "string",
           "x-go-name": "Description"
         },
+        "method": {
+          "$ref": "#/definitions/TemplateMethod"
+        },
         "versions": {
           "description": "available version for this application",
           "type": "array",
@@ -22936,9 +22939,6 @@
             "$ref": "#/definitions/FormField"
           },
           "x-go-name": "FormSpec"
-        },
-        "method": {
-          "$ref": "#/definitions/TemplateMethod"
         },
         "source": {
           "$ref": "#/definitions/ApplicationSource"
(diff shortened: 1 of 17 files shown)


ahmedwaleedmalik (approved):
/approve ⚡ 

Great job 👍

ahmedwaleedmalik (approved):
/lgtm

PR #10473, opened 2022-07-25
allow users to not specify .spec.pause for Clusters

xrstf:
Previously, when creating a cluster from YAML, you had to specify `spec: { pause: false }`, which is just annoying. This PR makes the field optional, as it is defaulted to false anyway.

```release-note
`spec.pause` does not need to be set for Clusters anymore (defaults to `false`)
```


ahmedwaleedmalik (approved):
/approve

PR #10476, opened 2022-07-25
Move ResourceQuota mutation to controller

### LINKED ISSUE #10246: Move ownership handling for ResourceQuotas
### What happened?
Right now our webhook sets the owner ref for a ResourceQuota.

This must not happen, as the webhook might not find the referenced project (due to its own caches). Owner references must be set by _controllers_, not webhooks. See https://github.com/kubermatic/kubermatic/pull/9441#issuecomment-1082354554 and following comments.

### Expected behavior
There should be a controller (either a new one or an existing one) that manages the ownership between ResourceQuotas and projects.

lsviben:
Removes ResourceQuota mutation webhook and replaces it with a controller. 

The job of the webhook was to set the label and owner ref of the RQ subject to the RQ. As the subject may not exist when the RQ is being created, using a mutation webhook blocks the creation of the RQ. Instead, use a controller which will reconcile when everything is ready.

Fixes #10246 

```release-note
NONE
```


xrstf (approved):
/approve

PR #10481, opened 2022-07-25
WIP: API endpoint to List OSPs

ahmedwaleedmalik:

This PR does the following:
- Grant OSM service account access to get,list OSPs from `kubermatic` namespace
- Introduce `/api/v2/seeds/{seed_name}/operatingsystemprofiles` to list OSPs

Fixes #

**Documentation**:

```release-note
Add endpoint to list operating system profiles at the seed level.
```


pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -533,10 +534,20 @@ func (r *Reconciler) ensureRoles(ctx context.Context, c *kubermaticv1.Cluster) e
...
+	if c.Spec.EnableOperatingSystemManager {
+		namedRoleCreatorGetters := []reconciling.NamedRoleCreatorGetter{
+			operatingsystemmanager.KubermaticOperatingSystemProfileRoleCreator,
+		}
+
+		if err := reconciling.ReconcileRoles(ctx, namedRoleCreatorGetters, seedNamespace, r.Client); err != nil {
+			return fmt.Errorf("failed to ensure RoleBindings: %w", err)
+		}
+	}
+
```
ahmedwaleedmalik: @xrstf PTAL at this and let me know what you think

pkg/handler/v2/operatingsystemprofile/operatingsystemprofile.go
```diff
@@ -0,0 +1,121 @@
...
+	{
+		Name: "osp-rockylinux",
+	},
+	{
+		Name: "osp-sles",
+	},
+	{
+		Name: "osp-ubuntu",
+	},
+}
```
ahmedwaleedmalik, line 61: I hate this part and I thought quite a lot before doing this. But tbh I couldn't think of any better solution for this.

pkg/provider/kubernetes/operating_system_profiles.go
```diff
@@ -0,0 +1,70 @@
...
+	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// PrivilegedOperatingSystemProfileProvider struct that holds required components of the PrivilegedOperatingSystemProfileProvider.
+type PrivilegedOperatingSystemProfileProvider struct {
+	privilegedClient ctrlruntimeclient.Client
+	namespace        string
```
lsviben, line 33: the namespace is not used

ahmedwaleedmalik: Good catch, let me fix that.

lsviben:
I dont know OSM TBH, but for me it looks like you could avoid these changes for creating roles/rolebindings in seed namespace for every user cluster OSM if the OSP would be cluster-scoped. Is there some particular reason why its not?

ahmedwaleedmalik:
> I dont know OSM TBH, but for me it looks like you could avoid these changes for creating roles/rolebindings in seed namespace for every user cluster OSM if the OSP would be cluster-scoped. Is there some particular reason why its not?

I understand your concern. OSM is deployed in the user cluster namespace, just like machine-controller. And it creates the CRs(not CRDs) for OperatingSystemProfiles on its own, in the user cluster namespace.

If someone wants to use those default OSPs, the ones I have hardcoded, everything works fine. They don't even need anything from the API tbh since the MC webhooks will take care of this defaulting.

But one of the core features of OSM is to allow users to consume custom OSPs. These OSPs should exist before cluster creation so we can't use the user cluster namespace for that. This is the reason why we chose the seed namespace aka `kubermatic` namespace and then added this RBAC.

ahmedwaleedmalik:
/hold 

Some more changes are required for RBAC.

lsviben:
> > I dont know OSM TBH, but for me it looks like you could avoid these changes for creating roles/rolebindings in seed namespace for every user cluster OSM if the OSP would be cluster-scoped. Is there some particular reason why its not?
> 
> I understand your concern. OSM is deployed in the user cluster namespace, just like machine-controller. And it creates the CRs(not CRDs) for OperatingSystemProfiles on its own, in the user cluster namespace.
> 
> If someone wants to use those default OSPs, the ones I have hardcoded, everything works fine. They don't even need anything from the API tbh since the MC webhooks will take care of this defaulting.
> 
> But one of the core features of OSM is to allow users to consume custom OSPs. These OSPs should exist before cluster creation so we can't use the user cluster namespace for that. This is the reason why we chose the seed namespace aka `kubermatic` namespace and then added this RBAC.
> 
> Cluster scoped OSPs are not possible since users can override and use custom version of OSM, just like MC, for user clusters as well.

Ok I see, didnt know that there are also OSPs in the user-cluster-namespace. These in the Seed namespace(kubermatic) are "custom" ones. Understood

pkg/controller/seed-controller-manager/kubernetes/resources.go
```diff
@@ -533,10 +534,20 @@ func (r *Reconciler) ensureRoles(ctx context.Context, c *kubermaticv1.Cluster) e
 		return fmt.Errorf("failed to ensure Roles: %w", err)
 	}
 
+	if c.Spec.EnableOperatingSystemManager {
+		namedRoleCreatorGetters := []reconciling.NamedRoleCreatorGetter{
+			operatingsystemmanager.KubermaticOperatingSystemProfileRoleCreator,
+		}
+
+		if err := reconciling.ReconcileRoles(ctx, namedRoleCreatorGetters, seedNamespace, r.Client); err != nil {
+			return fmt.Errorf("failed to ensure RoleBindings: %w", err)
```
xrstf, line 543: s/Binding//

pkg/resources/operatingsystemmanager/rbac.go
```diff
@@ -24,9 +24,11 @@ import (
 )
 
 const (
-	serviceAccountName = "kubermatic-operating-system-manager"
-	roleName           = "kubermatic:operating-system-manager"
-	roleBindingName    = "kubermatic:operating-system-manager"
+	serviceAccountName                              = "kubermatic-operating-system-manager"
+	roleName                                        = "kubermatic:operating-system-manager"
+	roleBindingName                                 = "kubermatic:operating-system-manager"
+	kubermaticOperatingSystemProfileroleName        = "kubermatic:operating-system-profile-viewer"
```
xrstf, line 32: That const name is a bit redundant, maybe `ospViewerRoleName` would be enough?

ahmedwaleedmalik:
Closing this PR in favor of https://github.com/kubermatic/kubermatic/pull/10532

PR #10498, opened 2022-07-26
setup dualstack smoketest with azure+(canal, cilium)

PratikDeoghare:
This adds a required test that checks Azure dualstack because running all of the dualstack tests is expensive and optional. 

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow/features.yaml
+++ b/.prow/features.yaml
@@ -203,7 +203,7 @@ presubmits:
               memory: 6Gi
 
   - name: pre-kubermatic-dualstack-e2e-smoke
-    run_if_changed: ".*(cilium|canal|dualstack|api|addon|defaults|cni|validation|operator|provider|machine|webhook|crd|.prow|go.mod|proxy|network).*"
+    run_if_changed: ".*(cilium|canal|dualstack|addon|cni|provider|machine|webhook|.prow|go.mod|proxy|network).*"
     decorate: true
     clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
     labels:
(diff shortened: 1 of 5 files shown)


ahmedwaleedmalik (changes requested):
Can you please rebase this PR to fix the commit messages?

.prow/features.yaml
```diff
@@ -180,6 +182,49 @@ presubmits:
...
+            privileged: true
+          resources:
+            requests:
+              memory: 4Gi
+              cpu: 2
+            limits:
+              memory: 6Gi
+
+  - name: pre-kubermatic-dualstack-e2e-smoke
+    run_if_changed: ".*(cilium|canal|dualstack|api|addon|defaults|cni|validation|operator|provider|machine|webhook|crd|.prow|go.mod|proxy|network).*"
```
ahmedwaleedmalik: Do we really want something this broad? Because in general, I think this will almost always run.

PratikDeoghare: We probably don't need all of that. Since the feature is new I have included anything that touches networking. Definitely can be improved in the future PRs.

ahmedwaleedmalik: I think this PR would be the best candidate to perform that improvement. Not sure why this should be done in some "future" PR.

rastislavs: I think we could definitely try it without:
 - api
 - defaults
 - validation
 - operator
 - crd

ahmedwaleedmalik (approved):
/approve

rastislavs:
/hold
needs rebase @PratikDeoghare

xrstf (approved):
/approve
/hold cancel

PR #10500, opened 2022-07-26
bump supported Kubernetes versions

xrstf:
This bumps the supported Kubernetes versions. I removed some minor versions that were never released in KKP, just because there is no point in releasing KKP 2.21 that then, for the first time, supports both 1.24.1 and 1.24.2.

I also removed `VERSION_TO_TEST` variables from our prow job files so that those versions do not accidentally get ouf of sync anymore (sometimes this lead to cluster updates running during a conformance test, which was spicy and could fail easily). The new testing jigs default to `defaults.DefaultKubernetesVersioning.Default` and thereby always test KKP's default version anyway.

```release-note
NONE
```

### DIFF
--- a/.prow/api.yaml
+++ b/.prow/api.yaml
@@ -36,8 +36,6 @@ presubmits:
           command:
             - "./hack/ci/run-api-e2e.sh"
           env:
-            - name: VERSION_TO_TEST
-              value: v1.23.8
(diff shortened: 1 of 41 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

PR #10505, opened 2022-07-26
Cloud Credential Controller

xrstf:
For a long time we have planned to remove the cloud provider credentials from the various Deployments in the cluster namespace. It's just plain wrong to have an `AWS_ACCESS_KEY=supersecret` env on the kube-apiserver Deployment.

To achieve this, this PR mirrors the credentials Secret for a usercluster from the `kubermatic` namespace into the cluster namespace. A new controller takes care of this simple procedure. This new controller also migrates cluster credentials (i.e. creates the Secret in the kubermatic Namespace and remove credentials from `.spec.cloud` in the Cluster object), something that previously was only done when a Cluster was created via the KKP API.

The new Secret is called `cloud-credentials` and will in the future be consumed by the various Deployments.

```release-note
Cloud provider credentials (`.spec.cloud` in a Cluster object) are transferred into a Secret by a controller, something previously only done during cluster creation when using the KKP dashboard. Now this procedure affects every Cluster object.
The cluster credentials are now also mirrored into the usercluster namespace.
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cluster-credentials-controller/controller.go
+++ b/pkg/controller/seed-controller-manager/cluster-credentials-controller/controller.go
@@ -1,0 +1,197 @@
+/*
+Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clustercredentialscontroller
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+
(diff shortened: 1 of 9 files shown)


pkg/controller/seed-controller-manager/cluster-credentials-controller/controller.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// patch in the cleanup finalizer first (the pkg/clusterdeletion takes care of cleaning up)
+	kuberneteshelper.AddFinalizer(cluster, kubermaticv1.CredentialsSecretsCleanupFinalizer)
+
+	if !equality.Semantic.DeepEqual(oldCluster.Finalizers, cluster.Finalizers) {
+		if err := r.Patch(ctx, cluster, ctrlruntimeclient.MergeFrom(oldCluster)); err != nil {
+			return nil, fmt.Errorf("failed to add finalizer: %w", err)
+		}
+
+		return &reconcile.Result{Requeue: true}, nil
```
lsviben, line 141: why are the requeue's after patches needed?

xrstf: They are not strictly necessary, but since we will cause a retrigger anyway because of the updates, it makes sense to just skip any further effort and wait for the next reconciliation. Ultimately this prevents a conflict later on (that is glossed over by our automatic retrying).

PR #10506, opened 2022-07-26
Use cloud credential secret for Deployments

xrstf:
This PR refactors all Deployments that get cloud credentials injected to take those from a Secret instead of inline (so no more sensitive data is put into Deployment objects).

```release-note
Cloud provider credentials are not put into environment variables for Deployments anymore, but instead Deployments reference Secrets.
```


ahmedwaleedmalik (approved):
/approve

PR #10532, opened 2022-07-27
API endpoints to list OperatingSystemProfiles

ahmedwaleedmalik:

The following endpoints were added to the list OperatingSystemProfiles at a global and cluster level, respectively:
- `/api/v2/seeds/{seed_name}/operatingsystemprofiles`
- `/api/v2/projects/{project_id}/clusters/{cluster_id}/operatingsystemprofiles`

Fixes #

**Documentation**:

```release-note
Add endpoint to list operating system profiles.
```


pkg/handler/v2/operatingsystemprofile/operatingsystemprofile.go
```diff
@@ -0,0 +1,171 @@
...
+		// Fetch OSPs for all the seeds.
+		var osps []osmv1alpha1.OperatingSystemProfile
+		for _, seed := range seeds {
+			ospList, err := operatingSystemProfile.ListUnsecured(ctx, seed.Namespace)
+			if err != nil {
+				return nil, common.KubernetesErrorToHTTPError(err)
+			}
+
+			osps = append(osps, ospList.Items...)
+		}
```
lsviben: The seeds will all have the same `kubermatic` namespace. And with this you will only get the OSP from the master cluster, which I guess is not your goal?

I understand that the custom OSPs are on the `kubermatic` namspace of the Seed clusters. So you would need to use the dynamic client that gets set based on the Seed in question. For that you would also need the {seed} in the endpoint.

Or if you want to take the custom OSPs from the master cluster, you dont need to iterate over the seeds, you can just get them from the `kubermatic` namespace. But if the OSPs reside in the master cluster, you will also need a synchronizer between master and seed clusters, which syncs the OSPs from master to all seeds (there is a bunch of similar resources already doing the same).

pkg/handler/v2/operatingsystemprofile/operatingsystemprofile.go
```diff
@@ -0,0 +1,171 @@
...
+
+		userClusterNamespace, err := clusterNamespaceFromContext(ctx, userInfoGetter, req.ProjectID, req.ClusterID)
+		if err != nil {
+			return nil, err
+		}
+
+		ospList, err := operatingSystemProfile.ListUnsecured(ctx, userClusterNamespace)
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
+		}
```
lsviben: This wont work when the clusters are on Seeds which are not also the Master cluster. You need a dynamic client

pkg/handler/v2/operatingsystemprofile/operatingsystemprofile.go
ahmedwaleedmalik: I'll properly look into this tomorrow. I think I went forward with the PR in a bit of haste. Which was wrong from my side.

lsviben: Dont worry, we can discuss it tomorrow :)

pkg/handler/v2/operatingsystemprofile/operatingsystemprofile_test.go
```diff
@@ -0,0 +1,266 @@
...
+			existingObjects: test.GenDefaultKubermaticObjects(
+				test.GenTestSeed(),
+				test.GenDefaultCluster(),
+			),
+			apiUser:            test.GenDefaultAPIUser(),
+			expectedHTTPStatus: http.StatusOK,
+			expectedResult:     []*apiv2.OperatingSystemProfile{},
+		},
+		{
+			name:      "John cannot list Bob's ApplicationInstallation",
```
lsviben:
```suggestion
			name:      "John cannot list Bob's OSP",
```

xrstf:
/hold

temporarily

PR #10533, opened 2022-07-27
allow to not specify --config when upgrading KKP

xrstf:
This makes the installer easier to use during upgrades. The config already exists in the cluster anyway, so we can just load it from there instead of forcing the admins to have it handy all the time. We could even do this for the Helm values, but one step at a time. :grin: 

```release-note
When updating an existing KKP installation, `--config` must not be specified for the kubermatic-installer. Instead the current configuration is loaded from the KKP master cluster.
```


ahmedwaleedmalik (approved):
/approve

Issue #10502, opened 2022-07-26
API Panics on Incomplete AppInstall


SimonTheLeg:
When sending an object that is missing for example the `.Status` field, the api panics. This happens due to a nilpointerexception in the conversion between external and internal types.

When touching this, it might make sense to fix #10365 with it.

Issue #10408, opened 2022-07-20
Type should be visible for ApplicationInstallations


SimonTheLeg:
This would make it easier in the Frontend to display the type next to each application, without the need to make an additional call to an appdef to figure that our

SimonTheLeg:
closed: this is already available in the status field :)

Issue #10365, opened 2022-07-18
POST/PUT for AppInstalls Should Not Contain Creation-/Deletiontimestamp


SimonTheLeg:
The POST and PUT endpoint for ApplicationInstallations currently expose the create- and deletiontimestamp from the server. This causes confusion among end-users, as these values should not be set and will automatically be rejected. Therefore it makes sense to remove them completely from the api

SimonTheLeg:
solved by #10579

PR #10535, opened 2022-07-27
applicationInstallationController: set conditions

### LINKED ISSUE #10546: ApplicationInstallationController should set the conditions
### What happened?
the controller does not set the condition so the UI can not get information about the installation status.

### Expected behavior

the condition for RetrivedManfifest and Ready should be set by the controller

vgramer:
this pr set the conditions in the status. the following change have been made:

* conditions type change from slice to `map[appskubermaticv1.ApplicationInstallationConditionType]appskubermaticv1.ApplicationInstallationCondition` for convenience when manipulating conditions.
* ApplicationInstallationCondition now contains the `LastTransitionTime` and `LastHeartbeatTime` (useful when an application is upgraded)
*  The `Apply`method of `Installer` was previously downloading the source and installing/upgrading the application. This method has been split into 2 functions. `DownloadSource` and  `Apply`  in order to be able to set the condition `ManifestRetrived`

On API side:
* add `method` field  in `apiv2.ApplicationInstallationStatus`.  it was previously in applicationVersion but when mode it to spec level, we forget to update the API part.
* when converting type from external to internal, we don't set the status as it makes no sense because it is only set by the controller and ignored for creating / updating operations.  Moreover, it avoids conversion problem when Do POST (see #10502). 

Fixes  #10546 #10408

**Documentation**:

```release-note
NONE
```


pkg/apis/apps.kubermatic/v1/application_installation.go
```diff
@@ -125,8 +125,9 @@ type Version struct {
 // ApplicationInstallationStatus denotes status information about an ApplicationInstallation.
 type ApplicationInstallationStatus struct {
 	LastUpdated metav1.Time `json:"lastUpdated,omitempty"`
```
vgramer: @SimonTheLeg i think we can remove this field as condition as `LastTransitionTime` and  `LastHeartbeatTime`. field. WDYT?

pkg/handler/v2/application_installation/application_installation.go
```diff
@@ -184,11 +204,6 @@ func convertExternalToInternal(app *apiv2.ApplicationInstallation) *appskubermat
 			Namespace: app.Namespace,
 		},
 		Spec: *app.Spec,
-		Status: appskubermaticv1.ApplicationInstallationStatus{
```
vgramer, line 187: status is only set by the controller and will be ignored by create/update operation. So it makes no sense to set it.

IMHO we should remove the status from `createApplicationInstallationReq` / `updateApplicationInstallationReq`

@SimonTheLeg WDYT

vgramer: there already is an issue for that #10365

pkg/apis/apps.kubermatic/v1/application_installation.go
vgramer: as discussed in meeting we can remove this field

pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
```diff
@@ -219,23 +220,62 @@ func (r *reconciler) getApplicationVersion(appInstallation *appskubermaticv1.App
...
+		return fmt.Errorf("failed to update status: %w", err)
+	}
+
+	oldAppInstallation = appInstallation.DeepCopy()
+	statusUpdater, installErr := r.appInstaller.Apply(ctx, log, r.seedClient, r.userClient, appInstallation, appSourcePath)
+
+	if installErr != nil {
+		r.setCondition(appInstallation, appskubermaticv1.Ready, corev1.ConditionFalse, "InstallationFailed", installErr.Error())
+	} else {
+		r.setCondition(appInstallation, appskubermaticv1.Ready, corev1.ConditionTrue, "InstallationSuccessful", "application sudccessfuly install or upgraded")
```
lsviben: typo
```suggestion
		r.setCondition(appInstallation, appskubermaticv1.Ready, corev1.ConditionTrue, "InstallationSuccessful", "application successfully installed or upgraded")
```

ahmedwaleedmalik (approved):
/approve

PR #10536, opened 2022-07-27
Fix finalizers getting overwritten by cloud & cluster credentials controller

xrstf:
Recently in #10359 I introduced a change where the AWS cloud provider would only cleanup if its finalizer is set. This was done to save on API calls, as during the cluster deletion the cloud controller might reconcile frequently.

This change has unearthed another problem: Sometimes the AWS cleanup finalizer got missing, leading to orphaned tags during e2e tests. We noticed once we ran into the max-tag limit for subnets multiple times a day.

This PR fixes that problem by changing the ClusterUpdater to use optimistic locking when finalizers are involved. Most providers forgot to specify the optimistic locking option, so I see this as prove that relying on the provider will lead to issues and and automatic selection is more reliable.

Additionally I changed the Try*Finalizer() functions to also use patching with optimistic locking, just as a minor optimizations now that I was sure how to properly handle finalizers with patches and since I saw that asserting `.DeepCopyObject().(ctrlruntimeclient.Object)` is not so uncommon.

The cloud provider part also affects 2.20. The credential secret stuff is 2.21+ only. So I expect a manual cherrypick.

```release-note
Fix finalizers on clusters sometimes getting overwritten by the cloud controller or cluster-credentials controller.
```

### DIFF
--- a/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
+++ b/pkg/controller/seed-controller-manager/cloud/cloud_controller.go
@@ -273,7 +273,7 @@ func (r *Reconciler) migrateICMP(ctx context.Context, log *zap.SugaredLogger, cl
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


moadqassem (approved):
/approve
/lgtm

PR #10537, opened 2022-07-27
Add externalclusters versioning configuration in KubermaticConfiguration

### LINKED ISSUE #10538: Add externalclusters versioning configuration in KubermaticConfiguration

As a user  
I would like to have a list of supported kubernetes versions for externalclusters
in order to have it while creating cluster.

### Solution details

- For the externalcluster providers where we a golang sdk method is not available to get the list.
we will have the available versions in KubermaticConfiguration

imharshita:

Add externalclusters versioning configuration in KubermaticConfiguration:
```
spec:
   versions:
       externalClusters:
           eks: 
             versions: 
             - x.x
             - y.y
             updates:
             - z.z
           aks: 
             versions: 
             - x.x
             - y.y
 ```

Fixes #10538

- These changes are mainly done for providers:
  EKS:
  - Available/Supported Versions to create cluster by provider
  - Update Version can be figured out using current version and the above available list
  
  AKS:
  - Available/Supported Versions to create cluster by provider
 
**Documentation**:

```release-note
KubermaticConfiguration containing ExternalCluster version configuration for providers like EKS
```


pkg/apis/kubermatic/v1/configuration.go
```diff
@@ -367,6 +367,28 @@ type KubermaticVersioningConfiguration struct {
 
 	// ProviderIncompatibilities lists all the Kubernetes version incompatibilities
 	ProviderIncompatibilities []Incompatibility `json:"providerIncompatibilities,omitempty"`
+
+	// ExternalClusters contains the available and default Kubernetes versions and updates for ExternalClusters.
+	ExternalClusters map[ExternalClusterProviderType]ExternalClusterProviderVersioningConfiguration `json:"externalClusters,omitempty"`
```
ahmedwaleedmalik, line 372: This looks like a super good addition. Although users will have to manually configure this in `KubermaticConfiguration` in order to create external clusters. Which adds an additional configuration layer for someone who simply just wants to create an external cluster.

- Will we have support in the dashboard to configure this? I don't want to push for something so late but that might be essential for this.
- What's your take on defaulting this list to the possible values? I'm aware that it might lead to some hardcoding in our code.
- Currently, if I create an external cluster directly via a CR(Not API/UI) with an invalid version, what will happen? 
- Currently, if I create an external cluster via API/UI with an invalid version, what will happen?

imharshita: Yes, thats what ee are planning to do , to have a default list in the kkp code so that user dont have to

imharshita: Currently if u provide a invalid version, (via api) ,it will return an error

imharshita: As cluster creation code is in api, not in controller. Creating a cr will not do anything to create an actual cluster.

xrstf: Compared to the Kubernetes versioning config, this is supposed to only take minor releases, which is why I pivoted to kind of preferring this solution (previously I was thinking about putting the versions for ExternalClusters in the KubermaticSettings, i.e. UI only). We already ship the default list of Kubernetes versions and keeping those updated is (I think) more work than updating the minor versions for external clusters.

But super happy, I am not.

ahmedwaleedmalik (approved):
/approve

PR #10542, opened 2022-07-28
fix RELEASES_TO_TEST not being optional enough

xrstf:
This makes sure that jobs can run with an automatic version selection, as originally intended in #10500.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik (approved):
/approve

PR #10545, opened 2022-07-28
[release/v2.20] Freeze kubermatic/mla commit to v0.2.2

wurbanski:
kubermatic/mla has been changed in ci non-backwards-compatible-way. This freezes on the last working release of mla.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/hack/ci/setup-mla.sh
+++ b/hack/ci/setup-mla.sh
@@ -28,6 +28,9 @@ source hack/lib.sh
 
 URL="https://github.com/kubermatic/mla.git"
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

xrstf:
We need the same for the master branch, or else when we branch away for 2.21, we will run into the same issue again. For machine-controller and OSM we do not allow these shenanigans, but require them to specifiy a fixed commit. Maybe for MLA we should do the same, so we don't break when doing a new release.

PR #10566, opened 2022-07-28
Add ApiserverIsRunning wrapper to the usercluster-webhook

xrstf:
This prevents the webhook from going into CrashLoops while the cluster is being created.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10575, opened 2022-07-29
Forbid users from disabling OSM after it has been enabled

ahmedwaleedmalik:

OSM is enabled by default for new clusters but existing clusters have to migrate to OSM manually. Due to this, we allow users to modify this field in the Cluster Spec so that they can enable OSM for existing clusters.

Although, disabling OSM after it has been enabled should never be allowed. Since machine deployments will have finalizers that will not get removed and the clusters will get stuck in deletion. Other resources,  created and managed by OSM like OSPs, OSCs, and secrets are not removed as well.

Fixes #

**Documentation**:

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #10577, opened 2022-07-29
Inject BUILD_ID and test name into cluster names again

xrstf:
This is something we already did maaany years ago, but in some refactoring craze, I removed the build ID from cluster names. After that, we could not easily identify whether for example the AWS tags belong to a "real" usercluster or are from some e2e test.

This PR brings back the build ID and also gives each cluster a more descriptive name. During this exercise I learned that since we prepend and append things to cluster names, we must restrict cluster names to be much shorter than anyone would think. While a DNS label can be 63 characters or so, due to the way AWS restrict tag lengths and us prepending "kubernetes-" and appending "-role-name", the effective max length for a cluster name is around 37 characters. I therefore adjusted our webhook to enforce this and had to be a bit more creative when naming the test ("etcd-recovery" instead of "etcd-launcher-recovery").

The Prow jobs got a higher memory limit because they started to get OOMKilled during compilations.

Since this PR focussed my on the cluster namespace handling, I also refactored the ClusterNameFromNamespace function, which previously was built assuming that a cluster namespace is ALWAYS `cluster-.....` But I do not want to make this assumption, there is a reason why we store the namespace name in the ClusterStatus and do not constantly rebuild it. To ensure that this logic ("deducing the cluster from the namespace name") doesn't grow like cancer, I refactored the function to list all Clusters and loop through them, finding the matching one. So this tiny refactoring is also part of this PR.

This tiny refactoring however showed that the etcd-launcher doesn't have permissions to LIST cluster objects, only to GET one specific cluster object. Since rewriting the rbac controller is too dangerous for me, @embik and me decided that it would be easier to just change the etcd-launcher to not have `-namespace` but `-cluster`. So that is all the change I did to the etcd-launcher's `main.go`. Looks horribly complicated in the diff, though.

And on this I noticed that we had a few event handlers that basically reimplemented `EnqueueClusterForNamespacedObject`, so I updated those controllers as well.

And lastly, there is a tiny bugfix to handle BYO cluster credentials, which I recently broke in #10505.

```release-note
NONE
```

### DIFF
--- a/.prow/ccm-migrations.yaml
+++ b/.prow/ccm-migrations.yaml
@@ -50,7 +50,7 @@ presubmits:
               memory: 4Gi
               cpu: 3.5
             limits:
-              memory: 4Gi
+              memory: 6Gi
 
   - name: pre-kubermatic-ccm-migration-vsphere-e2e
     run_if_changed: "(pkg/provider/cloud/vsphere/|pkg/resources/cloudcontroller/|pkg/test/e2e/ccm-migration/|.prow/ccm-migrations.yaml)"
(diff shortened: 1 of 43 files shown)


embik (approved):
/approve

ahmedwaleedmalik (approved):
/approve

PR #10595, opened 2022-08-01
event handling for 404 and 403 errorcodes

### LINKED ISSUE #10638: Handle 404 and 403 errors better in externalcluster controller
### What happened?

Handle 404 and 403 errors  better in externalcluster controller  to not reconcile external cluster controller when cluster  is not found or authorization failed.

### Expected behavior
Controller should instead emit events to inform user about the 404 and 403 errors.

- KKP version:

<details>

```yaml
# paste manifest here
```

</details>

imharshita:

Fixes #10638

Last commit reactors code to replace func argument from `ExternalClusterCloudSpec` => `ExternalClusterAKSCloudSpec` (for AKS), same for GKE and EKS as passing whole Spec is not required.

**Documentation**:

```release-note
none
```


pkg/util/errors/errors.go
```diff
@@ -21,6 +21,13 @@ import (
 	"net/http"
 )
 
+type StatusCode int
+
+const (
+	NotFoundStatusCode  StatusCode = 404
+	ForbiddenStatusCode StatusCode = 403
```
xrstf: You can just use the constants from `net/http`, no need to define your own set of HTTP status code constants.

pkg/provider/cloud/gke/provider.go
```diff
@@ -479,3 +481,31 @@ func createClient(ctx context.Context, serviceAccount string, scope string) (*ht
...
+	if err == nil {
+		return statusCode
+	}
+
+	var apiErr *googleapi.Error
+	if errors.As(err, &apiErr) {
+		return apiErr.Code
+	}
+
+	return statusCode
```
xrstf: Not sure why you created an unused variable here. This function can just be

```
func DecodeErrorCode(err error) int {
   var apiErr *googleapi.Error
   if errors.As(err, &apiErr) {
      return apiErr.Code
   }

   return 0
}
```

pkg/provider/cloud/gke/provider.go
```diff
@@ -479,3 +481,31 @@ func createClient(ctx context.Context, serviceAccount string, scope string) (*ht
 
 	return client, projectID, nil
 }
+
+func DecodeError(err error) error {
+	if err == nil {
+		return nil
+	}
```
xrstf: This nil check can be removed, I think. errors.As returns false if the error is nil: https://cs.opensource.google/go/go/+/refs/tags/go1.18.5:src/errors/wrap.go;l=91

pkg/provider/cloud/eks/provider.go
```diff
@@ -459,6 +460,23 @@ func DecodeAWSError(err error) error {
 	return err
 }
 
+func DecodeAWSErrorCode(err error) int {
+	var statusCode int
```
xrstf: Same comment as for gke.DecodeErrorCode()

ahmedwaleedmalik (approved):
/approve

PR #10603, opened 2022-08-01
Scrape kubelet resources metrics endpoint by default for user clusters

mate4st:

Scrape kubelet resources metrics endpoint by default for user clusters

```release-note
Make user cluster kubelet resource metrics available
```


xrstf (changes requested):
Please extend https://github.com/kubermatic/kubermatic/blob/master/cmd/conformance-tester/pkg/tests/metrics.go#L52 to prove that this new scraping rule actually works and continues to work in the future.

xrstf (approved):
/approve

PR #10606, opened 2022-08-01
AddonConfig Disable/Enable KKP auto assign IPAddressPool

### LINKED ISSUE #10547: Improvement of MetalLB Addon / IPAM Pool config
Following topics are missing in the current metallb / IPAM pool config

### 1) Reconciling of `ipampools.kubermatic.k8c.io`
Currently if an IPAM Pool get deleted, the user cluster addon object `ipaddresspools.metallb.io` doesn't get deleted automatically.
It seams that the `kubectl apply` only updates the additional address pool field, but the automatic generated ip address pool object stays. 

sig-cs-infra | … 2022-07-28 15-04-51

#### Why we need this?
Let's expect that a customer want to refactor their IPAM Pools and delete one pool. Then it get split up and reconfigured and re-applied. The old user cluster IP pool is still active and metallb will serve potetnial traffic to it.  A new cluster get created and get the same IP assigned and metallb start to serve traffic to this cluster as well. This will lead to the situation that we have potential 2 services with the same IP active. What should be avoided by the developed IPAMPool. Background, metallb make the virtual IPs available via ARP, so it's can happen that in this case two cluster will get mixed up traffic - debugging is really hard in this case, because somehow you will lost packages and you are blind to find out where it get routed to. 

### 2) Make automatic IP address pool optional and highlight the yaml field as additional IP address config option
To ensure full control from a user cluster admin perspective, he should have an option to disable the automatic IP address pool assignment. Also the `yaml` config should be described in a way  a enduser understand that it gets it additional to the default config. 
Cluster 'friendly-rosalind' in Project 'play' 2022-07-28 14-28-45(1)

adeniyistephen:
We added a new feature to enable and disable auto assign IPAddressPool for metallb.
Fixes #10547

This is important for the metallb addon coming in the next release for issue: #10547, it fixes the bug for users to be enable/disable IPAddressPool from the KKP IPAMPool on a cluster, some clusters might need the automatic IPAMPool while some might not need the IPAMPool.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/.prow/features.yaml
+++ b/.prow/features.yaml
@@ -358,7 +358,7 @@ presubmits:
               memory: 6Gi
 
   - name: pre-kubermatic-ipam-e2e
-    run_if_changed: "(pkg/controller/seed-controller-manager/ipam/|pkg/webhook/ipampool/|pkg/test/e2e/ipam/)"
+    run_if_changed: "(pkg/controller/seed-controller-manager/ipam/|pkg/webhook/ipampool/|addons/metallb/|pkg/test/e2e/ipam/)"
     decorate: true
     clone_uri: "ssh://git@github.com/kubermatic/kubermatic.git"
     labels:
--- a/addons/metallb/01_address-pools.yaml
+++ b/addons/metallb/01_address-pools.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{- if .Variables.autoIPAddressPool }}
+{{- if .Variables.disableAutoIPAMAllocation }}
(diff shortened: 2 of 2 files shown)


hbernardo:
Can you please add the metallb addon folder to IPAM e2e tests? File `.prow/features.yaml`:
```
...
  - name: pre-kubermatic-ipam-e2e
    run_if_changed: "(pkg/controller/seed-controller-manager/ipam/|pkg/webhook/ipampool/|addons/metallb/|pkg/test/e2e/ipam/)"
...
```

xrstf (approved):
/approve

xrstf:
Oh did I come in too early with my review?

PR #10625, opened 2022-08-01
Add projectId suffix to binding subjects

### LINKED ISSUE #10602: Groups allowing for project's permission leaks
### What happened?

I was able to create a cluster using cluster template wizard with `viewers` role assigned through `GroupProjectBinding`. Attempting to do the same with `UserProjectBinding` resulted in permission denied error as expected.

### Expected behavior

Permission denied error (403).

1. Login with your Google account and create two new projects.
2. In the first project, let's say A, create `GroupProjectBinding` between one of your Github groups (teams) e.g. `kubermatic:developmet` and assign `owners` role.
3. In the second project, let's call it  B, create another `GroupProjectBinding` between previously used group name and assign `viewers` role.
4. Switch accounts and login using Github.
5. Go to project B and open Cluster Template creation wizard.
6. Fill wizard's forms and on the last step click on the Create Cluster button.

- KKP version: 2.21 (dev)

### Additional information

Investigation showed that we create a global cluster role (with `create` action) and cluster role binding (with group we assigned through `GroupProjectBinding`) for any `GroupProjectBinding` with `editors` or `owners` roles. 

This global role allows for creation of resources like `clusters` or `userSshKeys` in the projects where user has only `viewers` role. The only requirement is that user has to have `editors` or `owners` role assigned through `GroupProjectBinding` in any other project available on KKP instance.

For `UserProjectBindings` we use group suffixes making it resistant to such behaviour.

UPB impersonated client:

GPB impersonated client:

Global role:

pkprzekwas:
Add project id suffix to role binding subjects to avoid unwanted role leaks.

Fixes #10602 

```release-note
NONE
```


pkg/provider/userinfo.go
```diff
@@ -52,7 +52,13 @@ func UserInfoGetterFactory(userProjectMapper ProjectMemberMapper) (UserInfoGette
 				return nil, err
 			}
 		} else {
-			groups.Insert(user.Spec.Groups...)
+			for _, group := range user.Spec.Groups {
+				groupName := group
+				if projectID != "" {
+					groupName += "-" + projectID
```
embik: Very minor and personal nit, but can we make this

```suggestion
					groupName += fmt.Sprintf("-%s", projectID)
```
?

embik (changes requested):
Looks good, I think this should fix our issue. Just a very minor nit.

pkprzekwas:
> Looks good, I think this should fix our issue. Just a very minor nit.

Corrected, as suggested.

embik (approved):
@pkprzekwas feel free to remove WIP status and add the 2.21 milestone if you think this is good to go, from my perspective it should fix our issue (although we should verify afterwards anyway).

embik (approved):

PR #10627, opened 2022-08-01
Use reconciling framework to manage the credentials secret

xrstf:
This is a simple refactoring to make use of the pre-existing reconciling framework instead of handrolling the credential secret creation/updates. The same applies to the code that creates the kubeconfig Secret for external clusters.

I also removed the "name" label, which was put on each and every credential secret. The original implementation stems from #3684, but I could not figure out why we ever added a redundant name label. I could imagine that this was a copy-paste mistake where the `name` label that we put on Service Account Token Secrets (which contains the human readable name that users can specify in the KKP dashboard when creating a new token) was simply accidentally copied into the code for managing the credential secrets (which never get a human readable name and so do IMHO not need a `name` label).

This PR also contains a super tiny fix for a wrong log line.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10631, opened 2022-08-01
Add comments to default k8s versions

imharshita:

**What does this PR do / Why do we need it**: Add comments to default k8s versions

Fixes #

**Documentation**:

```release-note
none
```

### DIFF
--- a/pkg/controller/operator/defaults/defaults.go
+++ b/pkg/controller/operator/defaults/defaults.go
@@ -302,13 +302,16 @@ var (
 	}
 
 	aksProviderVersioningConfiguration = kubermaticv1.ExternalClusterProviderVersioningConfiguration{
-		Default: semver.NewSemverOrDie("v1.22"),
+		Default: semver.NewSemverOrDie("v1.22.11"),
 		Versions: []semver.Semver{
 			// v1.24 is a Preview version (not Production ready).
 			// newSemver("v1.24"),
-			newSemver("v1.23"),
-			newSemver("v1.22"),
-			newSemver("v1.21"),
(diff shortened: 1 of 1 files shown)


xrstf:
Wait, wasn't the plan that we only have to maintain minor releases for external providers? Now we suddely have to keep up with all their patch releases as well?

In any case, please put a link in the code to the source wher you got that list from, regardless whether it's just minor versions or full versions.

imharshita:
> Wait, wasn't the plan that we only have to maintain minor releases for external providers? Now we suddely have to keep up with all their patch releases as well?
> 
> In any case, please put a link in the code to the source wher you got that list from, regardless whether it's just minor versions or full versions.

Yes the plan was to keep only major.minor versions but , it's not working out for aks.
for EKS its completely all right.
These is a temporary change.
I am working on fetching the list by making an api call to `GET https://management.azure.com/subscriptions/{subscriptionId}/providers/Microsoft.ContainerService/locations/{location}/orchestrators?api-version=2019-08-01
`
Doc: https://docs.microsoft.com/en-us/rest/api/aks/container-services/list-orchestrators?tabs=HTTP#orchestratorversionprofile

But it is not straightforward.
1. First, we need to fetch an authorization token using OAuth Authorization.
2. then we need to make a rest API call to get the list, then filter it out.

2nd is easy. 1. is something which I have not done before.
So, we have 2 options, either we wait till I figure it out and create a PR. (i will try asap)
or we keep this list till with major.minor.patch versions (only for aks) till I push my changes.
wdyt?

sorry for changing the plan, my bad.

xrstf (approved):
/approve

PR #10649, opened 2022-08-02
[release/v2.20] [EE] fix reconcile loop in allowed-registry controller (#10644)

xrstf:
This is a manual backport of #10644 because in KKP 2.20, we still use v1beta1 of the Gatekeeper APIs. I checked and both v1 and v1beta1 use the same pointer for this field.

```release-note
Fix reconcile loop in AllowedRegistry controller
```

### DIFF
--- a/pkg/ee/allowed-registry-controller/reconcile.go
+++ b/pkg/ee/allowed-registry-controller/reconcile.go
@@ -42,7 +42,6 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/utils/pointer"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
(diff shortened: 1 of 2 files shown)


embik (approved):

embik:
/lgtm cancel

```
# k8c.io/kubermatic/v2/pkg/ee/allowed-registry-controller
pkg/ee/allowed-registry-controller/reconcile.go:164:8: cannot use pointer.Bool(false) (type *bool) as type bool in field value
```

embik:
My IDE says this is the type definition as it exists in whatever version of our dependencies that we use:

```go
type Validation struct {
	// +kubebuilder:validation:Schemaless
	// +kubebuilder:validation:Type=object
	// +kubebuilder:pruning:PreserveUnknownFields
	OpenAPIV3Schema *apiextensionsv1.JSONSchemaProps `json:"openAPIV3Schema,omitempty"`
	// +kubebuilder:default=true
	LegacySchema bool `json:"legacySchema,omitempty"`
}
```

xrstf:
Yeah seems like checking v1beta1 on the master branch of OPA is not the same as v1beta1 on an older branch. So much for versioning.

But on the flipside, we can close this PR.

embik (approved):

PR #10660, opened 2022-08-03
applicationInstallationController: fix infinite reconciling loop

### LINKED ISSUE #10639: AppplicationInstallationController: update status trigger infinite reconcilation
### What happened?

When we create an applicationInstallation,  the status is set with  the Helm information and the condition `Ready`.  These changes to the status trigger an Update Event so the application is "re-installed", the  status is updated and so on and so on...

```diff
$ stalk applicationinstallations
--- ApplicationInstallation default/my-app v28744 (2022-08-01T14:21:12+02:00) (gen. 1)
+++ ApplicationInstallation default/my-app v28749 (2022-08-01T14:21:14+02:00) (gen. 1)
@@ -10,7 +10,7 @@
   generation: 1
   name: my-app
   namespace: default
-  resourceVersion: "28744"
+  resourceVersion: "28749"
   uid: 74862b5d-f028-4de8-9375-ec3a4f988fda
 spec:
   applicationRef:
@@ -47,7 +47,7 @@
       reason: DownaloadSourceSuccessful
       status: "True"
     Ready:
-      lastHeartbeatTime: "2022-08-01T12:21:12Z"
+      lastHeartbeatTime: "2022-08-01T12:21:13Z"
       message: application successfully installed or upgraded
       reason: InstallationSuccessful
       status: "True"
@@ -55,8 +55,8 @@
     info:
       description: Upgrade complete
       firstDeployed: "2022-08-01T12:18:46Z"
-      lastDeployed: "2022-08-01T12:21:11Z"
+      lastDeployed: "2022-08-01T12:21:13Z"
       status: deployed
     name: default-my-app
-    version: 98
+    version: 99
   method: helm
```
### Expected behavior
ApplicationInstallation should not be reinstalled on the status updates.

vgramer:
# What does this PR do / Why do we need it
When we create an application installation, the status is set with the Helm information and the conditions (e.g. Ready). These changes to the status trigger an Update Event so the application is "re-installed", the status is updated and so on and so on...

Fixes #10639 

# Special notes for your reviewer
As a consequence of using the predicate filter, the ReSync period of the controller, which is 10h (default value) is ignored. 
So the application will not force to be reconciled after 10h anymore. 

**Do we want to keep this behavior?**

It can esaly implemented by using `reconcile.Result{RequeueAfter: 10 * time.Hour}`

# Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
+++ b/pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
@@ -39,6 +39,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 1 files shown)


xrstf:
Maybe instead of filtering the events, the same approach we use for Cluster conditions and EtcdBackupConfig conditions should be applied here as well? As done recently in #10650.

vgramer:
@xrstf  hum... tricky question! 

If only the condition was updated, I would say yes. But the status is also updated with the HelmRelease information. We could apply the same strategy to  HelmRelease information as conditions (comparing fields ignored dates) but _technically_ we have no guarantee that fields are the same over 2 installations with the same parameters :/ 

IMHO: the best argument to use predicate is that the code is generic so other installation methods than helm can be used. I m pretty sure that the day we will add another method, we will forget this. So I prefer a more defensive programming strategy. 

By the way, I found this bug in literally 5 seconds thanks to stalk. so thank  ❤️❤️❤️

xrstf (approved):
Let's try it and see what happens :)

/approve

PR #10665, opened 2022-08-03
Update Go to 1.19.1

xmudrii:

This PR updates Go to 1.19.1 for all jobs and images used by KKP.

```release-note
KKP is now built using Go 1.19.1
The `quay.io/kubermatic/util:2.2.0` Docker image does not ship with `yq3` and `yq4` anymore, but only `yq` (version 4.x).
```


ahmedwaleedmalik:
@xmudrii Do we really need this for KKP 2.21?

xrstf:
I would vote that yes, we should go to Go 1.19. There's still a full month of QA and that should be enough to find any critical issues.

ahmedwaleedmalik:
> I would vote that yes, we should go to Go 1.19. There's still a full month of QA and that should be enough to find any critical issues.

Then we have a decision. I just want us to be consistent everywhere so I'd love if OSM, MC, KubeOne, and KKP are on the same version for Go. Although it's something that'd be nice to have; not mandatory.

xmudrii:
This is on hold until further notice because Go 1.19 breaks Swagger generation (https://github.com/go-swagger/go-swagger/issues/2759).

xrstf:
We're not patiently waiting for go-swagger/go-swagger/pull/2808 to be finished before we can progress. I don't think there's enough pressure to go to Go 1.19 for us at the moment in order to justify some workarounds in our codebase.

ahmedwaleedmalik (approved):
/approve 〽️

PR #10675, opened 2022-08-04
Unified PR template update

csengerszabo:
Unified PR template with emphasising documentation

Fixes #

**Documentation**:

This is not a user facing change, therefore no need for change in public documentation.

```release-note
NONE
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -28,11 +28,14 @@ Optionally add one or more of the following kinds if applicable:
 <!--
 Please do one of the following options:
 - Add a link to the existing documentation
-- Add links to the related documentation changes related to this pull request. E.g. the link to the kubermatic/docs pull request.
-- If no documentation change is applicable, then a brief explanation about the reasoning (like “not user-facing change”, “Will be updated along with the next PR” etc.)
+- Add a link to the kubermatic/docs pull request
+- If no documentation change is applicable then add:
+  - TBD (documentation will be added later)
+  - NONE (no documentation needed for this PR)
 -->
+```documentation
 
-Documentation link / PR link / Explanation
+```
 
 **Does this PR introduce a user-facing change? Then add your Release Note here**:
 <!--
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,6 +24,16 @@ Optionally add one or more of the following kinds if applicable:
 
 **Special notes for your reviewer**:
 
+**Does this PR introduce a user-facing change? Then add your Release Note here**:
+<!--
+Write your release note:
+1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
+2. If no release note is required, just write "NONE".
+-->
+```release-note
+
+```
+
 **Documentation**:
 <!--
 Please do one of the following options:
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,48 @@
...
+/kind design
+
+Optionally add one or more of the following kinds if applicable:
+/kind api-change
+/kind deprecation
+/kind failing-test
+/kind flake
+/kind regression
+/kind question
+-->
```
xmudrii: This is useful if we want to switch to the upstream release-notes generation tool. We did it for KubeOne and I'm quite happy with how it works, so maybe we should do the same for other projects.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,48 @@
...
+Please do one of the following options:
+- Add a link to the existing documentation
+- Add a link to the kubermatic/docs pull request
+- If no documentation change is applicable then add:
+  - TBD (documentation will be added later)
+  - NONE (no documentation needed for this PR)
+-->
+```documentation
+
+```
```
xmudrii: I would put this below the release-note block. I think the flow would be a little bit better.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,48 @@
...
-**Does this PR introduce a user-facing change?**:
-<!-- Write your release note:
+**Does this PR introduce a user-facing change? Then add your Release Note here**:
+<!--
+Write your release note:
 1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
 2. If no release note is required, just write "NONE".
 -->
 ```release-note
+NONE
```
xmudrii: I wouldn't default to NONE because users might just ignore it. If you don't specify anything, your PR will be blocked unless you specify a release note or NONE. In other words, you explicitly ask users to opt-in for NONE if that's really the case.

xmudrii (commented):
Some nits, otherwise it looks good to me.

.github/PULL_REQUEST_TEMPLATE.md
csengerszabo: these exact same /kind labels are good for everywhere?

xmudrii: I expect some minor changes, e.g. that we'll add some more labels/kinds, but other than that, I think those should be good for everyone.

csengerszabo:
@xmudrii I added you suggestions

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,48 @@
...
 
-**Documentation**:
-<!-- Add links to the related documentation changes related to this pull request. E.g. the link to the kubermatic/docs pull request. -->
+Optionally add one or more of the following kinds if applicable:
+/kind api-change
+/kind deprecation
+/kind failing-test
+/kind flake
+/kind regression
+/kind question
```
embik: I don't think `/kind question` is applicable since no PR should be a question. Can we remove it?

embik (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
csengerszabo: I removed it

csengerszabo (commented):
@embik I did the change you suggested

xmudrii (approved):

xrstf:
/hold

temporarily

PR #10677, opened 2022-08-04
Improve Conformance Tester

xrstf:
### What does this PR do / Why do we need it
This is the result of using the conformance-tester manually to work on all the KKP 2.21 QA.

### Background

The conformance-tester was originally written to be used by humans rather than exclusively on CI. However over time as people has left the project, we used the CI cluster more and more and so I, during refactorings/updates neglected and ultimately removed a lot of "usability" features.

This PR undoes a bit of that and then improves the tester a whooooole lot.

### This PR

Many of the improvements in this PR are interlinked and so it's hard to extract them back into individual PRs. Please read this text blob first to understand what's going on. Sorry to make you have to review this. But it's worth it :)

### `pkg/util/wait` takes a context to allow graceful shutdown

When hitting Ctrl-C, the root context gets cancelled. This will make calls like `client.Get()` fail, which inside a `wait.Poll()` loop will most likely result in a transient error. This is because pretty much none of our code probes the given context and/or checks whether the context is valid when an error happens. 

This in turn will make `wait.Poll()` continue to call the wait function, even though it's pointless. To combat this, all functions in the package now take a context and will stop looping when the context is cancelled/timed out.

As waiting pretty much always implies that we're waiting for _external_ state and getting external state always involves a context, all places that called the `Poll()` function and friends already had a context at hand.

This change alone caused all the changes in `pkg/test/e2e/*`.

### Introduces `pkg/util/flagopts.SetFlag`

This is a simple helper to make it easier to use `sets.String` in CLI flags. Nothing special. The unsued legacy `StringArray` helper was in turn removed.

### `-exclude-tests` flag

Instead of disabling _all_ tests at once, more fine-grained control is now given. Each of the test has a name (conformande, loadbalancer, storage, ....) and, just like distributions, the list of tests can be overriden.

### Reconciling UserSSHKeys

During cluster setup, when multiple clusters use the same keys of the project, we can run into conflicts and so need to reconcile. To make this easier, I extended the reconciling framework to cover UserSSHKeys.

### New Scenario Generator

Previously the individual scenarios (e.g. `awsScenario`) took care of multiplying all the various options to build a text matrix. This was unwieldy and not easy to maintain.

To improve this, a new `Generator` now exists. It can be fed with the chosen providers, distributions etc. and will spit out a `[]Scenario`. Each scenario is asked if it's valid before it's returned, so each provider can still "reject" combinations (like Anexia rejecting everything but Flatcar-based scenarios).

As all scenarios are based on the same information (provider, OS, version, ...), this is encapsulated in the `baseScenario`, which all other scenario types embed.

### `pkg/resources/machine` made reusable

Each of the scenarios has 2 pieces: the KKP API part and the KubeAPI part. This means each scenario has a `Cluster()` function that returns a `kubermaticv1.Cluster` for when we run with `-client=kube`, and it has an `APICluster()` which returns an `apimodels.CreateClusterSpec`. Same goes for specifying the machines.

Previously it was my plan to *not* reuse the KKP API code for building the new KubeAPI stuff. I consider the KKP API to be "legacy" and did not want to introduce a new dependency on legacy code.

However it is very complicated to actually fully assemble a ProviderSpec for MachineDeployments. For providers like Hetzner that only have a token and a network, it's easy. But look at how complex vSphere is and you soon change you mind.

So this PR changes the approach and reuses the `pkg/resource/machine` code for creating machines. Inside that package, the code for creating the `[provider]types.RawConfig` was made public so it can be reused from the conformance-tester. The tester then only has to create a `apiv1.NodeSpec` and let `pkg/resource/machine` take over the magic of combining the cluster info, the datacenter info and the nodespec (usually credentials, instance types, etc.) into a proper ProviderSpec, which can then later be embedded into a MachineDeployment.

This has the upside that KubeAPI based tests are now available for all supported providers, as it's not that hard to do it anymore. So this PR changed all Prow jobs to use `SETUP_MODE=kube` and flips the one `-headless` to be the one `-api` job. :)))

### Example Usage

The tester can be used like this:

```bash
make clean conformance-tester

_build/conformance-tester \
  -digitalocean-token "MY_DO_TOKEN" \
  -digitalocean-kkp-datacenter "do-ams3" \
  -hetzner-token "MY_HETZNER_TOKEN" \
  -hetzner-kkp-datacenter "hetzner-nbg1" \
  -providers "digitalocean,hetzner" \
  -distributions "ubuntu,centos,rockylinux,flatcar,rhel" \
  -releases "1.22,1.23,1.24" \
  -container-runtimes "docker,containerd" \
  -enable-osm \
  -client "kube" \
  -log-format "Console" \
  -name-prefix "kkp-qa" \
  -exclude-tests "conformance,telemetry" \
  -wait-for-cluster-deletion=false \
  -kubermatic-seed-cluster "kkp-qa-env" \
  -reports-root "$(realpath reports)"
```

```release-note
NONE
```


cmd/conformance-tester/pkg/clients/client_kube.go
```diff
@@ -200,75 +240,60 @@ func (c *kubeClient) CreateNodeDeployments(ctx context.Context, log *zap.Sugared
...
 
-	log.Info("Creating MachineDeployments...")
+	c.log(log).Info("Creating MachineDeployments...")
 	for _, md := range mds {
-		if err := userClusterClient.Create(ctx, &md); err != nil {
-			return fmt.Errorf("failed to create MachineDeployment: %w", err)
+		if err := wait.PollImmediateLog(ctx, log, 5*time.Second, time.Minute, func() (error, error) {
+			return userClusterClient.Create(ctx, &md), nil
+		}); err != nil {
+			return fmt.Errorf("didn't get NodeDeployments from scenario within a minute: %w", err)
```
embik: This error message doesn't look correct (copy-paste from the `wait.PollImmediate` above?).

cmd/conformance-tester/pkg/scenarios/hetzner.go
```diff
@@ -84,20 +55,29 @@ func (s *hetznerScenario) APICluster(secrets types.Secrets) *apimodels.CreateClu
 
 func (s *hetznerScenario) Cluster(secrets types.Secrets) *kubermaticv1.ClusterSpec {
 	return &kubermaticv1.ClusterSpec{
+		ContainerRuntime: s.containerRuntime,
+		Features: map[string]bool{
+			kubermaticv1.ClusterFeatureExternalCloudProvider: true,
+		},
```
embik: Why do this for Hetzner only? It doesn't look like it's done for any other provider, and afaik there's code somewhere else setting the feature flags if CCM is supported.

cmd/conformance-tester/pkg/types/options.go
```diff
@@ -139,17 +164,12 @@ func (o *Options) AddFlags() {
...
-	flag.StringVar(&o.WorkerName, "worker-name", "", "name of the worker, if set the 'worker-name' label will be set on all clusters")
-	flag.StringVar(&o.releasesFlag, "releases", o.releasesFlag, "a comma-separated list of Kubernetes releases (e.g. '1.24') to test")
-	flag.StringVar(&o.distributionsFlag, "distributions", o.distributionsFlag, "a comma-separated list of distributions to test (cannot be used in conjunction with -exclude-distributions)")
-	flag.StringVar(&o.excludeDistributionsFlag, "exclude-distributions", o.excludeDistributionsFlag, "a comma-separated list of distributions that will get excluded from the tests (cannot be used in conjunction with -distributions)")
-	flag.BoolVar(&o.OnlyTestCreation, "only-test-creation", false, "Only test if nodes become ready. Does not perform any extended checks like conformance tests")
 	flag.BoolVar(&o.PspEnabled, "enable-psp", false, "When set, enables the Pod Security Policy plugin in the user cluster")
+	flag.BoolVar(&o.OperatingSystemManagerEnabled, "enable-osm", false, "When set, enables Operating System Manager in the user cluster")
```
embik: OSM is enabled by default in KKP now. Should this flag maybe match that behaviour?

embik (changes requested):
Super small nits and questions, PR looks great overall.

cmd/conformance-tester/pkg/scenarios/hetzner.go
xrstf: You are right and I even already considered that, as client_kube.go already does

```go
	if cloudcontroller.ExternalCloudControllerFeatureSupported(scenario.Datacenter(), cluster, version.NewFromConfiguration(c.opts.KubermaticConfiguration).GetIncompatibilities()...) {
		cluster.Spec.Features = map[string]bool{kubermaticv1.ClusterFeatureExternalCloudProvider: true}
		if cloudcontroller.ExternalCloudControllerClusterName(cluster) {
			cluster.Spec.Features[kubermaticv1.ClusterFeatureCCMClusterName] = true
		}
		if cluster.Spec.Cloud.VSphere != nil {
			cluster.Spec.Features[kubermaticv1.ClusterFeatureVsphereCSIClusterID] = true
		}
	}
```

embik (approved):

PR #10682, opened 2022-08-05
applicationInstallationController: update status.HelmRelease even if an error occurred

### LINKED ISSUE #10681: Appplications: Helm release info should be updated even if an error occured
### What happened?

Currently, if an error occurs during the helm install or upgrade, we just return an error and don't set the HelmRelease Info in the status.

Depending on the error the HelmRelease Info may be or not be nil. If it's not nil, we should set the HelmRelease Info in the status.

Examples:

1. the helm files contain an invalid template; nothing will be installed and HelmRelease Info will be nil.
2. the resource created by helm does not satisfy a validation webhook; the application is partially installed and HelmRelease Info is not nil.

vgramer:
Depending on the error the HelmRelease Info may be or not be nil. If it's not nil, we should set the HelmRelease Info in the status. 

For example, if the resources created by helm do not satisfy a validation webhook, the application is partially installed and HelmRelease Info is not nil.

Fixes #10681 

The installer now always returns a non-nil `StatusUpdater` in order to simplify logic in the controller.

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/applications/providers/util/util.go
+++ b/pkg/applications/providers/util/util.go
@@ -49,7 +49,14 @@ func CleanUpHelmTempDir(cacheDir string, logger *zap.SugaredLogger) {
 }
 
 // StatusUpdater is a function that postpone the update of the applicationInstallation.
+// It used to set status's filed of a specific template Porvider (eg status.HelmRelease).
 type StatusUpdater func(status *appskubermaticv1.ApplicationInstallationStatus)
+
+// NoStatusUpdate is a StatusUpdater that does not update the status.
+// It useful in case an error happens and we don't have information to update the status.
+var NoStatusUpdate StatusUpdater = func(status *appskubermaticv1.ApplicationInstallationStatus) {
+	// NO OP
+}
 
 // GetCredentialFromSecret get the secret and returns secret.Data[key].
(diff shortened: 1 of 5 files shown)


pkg/applications/providers/util/util.go
```diff
@@ -49,8 +49,15 @@ func CleanUpHelmTempDir(cacheDir string, logger *zap.SugaredLogger) {
 }
 
 // StatusUpdater is a function that postpone the update of the applicationInstallation.
+// It used to set status's filed of a specific template Porvider (eg status.HelmRelease).
```
SimonTheLeg, line 52:
```suggestion
// It is used to set status's filed of a specific template Porvider (eg status.HelmRelease).
```

pkg/applications/providers/util/util.go
```diff
@@ -49,8 +49,15 @@ func CleanUpHelmTempDir(cacheDir string, logger *zap.SugaredLogger) {
...
 type StatusUpdater func(status *appskubermaticv1.ApplicationInstallationStatus)
 
+// NoStatusUpdate is a StatusUpdater that does not update the status.
+// It useful in case an error happens and we don't have information to update the status.
```
SimonTheLeg, line 56:
```suggestion
// It is useful in case an error happens and we don't have information to update the status.
```

SimonTheLeg (approved):
tiny nitpicks from my side. PR looks good to me

PR #10687, opened 2022-08-06
fix kc-status controller flipping the variables

xrstf:
This fixes the Status looking like this right now:

```yaml
status:
  kubermaticEdition: v2.21.0-alpha.1
  kubermaticVersion: EE
```

```release-note
NONE
```


embik (approved):
/approve

PR #10688, opened 2022-08-06
fix missing Registry overwrite in Canal 3.23 addon

xrstf:
It seems to be an oversight that we didn't use `Registry` in the 3.23 manifests. I also switched to quay.io, like we did in #10305. The `calico/node:v3.23.0` image is available as multiarch on quay.

```release-note
NONE
```


rastislavs:
Thanks @xrstf !
CC @sachintiptur 

/lgtm
/approve

PR #10718, opened 2022-08-10
Add user-cluster manager to user-cluster-webhook

### LINKED ISSUE #10716: Machines seems to be not validated by the user-cluster-webhook
### What happened?

I tried to create a new cluster after using all of my quota assigned to the project and it succeeded.

### Expected behaviour

I should be blocked at some point due to lack of resources in my project.

1. Create new project.
2. Go to admin panel and create resource quota for the new project with 2 CPUs limit.
3. Go back to project and provision first cluster allocating 2 CPUs
4. After first cluster is created, create another one.

- KKP version: 2.21 (dev)

### Additional information

Looks like user cluster webhook responsible for validating machines doesn't work. I checked that by tying to hit `/validate-cluster-k8s-io-v1alpha1-machine` resulting in 404 message.

pkprzekwas:
Due to introduced architecture we are forced to start two managers (one for each cluster: seed, user) in a single process of `user-cluster-webhook`. Without that, webhook for validation resource quota in machines doesn't work, which makes seanse as webhook server in which we register validator never starts.

Fixes #10716 

**What type of PR is this?**
/kind bug

Port for additional server has been chosen without a deep reflection. If anyone have better idea what ports we may use please feel free to chime in.

```release-note
NONE
```

### DIFF
--- a/cmd/user-cluster-webhook/main.go
+++ b/cmd/user-cluster-webhook/main.go
@@ -99,7 +99,7 @@ func main() {
 		log.Fatalw("Failed to create the user cluster manager", zap.Error(err))
 	}
 
-	// apply the CLI flags for configuring the webhook servers
+	// Apply the CLI flags for configuring the webhook servers.
 	if err := options.seedWebhook.Configure(seedMgr.GetWebhookServer()); err != nil {
 		log.Fatalw("Failed to configure webhook server", zap.Error(err))
 	}
@@ -124,10 +124,10 @@ func main() {
 	// /////////////////////////////////////////
 	// setup webhooks
 
-	// Setup the validation admission handler for ApplicationInstallation CRDs
+	// Setup the validation admission handler for ApplicationInstallation CRDs in seed manager.
 	applicationinstallationvalidation.NewAdmissionHandler(seedMgr.GetClient()).SetupWebhookWithManager(seedMgr)
 
-	// Setup Machine Webhook
+	// Setup Machine Webhook in user manager.
 	machineValidator, err := machinevalidation.NewValidator(seedMgr.GetClient(), userMgr.GetClient(), log, options.caBundle, options.projectID)
 	if err != nil {
 		log.Fatalw("Failed to setup Machine validator", zap.Error(err))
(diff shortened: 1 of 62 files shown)


cmd/user-cluster-webhook/main.go
```diff
@@ -77,9 +77,7 @@ func main() {
 		log.Fatalw("Failed to get user cluster kubeconfig")
 	}
 
-	// /////////////////////////////////////////
-	// create manager
-
+	// This case unusual in a sense that we need to configure and start managers for both user and seed clusters.
```
xrstf: This is missing an "is" ("This case IS unusual") and also, it's generally not _that_ unusual. Our master-ctrl-mgr does the same for master/seed clusters, even dynamically managing Managers at runtime (whenever new Seed objects appear or are updated). So this all is not such a crazy thing as you might think it is 😁

cmd/user-cluster-webhook/main.go
```diff
@@ -101,28 +99,35 @@ func main() {
...
+	addAPIs(userMgr.GetScheme(), log)
 
 	// /////////////////////////////////////////
 	// add pprof runnable, which will start a websever if configured
 
 	if err := seedMgr.Add(&options.pprof); err != nil {
-		log.Fatalw("Failed to add the pprof handler", zap.Error(err))
+		log.Fatalw("Failed to add the pprof handler (seed manager)", zap.Error(err))
+	}
+	if err := userMgr.Add(&options.pprof); err != nil {
```
xrstf: Does this make sense? We use the manager as a simple way to start the pprof stuff, but even with many managers there is no need to have multiple pprof things running in the same Go process, as they have no relation to the manager that started them.

pkg/resources/resources.go
```diff
@@ -933,6 +933,8 @@ const (
 	UserClusterWebhookDeploymentName        = "usercluster-webhook"
 	UserClusterWebhookServiceName           = "usercluster-webhook"
 	UserClusterWebhookServingCertSecretName = "usercluster-webhook-serving-cert"
+	UserClusterWebhookSeedListenPort        = 443
```
xrstf, line 936: Can we use non-privileged ports (i.e. >=1024)? That would not require devs to be root when testing the webhook locally. Maybe 6443 and 7443?

cmd/user-cluster-webhook/main.go
pkprzekwas: I though that relation between pprof things and managers has to be 1 to 1, but it makes sense to remove one from what you've explained.

xrstf (approved):
/approve

🤞

PR #10721, opened 2022-08-10
Update metering to version 1.0

### LINKED ISSUE #9856: Update Metering
Metering was updated to a new major version which changes the deployment and introduce a dependency to prom

Acceptance criteria:
Update Metering Deployment to the new Version

mate4st:

Removes  the legacy components of metering and deploys version 1.0 of metering.
This also introduce multiple types of reports, cluster and namespaced based. 
By default both types are generated, but it can be configured per report configuration. 

Fixes #9856 

**What type of PR is this?**
/kind feature

Since all data of legacy metering installation are synced in 5minutes intervals, we can safely delete the existing pvc.

```release-note
[Breaking] Update Metering to version 1.0. This changes the report csv format and the method of data collection. All previous generated reports will be still accessible via the dashboard. The new metering has a slightly different format and uses a different data source. This means that data collection will start from the beginning at the time of upgrading. Going back in time is not possible due to the change of the data source.
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1160
```


xrstf:
> The new metering installation should be in its own namespace

.... because? :) Please describe your reasonings for future reference.

WeirdMachine:
> > The new metering installation should be in its own namespace
> 
> .... because? :) Please describe your reasonings for future reference.



> > The new metering installation should be in its own namespace
> 
> .... because? :) Please describe your reasonings for future reference.

I think it make sense to have it in its own namespace, because its a more or less independent feature that could be seperated, similar to mla. However during testing I found out that this breaks reconcile where we use the seed owner reference because its not allowed to be cross namespaces. Therefore I reverted the namespace seperation and moved metering back to kubermatic namesapce.

pkg/ee/metering/prometheus/clusterole.go
```diff
@@ -0,0 +1,62 @@
...
+package prometheus
+
+import (
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// clusterRoleBindingCreator create a cluster role binding for the metering tool.
```
xrstf: Wrong comment

pkg/ee/metering/prometheus/clusterrolebinding.go
```diff
@@ -22,28 +22,33 @@
...
+package prometheus
 
 import (
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
 	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
 
 	rbacv1 "k8s.io/api/rbac/v1"
 )
 
 // clusterRoleBindingCreator create a cluster role binding for the metering tool.
```
xrstf: Wrong comment

pkg/ee/metering/prometheus/configmap.go
```diff
@@ -0,0 +1,514 @@
...
+package prometheus
+
+import (
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// prometheusStatefulSe creates the service account used by prometheus.
```
xrstf: Wrong comment.

pkg/ee/metering/prometheus/reconcile.go
```diff
@@ -0,0 +1,91 @@
...
+	Namespace = resources.KubermaticNamespace
+)
+
+// ReconcilePrometheus reconciles the prometheus instance used as a datasource for metering.
+func ReconcilePrometheus(ctx context.Context, client ctrlruntimeclient.Client, scheme *runtime.Scheme, getRegistry registry.WithOverwriteFunc, seed *kubermaticv1.Seed) error {
+	seedOwner := common.OwnershipModifierFactory(seed, scheme)
+
+	if err := persistentVolumeClaimCreator(ctx, client, seed); err != nil {
+		return fmt.Errorf("failed to reconcile PVC: %w", err)
+	}
```
xrstf: Why do we manually create a single PVC that is then used by potentially multiple Prometheus pods from the STS? Shouldn't the STS simply have a PVC template attached?

pkg/ee/metering/prometheus/svc.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"k8c.io/kubermatic/v2/pkg/controller/operator/common"
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// prometheusStatefulSe creates the service account used by prometheus.
```
xrstf: Wrong comment.

pkg/ee/metering/reconcile.go
```diff
@@ -179,50 +176,115 @@ func cleanupOrphanedReportingCronJobs(ctx context.Context, client ctrlruntimecli
...
-		return fmt.Errorf("failed to cleanup metering Deployment: %w", err)
+// fetchExistingReportingCronJobs returns a list of all existing reporting cronjobs.
+func fetchExistingReportingCronJobs(ctx context.Context, client ctrlruntimeclient.Client) (*batchv1.CronJobList, error) {
+	existingReportingCronJobs := &batchv1.CronJobList{}
+	listOpts := []ctrlruntimeclient.ListOption{
+		ctrlruntimeclient.InNamespace(meteringNamespace),
+		ctrlruntimeclient.ListOption(ctrlruntimeclient.MatchingLabels{common.ComponentLabel: meteringName}),
+	}
+	if err := client.List(ctx, existingReportingCronJobs, listOpts...); err != nil {
+		if !apierrors.IsNotFound(err) {
```
xrstf, line 186: Can a 404 really happen when listing CronJobs?

pkg/ee/metering/reconcile.go
```diff
@@ -179,50 +176,115 @@ func cleanupOrphanedReportingCronJobs(ctx context.Context, client ctrlruntimecli
...
+			return nil, fmt.Errorf("failed to list reporting cronjobs: %w", err)
+		}
 	}
+	return existingReportingCronJobs, nil
+}
+
+// undeploy removes all metering components expect the namespace and pvc.
```
xrstf: s/expect/except/

pkg/ee/metering/reconcile.go
```diff
@@ -179,50 +176,115 @@ func cleanupOrphanedReportingCronJobs(ctx context.Context, client ctrlruntimecli
...
+	}
+	if err := cleanupResource(ctx, client, key, &rbacv1.ClusterRoleBinding{}); err != nil {
+		return fmt.Errorf("failed to cleanup metering pull secret: %w", err)
+	}
+	if err := cleanupResource(ctx, client, key, &rbacv1.ClusterRole{}); err != nil {
+		return fmt.Errorf("failed to cleanup metering pull secret: %w", err)
+	}
+	if err := cleanupResource(ctx, client, key, &corev1.ServiceAccount{}); err != nil {
+		return fmt.Errorf("failed to cleanup metering pull secret: %w", err)
+	}
```
xrstf: All of those error messages are wrong.

pkg/ee/metering/reconcile.go
mate4st: I assume this happens when parallel reconcile is enabled or someone manually deleting it at the same time. However we can always ignore this error.

pkg/ee/metering/prometheus/reconcile.go
mate4st: Did so. Is there anything i need to take in consideration when using those?

pkg/apis/kubermatic/v1/datacenter.go
```diff
@@ -773,6 +773,9 @@ type SeedMLASettings struct {
 // MeteringConfiguration contains all the configuration for the metering tool.
 type MeteringConfiguration struct {
 	Enabled bool `json:"enabled"`
+
+	// +kubebuilder:default=kubermatic-fast
```
xrstf: Let's not hardcode this value into our CRDs. If anything we should do the defaulting in our own code and make this maybe dependent on the KubermaticConfiguration. "kubermatic-fast" was also originally only meant for SSD volumes for etcds, not for "everthing that KKP stores", so `kubermatic-fast` might even be the wrong default here.

I would just use the default StorageClass from the seed cluster, unless the admin configures a special SC in the Seed datacenter.

pkg/ee/metering/prometheus/reconcile.go
```diff
@@ -0,0 +1,87 @@
...
+	"k8c.io/kubermatic/v2/pkg/resources/reconciling"
+	"k8c.io/kubermatic/v2/pkg/resources/registry"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	Name      = "metering-prometheus"
+	Namespace = resources.KubermaticNamespace
```
xrstf: Please use the `seed.Namespace` instead of hardcoding it into the controller.

xrstf (approved):
/approve

pkprzekwas:
In order to add support for types in the dashboard we need to extend report configuration handler, but it can be done in a separate PR.

pkprzekwas (approved):

pkg/ee/metering/cronjob.go
```diff
@@ -38,136 +40,33 @@ import (
...
+				job.Labels = make(map[string]string)
 			}
-			labels[LabelKey] = reportName
-			job.SetLabels(labels)
+			job.Labels[common.NameLabel] = reportName
+			job.Labels[common.ComponentLabel] = meteringName
 
 			job.Spec.Schedule = mrc.Schedule
-			job.Spec.JobTemplate.Spec.Parallelism = pointer.Int32Ptr(1)
-			job.Spec.JobTemplate.Spec.Template.Spec.ServiceAccountName = meteringToolName
```
wozniakjan, line 53: I think this causes an issue for upgraded clusters, the fact `ServiceAccountName` is not explicitly removed causes old `CronJobs` to contain the old value and that `ServiceAccount` is purged as a legacy resource in the mechanism introduced in this PR.
https://github.com/kubermatic/kubermatic/blob/8fd688856e73375558ecb9969e34388158d0aff6/pkg/ee/metering/reconcile.go#L251-L253

PR #10720, opened 2022-08-10
Upgrade gimps to 0.5.0

hdurand0710:
Signed-off-by: Helene Durand <helene@kubermatic.com>

**What this PR does / why we need it**:
With gimps 0.4.1, we have the following error:
`% ./hack/verify-import-order.sh
[2022-08-10T15:59:47+02:00] Sorting import statements...
2022/08/10 15:59:53 Cannot check if file "/Users/helenedurand/go/src/github.com/k8c.io/kubermatic/pkg/handler/v2/external_cluster/aks.go" is generated: failed to parse code: 671:264: expected type, found ',' (and 10 more errors)`
Working fine with latest 0.5.0

**Which issue(s) this PR fixes**:

Fixes #

**What type of PR is this?**

/kind bug

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

PR #10723, opened 2022-08-11
KubeVirt Fix DataVolume usage at cluster creation time

### LINKED ISSUE #10699: KubeVirt impossible to use a Pre-allocated DV at cluster creation
### What happened?
Refer to https://github.com/kubermatic/kubermatic/issues/10486
Create a KubeVirt cluster and create a Pre-allocated DV `test-padv` at cluster creation time (step `Settings`).
At the next step `Initial Nodes`, it's impossible to use this pre-allocated DataVolume as the name must be of the form: `<namespace>/<dv name>` where namespace is the dedicated infra cluster namespace which is a dynamically allocated name `<cluster-xyz>`.

### Expected behavior
It should be possible to use the pre-allocated DataVolume at cluster creation time.

- KKP version: 2.21

<details>

```yaml
# paste manifest here
```

</details>

KubeVirt

Ubuntu 

### Additional information
The fix will consist in:
- in the UI, step `Initial Nodes`, only enter the DataVolume Name (same as set up at the previous `Settings` step)
- in KKP, prefix the DataVolume Name with the namespace <cluster-xyz/> name.

hdurand0710:

This PR allows, at cluster creation time, to create a DataVolume _(Settings tab)_ and use it as source of DataVolume cloning _(Initial Nodes tab)_.
It was not possible as the DataVolume name expected in the  _(Settings tab)_  was expected to be formatted: <namespace>/<name> with <namespace>= the KubeVirt infra cluster dedicated name, which is the cluster.Status.NamespaceName generated after.

The change is to have only the DataVolume name in the _(Settings tab)_ , and prefix it with the dedicated namespace in KKP when creating the initialNodeDeployment. 
We keep the format namespace/DataVolumeName in the RawConfig so we can extract the namespace and the name in the machine-controller to be able to clone with a source in a different namespace (even though it's not possible right now due to policies, we might work on this on the next release). 

For this release:
- KKP prefixes the DataVolumeName with the dedicated namespace. If the name is already with the format namespace/name, we keep it (but cloning will fail due to policies).
- machine-controller extracts namespace and name from this name for the DataVolume to clone.

Fixes #10699

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
TBD
The documentation will be done right after this PR, part of #10705.
```

### DIFF
--- a/cmd/kubermatic-api/swagger.json
+++ b/cmd/kubermatic-api/swagger.json
@@ -28470,7 +28470,7 @@
           "x-go-name": "PodAntiAffinityPreset"
         },
         "primaryDiskOSImage": {
-          "description": "PrimaryDiskOSImage states the source from which the imported image will be downloaded.",
+          "description": "PrimaryDiskOSImage states the source from which the imported image will be downloaded.\nThis field contains:\na URL to download an Os Image from a HTTP source.\na DataVolume Name as source for DataVolume cloning.",
           "type": "string",
           "x-go-name": "PrimaryDiskOSImage"
         },
--- a/pkg/resources/machine/common.go
+++ b/pkg/resources/machine/common.go
@@ -19,6 +19,8 @@ package machine
 import (
 	"errors"
 	"fmt"
+	"net/url"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 
@@ -508,7 +510,7 @@ func getGCPProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *ku
 	return EncodeAsRawExtension(config)
 }
 
-func GetKubevirtProviderConfig(_ *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*kubevirt.RawConfig, error) {
+func GetKubevirtProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*kubevirt.RawConfig, error) {
 	config := &kubevirt.RawConfig{
 		VirtualMachine: kubevirt.VirtualMachine{
 			Flavor: kubevirt.Flavor{
@@ -523,7 +525,7 @@ func GetKubevirtProviderConfig(_ *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec,
 						Size:             providerconfig.ConfigVarString{Value: nodeSpec.Cloud.Kubevirt.PrimaryDiskSize},
 						StorageClassName: providerconfig.ConfigVarString{Value: nodeSpec.Cloud.Kubevirt.PrimaryDiskStorageClassName},
 					},
-					OsImage: providerconfig.ConfigVarString{Value: nodeSpec.Cloud.Kubevirt.PrimaryDiskOSImage},
+					OsImage: providerconfig.ConfigVarString{Value: extractKubeVirtOsImageURLOrDataVolumeNsName(cluster.Status.NamespaceName, nodeSpec.Cloud.Kubevirt.PrimaryDiskOSImage)},
 				},
 			},
 			DNSPolicy: providerconfig.ConfigVarString{Value: dc.Spec.Kubevirt.DNSPolicy},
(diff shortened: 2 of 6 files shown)


hack/verify-import-order.sh
```diff
@@ -20,11 +20,11 @@ cd $(dirname $0)/..
 source hack/lib.sh
```
mfranczy: I belive you didn't want to push this file, right?

cmd/kubermatic-api/swagger.json
```diff
@@ -28470,7 +28470,7 @@
           "x-go-name": "PodAntiAffinityPreset"
         },
         "primaryDiskOSImage": {
-          "description": "PrimaryDiskOSImage states the source from which the imported image will be downloaded.",
+          "description": "PrimaryDiskOSImage states the source from which the imported image will be downloaded.\nFor now (might change later on) the Os Image contains:\na URL.\na DataVolume Name (not prefixed with namespace/).\nThe DataVolume name will be prefixed with the dedicated infra cluster namespace name (cluster-xyz/datavolumename)\nbefore storage in the provider config. It is then splitted in machine-controller to get the namespace and the name\nof the DataVolume to clone from.\nWe keep for now this mechanism to be able to clone the Data Volume in the future from different namespaces.",
```
mfranczy: I would remove this heads-up about temporary change
> For now (might change later on) the Os Image contains:\na URL.\na DataVolume Name (not prefixed with namespace/).\nThe DataVolume name will be prefixed with the dedicated infra cluster namespace name (cluster-xyz/datavolumename)

The plan is that we will have a drop down field for the UI so in the future users won't have to type anything in there. I think it's just ok to say that over this field we download an image from http source or we use data volume cloning.

pkg/machine/convert_test.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
mfranczy: 2022 :)

pkg/machine/convert.go
```diff
@@ -375,6 +377,24 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 	return cloudSpec, nil
 }
 
+func extractKubeVirtOsImageURLOrDataVolumeName(osImage string) string {
```
mfranczy: we already have the `extractKubeVirtOsImageURLOrDataVolumeNsName` function in `common`, cannot we merge those two? They works almost the same. You could return more values from the function if needed.

mfranczy (changes requested):

hack/verify-import-order.sh
hdurand0710: Yes, I did want to push it. 
I need to rebase, I did a specific PR yesterday for that. https://github.com/kubermatic/kubermatic/pull/10720.
Once rebased, it will go away.

pkg/machine/convert.go
mfranczy: I see you cannot return more values, but you can merge functionality of those functions anyway.

cmd/kubermatic-api/swagger.json
hdurand0710: Good point, I've updated accordingly.

hdurand0710:
@mfranczy , I pushed with all comments expect the merge of the `extract` methods as discussed, so you can have the latest version for further review.

pkg/machine/convert.go
mfranczy: As agreed, we will remove function `extractKubeVirtOsImageURLOrDataVolumeName` and we will allow to inject the cluster namespace afterwards. It's better option than introducing part of the code that will have to be removed in the next release due to our plans.

hdurand0710:
@mfranczy as discussed, I removed the `convert` part. 
When updating the MD, we now have the `namespace/dvname` displayed, working well, and that will be compliant with the next release where we can have a different namespace and provide it at cluster creation.

mfranczy (approved):
/approve
/lgtm

mfranczy (approved):
/approve
/lgtm

pkg/resources/machine/common.go
```diff
@@ -554,8 +556,24 @@ func GetKubevirtProviderConfig(_ *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec,
...
+	// config.VirtualMachine.Template.PrimaryDisk.OsImage.Value contains:
+	// - a URL
+	// - or a DataVolume name
+	// If config.VirtualMachine.Template.PrimaryDisk.OsImage.Value is a DataVolume, we need to add the namespace prefix
+	if _, err := url.ParseRequestURI(osImage); err == nil {
+		return osImage
+	}
+	// It's a DataVolume
+	// If it's already a ns/name keep it.
+	if nameSpaceAndName := strings.Split(osImage, "/"); len(nameSpaceAndName) >= 2 {
```
pkprzekwas, line 569: nit: wouldn't it be safer to use namespace name in the split function? I'm thinking about a situation where `osImage` would contain `/` in its name which is probably not possible.
```suggestion
	if nameSpaceAndName := strings.Split(osImage, namespace); len(nameSpaceAndName) >= 2 {
```

ahmedwaleedmalik (approved):
/approve ⚡

PR #10729, opened 2022-08-11
Add ClusterRole for listing namespaces in RBAC addon

ahmedwaleedmalik:

In KKP, by default when a user cluster is created then the person who is creating it is assigned the `cluster-admin` role. This works fine as expected. 

Although we ran into a few issues when we removed the `cluster-admin` role and tried to restrict users on that cluster to specific namespaces. One of the issues was that the users were unable to see the namespaces that they had access to in the Kubernetes dashboard. This makes sense since the dashboard will trigger an API call to list all namespaces and fail because the user doesn't have that authority. For more info https://github.com/kubernetes/dashboard/issues/6785

```
"namespaces is forbidden: User \"alexilaiho@gmail.com\" cannot list resource \"namespaces\" in API group \"\" at the cluster scope"
```

For an example to demonstrate this issue, let's assign a user admin access to the `kube-system` namespace:
1. He is unable to list namespaces and will always be redirected to the `default` namespace. Which he doesn't have access to in this case
Screenshot 2022-08-02 at 9 40 16 PM

2. If he directly modifies the URL and sets the query parameter for `namespace` to `kube-system` then everything works fine for them.
Screenshot 2022-08-02 at 9 40 11 PM

This can be super confusing for a basic user since they have no way to find this out. The same issue exists when someone is using `kubectl` as well.

As a workaround, for better user experience, we have decided to add a new cluster role for the RBAC addon which will give permissions to all the users to list namespaces.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
ClusterRole to list namespaces is shipped with the RBAC addon.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/addons/rbac/list-namespaces.yaml
+++ b/addons/rbac/list-namespaces.yaml
@@ -1,0 +1,27 @@
+# Copyright 2022 The Kubermatic Kubernetes Platform contributors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik:
/hold

We will discuss this internally in `sig-cluster-management` weekly and then move forward with this.

PR #10737, opened 2022-08-11
prevent accidental invalid config files in installer

xrstf:
A lot of time could have been saved in #10589 if the installer simply rejected invalid config files. Sadly it was one of the few spots where we hadn't migrated to strict unmarshalling yet.

/kind bug

```release-note
The `kubermatic-installer` now rejects `--config` files that are not actually valid `KubermaticConfiguration` objects.
```


ahmedwaleedmalik (approved):
/approve

PR #10751, opened 2022-08-15
Add EnableIngressHostname and IngressHostnameSuffix options to Openstack cloud spec

### LINKED ISSUE #10706: Adding enable-ingress-hostname setup as a field or annotation

As a user cluster admin
I would like to configure enable-ingress-hostname in cloud config
in order to see the end-users' IP addresses.

### Solution details

- Given there's a field to the cluster object or an annotation I can configure enable-ingress-hostname for the user cluster
- That field or annotation will derive the value and hand it over to the cloud config

### Use cases

Being able to view the IP address of the end users enables features like setting default language for them based on their location.

embik:
The Openstack CCM has some workaround for using the PROXY protocol, as using that on a load balancer seems to break when you try to connect to the public IP from another node or pod in the cluster (so in-cluster components cannot talk to the service). The workaround seems to set up a dummy DNS name on the LoadBalancer service. That is configured by `enable-ingress-hostname` and `ingress-hostname-suffix`. See [upstream docs](https://github.com/kubernetes/cloud-provider-openstack/blob/v1.24.2/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#load-balancer) for details on these options.

Upstream might remove this in the future once Kubernetes fixes the root issue via [KEP-1860](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1860-kube-proxy-IP-node-binding), so we should be ready to deprecate and remove this option in the future.

Fixes #10706

**What type of PR is this?**
/kind feature

```release-note
Add EnableIngressHostname and IngressHostnameSuffix options for Openstack (enables workaround in Openstack CCM for PROXY protocol client IP preservation)
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

ahmedwaleedmalik (approved):
/approve 🦀

PR #10753, opened 2022-08-15
Ship default AddonConfigs with KKP

### LINKED ISSUE #5499: Move the configuration of the AddonConfigs to the main repo

xrstf:
KKP has shipped addons since forever, but we never shipped the accompanying `AddonConfig` objects, which make the addons useful / functional. Some addons require configuration values and without us providing a sane default, the addons are basically useless.

This PR implements shipping default `AddonConfigs`. To allow admins to reconfigure the addons, a label is used to indicate ownership. This gives us the freedom to update the AddonConfigs later with new defaults, but makes it still easy for admins to override the configs with their own. It's also nicer than keeping the `AddonConfigs` inline in our KubermaticConfiguration (like we do for the list of accessible Addons).

/kind feature

Fixes #5499

```release-note
KKP will create default `AddonConfig` objects for the addons it ships. To customize them, remove the `app.kubernetes.io/managed-by` label from an AddonConfig, after which it will no longer be reconciled by KKP.
```


ahmedwaleedmalik (approved):
/approve

@xrstf Please create a testing ticket for this.

PR #10756, opened 2022-08-15
Use rfc1035 for metering report name validation

mate4st:
Fixes a bug where report names are invalid and cause the controller to throw errors during CronJob creation.

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/retest
/approve

PR #10762, opened 2022-08-15
Introduce ExternalClusterBringYourOwnProvider, make CloudSpec non-nil

xrstf:
This PR aligns the concept of a CloudSpec from ExternalClusters with what we already have for "regular" Clusters. For those we have a "bringyourown" provider as a stand-in for a cluster where you, the user, bring everything.

So far for ExternalClusters, we had a similar concept of "kubeconfig-only" / "generic" clusters, for which we have only a kubeconfig and nothing else. This was indicated by setting the CloudSpec to nil, which caused a number of bugs already because I never assumed that the CloudSpec could be nil. Also this prevents us from showing a meaningful value in the "kubectl get externalclusters" for the provider name.

To fix this, I reworked the CloudSpec to always exist and added a new external cloud provider, also called "bringyourown", to simplify the code.

Existing external clusters are migrated during upgrades with the installer.

/kind cleanup

```release-note
The `CloudSpec` of `ExternalClusters` must be set and use the new `bringyourown` provider when previously no cloud provider was configured. This is identical to how regular `Clusters` behave. Existing `ExternalClusters` are automatically migrated when using the kubermatic-installer, manual setups need to manually set `spec.cloudSpec.providerName = "bringyourown"` and `spec.cloudSpec.bringyourown: {}` after the CRDs were updated.
```


xrstf:
@imharshita PTAL

ahmedwaleedmalik (approved):
/approve

mfranczy:
@xrstf could you change commits messages from `woops`, `fix` or `ocd` to something more descriptive or squash those commits? It's harder later to go over commit logs.

ahmedwaleedmalik (approved):
/approve

xrstf:
/hold

needs #10774

PR #10764, opened 2022-08-15
Fix GCE tests

xrstf:
For whatever reason, I overlooked this in #10758. Probably because I made the changes in the dashboard and then assumed I was "done with Kubermatic".

```release-note
NONE
```

### DIFF
--- a/hack/local/run-api-e2e.sh
+++ b/hack/local/run-api-e2e.sh
@@ -40,7 +40,7 @@ if [ -z "${VAULT_ADDR:-}" ]; then
   export VAULT_ADDR=https://vault.kubermatic.com/
 fi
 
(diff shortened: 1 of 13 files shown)


hack/local/run-api-e2e.sh
```diff
@@ -110,7 +110,7 @@ metadata:
   namespace: kubermatic
 spec:
   gcp:
-    serviceAccount: ${GOOGLE_SERVICE_ACCOUNT}
+    serviceAccount: "${GOOGLE_SERVICE_ACCOUNT}"
```
xmudrii, line 113: Why are those two not base64 encoded?

xrstf: They are, in the chunk above (line 43).

xmudrii (approved):
/lgtm
/approve

PR #10774, opened 2022-08-16
Stabilize API e2e Tests

xrstf:
This fixes a typo that hides the true error. It also changes the waiting logic to be more robust by including the superset check in the wait loop, so that additional (cluster)roles do not confuse the test. Otherwise we get false errors like

> expected cluster roles [cluster-admin edit view], got [cluster-admin edit list-namespaces view]

/kind cleanup

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10781, opened 2022-08-17
Do not install gimps. Only tell the user to install it.

hdurand0710:

Update the `verify-import-order.sh` to:
- detect the OS and ARCH to download the right zip
- re-install gimps if already installed but with a version different from the expected one.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/hack/verify-import-order.sh
+++ b/hack/verify-import-order.sh
@@ -29,25 +29,9 @@ i686) ARCH="386" ;;
 i386) ARCH="386" ;;
 esac
 
-version=0.5.0
-# Remove gimps if already installed but not with the right version
-if [ -x "$(command -v gimps)" ]; then
-  iversion="$(gimps --version | grep version)"
-  if [ ${iversion##version: } != $version ]; then
-    echodate "gimps version is not the expected one... removing it before re-install with the right version"
-    rm /usr/local/bin/gimps
-  fi
-fi
-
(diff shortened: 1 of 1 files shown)


xrstf:
So I wonder if we really want this complexity. Maybe having code to install gimps was a mistake, as for most others we simply rely on people having them installed (like golangci-lint, wwhrd). Maybe we should just error out if gimps is not available and link to Github and tell people to just install it.

hdurand0710:
Ok, I followed your advice.

hack/verify-import-order.sh
```diff
@@ -19,16 +19,19 @@ set -euo pipefail
...
-  mv gimps /usr/local/bin/
+ARCH=$(uname -m)
+case $ARCH in
+armv6*) ARCH="armv6" ;;
+aarch64) ARCH="arm64" ;;
+x86) ARCH="386" ;;
+x86_64) ARCH="amd64" ;;
+i686) ARCH="386" ;;
+i386) ARCH="386" ;;
+esac
```
xrstf: I don't think we need this anymore?

hdurand0710: oh shame ! coffee is needed... :-)

xrstf (approved):
/approve

PR #10783, opened 2022-08-17
Rollback kubeconfig base64 encoding for KubeVirt functional tests

mfranczy:
We don't have to encode kubeconfig for KubeVirt tests, since data is taken from a secret it's already base64 encoded.

This PR fixes KubeVirt tests.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/cmd/conformance-tester/pkg/scenarios/kubevirt.go
+++ b/cmd/conformance-tester/pkg/scenarios/kubevirt.go
@@ -18,7 +18,6 @@ package scenarios
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
 
 	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
@@ -52,7 +51,7 @@ func (s *kubevirtScenario) APICluster(secrets types.Secrets) *apimodels.CreateCl
 				Cloud: &apimodels.CloudSpec{
 					DatacenterName: secrets.Kubevirt.KKPDatacenter,
 					Kubevirt: &apimodels.KubevirtCloudSpec{
-						Kubeconfig: base64.StdEncoding.EncodeToString([]byte(secrets.Kubevirt.Kubeconfig)),
+						Kubeconfig: secrets.Kubevirt.Kubeconfig,
 					},
 				},
 				Version: apimodels.Semver(s.version.String()),
@@ -67,7 +66,7 @@ func (s *kubevirtScenario) Cluster(secrets types.Secrets) *kubermaticv1.ClusterS
 		Cloud: kubermaticv1.CloudSpec{
 			DatacenterName: secrets.Kubevirt.KKPDatacenter,
 			Kubevirt: &kubermaticv1.KubevirtCloudSpec{
-				Kubeconfig: base64.StdEncoding.EncodeToString([]byte(secrets.Kubevirt.Kubeconfig)),
+				Kubeconfig: secrets.Kubevirt.Kubeconfig,
 			},
 		},
 		Version: s.version,


hdurand0710:
Reason for double encoding is:
https://github.com/kubermatic/kubermatic/blob/11216c095ca55db6829e43f8ff180ed67d3f630b/pkg/provider/kubernetes/credentials.go#L143
When using the `.Data` field of the secret, the value will be base64 encoded in the Secret.

But in the tests, we call it from:
https://github.com/kubermatic/kubermatic/blob/11216c095ca55db6829e43f8ff180ed67d3f630b/cmd/conformance-tester/pkg/scenarios/kubevirt.go#L70
where we encode the kubeconfig value.

The intermediate value is passed here:
https://github.com/kubermatic/kubermatic/blob/11216c095ca55db6829e43f8ff180ed67d3f630b/pkg/provider/kubernetes/credentials.go#L459

That's why we end up in the tests in a double encoded kubeconfig value for:
- cluster credential secret
and all that is created out of that.
The machine-controller is able to able this, but not the kubevirt ccm.

mfranczy:
Too many delete clicks for cherry-pick ;). We don't have to backport this PR for 2.19 and 2.20. 
Those branches have a different problem.. The CCM version used there is reading namespace from the passed kubeconfig so e2e-kubevirt has to point to the kube-system namespace (we can upgrade KubeVirt CCM as well which I think we should do).

xrstf:
> since data is taken from a secret it's already base64 encoded.

I don't quite understand that. Are you referring to the base64 encoding that Kubernetes secrets always have? Because the conformance-tester does

https://github.com/kubermatic/kubermatic/blob/master/cmd/conformance-tester/pkg/types/secrets.go#L179

> s.Kubevirt.Kubeconfig = test.SafeBase64Decoding(string(content))

, I don't think the kubeconfig is already encoded.

mfranczy:
> I don't quite understand that. Are you referring to the base64 encoding that Kubernetes secrets always have? Because the conformance-tester does

Yes I was refering to this. Damn I missed the part with decoding. But then why it's working now. Let me debug this. Without that change we have double base64 encoded kubeconfig and KubeVirt tests fails because of it. The only difference between master and older release branches is that KubeVirt setup mode is switched to "kube".

hdurand0710:
The reason is here:
The intermediate value is passed here:
https://github.com/kubermatic/kubermatic/blob/11216c095ca55db6829e43f8ff180ed67d3f630b/pkg/provider/kubernetes/credentials.go#L459

We put in the `.Data` an already encoded value -> it gets double encoded.

xrstf:
> The previous encoding is done:
> ...
> where we encode the kubeconfig value.

But this is exactly the line that this PR undoes, so where is the base64 encoding now happening?

xrstf:
What exactly is the intended format of the `Cluster.spec.cloud.kubevirt.kubeconfig` field in `pkg/apis/` and what is the format in the KKP API? Should this field base base64 encoded or should it be plaintext? I was under the impression it should be base64 encoded. ^^

* We take base64 in the dashboard.
* The machine-controller expects the kubeconfig as base64.
* The GCP ServiceAccount is also always encoded as base64.
* Ergo my assumption that `pkg/apis/kubermatic/v1/Cluster.spec.cloud.kubevirt.kubeconfig` should also be encoded.

xrstf:
Also I'm sorry this refactoring caused such a confusion for now, but once the dust has settled, I think we will all be happy :)

mfranczy:
> I was under the impression it should be base64 encoded. ^^

You're correct it should be. Let me go through the whole implementation again and see where is the probelm.

mfranczy:
huh? Was that merged? :D Anyway that fixes the KubeVirt CI but let me understand why, If needed I will come with another one. I think it's related to the client we are using in tests. TBD

xrstf:
At this point I would try to create a kubevirt cluster on an existing KKP 2.20 setup and just see what exactly is going on ^^

hdurand0710:
I confirm that on live code (not e2e tests): kubeconfig is not base64 encoded when we create a cluster.
in pkg/provider/kubernetes/credentials.go
`	spec := cluster.Spec.Cloud.Kubevirt`

```
2022-08-18T11:06:52.103+0200	error	kubernetes/credentials.go:448	HELENE2 cluster.Spec.Cloud.Kubevirt.KubeConfig apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUMvakNDQWVhZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwKY201bGRHVnpNQjRYRFRJeU1EUXdPREE1TkRRME0xb1hEVE15TURRd05UQTVORFEwTTFvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBT2JnCmlSazdIVjlIZVB4Nlp3SkdXcmpTbkJlcHBtYmZqNmhHUWdha2NmNHJyQjhEQVpOb2VkcWQ5V0hLVzdLaVJVcUIKRmJjNVpodWtkV0ZRQ2RqSjhHWHdMU2wzOG1yTWpyT3ByNUxscWQxZGlOOWtrR1cyQzBNSTVLbEc3a3VvSGRNaAo4WGkzVytTRGNKTHZDL0lCRkZSRWFMUGl5R3V2OWx0aUNjU3BGVE9sRlJGM2diY2FLWUswUzFFZm1ncmRXdE5KCjMrZ25TUThWYXhKd0VVbWw2WDMveWZaVEJuMkpmcGxGU2Jic0xXTmdHakQwNDFzSkM1S200OTRKNmxFbTg5MEUKZ2dCVVRxOENabFhzZHpkNkFJSjRMNzJnVlpqTmVldXpYODdGd0IxWno4RkxkMGMrVEU4ZmhrTTUycklOSm93WApXcDZwVXp6T0JRd2ZmdVZCemRrQ0F3RUFBYU5aTUZjd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0hRWURWUjBPQkJZRUZCdUNHcDVHZ0psNXg5T1oyMjl2alFpUVBCYXRNQlVHQTFVZEVRUU8KTUF5Q0NtdDFZbVZ5Ym1WMFpYTXdEUVlKS29aSWh2Y05BUUVMQlFBRGdnRUJBR0F5N1IwRHRpeVpja1NMQkNyWQpBZ0JadmF4RzlaL1hqZXhDTy9LcjRQUWFXa1NjUzRJT0RUL0J5Y2x6dTQ4TSs4eXAyN1d1a0VIT0JJRlVudlI2CmRXNnZ1NUdkd0lIRUIwSFdPL2s5S21Eem1La2ZlTlVLNkFtUHI1dTcyeU5HUTdpd1Q4cUc1d3lvQytrRTVURkQKVXZ2d2t3WXM3RVNscE1hWFltVGpDWHg1NUlia2FSeGxoaU1JM2MzUXZUVjFuaG1od3ZjMFBWZFZiSUZndFBmYwpja2RnUXNYK3hxV3hmL3d6OE81Z2N0QitOVTJQb2tvTzBVNnMyRThXY1hyeXZBR1FZYjZTS1ZjOTN6dnlHU1g3CnhkbHRsZ25zeGFPb3pabnVWdy83TVBpS0ZRUzErU1J1T284SjY0Y1ZFejZWVktaNXJpcUcxVzdaZjkyWXNuTmsKaGRRPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
    server: https://136.144.49.35:6443
  name: kubevirt-setup
contexts:
- context:
...
```

hdurand0710:
The place where it's decoded to a clear value:
https://github.com/kubermatic/kubermatic/blob/master/pkg/provider/cloud/kubevirt/provider.go#L81

This is wrong, we should not mutate a field in a `Validate`

mfranczy:
> I confirm that on live code (not e2e tests): kubeconfig is not base64 encoded when we create a cluster.

But input data for cluster creation is kubeconfig that must be base64 encoded, later is decoded.

> This is wrong, we should not mutate a field in a Validate

Totally agree with that, I see this code is very old. Additionaly it changes format of the value in the same field :-1: which is very confusing.

The whole refactoring of kubeconfig flow we must do in 2.22. Right now I am trying to get an answer about tests while it works now.

xrstf:
> Totally agree with that

It's impossible, but I agree _even more_ with this. I would have never expected this nugget in the kubevirt provider.

PR #10784, opened 2022-08-17
Remove report prefix from metering listing in api

mate4st:

Since the update of metering files are not prefixed with `report-` anymore. 
This is due to the fact that we do not need to distinguish between data and report csv files anymore. 

**What type of PR is this?**

/kind bug

```release-note
NONE
```

```documentation
NONE
```


xrstf (approved):
/approve

PR #10791, opened 2022-08-18
Fix metering reconcile

mate4st:

- Fix for reconcile metering prometheus sts.

- Fix for reconcile metering prometheus service. 

- As an addition the s3 buckets lifecycle configuration gets not updated if no retention is set. 
This prevents errors for s3 backends that do not support lifecycle configuration. 

- Adds dummy s3 metering secret to ci. 

/kind bug

```release-note
NONE
```

```documentation
NONE
```


xrstf (approved):
/approve

PR #10794, opened 2022-08-18
applicationDefinition: add defaultValues

### LINKED ISSUE #10664: Allowing for default values in ApplicationDefinition
After the first UI/UX workshops, we came to the conclusion, that it would be beneficial for KKP Admins to be able to set default values in an ApplicationDefinition, which then get shown in the values field for the user in the UI.

In the sig-app-mgmt meeting on 16th August, we decided to have these on a global level (per AppDef) as compared to per-version.
Our main arguments are that it is the simpler solution, it works in a majority cases, and that it is the defaults which can always be overridden. Plus we can also make it more adjustable in the future if we really need it. Furthermore we make the assumption that many community charts will keep their values structure

vgramer:

This PR add `defaultValues` in `applicationDefinition.spec`.  This field contains default values to show in the UI when creating an `applicationInstallation`

`applicationDefinition.spec.versions[].formFields` has been removed and replaced by `defaultValues`  for more fexibility

Fixes #10664

**What type of PR is this?**
/kind feature

```release-note
ApplicationDefinition can defined default values to show in the UI when creation an applicationInstallation
```

**Documentation**:

```documentation
TBD
```


ahmedwaleedmalik (approved):
/approve

PR #10795, opened 2022-08-18
Improve Seed cluster initialization, add ClusterInitialized condition

xrstf:
### What this PR does / why we need it

The KKP operator is supposed to handle the complete setup of KKP, so that adding a kubeconfig for a new cluster with nothing pre-installed will "just work".

To this end it needs to create the KKP/VPA/OSM CRDs, but for this to happen, the operator first had to wait for the seed-sync controller (in the master-ctrl-mgr) to do its job: The Deployment for the seed-ctrl-mgr is owned by the Seed object, so that object needs to exist first.

The reason the seed-sync is part of the master-ctrl-mgr is that it's a feature essential to _running_ KKP, not for setting it up. So even when the operator is broken or scaled down to 0, changes to Seeds must still be synced to their respective clusters.

However this caused an issue for new Seeds:

```
10 operator waits for seed-sync to create the Seed copy
20 seed-sync cannot do that until the operator creates the CRDs
30 GOTO 10
```

Additionally, since the seed-operator is trying to create watches inside the seed cluster, it will itself constantly explode, as the CRDs do not exist yet.

I first considered "hacking" this into the seed-operator code (basically adding lots of exceptions and corner cases), but abandoned this due to complexity and being a maintenance nightmare.

#### New `seed-init` controller

A new controller inside the operator has been added that has one simple task: Perform the initial, and only the initial setup of a seed cluster. This means setting up the CRDs and the very first copies of the KubermaticConfiguration and Seed objects.

After it has done so, it will set the `ClusterInitialized` condition on the `Seed` object, which is the new signal for every component in KKP that a seed can be worked with (it doesn't have to be fully reconciled yet, but at least static setups like creating watches will now work).

#### `ClusterInitialized` condition

This new condition is actually super handy. Instead of letting the seedlifecycle controller create controller instances for de-facto defunct seeds, it can now simply skip them entirely. This means that all controllers using the seedlifecycle will now not see seed cluster before they are ready.

This also means that the operator now no longer depends on the master-ctrl-mgr's seed-sync controller. The seed-sync is still vital for keeping the Seed/KC copies up-to-date, though.

/kind feature

```release-note
The KKP operator can now perform the complete initial setup for new seed clusters. If Minio/S3-exporter are not required, the KKP installer does not need to be used for setting up new / updating existing seed clusters.
```


embik (approved):
/approve

PR #10799, opened 2022-08-18
Metering: use reconciling framework, use seedsGetter more consistently

xrstf:
This does some quality of life changes to the new metering code.

* no more hardcoded "kubermatic" namespace
* use reconciling to manage secret
* use seedsGetter in more places

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/pkg/ee/metering/reconcile.go
+++ b/pkg/ee/metering/reconcile.go
@@ -51,7 +51,6 @@ import (
 )
 
 const (
-
 	// legacy naming.
 	meteringToolName = "kubermatic-metering"
 	meteringDataName = "metering-data"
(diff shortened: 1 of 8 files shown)


pkg/ee/metering/reconcile.go
```diff
@@ -238,8 +237,8 @@ func undeploy(ctx context.Context, client ctrlruntimeclient.Client, namespace st
 }
 
 // cleanupLegacyMeteringResources removes all active parts of the legacy metering installation.
-func cleanupLegacyMeteringResources(ctx context.Context, client ctrlruntimeclient.Client) error {
-	key := types.NamespacedName{Namespace: resources.KubermaticNamespace, Name: meteringToolName}
+func cleanupLegacyMeteringResources(ctx context.Context, client ctrlruntimeclient.Client, namespace string) error {
```
mate4st, line 245: We need to keep the hardcoded namespace here, as the legacy components were always deployed to kubermatic namespace. 

Ref: https://github.com/kubermatic/kubermatic/blob/release/v2.20/pkg/ee/metering/reconcile.go#L53

mate4st (changes requested):

pkg/ee/metering/reconcile.go
xrstf: Good catch!

ahmedwaleedmalik (approved):
/approve ⚡ 
/hold so that @WeirdMachine can take another look as well

mate4st (approved):

PR #10801, opened 2022-08-19
Set correct prefix for report listing

mate4st:

Fixes a bug where two report configuration with similar names interfere. Eg: daily would list daily-report.

/kind bug

```release-note
NONE
```

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10802, opened 2022-08-19
fix timing for OSM upgrade during KKP upgrade

xrstf:
During a KKP upgrade from 2.20 to 2.21 (on a shared master/seed system)

1. the `master-operator` will begin to reconcile the master cluster and in doing so, it will remove the `cluster-webhook` Service. From this point on, the cluster validation webhook is effectively dead. The service gets removed because we migrate to the new standalone kubermatic-webhook.
1. then the `seed-operator` tries to reconcile, but since it tried to update Cluster objects before fixing the webhook, it would end up in an endless loop or reconciling errors.

This PR simply moves the OSM stuff to happen later, after the webhook was reconciled.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -312,13 +312,6 @@ func (r *Reconciler) reconcileResources(ctx context.Context, cfg *kubermaticv1.K
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


xmudrii (approved):
/approve

ahmedwaleedmalik (approved):
/approve 🥲

PR #10803, opened 2022-08-19
don't spam misleading errors during upgrade

xrstf:
During the KKP upgrade it's perfectly normal and expected for the ReplicaSets to have no version label yet. This PR therefore reduces the log verbosity for this release, as hundreds of log messages can be super scary (even to me, and I implemented the log warning in the first place).

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10812, opened 2022-08-22
add info that release notes are being used for Changelog

SimonTheLeg:

Add a little info that the release-notes field is being used to generate the changelog

Fixes #

**What type of PR is this?**

/kind documentation

```release-note
none
```

**Documentation**:

```documentation
none
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -25,7 +25,7 @@ Optionally add one or more of the following kinds if applicable:
 
 **Does this PR introduce a user-facing change? Then add your Release Note here**:
 <!--
-Write your release note:
+Write your release note. Release notes are being used to generate the changelog:
 1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
 2. If no release note is required, just write "NONE".
 -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -25,7 +25,7 @@ Optionally add one or more of the following kinds if applicable:
 
 **Does this PR introduce a user-facing change? Then add your Release Note here**:
 <!--
-Write your release note:
+Write your release note. Release notes are being used to generate the Changelog:
```
embik: Uber nit: Changelog should be "changelog", it's not a title or something else that warrants a capital letter, just a normal noun.

embik (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
SimonTheLeg: that is a good point. I throught write it capitalized because of "the", but turns out that is not the case in English for regular nouns. Fixed

embik:
@csengerszabo I think you just adjusted PR templates to be the same across all projects. Should we add this to other projects' PR templates as well?

csengerszabo:
> @csengerszabo I think you just adjusted PR templates to be the same across all projects. Should we add this to other projects' PR templates as well?

That is true, we should keep them aligned to each other. Affected repositories:

- kubermatic
- kubeone
- dashboard
- metering
- mla
- machine-controller
- operating-system-manager
- start.kubermatic
- kubeone-for-devs

PR #10818, opened 2022-08-22
fix auto-update rule

xrstf:
The rule pointed to a version we don't actually support anymore.

/kind bug
/kind regression

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

embik (approved):

PR #10819, opened 2022-08-22
Include Equinix Metal as an alias for Packet to machine resource calculation

pkprzekwas:
We forgot that Packet's provider name has been changed to Equinix Metal. Update resource calculation adding Equinix.

**What type of PR is this?**
/kind bug

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10415, opened 2022-07-20
Enable OperatingSystemManager by default

ahmedwaleedmalik:
Signed-off-by: Waleed Malik <ahmedwaleedmalik@gmail.com>

**What does this PR do / Why do we need it**:
Operating System Manager is enabled by default and it's responsible for creating and managing the required configurations for worker nodes. Users can disable OSM per cluster basis.

Previously, since it was an experimental feature we had the `OperatingSystemManager` feature gate for it. But since it's considered GA now, we are going to remove that feature gate. 

**Does this PR close any issues?**:
Fixes #

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
* Operating System Manager is enabled by default and it's responsible for creating and managing the required configurations for worker nodes. 
* [ACTION REQUIRED] For existing clusters, users need to set `enableOperatingSystemManager` to true in the cluster spec to enable OSM. Existing `MachineDeployments` will not be rotated automatically. To use OSM for existing `MachineDeployments`, the user needs to update the `MachineDeployments` manually. An example for a somewhat benign change that could trigger rotation is to update the .spec.templates.metadata.annotations field of a MachineDeployment. This would result in the annotation being added to the machines and the machines would be rotated.
```

PR #10820, opened 2022-08-22
Delete cluster webhook before running migrations for OSM

ahmedwaleedmalik:

With 2.21 OSM will be enabled by default for new user clusters. Although for existing clusters a manual migration is required.

- https://github.com/kubermatic/kubermatic/pull/10415 took care of this migration but it was not tested properly, unfortunately. That change set didn't work because the `service` against the cluster admission webhooks wouldn't exist at the time of execution and thus all changes to clusters will effectively be rejected.
- https://github.com/kubermatic/kubermatic/pull/10802 was a follow-up PR to fix the issue described above. Although even that wouldn't work since the webhook explicitly denies requests where `enableOperatingSystemManager` is disabled after cluster creation.

To remediate this, in this PR we remove the cluster webhook altogether before performing this migration. This resolves both of the issues mentioned above. Webhook will be re-created after this migration by the seed operator https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/operator/seed/reconciler.go#L343

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -308,6 +308,13 @@ func (r *Reconciler) reconcileResources(ctx context.Context, cfg *kubermaticv1.K
 		return err
 	}
 
+	// This is a migration that is required to ensure that with the release of KKP v2.21
+	// we don't enable OSM for existing clusters.
+	// TODO: Remove this with KKP 2.22 release.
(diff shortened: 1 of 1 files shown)


xrstf (approved):
No idea if this is going to work, but since all OSM PRs land on my desk..

/approve

PR #10823, opened 2022-08-22
Add unit tests to ensure configured versions are coherent

xrstf:
This is a follow up to #10811 and #10818, ensuring that those issues will not surface again.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10826, opened 2022-08-22
changed failure policy for machine validating webhook to fail

lsviben:
Fixes/cleans-up the failurePolicy for the machine validationwebhookconfiguration which was set to Ignore during resource quota development. As now we support all providers, and the feature works, we can set it to Fail.

/kind bug
/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
/approve
/retest

PR #10832, opened 2022-08-23
Allow admins to interact with application installations

pkprzekwas:
At the moment API is throwing 403 whenever an administrator tries to list applicationinstallations of any cluster that doesn't belong to him. 

**What type of PR is this?**
/kind bug

```release-note
NONE
```


pkg/handler/v2/application_installation/application_installation.go
```diff
@@ -158,21 +158,36 @@ func UpdateApplicationInstallation(userInfoGetter provider.UserInfoGetter) endpo
...
 	}
 
+	if !userInfo.IsAdmin {
+		userInfo, err = userInfoGetter(ctx, projectID)
+		if err != nil {
+			return nil, err
+		}
+		cluster, err := clusterProvider.Get(ctx, userInfo, clusterID, &provider.ClusterGetOptions{})
+		if err != nil {
+			return nil, common.KubernetesErrorToHTTPError(err)
```
SimonTheLeg: I think we can move this part out of the if condition, as its being called both for admins and for non-admins, or?

pkg/provider/kubernetes/cluster.go
```diff
@@ -269,13 +269,19 @@ func (p *ClusterProvider) Get(ctx context.Context, userInfo *provider.UserInfo,
...
-		return nil, err
+
+	client := p.client
+
+	if !userInfo.IsAdmin {
+		var err error
+		client, err = createImpersonationClientWrapperFromUserInfo(userInfo, p.createSeedImpersonatedClient)
+		if err != nil {
+			return nil, err
+		}
```
SimonTheLeg: I have to admit, I do not fully understand what this code is doing. 

As I understand it, if the user is not an admin we overrwrite the client with one that is impersonated; otherwise you just get the normal client, which I assume has permissions to get all clusters?)

So would it not be better if `createImpersonationClientWrapperFromUserInfo` checks for if someone is admin and if so gives back and impersonation to do anything?

SimonTheLeg (commented):
some comments and questions from my side. I think it makes a lot of sense if @lsviben has a look as well. In addition to the review I also had a look at the implementation for [machinedeployments](https://github.com/kubermatic/kubermatic/blob/a7709145cff111aeb7011210a15fe33432fe38cf/pkg/handler/common/machine.go#L219), as they are similar, but now I am event more confused

pkg/handler/v2/application_installation/application_installation.go
pkprzekwas: Sure, but then I need to check if someone is an admin twice - which is fine I guess. I don't have a strong opinion here.

pkg/provider/kubernetes/cluster.go
pkprzekwas: > As I understand it, if the user is not an admin we overrwrite the client with one that is impersonated; otherwise you just get the normal client, which I assume has permissions to get all clusters?)

That's correct.

In my understanding the client that we assign here for admins (`p.client`) is the privileged one that is configured at the start of API - it's a master's admin token. It has nothing to do with impersonalisation as it is not assigned to any specific person. Impersonate client is created based on the data in OIDC token which is later used to map privileges in role bindings.

Regarding `createImpersonationClientWrapperFromUserInfo` function, it may not be simple to return impersonated client with all required permissions.

pkprzekwas:
> some comments and questions from my side. I think it makes a lot of sense if @lsviben has a look as well. In addition to the review I also had a look at the implementation for [machinedeployments](https://github.com/kubermatic/kubermatic/blob/a7709145cff111aeb7011210a15fe33432fe38cf/pkg/handler/common/machine.go#L219), as they are similar, but now I am event more confused

From what I see machine deployments use two types of providers - regular one, which uses impersonate client and privileged, which is using admin client.

pkg/provider/kubernetes/cluster.go
lsviben: instead of changing the cluster Get, use the GetUnsecured, which is intended for admins

SimonTheLeg: I see now.

pkprzekwas: Damn, it requires Project object. It's going to be some dependency hell but let me try.

pkprzekwas: Thanks for suggestion @lsviben. IMHO now it looks much better and what's more important doesn't modify the cluster provider.

PR #4627, opened 2019-11-05
get global settings endpoint

zreigz:
**What this PR does / why we need it**: get global settings endpoint.
This PR extends the user object for the admin flag. The admin user can get access to global settings.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #4625

```release-note
a new endpoint to retrieve global settings for admin:
GET /api/v1/admin/settings
```

PR #10838, opened 2022-08-23
Lowered default node count setting

lsviben:
Lowers the `defaultNodeCount` setting which is set upon KKP installation from 10 to 2. 10 seems like too much.

/kind cleanup
**What type of PR is this?**

```release-note
Lowered the default `defaultNodeSize` which is set in KubermaticSetting during KKP installation from 10 to 2. 
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
Just in case someone asks, the default of 10 was introduced in #4627 with no further comment.

/approve

PR #10849, opened 2022-08-24
KubeVirt: do not reconcile the namespace if the name is empty

### LINKED ISSUE #10843: KubeVirt: do not reconcile the namespace if name is empty
### What happened?
when creating a KubeVirt cluster, we get the error:
`failed cloud provider init: failed to reconcile Namespace: failed to ensure Namespace /: failed to get Object(*v1.Namespace): resource name may not be empty` 
a bunch of times.
We should not try to reconcile the namespace if the name is empty.

### Expected behavior

The mentioned error is not raised.

Just create a KubeVirt cluster.

- KKP version: 2.21

<details>

```yaml
# paste manifest here
```

</details>

KubeVirt

no impact.

hdurand0710:

Fixes #10843 

**What type of PR is this?**

/kind bug

When initialising the Cloud Provider and ensuring the creation of the namespace <cluster-xyz> in the infrastructure KubeVirt cluster, the reconcile was called several times until `cluster.Status.NamespaceName` is filled.
This was raising a bunch of times an error until this name is filled.

```release-note
KubeVirt: fix to ensure that we do not raise an error when reconciling the namespace in the infrastructure KubeVirt cluster, until we get the value of the namespace to create, avoiding transient errors.
```

**Documentation**:

```documentation
NONE
```
No documentation to updated, a bug fix.

### DIFF
--- a/pkg/provider/cloud/kubevirt/provider.go
+++ b/pkg/provider/cloud/kubevirt/provider.go
@@ -98,6 +98,12 @@ func (k *kubevirt) reconcileCluster(ctx context.Context, cluster *kubermaticv1.C
 		return cluster, err
 	}
 
+	// If the cluster NamespaceName is not filled yet, return a conflict error:
+	// will requeue but not send an error event
+	if cluster.Status.NamespaceName == "" {
+		return cluster, apierrors.NewConflict(kubermaticv1.Resource("cluster"), cluster.Name, fmt.Errorf("cluster.Status.NamespaceName for cluster %s", cluster.Name))
+	}
+
 	cluster, err = reconcileNamespace(ctx, cluster.Status.NamespaceName, cluster, update, client)
 	if err != nil {
 		return cluster, err


xrstf:
What about just returning nil? This is what most other controllers are doing when waiting for the cluster namespace?

ahmedwaleedmalik (approved):
/approve

hdurand0710:
On hold as it's not working as pushed.

hdurand0710:
The code for `cloud_controller`  is not like all other controller.
1- If we return (as the current PR version) : `cluster, nil` -> reconcile is succesful, namespace is not created until the next periodic reconcile.
2- if we return `nil, error` -> working but we get the error event in the UI 
3- if we return `nil, nil` we get an issue in the cloud_controller UpdateClusterStatus on a nil object would be….. not what we want -> need to change the `cloud_controller`


After discussion with @ahmedwaleedmalik :
1- a better fix would be to not econcile clusters in cloud controller manager if they don’t have `cluster.status.namespacedname` ->  we do not want to change this just before KKP 2.21 release
2- we will see for a patch release 2.21.1

-> adding KKP 2.22 milestone

xrstf:
> namespace is not created until the next periodic reconcile.

You're right, I forgot about the intentional slowdown we have in our cloud controller.

hdurand0710:
I've re-pushed with the original solution.
@ahmedwaleedmalik , @xrstf , what do we want to do ?

PR #10850, opened 2022-08-24
add semVer validation for AppDefs' version field

### LINKED ISSUE #10839: Discrepancy in the check between ApplicationDefinition and ApplicationInstallation
There is a discrepancy between the checks done in `ApplicationDefinition` and in `ApplicationInstallation`:
it's possible to create a `ApplicationDefinition` with `spec.versions.template.version`= `dev`
but when you want to use this version in the `ApplicationInstallation` you get the following error :

```
# applicationinstallations.apps.kubermatic.k8c.io "prometheus" was not valid:
# * spec.applicationRef.version: Invalid value: "dev": spec.applicationRef.version in body should match 'v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?'
#
```

The checks should be consistent between the 2 CRDs.

Additionally the fix for this should also include an update on [this docs page](https://docs.kubermatic.com/kubermatic/master/tutorials-howtos/applications/create-application-catalogue/), where the version `dev` is still being used

SimonTheLeg:
This PR makes two contributions:
1. Add semVer validation to the version field directly in the CRD
2. use semver.Version as the underlying type for versions

Fixes #10839 

**What type of PR is this?**

/kind bug

```release-note

```

**Documentation**:

```documentation

```

// TODO @SimonTheLeg create and link documentation PR


pkg/apis/apps.kubermatic/v1/application_definition.go
```diff
@@ -151,8 +151,12 @@ type ApplicationTemplate struct {
 }
 
 type ApplicationVersion struct {
-	// Version of the application (eg v1.2.3)
-	Version string `json:"version"`
+	// +kubebuilder:validation:Pattern:=v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?
```
xrstf, line 156: I really don't think we should copy/replicate some Go package's internal regex to parse version numbers. The same effect (rejecting invalid resources) can be achieved with webhooks as well.

pkg/apis/apps.kubermatic/v1/application_version.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	semverlib "github.com/Masterminds/semver/v3"
+)
+
+// +kubebuilder:validation:Type=string
+
+// Version wraps semverlib.Version. It is needed because kubebuilder does not accept structs with non-tagged fields, even if they have custom marshallers
+// With this the CRD resource will have Version as string but operator code can work directly with the semverlib.Version struct
+// (taken from https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/crd/testdata/cronjob_types.go#L283)
+type Version struct {
```
xrstf: What about using pkg/semver ?

SimonTheLeg: oh that actually looks like what I was about to build. Nice! :) I'll check it out. Do you by chance know if it supports apimachinery's [DeepEqual](https://github.com/kubernetes/apimachinery/blob/master/third_party/forked/golang/reflect/deep_equal.go#L272)?

I still have not figured out what DeepEqual actually calls on the objects that's why I am asking. Is it maybe just a func called `Equal`?

xrstf: I have never looked too deeply into our semver package. It existed since forever and just mostly works ^^

pkg/apis/apps.kubermatic/v1/application_definition.go
SimonTheLeg: yeah that is true. Since I'll have to touch the webhook anyways I am going to have a look. 

Initial idea was to make it clearer to the user that a valid semVer should be entered here. The other part is, it is essentially using the same regex as the masterminds/semver

But I'll have a look

xrstf: That is true, but I would consider the regex to be semver's implementation detail. What if at some point they update their parsing logic to not follow a regex, but be based on string splitting and stuff? Then we're suddenly left in the dark. A nice doc comment ("This must be a valid semver version, optionally with a leading 'v' prefix.") is much more helpful than presenting a long regex to the developer ^^

SimonTheLeg:
@xrstf @xmudrii I think the two of you are possibly the best people to judge 119e63a82f79f7b3b0cfffb7159bcd9fd513b54f

Essentially this commit makes a slight change to how we DeepCopy semVers. It is required because the old DeepCopy would cut off the "v" character of the string. The reason for that is in the underlying Masterminds/semver lib which does the same for its `.String()` func.
You can also practically try this out for yourself by reverting the changes in semver.go and run the unit test I added for semVer.

Please let me know if this change makes sense to your or if there are any known concerns.

pkg/controller/user-cluster-controller-manager/application-installation-controller/controller.go
```diff
@@ -212,7 +212,7 @@ func (r *reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, appI
 // getApplicationVersion finds the applicationVersion defined by appInstallation into the applicationDef and updates the struct appVersion with it.
 // An error is returned if the applicationVersion is not found.
 func (r *reconciler) getApplicationVersion(appInstallation *appskubermaticv1.ApplicationInstallation, applicationDef *appskubermaticv1.ApplicationDefinition, appVersion *appskubermaticv1.ApplicationVersion) error {
-	desiredVersion := appInstallation.Spec.ApplicationRef.Version.String()
+	desiredVersion := appInstallation.Spec.ApplicationRef.Version
 	for _, version := range applicationDef.Spec.Versions {
 		if version.Version == desiredVersion {
```
vgramer, line 217: should be `version.Version.Equal(desiredVersion)`

pkg/validation/application_installation.go
```diff
@@ -47,9 +47,9 @@ func ValidateApplicationInstallationSpec(ctx context.Context, client ctrlruntime
 
 	// Ensure that there is matching version defined in ApplicationDefinition
 	exists := false
-	desiredVersion := spec.ApplicationRef.Version.String()
+	desiredVersion := spec.ApplicationRef.Version
 	for _, version := range ad.Spec.Versions {
-		if version.Version == desiredVersion {
+		if *version.Version == *desiredVersion {
```
vgramer, line 52: same should use `Equal`

xrstf:
I kind of liked the normalisation that would happen when stringifying the version.

SimonTheLeg:
> I kind of liked the normalisation that would happen when stringifying the version.

I see, but in your opinion does this part change? I mean from my side I would argue that if you call `.String` on any semVer, it still normalizes it. But I feel like for a `DeepCopy`, it should copy the object as is and not transform it. Especially since it also is a guaranteed valid semVer before the transformation as well. 

Additionally I think this change increases the compatibility with other other third parties as well. For example if we normalize all of the semVers, we cannot use it in conjunction with tools like git or helm, who have their semVers as a string and do not do any normalization. I feel like both should be possible

xrstf:
I don't have strong feelings, I mostly fear that this change might break something somewhere down the line. But if others like th change, we can merge certainly merge it. I don't want to call the shots here.

vgramer:
I must admit I share the same fear as @xrstf.  Especially when I saw it introduce breaking changes in API.

Here one of the tests is falling in the `handler` package. 

```
k8c.io/kubermatic/v2/pkg/handler/v1/admin:TestGetAdmissionPluginEndpoint/scenario_3:_authorized_user_gets_plugin expand_less | 1s
-- | --
{Failed  admission_plugin_test.go:195: Expected response body to be  {"name":"eventRateLimit","plugin":"EventRateLimit","fromVersion":"1.13.0"}  got  {"name":"eventRateLimit","plugin":"EventRateLimit","fromVersion":"v1.13"}}
```

if the user (or dashboard) was expecting normalized SemVersions to be able to sort them, it will not work anymore.  The same goes if a user has a script that just does a string comparison.

xmudrii:
I agree with @xrstf and @vgramer. While, technically speaking, this might be a more correct approach because it's DeepCopy, I'm worried that we've got used to this behavior and we don't know what issues this change might cause. This doesn't mean we shouldn't change it, but we should somehow ensure that it's not going to break other stuff that uses the semver package. However, this might be harder, so I'm not sure if it's really worth the effort.

SimonTheLeg:
Yes you are right. I also think this is too risky to ship with v2.21. Exactly how you said with the already existing packages that use semver.

Going forward, I would propose the following:
1. I will close this PR for now. This way we can preserve our discussion, and it does not get mixed up with the stopgap solution (see bullet below). In addition I will create a new ticket in Application Backlog, so we can work on this topic after the release
2. I will create a new PR, which will just put the OpenAPI validation on the field. This will solve the [original issue](https://github.com/kubermatic/kubermatic/issues/10839). 

Now while the v2.21 state will not be super nice, the good news here is that even if we change the underlying type later, it will not affect the CRD spec and so has no impact on the end-user

vgramer:
@SimonTheLeg fine for me

PR #10857, opened 2022-08-25
Fix zero defaulting issue for ResourceQuotas

pkprzekwas:
Setting default values to zero blocks cluster creation if any of quota parameters is omitted in API request.

**What type of PR is this?**
/kind bug

```release-note
NONE
```


pkg/ee/resource-quota/handler.go
```diff
@@ -375,5 +375,17 @@ func convertToCRDQuota(quota apiv2.Quota) (kubermaticv1.ResourceDetails, error)
...
+		resourceDetails.CPU = nil
+	}
+	if resourceDetails.Memory.IsZero() {
+		resourceDetails.Memory = nil
+	}
+	if resourceDetails.Storage.IsZero() {
+		resourceDetails.Storage = nil
+	}
+
+	return *resourceDetails, nil
```
ahmedwaleedmalik: This seems a bit hackish to me. Shouldn't we just treat 0 as nil? For example, If I create the resource quota directly via CR and give `CPU` the value `0`. Then I'll end up with the same issue, right?

pkprzekwas: > If I create the resource quota directly via CR and give CPU the value 0. Then I'll end up with the same issue, right?

Yes. It actually may be a good idea as there is no value in distinguishing between `0` and `nil` unless we want to allow for blocking resource creation inside the project. But at this stage I don't want to risk changing CR and the way controllers interpret zero value  (cc @lsviben )

I'm currently testing changing values to pointers in the API model and replacing PATCH with PUT as an alternative solution.

ahmedwaleedmalik: I agree with you, for now, let's just settle with this.

ahmedwaleedmalik (approved):
/approve

pkg/ee/resource-quota/handler.go
pkprzekwas: I've replaced values with pointers and to make the API consistent with the behaviour under the hood introduced PUT in place of PATCH. I'll get rid of PATCH soon, but first I want to update the Dashboard, hence it got deprecated for now. Please take a look @ahmedwaleedmalik if possible.

pkprzekwas: oh, I see you already taken a glance :D

ahmedwaleedmalik: Yes :D

PR #10072, opened 2022-06-13
Add OIDC support to the K8S Dashboard integration

floreks:
**What does this PR do / Why do we need it**:
This PR changes the way Kubernetes Dashboard proxy integration works. It is now based on the OIDC flow and the authorization token returned from the identity provider. Thanks to the KKP Dashboard RBAC management it is now possible to dynamically configure per-user privileges inside K8S Dashboard.

**NOTE**: When this gets merged it will effectively break the `Open Dashboard` button in the KKP UI until other issues/PRs related to this epic will be done.

##### What's Changed
- Removed old dashboard proxy endpoints (ref: #10014):
  - `/api/v2/projects/{project_id}/clusters/{cluster_id}/dashboard/proxy`
  - `/api/v1/projects/{project_id}/clusters/{cluster_id}/dashboard/proxy`
- Added new endpoints to handle proxy calls
  - `api/v2/dashboard/login` - Handles the OIDC redirect and callback parts. Required parameters:
    - `projectID` and `clusterID` - used for the initial request to this endpoint
    - `state` and `code` - those are returned by the identity provider after successful authentication
  - `api/v2/projects/{project_id}/clusters/{cluster_id}/dashboard/proxy` - Required parameters:
    - `token` - bearer token retrieved from the identity provider after successful authentication
- Updated KKP - K8S Dashboard version compatibility matrix

##### Preview
https://user-images.githubusercontent.com/2285385/173351067-aeaf9695-47b4-4411-bebd-312329590936.mp4

**Does this PR close any issues?**:
Related to #9804
Fixes #10016
Fixes #10014

**Special notes for your reviewer**:
In order to test this locally:
1. Alter our dex configuration (`kubermaticIssuer` client) to whitelist the `redirectURI`: `http://localhost:8080/api/v2/dashboard/login`
  - `kubectl -n oauth edit configmap dex`
  - `kubectl -n oauth rollout restart deployment dex`
image
3. Create and wait for any cluster to start (using our dev env).
4. Run the API locally, i.e. `make clean && FEATURE_GATES=OIDCKubeCfgEndpoint=true KUBERMATIC_EDITION=ee KUBERMATIC_DEBUG=false ./hack/run-api.sh`
5. There are 2 ways to test the "frontend" part now:
  - run the KKP dashboard using `npm run start:local` after checking out this PR: https://github.com/kubermatic/dashboard/pull/4585
[...]

PR #10859, opened 2022-08-25
fix redirectURI being set for whole kubermatic-api instead of being used dynamically

### LINKED ISSUE #10825: Random failures with OIDC kubeconfig workflow
### What happened?

The OIDC Kubeconfig feature seems to be broken. While trying the workflow from the dashboard, we keep ending up with random errors, any one from this list:

```
- {"error":{"code":500,"message":"securecookie: base64 decode failed - caused by: illegal base64 data at input byte 3"}}
- {"error":{"code":400,"message":"incorrect value of state parameter: tmbhb7s9jj"}}
- {"error":{"code":400,"message":"incorrect value of cookie or cookie not set: http: named cookie not present"}}
- {"error":{"code":400,"message":"error while exchanging oidc code for token: oauth2: cannot fetch token: 400 Bad Request\nResponse: {\"error\":\"invalid_request\",\"error_description\":\"redirect_uri did not match URI from initial request.\"}"}}
```

Although without making any changes, sometimes it works.

Using our dashboard:
- Ensure that OIDC Kubeconfig is enabled in admin settings
- Create a Kubernetes Cluster
- Click on `Get Kubeconfig` to retrieve the OIDC Kubeconfig

Screenshot 2022-08-22 at 7 57 11 PM

Screenshot 2022-08-22 at 7 53 02 PM

- KKP version:

<details>

```yaml
# paste manifest here
```

</details>

lsviben:

Seems like this regression was introduced in  #10072 and continued  in #10102, where the redirectURI for the OIDC login (and later kubeconfig secret for webterminal) were being set globally for the whole kubermatic-api. 

As the kubeconfig OIDC endpoint expects to have its static redirectURI (which is set through the command line), when somebody accesses the dashboard, it changes the redirectURI for good, thus making the kubeconfig endpoint unusable.

This PR fixes that by dynamically setting the redirectURI for the OIDC dashboard and web-terminal requests. 

It would be good to clean this up more, for example the kubeconfig OIDC endpoint could do the same, and we could get rid of the static redirectURI that needs to be set in the command line or in the KubermaticConfiguration, but I will leave that for another time.

Fixes #10825 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/handler/auth/oidc.go
+++ b/pkg/handler/auth/oidc.go
@@ -64,12 +64,12 @@ type OIDCToken struct {
 type OIDCIssuerVerifier interface {
 	OIDCIssuer
 	TokenVerifier
-	RedirectURIPathSetter
+	RedirectURIPathGetter
 }
 
-type RedirectURIPathSetter interface {
-	// SetRedirectPath overrides configured path for a given URI
-	SetRedirectPath(path string) error
+type RedirectURIPathGetter interface {
+	// GetRedirectURI gets redirect URI for a given path
+	GetRedirectURI(path string) (string, error)
 }
 
 // OIDCIssuer exposes methods for getting OIDC tokens.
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
/approve

PR #10049, opened 2022-06-10
Add Azure CSI drivers (Azure Disk & Azure File)

embik:
**What does this PR do / Why do we need it**:

This adds Azure CSI drivers for both Azure Disk and Azure File. The default StorageClass is added via Azure Disk CSI.

By turning on CSI drivers for Azure, this PR also turns on CSI migration flags and deploys Azure clusters with external CCM + CSI drivers by default (this was blocked before as the CSI drivers were missing).

**Does this PR close any issues?**:
Fixes #9392

**Special notes for your reviewer**:

**Documentation**:

**Does this PR introduce a user-facing change?**:

```release-note
- Add CSI drivers for Azure Disk and Azure File
- New Azure clusters use external CCM by default
```

PR #10880, opened 2022-08-29
Re-add Azure CSI migration flags removed in CCM migration PR

embik:

I removed the flags in question by mistake in #10049. They need to be there because kube-controller-manager otherwise tries to migrate Azure disk and Azure file PVCs to the CSI drivers, but the drivers are only installed if a cluster was CCM migrated (which is not happening automatically).

Fixes #

**What type of PR is this?**
/kind bug
/kind regression

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
/approve

PR #10882, opened 2022-08-29
validate length of EAR encryption key

### LINKED ISSUE #10829: EAR enabled with invalid encryption key gets stuck in Pending state
If we enable EAR with an invalid value for encryption key say "password" then it gets in Pending state. 

In Pending state any changes are rejected. 

We should check for this invalid key earlier to avoid this situation. 

To reproduce: 

`kubectl  --type merge --patch-file patch.yaml` 

patch.yaml
```
spec:
  encryptionConfiguration:
    enabled: true
    resources:
      - secrets
    secretbox:
      keys:
        - name: encryption-key-2022-01
          value: password
```

PratikDeoghare:
Validates length of the encryption key used for EAR. Currently the length is not validated. 

Fixes #10829 

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/validation/cluster.go
+++ b/pkg/validation/cluster.go
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


PratikDeoghare:
Closing in favor of https://github.com/kubermatic/kubermatic/pull/10894 which includes these changes and more.

PR #10883, opened 2022-08-29
Move defaulting code

xrstf:
Previously, the defaulting code for Seed/KubermaticConfiguration objects lived in the _KKP Operator_ (`pkg/controller/operator/defaults`). This was nice while the operator was pretty much the only component that used KubermaticConfigurations.

However over time this has changed dramatically and nowadays most controllers in KKP make use of the `KubermaticConfigurationGetter`, which uses the defaulting code internally. Which means most of KKP's controllers and even the API depended on the operator code. Terrible.

So I planned on simply moving the defaulting code into the existing `pkg/defaulting`: `mv pkg/controller/operator/defaults/*.go pkg/defaulting/`.

However this unearthed some more import loops:

* `pkg/resources` was depending on `pkg/defaulting` and vice-versa.
* `pkg/provider` was depending on `pkg/defaulting` and vice-versa.

The provider<->defaulting loop was caused by the defaulting needing types like `provider.SeedGetter` and the provider needing the defaulting logic (as mentioned above, the `KubermaticConfigurationGetter` runs the defaulting logic).

I solved this loop by moving the _implementation_ for the getters into `pkg/provider/kubernetes`, while leaving the _types_ still in `pkg/provider` (like we do for all other providers).

Also, the helpers to determine cloud provider names (checking `cluster.Spec.Cloud.AWS != nil` etc.) have been moved to `pkg/apis/kubermatic/v1/helper` to make them easier to re-use without depending on the entire cloud provider package.

The resources<->defaulting loop was caused by ownership questions: Who owns the constant `DefaultNodePortRange`? Previously all `Default*` constants lived in `pkg/defaulting`, but this won't work when `pkg/resources` also needs them.

This got solved by moving constants that are needed for the `resources` package into the `resources` package. The `defaulting` package uses the `resources` package from now on, the other direction doesn't happen anymore. As much as possible, `Default*` constants should live in the `defaulting` package, but when needed, can be moved to `resources`.

Besides shuffling code around, this PR does not change any behaviour.

/kind cleanup

```release-note
NONE
```


pkg/defaulting/seed.go
```diff
@@ -0,0 +1,111 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
ahmedwaleedmalik:
```suggestion
Copyright 2022 The Kubermatic Kubernetes Platform contributors.
```

ahmedwaleedmalik (commented):
Just one nit otherwise great work 👍

ahmedwaleedmalik (approved):
/approve

PR #10889, opened 2022-08-30
Add report types to metering schedule configuration

### LINKED ISSUE #10815: Extend metering configuration API to support different report types
**User story**
As an API user, I would like to be able to define what kind of reports should be generated by my schedule configuration. At the moment specifying report type through API is not possible and by default schedule configuration creates both types of reports. For now available report types are `cluster`, `namespace`. It should be possible to define both of them if needed.

**Acceptance criteria**
- getting/creating/editing schedule configuration should allow for changing report type

pkprzekwas:
As seed CRD has been extended with additional report types we need to introduce this change to the API.

Fixes #10815 

**What type of PR is this?**
/kind feature

```release-note
Add `types` parameter, defining report types, to the `/api/v1/admin/metering/configurations/reports` POST and PUT.
```

**Documentation**:
```documentation
https://github.com/kubermatic/docs/blob/master/content/kubermatic/master/tutorials-howtos/metering/_index.en.md
```

### DIFF
--- a/pkg/ee/metering/report_config_handler.go
+++ b/pkg/ee/metering/report_config_handler.go
@@ -36,13 +36,17 @@ import (
 
 	apiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
+	"k8c.io/kubermatic/v2/pkg/handler/v1/common"
 	"k8c.io/kubermatic/v2/pkg/provider"
 	utilerrors "k8c.io/kubermatic/v2/pkg/util/errors"
 	"k8c.io/kubermatic/v2/pkg/validation"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	k8svalidation "k8s.io/apimachinery/pkg/util/validation"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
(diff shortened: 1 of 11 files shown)


pkg/ee/metering/report_config_handler.go
```diff
@@ -92,6 +97,18 @@ func (m createReportConfigurationReq) Validate() error {
 		}
 	}
 
+	if m.Body.Types != nil {
+		if len(*m.Body.Types) == 0 {
+			return utilerrors.NewBadRequest("at least one report type is required")
```
lsviben, line 102: should we leave some default type? Like the report type that was before set before we added these options?

pkprzekwas: Default types are applied when no `types` are provided (`m.Body.Types == nil`) in the incoming request. In case someone implicitly set `types: []`, then we want to throw an error. Or you mean sth else?

lsviben: I meant if we default for nil, why not default for empty list as well?

lsviben: But I dont have strong  feelinga about that, just wondering

pkprzekwas: I would prefer to leave an error when someone explicitly tries to set na empty list `[]`. Otherwise someone may think that no reports will be generated but instead we will generate both cluster and namespace reports.

PR #10891, opened 2022-08-30
Split semver package

### LINKED ISSUE #10877: Use semVer compatible data type for AppDefs
This ticket was mainly created, because we could not solve this issue properly before the release. It was more complicated than anticipated because our internal semver package's `DeepCopy` func cuts off the "v" character and lengthens strings like "v1" into "1.0.0". Unfortunately it is not as simple as just changing the DeepCopy func, because in other parts (e.g. [AdmissionPlugins](https://github.com/kubermatic/kubermatic/pull/10850#issuecomment-1228250685)) we make use of that cutting and lengthening and depend on it. 
For a more detailed discussion see [10850](https://github.com/kubermatic/kubermatic/pull/10850)

Therefore this ticket exists to find a way to use a semVer based type under the hood that preseves the original name.

**Implementation hints**

* There was already some work done in [10850](https://github.com/kubermatic/kubermatic/pull/10850) on this topic, which might be useful for future implementations
* Back from the previous state, I think there is no other way than to create another semVer type in the package, which preserves the original string name (but of course that is up for debate when implementing this ticket)

SimonTheLeg:

Creates a v1 and v2 version of the semver package. v2 differs from v1 in the following manner: Running DeepCopy(), String() will preserve the original manner the version was provided in as opposed to normalizing the string. Instead a Normalize() func is introduced. This behaviour is required for Applications where we interact with third party components (e.g. git, helm), which would break with the normalization. Since we use the normalization in some parts, the v1 version of the package was introduced and will be used in these places.

Applications will be the first component to make use of the v2 package. In order to keep this PR as small as possible, the changes to Apps will be rolled out in a separate commit. 

**What type of PR is this?**

/kind cleanup

```release-note
none
```

**Documentation**:

```documentation
none
```

### DIFF
--- a/cmd/conformance-tester/pkg/scenarios/base.go
+++ b/cmd/conformance-tester/pkg/scenarios/base.go
@@ -30,7 +30,7 @@ import (
 	"k8c.io/kubermatic/v2/cmd/conformance-tester/pkg/types"
 	apiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
 	kubermaticv1 "k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1"
-	"k8c.io/kubermatic/v2/pkg/semver"
+	k8csemverv1 "k8c.io/kubermatic/v2/pkg/semver/v1"
 	apimodels "k8c.io/kubermatic/v2/pkg/test/e2e/utils/apiclient/models"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -45,7 +45,7 @@ type Scenario interface {
 	OperatingSystem() providerconfig.OperatingSystem
 	ContainerRuntime() string
 	DualstackEnabled() bool
-	Version() semver.Semver
+	Version() k8csemverv1.Semver
 	Datacenter() *kubermaticv1.Datacenter
 	Name() string
 	Log(log *zap.SugaredLogger) *zap.SugaredLogger
@@ -68,7 +68,7 @@ type Scenario interface {
 type baseScenario struct {
 	cloudProvider    providerconfig.CloudProvider
 	operatingSystem  providerconfig.OperatingSystem
-	version          semver.Semver
+	version          k8csemverv1.Semver
 	containerRuntime string
 	dualstackEnabled bool
 	datacenter       *kubermaticv1.Datacenter
@@ -82,7 +82,7 @@ func (s *baseScenario) OperatingSystem() providerconfig.OperatingSystem {
 	return s.operatingSystem
 }
 
-func (s *baseScenario) Version() semver.Semver {
+func (s *baseScenario) Version() k8csemverv1.Semver {
 	return s.version
 }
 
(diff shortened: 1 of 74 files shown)


SimonTheLeg:
so now the rebase of 54 files should be done :D

xrstf:
Sorry to say this, but I really don't like us having 2 separate semver packages in our own codebase, plus using Masterminds/semver as a third semver package. 

> Since we use the normalization in some parts, the v1 version of the package was introduced and will be used in these places.

Maybe it makes more sense to explicitly apply the normalization in the places that currently already used the old semver package. It's not great, but IMHO much better than having 2 semver packages side by side in KKP.

embik:
Stupid question perhaps, but why do we need a semver type in the API here? It seems we don't really want it to be semver due to our external sources implementing semver in very different ways and without normalisation. Would it not be enough to keep this as a string and - when that semver information is really needed - parse it occasionally?

SimonTheLeg:
@embik The original idea was to keep them semVer to allow things like ordering them and to have direct protection against obvious non-semver strings (we were thinking of things like "master" or "dev" here).

> It seems we don't really want it to be semver due to our external sources implementing semver in somewhat different ways and without normalisation.

I think this is the semVer dilemma: One one hand the spec [explicitly forbids strings like v1.2.3](https://semver.org/spec/v2.0.0.html#is-v123-a-semantic-version), on the other the majority of implementations accept or use it. tbh I have no idea how exactly this came to be, but I think working around that fact is our best option.

xrstf:
So, totally out of left field, when updating our controller-runtime version i found that we now pull in `github.com/blang/semver/v4`:

```
$ go mod why -m github.com/blang/semver/v4
# github.com/blang/semver/v4
k8c.io/kubermatic/v2/cmd/seed-controller-manager
github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/defaulting
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/metrics
k8s.io/component-base/metrics
github.com/blang/semver/v4
```

Which means that KKP currently already has 3 semver implementations (ours, Mastermind and blang).

blang has a [ParseTolerant](https://pkg.go.dev/github.com/blang/semver/v4?utm_source=godoc#ParseTolerant) function, which does some normalization.

Maybe we should aim towards using a single implementation (blang?). Its normalization sounds similar to what this PR's new semver code would do.

moadqassem:
@xrstf Do you know which package/dependency is using blang? sorry I am too lazy to look for it, plus you probably already know 😂

xrstf:
Yes, I literally put it in my comment:

```
$ go mod why -m github.com/blang/semver/v4
# github.com/blang/semver/v4
k8c.io/kubermatic/v2/cmd/seed-controller-manager
github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/defaulting
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel
k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/metrics
k8s.io/component-base/metrics
github.com/blang/semver/v4
```

moadqassem:
> Yes, I literally put it in my comment ^^:
> 
> ```
> $ go mod why -m github.com/blang/semver/v4
> # github.com/blang/semver/v4
> k8c.io/kubermatic/v2/cmd/seed-controller-manager
> github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1
> k8s.io/apiextensions-apiserver/pkg/apiserver/schema/defaulting
> k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel
> k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/metrics
> k8s.io/component-base/metrics
> github.com/blang/semver/v4
> ```

Sorry, I overlooked it because was kinda focused more on the discussion when I asked the question 😉 Anyway, it seems that `github.com/blang/semver/v4` is a good option already, so why not just use it everywhere :)

SimonTheLeg:
with the move making this PR obsolete and a stop-gap solution in place (https://github.com/kubermatic/kubermatic/pull/11359), I am going to close this PR.

PR #10894, opened 2022-08-30
validate length of EAR encryption key

### LINKED ISSUE #10829: EAR enabled with invalid encryption key gets stuck in Pending state

PratikDeoghare:
 Validates length of the encryption key used for EAR. Currently the length is not validated.

Fixes https://github.com/kubermatic/kubermatic/issues/10829

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -185,11 +185,6 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		log.Fatalw("Unable to create the seed getter", zap.Error(err))
 	}
 
-	seedClientGetter, err := seedClientGetterFactory(rootCtx, mgr.GetClient())
-	if err != nil {
-		log.Fatalw("Unable to create the seed client getter", zap.Error(err))
-	}
-
 	var configGetter provider.KubermaticConfigurationGetter
 	if options.kubermaticConfiguration != nil {
 		configGetter, err = kubernetesprovider.StaticKubermaticConfigurationGetterFactory(options.kubermaticConfiguration)
@@ -232,7 +227,6 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		mgr:                  mgr,
 		clientProvider:       clientProvider,
 		seedGetter:           seedGetter,
-		seedClientGetter:     seedClientGetter,
 		configGetter:         configGetter,
 		dockerPullConfigJSON: dockerPullConfigJSON,
 		log:                  log,
--- a/cmd/seed-controller-manager/main.go
+++ b/cmd/seed-controller-manager/main.go
@@ -185,6 +185,11 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		log.Fatalw("Unable to create the seed getter", zap.Error(err))
 	}
 
+	seedClientGetter, err := seedClientGetterFactory(rootCtx, mgr.GetClient())
+	if err != nil {
+		log.Fatalw("Unable to create the seed client getter", zap.Error(err))
+	}
+
 	var configGetter provider.KubermaticConfigurationGetter
 	if options.kubermaticConfiguration != nil {
 		configGetter, err = kubernetesprovider.StaticKubermaticConfigurationGetterFactory(options.kubermaticConfiguration)
@@ -227,6 +232,7 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		mgr:                  mgr,
 		clientProvider:       clientProvider,
 		seedGetter:           seedGetter,
+		seedClientGetter:     seedClientGetter,
 		configGetter:         configGetter,
 		dockerPullConfigJSON: dockerPullConfigJSON,
 		log:                  log,
--- a/cmd/seed-controller-manager/wrappers_ce.go
+++ b/cmd/seed-controller-manager/wrappers_ce.go
@@ -24,8 +24,6 @@ import (
 
 	"k8c.io/kubermatic/v2/pkg/provider"
 	"k8c.io/kubermatic/v2/pkg/provider/kubernetes"
-	kubernetesprovider "k8c.io/kubermatic/v2/pkg/provider/kubernetes"
-
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -35,14 +33,6 @@ func addFlags(fs *flag.FlagSet) {
 
 func seedGetterFactory(ctx context.Context, client ctrlruntimeclient.Reader, options controllerRunOptions) (provider.SeedGetter, error) {
 	return kubernetes.SeedGetterFactory(ctx, client, options.seedName, options.namespace)
-}
-
-func seedClientGetterFactory(ctx context.Context, client ctrlruntimeclient.Reader) (provider.SeedClientGetter, error) {
-	seedKubeconfigGetter, err := kubernetesprovider.SeedKubeconfigGetterFactory(ctx, client)
-	if err != nil {
-		return nil, err
-	}
-	return kubernetesprovider.SeedClientGetterFactory(seedKubeconfigGetter), nil
 }
 
 func setupControllers(_ *controllerContext) error {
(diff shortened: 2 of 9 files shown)


pkg/controller/seed-controller-manager/encryption-at-rest-controller/controller.go
```diff
@@ -269,6 +335,38 @@ func (r *Reconciler) reconcile(ctx context.Context, log *zap.SugaredLogger, clus
 	}
 }
 
+func (r *Reconciler) validateSecretRef(cluster *kubermaticv1.Cluster, log *zap.SugaredLogger) (*reconcile.Result, error) {
+	if hasSecretKeyRef(cluster) {
```
PratikDeoghare: change to `!hasKeyRef { return } `

pkg/validation/cluster_test.go
```diff
@@ -886,3 +886,63 @@ func TestValidateGCPCloudSpec(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Log(test.name)
+			err := validateEncryptionConfiguration(test.clusterSpec, field.NewPath("spec", "encryptionConfiguration"))
+			t.Logf("%+v", err)
+			if (err != nil) != test.expectErr {
+				t.Logf("expected error: %t, got: %+v", test.expectErr, err)
```
embik: The whole logging here is probably for debugging the test, but it should probably be removed and as far as I can see, this test can currently simply not fail.

embik (changes requested):

pkg/validation/cluster_test.go
```diff
@@ -886,3 +886,61 @@ func TestValidateGCPCloudSpec(t *testing.T) {
...
+			expectErr: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			err := validateEncryptionConfiguration(test.clusterSpec, field.NewPath("spec", "encryptionConfiguration"))
+			if (err != nil) != test.expectErr {
+				t.Logf("expected error: %t, got: %+v", test.expectErr, err)
+			}
```
embik: This still doesnt fail the test if the error is not the expected error, no? I think we should be doing something like:

```go
assert.Equal(t, test.expectedError, err)
```

As the other test cases in this file do.

embik (changes requested):

pkg/validation/cluster_test.go
PratikDeoghare: Yes. Can make it happen.

PratikDeoghare: How about that?

cmd/seed-controller-manager/main.go
```diff
@@ -185,6 +185,11 @@ Please install the VerticalPodAutoscaler according to the documentation: https:/
 		log.Fatalw("Unable to create the seed getter", zap.Error(err))
 	}
 
+	seedClientGetter, err := seedClientGetterFactory(rootCtx, mgr.GetClient())
+	if err != nil {
+		log.Fatalw("Unable to create the seed client getter", zap.Error(err))
+	}
+
```
embik: I don't think this (and all the other stuff to get this into the EAR controller) is necessary; The seed-controller-manager mainly talks to its assigned Seed, so it only needs one Seed client. You can see other controllers using `mgr.GetClient()` to get a seed client here:

https://github.com/kubermatic/kubermatic/blob/b4a7df6129ef6a978a3f57b926ff24f987f8a429/pkg/controller/seed-controller-manager/preset-controller/controller.go#L69

https://github.com/kubermatic/kubermatic/blob/b4a7df6129ef6a978a3f57b926ff24f987f8a429/pkg/controller/seed-controller-manager/constraint-template-controller/controller.go#L85

https://github.com/kubermatic/kubermatic/blob/b4a7df6129ef6a978a3f57b926ff24f987f8a429/pkg/controller/seed-controller-manager/operating-system-profile-synchronizer/controller.go#L84

embik (changes requested):

cmd/seed-controller-manager/main.go
embik: Using the client embedded into the `Reconciler` struct for the encryption-at-rest-controller should suffice for what the code does, no?

https://github.com/kubermatic/kubermatic/blob/b4a7df6129ef6a978a3f57b926ff24f987f8a429/pkg/controller/seed-controller-manager/encryption-at-rest-controller/controller.go#L90

PratikDeoghare: Yes. It should work. I thought seed-ctrl-mgr could run outside the seeds it is managing and then it would be necessary to get the right client. 

But I see no other ctrl hanging on the seed-ctrl-mgr doing this. 

Will cleanup. :+1:

PratikDeoghare: Why is there `seedGetter` tho since you already know the seed?

embik: I assume because you might want to read information from the `Seed` object (e.g. specific configuration to change controller behaviour).

pkg/controller/seed-controller-manager/encryption-at-rest-controller/controller.go
```diff
@@ -177,9 +179,62 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	return *result, err
 }
 
+func getSecretKeyValue(client ctrlruntimeclient.Client, ref *corev1.SecretKeySelector, namespace string) ([]byte, error) {
+	secret := corev1.Secret{}
+	if err := client.Get(context.Background(), ctrlruntimeclient.ObjectKey{
```
embik: This should take a `ctx context.Context` argument and pass that to the client calls.

pkg/provider/kubernetes/seed.go
```diff
@@ -144,7 +144,7 @@ func SeedsGetterFactory(ctx context.Context, client ctrlruntimeclient.Client, na
 	}, nil
 }
 
-func SeedKubeconfigGetterFactory(ctx context.Context, client ctrlruntimeclient.Client) (provider.SeedKubeconfigGetter, error) {
+func SeedKubeconfigGetterFactory(ctx context.Context, client ctrlruntimeclient.Reader) (provider.SeedKubeconfigGetter, error) {
```
embik: Don't change this :D

pkg/validation/cluster_test.go
```diff
@@ -886,3 +886,66 @@ func TestValidateGCPCloudSpec(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateEncryptionConfiguration(t *testing.T) {
```
embik, line 890: Can we also add a test with invalid base64?

cmd/seed-controller-manager/wrappers_ce.go
```diff
@@ -24,7 +24,6 @@ import (
 
 	"k8c.io/kubermatic/v2/pkg/provider"
 	"k8c.io/kubermatic/v2/pkg/provider/kubernetes"
-
```
embik: don't remove this line.

cmd/seed-controller-manager/wrappers_ee.go
```diff
@@ -27,7 +27,6 @@ import (
 	groupprojectbindingcontroller "k8c.io/kubermatic/v2/pkg/ee/group-project-binding/controller"
 	resourcequotaseedcontroller "k8c.io/kubermatic/v2/pkg/ee/resource-quota/seed-controller"
 	"k8c.io/kubermatic/v2/pkg/provider"
-
```
embik: same here.

embik (changes requested):
Sorry that I keep throwing curveballs at you 😅

embik (approved):
/lgtm

PR #10895, opened 2022-08-30
Fix flakiness documentation  generation

vgramer:
the docs generation is now flaky. if you run this script in a loop on your machine, it will randomly spit out a diff:

screenshot-2022-08-30-140645

To generate the YAML CRS example, we use `genyaml`.
> `genyaml` is a simple documentation tool used to marshal YAML from Golang structs. It extracts *doc comments* from `.go` sources files and adds them as *comment nodes* in the YAML output.

`genyaml` parse of packages and  stores comments as a `map[typeName]map[tagName][Comment]`.  c.f.  [genyaml.go](https://github.com/kubernetes/test-infra/blob/6900af122506ff5431f61ecaa8f01674dd37ac99/pkg/genyaml/genyaml.go#L317-L382)

example: 
```go
comments = map[string]map[string][Comment]{
    "KubermaticUIConfiguration": map[string][Comment]{
        "dockerRepository": Comment{"string", false, "DockerRepository is the repository containing the Kubermatic dashboard image."},
        "dockerTag": Comment{"string", false, "DockerTag is used to overwrite the dashboard Docker image tag and is only for development\npurposes. "},
        "config": Comment{ "string", false, "Config sets flags for various dashboard features."},
        "resources" : Comment{ "ResourceRequirements", true, "Resources describes the requested and maximum allowed CPU/memory usage."},
        "replicas" : Comment{"int32", false, "Replicas sets the number of pod replicas for the UI deployment."},
    },
    "KubermaticProjectsMigratorConfiguration": map[string][Comment]{
        "dryRun":  Comment{"bool", false, "DryRun makes the migrator only log the actions it would take."},
    }
}
```

We have defined `ApplicationInstallationSpec` like this:
```go
type ApplicationInstallationSpec struct {
	// Namespace describe the desired state of the namespace where application will be created.
	Namespace NamespaceSpec `json:"namespace"`

	// ApplicationRef is a reference to identify which Application should be deployed
	ApplicationRef ApplicationRef `json:"applicationRef"`

	// Values describe overrides for manifest-rendering. It's a free yaml field.
	// +kubebuilder:pruning:PreserveUnknownFields
	Values runtime.RawExtension `json:"values,omitempty"`
	// As kubebuilder does not support interface{} as a type, deferring json decoding, seems to be our best option (see https://github.com/kubernetes-sigs/controller-tools/issues/294#issuecomment-518379253)
}

// AppNamespaceSpec describe the desired state of the namespace where application will be created.
type NamespaceSpec struct {
	// Name is the namespace to deploy the Application into.
	// Should be a valid lowercase RFC1123 domain name
	// +kubebuilder:validation:Pattern:=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
	// +kubebuilder:validation:MaxLength:=63
	// +kubebuilder:validation:Type=string
	Name string `json:"name"`

	// +kubebuilder:default:=true

	// Create defines whether the namespace should be created if it does not exist. Defaults to true
	Create bool `json:"create"`

	// Labels of the namespace
	// More info: http://kubernetes.io/docs/user-guide/labels
	// +optional
	Labels map[string]string `json:"labels,omitempty"`

	// Annotations of the namespace
	// More info: http://kubernetes.io/docs/user-guide/annotations
	// +optional
	Annotations map[string]string `json:"annotations,omitempty"`
}
```

The problem is the type `NamespaceSpec` already exists in `corev1`. So when `genyaml` parse the packages, it overrides `comments["NamespaceSpec"]` with one of the types (`appskubermaticv1` or `corev1`) leading to different results over different runs. 

So to fix the problem, we rename the type `appskubermaticv1.NamespaceSpec` to `appskubermaticv1.AppNamespace`

**What type of PR is this?**
/kind bug

To avoid changes in the swagger, In `pkg/api/v2/types.go`, I introduce type  `apiv2.ApplicationInstallationSpec`and use it instead of `appkubermaticv1 .ApplicationInstallationSpec`

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

xrstf:
Hetzner Helsinki is full.

/override pre-kubermatic-mla-e2e
/override pre-kubermatic-opa-e2e

PR #10903, opened 2022-08-31
Do not install all CRDs into all cluster kinds

xrstf:
Since KKP 2.21 the ApplicationInstallation CRD belongs to the *usercluster* (not the namespace on the seed, but the usercluster itself). This means that KKP will watch the CRD and ensure it exists. If changes to it are made, they are reverted (just like every other object we watch and reconcile).

However this is a problem when you create a usercluster and then try to install KKP into it. Right now, with KKP 2.21 providing the usercluster, one cannot install KKP 2.20 (for testing, for example):

```
INFO[14:36:45] Initializing installer…                     edition="Enterprise Edition" version=v2.20.6
INFO[14:36:45] Validating the provided configuration…
WARN[14:36:45]    kubermaticOperator.imagePullSecret is empty, setting to spec.imagePullSecret from KubermaticConfiguration
INFO[14:36:45] Provided configuration is valid.
INFO[14:36:46] Validating existing installation…
INFO[14:36:46] Existing installation is valid.
INFO[14:36:46] Deploying KKP master stack…
INFO[14:36:46]    Deploying kubermatic-fast StorageClass…
INFO[14:36:47]    StorageClass exists, nothing to do.
INFO[14:36:47]    Deploying nginx-ingress-controller…
INFO[14:36:47]       Deploying Helm chart…
INFO[14:36:47]       Re-installing because --force is set…
INFO[14:37:10]    Success.
INFO[14:37:10]    Deploying cert-manager…
INFO[14:37:11]       Deploying Custom Resource Definitions…
INFO[14:37:14]       Deploying Helm chart…
INFO[14:37:14]       Re-installing because --force is set…
INFO[14:37:24]    Success.
INFO[14:37:24]    Deploying Dex…
INFO[14:37:25]       Re-installing because --force is set…
INFO[14:37:28]    Success.
INFO[14:37:28]    Deploying Kubermatic Operator…
INFO[14:37:28]       Deploying Custom Resource Definitions…
ERRO[14:37:28] Operation failed: failed to deploy Kubermatic Operator: failed to deploy CRDs: failed to deploy CRD applicationinstallations.apps.kubermatic.k8c.io: CustomResourceDefinition.apiextensions.k8s.io "applicationinstallations.apps.kubermatic.k8c.io" is invalid: spec.scope: Invalid value: "Cluster": field is immutable.
```

In this scenario the installer is fighting with the usercluster-ctrl-mgr, and it's losing.

To prevent (rare) conflicts like this from happening, this PR adds a new annotation onto all our CRDs. This new `kubermatic.k8c.io/location` annotation will tell the installer/operator where each CRD is supposed to live. This way they can easily skip the ApplicationInstallation CRD.

This issue would otherwise re-surface everytime we introduce a new CRD in KKP that is managed inside userclusters.

CRDs without this new annotation will be treated like before (i.e. always installed).

/kind bug

```release-note
Fix rare CRD conflict when installing old KKP versions into userclusters created by a different KKP version.
```

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -35,7 +35,7 @@ go run sigs.k8s.io/controller-tools/cmd/controller-gen \
 
 annotation="kubermatic.k8c.io/location"
 locationMap='{
-  "applicationdefinitions.apps.kubermatic.k8c.io": "master",
+  "applicationdefinitions.apps.kubermatic.k8c.io": "master,seed",
   "applicationinstallations.apps.kubermatic.k8c.io": "usercluster",
   "addonconfigs.kubermatic.k8c.io": "master",
   "addons.kubermatic.k8c.io": "seed",
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -33,7 +33,7 @@ go run sigs.k8s.io/controller-tools/cmd/controller-gen \
   paths=./pkg/apis/... \
   output:crd:dir=./$CRD_DIR
 
-labelName="kubermatic.k8c.io/location"
+annotation="kubermatic.k8c.io/location"
 locationMap='{
   "applicationdefinitions.apps.kubermatic.k8c.io": "master",
   "applicationinstallations.apps.kubermatic.k8c.io": "usercluster",
@@ -67,7 +67,7 @@ locationMap='{
 }'
 
 failure=false
-echodate "Labeling CRDs"
+echodate "Annotating CRDs"
 
 for filename in $CRD_DIR/*.yaml; do
   crdName="$(yq4 '.metadata.name' "$filename")"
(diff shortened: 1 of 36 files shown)


hack/update-codegen.sh
```diff
@@ -24,9 +24,64 @@ CONTAINERIZE_IMAGE=golang:1.18.4 containerize ./hack/update-codegen.sh
...
 go run sigs.k8s.io/controller-tools/cmd/controller-gen \
   crd \
   object:headerFile=./hack/boilerplate/ce/boilerplate.go.txt \
   paths=./pkg/apis/... \
-  output:crd:dir=./pkg/crd/k8c.io
+  output:crd:dir=./$CRD_DIR
+
+annotation="kubermatic.k8c.io/location"
+locationMap='{
+  "applicationdefinitions.apps.kubermatic.k8c.io": "master",
```
vgramer:
```suggestion
  "applicationdefinitions.apps.kubermatic.k8c.io": "master,seed",
```

applicationDefinition CRs are replicated from master to seed so we must install applicationDefinition CRD into seeds.

vgramer (approved):
/lgtm

PR #10906, opened 2022-08-31
remove deprecated Cluster.Address field (replaced by Cluster.Status.address)

xrstf:
KKP 2.21 moved and migrated the address, it's now possible to get rid of the deprecated field alltogether.

/kind cleanup

```release-note
BREAKING CHANGE: The `address` field in the Cluster CRD was deprecated in KKP 2.21 and removed in this release. Use `status.address` instead. Existing clusters were migrated automatically by the seed-controller-manager in release 2.21.
```


pkg/resources/prometheus/configmap.go
```diff
@@ -392,7 +392,7 @@ scrape_configs:
   scheme: https
   tls_config:
 {{ .ApiserverTLSConfig | indent 4 }}
-  
+
```
xrstf, line 395: FYI, this tiny little change where I removed the trailing newlines caused the major formatting changes in the YAML fixtures. :grin:

ahmedwaleedmalik (approved):
/approve

PR #10907, opened 2022-08-31
Check KKP and seed versions in installer to prevent minor version skips

xrstf:
To ensure a safer upgrade experience, this PR adds additional validations to the installer. They are based on the work in 2.21, thanks to which we now have version information available for both KKP and each Seed :-)

When installing, the installer checks

1. That the KKP version (`KubermaticConfiguration.Status.KubermaticVersion`) is exactly the previous minor version (judging from the installer's own compiled-in version).
2. Each seed cluster satisfies the same condition.
3. Each seed, when a valid kubeconfig is available, must be properly reconciled. This check does not rely on the `status.phase` as that field is explicitly for informational purpose only.

When no version is available, no check happens. This is to allow for someone to create CRDs manually (for whatever reason) and preload the KubermaticConfiguration/Seed objects before running the installer.

The version check will fail if we ever release KKP 3.x, but there is little I can do about that now: when 3.0 hits, we must know the latest 2.x minor and right now, we don't.

/kind feature

```release-note
The KKP installer will ensure that neither the master nor any seed violate the KKP version skew policy, i.e. skipping a minor release during an upgrade is not permitted. Additionally, all seeds must be healthy for an upgrade to be possible. These changes are to ensure that smaller issues now do not lead to bigger problems during upgrades and migrations.
```

### DIFF
--- a/pkg/install/stack/kubermatic-master/validation.go
+++ b/pkg/install/stack/kubermatic-master/validation.go
@@ -32,6 +32,8 @@ import (
 	"k8c.io/kubermatic/v2/pkg/features"
 	"k8c.io/kubermatic/v2/pkg/install/stack"
 	"k8c.io/kubermatic/v2/pkg/install/util"
+	"k8c.io/kubermatic/v2/pkg/provider"
+	"k8c.io/kubermatic/v2/pkg/provider/kubernetes"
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

PR #10922, opened 2022-09-06
Migrate AWS and EKS providers to AWS SDK v2 for Go

### LINKED ISSUE #10841: Migrate to AWS SDK for Go v2
We're currently using the AWS SDK for Go in v1. The [v2 SDK](https://aws.github.io/aws-sdk-go-v2/docs/) has some neat stuff like being able to [re-use our contexts](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/#timeouts).

We should migrate to v2 to make sure we're staying up to date with AWS APIs and pull in potential stability improvements in communicating with the AWS APIs.

Corresponding issue for machine-controller: https://github.com/kubermatic/machine-controller/issues/1417

embik:

This PR migrates both the AWS and EKS code to the v2 of AWS SDK for Go. Some light API changes that seemed to rely on the internal details of the AWS SDK are adjusted. There's not really a reason to have a string pointer when you can have just a string.

The `Verifier` part of the `aws-iam-authenticator` code has been removed because it relied on partition information no longer exposed by the SDK. In addition, this code was unused in KKP, so removing it seemed the cleanest option.

Fixes #10841

**What type of PR is this?**

```release-note
The AWS and EKS cloud provider implementations now use the AWS Go SDK v2 to communicate with AWS
```

**Documentation**:

```documentation
NONE
```


xrstf:
@imharshita PTAL

ahmedwaleedmalik (approved):
Great job, especially with the context propagation.

/approve
/hold if @harshita-sharma012 wants to take a look as well

imharshita:
> Great job, especially with the context propagation.
> 
> /approve /hold if @harshita-sharma012 wants to take a look as well

yes please, i will finish my review by tomorrow, positively.

ahmedwaleedmalik:
> > Great job, especially with the context propagation.
> > /approve /hold if @harshita-sharma012 wants to take a look as well
> 
> yes please, i will finish my review by tomorrow, positively.

No worries at all, I was just bored so reviewed this :P

pkg/provider/cloud/eks/provider.go
```diff
@@ -432,66 +417,61 @@ func ListUpgrades(ctx context.Context,
...
-	var aerr awserr.Error
+	var aerr smithy.APIError
 	if errors.As(err, &aerr) {
-		// A service error occurred.
-		var reqErr awserr.RequestFailure
-		if errors.As(aerr, &reqErr) {
-			return utilerrors.New(reqErr.StatusCode(), reqErr.Message())
-		}
-		return errors.New(aerr.Message())
+		return utilerrors.New(500, fmt.Sprintf("%s: %s", aerr.ErrorCode(), aerr.ErrorMessage()))
```
imharshita: why hardcoded status code to 500?

imharshita (changes requested):

pkg/provider/cloud/eks/provider.go
xrstf: Good catch! I think we should get rid of this error handling code alltogether, to quote my commit:

> smithy.APIError implementations all seem to be doing the same, if not better string marshalling that we would do.
>
> Also using HTTP 500 (internal server error) for every possible fault is not ideal (ValidateCredentials() should not error out with an internal server error on bad credentials), and in general the code in the provider package should not return HTTP-related things, as controllers also use it.
>
> I removed the DecodeError() entirely, as it seems to serve no purpose anymore with SDK v2. :)

imharshita: but, this part of code from external cluster controller expects error to be an `utilerrors.HTTPError` :
https://github.com/kubermatic/kubermatic/blob/master/pkg/controller/master-controller-manager/external-cluster/cluster_controller.go#L106

either we can modify controller here, or I can create a separate pr as this will not immediately affect the controller unless `notFound` etc. clusters

embik: @imharshita can you please take another look, Christoph fixed this in https://github.com/kubermatic/kubermatic/pull/10922/commits/628c8727b937b7eb338b1eb009283f3ca61d56da I think.

embik: Sorry, seems I misunderstood this conversation. Looking.

embik: @imharshita I reworked the DecodeError function, it should now return what is expected by the controller code, if that information was included in the returned error. Please take another look.

imharshita: @embik  regarding hardcoding 500 code in DecodeError , I will fix that in a separate pr.

embik: the hardcoded 500 error is already resolved. We only fall back to 500 if no HTTP information was included in the error.

PR #10938, opened 2022-09-08
Enhance treatment of missing credential privilege for AKS external cluster

### LINKED ISSUE #10854: AKS external cluster: expose errors when having issues when creating/editing the cluster
For instance, when credentials have access to create a cluster but not to access the Kubeconfig, cluster is created, but it doesn't become stable in the UI:

In that case, it's not possible to identify what is the cause of the issue in the UI.

The same happens when trying to upgrade the controle plane version, for example, and Azure keeps the cluster in a "Failed" state because of subscription number of requests limitation.
image
Error log found in Azure console:
```
"{\"status\":\"Failed\",\"error\":{\"code\":\"ResourceOperationFailure\",\"message\":\"The resource operation completed with terminal provisioning state 'Failed'.\",\"details\":[{\"code\":\"GetClusterStateError\",\"message\":\"Client Error: Category: ClientError; Code: Unspecified; SubCode: TooManyRequestsHighCostGetVMScaleSet; Message: ; InnerMessage: ; Dependency: Microsoft.Compute/VirtualMachineScaleSet; OrginalError: autorest/azure: Service returned an error. Status=429 Code=\\\"OperationNotAllowed\\\" Message=\\\"The server rejected the request because too many requests have been received for this subscription.\\\" Details=[{\\\"code\\\":\\\"TooManyRequests\\\",\\\"message\\\":\\\"{\\\\\\\"operationGroup\\\\\\\":\\\\\\\"HighCostGetVMScaleSet30Min\\\\\\\",\\\\\\\"startTime\\\\\\\":\\\\\\\"2022-08-25T20:11:22.7378669+00:00\\\\\\\",\\\\\\\"endTime\\\\\\\":\\\\\\\"2022-08-25T20:15:00+00:00\\\\\\\",\\\\\\\"allowedRequestCount\\\\\\\":675,\\\\\\\"measuredRequestCount\\\\\\\":1807}\\\",\\\"target\\\":\\\"HighCostGetVMScaleSet30Min\\\"}] InnerError={\\\"internalErrorCode\\\":\\\"TooManyRequestsReceived\\\"}; AKSTeam: RP, Retriable: false, InternalRetryAfterTimer: 2022-08-25 20:15:00.748304999 +0000 UTC m=+9767.081282987\"}]}}"```

### PR DESCRIPTION
**Pull request by hbernardo**
Users get confused when their Azure credentials don't have all necessary permissions (e.g. cannot get the kubeconfig). In that case, they cannot see what the issue is. So this PR adds permissions validation before the creation of the AKS external cluster to prevent credentials issues after the creation of it.

Fixes #10854

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/provider/cloud/aks/provider.go
+++ b/pkg/provider/cloud/aks/provider.go
@@ -21,10 +21,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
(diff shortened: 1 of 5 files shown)


pkg/provider/cloud/aks/provider.go
```diff
@@ -341,6 +343,73 @@ func ValidateCredentials(ctx context.Context, cred resources.AKSCredentials) err
 	return DecodeError(err)
 }
 
+func ValidateCredentialsForResourceGroup(ctx context.Context, cred resources.AKSCredentials, resourceGroup string) error {
```
imharshita:
```suggestion
func ValidateCredentialsForResourceGroup(ctx context.Context, cred resources.AKSCredentials, resourceGroup string) bool {
```

pkg/provider/cloud/aks/provider.go
```diff
@@ -341,6 +343,73 @@ func ValidateCredentials(ctx context.Context, cred resources.AKSCredentials) err
...
+	}
+
+	aksClient, err := armcontainerservice.NewManagedClustersClient(cred.SubscriptionID, azcred, nil)
+	if err != nil {
+		return DecodeError(err)
+	}
+
+	_, err = aksClient.NewListPager(nil).NextPage(ctx)
+
+	return DecodeError(err)
```
imharshita:
```suggestion
roleclient, err := armauthorization.NewRoleDefinitionsClient(azcred, nil)
	if err != nil {
		return nil, DecodeError(err)
	}
	scope := "/subscriptions/" + cred.SubscriptionID + "/resourceGroups/" + resourceGroup
	var (
		hasSpecificRoles = map[string]bool{
			"Azure Kubernetes Service Contributor Role":   false,
			"Azure Kubernetes Service Cluster Admin Role": false,
		}
	)
	pagerrole := roleclient.NewListPager(scope,
		&armauthorization.RoleDefinitionsClientListOptions{Filter: nil})
	for pagerrole.More() {
		nextResult, err := pagerrole.NextPage(ctx)
		if err != nil {
			return nil, DecodeError(err)
		}

		for _, v := range nextResult.Value {
			roleName := *v.Properties.RoleName
			if _, ok := hasSpecificRoles[roleName]; ok {
				hasSpecificRoles[roleName] = ok
			}
		}
	}
	for role, hasSpecificRole := range hasSpecificRoles {
		if !hasSpecificRole {
		        return false
			// return nil, utilerrors.New(http.StatusForbidden, fmt.Sprintf("Missing role: %s", role))
		}
		return true
	}
```
you can return the error or return the boolean

pkg/provider/cloud/aks/provider.go
```diff
@@ -21,10 +21,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization"
```
imharshita, line 29: "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization"

imharshita:
hey, so in ValidateCredentialsForResourceGroup method the approach is to list the list of permissions  and check if the list includes the permissions reqired.
I would propose the approach of listing roles and then check if the list has required roles

imharshita:
hey, so in ValidateCredentialsForResourceGroup method the approach is to list the list of permissions  and check if the list includes the permissions reqired.
I would propose the approach of listing roles and then check if the list has required roles

PR #10948, opened 2022-09-10
Fix OSM migration on external Seed clusters

### LINKED ISSUE #10940: Disabling OSM on existing clusters during 2.21 upgrade fails on separate seeds
### What happened?

When upgrading a Multi-Seed KKP environment, reconciling the seed can fail with error events like these on the `Seed`:

```
Warning  SeedReconcilingError  59m (x6 over 3h57m)     kkp-seed-operator  aws-seed: [Cluster.kubermatic.k8c.io "<Cluster ID>" is invalid: spec.pause: Required value, Cluster.kubermatic.k8c.io "<Cluster ID>" is invalid: spec.pause: Required value]
````

Upgrading the seed to 2.21.0 does not progress as expected.

The issue is the OSM migration step using a `client.Update()` call, as that can easily hit different CRD versions being validated in different ways. A `Patch()` call should simplify the payload sent to the seed cluster and not trigger this issue (at least that is my assumption from looking at this; verify this is indeed the solution). https://github.com/kubermatic/kubermatic/blob/4c83dabf406ccdca8c2055b734266d92d66f57d0/pkg/controller/operator/seed/reconciler.go#L733

### Expected behavior

Upgrading seeds with existing clusters that do not have OSM enabled/disabled should work.

Not fully clear, but likely have a seed that does not live on the master cluster and have existing clusters without OSM enabled. Upgrading the environment should cause this.

- KKP version: 2.21.0

<details>

```yaml
# paste manifest here
```

</details>

N/A

N/A

#### Workaround

It should be possible to resolve this issue by patching all affected clusters on the seed cluster (do not apply this to clusters that have OSM enabled already):

```
$ kubectl patch cluster <CLUSTER_ID> --type=merge -p '{"spec":{"enableOperatingSystemManager":false}}'
```

xrstf:
During the 2.21 upgrade, if a cluster on an external Seed has no OSM flag yet, the operator will get stuck. See the inline comment in this patch for more details.

Fixes #10940

/kind bug

```release-note
Fix upgrades for external seeds that have clusters with no `enableOperatingSystemManager` flag yet, resulting in the seed-operator not being able to fully upgrade the seed cluster to 2.21.
```

### DIFF
--- a/pkg/controller/operator/seed/reconciler.go
+++ b/pkg/controller/operator/seed/reconciler.go
@@ -728,9 +728,14 @@ func (r *Reconciler) disableOperatingSystemManager(ctx context.Context, client c
 	// to do since for all the new clusters this field will be explicitly set to `true` via the webhooks.
 	for _, cluster := range clusterList.Items {
 		if cluster.Spec.EnableOperatingSystemManager == nil {
+			oldCluster := cluster.DeepCopy()
 			cluster.Spec.EnableOperatingSystemManager = pointer.Bool(false)
 
-			if err := client.Update(ctx, &cluster); err != nil {
+			// This must use Patch() instead of Update(); during the 2.20->2.21 change the spec.pause field was made optional
+			// and so an Update() from the operator on the master cluster (with 2.21 CRDs) will fail on external seed clusters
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

PR #10951, opened 2022-09-12
Fix webhook on seed-only clusters

### LINKED ISSUE #10941: kubermatic-webhook attempts to read kubeconfig secret on separate seed cluster
### What happened?

KKP 2.21 deploys a new `kubermatic-webhook` component into both master and seed clusters. When running dedicated seed clusters, one can run into errors like this:

```
(combined from similar events): failed to ensure addons: failed to ensure Addon <Cluster ID>/aws-node-termination-handler: failed to create *v1.Addon '<Cluster ID>/aws-node-termination-handler': admission webhook "addons.kubermatic.io" denied the request: addon mutation request <Request ID> failed: failed to get Seed client: failed to get kubeconfig secret "kubermatic/<Seed Name>-kubeconfig": Secret "<Seed Name>-kubeconfig" not found	
```

Which prevents creation of (most likely) various resources on the seed level as the webhook installed on seeds rejects them.

This happens because the `kubermatic-webhook` is using a `SeedClientGetter` to communicate with the underlying seed, but it is configured this way:

https://github.com/kubermatic/kubermatic/blob/4c83dabf406ccdca8c2055b734266d92d66f57d0/cmd/kubermatic-webhook/main.go#L307-L312

As such, it will expect a kubeconfig secret in the `kubermatic` namespace, something that is not synced to the seed cluster. So this only works on shared master and seed environments.

### Expected behavior

The various webhooks on a seed cluster work as intended and are able to talk to the underlying cluster.

Create a dedicated seed, install with KKP 2.21.0 and attempt to create clusters (this hasn't been tested yet but the issue should be reproducible).

- KKP version: 2.21.0

<details>

```yaml
# paste manifest here
```

</details>

N/A

N/A

### Additional information

We can probably fix this by deploying `kubermatic-webhook` with a special flag (or re-use an existing flag to extend the change in behaviour) that provides a `SeedClientGetter` that just returns the underlying client.

#### Workaround

To work around this issue, sync the `<Seed Name>-kubeconfig` secret that exists on the master cluster into the matching seed cluster (into the `kubermatic` namespace). This should allow the existing webhook to talk to the underlying seed until this bug is fixed.

xrstf:
The webhook on external seed clusters broke because no kubeconfig was available, breaking the seedKubeconfigGetter/seedClientGetter.

This PR wanted to implement @embik's suggestion to use the local client when possible. However the local client, governed by the Kubernetes service account, does not have permissions to do all the things a webhook needs to do, as it was built around the idea that the webhook _always_ goes through the Seed kubeconfig and thereby gains admin permissions. Fine-grained RBAC was not yet planned for the webhook.

This situation lead me to re-consider (see the commit log of this PR) and instead go with the "blunt" approach of copying the seed kubeconfig onto the seed clusters. The potential downside to this is that the webhook needs to connect "from the outside" to the cluster, incurring some network overhead.

An alternate idea would have been to take the kubeconfig and modify its URL to point to the local cluster, but this would require changing the certificate and seems very hacky (especially considering that in the future we want to go with a proper RBAC approach anyway).

For all these reasons this PR goes with the simplest solution (copying the kubeconfig 1:1) that solves the current issue (broken webhook on seed clusters) and defers any RBAC improvements to a future PR. Especially since this PR is going to be backported to 2.20/2.21, I don't want to introduce RBAC changes to those release branches.

Fixes #10941

**What type of PR is this?**
/kind bug

```release-note
Fix kubermatic-webhook failing to start on external seed clusters.
```

### DIFF
--- a/cmd/kubermatic-webhook/main.go
+++ b/cmd/kubermatic-webhook/main.go
@@ -55,6 +55,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
@@ -287,7 +288,8 @@ func createGetters(ctx context.Context, log *zap.SugaredLogger, mgr manager.Mana
 
 	// if master and seed clusters are shared, the webhook can be configured
 	// with a seed name; if no name is configured, the seed getter will simply
-	// return nil.
+	// return nil. Note that just because -seed-name is given does not mean the
+	// webhook runs on a seed-only cluster, it can very well be a shared master/seed.
 	// The kubermatic-operator takes care of setting the -seed-name flag properly.
 	if options.seedName != "" {
 		seedGetter, err = seedGetterFactory(ctx, client, options)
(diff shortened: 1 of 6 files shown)


ahmedwaleedmalik (approved):
/approve

PR #10961, opened 2022-09-12
update to envoyproxy go-control-plane 0.10.3

xrstf:
Over the years I was always scared off by the scary compile errors when upgrading the go-control-plane for envoy, but as it turns out, the necessary changes were actually trivial. So this PR finally gets rid of the superfluous `replace` directive for envoy.

Regarding the kafka override, I have no idea what this is good for and it doesn't seem to break anything when I remove it. So gone it is.

/kind chore

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #10984, opened 2022-09-15
seed-proxy: increase memory limit from 32Mi to 64Mi

vgramer:
On a customer setup, the memory limit is almost reached. unfortunately we don't have enough information to know why the limit is almost (it's not the case in our dev environment), so we increase the limit to the next power of 2.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
seed-proxy: increase memory limit from 32Mi to 64Mi
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #11004, opened 2022-09-20
Use yq4 in chart testing scripts

wurbanski:
Updates test-chart-rendering.sh to use yq4

**What type of PR is this?**
Add one of the following kinds:
/kind bug
/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/hack/test-chart-rendering.sh
+++ b/hack/test-chart-rendering.sh
@@ -28,15 +28,15 @@ REALDIR="$(cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))" && pwd)"
 source ${REALDIR}/lib.sh
 
 cd ${BASEDIR}
-chartname=$(yq '.name' Chart.yaml)
+chartname=$(yq e '.name' Chart.yaml)
 echodate "(Golden master) Testing ${chartname}..."
 exitCode=0
 
 set +o errexit
-helm version
+echodate "using helm: $(helm version)"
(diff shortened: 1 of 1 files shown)


hack/test-chart-rendering.sh
```diff
@@ -28,15 +28,15 @@ REALDIR="$(cd "$(dirname $(readlink -f "${BASH_SOURCE[0]}"))" && pwd)"
 source ${REALDIR}/lib.sh
 
 cd ${BASEDIR}
-chartname=$(yq '.name' Chart.yaml)
+chartname=$(yq e '.name' Chart.yaml)
```
xrstf, line 31: I don't understand why this is necessary to explicitly call the eval function. All other scripts everywhere work fine without the `e`.

wurbanski: me neither, but it didn't work locally for me and I wasn't sure if anyone actually ran this in a while ;-)

```
> test/test.sh 
Error: unknown command ".name" for "yq"
Run 'yq --help' for usage.
```

```
> yq -V
yq version 4.9.8
```

xrstf: Your version is from Jul 8, 2021, maybe update yq 😁

wurbanski: 🙈 yeah we can close it

PR #11005, opened 2022-09-20
Limit Data When Listing AppDefs and AppInstalls

### LINKED ISSUE #10406: LIST endpoint for AppDefs/AppInstalls should return minimal data
For AppDefs it makes sense to send all the data that is required for the catalogue view. For AppInstalls possibly the things shown in the list view 

For the catalogue view, we currently use:
- name
- description

For the AppInstall view, we currently use:
- name
- name in the AppRef
- namespace
- method
- status (to display the traffic light)

SimonTheLeg:
Signed-off-by: Simon Bein <simontheleg@gmail.com>**What this PR does / why we need it**:

Fixes #10406

**What type of PR is this?**

/kind cleanup

I verified if all the required fields for the FE are present by spinning up a local FE and hooking it into a locally running BE:

image

image

```release-note
- Limit Data When Listing AppDefs and AppInstalls
- Add GET endpoint for AppDefs
```

**Documentation**:

```documentation
NONE
```


SimonTheLeg:
/hold

There seems to be a problem with follow up requests after the initial LIST:

image

image

Waseem826:
>@Waseem826 This looks to me like we only run the "LIST" and then fetch all the information from there, but we do not run a specific GET when for example editing an application or creating a new one. Is that the case?

For editing an application: yes, we only make GET request to list all application installations and then use that information in edit dialog.
For creating new application, we only make GET request to list all application definitions and then use information from the selected application (in first step).

SimonTheLeg:
@Waseem826 Okay understood:

Then I would suggest I create a ticket in the dashboard repo to tackle this. From my point of view, I think it should be like this:

AppDefs: User wants to add a new App → "List" Request to display catalogue → User selects an App → Get Request the selected App
AppInstalls: User visits the Applications Tab of their cluster → "List" Request to display list of Installations → User edits an existing App → Get Request the selected App

I will then convert this PR to a draft until we have a branch in FE which has a fix for it. And then it makes sense to test both FE and BE branches at the same time to see if everything is working

Waseem826:
Sounds good @SimonTheLeg

Waseem826:
@SimonTheLeg we also need version, source and creation timestamp because we display them in table view of applications.

screenshot-localhost_8000-2022 09 21-14_47_54

SimonTheLeg:
@vgramer One thing from my side:

Currently I have applied a top level pruning to all the encapsulating fields. For AppDefs this is pretty straight forward, resulting essentiall in name and description being shown. For AppInstalls this results in a payload like this for the list endpoint:

```json
    [{
        "name": "prometheus",
        "creationTimestamp": "2022-09-19T09:21:41Z",
        "namespace": "prometheus",
        "spec": {
            "namespace": {
                "name": "prometheus",
                "create": true
            },
            "applicationRef": {
                "name": "prometheus",
                "version": "2.36.2"
            },
            "values": {}
        },
        "status": {
            "conditions": [
                {
                    "type": "ManifestsRetrieved",
                    "status": "True",
                    "lastHeartbeatTime": "2022-09-19T09:21:45Z",
                    "lastTransitionTime": "0001-01-01T00:00:00Z",
                    "reason": "DownaloadSourceSuccessful",
                    "message": "application's source successfully downloaded"
                },
                {
                    "type": "Ready",
                    "status": "True",
                    "lastHeartbeatTime": "2022-09-19T09:21:48Z",
                    "lastTransitionTime": "0001-01-01T00:00:00Z",
                    "reason": "InstallationSuccessful",
                    "message": "application successfully installed or upgraded"
                }
            ],
            "applicationVersion": {
                "version": "2.36.2",
                "template": {
                    "source": {
                        "helm": {
                            "url": "https://prometheus-community.github.io/helm-charts",
                            "chartName": "prometheus",
                            "chartVersion": "15.10.4"
                        }
                    }
                }
            },
            "method": "helm"
        }
    }
]
```

Now if we only go by what the FE needs, we could theoretically prune a little more, up to this (I think):

```json
[
    {
        "name": "prometheus",
        "spec": {
            "namespace": {
                "name": "prometheus",
                "create": true
            },
            "applicationRef": {
                "name": "prometheus",
                "version": "2.36.2"
            }
        },
        "status": {
            "conditions": [
                {
                    "type": "ManifestsRetrieved"
                },
                {
                    "type": "Ready"
                }
            ],
            "applicationVersion": {
                "version": "2.36.2",
                "template": {
                    "source": {
                        "helm": {
                        }
                    }
                }
            },
            "method": "helm"
        }
    }
]
```

However I am currently undecided if this is a good idea, as:
- it requires us to maintain more complexity in our api, by at least creating 3 more types
- it does create a not so nice looking empty field, because for `applicationVersion.template.source` we use directly the json keys `helm` or `git` instead of reading a json value

Wdyt?

vgramer:
@SimonTheLeg for the conditions we need more information because, if an error happens the condition  will be set false and the message will contains the error (e.g. helm apply error )

For the source part, i don't have strong opinion about it... We may use a string type instead if users don't need this info ?

SimonTheLeg:
as the code for the api has been moved to dashboard, please follow this PR for further updates:
https://github.com/kubermatic/dashboard/pull/5184

PR #10816, opened 2022-08-22
Anexia: allow extended disk configuration in AnexiaNodeSpec

LittleFox94:
**What this PR does / why we need it**:

This extends `apiv1.AnexiaNodeSpec` to allow extended disk configuration introduced in machine-controller with https://github.com/kubermatic/machine-controller/pull/1402.

Without this, Kubermatic API returns a 500 when trying to do any `MachineDeployment` operation on a cluster which has a `MachineDeployment` with the new `disks` attribute (and thus without the old `diskSize` attribute).

**What type of PR is this?**
/kind bug
/kind feature
/kind api-change

It's a bug (500 error) because of a new feature, which causes and api-change

**Special notes for your reviewer**:

Comments on the machine-controller PRs:
* https://github.com/kubermatic/machine-controller/pull/1405#issuecomment-1220500780
* https://github.com/kubermatic/machine-controller/pull/1402#issuecomment-1220506220

If the [backport to machine-controller v1.45](https://github.com/kubermatic/machine-controller/pull/1405) is approved, we need this backported to kubermatic version v2.19 and v2.20.

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Extend disk configuration for provider Anexia
```

**Documentation**:

```documentation

```

PR #11030, opened 2022-09-21
Fix Anexia disk config

LittleFox94:

There is a mistake in our new disk config handling code that slipped through in #10816 - this fixes that.

When having a `MachineDeployment` in a user cluster with only the new `disks` attribute set, API still responds with an error stating both the old and the new attribute are set:
`{"error":{"code":500,"message":"json: error calling MarshalJSON for type *v1.AnexiaNodeSpec: missing or invalid required parameter(s): both disks and diskSize configured but only one of those allowed"}}`

**What type of PR is this?**
/kind bug

I have backport branches prepared for v2.20 and v2.21 and will create the PRs once this here is approved.

```release-note bugfixes
Fix API error in extended disk configuration for provider Anexia
```

**Documentation**:

```documentation

```

### DIFF
--- a/pkg/machine/convert.go
+++ b/pkg/machine/convert.go
@@ -44,6 +44,7 @@ import (
 	apiv1 "k8c.io/kubermatic/v2/pkg/api/v1"
 
 	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/utils/pointer"
 )
 
 // GetAPIV1OperatingSystemSpec returns the api compatible OperatingSystemSpec for the given machine.
@@ -348,11 +349,10 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 				TemplateID: config.TemplateID.Value,
 				CPUs:       config.CPUs,
 				Memory:     int64(config.Memory),
-				DiskSize:   &[]int64{int64(config.DiskSize)}[0],
 			}
 
 			if config.DiskSize > 0 {
-				cloudSpec.Anexia.DiskSize = &[]int64{int64(config.DiskSize)}[0]
+				cloudSpec.Anexia.DiskSize = pointer.Int64(int64(config.DiskSize))
 			}
 
 			if diskCount := len(config.Disks); diskCount > 0 {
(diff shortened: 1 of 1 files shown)


pkg/machine/convert.go
```diff
@@ -348,11 +348,10 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 				TemplateID: config.TemplateID.Value,
 				CPUs:       config.CPUs,
 				Memory:     int64(config.Memory),
-				DiskSize:   &[]int64{int64(config.DiskSize)}[0],
 			}
 
 			if config.DiskSize > 0 {
-				cloudSpec.Anexia.DiskSize = &[]int64{int64(config.DiskSize)}[0]
+				cloudSpec.Anexia.DiskSize = &[]int64{int64(config.DiskSize)}[config.DiskSize]
```
xrstf: Surely `k8s.io/utils/pointer` has a helper that prevents this .... creative approach to creating pointers.

xrstf: Same goes for `cloudSpec.Anexia.Disks[diskIndex].PerformanceType = &[]string{diskConfig.PerformanceType.Value}[0]` a bit further down.

LittleFox94: Took that from other code in this code base, actually learned it here ^^`

xrstf: Also using [config.DiskSize] as the index here is probably also unintended :grin:

LittleFox94: But can definitely change it, it's at least the only occurrence in this file - maybe the original ones I saw were in machine-controller even

LittleFox94: > Also using [config.DiskSize] as the index here is probably also unintended grin

oof right, thanks, too long day ... will deliver a fixed version first thing tomorrow - before I make more humiliating stuff today ^^'

LittleFox94: addressed both problems, thanks again :)

xrstf: It's now so much prettier, don't you agree? :D

LittleFox94: it is! but as said, I just oriented myself on existing code at some place :P

xrstf: However, it seems that it doesn't compile anymore... Weird.

LittleFox94: huh? I'll take a look

LittleFox94: Fix incoming

LittleFox94: Sorry for the noise, should be fine now - locally ran the tests (`make test`) this time, successfully

xrstf (approved):
/approve

PR #11031, opened 2022-09-21
user MLA: Enable alert-management with Grafana

schnidrig:

Newer versions of Grafana now include much better alert management. This is a huge improvement over the Cortex Alertmanager. However, for this to work we need to adapt the grafana datasources created by the seed-controller-manager.

This PR adds the following:
- Alertmanager datasource in Grafana
- Specify in Loki & Prometheus datasources the corresponding Alertmanager datasource
- add the Rest-API URLs to the MLA-Gateway which are used for alertmangement.

It allows for:
- Displaying alerts in Grafana (instead of only Cortex-Alertmanager)
- Editing silences in Grafana

It does not enable editing alert rules in Grafana since the rules would later again be overwitten by the rules defined via Kubermatic-API.

**What type of PR is this?**
/kind feature

This PR also needs https://github.com/kubermatic/mla/pull/142
The alertmanger datasource is added when either user-mla monitoring or user mla logging is enabled.

```release-note
enable alert-management using Grafana
```


wurbanski:
Hi @schnidrig! Thanks for this contribution!

As you've already seen (😄) I'm in the process of making some changes to mla installation and repository structure, but after that first step is done, we will definitely get back to this. 
It seems like a good addition and I know that with this new approach to alerting in Grafana it is becoming much more popular, so we will do our best to facilitate this usage model.

PR #11035, opened 2022-09-22
add method to handle the os patch in the Patch Machine Deployment end…

### LINKED ISSUE #6799: Kubermatic API: Patch of MachineDeployments UX improvable
no high prio, but hard to find if you run into

**User Story**
If a user currently updates the machine deployment and changes the OS system over the KKP API endpoint:
```
/api/v2/projects/{project_id}/clusters/{cluster_id}/machinedeployments/{machinedeployment_id}
```
It seams, that existing Operating System does not changes if e.g. spec changes like here:

The get response still contains the old OS.

Only if also the old OS get set to null, the patch works.

**Acceptance criteria**
From the UX perspective this not ideal and could cause a lot of try-and-error at the use side. It would be better if the API:
- apply the patch correctly
- API spec gives an hint
- OR all operating system specs needs to be set, nevertheless if its's null or not (not preferred)

ahmadhamzh:
currently when the user send request to change the OS in machine deployment with the new OS it won't work till he set the value for the other OS to null in this pr we change that so the user can only give the value of the new OS 
Fixes #6799 

**What type of PR is this?**
/kind bug

```release-note
NONE
```
**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/handler/common/machine.go
+++ b/pkg/handler/common/machine.go
@@ -461,6 +461,20 @@ func PatchMachineDeployment(ctx context.Context, userInfoGetter provider.UserInf
 	if err != nil {
 		return nil, fmt.Errorf("cannot output existing node deployment: %w", err)
 	}
+	var unmarshalPatched *apiv1.NodeDeployment
+	if err := json.Unmarshal(patch, &unmarshalPatched); err != nil {
+		return nil, fmt.Errorf("cannot decode patched nodedeployment: %w ===", err)
+	}
+
+	selectedOperatingSystems := selectedOperatingSystems(unmarshalPatched.Spec.Template.OperatingSystem)
+
+	if selectedOperatingSystems > 1 {
+		return nil, fmt.Errorf("cannot have more than one os")
+	}
+
+	if selectedOperatingSystems == 1 {
+		nodeDeployment.Spec.Template.OperatingSystem = unmarshalPatched.Spec.Template.OperatingSystem
+	}
 
 	nodeDeploymentJSON, err := json.Marshal(nodeDeployment)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/handler/common/machine.go
```diff
@@ -461,6 +461,20 @@ func PatchMachineDeployment(ctx context.Context, userInfoGetter provider.UserInf
 	if err != nil {
 		return nil, fmt.Errorf("cannot output existing node deployment: %w", err)
 	}
+	var unmarshalPatched *apiv1.NodeDeployment
+	if err := json.Unmarshal(patch, &unmarshalPatched); err != nil {
+		return nil, fmt.Errorf("cannot decode patched nodedeployment: %w ===", err)
+	}
+
+	isOsCahnged := isOsCahnged(unmarshalPatched.Spec.Template.OperatingSystem)
```
ahmedwaleedmalik: Maybe we can rename this to something like `selectedOperatingSystems`? `isOsCahnged` is a bit ambigous since we are checking for the existence of OS in the payload.

pkg/handler/common/machine.go
```diff
@@ -862,3 +876,29 @@ func getNodeForMachine(machine *clusterv1alpha1.Machine, nodes []corev1.Node) *c
 	}
 	return nil
 }
+
+func isOsCahnged(os apiv1.OperatingSystemSpec) int {
```
ahmedwaleedmalik:
```suggestion
func isOsChanged(os apiv1.OperatingSystemSpec) int {
```

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik (approved):
/approve

dharapvj:
Why would we add method to patch one attribute at a time?!

The original issue was to say that if we provide updated MD definition, it should reconcile ask the fields.. otherwise POST body processing is ambiguous and is not consistent with processing logic for all the other POST calls e.g cluster, project etc

PR #11049, opened 2022-09-23
Add Kubernetes 1.25

xrstf:
This PR adds support for Kubernetes 1.25 to KKP and raises default version for new clusters to 1.24.6. The various CSI addons have been updated to their latest versions.

GCP is marked as unsupported for 1.25 because setting `CSIMigrationGCE=false` is not possible anymore and adding the GCP CSI driver addon is better left for another PR. The GCP CCM is the main blocking factor here, as we have not yet managed to actually get it to run and KKP has entangled CCM/CSI migration. For GCP we might have to break that rule and go with "in-tree CCM, out-of-tree CSI driver" to make 1.25 work and that's why I would want to do it in another PR.

/kind feature

```release-note
Add support for Kubernetes 1.25; raise default version for new clusters to 1.24.6.
```

**Documentation:**
```documentation
NONE
```


pkg/resources/cloudcontroller/aws.go
```diff
@@ -142,9 +142,11 @@ func awsDeploymentCreator(data *resources.TemplateData) reconciling.NamedDeploym
 func getAWSCCMVersion(version semver.Semver) string {
 	switch version.MajorMinor() {
 	case v124:
+		return "v1.24.0"
+	case v125:
 		fallthrough
 	//	By default return latest version
 	default:
-		return "v1.24.1"
+		return "v1.25.0"
```
kron4eg: there is v1.25.1 version of the registry.k8s.io/provider-aws/cloud-controller-manager

PR #11067, opened 2022-09-26
WIP - consolidate where externalCCM feature flag is set for new clusters

xrstf:
This PR moves the code to enable the externalCCM during cluster creation out of the handler package into our webhook. It's a simple defaulting that needs to happen _during_ creation (not after it), so a controller would be too late in setting the flag IMHO.

This PR then removes the 2 spots where the externalCCM feature gate was set previously (handler and kube_client for the conformance-tester), plus it simplifies some code around checking the feature gate.

Similarily, the flags for enabling OSM/UserSSHKeyAgent are defaulted by the webhook, so our handlers do not need to be concerned with this.

All of this is in the context of me refactoring our CCM/CSI migration handling a bit in preparation for AWS/GCE CSI changes in Kubernetes 1.25.

/kind cleanup

```release-note
NONE
```


xrstf:
Replaced by #11095.

PR #11079, opened 2022-09-27
Use `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes upstream images

### LINKED ISSUE #9987: Switch images from k8s.gcr.io to registry.k8s.io
As per [KEP 3000](https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/3000-artifact-distribution) (also see [this PR](https://github.com/kubernetes/kubernetes/pull/109938)), image distribution for Kubernetes upstream images is moving from k8s.gcr.io to registry.k8s.io. We should investigate if all images we need are available from there and make the switch. According to the email that went out on dev-kubernetes, this new registry will be default for 1.25.

## Acceptance Criteria

- All images that were using `k8s.gcr.io` as registry use `registry.k8s.io` so we do not drive any traffic to the old registry

embik:

Kubernetes upstream introduced a new registry URL called `registry.k8s.io`, which acts as a kind of proxy (well, technically it only redirects) for upstream images. This can speed up the image pull process if a local mirror is available that you're forwarded to (mainly on AWS, as far as I understand from upstream documentation) and is the suggested source of images now. Let's switch to it.

Fixes #9987

**What type of PR is this?**
/kind chore

```release-note
ACTION REQUIRED: Use `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes upstream images. It might be necessary to update firewall rules or mirror registries accordingly
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/addons/cluster-autoscaler/cluster-autoscaler.yaml
+++ b/addons/cluster-autoscaler/cluster-autoscaler.yaml
@@ -176,7 +176,7 @@ spec:
         app: cluster-autoscaler
     spec:
       containers:
-      - image: '{{ Registry "k8s.gcr.io" }}/autoscaling/cluster-autoscaler:{{ $version }}'
+      - image: '{{ Registry "registry.k8s.io" }}/autoscaling/cluster-autoscaler:{{ $version }}'
         name: cluster-autoscaler
         command:
         - /cluster-autoscaler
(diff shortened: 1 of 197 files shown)


xrstf (approved):
/approve

PR #10759, opened 2022-08-15
Use Secrets for cloud-configs on the seed cluster

xrstf:
### What this PR does / why we need it
This changes the storage for the cloud config and the CSI cloud config to use Secrets instead of ConfigMaps. I used Secrets for every cloud provider because it makes (keeps) the code much easier, also most providers actually do have credentials in their configs:

#### cloud config

* aws: clean
* azure: AD client secret
* openstack: username+password
* vsphere: username+password
* gcp: clean
* kubevirt: clean

#### csi cloud config

* vsphere: username+password
* vcd: clean
* nutanix: username+password
* kubevirt: kubeconfig

This PR also changes two constants:

* `CloudConfigConfigMapName` => `CloudConfigSeedSecretName`
* `CSICloudConfigName` => deleted (identical to `CSICloudConfigSecretName`)

To allow for a smoother transition, I implemented a simple fallback to the ConfigMap. This fallback can esily be removed in KKP 2.23.

### Which issue(s) this PR fixes
Fixes #7132

### What type of PR is this
/kind feature

### Does this PR introduce a user-facing change?
```release-note
The cloud-config and cloud-config-csi for userclusters is now stored in a Secret instead of a ConfigMap on the seed cluster (in each usercluster namespace).
```

PR #11095, opened 2022-09-28
Improve CCM/CSI Feature Gate logic

xrstf:
## What this PR does

KKP applied different rules for the `externalCloudProvider` feature flag, depending on whether KubeAPI or the KKP dashboard is used. This leads to unexpected results. This PR cleans up the CCM handling a bit to make it easier to understand and remove exceptions in the codebase.

The actual CSI migration logic and how the CLI flags are set on the various components is not affected by this PR. This is only about when and how the `externalCloudProvider` feature flag is set.

### External CCM Support Check

The code has been changed to rely on `resources.ExternalCloudControllerFeatureSupported()` in all cases, the hardcoded exceptions for Anexia/Kubevirt have been removed accordingly. The function was moved out of `pkg/resources/cloudcontroller/` because of an import loop, as now the `pkg/defaulting` also requires to check the CCM support.

The external CCM check has been removed alltogether in the KKP API and the KubeAPI conformance tester, as they can and should simply rely on the webhook doing the proper mutations.

The entire logic to check for external CCM support has been simplified. it now relies on the "provider incompatibility" mechanism, which has existed for some releases in KKP already. The most basic check is still hardcoded (that's why there is no incompat rule for AWS, as AWS is not even a `case` statement in Go), but version dependencies are now handled via the incompat rules.

As it turns out, for all Kubernetes versions supported by KKP for Openstack/Kubevirt/Azure, we already have external CCM support, so the checks (`OpenstackExternalCloudProviderSupported()`) could simply be removed alltogether.

The support for external CCM is now also checked during UPDATE calls, so one cannot enable it for AWS right now, for example.

### The case for `externalCloudController=false`

Now that the webhook defaults to enabling the CCM for new clusters, the CCM migration tests would have a hard time to create an "old" cluster that they can actually migrate. We considered 2 options:

1. Introduce a magic annotation that only works in dev mode.
2. Allow setting the feature to false and _do not overwrite_ that choice.

Not having an annotation seems cleaner and in due time we will have a check in the KKP installer that rejects further KKP upgrades when clusters without external CCM still exist.

Fun fact: Only because the CCM tests did _not_ use the KKP API were they still able to create clusters without external CCM. If the webhook already had worked consistently with the KKP API, the tests would have been NOPs.

### Reconciling Cleanups

The Deployment creators for the various CCMs have been harmonized and cleaned up a bit. Their handling for mounting the cloud-config, ccm-kubeconfig etc. is now more uniform. This also means that `SSL_CERT_FILE` is now consistently set.

### Cluster Mutation Handler

The handler was split into admission handler and mutator, simply because this made testing and understanding the mutator easier.

### Test Fixtures (aka "Minus 6000 lines?")

It seems that in previous PRs, for example #10759, did not remove leftover fixture files. In this case, as we do not generate fixtures for Secrets, it seems that the cloud-config fixtures "got lost", but that's just normal.

There are some changes in the fixtures caused by this PR, but most is just garbage collected.

### Outlook

A future, already prepared PR (the 2nd part of #11095) will then add support for the AWS CCM/CSI, bringing us one step closer to adding Kubernetes 1.25 support (1.25 enables CSIMigration by default, which tripped us up when adding support for it, and made us finally add AWS CCM support in the first place).

/kind cleanup

```release-note
By default, the `externalCloudProvider` feature will be turned on for all newly created clusters (if the provider supports it). Previously this only applied to Anexia/Kubevirt clusters (and to VSphere/Openstack/Azure clusters when using the KKP dashboard). Now the same rules apply no matter how the cluster is provisioned. The feature gate can still be explicitly set to `false`, but eventually all clusters must be migrated to the external CCM/CSI as in-tree support in Kubernetes is phased out.
```

## Documentation

```documentation
NONE
```


embik (approved):

lsviben:
opa e2e test are failing consistently, so something broke it. Ill try to debug to see what is going on

xrstf:
/hold

to prevent retries

xrstf:
I think I found it: The hcloud CCM is now dead because I removed the NODE_NAME env, leading to

> F0929 12:01:55.614990       1 controllermanager.go:124] Cloud provider could not be initialized: could not init cloud provider "hcloud": environment variable "NODE_NAME" is requiredF0929 12:01:55.614990       1 controllermanager.go:124] Cloud provider could not be initialized: could not init cloud provider "hcloud": environment variable "NODE_NAME" is required

As the hcloud-ccm runs on the seed, NODE_NAME is technically irrelevant, but if the CCM insists on it being defined....

embik (approved):

PR #11101, opened 2022-09-29
prevent dangling resource fixtures, ensure name/namespace are set

xrstf:
This PR improves the resource fixture tests by

1. ensuring that a realistic name/namespace are set
2. no dangling files remaing when `-update` is used
3. if dangling files are present, the test will fail

/kind cleanup

```release-note
NONE
```

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #11057, opened 2022-09-23
WIP - Implement External CCM / CSI for AWS

xrstf:
**What this PR does / why we need it**:
Kubernetes 1.25 makes the AWS CSI GA and enforces the CSI migration. To enable this on our side, this PR adds the first part of it, the CSI addon for AWS clusters. As we do not yet support 1.25 in KKP, this addon will so far not be deployed anywhere.

/kind feature

**Does this PR introduce a user-facing change? Then add your Release Note here**:
```release-note
Add AWS EBS CSI driver for Kubernetes 1.25+
```

PR #11102, opened 2022-09-29
Implement External CCM / CSI for AWS

xrstf:

Kubernetes 1.25 makes the AWS CSIMigration GA and enforces the CSI migration, unless one explicitly disables it by setting the `CSIMigrationAWS=false` feature flag. We took this opportunity to finally add support for the AWS CCM and CSI, as adding them together in one KKP release makes the CSI migration logic easier (AWS does not turn into a "special" provider with external CSI but in-tree CCM).

### EBS CSI

This PR adds a new `csi/aws-ebs` addon based on the official upstream Helm chart for the EBS CSI driver. This addon is deployed into all AWS clusters with `externalCloudProvider=true`. 

Likewise a new StorageClass is defined for when the CSI driver is enabled.

### AWS Provider in KKP

The provider code was adjusted to provide the new AWS IAM policies required. It turns out that our control-plane policy is already working fine for the CCM, so only the policy for the new CSI was added to the worker policy.

### AWS CCM

There's not much to say about the CCM. It works identical to all the others, uses the same flags and just does its job.

### Cluster Lifecycle

As AWS is now marked as a supported provider, the `externalCloudProvider` feature flag is enabled by default for all newly created clusters using Kubernetes 1.24+. Existing clusters will have to be migrated manually by setting the feature flag.

/kind feature

replaces #11057

## Release Note

```release-note
Add support for the AWS External CCM & EBS CSI driver for Kubernetes 1.24+; newly created AWS clusters will use the `externalCloudProvider` feature flag automatically, existing clusters must be migrated manually.
```

## Documentation

```documentation
https://github.com/kubermatic/docs/pull/1342
```


pkg/resources/cloudcontroller/aws.go
```diff
@@ -0,0 +1,149 @@
+/*
+Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
ahmedwaleedmalik:
```suggestion
Copyright 2022 The Kubermatic Kubernetes Platform contributors.
```

addons/Makefile
```diff
@@ -47,6 +47,22 @@ aws-node-termination-handler:
...
+	mkdir -p csi/aws-ebs
+	cat csi/aws-ebs/_header.txt > $(OUTPUT_FILE)
+	helm --namespace kube-system template aws-ebs-csi-driver aws-ebs-csi-driver/aws-ebs-csi-driver \
+	  --set 'controller.k8sTagClusterId=\{{ .Cluster.Name }}' \
+	  --set 'node.securityContext.seccompProfile.type=RuntimeDefault' \
+	  --set 'controller.securityContext.seccompProfile.type=RuntimeDefault' \
+	  >> $(OUTPUT_FILE)
+	cat csi/aws-ebs/_footer.txt >> $(OUTPUT_FILE)
+	sed -i 's/public.ecr.aws/{{ Registry "public.ecr.aws" }}/g' $(OUTPUT_FILE)
+	sed -i 's/k8s.gcr.io/{{ Registry "k8s.gcr.io" }}/g' $(OUTPUT_FILE)
```
xmudrii: Nit: maybe we could replace it with `registry.k8s.io` instead.

pkg/provider/cloud/aws/role_test.go
```diff
@@ -100,20 +100,32 @@ func TestEnsureRole(t *testing.T) {
...
 		cluster := makeCluster(&kubermaticv1.AWSCloudSpec{})
 		roleName := controlPlaneRoleName(cluster.Name)
-		policies := map[string]string{workerPolicyName: workerRolePolicy}
+
+		policy, err := getControlPlanePolicy(cluster.Name)
+		if err != nil {
+			t.Fatalf("failed to build the worker policy: %v", err)
+		}
+
+		policies := map[string]string{workerPolicyName: policy}
```
xmudrii: I'm a little bit confused here. Old tests were taking the worker policy, but new tests are taking the control plane policy. Is that as expected?

pkg/resources/cloudcontroller.go
```diff
@@ -50,6 +50,9 @@ func ExternalCloudControllerFeatureSupported(dc *kubermaticv1.Datacenter, cloudS
 
 		fallthrough
 
+	case cloudSpec.AWS != nil:
+		fallthrough
```
xmudrii: Shouldn't this take incompatibilities in the account similar to vSphere?

pkg/resources/cloudcontroller.go
```diff
@@ -50,6 +50,9 @@ func ExternalCloudControllerFeatureSupported(dc *kubermaticv1.Datacenter, cloudS
 
 		fallthrough
 
+	case cloudSpec.AWS != nil:
```
xmudrii: I can't leave a comment on L79, but shouldn't `MigrationToExternalCloudControllerSupported` function have a case for AWS?

pkg/resources/test/fixtures/deployment-aws-1.24.0-apiserver-externalCloudProvider.yaml
```diff
@@ -0,0 +1,417 @@
...
+        - --service-account-signing-key-file
+        - /etc/kubernetes/service-account-key/sa.key
+        - --api-audiences
+        - https://jh8j81chn.europe-west3-c.dev.kubermatic.io:30000
+        - --kubelet-preferred-address-types
+        - ExternalIP,InternalIP
+        - --cloud-provider
+        - aws
+        - --cloud-config
+        - /etc/kubernetes/cloud/config
```
xmudrii: This seems wrong if this is the API server with external CCM enabled. In that case, you shouldn't have those two flags at all.

pkg/resources/test/fixtures/deployment-aws-1.24.0-aws-cloud-controller-manager-externalCloudProvider.yaml
```diff
@@ -0,0 +1,139 @@
...
+        - https://apiserver-external.cluster-de-test-01.svc.cluster.local./healthz
+        - -insecure
+        - -retries
+        - "100"
+        - -retry-wait
+        - "2"
+        - -timeout
+        - "1"
+        - -command
+        - '{"command":"/bin/aws-cloud-controller-manager","args":["--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig","--cloud-config=/etc/kubernetes/cloud/config","--cloud-provider=aws","--cluster-cidr=172.25.0.0/16"]}'
```
xmudrii, line 35: Same as above, if this is with the external CCM, you shouldn't have cloud-provider and cloud-config flags at all.

pkg/provider/cloud/aws/policy.go
```diff
@@ -49,149 +53,304 @@ var (
...
       }
     },
     {
       "Effect": "Allow",
       "Action": [
-        "elasticloadbalancing:CreateLoadBalancer",
-        "elasticloadbalancing:CreateLoadBalancerPolicy",
-        "elasticloadbalancing:CreateLoadBalancerListeners",
-        "elasticloadbalancing:CreateListener",
-        "elasticloadbalancing:CreateTargetGroup"
```
xmudrii, line 133: Are we sure that we don't need those ELB policies any longer? What about clusters running the in-tree provider?

xmudrii (commented):
Some nits and questions, but otherwise, looks good.

pkg/resources/cloudcontroller.go
xrstf: That's why it's using `fallthrough` and the vsphere `case` statement uses `kubermaticv1.ProviderType(cloudSpec.ProviderName)` to work for all providers, not just vsphere.

pkg/resources/cloudcontroller.go
xrstf: I thought so at first, but I wasn't sure. For some providers we simply do not do a migration in the UI and I have no idea why or what's the criteria for that. I can change the code here :man_shrugging:

pkg/provider/cloud/aws/policy.go
xrstf: The permissions are still there, in the control plane policy. I just massively extended the worker policy, so maybe the diff made it hard to read?

pkg/resources/cloudcontroller.go
xmudrii: I'm not sure that I like that at all. Someone might add something in between that would break this workflow. It's better to avoid using fallthrough and just combine those cases into a single case.

pkg/resources/cloudcontroller.go
xmudrii: Looks good after the change.

pkg/provider/cloud/aws/policy.go
xmudrii: > The permissions are still there, in the control plane policy. I just massively extended the worker policy, so maybe the diff made it hard to read?

I was wondering if we need those permissions in the worker policy. I'm not sure if kubelet is handling anything related to LBs if the in-tree provider is used.

xmudrii: I see, thanks for clarifying it.

pkg/provider/cloud/aws/policy.go
```diff
@@ -49,149 +53,304 @@ var (
...
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateVolume"
       ],
+      "Resource": "*",
       "Condition": {
-        "Null": {
-          "ec2:ResourceTag/{{ .ClusterTag }}": "false"
+        "StringLike": {
+          "aws:RequestTag/{{ .EBSCSIClusterTag }}": "true"
```
xmudrii, line 110: How is this going to behave when we use the in-tree cloud provider? In that case, is `ebs.csi.aws.com/$CLUSTER_NAME` a valid condition since `kubernetes.io/aws-ebs` is used as a provisioner instead?

xmudrii (commented):
Just one more question.

pkg/provider/cloud/aws/policy.go
xrstf: screenshot-2022-10-05-132212

The new policy contains variants for both the old and the new tags, so I think we should be fine. This comes straight from upstream: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/example-iam-policy.json#L48-L83

xmudrii: @xrstf Thanks for the clarification!

xmudrii (approved):
Looks great! :heart: :zap: 

/lgtm
/approve

ahmedwaleedmalik (approved):
/approve

embik:
@xrstf please update docs to mention AWS CCM here: https://docs.kubermatic.com/kubermatic/main/tutorials-howtos/ccm-migration/. I've changed the docs section of this PR from `NONE` to `TBD` to track this.

PR #11054, opened 2022-09-23
[release/v2.20] Use unordered comparison of vsphere networks (#11048)

embik:
**What this PR does / why we need it**:

Manual backport of #11048.

**Which issue(s) this PR fixes**:

Fixes #

**What type of PR is this?**

/kind chore

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

Issue #10949, opened 2022-09-12
Fix CCM migration tests for release branch 2.20


ahmedwaleedmalik:
### What happened?

In CCM migration tests, the nodes are unable to join the user cluster and it results in failure for the tests.

Logs from `kubelet` running on the node:
```
Sep 12 07:39:30 ccm-migration-e2e-debug-645bb6f796-66b7q kubelet[21179]: E0912 07:39:30.843943   21179 kubelet_node_status.go:93] Unable to register node "ccm-migration-e2e-debug-645bb6f796-66b7q" with API server: Post "https://x5k8cmng44.kubermatic.127.0.0.1.nip.io:6443/api/v1/nodes": dial tcp 127.0.0.1:6443: connect: connection refused
```

### How is your environment configured?

- KKP version:
- Shared or separate master/seed clusters?:

### Provide your KKP manifest here (if applicable)

<details>

```yaml
# paste manifest here
```

</details>

xrstf:
Fixed by #11154

PR #11104, opened 2022-09-29
[release/v2.20] Fix vSphere network test

xrstf:
This is what #11054 tried to do, but much simpler and without relying on generics.

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
We already have a ticket for fixing the CCM migration tests https://github.com/kubermatic/kubermatic/issues/10949. I can take care of that on Monday so it's fine to override that test in this PR imo.

PR #11115, opened 2022-09-30
add EtcdLauncherNotEnabled phase to etcd restore

### LINKED ISSUE #9665: Add EtcdLauncherNotEnabled state to phases of EtcdRestore
EtcdLauncher is required for etcd restore feature to work properly. This story is part of enhancements planned to make it easy for users to detect when restores are not working properly. 

See discussion here https://github.com/kubermatic/kubermatic/pull/9633 and other linked PRs.

PratikDeoghare:
Adds `EtcdLauncherNotEnabled` phase to `EtcdRestore`. 

Fixes #9665 

**What type of PR is this?**

```release-note
EtcdRestores are moved to a 'EtcdLauncherNotEnabled' phase if required etcd-launcher is not enabled
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/apis/kubermatic/v1/etcd_restore.go
+++ b/pkg/apis/kubermatic/v1/etcd_restore.go
@@ -38,7 +38,7 @@ const (
 	EtcdRestorePhaseCompleted EtcdRestorePhase = "Completed"
 
 	// EtcdRestorePhaseEtcdLauncherNotEnabled value indicating that etcd-launcher is not enabled.
-	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdRestorePhaseEtcdLauncherNotEnabled"
+	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdLauncherNotEnabled"
 )
 
 // +kubebuilder:validation:Enum=Started;StsRebuilding;Completed;EtcdRestorePhaseEtcdLauncherNotEnabled
--- a/pkg/apis/kubermatic/v1/etcd_restore.go
+++ b/pkg/apis/kubermatic/v1/etcd_restore.go
@@ -41,7 +41,7 @@ const (
 	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdLauncherNotEnabled"
 )
 
-// +kubebuilder:validation:Enum=Started;StsRebuilding;Completed;EtcdRestorePhaseEtcdLauncherNotEnabled
+// +kubebuilder:validation:Enum=Started;StsRebuilding;Completed;EtcdLauncherNotEnabled
 
 // EtcdRestorePhase represents the lifecycle phase of an EtcdRestore.
 type EtcdRestorePhase string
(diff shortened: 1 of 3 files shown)


embik:
Might be a stupid question, but why are we adding this? Understanding this phase on a restore object would require knowledge of the dependency on etcd-launcher, no? Otherwise this would - as a user - not really help me understand what's wrong.

PratikDeoghare:
> Might be a stupid question, but why are we adding this? 

This was considered the best of of all ways (that we considered) of informing user that their etcd-restore might be broken. 

See https://github.com/kubermatic/kubermatic/pull/9633#discussion_r852708342 

> Understanding this phase on a restore object would require knowledge of the dependency on etcd-launcher, no? 

Yes. 
They learn anyway by raising ticket after debugging broken restore for a while. 
This can inform them that something is wrong and they can search with that string to find some docs. 
Saving them some debugging and saving us some ticket would be good, no?

embik:
Yeah disregard my previous comment, I now remember the discussion after you've linked. My bad!

pkg/apis/kubermatic/v1/etcd_restore.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// EtcdRestorePhaseCompleted value indicating that the old Etcd statefulset has completed successfully.
 	EtcdRestorePhaseCompleted EtcdRestorePhase = "Completed"
+
+	// EtcdRestorePhaseEtcdLauncherNotEnabled value indicating that etcd-launcher is not enabled.
+	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdRestorePhaseEtcdLauncherNotEnabled"
 )
 
 // +kubebuilder:validation:Enum=Started;StsRebuilding;Completed
```
embik: This needs to be updated to include the new phase. Will probably require updating the CRD codegen as well.

embik (changes requested):

pkg/apis/kubermatic/v1/etcd_restore.go
```diff
@@ -36,9 +36,12 @@ const (
 
 	// EtcdRestorePhaseCompleted value indicating that the old Etcd statefulset has completed successfully.
 	EtcdRestorePhaseCompleted EtcdRestorePhase = "Completed"
+
+	// EtcdRestorePhaseEtcdLauncherNotEnabled value indicating that etcd-launcher is not enabled.
+	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdRestorePhaseEtcdLauncherNotEnabled"
```
embik: Sorry I should have caught this earlier, but this should remove the `EtcdRestorePhase` prefix. It's not on the other phases either.

```suggestion
	EtcdRestorePhaseEtcdLauncherNotEnabled EtcdRestorePhase = "EtcdLauncherNotEnabled"
```

Same goes for the enum value of course.

embik (changes requested):
As is tradition with my reviews, I don't catch everything on first review. Apologies, but this should be fixed IMO.

embik (approved):

Issue #11063, opened 2022-09-26
kubermatic-installer and KKP controllers/operator should verify the Kubermatic edition on startup


xmudrii:
### Description of the feature you would like to add / User story

As a KKP admin
I would like to have kubermatic-installer and KKP controllers/operator verify the Kubermatic edition is correct
in order to avoid issues due to upgrading to a wrong edition

### Solution details

- If possible, kubermatic-installer should verify that the provided KubermaticConfiguration matches the kubermatic-installer's edition
- KKP controllers and operator should fail to start if KubermaticConfiguration is EE-based, but components are CE-based

### Use cases

It happens that users download CE-based kubermatic-installer and use it on EE-based KKP installation. That causes some non-subtle issues, such as KKP dashboard failing to load clusters. It might also happen that users provide CE-based images in KubermaticConfiguration and then use EE-based kubermatic-installer (see #11062).

xrstf:
Allowing to upgrade from CE to EE is an intended feature. 😁 Do we really want to require a dedicated `--allow-edition-change` flag in the installer? 🤔

xmudrii:
I think that use case is fine, however, we should consider a confirmation of some sort if you downgrade from EE to CE. In the most of cases it’s by mistake.

PR #11128, opened 2022-10-04
validate that admins do not accidentally change the KKP edition during updates

xrstf:
This implements the installer part of the ticket mentioned below, but skips the controller part, as I am not sure how we want to implement such a thing even (adding CLI flags to our controller-managers seems super weird).

Fixes part of #11063

/kind feature

```release-note
The `kubermatic-installer` will now reject installing a different KKP edition over an existing one (for example installing the Community Edition over a previous Enterprise Edition). This safety check can be disabled by adding `--allow-edition-change` to the installer. Installing the EE over the CE (i.e. upgrading) is supported, this change just prevents accidental mixups.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/cmd/kubermatic-installer/cmd_deploy.go
+++ b/cmd/kubermatic-installer/cmd_deploy.go
@@ -63,8 +63,9 @@ type DeployOptions struct {
 	SkipDependencies bool
 	Force            bool
 
-	StorageClass     string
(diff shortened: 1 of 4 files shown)


xmudrii (approved):
/lgtm
/approve

PR #11130, opened 2022-10-04
Configure machine flavor filters at datacenter level

### LINKED ISSUE #10207: Configure MachineDeploymentVMResourceQuota per Datacenter instead of Seed
**User Story** 
As a user, I would like to filter cloud instances(vms) based on resources(RAM and CPU) per datacenter, instead of having configured on a global level(master/seed). The reason behind that is, the need to have customize different datacenter based on the customers needs. 

For example, a customer would like to provision small instances in datacenter A while keeping large instance only in datacenter B. This scenario is pretty important while dealing with bare metal provisioning and with edge computing. At the moment this is not possible, once a user specifies a resource limits, it will be propagated across all datacenter, across all seeds.

The field `MachineDeploymentVMResourceQuota` should be probably moved to the seed object, to the datacenter field to be very precise. Users can set this in admin panel but they should be able to choose to which datacenter exactly. 

Another approach could be, moving this functionality to the preset instead of having as a global settings.

**Acceptance Criteria**
- Users should be able to specify custom  MachineDeploymentVMResourceQuota on a datacenter level instead of global level.

ahmedwaleedmalik:

This PR adds the ability to configure machine flavor filters at the datacenter level based on resources like CPU, Memory, and GPU. 

Fixes #10207

**What type of PR is this?**

/kind feature

There are two points that we need to make a decision on:
1. Should we remove the global limits defined in `KubermaticSettings`?
2. IMO it's better to use path param for `datacenter name` in the endpoints since it should be a mandatory value now. Instead of passing it in the header, which is the current implementation in the PR. On the other hand, if the user didn't specify the optional value in the header then they'll get the list of instances without any filtration. Which sounds fine to me as well. 
3. We'll add documentation for this feature at https://docs.kubermatic.com against the PR for the front-end.

```release-note
Support to filter machines based on resources(CPU, RAM, GPU) per datacenter
```

**Documentation**:

```documentation
NONE
```


lsviben:
about :
`Should we remove the global limits defined in KubermaticSettings?` 

IMO this is still useful as it gives us a "global limit". I would consider the datacenter ones as something that helps admins fine-tune the limits, but still keep the global one as a default

lsviben:
```IMO it's better to use path param for datacenter name in the endpoints since it should be a mandatory value now. Instead of passing it in the header, which is the current implementation in the PR. On the other hand, if the user didn't specify the optional value in the header then they'll get the list of instances without any filtration. Which sounds fine to me as well.```

I would set the datacenter name in the endpoints optional not to break the API. If its not set, just use the global/default one. So IMO dont add it to the path, and make changes to make it optional.

Other option would be to create new endpoints, with datacenter name in the path. Then we can slowly deprecate the old ones. 


Otherwise the changes look ok, except for the bunch of tests failing

ahmedwaleedmalik:
/retest

@lsviben PR is ready for review now ⚡

ahmedwaleedmalik:
/retest 
Flake

ahmedwaleedmalik:
> pre-kubermatic-dualstack-e2e-aws

TestCloudClusterIPFamily/c-aws-canal-IPv4+IPv6 🤷‍♂️

PR #11131, opened 2022-10-04
bump Dex to 2.35.1

xrstf:
An update for all Dex <= 2.34 versions is recommended: https://github.com/dexidp/dex/security/advisories/GHSA-vh7g-p26c-j2cw

Fixes #

**What type of PR is this?**
/kind chore

```release-note
Update Dex to 2.35.1 (contains [security fix](https://github.com/dexidp/dex/security/advisories/GHSA-vh7g-p26c-j2cw))
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/charts/oauth/Chart.yaml
+++ b/charts/oauth/Chart.yaml
@@ -15,7 +15,7 @@
 apiVersion: v1
 name: oauth
 version: v9.9.9-dev
-appVersion: v2.32.0
+appVersion: v2.35.1
 description: Dex
 keywords:
   - kubermatic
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

xrstf:
Even though I tried to avoid it by running the api-e2e test, this upgrade seems to have broken for us. Most likely because our dev environment runs on GKE (https://github.com/dexidp/dex/issues/2676).

I personally would simply revert this and wait for an upstream fix, rather than telling every customer to start creating a Service Account that they technically not needed before and not need now?

PR #11135, opened 2022-10-05
Do not require addons flags in `kubermatic-installer mirror-images`

embik:

The `mirror-images` subcommand requires one of the addons flags (`--addons-path` or `--addons-image`) to be set and fails otherwise. This does not make sense because the code is capable of falling back to the default image configured for addons:

https://github.com/kubermatic/kubermatic/blob/13bcaf1d655e33d517059f8cc8ea26a79124668a/cmd/kubermatic-installer/cmd_mirror_images.go#L165-L182

It is therefore not necessary to provide one of the flags. This PR removes the requirement.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
Do not require addons flags in `kubermatic-installer mirror-images` and fall back to default addons image
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
/approve

PR #11146, opened 2022-10-05
fix --config not being validated in mirror-images command

xrstf:
Since 2.21, the `--config` is optional for doing KKP upgrades, but it being nil can cause issues and I overlooked that the `mirror-images` command actually always needs a config. So this PR adds the missing validation for the CLI flag.

**What type of PR is this?**
/kind bug

```release-note bugfixes
Fix `--config` flag not being validated in `mirror-images` command in the KKP installer
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #11147, opened 2022-10-05
Improve registry overwrite mechanism

xrstf:
When mirroring images into a local registry, it's very easy to alter the RepoDigest of an image. This will lead to problems with custom registries in KKP, as an image like

```
quay.io/cilium/cilium:v1.11.10@sha256:abc123
```

gets rewritten to

```
myregistry.initech.local/cilium/cilium:v1.11.10@sha256:abc123
```

but the image on the local registry might have another repo digest.

There are supposedly ways around this ([crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane)?), but until those are implemented in KKP (in the `mirror-images` command) and to not force admins to follow a special procedure when mirroring images, this PR extends our registry overwrite mechanism.

Previously we would only take the registry itself into account, but with the new code, we look at and rewrite the entire Docker image (so `quay.io/kubermatic/foo:v1.2.3` instead of just `quay.io`). This allows the rewriting mechanism to behave smarter and -- for now -- always drop the digest if a local registry is used.

In the future we might make this configurable (e.g. adding an `keepImageDigests` option somewhere), but for now we drop the digest always when changing registries. The chances of a local registry getting poisoned with bad images is much smaller than with a public registry anyway.

The signature is `func(string) (string, error)`, which made all the creators a bit more tedious to write, but my thought was that the function can be used on user-supplied input where catching the error might be important. But in all cases in KKP, the images and tags are basically hardcoded, so for now I added a `Must()` helper to panic in cases that the hardcoded values are malformed.

**What type of PR is this?**
/kind bug
/kind feature

```release-note
Repo Digests are now dropped when a docker image is changed by using the overwrite-registry mechanism. Previously kept digests could lead to issues with mirrored Docker images.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/addons/csi/digitalocean/csi-driver.yaml
+++ b/addons/csi/digitalocean/csi-driver.yaml
@@ -64,7 +64,7 @@ spec:
           type: RuntimeDefault
       containers:
         - name: csi-provisioner
-          image: '{{ Registry "registry.k8s.io" }}/sig-storage/csi-provisioner:v3.2.1'
+          image: '{{ Image "registry.k8s.io/sig-storage/csi-provisioner:v3.2.1" }}'
           args:
             - "--csi-address=$(ADDRESS)"
             - "--default-fstype=ext4"
@@ -77,7 +77,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-attacher
-          image: '{{ Registry "registry.k8s.io" }}/sig-storage/csi-attacher:v3.5.0'
+          image: '{{ Image "registry.k8s.io/sig-storage/csi-attacher:v3.5.0" }}'
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
@@ -91,7 +91,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-snapshotter
-          image: '{{ Registry "registry.k8s.io" }}/sig-storage/csi-snapshotter:v6.0.1'
+          image: '{{ Image "registry.k8s.io/sig-storage/csi-snapshotter:v6.0.1" }}'
           args:
             - "--csi-address=$(ADDRESS)"
             - "--v=5"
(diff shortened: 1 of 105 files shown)


addons/csi/digitalocean/csi-driver.yaml
```diff
@@ -118,7 +118,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: csi-do-plugin
-          image: '{{ Registry "docker.io" }}/digitalocean/do-csi-plugin:v4.2.0'
+          image: '{{ Image "digitalocean/do-csi-plugin:v4.2.0" }}'
```
xmudrii, line 121: Why don't we keep `docker.io` here?

xmudrii (commented):
Looks great! Just one question, it affects some other addons as well.

pkg/controller/user-cluster-controller-manager/resources/resources/envoy-agent/daemonset.go
```diff
@@ -147,14 +159,14 @@ func getInitContainers(ip net.IP, versions kubermatic.Versions, registryWithOver
...
+	}, nil
 }
 
-func getContainers(versions kubermatic.Versions, registryWithOverwrite registry.WithOverwriteFunc) []corev1.Container {
+func getContainers(versions kubermatic.Versions, imageRewriter registry.ImageRewriter) ([]corev1.Container, error) {
 	return []corev1.Container{
 		{
 			Name:            resources.EnvoyAgentDaemonSetName,
-			Image:           fmt.Sprintf("%s/%s:%s", registryWithOverwrite(resources.RegistryDocker), envoyImageName, versions.Envoy),
+			Image:           registry.Must(imageRewriter(fmt.Sprintf("%s:%s", envoyImageName, versions.Kubermatic))),
```
xmudrii: Why is this changed from `versions.Envoy` to `versions.Kubermatic`?

addons/csi/digitalocean/csi-driver.yaml
xrstf: Because the function will prepend it by default anyway. To me the practice of prepending it manually was always a KKP-ism, necessary for easy search&replace registry changes. But with the new code, it's not necessary anymore, so I generally removed the docker.io/ prefix.

pkg/controller/user-cluster-controller-manager/resources/resources/envoy-agent/daemonset.go
xrstf: Ouh that was a copy-paste mistake.

xmudrii (approved):
/lgtm
/approve

PR #11156, opened 2022-10-07
Add project-aware cloud provider endpoints for GKE

### LINKED ISSUE #11163: Add project-scoped API endpoints for GKE
As part of #7260, we need to provide KKP API endpoints that are "project-aware", e.g. from `/v2/providers/gke/sizes` to `/v2/projects/{project_id}/providers/gke/sizes` because those API endpoints can use `Presets` as credentials, and we need information about the project context to determine whether a project-scoped preset can be used or not.

## Acceptance Criteria

- All provider-specific API endpoints for GKE gain a project-specific equivalent.

embik:

This PR adds new API endpoints for GKE that are project-aware, which means they are capable of using project-scoped presets (using a `Preset` with project restrictions via the existing API endpoints is not possible).

Fixes #11163

**What type of PR is this?**
/kind feature
/kind api-change

```release-note
Add API endpoints for GKE that allow using project-scoped Presets as credentials
```

**Documentation**:

```documentation
NONE
```


pkg/handler/v2/external_cluster/gke.go
```diff
@@ -1025,12 +1136,13 @@ func getSAFromPreset(ctx context.Context,
 	userInfoGetter provider.UserInfoGetter,
 	presetProvider provider.PresetProvider,
 	presetName string,
+	projectID string,
 ) (string, error) {
 	userInfo, err := userInfoGetter(ctx, "")
```
lsviben: use the projectID here as well, its needed to get the proper roles for the project. 

Although we dont use impersonation client (which uses the roles) to get the presets, better to add it then later look for bugs if the getting Preset implementation changes

lsviben (commented):
just a suggestion

pkg/handler/v2/external_cluster/gke.go
embik: As you said it doesnt affect anything right now but I totally agree, if we have the information why not pass it along. Added.

embik:
/retest

failed test looks unrelated.

PR #2320, opened 2018-11-01
Move metrics server to seed

mrIncompetent:
**What this PR does / why we need it**:
This PR moves the metrics server into the seed cluster.

Fixes #2125

**Release note**:
```release-note monitoring
[ACTION REQUIRED] move the metrics-server into the seed cluster. The metrics-server addon must be removed from the list of addons to install.
```

PR #11184, opened 2022-10-08
remove metrics-server addon

xrstf:
Back in #2320 we moved the metrics-server from the usercluster into the seed. We left the addon around for a bit, but by now it's horribly outdated and should just be removed alltogether.

/kind cleanup

```release-note
Remove `metrics-server` addon. The addon was deprecated in v2.9 and should not be part of any usercluster anymore.
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #11190, opened 2022-10-10
Fix Seed-Proxy ServiceAccount token not being generated

xrstf:
Since v1.24, Kubernetes does not auto-generate token Secrets for ServiceAccounts, but instead only generates tokens "on the fly" when a pod is actually mounting the ServiceAccount.

Since for the seed-proxy, we do not actually mount the SA anywhere, we do not get any tokens for it anymore, leading the master-ctrl-mgr to log

> {"level":"error","time":"2022-10-10T09:19:00.192Z","caller":"controller/controller.go:273","msg":"Reconciler error","controller":"kkp-seed-proxy-controller","object":{"name":"kubermatic","namespace":"kubermatic"},"namespace":"kubermatic","name":"kubermatic","reconcileID":"0bdd4a78-75ac-4bde-9a24-0a98d5418fda","error":"failed to reconcile: failed to fetch ServiceAccount: no Secret associated with ServiceAccount 'kubermatic/seed-proxy'"}

This PR fixes that bug by creating the necessary Secret ourselves so that Kubernetes will then fill-in the token. This PR also removes some migration code for when we moved resources around. That migration has happened like in 2.15 or so, a long time ago.

/kind bug
/kind regression

```release-note bugfixes
Fix Seed-Proxy ServiceAccount token not being generated
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/controller/master-controller-manager/seed-proxy/controller.go
+++ b/pkg/controller/master-controller-manager/seed-proxy/controller.go
@@ -63,6 +63,10 @@ const (
 	// inside the seed cluster.
 	SeedServiceAccountName = "seed-proxy"
 
+	// SeedSecretName is the name used for service accounts
+	// inside the seed cluster.
+	SeedSecretName = "seed-proxy-token"
+
 	// SeedMonitoringNamespace is the namespace inside the seed
 	// cluster where Prometheus, Grafana etc. are installed.
 	SeedMonitoringNamespace = "monitoring"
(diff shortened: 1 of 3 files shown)


vgramer (approved):
/lgtm

xrstf:
@vgramer Want to take another look? Now the PR should work :grin:

vgramer (approved):
/lgtm

PR #11198, opened 2022-10-11
Simplify expose-strategy tests

xrstf:
This PR slightly improves the expose strategy tests, which previously used a "must work N out of M times" approach, which I found weird and I could not see a reason for that in the original PR (#6745).

The PR changes the testing logic to just retry until it works, making the test a bit more flexible and robust against slow pod reconciling. I also found a way to get rid of the quite hackish `sed` usage to override the Kubernetes API server address by just using the `--server` flag. Also the error messages have been improved a bit, thanks to our polling functions.

/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/test/e2e/expose-strategy/client.go
+++ b/pkg/test/e2e/expose-strategy/client.go
@@ -17,14 +17,17 @@ limitations under the License.
 package exposestrategy
 
 import (
+	"context"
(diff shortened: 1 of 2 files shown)


vgramer (approved):
/lgtm

PR #11204, opened 2022-10-13
Add annotation to pause any resource updates

LittleFox94:

Add annotation on objects to pause any KKP automations on them, preventing them from being updated.

**What type of PR is this?**
/kind feature

There was a short discussion in Slack if this makes sense at all, stating the use case for that: https://kubermatic-community.slack.com/archives/C0PGCH99P/p1665649627499519

```release-note
Allow pausing resource updates by Kubermatic controllers by adding a special annotation
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/resources/reconciling/ensure.go
+++ b/pkg/resources/reconciling/ensure.go
@@ -34,6 +34,13 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// ResourceReconciliationPausedAnnotation is the name of the annotation to set to "true" to stop any
+	// automation from modifying it.
+	// This is only meant for development and should never be used in a production(-like) system.
+	ResourceReconciliationPausedAnnotation = "hacking.k8c.io/pause"
+)
+
 //go:generate go run ../../../codegen/reconcile/main.go
 
 // ObjectCreator defines an interface to create/update a ctrlruntimeclient.Object.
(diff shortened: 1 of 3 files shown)


LittleFox94:
This currently does not affect deleting resources in any way, didn't find a good place to add this.

mfranczy:
It's worth to document this annotation for devs, please do so here https://github.com/kubermatic/kubermatic/blob/master/docs/dev-getting-started.md

pkg/resources/reconciling/ensure.go
```diff
@@ -34,6 +34,12 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// ResourceReconciliationPausedAnnotation is the name of the annotation to set to "true" to stop any
```
xrstf, line 38: We should clearly document that doing so will "void any kind of warranty" (not the wording we should necessarily use, but it's the intent) and that there is no support for systems that make use of this hack. I can already imagine the horrors of debugging for 3 days only to find a stray annotation somewhere. 🙈  (though the new warning log message that you put in place already helps a lot!)

LittleFox94: extended that comment with a note for that and also added that to the new development docs section

xrstf (approved):
/approve

Let's see how this works out :)

LittleFox94:
> /approve
> 
> Let's see how this works out :)

I hope not problematic :upside_down_face:

PR #11221, opened 2022-10-20
Prepare scripts for automated deployment of UserCluster MLA on dev

wurbanski:
Implements changes required to automate UserCluster MLA deployments on dev cluster. Part 1 of 2 (TBD: infra PR)

Fixes #

**What type of PR is this?**
/kind feature

Needs the accompanying PR in infra (TBD once we settle if this approach is correct)

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/hack/ci/deploy.sh
+++ b/hack/ci/deploy.sh
@@ -123,7 +123,7 @@ usercluster-mla)
     --config "$KUBERMATIC_CONFIG" \
     --helm-values "$VALUES_FILE" \
     --mla-include-iap \
-    --timeout=1500s
+    --timeout=30m
   ;;
 
 kubermatic)
--- a/hack/ci/deploy.sh
+++ b/hack/ci/deploy.sh
@@ -117,6 +117,15 @@ logging)
   deploy "promtail" "logging" charts/logging/promtail/
   ;;
 
+usercluster-mla)
+  echodate "Running Kubermatic Installer for UserCluster MLA..."
+  ./_build/kubermatic-installer deploy usercluster-mla \
+    --config "$KUBERMATIC_CONFIG" \
+    --helm-values "$VALUES_FILE" \
+    --mla-include-iap \
+    --timeout=1500s
+  ;;
+
 kubermatic)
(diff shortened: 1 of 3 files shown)


wurbanski:
@xrstf can you take a look at this? I don't want to build entire KKP in this, but I need the installer - is this a good approach here or should I do it differently?

hack/ci/deploy.sh
```diff
@@ -117,6 +117,15 @@ logging)
   deploy "promtail" "logging" charts/logging/promtail/
   ;;
 
+usercluster-mla)
+  echodate "Running Kubermatic Installer for UserCluster MLA..."
+  ./_build/kubermatic-installer deploy usercluster-mla \
+    --config "$KUBERMATIC_CONFIG" \
+    --helm-values "$VALUES_FILE" \
+    --mla-include-iap \
+    --timeout=1500s
```
xrstf: Can we please use a more descriptive timeout, like "1500000000000us"? Or what about simply "25m"?

xrstf:
I don't quite understand the issue here. You want to use the installer, but not build it? And is this PR done or are you still working on it?

hack/ci/deploy.sh
wurbanski: it's just `1500000000us` - your proposal would take much too long IMO

wurbanski:
I just didn't know if I should rebuild the installer in the MLA job or is it possible to just use the binary from somewhere else. If I want to rebuild it here before I run it, I will need more resources than if I were only to install. Thought that maybe you would have some idea of how the entire system should run optimally resource-wise, but if it's fine to rebuild it, then we can go with this. 

I used WIP but I meant to just hold it.

/hold

xrstf:
Yeah to be on the safe side you have to rebuild the installer. We generally try to keep Prowjobs self contained and not have dependencies between jobs, plus just compiling the amd64 binaries should be somewhat "quick" (no need to start compiling the arm64 binaries like the user-ssh-keys-agent, which would take ages) :)

xrstf (approved):
/approve

PR #11226, opened 2022-10-22
Remove redundant E2E jigging code in CCM Migration tests

xrstf:
The CCM tests have their own code to define how clusters are created, waited for, filled with machines etc. -- basically replicating what pkg/test/e2e/jig/ is attempting to do for all e2e tests.

This PR refactors the CCM tests to use the existing jigs and extends the existing jigs to handle the new providers.

### Part 1: New `jig.*Credentials` structs

Previously the e2e tests had CLI flags for all provider credentials (like `-aws-access-key-id`) and they all kept the credentials in separate global variables. Some tests even made use of presets and the ccm-migration tests created credential secrets (something the cluster-credentials-controller would do based on the cluster spec).

To simplify the code a bit, new structs for provider credentials were introduced in the `jig` package, which are based on environment variables instead of CLI flags. This is to prevent leaking vars when viewing processes.

The choice of datacenter was also moved into the credentials, so the singular `-kkp-datacenter` flag was replaced with `-[provider]-kkp-datacenter`. This is because the ccm-migration test handles multiple providers and it would be ugly to squeeze all that into a single CLI flag. It also makes the code a bit neater as the info about the chosen datacenter doesn't have to be passed along separately anymore, but instead it's contained in the `*Credentials` struct.

Half of this PR is therefore just adjusting all the e2e tests the same way: remove the custom credential stuff and replace it with the appropriate struct from the `jig` package. This applies to all e2e tests except the ccm-migration tests, which got even more refactoring.

### Part 2: Extending provider support in Jigs

The MachineJig now supports Azure, Openstack, vSphere and Hetzner. Sadly this means lots of code duplication from `pkg/resources/machine/common.go`'s logic to default the MachineSpecs, but the `pkg/resources/machine` stuff depends on `/pkg/api/` and since we're in the process of removing the API-related code, I do not want to create further dependencies to it.

Once the API gets removed, I can imagine the code from the jigs to be "promoted" to standalone helpers to assemble MachineSpecs (i.e. refactor `pkg/resources/machine` to basically be the code of `pkg/test/e2e/jig/machine.go`).

### Part 3: Refactor ccm-migration tests

With parts 1 and 2 in place, the ccm-migration tests were then massively thinned out. All the cluster creation, waiting, machine stuff code was removed and it is now all based on the `pkg/test/e2e/jig` code. For each cloud provider, basically only the single call to `jig.New[Provider]Cluster` and a `CheckComponents()` remain.

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


embik (approved):
/approve

PR #11251, opened 2022-10-28
Don't update version of mla-secrets chart automatically

wurbanski:
This chart renders secrets randomly if they are not set and we don't want it to change uncontrollably, hence we keep specific version to not force an update.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xrstf (approved):
/approve

Issue #11224, opened 2022-10-21
Remove API's code from kubermatic repository


pkprzekwas:
Once API code and automations are migrated to dashboard's repository we have to remove remaining API code from kubermatic repository.

PR #11258, opened 2022-10-29
Remove kubermatic-api cmd, apiclient and related scripting

xrstf:
This is the first, easy part of removing the KKP API codebase: Just deleting the `cmd`, the `apiclient` and stripping down all the scripts and stuff. Disecting the handlers and providers is a bit more work and left for the next PR after this one.

/kind cleanup

Part of #11224.

```release-note
Remove `kubermatic-api` cmd, `swagger.json` and the generated API client after the API-related code was moved into the dashboard repository.
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve 💯 
/hold not sure if you need a dedicated review from the folks who worked on API deprecation

PR #11265, opened 2022-10-31
Remove unused kubernetes providers

xrstf:
This is part 3 of #11224. After the basic command was removed in #11258, then the API handlers got removed in #11262, this PR now goes deeper and removes most of the kubernetes providers.

In general, providers should only have ever been used inside the API, but some controllers started to use them as well. The abstraction that the providers ... provide (sigh) do generally not seem that useful to us and we rather want to have some helper functions to collect shared code. That's why the provider usage inside controllers is something to refactor over time.

This PR ...

* gets rid of all kubernetes providers
  * except for ExternalCluster, which is the single base codebase for the external cluster controller
  * seed and kubermaticconfig getters will remain in the KKP codebase
* removes the Machine conversion code (a better version of this is going to be implemented based on the new tests jigs)
* fixes the e2e cluster jig, which was based on the cluster provider (what was I thinking?!)

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


xrstf:

vCenter is offline again.

PR #11267, opened 2022-11-01
Fix: DNAT controller iptables-restore call blocking for big user clusters

### LINKED ISSUE #9995: DNAT fails to install iptables rules in big clusters

xrstf:
This adopts the changes from #9996 after the original author has seemingly abandoned the PR.

Fixes https://github.com/kubermatic/kubermatic/issues/9995

**What type of PR is this?**
/kind bug

```release-note
Fix: DNAT controller doesn't install NAT rules for big clusters
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/controller/kubeletdnat-controller/controller.go
+++ b/pkg/controller/kubeletdnat-controller/controller.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve ⚡ 
/retest

PR #11271, opened 2022-11-01
remove unneeded and outdated addon testdata

xrstf:
As evidenced by the branch name of this PR, this started with the intent to add _more_ testdata to `pkg/addon/`, because as of now we only tested AWS and Digitalocean clusters. But then I realized that instead of collecting test YAML data files, we could just generate clusters with various combinations of features in code and then I remembered that this is exactly what the `pkg/install/images/` tests are already doing.

So I removed the needless (and outdated) testdata files and left comments explaining how the unit tests relate.

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #11279, opened 2022-11-02
Consolidate cloud provider code

xrstf:
The old KKP API had cloud provider code in it (i.e. in `pkg/handler/`) to validate credentials, list VM sizes etc. in order to prepopulate the dropdowns in the KKP dashboard. While the functions to list sizes have already been removed, the remaining cloud functions in `pkg/handler/` are used by the EE ResourceQuota webhook and so are still required in KKP.

This PR moves them into the "big" provider packages (i.e. move `pkg/handler/common/cloud/aws.go` to `pkg/provider/cloud/aws/`) and cleans the code up a bit by introducing a helper struct (`provider.NodeCapacity`) primarily to get rid of things like `apiv2.AWSSize`, as `pkg/api/` is also on its way out.

I also think Nutanix was broken, because it did `fmt.Sprintf("%d%s", ...)`, but the first value was a pointer, which should always have produced invalid output.

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


xrstf:
@pkprzekwas also need an /lgtm 😁

PR #11281, opened 2022-11-02
Update Go dependencies

xrstf:
It's that time again to update all of our Go dependencies to the latest versions.

* etcd has not been bumped so it stays in-sync with the version we deploy into clusters (3.5.4)
* OPA has not been bumped because that often causes trouble down the line
* The old `...Ptr()` functions are now deprecated and had to be replaced to appease golangci-lint.

/kind chore

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve ⚡

PR #11286, opened 2022-11-03
remove support for Kubernetes 1.22 user clusters

xrstf:
Kubernetes 1.22 went End-of-Life on October 22nd and so this PR removes support for creating new clusters using 1.22. As usual manifests and code to _reconcile existing_ clusters are kept so that those can be migrated to 1.23 and don't just suddenly stop working. Upgrades are not enforced though, users have to do it themselves.

The web-terminal image has been built and pushed already.

**What type of PR is this?**
/kind chore

```release-note
Remove support for creating new Kubernetes 1.22 user clusters
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (commented):
/approve

PR #11298, opened 2022-11-07
Support for OpenStack server groups

### LINKED ISSUE #8827: Add servergroup to openstack machine deployment
Starting with machine-controller 1.42.2 the machine controller can handle server group, which are needed to set (soft)(anti)affinity.

This should be possible from the UI/API. 

The only difference to other components like secgroup and network, the default should stay none.
I am not sure if we need to be able to use pre-exsiting ones, or just a drop down for all 5 modes are enough.

Possible modes would be:

- none
- affinity
- soft affinity
- anti affinity
- soft anti affinity

ahmedwaleedmalik:

Support for OpenStack server group in initial machine deployments.

Fixes #8827 
xref https://github.com/kubermatic/dashboard/pull/5201

**What type of PR is this?**

/kind feature

```release-note
Support for using server groups with OpenStack
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/api/v1/types.go
+++ b/pkg/api/v1/types.go
@@ -1720,6 +1720,9 @@ type OpenstackNodeSpec struct {
 	// Max time to wait for the instance to be ready, i.e. 10s/1m
 	// required: false
 	InstanceReadyCheckTimeout string `json:"instanceReadyCheckTimeout"`
+	// UUID of the server group, used to configure affinity or anti-affinity of the VM instances relative to hypervisor
+	// required: false
+	ServerGroup string `json:"serverGroup"`
 }
 
 func (spec *OpenstackNodeSpec) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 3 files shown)


pkg/api/v1/types.go
```diff
@@ -1720,6 +1720,9 @@ type OpenstackNodeSpec struct {
 	// Max time to wait for the instance to be ready, i.e. 10s/1m
 	// required: false
 	InstanceReadyCheckTimeout string `json:"instanceReadyCheckTimeout"`
+	// UUID of the server group, used to configure affinity or anti-affinity of the VM instances relative to hypervisor
+	// required: false
+	ServerGroup string `json:"serverGroup"`
```
embik, line 1725: Should this (and the struct field later on) be tagged `omitempty` maybe, so it's not included in JSON if it's not set?

ahmedwaleedmalik: It wouldn't really matter since in both cases the field is initialized as empty in the MachineDeployment.

embik (approved):

embik (approved):

PR #11300, opened 2022-11-07
do not require go-openapi just to handle aggregate errors

xrstf:
We already have a package that can handle multiple errors and that package does not pretent that those errors are all validation related. So I think we can just use one over the other, as ultimately the aggregated errors are just logged and nothing depends on them being a go-openapi Composite Error.

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/.wwhrd.yml
+++ b/.wwhrd.yml
@@ -27,7 +27,6 @@ allowlist:
   - ISC
 
 exceptions:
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
/approve

